PR #427, opened 2018-06-18
Typo in JS/TS example code

twpol:
Pretty sure the code should be `new` here. :)


joeduffy:
Indeed, `ew` is not yet an operator, as far as I know ... but it should be a way!  Perhaps to comment out code smells?

In any case, thank you @twpol for the fix.  Merging.

PR #430, opened 2018-06-18
Explain in a little detail what the easy install scripts do

ellismg:
Could use your help wordsmithing this, @lindydonna.


joeduffy:
I would try to get it down to two sentences max.  (Parenthetical remarks count as sentences.)

lindydonna:
@joeduffy @ellismg I added another commit with a proposed change. I consolidated the note to cover all platforms.

lindydonna:
@joeduffy It's 3 sentences now, but one is very short. :)

PR #435, opened 2018-06-19
Build the search index ahead of time during site generation

justinvp:
Follow-up from #419: Build the search index ahead of time during site generation.

Also, stop intercepting the form submit event when on `/search.html` (to avoid subsequent GET requests). It's simpler to just always make a new GET request to `/search.html` each time a search is done. This addresses https://github.com/pulumi/home/issues/309.

### DIFF
--- a/js/search-worker.js
+++ b/js/search-worker.js
@@ -2,38 +2,18 @@
 
 importScripts("/js/lunr.min.js");
 
-var data = {};
 var index;
 
-// Download the search data synchronously.
+// Download the search index synchronously.
 var req = new XMLHttpRequest();
 var async = false;
-req.open("GET", "/search-data.json", async);
+req.open("GET", "/search-index.json", async);
 req.send();
 if (req.readyState === req.DONE && req.status === 200) {
-    data = JSON.parse(req.responseText);
-    loadIndex();
+    var data = JSON.parse(req.responseText);
+    index = lunr.Index.load(data);
 } else {
-    console.log("problem downloading search-data.json");
-}
-
-function loadIndex() {
-    index = lunr(function() {
-        this.ref("id");
-        this.field("title", { boost: 10 });
-        this.field("content");
-
-        for (var key in data) {
-            var item = data[key];
-            if (item.title || item.content) {
-                this.add({
-                    id: key,
-                    title: data[key].title,
-                    content: data[key].content
-                });
-            }
-        }
-    });
+    console.log("problem downloading search-index.json");
 }
 
 function search(query) {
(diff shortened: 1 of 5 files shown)


js/search-worker.js
```diff
@@ -42,11 +22,11 @@ function search(query) {
...
             return results.map(function(result) {
-                return {
-                    score: result.score,
-                    url: result.ref,
-                    title: data[result.ref].title
-                };
+                // Extract the url and title from the ref string (they're separated by "|").
+                var i = result.ref.indexOf("|");
+                var url = result.ref.substring(0, i);
+                var title = result.ref.substring(i + 1);
```
chrsmith: Is this going to cause a problem if there is a page somewhere without a title? Is it worth checking the length of `result.ref`?

js/search-worker.js
```diff
@@ -2,38 +2,18 @@
 
 importScripts("/js/lunr.min.js");
 
-var data = {};
 var index;
 
-// Download the search data synchronously.
+// Download the search index synchronously.
 var req = new XMLHttpRequest();
 var async = false;
```
chrsmith: Nit: Given that `async` is a keyword in TypeScript, which we use in a lot of places, perhaps `doAsync` would be a better variable name? Alternatively, since it is only used in one place, perhaps just `false /* async */` would be sufficient?

scripts/build-search-index.js
```diff
@@ -0,0 +1,42 @@
...
+        this.ref("ref");
+        this.field("title", { boost: 10 });
+        this.field("content");
+
+        for (var url in data) {
+            const page = data[url];
+            if (page.title || page.content) {
+                // We use '|' to separate the url and title in the ref, so
+                // ensure we haven't inadvertently introduced a page that
+                // contains a '|' in its url.
```
chrsmith: Nit: URL, all caps.

js/search-worker.js
justinvp: We do have pages that don't have titles and this code works for those cases. JavaScript's [`substring`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) is pretty forgiving -- if you pass an index greater than or equal to the string length it returns an empty string, e.g. all the following cases return `""`:

```javascript
"/url|".substring(5);
"/url|".substring(6);
"/url|".substring(100);
```

I did add a check that throws if `"|"` isn't found.

js/search-worker.js
justinvp: Done.

scripts/build-search-index.js
justinvp: Done.

PR #436, opened 2018-06-19
Use friendlier clouds page for 404

### LINKED ISSUE #433: Use friendlier clouds page for 404s (and address the logo, font, etc.)
Currently:

screen shot 2018-06-18 at 3 44 46 pm

justinvp:
Use the friendlier clouds page for 404 and fix the logo and fonts. Also remove welcome.html as we don't need it anymore.

Before:

screen shot 2018-06-18 at 3 44 46 pm

After:

screen shot 2018-06-19 at 4 02 18 pm

Fixes #433


justinvp:
@chrsmith, after I made these changes I discovered the nicer way you abstracted things for the 404/500 pages for the WWW site. I considered refactoring to do something similar here, but didn't want to spend too much more time on it. If we need a 500 error page here, I'll do that then.

search-data.json
```diff
@@ -18,8 +18,7 @@ Exclude some pages that we don't want to index.
   page.url == '/packages/pulumi-kubernetes/' or
   page.url == '/redirects.json' or
   page.url == '/search-data.json' or
-  page.url == '/search.html' or
-  page.url == '/welcome.html' -%}
+  page.url == '/search.html' -%}
```
chrsmith: Should this be in the current PR, or is this a relic from a previous commit on your machine?

justinvp: This PR deletes `/welcome.html`, so it’s no longer needed here.

PR #444, opened 2018-06-22
Add page titles

### LINKED ISSUE #438: The title of every page is "Pulumi Documentation"
Also, when emitting the title on the page, we always emit the `<h1>{{ page.title }}</h1>` even if the page doesn't have a title (some pages don't, like the CLI documentation).

justinvp:
Use the page's title for the document title. Also, if the page has no title, don't emit the `<h1></h1>` on the page.

Fixes #438

### DIFF
--- a/index.md
+++ b/index.md
@@ -1,5 +1,4 @@
 ---
-title: Welcome to Pulumi
 layout: default_index
 ---
 
--- a/404.html
+++ b/404.html
@@ -1,4 +1,5 @@
 ---
+title: Page Not Found
 layout: 404
 ---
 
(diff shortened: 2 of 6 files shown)


index.md
```diff
@@ -1,5 +1,4 @@
 ---
-title: Welcome to Pulumi
```
justinvp: This way, the home page's title will be "Pulumi" instead of "Welcome to Pulumi - Pulumi".

joeduffy: We should probably match the www site for title here. This is very important to nail for SEO.

justinvp: Yes, this does match the www title here. It's just "Pulumi".

justinvp:
> Do we need to do a pass over content files and confirm we are setting title in our frontmatter?

If I recall correctly, the last time I looked (as part of the search work) we were setting the title everywhere except the CLI command docs (as those markdown files are generated without the frontmatter) and some manual redirect pages where it won't matter for this. But I'll do another pass.

PR #446, opened 2018-06-25
Add FAQ

ellismg:
Add a FAQ with some questions we've heard over the past week since launch.

### DIFF
--- a/_data/reference.yaml
+++ b/_data/reference.yaml
@@ -72,4 +72,5 @@ toc:
   path: /reference/cd
   section:
   - reference/cd-travis.md
+- reference/faq.md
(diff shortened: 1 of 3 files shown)


joeduffy:
It's probably worth adding anchors so we can link to specific answers when someone asks e.g. in Slack

ellismg:
> It's probably worth adding anchors so we can link to specific answers when someone asks e.g. in Slack

It looks like the system already generates the anchors, but there is more work to be done to get the nice experience where when you hover over the header an anchor icon displays.  I spent a few mins fighting with doing this to no real success, so I'll and what we have so we can at least start pointing folks on the slack to the page.

PR #460, opened 2018-06-29
Add anchor tags to our headers

### LINKED ISSUE #454: Add anchor tags
Our markdown generator already adds an anchor tags (which is great!) but the only way to get at the values is to view source and build the URL yourself.

It would be great if we could have little on hover anchors using anchor-js, similar to what is described here: https://milanaryal.com.np/adding-hover-anchor-links-to-header-on-github-pages-using-jekyll/

ellismg:
This uses anchorjs to include anchor tags on all of our header
elements.  You get the nice hover behavior where if you hover over a
header, a little anchor link pops up and you can right click it and
copy the link and then share it.

There were a handfull of places where we explicity used header
elements and we didn't actually want links. In these places, I've
applied the `no-anchor` class and instructed anchorjs not to decorate
headers with this class.

It's possible that there are few other places we'll want to do this,
but from my grepping it looked like this would be sufficent. If we
find more places, we can just assign this class by hand.

Fixes #454

### DIFF
--- a/404.html
+++ b/404.html
@@ -32,7 +32,7 @@ layout: 404
 </div>
 
 <div id="clouds-and-beams">
-    <h1>404</h1>
-    <h2>Sorry, the page you're looking for could not be found.</h2>
+    <h1 class="no-anchor">404</h1>
+    <h2 class="no-anchor">Sorry, the page you're looking for could not be found.</h2>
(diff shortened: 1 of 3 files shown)


ellismg:
In case it was not obvious, this change is also applied to all our package documentation which is actually pretty sweet, because now we can easily link to specific members in slack.

Issue #466, opened 2018-07-02
Sitemap needed


mehzer:
We should connect .io to Google Webmaster tools, ensure we have an up-to-date Sitemap and have it submitted etc. 

Probably through use of a Jekyll plugin to generate.

chrsmith:
Closing out, as it looks like this was fixed in the linked PR. https://pulumi.io/sitemap.xml returns what appears to be a sitemap.

PR #467, opened 2018-07-02
Added Sitemap plugin and config

mehzer:
Addresses #466 by using https://github.com/jekyll/jekyll-sitemap 
Adjusted _config to add url: and the plugin reference

Produces pulumi.io/sitemap.xml


_config.yml
```diff
@@ -8,6 +8,9 @@
 # jekyll-redirect-from, and we don't need those to be fully qualified.
 pulumi_url: https://pulumi.io
 
+# Needed for jekyll-sitemap
+url: "https://pulumi.io"
```
justinvp: Now that we need this for the sitemap, would you mind adding a commit that removes the `pulumi_url` variable above along with the comment, and replace it's use in `head.html` (e.g. `site.pulumi_url` => `site.url`)?

_config.yml
```diff
@@ -18,6 +21,7 @@ plugins:
   - jekyll-redirect-from # for 301 redirects
   - jekyll-relative-links # to convert internal MD links to HTML
   - jekyll-optional-front-matter # for Markdown without front matter (e.g., CLI reference)
+  - jekyll-sitemap # produce a sitemap.xml file for submission to search engines
```
justinvp: We should exclude `sitemap.xml` from the search index by adding a check to `search-data.json`.

_config.yml
mehzer: Made this change

_config.yml
mehzer: Made this change

mehzer:
Fixed per your feedback. Not sure on other exclusions at this time.

PR #485, opened 2018-07-16
Updated cd-travis.md

Jtango18:
- Removed `pulumi init` call, which fails since this command was removed in #0732b05
- Ensure update-stack.sh is executable as part of pre-init(this may be out of scope for this documentation, but is useful in getting new users up and running quickly)
- Add detail of where to find Pulami Access Token - this doesn't seem to appear in any current documentation.

### DIFF
--- a/reference/cd-travis.md
+++ b/reference/cd-travis.md
@@ -19,24 +19,25 @@ which Pulumi stacks.  You will want to customize this.
 
 First, add your Pulumi access token to your Travis settings as the `PULUMI_ACCESS_TOKEN` environment variable, either
 via the `travis.yml` file, Travis CLI, or the Travis Web UI, however you typically do it ([see here for more
-information](https://docs.travis-ci.com/user/environment-variables/)).  This ensures that you can login to the service.
+information](https://docs.travis-ci.com/user/environment-variables/)).  This ensures that you can login to the service. Your Pulumi Access Tokens are managed thorugh your account, and can be found [here](https://app.pulumi.com/account/tokens)
 
 > **Note:** You'll also need to configure your cloud provider credentials in the same way you do on your development
 > machines.  This most often means provisioning a key specifically for CI, and then setting the environment variables so
 > that the Travis machine can perform the necessary operations.  If you are using AWS, for example, you'll want to set
 > `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`.  Please consult your cloud provider's documentation.
 
-Next, add three things to your `travis.yml` file.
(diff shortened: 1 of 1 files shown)


reference/cd-travis.md
```diff
@@ -19,24 +19,25 @@ which Pulumi stacks.  You will want to customize this.
 
 First, add your Pulumi access token to your Travis settings as the `PULUMI_ACCESS_TOKEN` environment variable, either
 via the `travis.yml` file, Travis CLI, or the Travis Web UI, however you typically do it ([see here for more
-information](https://docs.travis-ci.com/user/environment-variables/)).  This ensures that you can login to the service.
+information](https://docs.travis-ci.com/user/environment-variables/)).  This ensures that you can login to the service. Your Pulumi Access Tokens are managed thorugh your account, and can be found [here](https://app.pulumi.com/account/tokens)
```
chrsmith: Minor typeo "thorugh".

reference/cd-travis.md
```diff
@@ -19,24 +19,25 @@ which Pulumi stacks.  You will want to customize this.
...
-`$PATH`, initialize the workspace, and then login to the service:
+`$PATH`:
 
     ```yaml
     before_install:
         # Install the Pulumi SDK, add it to the $PATH, and initialize the workspace.
         - curl -L https://get.pulumi.com | bash -s -- --version 0.14.2
         - export PATH=$PATH:$HOME/.pulumi/bin
-        - pulumi init
+        # Enable update-stack script to be executable by TravisCI
```
chrsmith: Thanks for catching the `pulumi init`. As far as these new lines, I don't think they are necessary. The `update-stack.sh` file should be checked into the git repo with the `+x` permission, not have it added as part of the build process.

Calling this out is probably a good idea to avoid people getting tripped up, but perhaps we should mention it in the section above where we link to the `update-stack.sh` file in the examples repo? Something like "Be sure to run `chmod +x ./scripts/update-stack.sh` before checking in, so Travis can execute the script."

reference/cd-travis.md
```diff
@@ -19,24 +19,25 @@ which Pulumi stacks.  You will want to customize this.
...
+        - chmod +x ./scripts/update-stack.sh
```
lukehoban: This ideally shouldn't be needed inside the `.travis.yaml` as the file should be marked executable in source control.

In the example repo linked above, the `update-stack.sh` file is indeed marked executable.

Perhaps the piece that's missing is an e2e working example repo that can be cloned and used based on these directions?

lukehoban:
Thanks @Jtango18!

PR #487, opened 2018-07-19
Removed the forbidden symbol - in some variables declaration

Tirke:
Dash is not an allowed symbol in a JS variable name.


Tirke:
By digging up a little bit, I think my change will be invalided by the next doc generation.

chrsmith:
@Tirke yeah, it looks like you are right. I filed issue https://github.com/pulumi/docs/issues/488 to track fixing the source of this. (Since even if we landed the PR, it would just be overwritten the next time we render the docs.)

If you want to take a try at the Go code, feel free to take a crack at it. Otherwise I'll take a look soon.

PR #490, opened 2018-07-19
Fix tscdocgen pkgvar casing

### LINKED ISSUE #488: tscdocgen produces invalid output for package names with hyphens
The tool we use to generate documentation from our TypeScript-based Pulumi packages produces incorrect output for package names that contain invalid characters. e.g. we generate local variable names `pulumi-aws` when we should instead be using `pulumiAws`.

I believe the change needs to be made somewhere in:
https://github.com/pulumi/docs/tree/master/tools/tscdocgen

See PR https://github.com/pulumi/docs/pull/487 for an example of this causing problem in the resulting Markdown files.

Tirke:
Fixes #488 .

Disclaimer: I'm not a Go expert, I didn't want to introduce too much complexity so I did the easiest thing, so that's just adding a function in main.go. But this is a typical utils function. I also didn't make it too generic but it should handle the common use case just fine.

If this is not good, just tell me and I will update.

### DIFF
--- a/tools/tscdocgen/main.go
+++ b/tools/tscdocgen/main.go
@@ -143,7 +143,7 @@ func camelCase(s string) string {
 
 	var prev rune
 	for _, curr := range s {
-		if !(curr == '-') {
+		if curr !== '-' {
 			if prev == '-' {
 				buffer = append(buffer, unicode.ToUpper(curr))
 			} else {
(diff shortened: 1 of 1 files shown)


tools/tscdocgen/main.go
```diff
@@ -539,7 +559,7 @@ func createLabel(node *typeDocNode, parent *typeDocNode) string {
 			return fmt.Sprintf("let %s", node.Name)
 		}
 
-	// For others, we will generate a full signature.
+		// For others, we will generate a full signature.
```
lukehoban: These look like spurious new whitespace.

PR #494, opened 2018-07-25
Update changelog for v0.14.3

ellismg:
I did a walk over all of the changes in the repositories and called out what looked interesting.  Let me know if you think there's other stuff that we want to note or we want to re-word some of the detailed descriptions.


install/changelog.md
```diff
@@ -22,6 +22,11 @@ title: "Change Log"
         </tr>
     </thead>
     <tbody>
+        <tr>
+            <th scope="row"><a href="#v142">0.14.3</a></th>
```
lukehoban: `#v143`?

Issue #483, opened 2018-07-09
Add test to verify format of dashboard.json


justinvp:
If we're going to continue using dashboard.json as the source for some dashboard cards, we should consider adding a test that verifies the format to avoid accidental JSON parsing issues (such as #481).

PR #496, opened 2018-07-27
Promote to production

ellismg:


dashboard.json
```diff
@@ -46,11 +51,6 @@
                 "url": "https://pulumi.io/install/changelog.html#v141",
                 "date": "2018-06-29"
             },
```
justinvp: This trailing comma isn't valid JSON. This broke the dashboard the last time around. #483 tracks adding a test to avoid breaking this by accident.

justinvp: #497

ellismg: Thanks, Justin!

PR #507, opened 2018-08-01
Add a Troubleshooting page for dealing with broken stacks

swgillespie:
Would like to link to this from the CLI instead of spewing large amounts of text to users whenever these errors occur.

This is WIP since I can't get Jekyll working locally. I'm relying on the CI to let me preview this text.

### DIFF
--- a/reference/troubleshooting.md
+++ b/reference/troubleshooting.md
@@ -1,0 +1,62 @@
+---
+title: Troubleshooting
+---
+
+Pulumi tries very hard to ensure that your infrastructure is always in a known and predictable state.
+However, the reality is that sometimes things go wrong. If you can't update your stack, or there's
+some other problem that is preventing you from being productive with a Pulumi stack, you've come to the
+right place.
+
+## Common Problems
+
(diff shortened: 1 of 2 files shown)


reference/troubleshooting.md
```diff
@@ -0,0 +1,62 @@
...
+#### Quick Summary
+
+Run `pulumi cancel` to cancel the update.
+
+> Warning! If you cancel another person's update, their update will fail immediately.
+
+### [500] Internal Server Error
+
+The Pulumi command-line tool interacts with the Pulumi web service throughout the course of an update. If the
+service is experiencing an outage, it is possible that users of the command-line tool may see this error message
```
lukehoban: Instead of talking about an "outage", I'd just refer to "if the service is unable to process the update".  The 500 could be returned for many reasons, most of which are about isolated or transient bugs, not a full "outage".

swgillespie:
@lukehoban you're right, thanks. I was hoping that the netlify preview thing would show up in this PR, but I guess it didn't.

PR #515, opened 2018-08-11
Add docs for using AWS CodeBuild

chrsmith:
This PR describes a minimal example of updating a Pulumi stack using AWS Code Services, specifically CloudBuild.

### DIFF
--- a/reference/cd-aws-code-services.md
+++ b/reference/cd-aws-code-services.md
@@ -1,0 +1,91 @@
+---
+title: AWS Code Services
+---
+
+[Amazon Code Services](https://aws.amazon.com/products/developer-tools/) encompases a variety
+of specific tools for CI/CD, including [CodePipeline](https://aws.amazon.com/codepipeline/),
+[CodeBuild](https://aws.amazon.com/codebuild/), [CodeDeploy](https://aws.amazon.com/codedeploy/),
+and others.
+
+To incorporate updating Pulumi stacks into an AWS Code Services-managed CI/CD system, you'll
+want to use CodeBuild. Pulumi needs to execute a built program in order to determine the desired
+state of cloud resources, and CodeBuild provides a compute environment to do just that.
+
+If you are using CodePipeline, you can then create a new pipeline stage which triggers the
+CodeBuild project. Allowing you to update a Pulumi stack wherever it makes sense in your existing
+pipeline.
+
+## Configuring CodeBuild
+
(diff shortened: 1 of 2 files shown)


reference/cd-aws-code-services.md
```diff
@@ -0,0 +1,91 @@
...
+[Amazon Code Services](https://aws.amazon.com/products/developer-tools/) encompases a variety
+of specific tools for CI/CD, including [CodePipeline](https://aws.amazon.com/codepipeline/),
+[CodeBuild](https://aws.amazon.com/codebuild/), [CodeDeploy](https://aws.amazon.com/codedeploy/),
+and others.
+
+To incorporate updating Pulumi stacks into an AWS Code Services-managed CI/CD system, you'll
+want to use CodeBuild. Pulumi needs to execute a built program in order to determine the desired
+state of cloud resources, and CodeBuild provides a compute environment to do just that.
+
+If you are using CodePipeline, you can then create a new pipeline stage which triggers the
```
joeduffy: It would be cool eventually to have a Pulumi program for this!

reference/cd-aws-code-services.md
```diff
@@ -0,0 +1,91 @@
...
+
+`update_pulumi_stack.sh` is the minimal set of steps for updating a Pulumi stack.
+
+It runs `npm` commands to download the dependencies of the Pulumi program, and then builds it.
+And then uses the Pulumi CLI to select the stack and perform the update.
+
+You'll want to modify this script depending on the language used for your program, how it is
+built, etc.
+
+```bash
```
joeduffy: Should we put this in a repo instead of copy-and-pasting it into multiple places?

reference/cd-aws-code-services.md
chrsmith: Added the idea to the Hackathon Ideas wiki. This would come in handy in a lot of scenarios when using AWS.

reference/cd-aws-code-services.md
chrsmith: Since it is less than 10 lines, I'm inclined to avoid introducing a level of indirection. But if we were to create a more complicated Bash script with more bells and whistles, having a go-to example makes sense.

The problem are a lot of policy level decisions and/or configuration that needs to be provided. So just saying "use this file" probably isn't enough without some associated documentation for how to use it. IMHO that's where we went wrong with our [earlier Travis CI example](https://github.com/pulumi/examples/blob/master/misc/travis/update-stack.sh).

PR #521, opened 2018-08-13
Upgrade to 0.15.0-rc2

ellismg:


chrsmith:
The are some examples in this repo that mention a specific Pulumi version, e.g. our docs for configuring Travis CI. I assume we'll want to change those too when we ship `0.15.0` (without the `-rc2` big)?

ellismg:
> The are some examples in this repo that mention a specific Pulumi version, e.g. our docs for configuring Travis CI. I assume we'll want to change those too when we ship 0.15.0 (without the -rc2 bit)?

Yes.  I usually do that as part of adding the release notes and bumping the latest version.

PR #525, opened 2018-08-14
Update API docs for 0.15.0

lukehoban:
Update all reference API docs to 0.15.0 versions.


lukehoban:
May want to review the individual commits.  I cleaned up some of the doc generation of privates and unexported members as part of this.

PR #528, opened 2018-08-14
Amend release notes

chrsmith:
Hey @joeduffy, is this what you had in mind for adding details about changes to the Pulumi Service and the GitHub app to the v0.15.0 release notes? If so, I can merge this into the `ellismg/release-notes-0.15.0` branch.

### DIFF
--- a/install/changelog.md
+++ b/install/changelog.md
@@ -112,6 +112,22 @@ While a `tsconfig.json` file is no longer required, as Pulumi uses intelligent d
 - The CLI no longer emits warnings if it can't detect metadata about your git enlistement (for example, what GitHub project it coresponds to).
 - The CLI now only warns about adding a plaintext configuration in cases where it appears likely you may be storing a secret.
 
+### Pulumi Service
+
+Since the last release, the Pulumi Service made some key improvements around the handling of update logs. When viewing stack update logs that include the output of other
+programs, like `Docker`, the resulting logs should now render correctly as we as load much faster.
+
+The biggest new feature however, is the introduction of the new [Pulumi GitHub application](https://github.com/apps/pulumi).
+
+#### Pulumi GitHub App (preview)
+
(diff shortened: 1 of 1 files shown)


install/changelog.md
```diff
@@ -112,6 +112,22 @@ While a `tsconfig.json` file is no longer required, as Pulumi uses intelligent d
 - The CLI no longer emits warnings if it can't detect metadata about your git enlistement (for example, what GitHub project it coresponds to).
 - The CLI now only warns about adding a plaintext configuration in cases where it appears likely you may be storing a secret.
 
+### Pulumi Service
+
+Since the last release, the Pulumi Service made some key improvements around the handling of update logs. When viewing stack update logs that include the output of other
```
lukehoban: No need for "Since the last release" - that's implicit in this being a changelog.

install/changelog.md
```diff
@@ -112,6 +112,22 @@ While a `tsconfig.json` file is no longer required, as Pulumi uses intelligent d
...
+programs, like `Docker`, the resulting logs should now render correctly as we as load much faster.
```
lukehoban: `Docker` -> `docker`.

I don't think I follow this sentence - is it just a bug fix?  Do we need to include it here?

install/changelog.md
chrsmith: Done.

install/changelog.md
chrsmith: It is exactly just a bug fix, and I agree it is a stretch to add it here.

Would you prefer we promote `Pulumi GitHub App (preview)` to an L3 heading (sibling of `Pulumi CLI`) and drop the `Pulumi Service` section entirely? Given the lack of meaningful features we can talk about for the milestone, that might be the right call.

ellismg:
I'm going to merge this (it looks pretty-reasonable) so I can promote the changelog.

PR #529, opened 2018-08-14
Promote master to production

ellismg:


ellismg:
Holding this on some changes from @justinvp to update dashboard.json

PR #537, opened 2018-08-15
Production deployment

lukehoban:


chrsmith:
LGTM. Note that this PR updates the build process slightly, so we should see a PR comment with the preview against the `pulumi.io-production` stack. We also should be using `travisqueue` to run builds concurrently on travis.

PR #548, opened 2018-08-17
Fix TOC listing for reference docs

### LINKED ISSUE #470: Package reference docs now include TOC for AWS landing pages
I assume these pages used to show the standard "reference" TOC, but they now show the AWS landing pages, even for packages like `@pulumi/kubernetes`:

screen shot 2018-07-02 at 4 23 07 pm

justinvp:
No longer shows the incorrect AWS TOC for References.

This change simplifies how the current TOC is determined based on the path of the page.

Instead of listing paths to the .md files, list out the path and title for the pages explicitly. This avoids having to lookup the page in `site.pages` to get the title and path for the page, which takes forever for the reference docs, due to how many there are.

The build now runs in 40 seconds, instead of over 5 minutes it would have taken with the fixed Reference TOC.

Also fixed the FAQ and Known Issues under References to actually show the selected style when selected.

Fixes #470.


_data/reference.yaml
```diff
@@ -2,26 +2,42 @@ nav_section: "Reference"
...
+  - title: Python
+    path: /reference/python.html
+  - title: "Tutorial: Pulumi Components"
+    path: /reference/component-tutorial.html
+  - title: Terraform Providers
+    path: /reference/terraform.html
+  - title: How Pulumi Works
+    path: /reference/how.html
+  - title: Troubleshooting
+    path: /reference/troubleshooting.html
```
justinvp: It'd be nice to not have to specify the title here explicitly, but the page lookup to get the title and resulting path from the `.md` causes the build to run for over 5 minutes. With these change the build goes down to ~40 seconds on my machine, so I think it's worth manually specifying the titles here.

joeduffy: Believe it or not, I prefer listing them here explicitly. It's always been confusing to me that we had a mix of explicit and implicit, and this way you can easily see the full TOC labels in one place. Nice win on build time too 😊

Issue #550, opened 2018-08-19
Add code snippets for tutorials


joeduffy:
Our quickstart tutorials are a little awkward to use. We often tell you to "copy this code and replace the contents of `index.js` with it," but the size of code snippets can be quite large in some cases.

We should add a notion of real code snippets, similar to the kubernetes.io website:

![image](https://user-images.githubusercontent.com/3953235/44310522-c8e23780-a3a5-11e8-8393-da36bfbd3f1e.png)

Notice that there is a header complete with the name of the file and a handy little "copy to clipboard" shortcut, which makes pasting to `index.js` slightly less awkward.

I also think we should create a template for each quickstart and tell folks to use that flow instead, e.g. `pulumi new https://github.com/pulumi/...`. That way they don't even need to do the copy and paste dance. I find it cumbersome enough personally that I wouldn't be surprised if it results in dropoff.

justinvp:
Looking at the remaining work on my plate this week, it's going to be difficult to get this wrapped up by EOD Friday. Proactively moving out. Would love to spend a little time on this and some of the other pulumi.io improvements assigned to me the next milestone.

joeduffy:
Now that https://github.com/pulumi/docs/issues/1023 has taken over the "copy to clipboard" aspect of this, I think the only remaining thing to do is add a title bar for the code snippets.

@justinvp does that sound about right?

bermudezmt:
We've addressed via #1363

Issue #500, opened 2018-07-30
[Recurring] Improve onboarding docs flow


joeduffy:
The onboarding docs aren't as smooth as they could be, and suffer from a few artifacts of where we happened to be at launch, combined with vigorous and continuous refactoring leading up to launch.

A few examples

* The flow makes it look like Pulumi only works on AWS
* We plaster the `curl` command in too many places
    - often repetitively, and
    - without sufficient context for what to do next
* The Tutorials landing page awkwardly tells you to `pulumi new`

More substantively, however, we would benefit from unifying our Install, Tutorials, and Tour pages, rather than having distinct flows for them. And it'd be great if you chose up front whether you wanted to take the tour on AWS, Azure, Google Cloud, or Kubernetes.

lukehoban:
Per offline conversation - a suggestion was to make the following changes related to install steps:

- On pulumi.io, remove the curl line from the landing page, and rework the three buttons currently under it to have "Take A Tour", "Install" and "Examples" in that order (i.e. put "Take A Tour" before install, since the tour is perhaps the easiest way to get a taste of Pulumi.
- On https://pulumi.io/install/ change the text "Run the following command to install the latest version of Pulumi:" to something like "Run the following command to install the latest version of Pulumi or download and extract it manually" and that second part links to https://pulumi.io/install/#instructions, which is a little further down the page)
- Add yellow nag text to the end of `pulumi` invocations when there is a newer version, with a link to something like https://pulumi.io/install/upgrade.html, which explains that you just need to run that command again or download a new version and extract it.

https://github.com/pulumi/pulumi/issues/1686 is tracking the third bullet, but the other two may be worth taking into consideration as part of this work item.

justinvp:
Close in favor of #1033 (and #997)?

joeduffy:
@justinvp I want to keep this open as a reminder to keep making progress every sprint. I'll move out of M23, however, since we have the concrete work items filed already.

PR #552, opened 2018-08-21
Add a Kubernetes quickstart

joeduffy:
This adds a Kubernetes quickstart based on the Guestbook application,
and also updates our homepage and footer navigation to more closely
align to the more recent pulumi.com updates where Kubernetes takes the
place of CoLaDa as our "fourth pillar."


_data/quickstart.yaml
```diff
@@ -11,3 +11,7 @@ toc:
    - quickstart/aws-ec2.md
    - quickstart/aws-extract-thumbnail.md
    - quickstart/aws-s3-website.md
+- title: Kubernetes Tutorials
+  path: /quickstart/kubernetes
+  section:
+  - quickstart/kubernetes-guestbook.md
```
lukehoban: Indentation here is slightly different - I assume it doesn't matter - but probably worth being consistent (and who knows with YAML what will cause things to mis-parse).

_sass/_pulumi.scss
```diff
@@ -184,7 +184,6 @@ svg {
 
 .what-is-pulumi-cta {
     font-weight: 600;
-    text-transform: uppercase;
```
lukehoban: :+1:

quickstart/kubernetes-guestbook.md
```diff
@@ -0,0 +1,373 @@
+---
+title: "Create a Guestbook with Redis and PHP"
```
lukehoban: Here you call out PHP, below you call out Nginx.  Since PHP is mostly invisible to the authoring experience below, perhaps stick with Nginx here as well?

quickstart/kubernetes-guestbook.md
```diff
@@ -0,0 +1,373 @@
...
+---
+
+In this tutorial, we'll build and deploy
+[the standard Kubernetes Guestbook example](https://github.com/kubernetes/examples/tree/master/guestbook) using Pulumi.
+
+The guestbook is a simple, multi-tier web application that uses Redis and Nginx, powered by Docker containers and
+Kubernetes. The primary difference between this example and the standard Kubernetes one is that we'll be authoring it in
+TypeScript instead of YAML, and we'll deploy it using `pulumi` rather than `kubectl`. This gives us the full power of
+real languages, combined with immutable infrastructure with diffable updates, alongside reliable orchestration, status,
+and robust state management.
```
lukehoban: Perhaps we could simplify this last sentence?

quickstart/kubernetes-guestbook.md
```diff
@@ -0,0 +1,373 @@
...
+    ```shell
+    $ pulumi new kubernetes-typescript --dir guestbook && cd guestbook
+    ```
+
+    There are a variety of ways to create a new project -- for instance, we could have started by cloning
+    [the above repo](https://github.com/pulumi/examples/tree/master/kubernetes-ts-guestbook) -- but templates are easy
+    and allow us walk through the process of creating a new project from scratch. This command will have initialized
+    a fresh project in the `guestbook` directory, and our command has `cd`'d into it.
+
+2.  Next, replace the minimal contents of the template's `index.js` file with the full guestbook:
```
lukehoban: Is this TypeScript or JavaScript (the template above says `typescript`, this says `.js`, the code uses TypeScript features and is in a `js` code block)?

quickstart/kubernetes-guestbook.md
```diff
@@ -0,0 +1,373 @@
...
+                            // environment variable to find the master service's host. To do so, comment out the
+                            // 'value: "dns"' line above, and uncomment the line below:
+                            // value: "env",
+                        }],
+                        ports: [{ containerPort: 80 }],
+                    }],
+                },
+            },
+        },
+    });
```
lukehoban: We should possibly consider using the version of this code in https://github.com/pulumi/examples/tree/master/kubernetes-ts-guestbook instead of the version above.  It highlights more Pulumi concepts instead of being a literal port of the `.yaml` files, and is generally "nicer".

cc @hausdorff

quickstart/kubernetes-guestbook.md
```diff
@@ -0,0 +1,373 @@
...
+    ```
+
+    This is a lot of code, but it just creates three Kubernetes Services, each with an associated Deployment. The full
+    Kubernetes object model is available to us. Programming directly against the raw object model is a great way to ease
+    into using Pulumi, and gives us the full power of Kubernetes right away; however, we will also soon see that real
+    languages can help greatly simplify examples like this, thanks to the added abstraction of classes and functions.
```
lukehoban: > however, we will also soon see that real languages can help greatly simplify examples like this, thanks to the added abstraction of classes and functions.

I don't see any followup on this point in the text here.  

Should we try to just use an example up front that is more Pulumi-natural, and avoid excusing the "lot of code" (highlighting instead how abstractions have helped us reduce repetition and capture common patterns and reuse)?

quickstart/kubernetes-guestbook.md
```diff
@@ -0,0 +1,373 @@
...
+    The `pulumi stack output` command prints exported program variables:
+
+    ```shell
+    $ pulumi stack output frontendIP
+    ```
+
+    The value of the `frontendIP` variable matches exactly the `frontend`'s `CLUSTER-IP` reported by `kubectl`:
+
+    ```
+    10.102.193.86
```
lukehoban: Should this be part of the code block above?

quickstart/kubernetes-guestbook.md
```diff
@@ -0,0 +1,373 @@
...
+      ...
+    </html>
+    ```
+
+    If you are instead running this program in a real cluster, you can change `ClusterIP` to `LoadBalancer` above,
+    run `pulumi up` to update it, and then you can simply access your Guestbook application with:
+
+    ```shell
+    $ curl $(pulumi stack output frontendIP)
+    ```
```
lukehoban: I actually like the idea of putting all this logic in the code as a config var (as the examples repo version of this does), so that it is in code instead of in prose.  I feel like it's an opportunity to comment on how it is easy to make the deployment adaptive via this configuration?

quickstart/kubernetes.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Kubernetes
+---
+
+To learn how to create Kubernetes applications using Pulumi, please explore any of these tutorials:
+
+* [Create a Guestbook with Redis and PHP](./kubernetes-guestbook.html): Build a simple multi-tier
+  application -- a guestbook -- using Kubernetes
```
lukehoban: Do we need this layer of indirection?  The other top level topics just have one example each directly linked from the TOC, and finish with links off to other "next steps" for that area in other places (examples, etc.).  Should we do the same here for now?

_data/quickstart.yaml
joeduffy: Good catch, will fix :+1:

quickstart/kubernetes-guestbook.md
joeduffy: I actually copied this from the Kubernetes quickstart on which I based this example (both the kubernetes.io and GKE docs refer to it under this name). I actually don't have a huge preference on this, so if -- knowing that I copied it -- you'd still prefer the rename, just lmk.

quickstart/kubernetes-guestbook.md
joeduffy: :+1:

quickstart/kubernetes-guestbook.md
joeduffy: I switched halfway through to TypeScript, since the example I'm basing it on and our current Kubernetes template is in TypeScript. I realize we tend to be JavaScript-first, although now that we have the first class TypeScript provider, I suspect it matters less (no manual `tsc` steps).

joeduffy: And, of course, I will fix up to be consistent -- good catch.

quickstart/kubernetes-guestbook.md
joeduffy: I looked at doing that, but it was a little too "advanced" for a getting started tutorial, IMHO. Plus, it involves multiple files, which just makes the overall tutorial a little bit harder to follow.

As you noted below, I alluded to building on this example to demonstrate the benefits of "real code" and I thought this would be a nice progression (basically, "quickstart 1a").

quickstart/kubernetes-guestbook.md
joeduffy: Yeah, we need to build on this and add the aforementioned extension. I think it's fine to mention in passing just in case someone starts to think "hey, this just looks like YAML written in JavaScript."

joeduffy: (and once we have it, we can hyperlink to it)

quickstart/kubernetes-guestbook.md
joeduffy: I started explicitly factoring these out. Definitely up for feedback on this. I noticed the Kubernetes tutorials don't put the output and commands together which, I gather, makes it clearer what you're meant to copy and paste, versus just output you should expect. (Coupled with [small touches like copy-to-clipboard icons](https://github.com/pulumi/docs/issues/550) for the commands, this would matter more.)

I don't feel strongly but I do think it flows nicer.

quickstart/kubernetes-guestbook.md
joeduffy: I'm fine doing that, but I'm going to flip the polarity (and think we should on the example also). Basically, assume there is no load balancer unless explicitly stated; this is simpler, and for the newbie users that are relying on Minikube, it's one less thing to grok when you're just trying to get up and running. Presumably folks with a real cluster will actually know they need a load balancer, whereas these newer users won't know they don't need one.

quickstart/kubernetes.md
joeduffy: I plan to refactor all of these into AWS, Azure, Google Cloud, Kubernetes, and Cloud Framework, each with its own landing page of multiple tutorials, as part of https://github.com/pulumi/docs/issues/500. This is in line with that. If you hate it, we can consider alternative approaches, but this seems like a pretty intuitive way to go about it.

ericrudder:
I think flipping polarity is probably a good idea, however, it is amazing
how often I see this referred to online tripping folks up ... I probably
have to do more queries on K basics than you guys in order to understand
what i am looking at, and the confusion runs wide

PR #555, opened 2018-08-23
Fix typos

justinvp:


justinvp:
This is no longer relevant after Joe's changes.

PR #564, opened 2018-08-31
Fix install links

lukehoban:


ericrudder:
Do we want to fix w. redirect?

A quick look indicated we might have in other spots (e.g. quickstart/kubernetes/tutorial-guestbook.md)

We can add one of fixes if you prefer ... I think we probably start with an example base and edit a fair amount.

joeduffy:
Strange, the file does have a redirect at the top: https://raw.githubusercontent.com/pulumi/docs/master/quickstart/aws/setup.md. I wonder if this didn't work because the link is relative.

I think maybe changing it to `redirect_from: install/aws.html` might do the trick.

joeduffy:
Ah, I see, it's just due to the relative path issues. The redirect is in place, but after the recent changes, this resolves to `quickstart/install` which is not a valid target. LGTM, merging now, I'll deploy as-is.

PR #566, opened 2018-08-31
Add some comparisons to other tools

joeduffy:
Fixes pulumi/pulumi#549.


joeduffy:
I'll see what I can do. I'm thinking we move them under `reference/vs`, which at least helps a bit with SEO. It also gets them out of the QuickStart flow. I am a little on the fence about this; if we had collapsable nav, I think I'd prefer for it to be underneath our "intro" material since I expect new users to wonder about this. But absent that, I'd prefer not to clutter things up needlessly.

quickstart/comparisons/cloud_templates.md
```diff
@@ -0,0 +1,40 @@
...
+These configuration files are often uploaded to a hosted service in the target cloud, where it will then process the
+files to create, update, or delete resources as needed.
+
+Pulumi's model shares a lot with these systems. Although Pulumi programs are written in imperative, familiar languages,
+they are ultimately evaluated to produce a similar set of create, update, or delete operations for your cloud resources.
+
+This is where Pulumi begins to diverge from these other solutions, however.
+
+Pulumi lets you use your favorite languages. Instead of bespoke templating solutions, often tacked on to a markup
+in an awkward and hard-to-learn and -remember way, Pulumi programs use real code. Thanks to simple things like
```
ericrudder: did you want all dashes until remember or just have one extra?

joeduffy: I did, as I meant hard-to-learn and hard-to-remember, but I'll admit this is an awkward phrasing. I can look into tidying it up.

PR #567, opened 2018-09-01
Deploy broken link fixes to production

joeduffy:


ericrudder:
@joeduffy i noticed https://pulumi.io/quickstart/kubernetes/tutorial-guestbook.html still has the broken install link; this is the thing you said you thought was fixed in the meeting

PR #584, opened 2018-09-11
Update API docs to latest released versions

lukehoban:
Updates CLI and API reference docs to latest released versions.


lukehoban:
> I glanced quickly at this to try to understand where the broken links were coming from, but I was unable to figure it out.

Pushed a commit that should fix this.

PR #589, opened 2018-09-18
Create UnwrapBlog.md

CyrusNajmabadi:
Tagging @hausdorff @lukehoban @mehzer 'Unwrap' blog post. Take a look at let me know what you think!

Alternative title: Advanced TypeSript types FTW!


CyrusNajmabadi:
Tagging @hausdorff @lukehoban @mehzer   'Unwrap' blog post.  Take a look at let me know what you think!

CyrusNajmabadi:
Alternative title: Advanced TypeSript types FTW!

CyrusNajmabadi:
Tagging @danielrosenwasser as well :)

mehzer:
I like the alt title better. Looks like a great post, can't comment on the
subject matter as its way over my head, but the writing is clear and good
tone.

CyrusNajmabadi:
Updated title!

UnwrapBlog.md
```diff
@@ -0,0 +1,92 @@
...
+
+```ts
+interface VpcInput {
+    ipAddresses: Input<Input<string>[]>;
+}
+```
+
+Breaking that down, it's basically saying "I'll take an an array of strings.  But those actual string values can individually be strings, Promises-of-strings, or Strings-carrying-dependency information." It even goes one step beyond that, because the entire array itself could be just an array-of-those-things, or a Promise-of-array-of-those-things, or an Array-of-those-things-with-dependency information.  Whew...!
+
+This approach is great for people producing the data.  As a producer i could do:
```
ellismg: nit: Capitalize I

UnwrapBlog.md
```diff
@@ -0,0 +1,92 @@
...
+
+This complex type can often show up in forms like:
+
+```ts
+interface VpcInput {
+    ipAddresses: Input<Input<string>[]>;
+}
+```
+
+Breaking that down, it's basically saying "I'll take an an array of strings.  But those actual string values can individually be strings, Promises-of-strings, or Strings-carrying-dependency information." It even goes one step beyond that, because the entire array itself could be just an array-of-those-things, or a Promise-of-array-of-those-things, or an Array-of-those-things-with-dependency information.  Whew...!
```
joeduffy: Nit, "an an".

UnwrapBlog.md
```diff
@@ -0,0 +1,92 @@
...
+
+```ts
+type Input<T> = T | Promise<T> | Output<T>;
+
+// Like a Promise<T>, but also keeps track of Dependency information so we can
+// figure out what the upstream/downstream impact is of any value change.
+interface Output<T> { /* ... */ }
+```
+
+This complex type can often show up in forms like:
```
joeduffy: Maybe

> By convention, all resource input properties are `Input<T>`s, and al output properties are `Output<T>`s. These complex types can often show up in forms like:

?

UnwrapBlog.md
```diff
@@ -0,0 +1,92 @@
+Advanced TypeScript type FTW!
+
+In Pulumi we have a data model that allows people to express complex data that may *eventually* be available. Traditional JS programming might expose that as a Promise<T>, but we've taken that one step further by introducing a type we call:
```
joeduffy: Maybe one more sentence to motivate? E.g.

> We at Pulumi love TypeScript for cloud apps and infrastructure, because of its rich type system and great ahead-of-time typechecking -- making for a more productive inner loop and helping to find errors sooner. The typesystem magic behind how this works for infrastructure as code, however, can be fascinating!

UnwrapBlog.md
```diff
@@ -0,0 +1,92 @@
...
+```ts
+new Vpc({ ipAddress: ["an-address"] }); // or even 
+new Vpc({ ipAddresses: ["an-address", computeAddressAsynchronously() ]}); // or even
+new Vpc({ ipAddresses: ["an-address", computeAddressAsynchronously(), someDependency.address ]}); // or even
+new Vpc({ ipAddresses: computeAnArrayAnyOfTheAboveAsynchrousnously() }); // etc. etc.
+```
+
+It's super flexible, and let's people naturally produce the values in whatever way they want without having to work too hard to make it conform to some very narrow type.
+
+However, while easy to produce, this can sometimes be difficult to consume.  For example, if you wanted to take in that `Input<Input<string>[]>` and then check if there was a very specific value in it, you might have to go through and do a bunch of work.  You'd have to first deal with potentially getting a `Input<string>[]` or a `Promise<Input<string>[]>` or even an `Output<Input<string>[]>`. Then, once you even got to the underlying array, you'd have to deal with each potential element in it being a `string`... or a `Promise<string>`... or a... well... you get the idea.  It's not fun :)
```
joeduffy: Nit, swap real emoji smile for ":)"?

UnwrapBlog.md
joeduffy: Nit, "JS" => "JavaScript"?

UnwrapBlog.md
```diff
@@ -0,0 +1,92 @@
...
+
+We knew that providing the implementation such a function itself was not going to the true hard part.  After all, we already had the code to peel back one layer.  So peeling back all the layers was more a matter of just having the function call itself recursively when encountering things like arrays (or other objects that themselves contained `Input`s).  The hard part came in when were trying figure out how would we even express this in TypeScript?.  After all, what's the type signature when you effectively to say:
+
+```ts
+function unwrap(input: Input<...MaybeHasInputsOrPromisesOrOutputsArbitrarilyDeeper...>): SomeThingWithAllTheInputsPromisesAndOutputsErased;
+```
+
+This was a big headscratcher for a while.  But after exploring some of the powerful new features that TypeScript has been bringing about in recent releases, we finally found a great way to manage it.  In the end, it ends up looking something like this:
```
joeduffy: Any way to link "powerful new features" to release notes or examples? E.g., for this sort of thing, I'd usually link powerful, new, and features to different examples of our favorite features here.

joeduffy: Actually, I see that you link below, so this is probably okay as-is.

UnwrapBlog.md
```diff
@@ -0,0 +1,92 @@
...
+
+Put together, the `Unwrap<T>` is effectively itself a recursively defined structural pattern matching type.  You can see it reference itself once you walk into `UnwrappedArray<T>` and `UnwrappedObject<T>` types. As it recurses, it matches the type it has and then 'peels away' the wrappings that have been caused by `Promise<T>` and `Output<T>`.  As it hits Arrays and Objects, it dives deeper, unwrapping along the way.  This is all manageable in TypeScript because it does this sort of matching and unwrapping in a lazy fashion, allowing us to express what is effectively an infinitely large pattern-matching type, without it having any trouble understanding it or applying it!
+
+It amazes us how well this works and how amazing TypeScript is here.  As you can see here, TS completely understands what's going on and will give you all the great type-checking support and tooling that we know and love:
```
joeduffy: Nit, we should try to be consistent -- either it's TypeScript or TS... I prefer TypeScript.

UnwrapBlog.md
```diff
@@ -0,0 +1,92 @@
...
+
+![image](https://user-images.githubusercontent.com/4564579/45712608-9e6cd100-bb41-11e8-90dd-b4fe79d236fe.png)
+
+Since adding support for this, we've been able to start using this new function across our own codebase and we've seen things get dramatically simpler.  The pyramid-of-doom goes away and code becomes much cleaner and easier to read and maintain.  JavaScript helped us write this system easily, but thanks to TypeScript we can express these complex intents and have the safety and understanding to do this sort of thing with confidence.
```
joeduffy: Any good before/after example to show?

joeduffy:
Love it! Minor comments throughout.

UnwrapBlog.md
CyrusNajmabadi: I didn't want to bring up 'Resources', because that's a whole new concept in itself.

CyrusNajmabadi: That's why i just leave it as Inputs, and Outputs, and just briefly describe why an 'Output' would even exist.

UnwrapBlog.md
```diff
@@ -0,0 +1,92 @@
+Advanced TypeScript type FTW!
+
+In Pulumi we have a data model that allows people to express complex data that may *eventually* be available. Traditional JavaScript programming might expose that as a Promise<T>, but we've taken that one step further by introducing a type we call:
```
lukehoban: Hyperlink to pulumi.io?

Perhaps expand on the first sentence to include that we have a JavaScript/TypeScript SDK, and that the SDK needs a way to expose properties that may eventually be available along with their dependencies.  Just to set context a bit.

UnwrapBlog.md
lukehoban: I like something like what @joeduffy suggests - maybe just "all inputs to Pulumi objects"?

UnwrapBlog.md
```diff
@@ -0,0 +1,92 @@
...
+    ipAddresses: Input<Input<string>[]>;
+}
+```
+
+Breaking that down, it's basically saying "I'll take an array of strings.  But those actual string values can individually be strings, Promises-of-strings, or Strings-carrying-dependency information." It even goes one step beyond that, because the entire array itself could be just an array-of-those-things, or a Promise-of-array-of-those-things, or an Array-of-those-things-with-dependency information.  Whew...!
+
+This approach is great for people producing the data.  As a producer I could do:
+
+```ts
+new Vpc({ ipAddress: ["an-address"] }); // or even
```
lukehoban: How is the first of there related to the type above?  This seems like it wouldn't work.

UnwrapBlog.md
CyrusNajmabadi: I didn't want to point at actual examples, because the blog post is a bit of a simplification.  For example, our 'unwrap' is actually called 'output' (and i didn't want to explain that), and we return an `Output<Unwrap<T>>` and i didn't want to have to explain that.  And the code that uses that needs to use .apply, and i didn't want to explain that :)

The meat of the type system and how we use it is what i was trying to convey here.  And, to keep things on that point, i simplified things i felt would just not substantively contribute to that message.

UnwrapBlog.md
```diff
@@ -0,0 +1,92 @@
...
+
+Our core goal was to change our 'unwrap' function from just peeling away one layer of this onion, to peeling away *all* layers.  So, for example, if you started with an `Input<Input<string>[]>` you would just get back a plain old `string[]` that you could use completely naturally in all the ways you'd expect.  We were initially uncertain if we could make this work, but thanks to the power of JavaScript and the expressivity of TypeScript, we think we found a great solution here.
+
+We knew that providing the implementation such a function itself was not going to the true hard part.  After all, we already had the code to peel back one layer.  So peeling back all the layers was more a matter of just having the function call itself recursively when encountering things like arrays (or other objects that themselves contained `Input`s).  The hard part came in when were trying figure out how would we even express this in TypeScript?.  After all, what's the type signature when you effectively to say:
```
lukehoban: Making the point that this function is in fact really *easy* to write in JavaScript, but just not as easy to *type* might help here.  That's sort of the crux of why this is interesting/exciting - a reasonably easy/common thing to do in JavaScript, which can now be typed accurately in TypeScript.

CyrusNajmabadi:
Ok.  Have incorporated things.  @mehzer Feel free to run with htis.

UnwrapBlog.md
CyrusNajmabadi: I thought that was made with: "The hard part came in when were trying figure out how would we even express this in TypeScript?"

UnwrapBlog.md
CyrusNajmabadi: An array of strings, is itself an ```Input<Input<string>[]>```

UnwrapBlog.md
lukehoban: Yep - point is made - was more thinking of emphasizing how easy/normal it is to write this sort of thing in JavaScript - and hence why this is a problem/solution that is broadly interesting.

UnwrapBlog.md
CyrusNajmabadi: oh, i see.  this is about saying ipAddress.  yes, that was a typo.  fixed!

UnwrapBlog.md
CyrusNajmabadi: I've tried to clean this up a tad :)

UnwrapBlog.md
CyrusNajmabadi: Added links.

CyrusNajmabadi:
Alright.  All feedback incorporated!

mehzer:
OK, I'll stage on HubSpot and give you a final review @CyrusNajmabadi
<https://github.com/CyrusNajmabadi>

CyrusNajmabadi:
@mehzer I did make some last minute changes.  so you may want to recopy/paste.

CyrusNajmabadi:
Closing out now that @mehzer has his copy.

PR #633, opened 2018-10-30
Update Changelog for 0.16.2

ellismg:


dashboard.json
```diff
@@ -36,6 +36,11 @@
         "title": "News",
         "icon": "radio",
         "items": [
+            {
```
joeduffy: @justinvp Do we serve this up anywhere anymore?

justinvp: This file (`dashboard.json`) is no longer used and can be deleted.

justinvp: Filed #634

PR #635, opened 2018-10-30
Delete dashboard.json

### LINKED ISSUE #634: Delete dashboard.json
It's no longer used and can be deleted.

justinvp:
It's no longer used, so no need to keep it around.

Fixes #634


infrastructure/index.ts
```diff
@@ -32,15 +32,6 @@ const contentBucket = new aws.s3.Bucket(
             indexDocument: "index.html",
             errorDocument: "404.html",
         },
-        // The docs website serves a file dashboard.json, which is read by a few services
-        // to surface recent updates. We need to set CORS headers for that file.
-        corsRules: [
```
justinvp: I considered keeping these around for now, but I don't think we have a reason to.

PR #642, opened 2018-11-06
Update changelog for 0.16.3

ellismg:
I did not update the `latest-version` text file, since there were no differences in the CLI between 0.16.2 and 0.16.3

### DIFF
--- a/reference/changelog.md
+++ b/reference/changelog.md
@@ -25,6 +25,11 @@ redirect_from: /install/changelog.html
     <tbody>
 
         <tr>
+            <th scope="row"><a href="#v161">0.16.3</a></th>
+            <td>2018/10/29</td>
+            <td>{% include sdk-links.html version='0.16.3' %}</td>
+        </tr>
+        <tr>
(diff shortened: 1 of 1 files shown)


ellismg:
I will admit that I am surprised that there are changes to any resources aside from `changelog.html` in the preview.

reference/changelog.md
```diff
@@ -87,6 +92,18 @@ redirect_from: /install/changelog.html
...
+
+Released on November 6, 2018
+
+### Pulumi CLI
+
+There are no changes from the previous version of the Pulumi CLI, but it shares the same version as the `@pulumi/pulumi` package, which was updated
+
+### @pulumi/pulumi 0.16.3
+
+- Support closure seralization when running on Node 11 [pulumi/pulumi#2101](https://github.com/pulumi/pulumi/pull/2101)
```
lukehoban: How about we just say - "Fully support Node.js 11".  (folks may not realize that closure serialization was the only thing not yet working on Node 11).

ellismg: Can do!

PR #658, opened 2018-11-16
Update Kubernetes tutorial index

hausdorff:


_data/quickstart.yaml
```diff
@@ -26,8 +26,10 @@ toc:
   path: /quickstart/kubernetes/index
   section:
    - quickstart/kubernetes/setup.md
-   - quickstart/kubernetes/tutorial-stateless-app.md
-   - quickstart/kubernetes/tutorial-guestbook.md
+   - quickstart/kubernetes/tutorial-exposed-deployment.md
+   - quickstart/kubernetes/tutorial-configmap-rollout.md
+   - quickstart/kubernetes/tutorial-wordpress-chart.md
+   - quickstart/kubernetes/tutorial-p8s-rollout.md
```
lukehoban: Any reason not to also link the "stateless-app" and "guestbook" here?  In particular since those are things which are well known from kubernetes.io docs?

lukehoban:
Also - these pages have quite long titles.  Do they look good in the TOC?  We may have to shorten the titles to ensure they read well in the TOC.

_data/quickstart.yaml
hausdorff: I think they're less compelling. We can link them if you feel really strongly about it, but I think the guestbook (_e.g._) is better suited to an example.

lukehoban: Let's keep guestbook here at the end.  Can drop "stateless-app" as that is more or less replaced by "exposed-deployment".

Issue #608, opened 2018-10-08
Document `pulumi state` command


swgillespie:
We should document the `pulumi state` command and how to use it on the Troubleshooting page once https://github.com/pulumi/pulumi/pull/2024 lands.

chrsmith:
We didn't end up taking this in the v0.16.0 release, so moving to the next milestone.

PR #671, opened 2018-11-21
Document the 'pulumi state' command

swgillespie:
The docs website already has autogenerated documentation for 'pulumi
state' thanks to Cobra, so this commit links to it from the
troubleshooting doc and inserts it into the sidebar for easy navigation.

Fixes https://github.com/pulumi/docs/issues/608


joeduffy:
LGTM :+1:

PR #673, opened 2018-11-21
Add quickstart notes on how to install AWS plugin

swgillespie:
Customers sometimes run into issues with the AWS quickstart when their
AWS plugin did not automatically install as part of their package
installation. In the event that NPM did not correctly install their
plugin, this commit adds notes to the relevant areas that instruct users
on how to install the plugin manually.

Fixes https://github.com/pulumi/pulumi/issues/2097

### DIFF
--- a/quickstart/aws/tutorial-containers-ecs-fargate.md
+++ b/quickstart/aws/tutorial-containers-ecs-fargate.md
@@ -102,6 +102,10 @@ In this tutorial, we'll use JavaScript to build and deploy a simple container to
     Update duration: 3m53.44141303s
     ```
 
+> ***Note***: If you see an error message like `no resource plugin 'aws' found in the workspace or on your $PATH`, you
+> can manually install the Pulumi AWS plugin with `pulumi plugin install resource aws v$(npm info @pulumi/aws
+> version)`.
+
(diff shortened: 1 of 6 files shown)


swgillespie:
> Hmm - I don't think this is how we should "fix" pulumi/pulumi#2097. I'd prefer to make sure this doesn't happen, and in the very rare case that it does due to things entirely out of our control, find a way to provide richer information in the context of the failure.

We already do this now, to be clear: https://github.com/pulumi/pulumi/pull/2040.

If we think this is sufficient, then sure, we can close this PR 😄 . However, there is evidence that users do occasionally run into this issue. I'm fine either way.

lukehoban:
> We already do this now, to be clear: pulumi/pulumi#2040.

Nice!  I didn't realize we had already done that.  Personally I think that's a sufficient alternative to this docs change - gives an actionable fix immediately at the point of the error - which is frankly even better than in the docs.

I'm still unclear on why we expect this to be common for users to hit - and whether that is because we are doing something wrong somewhere else or just because we believe there Internet is unreliable.

lukehoban:
Per notes in pulumi/pulumi#2097 we can close this out.

Issue #601, opened 2018-09-29
Installation is too buried


joeduffy:
After the change to streamline the Getting Started workflow, we went from zero clicks to three clicks to find the installation instructions. This feels like a step backwards. Let's see if we can find a middle ground.

lukehoban:
Related to this - I've noticed a couple other pieces of information related to installation are now harder (or impossible?) to find:
1. What is the version of the current release I will get if I follow the installation instructions (and generally - what is the current version).
2. The changelog with links to historical download version.
3. Details on manually installing from tarball.

joeduffy:
For 2, it is under Reference at https://pulumi.io/reference/changelog.html. Perhaps too hard to find.

For 3, if you select "Manual Install" in the dropdown, you will see the manual install steps.

Issue #572, opened 2018-09-04
Fix up some nav issues


joeduffy:
* Collapsable left hand nav
* Put the Tour underneath Getting Started
* Split Reference into Guide vs. Reference
    - Concepts, CI/CD, and Pulumi vs. Others, FAQ, Change Log into Guide
    - Packages and CLI Commands underneath Reference

joeduffy:
* Move GitHub star button to top nav just like pulumi.com

joeduffy:
This has been basically been done, and is now superseded by https://github.com/pulumi/docs/issues/744.

PR #676, opened 2018-11-25
Improve the landing page

joeduffy:
This change improves the landing page slightly:

* Shrink the Pulumi logo on the landing page a little bit.

* Improve the sub-heading to just "Build software for any cloud
  using your favorite language," from the previous wordy blurb.

* Drop the "Tour" button, and add an "Install" button in its place.

* Colorize all buttons the same. The prior approach of making
  "Get Started" being a different color had the confusing effect of
  making it seem like you were actively in the Getting Started section.

* Move the GitHub star button from just the homepage to the upper right
  corner of the nav bar, similar to what we do on www.pulumi.com.

* Add a band of cloud logos just beneath the buttons. This only
  includes AWS, Azure, GCP, and Kubernetes -- I've dropped OpenStack
  and the Pulumi Cloud Framework, asthey are lesser known/used.

* Add a band with "Why Pulumi?", "Alternatives", and "Community,"
  with little blurbs to tell people what Pulumi *is* and why you
  might be interested in using it.

This is part of the recurring task, pulumi/docs#500, improves
pulumi/docs#572, and fixes pulumi/docs#601.

There is more to do here -- notably, I'd like some code snippets
and perhaps a CLI screenshot somewhere on this page, perhaps as a
distinct band above or beneath the "Why Pulumi?", etc. table.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -1,6 +1,4 @@
 <header class="main-header">
-    <a href="https://github.com/pulumi/pulumi"><img style="position: absolute; top: 45px; right: 0; border: 0;" src="https://s3.amazonaws.com/github/ribbons/forkme_right_orange_ff7600.png" alt="Fork me on GitHub"></a>
-
     <div class="utility-bar hidden-xs hidden-sm">
         <div class="container-fluid">
             <div class="row">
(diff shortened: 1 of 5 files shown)


joeduffy:
Before:

image

After:

image

Issue #202, opened 2018-04-14
Update Jekyll configuration to not emit .html extension


lindydonna:
A friend pointed out that our docs URLs look a bit "old school" with the `html` extension. There should be a simple Jekyll config that fixes this.

lblackstone:
Here are a couple related links:
https://medium.com/@mittalyashu/how-to-remove-html-extension-in-jekyll-efeb48bd725f
https://github.com/pulumi/docs/pull/684

Summary:
Jekyll will generate links without the `.html` extension if they include a trailing `\`. A possible alternative would be creating "duplicate" S3 buckets for links with both a `link.html` key and a bare `link` key.

chrsmith:
With #684, this should be "fixed". And it's just a matter of linking to the non-`.html` versions of files, and adding redirects to canonicalize the cleaner URLs. Let me see the extent to which we need to change things here, and if there is anything missing.

chrsmith:
Looks like @cnunciato is a better owner for this in the current milestone. Here's my thinking about how to tackle this in the short-term:

- Move to a world where we just emit pages as `/foo/index.html` instead of `foo.html`. See https://github.com/pulumi/docs/pull/1022 and https://github.com/pulumi/docs/pull/1014 as examples of this type of approach.

An approach we _could_ take, but probably shouldn't:

- Modify `infrastructure/index.ts` to make a copy of any `.html` files it puts into S3 and removes the extension, but sets the appropriate `Content-Type` header. So we create both a `/foo.html` and `/foo` object in the S3 bucket, which would allow us to serve `/foo` and `/foo.html` when deployed.

  This actually works surprisingly well, but seems dubious because we would be fighting against Jekyll (which will always emit the .html file). It seems like if we wanted to do this, we should instead not serve content directly from S3, but standup some small web server to have the logic to map a request for `/foo` to check for an S3 object `/foo`, and if not present, fallback to `/foo.html`. (Essentially what the built-in webserver does when you run `jekyll serve`.)

cnunciato:
Waiting on the outcome of #1034 on this (since it'd be handled by Hugo out of the box).

justinvp:
Taking this over as I've already taken care of it as part of #1088

PR #684, opened 2018-11-26
Trim .html postfix on S3 bucket keys

lblackstone:


lblackstone:
This won't work like I was thinking. Closing this one out and will try a different approach.

chrsmith:
Thanks for taking a swing at this issue, which should fix #202 . What you have is close to what we want, but as you already are aware, would break all of our existing links to files ending in `.html`.

An alternative would be to create a copy of the file but without the `.html` extension when we upload it to S3. That way, in addition to `foo.html`, we would also have `foo` in the S3 bucket as well (which would be served at http://.../foo (without the .html suffix).

To do that, the necessary change would be to do something like: if the file extension is .html and the file without the .html extension isn't already found (i.e. don't clobber/overwrite), then create a _second_ `aws.s3.BucketObject` object with the trimmed key, like you have in the PR.

The only additional nuance is that we'd probably need to set the `Content-Type` property for the file, since without the `.html` extension it might default to the wrong thing. But as long as we keep the `contentType: getMimeType(filePath)` line, we'll probably be just fine.

joeduffy:
That sounds like what I was thinking too. The only hitch is whether there could be legitimate scenarios where we have both `foo.html` and `foo` as objects already, in which case the simple algorithm suggested would end up overwriting a legitimate file. Is this something we need to worry about?

PR #686, opened 2018-11-27
Update readme

atpoirie:
Update to correct region name and use latest AMZN Linux AmI


lukehoban:
BTW - the good news is that the comment was wrong and the previous AMI was indeed in `us-east-1` (or else this quickstart would not have been working as documented).  However, great to fix the comment and update to the latest AMI.

PR #689, opened 2018-11-28
Use shorter shortlinks

lblackstone:
Rather than needing to choose a meaningful name for each "shortlink",
use an alphanumeric 6-character link. Updated README with a command
to generate a link in this format.


lblackstone:
@chrsmith Added a script and updated the README

scripts/create_short_url.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+
```
chrsmith: Let's add a big long, comment explaining what this script does.

scripts/create_short_url.sh
```diff
@@ -0,0 +1,45 @@
...
+set -o errexit -o pipefail
+cd "$(dirname "${BASH_SOURCE}")/.."
+
+if [[ $# -ne 2 && $# -ne 3 ]]; then
+    cat << EOF
+Usage: create_short_url.sh <target_link> <filename> [shortlink]
```
chrsmith: Nit: If we ever rename the script this would be out of date, `$0` can be used instead of `create_short_url.sh`.

scripts/create_short_url.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# If shortlink not provided, generate a random 6-character alphanumeric link
+if [[ ! ${SHORTLINK} ]]; then
+    PERMALINK="$(LC_ALL=C tr -dc 'a-z0-9' </dev/urandom | head -c6 ; echo)"
+else
+    PERMALINK="${SHORTLINK}"
+fi
+
+RELATIVEPATH="./shortlinks/${FILENAME}.md"
+FILEPATH=$(echo "$(cd "$(dirname "$RELATIVEPATH")"; pwd -P)/$(basename "$RELATIVEPATH")")
```
chrsmith: To cut down on the nested expressions, could we have `FOO=$(cd "$(dirname "$RELATIVEPATH")"; pwd -P)` and then reference `$FOO` here instead?

chrsmith: Nit: To be consistent with the rest of the file, we should use `${RELATIVEPATH}`, right?

PR #693, opened 2018-11-28
Update docs for v0.16.6

pgavlin:


pgavlin:
I'm not sure that we actually want to update the package docs for anything besides `@pulumi/pulumi`. Thoughs @lukehoban @ellismg ?

ellismg:
I think you are correct about just updating @pulumi/pulumi docs.

I am 87% sure that the trailing newline in latest-version is not going to cause problems but we have never done it before, so do you mind removing it until we can actually test how the bash script handles it?


LGTM otherwise.

reference/changelog.md
```diff
@@ -24,6 +24,11 @@ redirect_from: /install/changelog.html
     </thead>
     <tbody>
 
+        <tr>
+            <th scope="row"><a href="https://github.com/pulumi/pulumi/blob/v0.16.6/CHANGELOG.md#0166-released-novemeber-28th-2018">0.16.6</a></th>
```
justinvp: I [fixed](https://github.com/pulumi/pulumi/pull/2249) the `novemeber` typo in CHANGELOG.md and didn't realize it was being referenced here. We should fix here as well.

PR #711, opened 2018-12-11
Force Powershell to use TLS1.2

ellismg:
We will be moving get.pulumi.com over to requring TLS 1.2 in a short
while. When we do so, some versions of PowerShell will not establish a
secure connection to this host by default.

We must force PowerShell to use TLS 1.2 in this case, otherwise our
call to download the install.ps1 script and then run it would fail.


ellismg:
Validated that without this extra piece, powershell refused to download from get.pulumi-staging.io, which now has the change to require TLS 1.2

PR #714, opened 2018-12-14
Release 0.16.8 CLI

ellismg:


ellismg:
> I guess there were no API or CLI changes that needed updating in docs?

Correct.  The payload only contains a single fix in the implementation of the CLI.  No other changes:

PR #710, opened 2018-12-11
Add Kubernetes the Prod Way

hausdorff:

PR #716, opened 2018-12-18
Address comments in #710

hausdorff:


hausdorff:
@lukehoban LGTM'd this over slack.

PR #719, opened 2018-12-21
Flesh out programming model docs

lukehoban:
Also add shortlink for outputs for use in error messages.

Also adds language chooser to all code examples in the programming model docs. This helps ensure these docs are useful across languages.  There are many places where topics covered in these docs are not yet supported, and that is called out more explicitly now.

### DIFF
--- a/reference/programming-model.md
+++ b/reference/programming-model.md
@@ -2,7 +2,7 @@
 title: "Programming Model"
 ---
 
-In Pulumi, [resources](programming-model.html#resources) are defined by allocating resource objects in a program, such as `new aws.ec2.Instance(...)`.  The first argument passed to the resource constructor is its `name`, which must be unique within the Pulumi program. To create dependencies between resources, just reference the [output properties](#outputs) of a resource. For example, this definition of an EC2 instance creates a dependency on a `SecurityGroup`:
+In Pulumi, [resources](#resources) are defined by allocating resource objects in a program, such as `new aws.ec2.Instance(...)`.  The first argument passed to the resource constructor is its `name`, which must be unique within the Pulumi program. To create dependencies between resources, just reference the [output properties](#outputs) of a resource. For example, this definition of an EC2 instance creates a dependency on a `SecurityGroup`:
 
 
 ```js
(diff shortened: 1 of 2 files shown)


reference/programming-model.md
```diff
@@ -33,80 +64,243 @@ The [@pulumi/pulumi] package is the core library for working with the Pulumi pla
...
+All resource constructors also accept a third argument which can provide the following additional properties. 
+- `dependsOn` - a list of explicit resource dependencies
 - `protect` - whether to mark a resource as protected. A protected resource cannot be deleted directly: first you must set `protect: false` and run `pulumi update`. Then, the resource can be deleted, either by removing the line of code or by running `pulumi destroy`.
 - `parent` - optional parent for the resource. See [Components](#components).
 - `provider` - optional provider for the resource. See [Providers](#providers).
 
 ## Resource outputs {#outputs}
 
-<!-- TODO: add direct anchor link to `apply` once #333 is fixed -->
+Outputs are a key part of how Pulumi tracks dependencies between resources.  Because the values of Outputs are not available until resources are created, these are represented using a special [Output] type which internally represents two things:
```
joeduffy: Does `[Output]` need to be `[pulumi.Output]`?

reference/programming-model.md
```diff
@@ -33,80 +64,243 @@ The [@pulumi/pulumi] package is the core library for working with the Pulumi pla
...
+1. An eventually available value of the output
+2. The depdency on the source(s) of the output value
+
+The output properties of all resource objects in Pulumi have type [`Output`][pulumi.Output]. Resource inputs have type [`Input`][pulumi.Input], which accepts either a raw value or an output from another resource. This allows dependencies to be inferred, including ensuring that resources are not created or updatred until all their dependencies are availabe and up to date.
```
joeduffy: Do you want to mention that `Input` can also be a `Promise`?

Also, in general, I often draw an analogy between `Output`s and `Promise`s -- e.g., "`Output`s are just `Promise`s that carry dependency information." Not sure if that analogy makes sense to state explicitly here.

reference/programming-model.md
```diff
@@ -33,80 +64,243 @@ The [@pulumi/pulumi] package is the core library for working with the Pulumi pla
...
+url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error) { 
+    return "https://" + dnsName, nil 
+})
+```
+
+The `apply` method accepts a callback which will be passed the value of the `Output` when it is availabe, and which returns the new value.  The result of the call to `apply` is new `Output` whose value is the value returned from the callback, and which includes the dependencies of the original `Output`.  If the callback itself returns an `Output`, the dependencies of that output are unioned into the dependencies of the returned `Output`.
+
+> Note: The `Output` itself cannot be used directly in string concatenation or other operations, as it is not itself the value of the output.  To transform the value of the output (when it becomes available), the `apply` method should be used instead.
+
+To combine multiple `Output`s into a transformed value, use [pulumi.all].  This allows a new value to be constructed from several inputs, such as concatenating outputs from two different resources together, or constructing a policy document using information from several other resources.
```
joeduffy: I wonder if `apply` and `all` should get their own H3 headers, so we can more easily link to them.

reference/programming-model.md
```diff
@@ -33,80 +64,243 @@ The [@pulumi/pulumi] package is the core library for working with the Pulumi pla
...
+let connectionString = pulumi.all([sqlServer.name, database.name])
+                             .apply(([server, db]) => `Server=tcp:${server}.database.windows.net;initial catalog=${db}...`);
+```
+
+```python
+# `all` is not yet available in Python
+```
+
+```go
+// `all` is not yet available in Go
```
joeduffy: Should we link to a GitHub issue tracking this, so folks can vote or ask about it as needed?

PR #724, opened 2018-12-25
Docs update for pulumi 0.16.9

ellismg:


reference/changelog.md
```diff
@@ -23,6 +23,11 @@ redirect_from: /install/changelog.html
         </tr>
     </thead>
     <tbody>
+        <tr>
+            <th scope="row"><a href="https://github.com/pulumi/pulumi/blob/master/CHANGELOG.md#0169-released-december-24th-2018">0.16.9</a></th>
```
ellismg: The `master` here is unfortunate, but correct.  I forgot to actually put the release date in CHANGELOG.md before doing the release, so the tagged commit does not have the correct title and hence does not have the correct link.

PR #726, opened 2018-12-26
Unset the body background color

cnunciato:
Applying a dark background color to the body tag causes some browsers (Chrome, Firefox, Safari) to paint the scroll thumb in a light color in order to provide suitable contrast, but since we’re setting the background color of page _content_ back to white, the scroll thumb gets lost, which makes it difficult to visualize the size of a page and one’s scroll position within it.

Before & after:


joeduffy:
LGTM, good find! 👍

PR #730, opened 2019-01-01
Added plugin section to Azure setup instructions

DaRosenberg:


lukehoban:
@DaRosenberg Do you recall what set of steps you followed that lead to you getting an error here?  In general, our goal is that `pulumi plugin install` is not needed under any normal circumstances - that should happen automatically as part of installing NPM or PyPI packages.  It may make sense to add additional docs here - but ideally we would not be guiding folks to do this as part of the mainline experience as they shouldn't have to (and we should fix or document any bugs that are currently getting in the way of that).

DaRosenberg:
@lukehoban I first followed the steps in this page just to get Azure CLI 2.0 installed and do the authentication. I then created a new project based on the `typescript-azure` template. I believe the `npm install` happened automatically at this point, because I got full IntelliSense for all the Azure-specific stuff. As soon as I tried to do a `pulumi preview` on that new project however, Pulumi showed an error message which stated that the plugin had to be installed.

DaRosenberg:
@lukehoban I keep getting hit by this, now in a CI/CD scenario on Azure DevOps.

I have my release pipeline set up to first install Pulumi like `curl -fsSL https://get.pulumi.com | sh`, then do an `npm install` and then do a `pulumi up`. All three steps run as expected, but the last one fails like so:

```
2019-01-20T20:54:00.7752180Z ++ pulumi up --yes
2019-01-20T20:54:01.5463334Z Previewing update (retrospect-svcReplika-dev):
2019-01-20T20:54:02.6144812Z error: could not load plugin for azure provider 'urn:pulumi:retrospect-svcReplika-dev::RetrospectSvcReplika::pulumi:providers:azure::default': no resource plugin 'azure-v0.16.6' found in the workspace or on your $PATH, install the plugin using `pulumi plugin install resource azure v0.16.6`
2019-01-20T20:54:02.6200241Z ##[debug]Exit code 255 received from tool '/bin/bash'
2019-01-20T20:54:02.6209895Z ##[debug]STDIO streams have closed for tool '/bin/bash'
2019-01-20T20:54:02.6283375Z ##[error]Bash exited with code '255'.
```

lukehoban:
Sorry for a long delay @DaRosenberg.  

We really want to avoid users having to explicitly run `pulumi plugin install` manually in any normal situation.  So I'd prefer not to take this PR as-is.  I *would* still love to get to the bottom of why plugins aren't getting installed correctly for you.  If you are still seeing this - could you open an issue to track in the https://github.com/pulumi/pulumi repo?  In particular, I'd love to see the output of your `npm install` command in CI, as well as the output of `pulumi plugin ls` after running `npm install`.

Issue #738, opened 2019-01-09
Formatting issue causing nested elements


clstokes:
At https://pulumi.io/reference/pkg/nodejs/@pulumi/aws/config/#insecure it seems an HTML element isn't getting closed or some code elements aren't getting escaped properly.

Theory about the issue from Joe:
> It's generating `__config.getObject<boolean>` and, for some reason, the `<boolean>` bit isn't getting translated to `&lt;boolean&gt;` like it should be.

Screenshot for posterity's sake:
![image](https://user-images.githubusercontent.com/293763/50869467-ede14200-1368-11e9-80ab-72c588893176.png)

PR #739, opened 2019-01-09
HTML escape default values

joeduffy:
I hadn't realized that the DefaultValue fields for let/const
initializers would contain complex expressions, in the form of
strings. By emitting them as-is, we can end up emitting text like

    let foo = __config.getObject<boolean>("foo");

in the middle of a chunk of HTML. The <boolean> part is then
interpreted as an HTML tag, and it leads to awful <div> mismatches
that garble the layout of the remainder of the page.

The fix here is to simply HTML encode those values in the result.

Fixes pulumi/docs#738.


joeduffy:
@clstokes I've regenerated the AWS docs with the fix, so that you can eyeball that it did the right thing; see https://github.com/pulumi/docs/pull/739/commits/74e17ff8c3b3eaf8c3411e9105f2db3da1f81311.

Note that I'm going to regenerate all of the docs as a follow-on commit, which will be a big delta, as we've released newer package versions and so it'll contain new APIs, etc, in addition to the fix.

PR #740, opened 2019-01-09
Expand the AWS examples we have in Docs

CyrusNajmabadi:


CyrusNajmabadi:
Added two more pure AWS tutorials.  @chrsmith is there somethign else i need to do here?

PR #742, opened 2019-01-10
added credentials example for Windows users

casey-robertson:
As a sometimes Windows user myself and to help out others, added the command to set environment variable for that platform.

### DIFF
--- a/quickstart/gcp/setup.md
+++ b/quickstart/gcp/setup.md
@@ -31,8 +31,16 @@ To communicate your credentials to the Pulumi Google Cloud Platform Provider,
 export the contents of your credentials file to the `GOOGLE_CREDENTIALS`
 environment variable:
 
+Linux and Mac OS X
+
 ```bash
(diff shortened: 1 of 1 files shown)


quickstart/gcp/setup.md
```diff
@@ -31,10 +31,18 @@ To communicate your credentials to the Pulumi Google Cloud Platform Provider,
...
+
 ```bash
 export GOOGLE_CREDENTIALS=$(cat credentials.json)
 ```
 
+Windows Powershell
+
+```bash
+$env:GOOGLE_CREDENTIALS="C:\PATH\TO\CREDENTIALS\credentials.json"
+```
```
lukehoban: Does this do the same thing as the `bash` above?  In particular, does it include the contents of the file as the value of the `GOOGLE_CREDENTIALS` environment variable - or does it set the name of the file as the value of the environment variable?

Issue #744, opened 2019-01-12
Overhaul the ToC


joeduffy:
We'd like to overhaul the ToC for the website. This includes

- [x] Make the left hand nav collapsable
- [x] Add "mini ToCs" for complex pages
- [x] Rearrange the ToC
- [ ] Merge disparate sections and remove duplication (Tour/Reference)
- [ ] Turn the Quickstart into a more "hands on tutorial," with a Tour-like style

clstokes:
Looks awesome! I noticed a small strange behavior. If you click the header of a sidebar menu item that's already open, the page kind of "jerks" a little bit and the menu doesn't collapse. Clicking the toggleButton icon works as expected.

<details>
  <summary>GIF</summary>

![screencast 2019-01-13 10-19-10](https://user-images.githubusercontent.com/293763/51089279-7cfea900-171f-11e9-88ac-6b91b48d3d07.gif)

</details>

justinvp:
Should we close this in favor of https://github.com/pulumi/docs/issues/831?

joeduffy:
Will look at this as we overhaul the current architecture next sprint.

joeduffy:
https://github.com/pulumi/docs/issues/831

PR #745, opened 2019-01-12
Make the left nav collapsable

joeduffy:
This change spiffs up the left nav a little bit. This includes a little
bit nicer styling (bigger fonts, light grey background with a border,
etc), but more notably, the ability to collapse sections. This makes
leaping around between sections a lot smoother.

Part of pulumi/docs#744.

### DIFF
--- a/_includes/toc.html
+++ b/_includes/toc.html
@@ -19,7 +19,7 @@ include.tree is the actual set of sections to render. It is an array with shape:
             {% assign path  = found_page.url %}
         {% endif %}
     {% endcapture %}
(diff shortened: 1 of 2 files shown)


joeduffy:
@cnunciato Merging proactively so I can keep chipping away. Post-merge feedback welcome.

cnunciato:
No worries. Only thing that caught my eye was that FAQ and Changelog are collapsible also, so I'm wondering how simple it'd be to disable the toggler for those two.

joeduffy:
Good catch. Let me see what I can do about that before promoting to production.

PR #747, opened 2019-01-12
Hide expand/collapse buttons for empty sections

joeduffy:
This change hides the expand/collapse buttons for sections that
don't have any sub-navigation within them.


joeduffy:
@cnunciato This templating is kind of yucky, but it gets the job done.

image

PR #749, opened 2019-01-12
Add the ability to have "mini" ToCs

joeduffy:
This change introduces the notion of a "mini" ToC, which simply
helps navigate within the H2s in a given page. This is useful for
sections like our Programming Model topic which contains lots of
goodies, but which aren't obvious at a glance. I've only added it
here but will undoubtedly add it in more places as appropriate
throughout the course of tackling pulumi/docs#744.


joeduffy:
For example:

image

PR #752, opened 2019-01-12
Add a sign in button to the top nav

joeduffy:
It's unfortunate that we don't currently have a sign in button
prominently displayed on our docs website. This is a nice "teaser"
to get you to sign up. This change rectifies this, adding it to the
right of the search box. We don't have the necessary horizontal space
on mobile, however, so it's hidden there.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -51,11 +51,12 @@
                             </span>
                         </button>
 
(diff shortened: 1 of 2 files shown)


joeduffy:
image

PR #753, opened 2019-01-12
Remove extra hyphen from string when getting env from stack name

hobbsh:
Basically `env` here ends up with a preceding hyphen when inferring environment from the stack name (i.e. `lumo-infra-staging` results in env being `-staging`). 

I'm no typescript whiz so there is probably a better way of doing this. Overall, maybe it would be better to recommend using a variable in the stack config instead of parsing the stack name? So do a `pulumi config set lumo-infra:environment staging` and:

```
export const config = new pulumi.Config("lumo-infra");
export const env = config.require("environment");
```


lukehoban:
I believe this was fixed as part of some larger changes to this section - closing this out.

PR #755, opened 2019-01-12
Majorly overhaul the ToC

joeduffy:
This changes many aspects of the ToC.

I'm sure I'll miss some of them; however:

* Change the top-level nav to: Get Started, Documentation, and APIs.

* Eliminate the Tour. I'll bring back pieces of it in a subsequent
  commit, but it is largely duplicative with our other material.

* Add a Documentation landing page with a bit of helpful text about
  navigating around, rather than dumping people straight into Concepts.

* Tweak lots of headers so that they are more compelling. For instance,
  rather than "Cloud Console," we call it "Teams and Collaboration."
  In general, the entire service documentation is now more oriented
  around the jobs to be done, rather than being driven by a very
  functional "the service has feature X" point of view.

* Add the mini-ToC to a few more pages that benefit from it.

* Add a section in the Troubleshooting page about using Slack to get help.

* Add the "Install Pulumi" button in a few new areas.

* Add "Install Node.js" and "Install Python" buttons on the language pages.

* Tweak a bit of styling to improve whitespace, padding, etc. Also
  add hanging indentation on the L2 nav topics, so they look cleaner
  in the event that they run over (which we generally try to avoid
  but run into here and there).

* Add Supported Clouds and Supported Languages landing pages.

* Use the language switcher in the Configuration section, and also
  add a Go example.


joeduffy:
This isn't very easy to discern from the commits themselves. Here's a screenshot:

image

PR #761, opened 2019-01-15
Python documentation generator

swgillespie:
This PR contains a work-in-progress documentation generator for Python. A new tool, `pydocgen`, uses Sphinx and some light templating in order to generate a tree of Markdown files suitable as input to Jekyll, which in turn can be integrated to our documentation in the same manner as our NodeJS docs.

The magic happens in `tools/pydocgen/pydocgen/__init__.py`, so start there if you want understand what's going on here. I have tried to document that script as best I can, so I hope that it is self-explanatory enough that I don't have to write much within this PR to describe how it works, but the upshot is that `pydocgen` is a Python script that drives Sphinx in order to auto-generate a lot of documentation from our existing Python packages.

It's currently not ready for primetime (and thus WIP) for a few reasons:
1. The generated documentation for TF-based providers is very bad without https://github.com/pulumi/pulumi-terraform/pull/304
2. I haven't tried to generate documentation for `pulumi_kubernetes` but I expect that there will be some work getting the kubernetes code generator to generate sphinx-compatible docstrings
3. The resulting HTML output by `pydocgen` needs to be styled in order to not look terrible in the context of our website, so there's a non-trivial amount of CSS work to be done before we can be happy with the result.

However, I would appreciate people's reviews and thoughts on the process here. It's certainly ad-hoc but based on my preliminary results it definitely will get the job done for us.

In this PR, you'll see references to `pulumi_random` and `pulumi_aws` as the only providers - this is because I hacked them up locally to build with my `pulumi-terraform` changes. In the future, we will include all providers that we intend to generate documentation for in this list. I picked `pulumi_random` and `pulumi_aws` because they are vastly different in terms of directory structure (`pulumi_aws` has many submodules, `pulumi_random` does not), so it provides a good test of `pydocgen` operating in two different modes.


swgillespie:
Also worth noting is that `pydocgen` does not handle the `pulumi` package particularly well, since it is tuned to autogenerate documentation from providers. I think that our best bet going forward is to manually provide Sphinx with input files for `pulumi` and run it through the rest of the Sphinx pipeline. This is a minimal amount of work and provides us with a high degree of control over the output of the `pulumi` documentation.

joeduffy:
I'm curious to hear your thoughts on how to integrate this into our overall docs structure.

For instance, the left hand nav currently lists all packages

image

And then the page you land on gives you a switcher at the top

image

...which of course currently says that Python is unavailable

image

It definitely feels like we'll want to make it easy to switch between languages -- versus, for instance, having _two_ package links everywhere (one to the Node.js docs, one to the Python docs) -- but the specific switcher UI element isn't quite right as it stands.

Btw, if we like the switcher UI, and just want to make it link to a different page when you click on one language or the other, this wouldn't be very difficult to do -- I'm sure either @justinvp or I could do it.

swgillespie:
> Btw, if we like the switcher UI, and just want to make it link to a different page when you click on one language or the other, this wouldn't be very difficult to do -- I'm sure either @justinvp or I could do it.

I wouldn't say that design is my strong suit, but my first inclination is to go down this route, at least for now until we get a feel for how it looks. Is it hard to do this?

joeduffy:
@justinvp Could you take a look at this possibility?

justinvp:
> @justinvp Could you take a look at this possibility?

Sure. Are we aiming to land this in the next couple of days or can it wait until sometime next week?

swgillespie:
> Didn't follow -- why is this work serialized?

I made changes to tfgen to emit doc comments that this tool is capable of ingesting (https://github.com/pulumi/pulumi-terraform/pull/304). I'm unable to bring those changes to upstream TF repos until https://github.com/pulumi/pulumi-terraform/issues/308 is fixed, which @jen20 is going to do as part of https://github.com/pulumi/pulumi-terraform/issues/306.

This tool generates terrible documentation for packages without https://github.com/pulumi/pulumi-terraform/pull/304. We could hack it up locally such that we use a fork of pulumi-terraform to build a bunch of packages and then style the HTML that comes out of those. If we're pressed for time, we could do that, but it's been a pain to do that on my machine while maintaining a sane dev environment.

I could just generate a lot of markdown using the above method and then send it to you @justinvp - would that be sufficient for you?

swgillespie:
At this point I believe that this PR is ready for review. I recommend checking out the repo and firing up `jekyll serve` so that you can take a look at what these changes look like in a web browser.

First, I want to stress that there is still a lot more work to be done until we can be satisfied here:

1. This PR does not include documentation for `pulumi`, the Python SDK, since the documentation that it generates with the tool in this PR looks quite bad. We need to take a few hours to sit down and convince Sphinx to produce high-quality documentation for the `pulumi` package. This is not hard, but it does take time.
2. The documentation that this tool generates is at times buggy. There are likely going to be a smattering of bug fixes in both this tool and `pulumi-terraform` in order to get satisfactory results.
3. Sphinx emits a lot of warnings. This implies that there are bugs in `pulumi-terraform` around docstring generation that we will need to fix.

Furthermore, we'll probably want to think even broader about what form we'd like the docs to take on this website. I was unable to accomplish the "language switcher" approach in a reasonable amount of time so I instead opted to split the packages by language in the sidebar. I think that this is sufficient for now so that we can get this checked-in and begin iterating on it.

In summary, this PR gets us reasonable documentation for the `aws`, `azure`, and `gcp` packages, but we will need to continue to invest in Python documentation in order for the user experience to not be terrible. I do think that this PR is a good start and we should merge it, so that we can begin improving it.

_sass/_pulumi.scss
```diff
@@ -929,3 +929,109 @@ h4 { font-size: 22px; font-weight: 500; }
...
+// Our style selectors are split into two toplevel selectors: a div starting with "module-",
+// and a div with class "section" whose ID begins with "pulumi". These two things identify
+// the two kinds of pages that Sphinx emits: the first kind is a module page, containing
+// information on on members within a module, and the second kind is a table of contents.
+
+// This selector is for the module page.
+div[id^="module-"] {
+    // Don't show the header link - there's an anchorjs link on this header that does the
+    // job just as well.
+    h1[id$="¶"] {
```
justinvp: BTW, I was running into the following error when trying to build:

```
  Conversion error: Jekyll::Converters::Scss encountered an error while converting 'css/styles.scss':
                    Invalid US-ASCII character "\xC2" on line 950
```

Looks like [escaping](https://www.w3.org/International/questions/qa-escapes#cssescapes) the character fixes the issue:

```suggestion
    h1[id$="\B6"] {
```

joeduffy:
> This PR does not include documentation for pulumi, the Python SDK, since the documentation that it generates with the tool in this PR looks quite bad. 

I would rather have bad documentation than no documentation, honestly.

> I instead opted to split the packages by language in the sidebar. 

I'll take a look at this to see what it looks like, but I'd really like to avoid complicating our nav story again, since we just simplified it. I can probably implement the switcher as a follow up to your change.

joeduffy:
Oh one quick question, I pulled locally and am running and  I only see four packages. Are the others on their way or am I just missing something?

swgillespie:
Per discussions with people, I'm going to go ahead and merge this PR as-is. We can incrementally improve our documentation as we go.

I'll open up an issue with concrete follow-up tasks.

PR #764, opened 2019-01-16
Add info about GitLab-backed Pulumi organizations.

### LINKED ISSUE #754: Document GitLab support
https://pulumi.io/reference/service/editions.html mentions GitHub, but not GitLab.

praneetloke:
Resolves #754. Here's a preview of how these changes look when they are rendered.

screenshot_2019-01-16 organizations

### DIFF
--- a/reference/service/orgs.md
+++ b/reference/service/orgs.md
@@ -18,11 +18,52 @@ Adding a new Pulumi organization can be done directly from the Pulumi Cloud Cons
     <button class="button">ADD ORGANIZATION</button>
 </a>
 
-## GitHub-backed Organizations
+## GitHub and GitLab-backed Organizations
 
-Organizations using the Pulumi Team and Enterprise editions may be backed by a GitHub organization,
-enabling you to manage user access and team permissions by using your existing GitHub access
+Organizations using the Pulumi Team and Enterprise editions may be backed by a GitHub organization or GitLab group,
+enabling you to manage user access and team permissions by using your existing access
 controls.
+
+Here's a table to help you map the terms used by GitHub and GitLab to Pulumi's nomenclature.
+
+<style>
+table {
+    margin-bottom: 20px;
+}
+
+td, th {
+    padding: 8px 8px;
+    text-align: center;
+    border: 1px solid rgba(0,0,0,0.13);
+}
+
+thead tr th {
+    color: #00acf2;  /* $primary2, blue */
+    font-weight: 800;
+}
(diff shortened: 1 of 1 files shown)


praneetloke:
Ping @chrsmith.

joeduffy:
LGTM.

One minor idea, should the vendor specific terms in the table be hyperlinked to the GH/GL docs for that concept?

praneetloke:
@joeduffy what do you think of this?

The links I added are (in-order):
https://about.gitlab.com/2017/09/11/comparing-confusing-terms-in-github-bitbucket-and-gitlab/

https://github.com/collab-uniba/socialcde4eclipse/wiki/How-to-setup-a-GitHub-organization,-project-and-team

https://docs.gitlab.com/ce/user/group/

screenshot_2019-01-18 organizations

reference/service/orgs.md
```diff
@@ -18,12 +18,53 @@ Adding a new Pulumi organization can be done directly from the Pulumi Cloud Cons
     <button class="button">ADD ORGANIZATION</button>
 </a>
 
-## GitHub-backed Organizations
+## GitHub and GitLab-backed Organizations
```
chrsmith: Since you go into details about the different 3rd party organization types below (the table is super-useful), we can probably make the heading more generic. Perhaps the section heading could be "Imported Organizations", or "Importing from 3rd Party Services"?

reference/service/orgs.md
```diff
@@ -18,12 +18,53 @@ Adding a new Pulumi organization can be done directly from the Pulumi Cloud Cons
...
 
-Organizations using the Pulumi Team and Enterprise editions may be backed by a GitHub organization,
-enabling you to manage user access and team permissions by using your existing GitHub access
+Organizations using the Pulumi Team and Enterprise editions may be backed by a GitHub organization or GitLab group,
```
chrsmith: Again, I'd just drive users to the table below, since that is the most efficient way we'll communicate the types of organizations we support. And also how Teams would map.

Perhaps:

"""Organizations using the Pulumi Team and Enterprise editions may be backed by an external service, such as GitHub or GitLab. You can enforce user access and team permissions in the Pulumi organization by using your existing access controls on another service.

Here's a table describing the mapping between supported 3rd party services and how they are mirrored into Pulumi:"""

Does it make sense to add another row with "Team" as well? Which IIRC in the GitLab case is "Sub-group"?

PR #765, opened 2019-01-17
Update for pulumi/pulumi 0.16.11

ellismg:


joeduffy:
LGTM. A few notes

* I didn't see any reference to the new `interpolate`, etc functions in the changelog. @CyrusNajmabadi 

* Do we have any conceptual documentation we are releasing for stack tagging? A blog? We should ship a new major feature with at least one or the other -- ideally both. @justinvp

ellismg:
There is a comment about the interpolation functiona as the first item in the changelog.

https://github.com/pulumi/pulumi/blob/v0.16.11/CHANGELOG.md

lukehoban:
> Any reason we aren't doing this?

We didn’t do this initially because it can’t be done as part of the single initial commit for a PR (which would need to reference itself).  We thought about adding tooling to automate this based on some convention, but felt that was complexity overkill. 

I think at this point we should just:
1. In general, add additional commits after PR creation to link back to PR
2. During cleanup and prep for release add any missing PR links (like we do for missing line items)

I think both are in practice relatively low cost and easy to adhere to if everything else already has links.

joeduffy:
That sounds good. I actually meant the issues, not the PRs, which generally do exist before the commit. But both are interesting for sure

PR #768, opened 2019-01-17
Update programming-model.md to include info on concat/interpolate.

CyrusNajmabadi:


reference/programming-model.md
```diff
@@ -184,6 +184,88 @@ def split(input):
 // See https://github.com/pulumi/pulumi/issues/1614.
 ```
 
+#### Working with Outputs and strings {#ouputs-and-strings}
+
+It's very common want to build a string to use out of the values contained in `Outputs`.  Common uses for this are to either provide a custom [stack output](#stack-outputs), or to provide a dynamically compute string as an [Input](https://pulumi.io/reference/pkg/nodejs/@pulumi/pulumi/#Input) to another Resource.  For example, say you had the following:
```
swgillespie: grammar: "dynamically computed"

reference/programming-model.md
```diff
@@ -184,6 +184,88 @@ def split(input):
...
+// See https://github.com/pulumi/pulumi/issues/2366.
+```
+
+```go
+// Helpers for combining Outputs into strings are not yet available in Go.
+// 
+// See https://github.com/pulumi/pulumi/issues/1614.
+```
+
+However, this is quite verbose and unweildy.  To make this easier, Pulumi exposes two helpers `concat` and `interpolate`  to make this more convenient.  They can be used as follows:
```
swgillespie: spelling: "unwieldy"

reference/programming-model.md
```diff
@@ -184,6 +184,88 @@ def split(input):
...
+```typescript
+const hostName: Output<string> = // get some Output
+const port: Output<number> = // get some Output
+
+// Would like to produce a string equivalent to: http://${hostname}:${port}/
+const url: Output<string> = // ?
+```
+
+```python
+// Helpers for combining Outputs into strings are not yet available in Python.
```
swgillespie: It'll render better if we use `#` for comments + python syntax highlighting

Issue #769, opened 2019-01-18
Serverless events are missing from docs


joeduffy:
Some combination of the way we generate the `d.ts` files for serverless events -- using overlays -- plus the doc generator means that we don't have `on*` methods in our documentation! 😦

PR #770, opened 2019-01-20
Merge nodes and modules upon conflicts

joeduffy:
This changes our TypeScript doc generation to merge modules and their
constituent nodes upon a name collision. This happens with the way we
do monkey patching of our serverless events, but is a general problem
that can come up with various legal TypeScript patterns.

For now, this simply entails:

- Treat `declare module { ... }` as being exported; prior to
  this change, we assumed it was internal, and skipped over them.

- Normalize names like "./x" to being a child of the parent module.

- Merge conflicting modules.

- Merge conflicting type nodes. The only legal combinations we
  support right now are interface+interface, class+interface, and
  interface+class. We prefer to choose the class as the "primary"
  for purposes of display (so it shows up as `class Foo` with
  its constructor function, and so on), and generally merge the
  child nodes, interfaces implemented, and source locations.

This fixes pulumi/docs#769. I will re-generate the docs with this
fix in a separate commit after various other changes have landed.

### DIFF
--- a/tools/tscdocgen/main.go
+++ b/tools/tscdocgen/main.go
@@ -269,15 +269,12 @@ func (e *emitter) emitMarkdownModule(name string, mod *module, root bool) error 
 
 		members = append(members, member)
 	}
-	sort.Strings(files)
-	sort.Slice(members, func(i, j int) bool {
-		if members[i].Label != members[j].Label {
-			return members[i].Label < members[j].Label
-		}
-		return members[i].Name < members[j].Name
-	})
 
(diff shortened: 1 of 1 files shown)


joeduffy:
Note: I am fairly certain I didn't implement the full module merging rules (unless I got ridiculous lucky -- unlikely), but this is enough to get through all of our packages.

joeduffy:
Merging as this seems to fix everything. Happy to respond to CR feedback afterwards!

PR #771, opened 2019-01-20
Improve the code colorization

joeduffy:
This fixes some anomalies in our code colorization, while also
generally adopting VS Code's default "light" color theme.


joeduffy:
Example before:

image

Example after:

image

PR #783, opened 2019-01-22
Correct typo for mounting a volume

tvalenta:
s;moung;mount;g


lukehoban:
Thanks @tvalenta!

These docs are generated from source code - so to really fix this we'll need to fix https://github.com/pulumi/pulumi-cloud/blob/2de4bd3b77567899f995fce87fe7c661d4b4b91f/api/service.ts#L97 instead/as well.

tvalenta:
Ah, of course!

tvalenta:
Ah, of course!

Issue #736, opened 2019-01-08
KTPW Lab2 doesn't document required config variables


lblackstone:
Running `pulumi up` as specified in https://pulumi.io/quickstart/k8s-the-prod-way/infrastructure.html will result in errors if required config values are not set first. Document how to do this.

hausdorff:
Hmm, this documentation actually _does_ tell you how to set the appropriate config values, @lblackstone? I'm not sure what's missing.

lblackstone:
Yep, looks like it's there now. Not sure if it was added, or if I just missed it...

PR #786, opened 2019-01-24
Remove KtPW from the TOC

hausdorff:


lukehoban:
Should we fix https://github.com/pulumi/docs/issues/736 as part of this?

PR #795, opened 2019-01-26
Update the SAML docs with settings for IDP-initiated SAML flows

praneetloke:
- Update the Okta doc to instruct users to set the Default Relay State in their config.
- Update the GSuite doc to instruct users to set the Start URL.

Both of these changes are related to supporting IDP-initiated SAML flows involving the RelayState param. An IDP-initiated login is one where the SP is not soliciting the IDP for authentication, instead, it is the other way around, the IDP allows the user to login into the IDP portal and then click on "shortcuts" to apps their identity is integrated with. This causes the IDP to call the ACS URL with a POST request. At this point, the SAML lib that we use would fail since it cannot identify the request ID, since there isn't one as it was not initiated by the SP.

SP = Service Provider (Pulumi Cloud Console)
IDP = Identity Provider (Okta, GSuite etc.)

### DIFF
--- a/reference/service/saml-gsuite.md
+++ b/reference/service/saml-gsuite.md
@@ -40,6 +40,7 @@ Pulumi organization were `acmecorp`, those values would be:
 
     * ACS URL: `https://api.pulumi.com/login/acmecorp/sso/saml/acs`
     * Entity ID: `https://api.pulumi.com/login/acmecorp/sso/saml/metadata`
+    * Start URL: `https://api.pulumi.com/login/acmecorp/sso`
 
     ![Step 4: Provide ACS and metadata URLs](../../images/reference/service/saml-gsuite/gsuite-dialog-step-4.png)
 
--- a/reference/service/saml-okta.md
+++ b/reference/service/saml-okta.md
@@ -49,6 +49,7 @@ tbody tr td {
 | --------------- | ----- |
 | Single sign on URL | https://api.pulumi.com/login/robot-co/sso/saml/acs |
 | Audience URI | https://api.pulumi.com/login/robot-co/sso/saml/metadata |
+| Default Relay State | https://api.pulumi.com/login/robot-co/sso |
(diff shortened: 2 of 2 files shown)


reference/service/saml-gsuite.md
```diff
@@ -40,6 +40,7 @@ Pulumi organization were `acmecorp`, those values would be:
 
     * ACS URL: `https://api.pulumi.com/login/acmecorp/sso/saml/acs`
     * Entity ID: `https://api.pulumi.com/login/acmecorp/sso/saml/metadata`
+    * Start URL: `https://api.pulumi.com/login/acmecorp/sso`
```
cnunciato: Given the way this reads, I'd probably suggest pulling it into its own line, maybe just under the image, since these two bullets and the graphic refer to the required fields, and this third field is both optional and useful enough to warrant a bit of explanation. (We used Okta at Chef, and these shortcuts came in really handy! I didn't realize Google could do something similar.) Maybe something like:

```
    ![Step 4: ...step-4.png)

   If you'd like to let your users launch the Pulumi Cloud Console from within the Google Apps menu, you can set an optional **Start URL**  as well &mdash; for example, `https://api.pulumi.com/login/acmecorp/sso`. 

   Leave the other fields ...
```

...or the like?

chrsmith: +1 to having a little more copy about the IdP initiated login scenario.

If this Start URL is just POSTed to directly from the IdP, we could work around the library issue by just having it go directly to the `https://**app.pulumi.com**/welcome/acmecorp/sso` (And just treat the POST like a GET.) Do I have that right? (I like the approach you're taking of fixing the issue in the SAML library, just trying to understand the issue.)

praneetloke:
Closing for now. Will re-open this when some related changes in the service is ready to be merged as well. We are trying to get a change upstream'd into a dependent library.

PR #797, opened 2019-01-28
Minor edit: add "go" language.

andrewlayman:


lukehoban:
Merging for now, though we will likely go the other direction soon and remove more mentions of Go.  But better to be consistent for now.

PR #799, opened 2019-01-28
Minor edits.

andrewlayman:
But please verify the formatting for the escaped backets in \[secret\].


lukehoban:
I actually don't think we want to take the changes here - some are technically not correct, and some of the cross linking won't work correctly as written (it is trying to cross link across pages I believe).  Closing this out, and we can open more targetted PRs if needed.

PR #802, opened 2019-01-28
Minor edit to How Pulumi Works

andrewlayman:


lukehoban:
I don't see any actual diff here - so will close this out.

PR #804, opened 2019-01-28
Clarify the Paths example

andrewlayman:


lukehoban:
I think it's okay to leave this out - it distracts from the topic-at-hand, and is easy enough to look up in search if needed.

PR #805, opened 2019-01-28
Updates for v0.16.12

ellismg:


ellismg:
Going to go ahead with this for now, @lukehoban.  The genie is a bit out of the bag on the new CLI anyway, since an external contributor helpfully bumped the homebrew version over the weekend.  Might as well get our website updated as well.

PR #808, opened 2019-01-28
Clarify the each project has 1 program

andrewlayman:


lukehoban:
The following sentence mentions "for the Pulumi program" which I think makes this clear enough - and we'd probably want to go into more depth to flesh out this relationship more.  I'll close this one out for now - but we may want to find another place to make this relationship more clear.

PR #809, opened 2019-01-28
Typo correction

andrewlayman:


andrewlayman:
One typo was missed.  I'll make the correction in the code.

PR #816, opened 2019-01-30
Rework StackReferences documentation

ellismg:
When constructing a stack reference the string must now be of the form
`<organization>/<project>/<stack>`. I have rewritten the content here
to use this new form, as well as simplified the example so it exploits
the fact that stack names may now be duplicated across projects to
remove some of the extra logic that was not core to the example.

### DIFF
--- a/reference/organizing-stacks-projects.md
+++ b/reference/organizing-stacks-projects.md
@@ -80,14 +80,14 @@ you may have inter-dependencies between your stacks -- something that Pulumi sup
 
 ## Inter-Stack Dependencies
 
-Let's imagine you decide to define your cluster infrastructure in one project and consume it from another.
-Perhaps one project, `acmecorp-infra`, defines your Kubernetes cluster and another, `acmecorp-services`, deploys
+Let's imagine that acmecorp decides to define its cluster infrastructure in one project and consume it from another.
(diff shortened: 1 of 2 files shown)


ellismg:
After deploying this change to production, I plan to link to the updated page from the `CHANGELOG.md` file in pulumi/pulumi

reference/organizing-stacks-projects.md
```diff
@@ -104,16 +104,19 @@ The Pulumi programming model offers a way to do this with its `StackReference` r
...
+const infra = new pulumi.StackReference(`acmecorp/infra/${env}`);
 const provider = new k8s.Provider("k8s", { kubeConfig: infra.getOutput("kubeConfig") });
 const service = new k8s.v1.core.Service(..., { provider: provider });
 ```
 
-The `StackReference` constructor takes as input the name of another stack from which to fetch outputs. We
-need to do a little parsing here to map, say, `acmecorp-services-production` to `acmecorp-infra-production`.
-But once we have that resource, we can fetch the `kubeConfig` output variable with the `getOutput` function.
-From that point onwards, Pulumi understands the inter-stack dependency for scenarios like cascading updates.
+The `StackReference` constructor takes as input the a string of the form `<organization>/<project>/<stack>`, and lets
```
lukehoban: `the a` => `a`

PR #822, opened 2019-01-30
Describe the source of the snapshot file.

andrewlayman:


andrewlayman:
I'll make the change in the preview.go file.

PR #824, opened 2019-01-30
Explain what stacks are in more detail.

andrewlayman:


lukehoban:
I don't think we'll want to make this change to the core CLI docs - so closing this one out.

PR #829, opened 2019-01-31
Keep API Reference open when clicking a package

justinvp:
And make sure the Node.js and Python items show as selected when chosen.


joeduffy:
Magic ✨

If it works, :+1:, LGTM to me.

PR #830, opened 2019-01-31
Deploy to Production: Keep API Reference open when clicking a package

justinvp:
And make sure the Node.js and Python items show as selected when chosen.


justinvp:
I plan on merging this when green. Fix can be seen at https://staging.pulumi.io

Issue #791, opened 2019-01-25
Pulumi Python SDK needs documentation


swgillespie:
We need to document the `pulumi` package in addition to the provider documentation already present on the site. The doc generating tool in this repo generates poor documentation for `pulumi` and so we'll likely need to craft some Sphinx input by hand.

swgillespie:
Fixed by https://github.com/pulumi/docs/pull/833.

PR #833, opened 2019-01-31
Add Pulumi Python SDK Docs

swgillespie:
This PR augments the documentation generator to generate documentation
for the Pulumi package in addition to providers. Instead of
autogenerating all documentation as we do for providers, this commit
opts to autogenerate only the API documentation, while interspersing it
with prose.

In addition, the addition of the Pulumi Python SDK to Sphinx's input
allows Sphinx to intelligently generate cross-reference links for types
such as pulumi.ResourceOptions. There's not a whole lot of cross-linking
happening, but the fact that we're now doing it a little bit
demonstrates that it is possible.


swgillespie:
Key files in this PR:

* [`pulumi.rst`](https://github.com/pulumi/docs/compare/swgillespie/pulumi-docs?expand=1#diff-d5bf7c984b191b89c451585290f7d7cc): This is a Sphinx input file that generates the Pulumi Python SDK documentation.
* [`pydocgen/__init__.py`](https://github.com/pulumi/docs/compare/swgillespie/pulumi-docs?expand=1#diff-1439ebcb9150ab1b09b0cf30686e5e45): There are minor changes in this PR to the documentation tool that generates the input that we're feeding directly to Jekyll.

swgillespie:
Fixes https://github.com/pulumi/docs/issues/791.

PR #844, opened 2019-02-04
Fix documentation for packages mixing __all__ and top-level resources 

swgillespie:
This PR contains a few fixes to the documentation generator tool:

1. The `config` module in `tfgen`-generated packages can't be safely imported outside of a Pulumi program. Because of this, Sphinx also can't import it and is unable to generate documentation for it. It's better for us in this case to simply skip it, which this PR now does.
2. When rendering single-package modules (e.g. `pulumi_random`, `pulumi_vsphere`, etc, any module that doesn't have any submodules other than `config`), we must pass `ignore-module-all` to Sphinx, since Sphinx uses `__all__` to determine the public surface area of a module and we are lying a little bit about what our public surface area is.

There are a few diffs in existing docs with these changes, but more importantly this unblocks the addition of docs for a few other packages that weren't working before (in a followup PR).

### DIFF
--- a/tools/pydocgen/pydocgen/__init__.py
+++ b/tools/pydocgen/pydocgen/__init__.py
@@ -254,6 +254,5 @@ def main():
         if path.exists(tempdir):
             shutil.rmtree(tempdir)
         if path.exists(outdir):
-            pass
             shutil.rmtree(outdir)
 
--- a/tools/pydocgen/pydocgen/__init__.py
+++ b/tools/pydocgen/pydocgen/__init__.py
@@ -152,7 +152,7 @@ def generate_sphinx_files(ctx: Context):
 def should_generate_multimodule(module):
     """
     Returns whether or not we should generate a multi-page page tree for this particular module or if we should fit
-    evernthing on a single page.
+    everything on a single page.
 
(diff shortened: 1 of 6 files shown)


swgillespie:
@lukehoban This fixes the issues you were seeing last friday when trying to add a few new packages to the Python doc generator.

tools/pydocgen/pydocgen/__init__.py
```diff
@@ -134,19 +134,38 @@ def generate_sphinx_files(ctx: Context):
...
             create_dir(ctx.tempdir, "providers", provider.package_name)
             for module in all_modules:
                 dest = path.join("providers", provider.package_name, f"{module}.rst")
                 module_meta = {"name": module, "full_name": f"{provider.package_name}.{module}"}
                 render_template_to(ctx, dest, module_template, module=module_meta)
 
+def should_generate_multimodule(module):
+    """
+    Returns whether or not we should generate a multi-page page tree for this particular module or if we should fit
+    evernthing on a single page.
```
justinvp: Nit:

```suggestion
    everything on a single page.
```

PR #858, opened 2019-02-08
Small fixes to k8s docs

hausdorff:


_data/quickstart.yaml
```diff
@@ -30,7 +30,6 @@ toc:
   path: /quickstart/kubernetes/index
   section:
    - quickstart/kubernetes/setup.md
-   - quickstart/kubernetes/tutorial-exposed-deployment.md
```
lukehoban: I don't think we want to make this change do we?

lukehoban:
I don't think we want these changes - please re-open if these are intentional.

PR #869, opened 2019-02-09
GKE & EKS: add clarifying info based on TGIK8s feedback

metral:


metral:
@joeduffy @lukehoban I believe I've incorporated all feedback from TGIK and listed here. PTAL

_includes/gcp-gke-prereqs.md
```diff
@@ -4,8 +4,12 @@
 1. [Install Node.js version 6 or later](https://nodejs.org/en/download/)
 1. Install a package manager for Node.js, such as [npm](https://www.npmjs.com/get-npm) or [Yarn](https://yarnpkg.com/en/docs/install).
 1. [Install Google Cloud SDK (`gcloud`)](https://cloud.google.com/sdk/docs/downloads-interactive)
-1. Configure Auth Options
-    1. `gcloud` Login
+1. Configure GCP Auth
+
+    Choose **either option** below to configure your local machine with GCP
```
joeduffy: I think this is fine. I do wonder, however, if we should even be presenting an option. An alternative is to do something more like:

Login using `gcloud`

```bash
... etc etc etc ...
```

> Note that this auth mechanism is meant for inner loop developer workflows. If you want to run this example in an unattended service account setting, such as in CI/CD, please [follow instructions to configure your service account].

I can just still see people going down the rabbit's hole of (2) and for a tutorial, I suspect it is seldom needed.

quickstart/aws/tutorial-eks.md
```diff
@@ -11,22 +11,45 @@ In this tutorial, we'll launch a new Managed Kubernetes cluster in Elastic Conta
...
     ```bash
     $ pulumi new aws-typescript --dir eks-hello-world
     ```
 
-    Enter in a project name, description, stack name, and AWS region to
-    deploy into. We recommend using `us-west-2` to host your EKS cluster.
+    * Enter in a Pulumi project name, and description to detail what this
+      Pulumi program does
+    * Enter in a name for the [Pulumi stack](https://pulumi.io/reference/stack.html), which is an instance of our Pulumi program, and is used to distinguish amongst different development phases and environments of your work streams.
+    * Select 'no' when prompted to 'perform this update,' as we'll be
```
joeduffy: @justinvp Is the plan to stop prompting to do an update? This was really confusing for Joe and I still find it very jarring personally, since 9/10, I want "no".

quickstart/gcp/tutorial-gke.md
```diff
@@ -401,3 +377,7 @@ Managed Kubernetes cluster on GCP GKE.
 
 For a follow-up example on how to use Pulumi programs to create a Kubernetes
 apps on your new cluster, see [Kubernetes Tutorial: Getting Started With Pulumi](../kubernetes/tutorial-configmap-rollout.html).
+
+We also encourage you to watch Joe Beda, co-founder of Kubernetes and Heptio,
+[take Pulumi for a spin on GKE](https://youtu.be/ILMK65YVSKw) in an
+episode of [TGIK8s](https://github.com/heptio/tgik).
```
lukehoban: Should we just embed the YouTube video inline here.

metral: Embedding [doesn't seem possible](https://stackoverflow.com/a/16079387/2274669) in markdown. Do you have a suggestion?

lukehoban: I think you can just include the inline HTML.  For example: https://stackoverflow.com/questions/10529859/how-to-include-video-in-jekyll-markdown-blog

metral: Ah ok, wasn't aware we used jekyll. I've gone ahead and included the inline HTML in the most recent commit

quickstart/aws/tutorial-eks.md
justinvp: I also find it jarring. It's on my list of things to address in the CLI this milestone.

PR #871, opened 2019-02-11
Deploy latest to production

lukehoban:


lukehoban:
> I'd like to try to sneak in [pulumi/examples#229](https://github.com/pulumi/examples/pull/229) into this push to prod if possible to incorporate the TGIK feedback for GKE and EKS once it passes tests & is approved. But we can also do that in a future PR if needed.

Isn't that a change on the `examples` repo, which happens independently?  Or was there another part of that to follow up in the docs site as well?

Either way - we should just push to production after effectively every change here - no need to wait on anything in particular.

metral:
Sorry, linked the wrong PR. SGTM

PR #873, opened 2019-02-11
Add a ref to `Pulumi.<stack>.yaml` to pulumi_config.md

andrewlayman:
Add a reference to `Pulumi.<stack name>.yaml` to pulumi_config.md.


lukehoban:
I expect we won't want to make this change in the CLI docs so closing this one out.

PR #879, opened 2019-02-12
Add the filename arg to the descr of stack graph

andrewlayman:
Add the required filename argument to the description of `pulumi stack graph` in file` pulumi_stack_graph.md`.


lukehoban:
All of the docs under `/reference/cli` are auto-generated.  We'll need to make these fixes in https://github.com/pulumi/pulumi instead (and then release a new CLI with those fixes, and then come back here and regenerate docs for that release).

lukehoban:
I don't think this is compatible with how these docs are autogenerated so closing this out.

PR #881, opened 2019-02-12
Az login before pulumi new in tutorial-container

andrewlayman:
Move the instruction to 'Az login' before 'pulumi new' in file tutorial-container-webs

### DIFF
--- a/quickstart/azure/tutorial-container-webserver.md
+++ b/quickstart/azure/tutorial-container-webserver.md
@@ -11,7 +11,8 @@ In this tutorial, we'll use JavaScript to deploy a simple webserver Container In
 
 ## Create a Virtual Machine with SSH access {#webserver}
 
-1.  In a new folder `webserver`, create an empty project with `pulumi new`. Make sure you have run `az login` or configured credentials for Azure.
+1.  Make sure you have run `az login` or configured credentials for Azure.
+1.  In a new folder `webserver`, create an empty project with `pulumi new`. 
     ```
     $ pulumi new azure-javascript --dir webserver
     $ cd webserver


quickstart/azure/tutorial-container-webserver.md
```diff
@@ -11,7 +11,8 @@ In this tutorial, we'll use JavaScript to deploy a simple webserver Container In
 
 ## Create a Virtual Machine with SSH access {#webserver}
 
-1.  In a new folder `webserver`, create an empty project with `pulumi new`. Make sure you have run `az login` or configured credentials for Azure.
+1.  Make sure you have run `az login` or configured credentials for Azure.
```
lukehoban: This is covered as part of the `Prerequisites` above under the [Configure Azure credentials] link.  (and technically, it is not required depending on which approach you take for configuring your Azure credentials).

I'm not sure it makes sense to re-include everything there into this (and all other) tutorial texts.  We could consider make the "Prerequisites" more of proper steps here so that it's more clear they have to be completed.

I'd also generally love to maintain some horizontal consistency across these tutorials - so any change like this we would ideally make consistently across all tutorials.

andrewlayman: I agree re having horizontal consistency. Would you like me to create an include file for the Azure prerequisites and another for AWS and each other cloud used in a tutorial?

How about I change the text in the prerequisites section to say "Configure Azure credentials or Login to Azure"?

justinvp:
Thanks, Andrew, but we'll leave this one as-is for now. We're planning some improvements across all tutorials.

Issue #790, opened 2019-01-25
Python docs contain unrendered Markdown 


swgillespie:
`tfgen` emits documentation strings in Markdown format and, unfortunately, Sphinx wants them to be in reStructuredText (as is the standard in the Python ecosystem). This results in [the docs sometimes containing unrendered Markdown](https://staging.pulumi.io/reference/pkg/python/pulumi_gcp/appengine/).

We have written a [Sphinx extension that converts Markdown to RST](https://github.com/pulumi/docs/blob/master/tools/pydocgen/pydocgen/templates/markdown_docstring.py), but it is currently turned off because tfgen emits parameter names that are valid Markdown (`__name__` and `__opts__`), which breaks Sphinx.

We should consider 1) being more intelligent in tfgen when generating docstrings and perhaps attempt to convert them to RST or 2) rename `__name__` and `__opts__` to not be Markdown and instead have normal names. We should arguably be doing 2) anyway.

Issue #793, opened 2019-01-25
GCP docs have random HTML tags strewn about


swgillespie:
For example:

<img width="1023" alt="screen shot 2019-01-25 at 10 52 53 am" src="https://user-images.githubusercontent.com/1871912/51766428-66494200-208f-11e9-81a7-1f68a9bc4fed.png">
<img width="1055" alt="screen shot 2019-01-25 at 10 52 30 am" src="https://user-images.githubusercontent.com/1871912/51766429-66494200-208f-11e9-8154-f86bcba25e90.png">

Obviously this is not great - we should figure out what's going on here.

swgillespie:
Fixed by https://github.com/pulumi/docs/pull/886.

PR #886, opened 2019-02-13
Convert Markdown in docstrings to RST

swgillespie:
Since TFGen most often generates docstrings in Markdown, this commit
dynamically translates a docstring's Markdown into the reStructuredText
format that Sphinx expects. The result is that the Sphinx-generated
documentation generally looks the way that Markdown would - links render
correctly, notes get turned into blockquotes, and in general the docs
don't have inline unrendered Markdown in them.

Fixes https://github.com/pulumi/docs/issues/790


swgillespie:
The only code change here is uncommenting a single line in `conf.py` that activates [this file](https://github.com/pulumi/docs/blob/master/tools/pydocgen/pydocgen/templates/markdown_docstring.py), which is a Sphinx extension that hooks the docstring parsing process and converts the Markdown within them to reStructuredText on-the-fly. It was previously disabled until https://github.com/pulumi/pulumi-terraform/pull/320 landed.

swgillespie:
Also fixes https://github.com/pulumi/docs/issues/793, somehow.

PR #888, opened 2019-02-13
Update OpenStack Node.js and all providers Python docs

jen20:


jen20:
I'll open a new pull request for this, since we also have new versions of several other packages.

PR #899, opened 2019-02-18
Fixing broken link in faq.md

dcaba:
It was referencing an html instead of the md... unless these pages are also populated into a different place (website) as html, looks wrong


dcaba:
Mm, i see these markdowns are somehow rendered as html, so the links in the pulumi site actually work https://pulumi.io/reference/state.html ... maybe we should put here the absolute path to the pblic page so it does not lead into confusion?

lukehoban:
Yes - this currently only works correctly as currently written.  We would likely need to move to a new content engine (away from Jekyll) to address this deeply.  For now, we'll keep the content as-is so that it continue to render correctly on https://pulumi.io.

PR #901, opened 2019-02-19
Link to status.pulumi.com

ellismg:


reference/troubleshooting.md
```diff
@@ -50,8 +50,9 @@ The Pulumi command-line tool interacts with the Pulumi web service throughout th
 service is unable to process an update, it is possible that users of the command-line tool may see this error message
 throughout the course of an update.
 
-We take great pride in service uptime and work rapidly to fix service interruption. You can follow our
-[official Twitter account](https://twitter.com/PulumiCorp) to keep up-to-date on service interruptions.
+We take great pride in service uptime and work rapidly to fix service interruption and use our [status page](https://status.pulumi.com) to
+communicate information about service outages.
```
justinvp: Nit: I'll defer to you, but just wondering if we should use a word less strong than "outages" (e.g. "incidents", "status", etc.?) as full-on outages will hopefully be rare.

```suggestion
communicate information about service incidents.
```

PR #919, opened 2019-03-05
UPDATE with pulumi new info

jkogut:
**Following:**

Get Started ---> Download and install ---> After installation, choose a cloud to get started: ---> Kubernetes

found no information about requirements for project existence, therefore was hit with:

`error: no Pulumi project found in the current working directory`

To avoid this, my suggestion is to mention about the need of project creation (or existence).


lukehoban:
I'm not quite sure this is the right place to call this out - as it's not quite a "pre-requisite", it's more part of the getting started flow to actually get to a working Pulumi program.  

I opened https://github.com/pulumi/docs/issues/997 to track really making sure we nail this flow through the docs overall.  I think we can solve for the problem you originally hit here in a deeper way that will help ensure that any path through the docs makes it clear what steps are needed to get to a working Pulumi deployment.  For now, I'm going to close this out, but if you have thoughts on what would be the highest impact things we could do here - please do add a note to https://github.com/pulumi/docs/issues/997 .

PR #924, opened 2019-03-07
Update docs for 0.17.X

ellismg:


ellismg:
Note: This removes `aws-serverless` from the api docs and renames `aws-infra` to `awsx`

PR #930, opened 2019-03-11
Update programming-model.md

CyrusNajmabadi:
Adding informaiton about lifting properties.

### DIFF
--- a/reference/programming-model.md
+++ b/reference/programming-model.md
@@ -224,7 +224,7 @@ let certValidation = new aws.route53.Record("cert_validation", {
 ```python
 certificate = aws.acm.Certificate("cert",
   domainName: "example.com",
-  validationMethod: "DNS",
+  validation_method="DNS",
   
 record = aws.route53.Record("validation",
   # Need to pass along a deep subproperty of this Output
--- a/reference/programming-model.md
+++ b/reference/programming-model.md
@@ -189,9 +189,8 @@ url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error) 
 
 The `apply` method accepts a callback which will be passed the value of the `Output` when it is available, and which returns the new value.  The result of the call to `apply` is a new `Output` whose value is the value returned from the callback, and which includes the dependencies of the original `Output`.  If the callback itself returns an `Output`, the dependencies of that output are unioned into the dependencies of the returned `Output`.
 
-> Note: The `Output` itself cannot be used directly in string concatenation or other operations, as it is not itself the value of the output.  To transform the value of the output (when it becomes available), the `apply` method should be used instead.
+> Note: Several common types of transformations can be done more convienently.  See (Accessing properties of an Output)[#lifting] for how to access Output value properties simply.   Also, `Output` itself cannot be used directly in string concatenation as it is not itself the value of the output.  See (Working with Outputs and strings)[#ouputs-and-strings] for examples of how to more simply work use the two together.  For cases where these convenience forms are not sufficient, `.apply` is available the most general way to transform one `Output` into another.  
 
-`.apply` is the most general way to transform one `Output` into another.  However, several common types of transoformations can be done more conveinently.  See (Accessing properties of an Output)[#lifting] and (Working with Outputs and strings)[#ouputs-and-strings] for examples of this.
 
 ##### Accessing properties of an Output {#lifting}
 
--- a/reference/programming-model.md
+++ b/reference/programming-model.md
@@ -195,7 +195,7 @@ The `apply` method accepts a callback which will be passed the value of the `Out
 
 ##### Accessing properties of an Output {#lifting}
 
-It is common to need to only access some subproperty of an `Output` in order to pass in that subproperty to another `Resource`.  For example, when using ACM certificates one would need to write:
+It is common to need to only access some subproperty of an `Output` in order to pass in that subproperty to another `Resource`.  For example, when using ACM certificates one might write:
 
 {% include langchoose.html %}
 
@@ -205,6 +205,7 @@ let certCertificate = new aws.acm.Certificate("cert", {
   validationMethod: "DNS",
 });
 let certValidation = new aws.route53.Record("cert_validation", {
+  // Need to pass along this Output's deep subproperty
   records: [certCertificate.domainValidationOptions.apply(domainValidationOptions => domainValidationOptions[0].resourceRecordValue)],
   ...
 ```
@@ -215,6 +216,7 @@ let certCertificate = new aws.acm.Certificate("cert", {
   validationMethod: "DNS",
 });
 let certValidation = new aws.route53.Record("cert_validation", {
+  // Need to pass along this Output's deep subproperty
   records: [certCertificate.domainValidationOptions.apply(domainValidationOptions => domainValidationOptions[0].resourceRecordValue)],
   ...
 ```
@@ -225,6 +227,7 @@ certificate = aws.acm.Certificate("cert",
   validationMethod: "DNS",
   
 record = aws.route53.Record("validation",
+  #Need to pass along this Output's deep subproperty
   records: [certificate.domain_validation_options.apply(
       lambda domain_validation_options => domain_validation_options[0].resource_record_value
   )],
--- a/reference/programming-model.md
+++ b/reference/programming-model.md
@@ -229,7 +229,7 @@ certificate = aws.acm.Certificate("cert",
 record = aws.route53.Record("validation",
   # Need to pass along a deep subproperty of this Output
   records: [certificate.domain_validation_options.apply(
-      lambda domain_validation_options => domain_validation_options[0].resource_record_value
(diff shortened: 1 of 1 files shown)


reference/programming-model.md
```diff
@@ -189,9 +189,99 @@ url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error)
...
+```
+
+```python
+certificate = aws.acm.Certificate("cert",
+  domainName: "example.com",
+  validationMethod: "DNS",
+  
+record = aws.route53.Record("validation",
+  records: [certificate.domain_validation_options.apply(
+      lambda domain_validation_options => domain_validation_options[0].resource_record_value
```
swgillespie: Can you test this:

```python
record = aws.route53.Record("validation", records=[
  certificate.domain_validation_options[0].resource_record_value
])
```

You should not need the apply in there.

swgillespie:
```suggestion
      lambda domain_validation_options: domain_validation_options[0].resource_record_value
```

reference/programming-model.md
```diff
@@ -189,9 +189,99 @@ url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error)
...
+  ...
+```
+
+```python
+certificate = aws.acm.Certificate("cert",
+  domainName: "example.com",
+  validationMethod: "DNS",
+  
+record = aws.route53.Record("validation",
+  records: [certificate.domain_validation_options.apply(
```
swgillespie:
```suggestion
  records=[certificate.domain_validation_options.apply(
```

reference/programming-model.md
```diff
@@ -189,9 +189,99 @@ url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error)
...
+  validationMethod: "DNS",
+});
+let certValidation = new aws.route53.Record("cert_validation", {
+  records: [certCertificate.domainValidationOptions.apply(domainValidationOptions => domainValidationOptions[0].resourceRecordValue)],
+  ...
+```
+
+```python
+certificate = aws.acm.Certificate("cert",
+  domainName: "example.com",
```
swgillespie:
```suggestion
  domain_name="example.com",
```

reference/programming-model.md
```diff
@@ -189,9 +189,99 @@ url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error)
...
+  validationMethod: "DNS",
```
swgillespie:
```suggestion
  validation_method="DNS",
```

reference/programming-model.md
```diff
@@ -189,9 +189,99 @@ url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error)
...
+});
+let certValidation = new aws.route53.Record("cert_validation", {
+  records: [certCertificate.domainValidationOptions[0].resourceRecordValue],
+  ...
+```
+
+```python
+certificate = aws.acm.Certificate("cert",
+  domainName: "example.com",
```
swgillespie:
```suggestion
  domain_name="example.com",
```

reference/programming-model.md
```diff
@@ -189,9 +189,99 @@ url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error)
...
+  validationMethod: "DNS",
```
swgillespie:
```suggestion
  validation_method="DNS",
```

reference/programming-model.md
```diff
@@ -189,9 +189,99 @@ url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error)
...
+  
+record = aws.route53.Record("validation",
+  records: [certificate.domain_validation_options[0].resource_record_value],
```
swgillespie:
```suggestion
  records=[certificate.domain_validation_options[0].resource_record_value],
```

reference/programming-model.md
```diff
@@ -189,9 +189,102 @@ url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error)
 
 The `apply` method accepts a callback which will be passed the value of the `Output` when it is available, and which returns the new value.  The result of the call to `apply` is a new `Output` whose value is the value returned from the callback, and which includes the dependencies of the original `Output`.  If the callback itself returns an `Output`, the dependencies of that output are unioned into the dependencies of the returned `Output`.
 
-> Note: The `Output` itself cannot be used directly in string concatenation or other operations, as it is not itself the value of the output.  To transform the value of the output (when it becomes available), the `apply` method should be used instead.
+> Note: The `Output` itself cannot be used directly in string concatenation or other operations, as it is not itself the value of the output.  To transform the value of the output (when it becomes available), the `apply` method should be used instead.
```
lukehoban: Trailing white space?

reference/programming-model.md
```diff
@@ -189,9 +189,102 @@ url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error)
...
 
-> Note: The `Output` itself cannot be used directly in string concatenation or other operations, as it is not itself the value of the output.  To transform the value of the output (when it becomes available), the `apply` method should be used instead.
+> Note: The `Output` itself cannot be used directly in string concatenation or other operations, as it is not itself the value of the output.  To transform the value of the output (when it becomes available), the `apply` method should be used instead.  
 
-#### All {#all}
+`.apply` is the most general way to transform one `Output` into another.  However, several common types of transoformations can be done more conveinently.  See (Accessing properties of an Output)[#lifting] and (Working with Outputs and strings)[#ouputs-and-strings] for examples of this.
```
lukehoban: Should this be above or somehow integrated with the “note:” section above?  It seems to now slightly contradict what’s mentioned there.

reference/programming-model.md
```diff
@@ -189,9 +189,102 @@ url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error)
...
+
+##### Accessing properties of an Output {#lifting}
+
+It is common to need to only access some subproperty of an `Output` in order to pass in that subproperty to another `Resource`.  For example, when using ACM certificates one might write:
```
lukehoban: Perhaps just “property” instead of “subproperty”.

reference/programming-model.md
```diff
@@ -189,9 +189,102 @@ url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error)
...
+  ...
+```
+
+```go
+// Helpers for accessing properties are not yet available in Go.
+// 
+// See https://github.com/pulumi/pulumi/issues/1614.
+```
+
+To make this kind of property and array-element access more simple, `Outputs` 'lift' the properties of the values they wrap to themselves as well.  For example, if the wrapped property contains a value `prop: string`, then the `Output` wrapping that has a property `prop: Output<string>`.  This 'lifting' is just shorthand for writing out the full `.apply` call.  In other words, an expression of the form `someOutput.someProp` is equivalent to `someOutput.apply(o => o.someProp)`.  Similarly, `someOutput[val]` is equivalent to writing `someOutput.apply(o => o[val])` (with the right syntax for each respective language).  Given this, the above examples can be rewritten into:
```
lukehoban: I think we can be more direct here and say “you can access properties of an output directly, and can transform an output with apply”. 

In particular, I don’t think we have to show both the apply and property access versions here - just showing how to access properties seems fine for this section - and noting the limitations that will require using the more general apply.

reference/programming-model.md
```diff
@@ -189,9 +189,100 @@ url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error)
...
+  ...
+```
+
+```go
+// Helpers for accessing properties are not yet available in Go.
+// 
+// See https://github.com/pulumi/pulumi/issues/1614.
+```
+
+This greatly helps the clarity of the final code, while not losing any important dependency information that is needed for properly creating and maintaining the stack.
```
lukehoban: I think we can avoid words like “greatly” here.  Also not exactly sure how the dependency information part is related - that’s a given for anything related to Ouput right?  Doesn’t seem that property access changes it?

Might be worth adding note here on calling out the related scenarios which will require using apply (method calls, passing an output to another function, etc. ).

CyrusNajmabadi: > that’s a given for anything related to Ouput right? Doesn’t seem that property access changes it?

I want to help people realize you get an Output back, and it's not like `x.y.z` is just giving a string back that loses it.

reference/programming-model.md
CyrusNajmabadi: fixed.

reference/programming-model.md
CyrusNajmabadi: done.

reference/programming-model.md
CyrusNajmabadi: Made more direct.

reference/programming-model.md
CyrusNajmabadi: Combined and rewordded.

reference/programming-model.md
CyrusNajmabadi: removed the 'greatly' part.

PR #931, opened 2019-03-11
Create cyblog.md

CyrusNajmabadi:


cyblog.md
```diff
@@ -0,0 +1,75 @@
...
...orks at runtime.  There are many things a Proxy lets you control, however for our needs the most important bit was that it allows you to override how member-lookup works.  In other words, when you write `someProxy.someMember` you get a chance to determine what should happen an what `someMember` should actually return!  With that flexible interception point available, we actually took our core `Output` type and made it into a `Proxy`.  We then added the right interception code so that if you ever write `someOutput.someMember` that that gets translated exactly into `someOutput.apply(o => o.someMember)`.   his also works just fine for array-accesses (which are just property-lookups from JavaScript's perspective).  In other words, at runtime, the member-lookup form and the `.apply` form will be equivalent (except that the former is so much nicer to write!).  This is why, for example, `certCertificate.domainValidationOptions[0].resourceRecordValue` will have the correct value (with all the right dependency information).  At runtime it really will be equivalent to the original `certCertificate.apply(certCertificate => certCertificate.domainValidationOptions[0].resourceRecordValue)` form.
```
swgillespie: What about this, for grammar (avoiding use of the second person):

"Proxies allow users to write JavaScript code that intercepts and overrides many facets of JavaScript runtime behavior."

swgillespie: In general I think grammatically we should avoid the second person in this paragraph, since we're not actually instructing the reader to do something; we're describing that proxies allow *us* to overload aspects of JavaScript instead of *you* the user of `Proxy`, since in the end we are the ones handling the proxy.

Is that reasonable, you think?

cyblog.md
```diff
@@ -0,0 +1,75 @@
+### Simplified Outputs in Pulumi 0.17.0
+
+The `Output<T>` type is used an enormous amount in any Pulumi application.  It is the way that Resources expose their values, and it is commonly used to pass values from one Resource to another.  Importantly `Outputs` are a key part of how Pulumi tracks dependencies between resources.  In fact, Outputs are similar to promises/futures that you may be familiar with from other programming models but also carry along dependency information.
```
swgillespie: Nit: comma splices

```suggestion
The `Output<T>` type is used an enormous amount in any Pulumi application.  It is the way that Resources expose their values and it is commonly used to pass values from one Resource to another.  Importantly, `Outputs` are a key part of how Pulumi tracks dependencies between resources.  In fact, Outputs are similar to promises/futures that you may be familiar with from other programming models but also carry along dependency information.
```

cyblog.md
```diff
@@ -0,0 +1,75 @@
...
+
+const certValidation = new aws.route53.Record("cert_validation", {
+    records: [certCertificate.domainValidationOptions.apply(domainValidationOptions => domainValidationOptions[0].resourceRecordValue)],
+    ttl: 60,
+    type: certCertificate.domainValidationOptions.apply(domainValidationOptions => domainValidationOptions[0].resourceRecordType),
+    zoneId: zone.apply(zone => zone.id),
+});
+```
+
+In particular, creating the `aws.route53.Record` involve a fair amount of complexity with those arrow functions.  i.e.:
```
swgillespie: Singular subject

```suggestion
In particular, creating the `aws.route53.Record` involves a fair amount of complexity with those arrow functions.  i.e.:
```

cyblog.md
```diff
@@ -0,0 +1,75 @@
...
+```
+
+That's a lot nicer than before!  The following improvements happened:
+1. there's no more lambda!
+2. there's no need for a repetive lambda parameter (which might conflict with some other name in scope).
+3. the code *is* just pure simple *idiomatic* JavaScript/TypeScript that clearly conveys its intent.
+
+Importantly, no information has been lost here.  The exact same dependency information flows along here like it did before.  And, thanks to TypeScript's flexible type system, the above is totally typesafe and will let you know the right types of things and will still error if you happen to make mistakes.
+
+So, how was this done?  Well, the core part of the change is thanks to a little-known feature of JavaScript: (Proxies)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy].
```
swgillespie: I think something funny is happening with the Markdown here, GitHub doesn't seem to render this link correctly?

cyblog.md
```diff
@@ -0,0 +1,75 @@
...
+
+Now, while this was fairly easy to get working at runtime from a JavaScript perspective, it was a little more challenging to figure out how to make this work in TypeScript's typing system.  For example, if you had a value like so:
```
swgillespie: We started off this post in the third person ("It allows one to pass a piece of code that will be applied to the underlying value...") and ended up addressing the second person here ("For example, if you had a value like so..."). What about using the first person plural for this segment, since we're introducing a hypothetical: "For example, suppose we had a value like so"?

cyblog.md
```diff
@@ -0,0 +1,75 @@
...
+```ts
+    records: [certCertificate.domainValidationOptions.apply(domainValidationOptions => domainValidationOptions[0].resourceRecordValue)],
+    ttl: 60,
+    type: certCertificate.domainValidationOptions.apply(domainValidationOptions => domainValidationOptions[0].resourceRecordType),
+    zoneId: zone.apply(zone => zone.id),
+```
+
+Yikes!  This is so verbose, it doesn't even fit on the width of the page!
+
+The idea of the `.apply` function is similar to `Promise.then`.  It allows one to pass a piece of code that will be applied to the underlying value, and will return an `Output` that then points to the transformed underlying value.  Importantly, the new `Output` will still track dependency information properly.  In the above example the `aws.route53.Record` will know that it both depends on `certCertificate` and on `zone`, even though neither of those resources (or their direct Outputs) are passed directly do the constructor.
```
swgillespie: Nit: comma splice
```suggestion
The idea of the `.apply` function is similar to `Promise.then`.  It allows one to pass a piece of code that will be applied to the underlying value and will return an `Output` that then points to the transformed underlying value.  Importantly, the new `Output` will still track dependency information properly.  In the above example the `aws.route53.Record` will know that it both depends on `certCertificate` and on `zone`, even though neither of those resources (or their direct Outputs) are passed directly do the constructor.  
```

cyblog.md
joeduffy: I think the `(`s and `)s` got swapped with the `[`s and `]`s.

CyrusNajmabadi:
Thanks @joeduffy .  That was it.

Issue #815, opened 2019-01-30
Document how to use links to Kubernetes resources


lukehoban:
In the Kubernetes setup/getting-started page we should add a note about using the links to Kube resources from the pulumi console - indicating;
1. They point to the Kubernetes Dashboard which you'll need to install for these links to work
2. That you'll need to run `kubectl proxy` to be able to proxy local access to the dashboard (which should be running inside the cluster not exposed directly to the internet)

lblackstone:
Related: https://github.com/pulumi/pulumi-service/issues/3172

PR #941, opened 2019-03-18
Add note about Dashboard links for Kubernetes

lblackstone:
Fixes https://github.com/pulumi/docs/issues/815


lblackstone:
> is this a good place to call out the potential security risks with running the dashboard, given it needs a root like serviceAccount, and most tend to be publicly exposed?

I linked directly to the dashboard repo rather than giving the one-liner to install, so users should see all the scary warnings in their docs.

PR #948, opened 2019-03-25
Fix awsx nodejs source links

### LINKED ISSUE #947: AWSX nodejs source links are incorrect
They are currently prefixed with:

https://github.com/pulumi/pulumi-awsx/blob/master/nodejs

But the source actually lives at:

https://github.com/pulumi/pulumi-awsx/blob/master/nodejs/awsx

justinvp:
Fixes #947


justinvp:
I'm going to merge this because I have some more changes I want to built on-top of this (and this is a pretty straightforward change, IMO). Happy to address any feedback subsequently.

PR #949, opened 2019-03-25
Fix link in GCP/GKE tutorial

hausdorff:


hausdorff:
Oh, I just had an idea. If add a bit of text that links to that YAML file, the broken-link-checker will actually tell us if we've moved it. I'll do that now.

hausdorff:
@ekrengel I am super sympathetic to the idea of moving this, but I'm not sure what to do immediately. I'd actually prefer that we auto-generate this page from the actual README.md files in the examples folders, but that's just me.

ekrengel:
I think this is fine for now. I'll take down a note to think about it more

PR #983, opened 2019-04-09
Tweaks to GCP Setup

### LINKED ISSUE #969: Recommend using application-default credentials in gcp quickstart
Instead of using a service account, we should push folks towards using `gcloud auth application-default login` in our gcp quickstart.

### LINKED ISSUE #970: Improve https://pulumi.io/quickstart/gcp/setup.html
The steps on https://pulumi.io/quickstart/gcp/setup.html heavily bias toward setting up service accounts, which is a generally difficult and more advanced model for using GCP locally.

We should start with the simple approach of `gcloud auth application-default login` and then include additional details on how to do service accounts after that.

ellismg:
Push folks towards using `gcloud auth application-default login` by
default and then move the service account stuff to it's own page.

Fixes #969
Fixes #970


quickstart/gcp/setup.md
```diff
@@ -4,44 +4,19 @@ redirect_from: /install/gcp.html
...
 
 [Pulumi Google Cloud Platform Provider]: ./index.html
-[Google Cloud Platform Credentials]: https://console.cloud.google.com/apis/credentials
 
 The [Pulumi Google Cloud Platform Provider] needs to be configured with Google credentials
 before it can be used to create resources.
 
-You will need to provide the Google Cloud Platform Provider with your Google service account private key. You can
-create and download credentials using the [Google Cloud Platform Credentials] page on the Google Cloud Platform Console.
+When developing locally, we recommend that you use `gcloud login` to configure your account credientals:
```
lukehoban: `credientals` => `credentials`

quickstart/gcp/service-account.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Setup Service Account
+---
+
+[Google Cloud Platform Credentials]: https://console.cloud.google.com/apis/credentials
+
+You will need to provide the Google Cloud Platform Provider with your Google service account private key. You can
```
lukehoban: Perhaps an intro paragraph about the scenario and value for using a Service Account - to be able to run Pulumi deployments with a specific set of scoped credentials not associated with an individual user - whether in CI or for more control over the account role used locally for development.

ellismg: Added this:

>Using a Google service account allows you to use Pulumi in a non-interactive setting (for example CI/CD systems, where a person can not complete the normal `gcloud auth application-default login` flow). A service account can also be used when developing locally to ensure a specific set of scoped credentials not tied to a user account are used. This can be useful even when developing locally to give you more control over the account role used for deployment.

To the start of the page.

quickstart/gcp/setup.md
ellismg: Fixed, thanks!

quickstart/gcp/setup.md
```diff
@@ -56,4 +31,10 @@ when creating a resource
 creating a resource.
 
 If these values are not provided, some resources may require you to provide these
-values as `project`, `region`, or `zone` properties.
+values as `project`, `region`, or `zone` properties.  You may also set these values per project using `pulumi config`:
+
+```bash
+$ pulumi config set gcp:project <your-project-here>
+$ pulumi config set gcp:region <your-region-here>
+$ pulumi config set gcp:region <your-zone-here>
```
joeduffy: Did you mean to say `pulumi config set gcp:zone <your-zone-here>`, instead of `... gcp:region ...`?

PR #989, opened 2019-04-09
IP is not a value available when running via EKS

cwiggs:


lukehoban:
I’m not sure this change is correct by itself. @metral could you take a look?

metral:
Ah, yes this PR is correct, it should be `hostname`. I must've copy/pasted it by accident with GKE's tutorial. I'm glad this got caught, thanks @cwiggs!

PR #995, opened 2019-04-23
Docs for configuring Azure AD as Pulumi SAML IdP

chrsmith:
Adds documentation for how to configure Azure Active Directory as a SAML Identity Provider for the Pulumi Service. (@clstokes did all the work here, I just converted his doc into Markdown.)

This PR also goes through our existing SAML configuration guides for Okta and G Suite and does some general cleanup. Trying to make the three SAML docs somewhat more consistent with one another.

It also explicitly links to the SAML configuration pages from the `orgs` page, since previously we didn't reference the SAML configuration docs anywhere. (Since there wasn't a way to actually create a SAML org to start the process.)


reference/service/saml-aad.md
```diff
@@ -0,0 +1,111 @@
...
+  into the **Identity Provider Metadata** field.
+
+    ![Provide the XML IdP descriptor](../../images/reference/service/saml-aad/pulumi-saml-settings-page.png)
+
+1. Click **Save**.
+
+## Signing into Pulumi using Azure AD
+
+Once your Azure AD application is created, and its configuration data passed to Pulumi, you can now
+log into the Pulumi Cloud Console using your SAML SSO credentials.
```
cnunciato: Bit of a nit, but we should probably choose either "sign in" or "log in" for this sort of thing. Personally I'm partial to "sign in" (and use `login` more as a noun -- e.g., your login === your username), as it's a bit friendlier/more colloquial, but could go either way -- I just think we should be consistent. E.g.:

&times; Please log in
&times; Please login
&times; Please signin
&check; Please sign in

reference/service/saml-okta.md
```diff
@@ -47,11 +44,12 @@ tbody tr td {
...
 | General Setting | Value |
 | --------------- | ----- |
-| Single sign on URL | https://api.pulumi.com/login/robot-co/sso/saml/acs |
-| Audience URI | https://api.pulumi.com/login/robot-co/sso/saml/metadata |
+| Single sign on URL | https://api.pulumi.com/login/**acmecorp**/sso/saml/acs |
+| Audience URI | https://api.pulumi.com/login/**acmecorp**/sso/saml/metadata |
 | Name ID format | Unspecified |
 | App username | Email |
 
+<p><!-- space between table and text --></p>
```
cnunciato: 😱 😄  (But we really should be using CSS for this instead. The rework coming up can hopefully address this sort of thing.)

reference/service/saml-okta.md
```diff
@@ -92,26 +90,23 @@ a user's identity.
 
 ![SAML Application Metadata](../../images/reference/service/saml-okta/okta-xml-descriptor.png)
 
-With the block of XML text in your clipboard, navigate to your SAML organization's page on the
-Pulumi Cloud Console. The exact URL depends on the name of the organization, but it will be
-something like:
-https://app.pulumi.com/robot-co/settings
+With the block of XML text in your clipboard, open to the Pulumi Cloud Console and navigate to your SAML organization. Click the
```
cnunciato: `open to the` -> `open the`?

reference/service/saml-okta.md
```diff
@@ -92,26 +90,23 @@ a user's identity.
...
+**Settings** tab, and then select **SAML SSO**.
 
-On the organization setting's page, paste the IDP metadata descriptor into the bottom card
-titled "SAML SSO Settings". Then click the "SAVE" button at the bottom of the card.
+Paste the IDP metadata descriptor into the bottom card
```
cnunciato: Another small thing -- we're sort of using a mix of "IdP" and "IDP". Should probably just pick one and stick with it.

reference/service/saml-aad.md
chrsmith: Sounds good to me. I never know which to use, so I'll take a quick pass and try to standardize on "sign in".

reference/service/saml-okta.md
chrsmith: Yes, we should. Please forgive this obviously flawed code smell.

I'm intentionally not touching the CSS in hopes that the rework will fix all known problems.

reference/service/saml-okta.md
chrsmith: Yes, done.

reference/service/saml-okta.md
chrsmith: Noted, will do. (Just use whatever we seem to use more often.)

Issue #996, opened 2019-04-23
AWSX API Gateway not showing up in docs


joeduffy:

joeduffy:
@CyrusNajmabadi any idea why these are missing altogether from https://pulumi.io/reference/pkg/nodejs/@pulumi/awsx/index.html?

CyrusNajmabadi:
I do.  Not sure why these docs aren't there.  I honestly don't really know how our docs pipeline works.  Going to try to learn a bit about it.

CyrusNajmabadi:
Looks like typedoc is failing on our code:

```
Using TypeScript 3.1.6 from /mnt/c/go/src/github.com/pulumi/docs/node_modules/typedoc/node_modules/typescript/lib
Error: /mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/apigateway/api.ts(617)
 Argument of type '{ [securityDefinitionName: string]: SecurityDefinition; } | undefined' is not assignable to parameter of type '{}'.
  Type 'undefined' is not assignable to type '{}'.
Error: /mnt/c/go/src/github.com/pulumi/pulumi-awsx/nodejs/awsx/apigateway/api.ts(630)
 Object is possibly 'undefined'.
```

Looking to try to upgrade the version we use so it can actually understand the modern TS code we use.

PR #999, opened 2019-04-23
Update docs

CyrusNajmabadi:
Fixes https://github.com/pulumi/docs/issues/996

Had to upgrade the version of typedoc we use.  Also worked around a couple of crashes in our go doc conversion tool.


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "dependencies": {
     "broken-link-checker": "^0.7.8",
-    "typedoc": "^0.13.0",
-    "typescript": "^3.1.4"
+    "typedoc": "^0.15.0-0",
+    "typescript": "^3.4.5"
```
CyrusNajmabadi: needed to understand latest typescript features we use.

tools/tscdocgen/main.go
```diff
@@ -948,11 +948,15 @@ func createTypeLabel(t typeDocType, indent int) string {
...
-				createTypeLabel(index.Type, indent))
+		} else if len(decl.IndexSignatures) == 1 {
+			index := decl.IndexSignatures[0]
+			if len(index.Parameters) == 1 {
+				return fmt.Sprintf("{[%s: %s]: %s}",
+					index.Parameters[0].Name,
+					createTypeLabel(index.Parameters[0].Type, indent),
+					createTypeLabel(index.Type, indent))
+			}
+			return "{ ... }"
```
CyrusNajmabadi: had to fixup typedoc support since it now lists index signatures in an array.  no clue if this is the right ting to do, but it works.

PR #1007, opened 2019-04-24
Update installer links & CLI docs

metral:


metral:
@ellismg feedback has been addressed.

PR #1015, opened 2019-04-25
~index: Fix Slack link at bottom of page.

clstokes:


chrsmith:
Thanks for fixing this.

PR #1019, opened 2019-04-26
Prevent selection of `$` for bash snippets

justinvp:
bash-no-select

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -6,3 +6,4 @@
     crossorigin="anonymous"></script>
 <script src="/js/main.js"></script>
 <script src="/js/langchoose.js"></script>
+<script src="/js/noselect.js"></script>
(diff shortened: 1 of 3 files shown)


_includes/scripts.html
```diff
@@ -6,3 +6,4 @@
     crossorigin="anonymous"></script>
 <script src="/js/main.js"></script>
 <script src="/js/langchoose.js"></script>
+<script src="/js/noselect.js"></script>
```
justinvp: Ideally we'd pack/minify the JavaScript, but we're not doing that already.

justinvp:
I opened #1024 to discuss whether we want to keep the `$` going forward (across the board), and #1023 to track adding a "copy to clipboard" button for snippets. Until then, this seems like a nice little improvement to make it a bit easier to copy commands.

justinvp:
Gonna go ahead and merge. Happy to address feedback subsequently.

Issue #1020, opened 2019-04-27
Reference doc links drift over time


CyrusNajmabadi:
A core problem we have with our reference docs is they they embed links to our actual code to the master branch.  i.e. https://pulumi.io/reference/pkg/nodejs/@pulumi/awsx/apigateway/#metric has a link to https://github.com/pulumi/pulumi-awsx/blob/master/nodejs/awsx/apigateway/metrics.ts#L97

This is problematic since the links drift as we update master, making them potentially confusing/useless.  It's particularly bad because the docs are supposed to represent `latest` npm release for these packages, but instead point at something else entirely.

This shouldn't really be difficult to fix.  We just need to embed the links to the files with the appropriate permalink form that contains the actual tree-hash. i.e. https://github.com/pulumi/pulumi-awsx/blob/a64b1117969d34a54cf53f40abbcf9248a643e58/nodejs/awsx/apigateway/metrics.ts#L97

PR #1021, opened 2019-04-27
Use permalinks in our docs so that they don't drift when master changes.

CyrusNajmabadi:
Fixes https://github.com/pulumi/docs/issues/1020


scripts/run_typedoc.sh
```diff
@@ -18,6 +18,9 @@ TOOL_APIDOCGEN="go run ./tools/tscdocgen/*.go"
 PULUMI_DOC_TMP=`mktemp -d`
 PULUMI_DOC_BASE=./reference/pkg/nodejs/@pulumi
 
+# Set this to 1 to run all generation in parallel.
+PARALLEL=1
```
CyrusNajmabadi: this is useful locally as it makes things about 3x faster on my machine.  On a build machine it sucks since all the output is interleaved.

scripts/run_typedoc.sh
```diff
@@ -40,36 +43,67 @@ generate_docs() {
         echo -e "\033[0;93mGenerating typedocs\033[0m"
         pushd ../$2
         if [[ -z "$NOBUILD" ]]; then
+            git clean -xdf
             make ensure && make build && make install
```
CyrusNajmabadi: added this because i've had issues with some node module stuff not being pristine when doing thsi stuff.

scripts/run_typedoc.sh
```diff
@@ -40,36 +43,67 @@ generate_docs() {
...
-generate_docs "random" "pulumi-random" "sdk/nodejs"
-generate_docs "vsphere" "pulumi-vsphere" "sdk/nodejs"
+REPOS=(
+    "aws,pulumi-aws,sdk/nodejs"
+    "awsx,pulumi-awsx/nodejs/awsx"
+    "azure,pulumi-azure,sdk/nodejs"
+    "azure-serverless,pulumi-azure-serverless/nodejs/azure-serverless"
+    "cloud,pulumi-cloud/api"
+    "cloud-aws,pulumi-cloud/aws"
+    "cloud-azure,pulumi-cloud/azure"
```
CyrusNajmabadi: @lukehoban i think you're removin azure-serverless/cloud-aws/cloud-azure.  Do you just want me to do it?

PR #1022, opened 2019-04-27
Remove unnecessary links to foo/index.html

chrsmith:
Serving website content directly from S3 leaves a few things to be desired opposed to a proper web server like NGINX. However, it is configured to automatically serve `foo/index.html` for requests to `foo/`. ([Pulumi code](https://github.com/pulumi/docs/blob/master/infrastructure/index.ts#L32), [AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/dev/IndexDocumentSupport.html))

This PR did a grep for all cases where we linked to `/index.html` and just sampled it to `/`, even in semi odd cases like `../index.html` or `./index.html`. (Trying links like `./` and `../` worked fine, at least on Safari where I tested it.)

I also made the change in `tsdocgen` to fix the files it generates, but when trying to run it locally there were a lot of unexpected content diffs. So I'll need to follow up with someone about what the right steps are for regenerating those files.

### Other Changes

I also fixed references to "GitHub" that had the incorrect capitalization. (It should have both a capital G and H.)

I changed the markdown language hint from `bash` to `shell` in a few places, but that's actually unnecessary. (I didn't read the [Rogue docs](https://github.com/jneen/rouge/wiki/List-of-supported-languages-and-lexers) close enough, and evidently `bash` is an alias for `shell`, so those could be left unchanged.)

However, there was one case where there wasn't a language hint at all. So adding `shell` is a big improvement for that snippet.


_data/reference.yaml
```diff
@@ -116,67 +116,67 @@ toc:
       sidenav_disable_prefix_selection: true
       section:
       - title: pulumi
-        path: /reference/pkg/nodejs/@pulumi/pulumi/index.html
+        path: /reference/pkg/nodejs/@pulumi/pulumi/
```
justinvp: I pulled down your branch locally to test whether TOC selection still works with these changes, and looks like there's a subtle bug: If I select "cloudflare", both "cloud" and "cloudflare" show as selected.

Screen Shot 2019-04-29 at 8 22 18 AM

The TOC selection code is a nightmare. Would love to burn it down and replace with something cleaner / more maintainable.

https://github.com/pulumi/docs/blob/master/_includes/toc.html
https://github.com/pulumi/docs/blob/master/_includes/toc_sub.html

chrsmith: WE WILL KILL IT WITH FIRE

ericrudder:
better to just let someone on the team take a stab at it

justinvp:
Going to close as this is no longer relevant after moving to Hugo.

PR #1032, opened 2019-05-01
Add docs for `ignoreChanges`

lukehoban:


reference/programming-model.md
```diff
@@ -109,19 +109,22 @@ The `args` provided to a resource determine what inputs will be used to initiali
...
 
 ###### `deleteBeforeReplace`
-Optionally specify that replacements of the resource will delete the existing resource before creating it's replacement.  This will lead to downtime during the replacement, but may be necessary for some resources that manage scarce resources behind the scenes.  The default is `false`.
+Specify that replacements of the resource will delete the existing resource before creating it's replacement.  This will lead to downtime during the replacement, but may be necessary for some resources that manage scarce resources behind the scenes.  The default is `false`.
+
+###### `ignoreChanges`
+Provide a list of properties which will be ignored as part of updates.  The value of the property will be used for newly created resources, but will not be used as part of updates to resources.  This can be used to avoid changes in properties leading to diffs, or to change defaults for a property without forcing all existing deployed stacks to update or replace the affected resource.
```
ericrudder: Provides a list of properties which will be ignored as part of updates.  The value of the property will be used for newly created resources, but will not be used as part of updates.  This is typically used to avoid changes in properties leading to diffs and to change defaults for a property without forcing all existing deployed stacks to update or replace the affected resource.

reference/programming-model.md
```diff
@@ -109,19 +109,22 @@ The `args` provided to a resource determine what inputs will be used to initiali
 All resource constructors also accept an `options` argument which can provide the following additional resource options controlling how the resource will be managed by Pulumi. 
 
 ###### `dependsOn`
-Optionally provides a list of explicit resource dependencies to add to the implicit dependencies from inputs to the resource.  Every resource referenced directly or indirectly by an `Output` passed in to the resource constructor will implicitly be included, so this is only needed when the dependency is on something that is not already an input to the resource.  The default is `[]`.
+Provides a list of explicit resource dependencies to add to the implicit dependencies from inputs to the resource.  Every resource referenced directly or indirectly by an `Output` passed in to the resource constructor will implicitly be included, so this is only needed when the dependency is on something that is not already an input to the resource.  The default is `[]`.
```
ericrudder: Provides a list of explicit resource dependencies to add to the resource.  Every resource referenced either directly or indirectly by an `Output` that is passed in to the resource constructor will implicitly be included, so this additional information is only needed when the dependency is on something that is not already an input.  The default is `[]`.

PR #1035, opened 2019-05-01
Update webhooks.md with change in wording

casey-robertson:
Modified statement to align with intent.


lukehoban:
I believe the original text is closer to the intent - one of the actions that will trigger a WebHook is modifying the membership of a team. Possible that wording can be improved here though. Cc @chrsmith

casey-robertson:
Ok apologies about that.  No worries if I misinterpreted the meaning.

PR #1054, opened 2019-05-09
Update version to 0.17.10

lblackstone:


justinvp:
Actually, looks like `reference/changelog.md` needs to be updated as well like in @lukehoban's PR: #1055. Since the CLI docs didn't change (other than the gen date), we could just take #1055 instead.

lukehoban:
> Since the CLI docs didn't change (other than the gen date), we could just take #1055 instead.

Went ahead and merged https://github.com/pulumi/docs/pull/1055, so will close this one out.

Issue #1036, opened 2019-05-01
pulumi.io bug-fixes and general style improvements 


cnunciato:
This meta-issue is intended to contain a proposed set of visual fixes and overall improvements to pulumi.io.

* [x] Fix the search-box styling. Round the corners on the input field, and either add a little spacing between the activation button and the input, or remove the spacing and make the button and input look more unified. The activation button probably needn't be blue, either, as that competes with the SIGN UP button beside it.

* [x] In general, button styles look a bit off as well: the single-pixel outline is unique to pulumi.io (so should be removed), paddings, font weights, etc. We should generally apply the same treatments we're using on pulumi.com.

* [ ] Cloud-provider logos on the home and quick-start pages should be converted from PNGs to SVGs, and they should look more interactive as well. (They currently seem more decorative/informative.)

* [x] We should remove the underlines in most normal hrefs/links. The bright-blue coloring is clear enough.

* [x] The gray headings on the home page (Why Pulumi? etc.) seem a bit drab. They should be given equal or greater contrast then the content below them. We should consider styling them similarly to the Any Code, Any Cloud items on the pulumi.com home page, with higher-contrast headings, for example.

* [ ] The edges of the orange images below that section (in Containers, Serverless, etc.) look fuzzy against the gray background. We could probably make these stand out better by introducing some alternative color and/or contrast.

* [x] The bright blue looks pretty dissonant in this palette. I realize we're using it elsewhere, and it works for links, but it _really_ stands out against when used for buttons and headings in page bodies. There might be a better choice for some of these.

* [x] Items in the TOC navigation should get some sort of a hover effect, like a color change similar to the color they'd get when in the active/selected state, or maybe blue, since that's the color of in-page links.

* [x] The text of TOC items aliases poorly on normal-resolution monitors, and we probably don't need the one-pixel separators between the TOC items themselves, as they're distinguished clearly enough by their color/weight, the spacing between them, and the indentation of their children. (Down with visual noise! :smile:)

* [x] Bullets beside bright-blue/linked text should be text-colored rather than link-colored (as they're not linked themselves).

* [x] In-page TOCs should look more like TOCs, i.e., distinguished from the content below them; even a simple one-pixel seam between them and the content they refer to could help make this separation clearer. Also, since they aren't steps in a sequence, we should remove the lines that connect them. And the bullets themselves should line up vertically with the bulleted items in the pages below them.

[...]

PR #1064, opened 2019-05-10
Fix search-box styling, adjust button styles

cnunciato:
This change improves the layout of the search box and sign-up button for both mobile and desktop, and adjusts button borders to match their hover-state background colors.

**Also, for best results, ignore whitespace**, as this also converts affected files from tabs to spaces.

Part of https://github.com/pulumi/docs/issues/1036.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -60,13 +60,13 @@
                             </span>
                         </button>
 
-                        <span class="right-nav-block">
+                        <div class="right-nav-block">
                             <form id="search-form" action="/search.html" method="get">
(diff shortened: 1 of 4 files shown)


cnunciato:
Out of curiosity, anyone prefer a more muted search button to the blue? Just to minimize the number of blue things calling for your attention?

ericrudder:
we're actually "supposed" to be like the G cloud console, which is even
"more muted"  (white in search box) ... :)

[image: image.png]

i assume if you actually change to match, they will immediately change
their design!

cnunciato:
All right, I've updated the PR to use the muted version instead.

PR #1076, opened 2019-05-16
Create an initial AWSX/Crosswalk User Guide

joeduffy:


_data/reference.yaml
```diff
@@ -91,6 +91,25 @@ toc:
     path: /reference/javascript.html
   - title: Python
     path: /reference/python.html
+- title: Crosswalk for AWS
```
ericrudder: Pulumi Crosswalk for AWS?

_data/reference.yaml
```diff
@@ -91,6 +91,25 @@ toc:
     path: /reference/javascript.html
   - title: Python
     path: /reference/python.html
+- title: Pulumi Crosswalk for AWS
```
justinvp: With the move to Hugo, we can now define the TOC for this section in each page's front matter.

First, all the markdown files need to move to `/content/reference/crosswalk-aws` and `index.md` needs to be renamed `_index.md`.

Then you can follow a pattern similar to the files inside the [`/content/reference/vs`](https://github.com/pulumi/docs/tree/1ffc212d434c5d5933962941b3209c003ae12325/content/reference/vs) directory.

Inside `_index.md` add front matter like:

```
---
title: "Pulumi Crosswalk for AWS"
menu:
  reference:
    identifier: crosswalk-aws
    weight: 8
---
```

Regarding the `weight`: You'll need to tweak the weight in other files for the sections below this one, if you want `Pulumi Crosswalk for AWS` to come before `Compared to Others` and later sections.

Then in `vpc.md`:

```
---
title: "AWS Virtual Private Cloud (VPC)"
menu:
  reference:
    parent: crosswalk-aws
    name: "Virtual Private Cloud (VPC)"
    weight: 1
---
```

- `parent` is the parent menu item, the identifier we used in `_index.md`.
- `name` is an alternate name to use (otherwise if this isn't specified the title of the page is used)
- `weight` to order the items

And in `elb.md`:

```
---
title: "AWS Elastic Load Balancing (ELB)"
menu:
  reference:
    parent: crosswalk-aws
    name: "Elastic Load Balancing (ELB)"
    weight: 2
---
```

etc.

Or you could choose not to specify a weight for these child pages, and just let Hugo order them alphabetically based on the `name` (or `title` if a `name` wasn't specified).

reference/crosswalk-aws/cloudwatch.md
```diff
@@ -0,0 +1,15 @@
+---
+title: "AWS CloudWatch Metrics, Alarms, and Dashboards"
+---
+
+{% include mini-toc.html %}
```
justinvp: With the move to Hugo, this will need to be changed to use the `mini-toc` shortcode:

```
{{< mini-toc >}}
```

reference/crosswalk-aws/vpc.md
```diff
@@ -0,0 +1,499 @@
...
+
+// Export a few interesting fields to make them easy to use:
+export const vpcId = vpc.id;
+export const vpcPrivateSubnetIds = vpc.privateSubnetIds;
+export const vpcPublicSubnetIds = vpc.publicSubnetIds;
+```
+
+This example reads the default VPC and exports some of its properties for easy consumption. `pulumi up` will show:
+
+```bash
```
justinvp: Despite downgrading to an earlier version of Hugo (v0.55.4) to workaround some markdown rendering regressions, I still found one case that Hugo was having trouble rendering on v0.55.4. For some reason, specifying a language for code fences that contain the output of `pulumi up` was causing the code fence not to render as a code snippet.

I had to workaround this by remove the wrapping ```bash ... ``` and indenting the code snippet twice (e.g. 8 spaces) for it to render correctly as a code snippet (just without a language for highlighting, which is reasonable to live without in this case).

I need to track down whether there is already a bug open on this or if I need to open one against [Hugo's default markdown renderer](https://github.com/russross/blackfriday).

It *might* not be an issue here, as the places I was running into this was in our tutorials where the code snippets are inside lists of steps, and haven't narrowed it to a minimal repro. So just an FYI if this doesn't render as you expect, try removing the wrapping ```bash ... ``` and indenting twice instead.

justinvp:
I rebased these changes against the latest in master and fixed things up for Hugo in the following branch: https://github.com/pulumi/docs/tree/joeduffy/1071_awsx_user_guide-hugo

joeduffy:
Thank you, Justin!

joeduffy:
Superseded by https://github.com/pulumi/docs/pull/1136.

PR #1079, opened 2019-05-16
Remove font-smoothing mixin

cnunciato:
The properties set in this mixin seem to be causing problems with font rendering and legibility. Since they are generally not recommended, we should remove them.

Also, lotsa tabs -> spaces, so suggest hiding whitespace changes. 😄 


joeduffy:
LGTM, thank you!

PR #1088, opened 2019-05-23
Convert site from Jekyll to Hugo

### LINKED ISSUE #202: Update Jekyll configuration to not emit .html extension

### LINKED ISSUE #1034: Migrate docs site to Hugo
Before we spend a bunch more time investing in improving our use of Jekyll (e.g. addressing ToC issues, dev innerloop, build times, etc.), let'd do a time-boxed investigation (1-2 days) of migrating the docs site over to Hugo (align with www site).

Can we do it? Will it improve things? Will it not break things?

justinvp:
We're now using Hugo to generate the site 🎉

More details on the change:

- Index files are now `_index.md` instead of `index.md`.
- All markdown content files live under the `/content` directory, so `/aws`, `/quickstart`, `/reference`, `/shortlinks`, and `/tour` all now live under `/content`.
	- With the Jekyll site, we were generating our Node.js API docs under `/reference/pkg/nodejs/@pulumi/*`. We could continue to do that with Hugo (though, within the `/content` parent folder), but when generating the site, Hugo elides the `@` character from `@pulumi`, so the generated URLs would be `/reference/pkg/nodejs/pulumi/*`. To avoid any confusion over the difference between having `@pulumi` as the source and `pulumi` as the generated folder, I’ve renamed `@pulumi` in the source to `pulumi`. All existing packages currently published under `@pulumi` get an [alias](https://gohugo.io/content-management/urls/#aliases) that redirects to the new location under `pulumi`.
- Files under `/_sass`  are now under `/assets/sass`. We use [Hugo Pipes](https://gohugo.io/hugo-pipes/scss-sass/) to process and minify the SASS to CSS.
- Files in `/_includes` and `/_layouts` are now under `/layouts`
	- Partial layouts are under `/layouts/partials`. These are included in other layouts. For example, we have `head.html`, `header.html`, and `footer.html`, etc. partials that we include in the default `/layouts/_default/baseof.html` layout.
	- [Shortcodes](https://gohugo.io/content-management/shortcodes/) are under `/layouts/shortcodes`. These are things that can be included in markdown files. Hugo has some built-in shortcodes that we use (such as `relref` for links), but also supports custom shortcodes, and we have a bunch of custom shortcodes. For example, we have a `langchoose.html` shortcode that we include in some of our markdown files via `{{< langchoose >}}`.
- Static files are in  the `/static` folder. So files that should be in the root (e.g. https://pulumi.io/latest-version) live in the `/static` folder. As do other static files like:  `/static/js`,  `/static/images`, and `/static/fonts`.
	- Note: We’re not currently bundling/minifying our JavaScript code, but this is something we should consider as part of moving to pulumi.com. The www site uses `grunt` to do some processing, so we could leverage that (although, it’d be nice to move to `webpack`, and also use TypeScript).
- We use Hugo’s built-in [`relref`](https://gohugo.io/content-management/shortcodes/#ref-and-relref) shortcode when linking to pages. This way, we don’t have to link to what we think the generated URL will be; we can just link to the markdown page we want, and Hugo will take care of translating it to a relative URL. If Hugo can’t find the page, it will error, so it provides some nice verification of links.
- I’ve recreated the TOC for each section to match the existing Jekyll site using [Hugo Menus](https://gohugo.io/content-management/menus/).  These are mostly defined in each page’s front matter. However, for now, the menus for the CLI commands and API reference are specified in the `config.toml` file. I did this for expedience, but we should be able to move these into the generated page’s front matter so they don’t have to be specified explicitly in the `config.toml` at all.
- Many of our markdown files contain HTML. Any markdown content specified within HTML tags is not usually rendered in Markdown. With Jekyll, there was a way to workaround this, by including a `markdown="1"` in the surrounding HTML tag. Hugo’s markdown renderer doesn’t support `markdown="1"`, but we can workaround it by using a custom shortcode that I added. Any markdown inside HTML tags that you want rendered as markdown, just wrap in `{{% md %}}` and `{{% /md %}}`.
- We’re now using pretty URLs. I’ve set up [aliases](https://gohugo.io/content-management/urls/#aliases) for all the old ugly `foo.html` style URLs, so old links should continue to work (redirecting to the new URL). Unfortunately, all the aliases adds to the build time, so it's now closer to 9 seconds (still a huge improvement over Jekyll!)
	- When we move the content to pulumi.com, we can leverage S3 (or Cloudfront) redirects (that do real 301 redirects) and get rid of all these aliases.

Fixes #1034
Fixes #202


Makefile
```diff
@@ -55,18 +43,19 @@ test:
...
+		--exclude "/reference/changelog" \
 		--exclude "https://github.com/pulumi/docs/edit/master" \
 		--exclude "https://www.linkedin.com/company/pulumi/" \
 		--exclude "https://marketplace.visualstudio.com/items?itemName=pulumi.build-and-release-task"
 
 .PHONY: validate
 validate:
-	bundle exec jekyll serve --strict_front_matter --host 0.0.0.0 --incremental --detach >/dev/null 2>&1
+	hugo server -D --disableBrowserError --disableLiveReload &>/dev/null &
+	while ! nc -z localhost 1313; do sleep 0.1; done
```
justinvp: It takes a little bit of time for the `hugo server` to start up, so this is to wait until port 1313 is open before proceeding with the broken link validation. I'm not sure if `nc` is available in Travis. We'll see.

Makefile
```diff
@@ -89,13 +78,40 @@ ifeq ($(TRAVIS_BRANCH),production)
 endif
 
 .PHONY: travis_push
-travis_push:: banner ensure build deploy
+travis_push::
+	$(MAKE) banner
+	$(MAKE) ensure
+ifeq ($(TRAVIS_BRANCH),master)
+	export HUGO_BASEURL = https://staging.pulumi.io/
```
justinvp: When running `hugo` to build the site, if we're on the master branch, I want to specify the `HUGO_BASEURL` env variable so that `https://staging.pulumi.io/` is used for the base URL instead of the default `https://pulumi.io/` specified in `config.toml`. I *think* exporting it like this here before calling the `build` target below will work (the `hugo` command part of the `build` target needs to see the variable, but I haven't confirmed.

Repeating this for each `travis_` target is pretty lame. I'm sure there's a much more elegant way of handling this, and would love any suggestions, although, this Makefile isn't long for this world...

Makefile
```diff
@@ -5,45 +5,33 @@ default: banner generate build
...
-	@bundle exec jekyll serve --strict_front_matter --host 0.0.0.0 --incremental
+	hugo server -D
 
 .PHONY: generate
 generate:
 	@echo -e "\033[0;32mGENERATE:\033[0m"
 	./scripts/run_typedoc.sh
 	./scripts/generate_python_docs.sh
-	pulumi gen-markdown ./reference/cli
+	pulumi gen-markdown ./content/reference/cli
```
justinvp: I have a change to `pulumi` that I'll open a PR for and add the link here.

justinvp:
Unfortunately, it looks like there are a lot of files that git thinks are Added + Deleted instead of Moved. For these, I didn't make any substantial changes to the content, only minor tweaks such as to front matter and updating links to use the `{{< relref "page.md" >}}` shortcode.

Makefile
justinvp: Yeah, [doesn't work](https://travis-ci.com/pulumi/docs/builds/112964262).

```
export HUGO_BASEURL = https://staging.pulumi.io/
/bin/sh: 1: export: : bad variable name
make: *** [travis_pull_request] Error 2
The command "make travis_${TRAVIS_EVENT_TYPE}" exited with 2.
```

Will fix.

justinvp:
> Unfortunately, it looks like there are a lot of files that git thinks are Added + Deleted instead of Moved. 

Let me rework the history a bit to maintain the renames.

content/quickstart/aws/tutorial-ec2-webserver.md
```diff
@@ -51,46 +55,43 @@ In this tutorial, we'll use JavaScript to deploy a simple webserver EC2 instance
...
-
-        Type                      Name                               Plan       Info
-    +   pulumi:pulumi:Stack       webserver-webserver-dev            create     
-    +   ├─ aws:ec2:SecurityGroup  webserver-secgrp                   create     
-    +   └─ aws:ec2:Instance       webserver-www                      create     
-    
-    info: 3 changes previewed:
-        + 3 resources to create
-    ```
+        $ pulumi up
```
justinvp: Hugo's markdown renderer was having trouble rendering this with the ```bash. To workaround, I can just indent twice (e.g. 8 spaces) and it will be treated as a code snippet and renders as expected. It's not perfect as we lose the language specifier, but for these cases it's not a big deal.

content/quickstart/aws/tutorial-rest-api.md
```diff
@@ -29,16 +34,16 @@ pulumi new hello-aws-javascript
 
 You can accept the defaults for this command. For instance, you can change the AWS region to `us-west-2`.
 
-![Run Pulumi new](/images/quickstart/hello/Quickstart1.png){:width="700px"}
+<img src="/images/quickstart/hello/Quickstart1.png" alt="Run Pulumi new" width="700">
```
justinvp: Hugo doesn't support the `{:...}` syntax that Jekyll does for specifying widths and adding classes, so just using an `` tag for these instead.

PR #1089, opened 2019-05-23
Update tutorial-eks.md

CyrusNajmabadi:


quickstart/aws/tutorial-eks.md
```diff
@@ -45,17 +45,17 @@ In this tutorial, we'll launch a new Managed Kubernetes cluster in Elastic Conta
...
     const name = "helloworld";
 
     // Create an EKS cluster with non-default configuration
-    const vpc = new awsinfra.Network("vpc", { usePrivateSubnets: false });
+    const vpc = new awsx.ec2.Vpc("vpc");
     const cluster = new eks.Cluster(name, {
-        vpcId: vpc.vpcId,
-        subnetIds: vpc.subnetIds,
+        vpcId: vpc.id,
+        subnetIds: vpc.publicSubnetIds,
```
lukehoban: Is this right?  We are changing the example from "public" to "public+private", but then passing the "public" subnets to the cluster, which won't work, right?

CyrusNajmabadi: Why woudl the public subnets not work?  Note: i will change this to be fully public.  however, the presence of private subnets shoudln't affect your ability to use the public ones, right?

CyrusNajmabadi:
> Should we also update all the other mentions of aws-infra in this docs repo while we are at it?

Definitely.

quickstart/aws/tutorial-eks.md
```diff
@@ -108,24 +108,8 @@ In this tutorial, we'll launch a new Managed Kubernetes cluster in Elastic Conta
...
-    +      ├─ aws:ec2:RouteTable                      vpc                               	create
-    +      ├─ aws:ec2:NatGateway                      vpc-nat-1                         	create
-    +      ├─ aws:ec2:RouteTableAssociation           vpc-nat-1                         	create
-    +      ├─ aws:ec2:NatGateway                      vpc-nat-0                         	create
-    +      ├─ aws:ec2:RouteTableAssociation           vpc-nat-0                         	create
-    +      ├─ aws:ec2:RouteTable                      vpc-nat-1                         	create
-    +      ├─ aws:ec2:RouteTable                      vpc-nat-0                         	create
-    +      ├─ aws:ec2:RouteTableAssociation           vpc-1                             	create
-    +      └─ aws:ec2:RouteTableAssociation           vpc-0                             	create
+    ...
```
justinvp: FYI, I just merged the change to convert the site from Jekyll to Hugo, which will cause a conflict here (sorry!). As part of that, I had to tweak these ```bash code snippets to workaround a markdown rendering issue with Hugo. For some reason, Hugo's markdown renderer does not like all the `+` in these code snippets.

To workaround, I removed the ```bash from these snippets, and instead just indented twice (8 spaces), which allows it to render as a code snippet (and render correctly). We lose the `bash` language, but that doesn't matter much for these particular snippets.

quickstart/aws/tutorial-eks.md
lukehoban: I am not exactly sure that it doesn’t work (though I suspect it might not), but it is certainly not common to have private sinners but then run compute in the public ones. The typical (only?) intent of public+private subnet arrangement is to have compute on private and ingress/egress through public.

lukehoban:
Rebased on top of the latest Hugo site and took care of the rest of aws-infra references in #1093.

PR #1096, opened 2019-05-28
API doc updates

justinvp:
Updates from running `./scripts/run_typedoc.sh`. There were some updates to `@pulumi/pulumi` that I excluded as it looks like the v0.17.14 tag was just created a few hours ago but hasn't been "released" on the docs site yet.

@stack72, @jen20, @lukehoban FYI


justinvp:
Oops, need to re-run for `@pulumi/aws` with latest `tf2pulumi`. Will push fix in a moment.

Issue #598, opened 2018-09-24
More documentation for using Pulumi in CI


chrsmith:
Add more docs for how to use Pulumi with CI systems, like Circle CI, Jenkins, Code Deploy, etc.

Bonus points for how the Pulumi GitHub app fits into them too. (i.e. what environment variables need to be set.)

joeduffy:
In order:

- [ ] GitLab
- [ ] Jenkins
- [ ] Azure DevOps
- [ ] Circle CI

joeduffy:
@praneetloke I think we made a good dent in this for M19. Moving out for now, feel free to opportunistically continue fleshing things out now and in M20, but great progress so far.

praneetloke:
I just unchecked Jenkins from the above list. Not sure how it was checked-off, but I don't see a related doc for it on our site.

chrsmith:
Added `P1` label since this is top of mind for the current milestone. @praneetloke , I'm happy to assist if you'd like some help.

kostis-codefresh:
Hello. I work for Codefresh and I can easily prepare a CI/CD page on how Codefresh and Pulumi can work together. There is already an example on our docs https://codefresh.io/docs/docs/yaml-examples/examples/pulumi/

praneetloke:
@kostis-codefresh that would be great! Please feel free to send us a PR.

kostis-codefresh:
@praneetloke PR is here https://github.com/pulumi/docs/pull/1087 :-)

chrsmith:
I believe we added some new docs here, but there's still more we can do. So keeping this open for the next milestone as well.

chrsmith:
Moving over to @praneetloke for the current milestone. It looks like we have pretty solid coverage, and with our hooks / documentation for adding new CI/CD systems, I think we have all of our bases covered.

If we just added Jenkins this milestone I think we'd be in great shape. /cc https://github.com/pulumi/pulumi/issues/2106.

joeduffy:
@praneetloke Do you have cycles to tackle Octopus and Spinnaker?

praneetloke:
Moving to M28, but I have a doc for Octopus Deploy in-progress that I plan on completing today.

stack72:
Closing this out - we can revisit this as more concrete items / use cases rather than a top level open goal

PR #1112, opened 2019-05-31
Add docs to show how to add support for detecting a new CI system in the CLI

praneetloke:
Related to #598.

Here's how it looks when it's rendered:

Screenshot_2019-05-31 Supporting A New CI System


ericrudder:
thanks for adding!

can we use proper casing for all headings?

iow:  Supporting New CI Systems  instead of Supporting new Ci systems

etc.

praneetloke:
@ericrudder thanks for the feedback! I've made the updates and updated the screenshot.

.gitignore
```diff
@@ -2,4 +2,8 @@
 /resources/
 /public/
 node_modules
+_site/
```
praneetloke: @justinvp fyi I needed to add these entries. Running `make serve` was generating these.

justinvp: @praneetloke, these should not be needed with Hugo. We used to have similar entries in this file when we were using Jekyll. Are you sure you don’t have older directories left over from before the Hugo migration? Try running `git clean -xdf`.

content/reference/cd-supporting-new-ci.md
```diff
@@ -0,0 +1,80 @@
+---
+title: Supporting A New CI System
+aliases: ["cd-supporting-new-ci.html"]
```
justinvp: This line is not needed and should be deleted. Existing files that had previously been published with Jekyll using the ugly URLs have these aliases (redirects), but new content doesn’t need it.

```suggestion
```

.gitignore
praneetloke: Ohh that might be it. Let me remove them.

praneetloke: Confirmed. It was leftover from before. I have reverted this change.

content/reference/cd-supporting-new-ci.md
praneetloke: Done.

content/reference/cd-supporting-new-ci.md
```diff
@@ -0,0 +1,80 @@
...
+expanded_url: /reference/cd/
+menu:
+  reference:
+    parent: cd
+---
```
justinvp: It might be worth adding `weight: 100` under `parent` (and same indentation as `parent`), so that this article sorts to the bottom of the other CD articles in the table of contents on the left. (100 is just an arbitrary large value). What do you think?

praneetloke: Yep. I like that!

praneetloke: Hmm adding `weight: 100` is causing this to be the first link rather than last. Actually, any non-zero value for `weight` is doing that. Must be because the other entries don't have a `weight`?

content/reference/cd-supporting-new-ci.md
```diff
@@ -0,0 +1,79 @@
...
+In order to add support for your CI system, you should be somewhat familiar in working with [Go](https://golang.org/).
+
+## How Does It Work?
+
+There are 2 parts to the CLI for it to be able to correctly detect the environment variables of a CI system.
+
+1. Detecting in which CI system the Pulumi CLI is currently running
+1. Detecting additional environment variables for a specific CI system
+
+In the `pulumi/pulumi` repo the `pkg/util/ciutil` contains all of the logic necessary for the above. Every stack `preview` or `update` calls the functions exported from this package.
```
justinvp: Add a link around the `pulumi/pulum` repo? And also add links for any files or directories mentioned throughout this article?

praneetloke: Done.

content/reference/cd-supporting-new-ci.md
justinvp: Ah, OK. (Sorry, reviewing from my phone and not at the computer to try it myself). Yeah, probably because the other pages don’t have explicit weights. I was hoping we wouldn’t have to add explicit 1, 2, 3, etc. to each of the pages to manually control the order (I did it that way previously). I do wonder if we set all the other pages to have the same explicit weight (e.g. setting them all to 1) if they would still be ordered alphabetically by Hugo, or if that makes Hugo sort them randomly. Anyway, no need to address as part of this PR, I can experiment later.

content/reference/cd-supporting-new-ci.md
```diff
@@ -0,0 +1,79 @@
...
+menu:
+  reference:
+    parent: cd
+---
+
+## Overview
+
+The Pulumi CLI already supports enhanced metadata for several popular CI/CD systems. We have tried to make it easy for contributors to add/update support for CI systems. This document walks-through how you can modify the Pulumi CLI to support your own CI system or perhaps even update an existing one.
+
+The detection of metadata in a CI environment depends on some key environment variables that a CI system injects into the build environment of their build agents. The Pulumi CLI uses those environment variables to set try and determine the values for the following properties:
```
justinvp: “to set try and determine” doesn’t sound right to me

content/reference/cd-supporting-new-ci.md
```diff
@@ -0,0 +1,79 @@
...
+
+- CI System Name
+- Build Type
+- Build URL
+- Commit SHA
+- PR Number
+- Commit Message
+
+The above metadata about your CI environment is then used to provide an enhanced experience in the Pulumi Console when you look at your stack activity log. The metadata from your CI environment combined with the information about your Git repository allows us to provide links to things such as the Pull Request on github.com or the commit that triggered a PR build.
```
justinvp: Link around Pulumi Console?

content/reference/cd-supporting-new-ci.md
```diff
@@ -0,0 +1,79 @@
...
+
+- Add a new file to the [`pkg/util/ciutil`](https://github.com/pulumi/pulumi/tree/master/pkg/util/ciutil) folder with the name of the CI system for which you are adding support.
+- Define a new struct for the CI system and add `baseCI` to its definition. Refer to any of other pre-existing implementations, such as the [`travis.go`](https://github.com/pulumi/pulumi/blob/master/pkg/util/ciutil/travis.go) file.
+- Implement the `DetectVars()` function. This is the function where you should construct a new instance of the `Vars` struct and set its properties accordingly.
+
+That's it! Send us a new [PR](https://github.com/pulumi/pulumi/pulls) in the [`pulumi/pulumi`](https://github.com/pulumi/pulumi) repo with your addition. We would love to help you get that merged as soon as possible.
+
+## Using A Fallback
+
+If the CI system you are using is not current detected by Pulumi. You can set the following env vars, as a fallback, yourself. The CI detection code will assume a "Generic CI" system name.
```
justinvp: current => currently

And:

“Pulumi. You” => “Pulumi, you”

?

content/reference/cd-supporting-new-ci.md
praneetloke: OK sounds good. I'll leave it as-is in this PR.

content/reference/cd-supporting-new-ci.md
praneetloke: Ah fixed it! I must have changed my mind mid-sentence. 😁

content/reference/cd-supporting-new-ci.md
praneetloke: Done. I also rephrased the sentence a bit.

PR #1113, opened 2019-06-03
Ensure markdown is rendered for API docs code examples 

### LINKED ISSUE #1103: Unrendered markdown in some API docs
There are still some places where the markdown isn't being rendered as expected.

For example: https://pulumi.io/reference/pkg/nodejs/pulumi/digitalocean/#SshKey

Screen Shot 2019-05-29 at 1 48 49 PM

justinvp:
Hugo's markdown renderer will not render markdown between HTML tags, and does not support the `markdown="1"` attribute like Jekyll does to enable this. Instead, we have a `md` shortcode that be used around content that should be rendered as markdown. This change fixes our template for Node.js API docs to wrap some markdown content, so that it is rendered correctly, and regenerates all the Node.js docs.

Fixes #1103


joeduffy:
LGTM, thanks Justin!

PR #1120, opened 2019-06-04
Add proper 301 redirects for our Node.js API docs

### LINKED ISSUE #1119: Anchors are dropped from redirects
With the move to Hugo, the Node.js API docs have moved from `reference/pkg/nodejs/@pulumi/` to `reference/pkg/nodejs/pulumi/` as Hugo strips the `@` when generating the site.

We have Hugo aliases in place that redirect from the old locations to the new locations. However, these are based on [meta refresh](https://www.w3.org/TR/WCAG20-TECHS/H76.html), which does not pass along any anchors that may be specified in the source URL. This is problematic as many of our API docs rely on anchors when linking to specific identifiers.

The fix is to use real 301 redirects instead of these client-side redirects, as browsers will pass along the anchor when 301 redirects are encountered.

justinvp:
With the move to Hugo, the Node.js API docs have moved from `reference/pkg/nodejs/@pulumi/` to `reference/pkg/nodejs/pulumi/` as Hugo strips the `@` when generating the site.

We have Hugo aliases in place that redirect from the old locations to the new locations. However, these are based on client-side [meta refresh](https://www.w3.org/TR/WCAG20-TECHS/H76.html), which does not pass along any anchors that may be specified in the source URL. This is problematic as many of our API docs rely on anchors when linking to specific identifiers.

The fix is to use real 301 redirects instead of these client-side redirects, as browsers will pass along the anchor when 301 redirects are encountered.

---

I confirmed the fix in my dev stack. For example try navigating to: https://hugo.pulumi.io/reference/pkg/nodejs/@pulumi/azure/keyvault/#example-usage-generating-a-new-certificate

---

A follow-up would be to remove the Hugo aliases for the Node.js docs, as they are no longer needed with this in place. But I'll hold off on doing that, as we'll be removing unnecessary Hugo aliases and using proper 301 redirects as we migrate the docs from pulumi.io to www.pulumi.com.

Fixes #1119

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -31,6 +31,15 @@ const contentBucket = new aws.s3.Bucket(
         website: {
             indexDocument: "index.html",
             errorDocument: "404.html",
+            routingRules: JSON.stringify([{
+                "Condition": {
+                    "KeyPrefixEquals": "reference/pkg/nodejs/@pulumi/",
+                },
+                "Redirect": {
+                    "HostName": config.targetDomain,
+                    "ReplaceKeyPrefixWith": "reference/pkg/nodejs/pulumi/",
+                },
+            }]),
         },
     },
     {


justinvp:
One thing to note: This addresses the issue for our Node.js docs, but does not address for links elsewhere in the site. So if we have any links to the old ugly style URLs (e.g. https://pulumi.io/reference/programming-model.html#programs), the client-side redirects will redirect to the new location (e.g. https://pulumi.io/reference/programming-model/), but the anchor will unfortunately be dropped. We'll address this soon with proper 301 redirects for the whole site when we move the content to www.pulumi.com.

lukehoban:
> So if we have any links to the old ugly style URLs (e.g. https://pulumi.io/reference/programming-model.html#programs), the client-side redirects will redirect to the new location (e.g. https://pulumi.io/reference/programming-model/), but the anchor will unfortunately be dropped. 

FWIW - I do expect we have quite a few of those out int the wild - especially for `programming-model.html`.

lukehoban:
> So if we have any links to the old ugly style URLs (e.g. https://pulumi.io/reference/programming-model.html#programs), the client-side redirects will redirect to the new location (e.g. https://pulumi.io/reference/programming-model/), but the anchor will unfortunately be dropped. 

FWIW - I do expect we have quite a few of those out int the wild - especially for `programming-model.html`.

PR #1122, opened 2019-06-05
Fix strange markdown rendering in API docs

### LINKED ISSUE #1118: Some API examples include escaped HTML
For example: https://pulumi.io/reference/pkg/nodejs/pulumi/aws/s3/#static-website-hosting

Screen Shot 2019-06-04 at 11 12 48 AM

And looks like some `*` are missing from imports.

justinvp:
Hugo's markdown renderer does not support rendering markdown inside HTML tags and does not support the `markdown="1"` attribute on HTML tags that Jekyll supports to enable markdown rendering inside the tags.

To work around this limitation, we have a custom `md` shortcode that can be used inside HTML tags to wrap markdown that should be rendered. However, there was a bug in our `md` shortcode. We use the `{{% md %}}` syntax instead of `{{< md >}}` because we intend to have markdown inside the shortcode, and inside our `md` layout, we use `{{ markdownify .Inner }}` to render the `.Inner` content as markdown. However, in some cases, when not actually inside HTML tags, `.Inner` will end up being already rendered markdown. So we end up passing the already rendered markdown through the `markdownify` function, resulting in some wacky rendering.

The fix is to specify the behavior of `{{% %}}` per https://gohugo.io/content-management/shortcodes/#shortcodes-with-markdown, to consistently render `.Inner` as markdown whether it's inside HTML tags or not, and emit `.Inner` directly without passing it through `markdownify`.

## Before

https://pulumi.io/reference/pkg/nodejs/pulumi/aws/s3/#static-website-hosting

Screen Shot 2019-06-04 at 5 51 22 PM

## After

Screen Shot 2019-06-04 at 5 51 35 PM

Fixes #1118

### DIFF
--- a/layouts/shortcodes/md.html
+++ b/layouts/shortcodes/md.html
@@ -1,5 +1,9 @@
 {{/*
     This shortcode is used to render markdown inside HTML tags, which Blackfriday,
     the markdown processor that Hugo uses, doesn't currently support.
+    We specify version = 1 so that `.Inner` is rendered as markdown consistently,
+    whether it ends up being within HTML tags or not.
+    See https://gohugo.io/content-management/shortcodes/#shortcodes-with-markdown.
 */}}
-{{ markdownify .Inner }}
+{{ $_hugo_config := `{ "version": 1 }` }}
+{{ .Inner }}


justinvp:
Gonna merge and deploy to production to fix the rendering issues in the API docs. Happy to address any feedback subsequently.

PR #1128, opened 2019-06-06
Move content into content/docs

cnunciato:
This change shifts all docs content into a subdirectory (/docs) and adjusts navigation, TOCs and doc generation accordingly. It does not include any redirects or aliases yet -- just moves all content into a new directory, and makes everything navigable.


content/docs/_index.md
```diff
@@ -0,0 +1,170 @@
...
+<div class="card-table" style="margin-top: 24px">
+    <a href="https://www.pulumi.com"><img src="/images/logo/logo.svg" alt="Pulumi" width="300"></a>
+    <h2 class="get-to-the-cloud no-anchor">
+        Cloud Native Infrastructure as Code
+    </h2>
+    <h3 class="get-to-the-cloud-sub no-anchor">
+        Provision infrastructure on any cloud using your favorite language.
+    </h3>
+    <p style="text-align: center; margin-bottom: 0">
+        <a href="{{< relref "/docs/quickstart/install.md" >}}"><button class="button primary">INSTALL</button></a>
```
justinvp: This file (among many) isn't showing up as a move. Since the content changed too much, git thinks we deleted a file and added a new one. This is the downside of moving the files and updating their content in the same commit. In order to maintain history for a file, I had to move the files in one commit but not change their content (so git tracks it as a move), and then in a subsequent commit modify the content of the moved files to update links, etc.

cnunciato: Oh, I see -- I'll try that.

justinvp: If you want to save yourself a bunch of hassle breaking this up into two commits, copy the whole `/content/docs` directory with these changes to a temporary directory somewhere on your machine for temporary safe keeping.

Then start fresh and move the /content into /content/docs, but don't modify any files. Commit.

Then delete everything in /content/docs and replace it with the directory you stashed away, which should have all the modifications. And commit those changes.

cnunciato: Ok, I think this should be good now. Got through a full docs run and while I hit a few warnings, things seem to have completed successfully.

content/docs/reference/pkg/nodejs/pulumi/aws/_index.md
```diff
@@ -304,38 +304,10 @@ deployments.
 Use this data source to get the ID of a registered AMI for use in other
 resources.
 
-## Example Usage
```
justinvp: It looks like examples are missing in all these re-generated docs. I'm guessing the `tf2pulumi` tool isn't found on your `$PATH`?

On my various machines, I've cloned https://github.com/pulumi/tf2pulumi in my go directory, and then I run `make ensure` and `make build && make install`.

cnunciato:
@justinvp Thanks much, that did it. All should be good now. I'll go ahead and merge.

PR #1129, opened 2019-06-07
Advertise 0.17.16

ellismg:


ellismg:
> LGTM - should we mark that v0.17.15 has errors so users skip it?

Not sure - On the one hand, that's a decent idea.  On the other hand, we've had a few of these "we really should have not released this build because it had a bad bug, let's just cut a new build with the fix and move forward" and we don't try to call these things out explicitly.  I'd rather just funnel folks into the latest version regardless of the changes we make.

metral:
SGTM 👍

content/reference/changelog.md
```diff
@@ -28,6 +28,16 @@ menu:
         </tr>
     </thead>
     <tbody>
+        <tr>
+            <th scope="row"><a href="https://github.com/pulumi/pulumi/blob/master/CHANGELOG.md#01716-released-june-6-2019">0.17.16</a></th>
+            <td>2019/06/06</td>
+            <td>{{< sdk-links version="0.17.16" >}}</td>
+        </tr>
+        <tr>
+            <th scope="row"><a href="https://github.com/pulumi/pulumi/blob/master/CHANGELOG.md#01714-released-june-5-2019">0.17.15</a></th>
```
justinvp:
```suggestion
            <th scope="row"><a href="https://github.com/pulumi/pulumi/blob/master/CHANGELOG.md#01715-released-june-5-2019">0.17.15</a></th>
```

PR #1132, opened 2019-06-07
Make search work

cnunciato:
Accounts for the shift into /docs.


Makefile
```diff
@@ -30,7 +30,7 @@ generate:
 build:
 	@echo -e "\033[0;32mBUILD:\033[0m"
 	hugo
-	node ./scripts/build-search-index.js < ./public/search-data/index.json > ./public/search-index.json
+	node ./scripts/build-search-index.js < ./public/docs/search-data/index.json > ./public/docs/search-index.json
 	rm -rf ./public/search-data
```
justinvp:
```suggestion
	rm -rf ./public/docs/search-data
```

layouts/docs/search-data.json
```diff
@@ -0,0 +1,10 @@
+{
+{{- $pages := (where .Site.Pages ".Params.noindex" "ne" true) -}}
```
justinvp: Delete `layouts/_default/search-data.json` so this shows up in git as a move?

layouts/docs/search.html
```diff
@@ -0,0 +1,7 @@
+{{ define "main" }}
```
justinvp: Delete `layouts/_default/search.html` so this shows up in git as a move?

layouts/docs/search-data.json
cnunciato: Ah, totally thought I had. Good catch, thanks.

PR #1136, opened 2019-06-08
Add new documentation

joeduffy:
This adds a new area of documentation for our upcoming release.

We are taking a slightly different approach with this content, and describing how to accomplish end to end solutions, sometimes spanning multiple packages.

There are many known areas of improvement, most notably, screenshots and images, to break apart some of the "walls of text." I will try to do some, but we may have to go live with just the text to start.


content/reference/crosswalk-aws/lambda.md
```diff
@@ -0,0 +1,375 @@
...
+}
+
+tbody tr td:first-child {
+    text-align: left;
+}
+</style>
+
+| AWS Service | Event | Description |
+|-------------|-------|-------------|
+| API Gateway | [awsx.apigateway.API](/reference/pkg/nodejs/@pulumi/awsx/apigateway/#API) | create serverless APIs using a simple approach |
```
justinvp:
```suggestion
| API Gateway | [awsx.apigateway.API]({{< relref "/reference/pkg/nodejs/pulumi/awsx/apigateway#API" >}}) | create serverless APIs using a simple approach |
```

Two comments (applies throughout):

1. The URL to the reference docs should not include the `@` in `@pulumi`. Hugo strips the `@` in the published directory structure, so the URLs are now `/reference/pkg/nodejs/pulumi/...`. We do have redirects on pulumi.io that redirect from `@pulumi` to `pulumi` based routes, and will have redirects that go from pulumi.io to their new homes on www.pulumi.com when we flip the migration switch, but we aren't planning to keep around the redirects within www.pulumi.com.

2. Also, I've been leveraging the [`relref` shortcode](https://gohugo.io/content-management/shortcodes/#ref-and-relref) for links within the site. With that, Hugo will ensure the page being linked to exists, and if not, will fail the build, which is especially helpful as we shift content around in subdirs as part of the website migration. (It also allows linking directly to `foo.md` pages, so you don't have to guess what the published URL is going to be).

joeduffy:
@ericrudder Appreciate the editing feedback. Would you mind using [suggested changes](https://haacked.com/archive/2019/06/03/suggested-changes/) for typos and wordsmithing? It's 100x easier to incorporate into docs changes like these. Thank you!

content/reference/crosswalk-aws/lambda.md
joeduffy: Re (1), yeah, I found out the hard way about this. (I was using `#` anchors, and they were getting stripped during the redirection.) I'll do a pass to clean these up -- thanks for catching!

Re (2), ahh, ok, I will see about changing all the URLs I've got in there to adopt this.

joeduffy: Done, thanks again :+1:

content/reference/crosswalk-aws/lambda.md
```diff
@@ -0,0 +1,379 @@
...
+// Look up an S3 bucket that already exists in our account.
+const docsBucket = new aws.s3.Bucket.get("docs", "arn:aws:s3:::docs-4f64efc");
+
+// Create an AWS Lambda event handler on our bucket using magic functions.
+docsBucket.onObjectCreated("docsHandler", (e) => {
+    // your lambda code goes here
+});
+```
+
+Wn you run `pulumi up`, you'll still see this bucket resource but notice it says `read` instead of `create` or `update`.
```
aydrian:
```suggestion
When you run `pulumi up`, you'll still see this bucket resource but notice it says `read` instead of `create` or `update`.
```

content/reference/crosswalk-aws/api-gateway.md
```diff
@@ -0,0 +1,1043 @@
...
+To ease population of the S3 bucket to serve from, and connecting it to the API Gateway, you specify the local path --
+either a file or an entire directory -- and Pulumi will upload the contents as S3 objects.
+
+Let's say we have a directory `www` containing a single `index.html` file:
+
+```html
+<h1>Hello, API Gateway + S3!</h1>
+```
+
+The following program will create an API Gateway that serves this content at the `/www` URL:
```
aydrian: Did you mean the `/` URL? The path in the code is /.
```suggestion
The following program will create an API Gateway that serves this content at the `/` URL:
```

joeduffy: Good catch, thank you!

PR #1137, opened 2019-06-08
Port www head and toolchain fixes

chrsmith:
The `www` repo had some minor changes to the `head` partial, to clean up some of the meta tags we use for SEO. (Such as auto including the twitter author based on `authors` frontmatter.)

This PR just merges those changes in, and does some rearranging of the `head` partial. The only meaningful change should be the use of `PostCSS` as part of the toolchain. This was something we required to keep compatibility with the `www` repo when switching to the Sass build as part of the Hugo pipelines. While the "auto prefixer" feature is something I'd prefer we don't rely on, the CSS minification is better than what comes out fo the box. (And since we are including all of Bootstrap, etc. it makes a difference.)

Happy to go over some of the changes in more detail if anything isn't clear.


layouts/partials/head.html
```diff
@@ -1,62 +1,85 @@
 <head>
-    <meta charset="UTF-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
-    <meta name="google-site-verification" content="j-Opn93zR4YQcV5bbYN99BmNc48BT5mzd2VpzKaW9YY" />
+<meta charset="UTF-8">
```
cnunciato: Why remove the indentation of these, as they’re children of `head`?

chrsmith: I have no explanation, I'll make the change before submitting.

PR #1148, opened 2019-06-11
Clean up blog layout

chrsmith:
This PR does a solid pass at improving the UX of the blog layout and styling. While there is still a lot of room to improve (e.g. the paginator isn't styled at all), it is a good enough start to start with.

The changes to be sure to review are the edits to `partials/head.html`, which cleans up some redundant `<meta>` tags and hooks up PostCSS for our Sass build pipeline. (These were changes we made in the `www` repo, and seem like reasonable things to carry forward since the CSS magnification surprisingly helps a lot.)

Note that this does mean you'll need to run `{yarn | npm} install` in the repo's root before you can build again.

---

List page
image

Single page
image

Tag page
image

Author page
image

"Jank AF" paginator I didn't get to styling yet
image

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -3,60 +3,91 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
     <meta name="google-site-verification" content="j-Opn93zR4YQcV5bbYN99BmNc48BT5mzd2VpzKaW9YY" />
-
     {{ if isset .Params "redirect_to" }}
     <meta http-equiv="refresh" content="0; url={{ .Params.redirect_to }}" />
     {{ end }}
 
-    <title>{{ .Title }}{{ if not .IsHome }} - {{ end }}{{ .Site.Title }}</title>
-
-    {{ if isset .Params "meta_title" }}
-    <meta property="og:title" content="{{ .Params.meta_title }}">
-    <meta name="twitter:title" content="{{ .Params.meta_title }}">
-    {{ else if isset .Params "title" }}
-    <meta property="og:title" content="{{ .Params.title }}">
-    <meta name="twitter:title" content="{{ .Params.title }}">
-    {{ else }}
-    <meta property="og:title" content="{{ .Site.Title }}">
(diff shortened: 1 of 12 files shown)


layouts/partials/head.html
```diff
@@ -3,60 +3,91 @@
...
+    <meta property="og:site_name" content="pulumi">
+
+    <meta name="twitter:card" content="summary">
+    <meta name="twitter:site" content="@PulumiCorp">
+
+    {{ if isset .Params "meta_twitter_creator" }}
+        <meta name="twitter:creator" content="{{ .Params.meta_twitter_creator }}">
+    {{ else if isset .Params "authors" }}
+        {{ range $idx, $authorName := .Param "authors" }}
+        {{ $authorData := index $.Site.Data.team.team $authorName }}
```
cnunciato: Wondering what you all think about indentation and template expressions in Hugo. 

It seems like common in Hugo circles to stack nested expressions atop one another, without any indentation, presumably to achieve nicely indented rendered markup. But I personally find this approach optimizes for the wrong thing; we should be optimizing for the reader instead, since nobody’s looking at rendered markup anyway, and if they are, they’re probably using developer tools, which will render the DOM with proper indentation irrespective of how it’s laid out in raw source. I’d like to propose we indent these expressions to reflect how they’re evaluated, so it’s easier to read them (to see what’s contained in what, etc.). Anyone disagree?

chrsmith: I'm pretty sure I've been inconsistent in my Hugo formatting, specifically because I haven't figured out what the right balance is. As far as "what we should be doing", I am 110% on-board with treating the Hugo markup as code, so that it is easier to read/understand. 

My only reservation -- and this isn't specific to this issue -- is that as we optimize the templates for being easier to read, the end result looks less and less clean. All the white space, indentation, and "newlines at the end of the file" do add up. If you view the source for a generated source page, it seems like we would strongly benefit from some HTML minified, if such a thing exists and wouldn't ruin our code samples.

Anyways, I'll check the code in just to keep things moving, but I'm glad you mentioned something and will use that "optimizer for reader" approach in future edits.

cnunciato: Sounds good. FWIW, Hugo does have a `--minify` option that could help with the whitespace thing; we could give that a try just to see how it performs for us.

PR #1149, opened 2019-06-11
Fix code snippet scroll behavior

justinvp:
I noticed this while working on the get started changes. Fix code snippet scroll behavior after the Hugo move, where the generated code for code snippets is slightly different.

Before

Screen Shot 2019-06-11 at 9 50 58 AM

After

Screen Shot 2019-06-11 at 9 50 28 AM

### DIFF
--- a/assets/sass/_pulumi.scss
+++ b/assets/sass/_pulumi.scss
@@ -1,22 +1,14 @@
 @import "pulumi-constants";
 @import "pulumi-toc";
 
-// Drop the rounded corners for code snippets. Ideally we would just do this
-// for the code associated with our language chooser. However, there isn't a
-// way to associate the two. (Since the N language code blocks are siblings
-// to the lang chooser tabs.)
-pre.highlight {
-    border-radius: 0 !important;
-}
-
 // Force code snippets to scroll horizontally rather than word wrapping.
-pre.highlight code {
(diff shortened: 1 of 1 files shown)


assets/sass/_pulumi.scss
```diff
@@ -1,22 +1,14 @@
 @import "pulumi-constants";
 @import "pulumi-toc";
 
-// Drop the rounded corners for code snippets. Ideally we would just do this
-// for the code associated with our language chooser. However, there isn't a
-// way to associate the two. (Since the N language code blocks are siblings
-// to the lang chooser tabs.)
-pre.highlight {
-    border-radius: 0 !important;
```
justinvp: Later in the file we set the border-radius to 2px, so this has no effect. The 2px radius looks fine to me with the lang chooser tabs (if the radius was 4px, it wouldn't look good).

assets/sass/_pulumi.scss
```diff
@@ -323,10 +315,6 @@ pre {
     margin-bottom: 24px;
 }
 
-pre.highlight {
-    border-radius: 2px !important;
```
justinvp: This is not necessary as the block above sets the `border-radius` for `pre` to `2px`.

PR #1153, opened 2019-06-11
Remove hard-coded version is CI example scripts

### LINKED ISSUE #1145: show version string in bash scripts?
for example, on this page https://pulumi.io/reference/cd-aws-code-services, should we really have the version string show (here 16.2) ... can we just omit completely?

ellismg:
For folks who are going to copy these scripts, it's likely better for
them to just be flowing to the newest version instead of being locked
to an increasingly out of date CLI.

Fixes #1145

### DIFF
--- a/content/reference/cd-aws-code-services.md
+++ b/content/reference/cd-aws-code-services.md
@@ -63,7 +63,7 @@ phases:
   install:
     commands:
       # pulumi
-      - curl -L https://get.pulumi.com/ | bash -s -- --version 0.16.2
+      - curl -L https://get.pulumi.com/ | sh
       - export PATH=$PATH:$HOME/.pulumi/bin
   build:
     commands:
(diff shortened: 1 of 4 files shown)


content/reference/cd-aws-code-services.md
```diff
@@ -63,7 +63,7 @@ phases:
   install:
     commands:
       # pulumi
-      - curl -L https://get.pulumi.com/ | bash -s -- --version 0.16.2
+      - curl -L https://get.pulumi.com/ | sh
```
justinvp: Is there a reason we just specify `-L` instead of `-fsSL` like we mention at https://pulumi.io/quickstart/install/?

ellismg: Not really, I'll fix this up.

PR #1163, opened 2019-06-13
Fix issue preventing use of fixed width font in code snippets

### LINKED ISSUE #1162: Code snippets not using fixed width font
Chris noticed that some parts of code snippets aren't using the right font:

Screen Shot 2019-06-13 at 9 33 56 AM

The problem is due Chroma annotating certain portions of code snippets with a `p` class, and us having some .scss that was originally copied over from the www site that defines a `p` class to have the same styles as the `p` tag.

justinvp:
We have some typography scss copied over from the www site that defines styles for both the `p` tag as well as a `p` class. This isn't ideal because portions of code snippets are now annotated with a `p` class by Chroma for code highlighting, resulting in the wrong font being used.

The fix is to simply remove the `p` class as it's not needed (it was only being used elsewhere in the scss with `@extend .p`, but those uses can just extend the tag instead of the class, e.g. `@extend p`.

Before

Screen Shot 2019-06-13 at 9 53 15 AM

After

Screen Shot 2019-06-13 at 9 52 26 AM

Fixes #1162

Note: I've only removed the unnecessary `p` class in this change, but as part of the website work this milestone, we should rationalize this scss. There's really no need to have all these classes with the same name as the tags.

### DIFF
--- a/assets/sass/partials/_forms.scss
+++ b/assets/sass/partials/_forms.scss
@@ -1,10 +1,10 @@
 //basic form styles
 label{
-    @extend .p;
+    @extend p;
 }
 
 input[type="text"], input[type="email"], input[type="password"] {
-	@extend .p;
+	@extend p;
 	margin-bottom: 0;
     border: 1px solid #CECECE;
     padding: 16px 17px;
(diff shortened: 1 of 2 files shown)


cnunciato:
> as part of the website work this milestone, we should rationalize this scss

On it. 😄

PR #1170, opened 2019-06-14
Add RSS for the blog

cnunciato:
Note that this renders an RSS `link` for the blog on all pages of the site, not just those under /blog.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -2,7 +2,13 @@ baseURL = "https://pulumi.io/"
 languageCode = "en-us"
 title = "Pulumi"
 
-disableKinds = ["category", "RSS"]
+disableKinds = ["category", "taxonomyTerm"]
+sectionPagesMenu = "main"
+pygmentsCodeFences = true
+pygmentsUseClasses = true
+
+[outputFormats.RSS]
+    baseName = "rss"
 
 pygmentsCodeFences = true
 pygmentsUseClasses = true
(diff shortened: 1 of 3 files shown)


config.toml
```diff
@@ -2,20 +2,21 @@ baseURL = "https://pulumi.io/"
 languageCode = "en-us"
 title = "Pulumi"
 
-disableKinds = ["category", "RSS"]
+disableKinds = ["category", "taxonomyTerm"]
```
cnunciato: Removing `taxonomyTerm` as well here just to silence a build warning: 

```
WARN 2019/06/14 10:48:59 found no layout file for "HTML" for "taxonomyTerm"
```

I don't think we _need_ this, but I'm not yet an expert on Hugo taxonomies, so given our author and tag pages are rendering properly, it seems fine to suppress it. @justinvp may know better, though.

layouts/blog/rss.xml
```diff
@@ -0,0 +1,29 @@
+<rss version="2.0">
+    <channel>
+        <title>Pulumi Blog</title>
+        <link>{{ .Permalink }}</link>
+        <description>Infrastructure as Code for AWS, Azure, Google Cloud, including Kubernetes, Serverless, Containers, for Developers and DevOps.</description>
+        <language>{{ .Site.LanguageCode }}</language>
+        <pubDate>{{ .Date.Format "Mon, 02 Jan 2006" | safeHTML }}</pubDate>
```
justinvp: According to the [RSS 2.0 Spec](https://cyber.harvard.edu/rss/rss.html), these should conform to the Date and Time Specification of [RFC 822](https://www.ietf.org/rfc/rfc822.txt), which I believe should include the time and timezone.

From https://gohugo.io/templates/rss/:

```
<pubDate>{{ .Date.Format "Mon, 02 Jan 2006 15:04:05 -0700" | safeHTML }}</pubDate>
```

layouts/blog/rss.xml
```diff
@@ -0,0 +1,29 @@
+<rss version="2.0">
```
justinvp: Consider adding the following as the first line:

```
{{ printf "<?xml version=\"1.0\" encoding=\"utf-8\" standalone=\"yes\" ?>" | safeHTML }}
```

Per https://gohugo.io/templates/rss/

PR #1175, opened 2019-06-15
Switch to TailwindCSS

cnunciato:
This teeny little PR converts the website from Bootstrap (and a ton of custom styles) to [TailwindCSS](https://tailwindcss.com/), and takes an initial (but not totally complete) pass at styling the Docs and Blog sections of the site. The PR is intended to keep others unblocked with regard to what's under the `content` directory while I continue to hack away at pretty much everything else.

I've worked very quickly here, but I've done my best to try to keep the commits relatively small and well contained, so I apologize if this is difficult to follow. The effort, though, is in a good enough place structurally that I believe it's worthy of integrating and moving forward with.

Do note that much visual refinement remains, so don't read this PR as much more than a coarse-grained representation of what the end result will be -- but that said, the end's not too far off at this point, and Tailwind is proving to be a joy to work with as well.


content/legal/privacy.md
```diff
@@ -1,13 +1,9 @@
 ---
-title: "Privacy"
-date: 2018-06-07T15:26:26-07:00
+title: "Privacy Policy"
+url: "/privacy"
```
justinvp: Why is `privacy` staying in the root, but all the other ones are now going to be outputted under `/legal`? (We should check with Eric if any of these need to remain in the root for some reason. If we do end up moving them to be outputted under `/legal`, I'll have need to add redirects from the old location in the root to the new location under `/legal`.

cnunciato: Oh, that was unintentional. I wanted to create some navigation between them, but it seems better to leave them where they are, yes. I'll see if i can make that work.

cnunciato: All right, fixed up. Thanks for the catch!

PR #1202, opened 2019-06-21
Change concepts/ to redirect to reference/concepts/ instead of tour/

justinvp:
Having `concepts/` redirect to `reference/concepts/` is better than `tour/`.


justinvp:
Minor cleanup of this redirect within the old site, as it'll make the redirect from the old site to new site cleaner as well.

Issue #1023, opened 2019-04-29
"Copy to clipboard" button for code snippets


justinvp:
Code snippets should have "copy to clipboard" buttons. Pulled out from #550.

ericrudder:
similar request from subha:

BTW, a quick Copy button here would be nice.
https://pulumi.io/quickstart/azure/install-pulumi/

PR #1204, opened 2019-06-21
Add a "copy to clipboard" icon to the langchooser

joeduffy:
This change adds a copy icon to the upper right of the langchooser,
and copies the selected code body to the clipboard when clicked.

This is part of pulumi/docs#1023. I do think we should explore whether
we can use this in other places -- e.g., none of the API docs samples
will benefit from this because they don't use the langchooser -- but
doing so is a little more complex. We could just look for <div
class="highlight"> elements, but it's not clear, visually, where we'd
put the icon. Perhaps floating with a negative offset in the upper
right would work, but there are also plenty of <div class="highlight">
elements we probably wouldn't want this for.

### DIFF
--- a/assets/js/langchoose.js
+++ b/assets/js/langchoose.js
@@ -84,6 +84,45 @@ function selectLanguage(lang) {
     });
 }
 
+// copyNextCodeBlock copies the contents of the currently active code block following the element,
+// elem, which is the "copy" icon being clicked.
+function copyNextCodeBlock(elem) {
+    // Select the next visible code block.
+    var codes = $(elem).closest(".mdl-tabs").siblings(".highlight:visible");
+    if (codes.length) {
+        // Create an invisible text block containing the text from the code block, copy it to the
+        // clipboard, and then delete the invisible textarea.
+        var code = codes[0];
+        var text = document.createElement("textarea");
+        text.style.position = "fixed";
+        text.style.top = text.style.left = 0;
+        text.style.width = "2em";
+        text.style.height = "2em";
+        text.style.padding = 0;
+        text.style.border = "none";
+        text.style.outline = "none";
+        text.style.boxShadow = "none";
+        text.style.background = "transparent";
+        text.value = code.innerText;
+        document.body.appendChild(text);
+
+        // Copy the text.
+        text.focus();
+        text.select();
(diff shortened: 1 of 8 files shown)


joeduffy:
image

assets/js/langchoose.js
```diff
@@ -84,6 +84,45 @@ function selectLanguage(lang) {
...
 }
 
+// copyNextCodeBlock copies the contents of the currently active code block following the element,
+// elem, which is the "copy" icon being clicked.
+function copyNextCodeBlock(elem) {
+    // Select the next visible code block.
+    var codes = $(elem).closest(".mdl-tabs").siblings(".highlight:visible");
+    if (codes.length) {
+        // Create an invisible text block containing the text from the code block, copy it to the
+        // clipboard, and then delete the invisible textarea.
```
justinvp: Why not use [`clipboard-polyfill`](https://www.npmjs.com/package/clipboard-polyfill) like we do in the service?

https://github.com/pulumi/pulumi-service/blob/22dc0c6409f7a5f244c3c8871eaccd217e6b5b93/cmd/console2/src/app/settings/tokens/token-settings.component.ts#L70

```typescript
(clipboard as any).writeText(value);
```

And let the polyfill handle older browsers?

assets/js/langchoose.js
```diff
@@ -84,6 +84,45 @@ function selectLanguage(lang) {
     });
 }
 
+// copyNextCodeBlock copies the contents of the currently active code block following the element,
+// elem, which is the "copy" icon being clicked.
+function copyNextCodeBlock(elem) {
```
justinvp: I was going to be a bit smarter about copying `language-bash` and `language-shell`, stripping `"$ "` at the beginning of any lines.

I could also see needing to have a way to opt-out or specify which lines should be copyable (in the case of `language-bash` that has an initial command, and then the output subsequently, where we'd just want to copy the command.).

justinvp:
@joeduffy, if you'd like, I can take care of fixing the merge conflicts and addressing the feedback this week.

joeduffy:
Sounds great, I appreciate it!

justinvp:
Replaced by #1363

PR #1205, opened 2019-06-21
pulumi_aws/ecs/service : fixed parameter name

burnzy:
Updated  "ordered_placement_strategy" parameter to correct one: "ordered_placement_strategies"


justinvp:
Hi @burnzy! Thanks so much for the PR! Unfortunately, this is a generated page, so any change here will just be overwritten when the API docs are regenerated, and it doesn't look like the fix is as simple as a typo fix. `pulumi-terraform` [pluralizes names](https://github.com/pulumi/pulumi-terraform/blob/13fc8ad0d1dcc1d73aa294c71a447de579fb4a63/pkg/tfbridge/names.go#L57) that will become array-shaped Pulumi values, and it doesn't look like we're doing the right thing when generating the docs.

I've opened https://github.com/pulumi/pulumi-terraform/issues/398 to track this issue.

Thanks again!

burnzy:
I figured they were auto-generated, but I saw the " Edit This Page" link there so thought I'd give it a try anyway. Thanks for the explanation 👍

PR #1206, opened 2019-06-22
Add fusion stack

justinvp:
Any changes to the `fusion` branch will be deployed to https://fusion.pulumi.io/.

PR #1207, opened 2019-06-23
Fix broken images in fusion branch

justinvp:
Fixes broken images in the fusion stack. See https://github.com/pulumi/docs/pull/1206#issuecomment-504700119).

Instead of trying to deal with whether `BaseURL` has a trailing slash or not, there were many places that were prefixing `BaseURL` where a relative URL would work just fine. In such cases, just delete the unnecessary `BaseURL`. In the small number of remaining cases where we want to emit an absolute URL, use the `absURL` function instead, which is smart about handling missing/extra slashes.


justinvp:
I'm going to merge this to fix broken images on https://fusion.pulumi.io, but happy to address any feedback subsequently.

Issue #1213, opened 2019-06-24
GREAT WEBSITE MIGRATION OF 2019


justinvp:
Rollup of remaining tasks to rollout the combined site. We'll of course have some additional follow-up work afterwards to further deduplicate content, improve the structure & information architecture, etc.

### P1
- [x] [cnunciato] Create `robots.txt`
- [x] [cnunciato] Restore `meta_*` items that were deleted
- [x] [cnunciato] Docs: Search form and results styling
- [x] [cnunciato] Review pricing page (Chris has some feedback)
- [x] [justinvp] Another `pulumi/docs` branch integration from  `master` => `fusion`  (changes since 6/17) (#1220)
- [x] [justinvp] Improve `/docs` page (no longer needs to look like a website “homepage”; add Pulumi intro and diagrams) (#1239)
- [x] [justinvp] Regenerate all API docs (#1227)
- [x] [justinvp] Remove aliases (https://github.com/pulumi/docs/pull/1210)
- [x] [justinvp] Update CLI `gen-markdown` command for new site (https://github.com/pulumi/pulumi/pull/2869)
- [x] [chrsmith] Add latest blog posts (changes since 6/16) (https://github.com/pulumi/docs/pull/1218)
- [x] [chrsmith] Add latest blog posts (changes since 7/1)

### P2
- [x] [cnunciato] Add the "hover menu" for top-level nav items (quick links into the section)
- [x] [cnunciato] Code samples aren’t getting syntax highlighting in www content
- [x] [cnunciato] Mobile experience review / tweaks
- [x] [justinvp] Port change to hide language prologue classes by default (https://github.com/pulumi/docs/pull/1212)
- [x] [justinvp] Nav bar should show which section is active (#1216)

### P3
- [x] [cnunciato] Some heroes still need their background gradients/glows
- [x] [cnunciato] a few places still need those purple-and-gray waves 
- [x] [cnunciato] Some color combinations still aren’t quite right

### P4
- [ ] [justinvp, #1325] Fix styles used on AWS Crosswalk images to use Tailwind
- [ ] [justinvp, #1325] Looks like we have both `/image/pulumi.png` and `/social/pulumi.png` used as meta images. We should standardize on `/image/pulumi.png` and update any [redirects](https://github.com/pulumi/home/pull/507) as appropriate.
- [ ] [cnunciato, #1277] We seem to be generating an `rss.xml` file for all sections; is there a way to make it only generate for `/blog/rss.xml`?

## Rollout

[...]

PR #1217, opened 2019-06-24
Add search

cnunciato:
Adds the search form to docs sidebars.

(GIF compression clobbers it, but results are shown in containers also on the desktop.) 

Part of #1213.

### DIFF
--- a/content/docs/search.md
+++ b/content/docs/search.md
@@ -1,7 +1,5 @@
 ---
 title: Search Results
 layout: search
-aliases:
-    - search.html
 noindex: true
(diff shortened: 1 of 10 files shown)


content/docs/search.md
```diff
@@ -1,5 +1,7 @@
 ---
-layout: "search"
-aliases: ["search.html"]
+title: Search Results
+layout: search
+aliases:
+    - search.html
```
justinvp: FYI, we don't need the alias anymore; it's been deleted in #1210

PR #1223, opened 2019-06-25
Remove top margin from code snippets under lang tabs

justinvp:
Before (master)

Screen Shot 2019-06-25 at 9 44 40 AM

Screen Shot 2019-06-25 at 9 44 48 AM

Before (fusion)

Screen Shot 2019-06-25 at 9 40 29 AM

Screen Shot 2019-06-25 at 9 40 55 AM

After (fusion)

Screen Shot 2019-06-25 at 9 39 55 AM

Screen Shot 2019-06-25 at 9 40 45 AM

### DIFF
--- a/assets/sass/_langchoose.scss
+++ b/assets/sass/_langchoose.scss
@@ -117,6 +117,21 @@ $tab-font-size: 14px !default;
     cursor: pointer;
 }
 
+// Remove the top margin from code snippets under lang chooser tabs so the tabs sit directly on top.
+.langtab-tabs + .clearfix + .highlight,
+.langtab-tabs + .clearfix + .highlight + .highlight,
+.langtab-tabs + .clearfix + .highlight + .highlight + .highlight,
+.langtab-tabs + .clearfix + .highlight + .highlight + .highlight + .highlight,
+.language-prologue-javascript + .highlight,
+.language-prologue-typescript + .highlight,
+.language-prologue-python + .highlight,
+.language-prologue-go + .highlight
+{
+    .chroma {
+        margin-top: 0;
+    }
+}
+
 // Hide adjacent siblings of .language-prologue-* classes from the outset, so they don't all initially
 // show as visible before all but the selected language are hidden.
 .language-prologue-javascript + *,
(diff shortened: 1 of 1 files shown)


assets/sass/_langchoose.scss
```diff
@@ -117,6 +117,21 @@ $tab-font-size: 14px !default;
     cursor: pointer;
 }
 
+// Remove the top margin from code snippets under lang chooser tabs so the tabs sit directly on top.
+.langtab-tabs + .clearfix + .highlight,
+.langtab-tabs + .clearfix + .highlight + .highlight,
+.langtab-tabs + .clearfix + .highlight + .highlight + .highlight,
+.langtab-tabs + .clearfix + .highlight + .highlight + .highlight + .highlight,
```
justinvp: This is all pretty gross/lame -- but I couldn't think of a better way to select each of the possible code snippets under the lang tabs. Open to feedback if there's a better way to do it.

cnunciato: Ah! This is a job for the [general sibling combinator](https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator):

```
.mdl-tabs {
    ...

    ~ .highlight > .chroma {
        margin-top: 0;
    }
}
```

... seems to do the trick.

cnunciato: ... although, I wonder if that messes up anything _below_ the language chooser. 🤔 Probably does.

justinvp: Thanks! TIL 😄. From poking around the site, it seems to work OK and not disrupt anything below the language chooser.

assets/sass/_langchoose.scss
```diff
@@ -117,6 +117,13 @@ $tab-font-size: 14px !default;
     cursor: pointer;
 }
 
+.langtab-tabs {
```
justinvp: Instead of putting this in `.mdl-tabs` I kept it more specific to the lang chooser, as I plan to extract out the tabs into their own more general purpose component, because I want to reuse tabs for things other than the language chooser. For example, one piece of feedback we got on the new get started guide (from Twitter) was to just have tabs for platform selecting an OS when installing Pulumi (as it wasn't immediately obvious we supported Windows since we do platform detection and only show a single platform, macOS, when viewing from an iPhone).

justinvp:
> At some point it might be nice to revisit this component and do a little refactoring

Yep.

PR #1225, opened 2019-06-25
Add a robots.txt file

cnunciato:
This change adds a robots.txt template authorizing seach-engine indexing for all pages of the production site.

It assumes that all non-production builds will override the `baseURL` configuration property with something other than `https://pulumi.com/`.

Part of #1213.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1,4 +1,4 @@
-baseURL = "https://pulumi.com/"
+baseURL = "https://www.pulumi.com/"
 languageCode = "en-us"
 title = "Pulumi"
 
--- a/config.toml
+++ b/config.toml
@@ -1,4 +1,4 @@
-baseURL = "https://pulumi.io/"
+baseURL = "https://pulumi.com/"
 languageCode = "en-us"
 title = "Pulumi"
 
@@ -6,6 +6,7 @@ disableKinds = ["category", "taxonomyTerm"]
 sectionPagesMenu = "main"
 pygmentsCodeFences = true
 pygmentsUseClasses = true
+enableRobotsTXT = true
 
 [outputFormats.RSS]
     baseName = "rss"
--- a/layouts/robots.txt
+++ b/layouts/robots.txt
@@ -1,0 +1,5 @@
+User-agent: *
+{{- /* baseURL is a template.URL, so we need to convert it before we compare it.  */ -}}
+{{- if ne (string .Site.BaseURL) "https://pulumi.com/" }}
+Disallow: *
+{{ end }}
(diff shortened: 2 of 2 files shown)


config.toml
```diff
@@ -1,11 +1,12 @@
-baseURL = "https://pulumi.io/"
+baseURL = "https://pulumi.com/"
```
justinvp: Shouldn't this be `https://www.pulumi.com/` (with the `www`)?

Right now `https://pulumi.com` will redirect to `https://www.pulumi.com`. Personally, I wish it was the other way around and `https://pulumi.com` was the canonical URL. @chrsmith, do you remember why we decided to go with `www.pulumi.com` over `pulumi.com`? Should we revisit as part of this website merge effort? (I'm a little afraid if we switch, it'll somehow affect our SEO; but if we want to switch, for a more succinct URL, it'd probably be better to do it sooner than later).

layouts/robots.txt
```diff
@@ -0,0 +1,5 @@
+User-agent: *
+{{- /* baseURL is a template.URL, so we need to convert it before we compare it.  */ -}}
+{{- if ne (string .Site.BaseURL) "https://pulumi.com/" }}
```
justinvp: I wonder if we should add the following to our `config.toml` under `[params]`:

```toml
[params]
    productionURL = "https://www.pulumi.com/"
    # ...
```

And then change this comparison to use that, e.g. something like:

```
{{- if ne (string .Site.BaseURL) .Site.Params.productionURL }}
```

Alternatively (and perhaps better)...

I sent you and @chrsmith an email pointing to Hugo's Configuration Directory and `hugo --environment` docs: https://gohugo.io/getting-started/configuration/#configuration-directory

If/when we end up leveraging that to have separate config.toml variables for staging vs. production (and therefore are using `hugo --environment staging` and `hugo --environment production` when building for staging vs. production), we could do something like:

```
{{- if ne .Hugo.Environment "production" }}
```

config.toml
cnunciato: This is one of those things that feels a bit squishy to me, but if I had to make a call now, I'd probably suggest going with `www`, yeah. Other than aesthetics, there doesn't seem to be any downside to it, whereas with a bare domain, a handful of technical issues come into play (cookies, DNS options, etc.) that could present complications for us down the road. So I'm okay rolling with www if you all are.

layouts/robots.txt
cnunciato: Yep, agreed. I read up a bit on environments yesterday and that seemed like the way to go; I figured we'd do that in the shortish term post-cutover, and come back to make an adjustment here accordingly.

PR #1237, opened 2019-06-26
Blog layout and styling improvements

chrsmith:
All of the content changes are specific to replacing "4x space indentation" for code blocks, to using an explicit "```typescript". Since the former wasn't consistently applying syntax highlighting.

If you could review the changes to the `/layouts/` folder, I'd love to get your take on if I'm using TailWind CSS the right way. Things appear to be better:

Adjustments to blog sidebar, such as adding the "recent posts" list and making the tag cloud elements a little smaller.

image

On mobile the sidebar hides and is replaced with a toggleable menu. Notice how the images in the post summaries no longer appear to float.

image

Expanded

image

The `partials/blog/poster.html` now puts a max size on the image and floats it to the right, which makes them less overbearing.

image

Add a very basic footer to individual blog posts, with the date and tags.
image

---

**[chrsmith] Syntax highlighting seems to be missing from at least some posts (https://fusion.pulumi.io/blog/running-containers-in-aws-the-lowdown-ecs-fargate-and-eks/)**

Evidently syntax highlighting is only applied sometimes if you indent a code block with 4x spaces. (Perhaps
the heuristic it uses to determine the language isn't able to for shorter snippets?)

Switched to using explicit ```typescript (or similar) blocks.

**[chrsmith] Blog code blocks have awkward "scroll bars?" even when there is not content to scroll?**

I saw this in the past, but couldn't find a repro anywhere. So perhaps this was fixed with some other
CSS change we took. If we see it again I can take a look.

**[chrsmith] Sort order for Author and Tag pages was incorrect**

Fixed.

**[chrsmith] Also, was nice that these had counts on them previously so you could quickly get a feel for what kind of other content was available related to topics of interest.**

Won't fix. This adds a lot of visual noise for little benefit. e.g. would you not click the tag if there was only 3 posts under that topic?

**[chrsmith] size of titles relative to other content on the page feels too small.**

That's reasonable feedback, though we are using the same font-size for headings throughout
the site. So let's leave it as-is for now, and we can touch it up if you still feel the same
way later.

**[chrsmith] We intentionally don't list blog dates on them, to keep them more timeless, let's please remove**

Done. The dates are still listed in the "blog listing page", but at the bottom of the post on individual pages.

**[chrsmith] On mobile, blog tags take up a lot of vertical real estate, including on blog posts themselves -- maybe just show on the homepage, and maybe put in a drop-down?**

Yes. They also take up a lot of space after switching to the "earlier" tags. Adjusting default size, and hiding on mobile.

- Made the elements smaller, so they fill the avialable space better.
- Also hid the entire tag cloud on mobile, though it can be displayed after clicking on a hamberger menu.

**[chrsmith] The preview images are too large, let's keep it similar to our old format of having them right aligned and smaller**

Done.

**[chrsmith] I didn't see where the blog article lists its tags; this used to be at the bottom (see "Topics: x, y, z").**

Done.

**[chrsmith] Blog missing any "related posts" or "recent posts" sidebar - hard to get a feel for what other content is available when you are on a post**

Done, added recent posts lists (for large layouts only.)

Fixing another batch of issues from https://github.com/pulumi/docs/issues/1213.


chrsmith:
@cnunciato All of those sound great. Commit away, then squash + merge. 👍

layouts/partials/blog/sidebar.html
```diff
@@ -2,31 +2,62 @@
...
+        <ul class="list-none p-0 mb-4 inline-flex flex-wrap text-xs">
+            {{ $title := .Title | lower }}
+            {{ range $tag, $_ := .Site.Taxonomies.tags }}
+                {{ if $tag }}
+                    <li>
+                        {{ $classes := "bg-white border border-gray-200 text-blue-600 hover:border-blue-600" }}
+                        {{ if eq $tag $title }}
+                            {{ $classes = "border-blue-600 bg-blue-600 text-white hover:text-white" }}
+                        {{ end }}
+                        <a class="px-2 py-1 mr-1 cursor-auto rounded transition-all {{ $classes }}" href="{{ .Page.Permalink }}">
```
justinvp: Why `cursor-auto`?

I was playing around with the site with this change applied, and clicking the tags felt off without the regular pointer cursor for links.

tags

justinvp: Here's what it'd look like with standard pointer cursor for links, which feels better to me:

tags2

cnunciato: Oh, no! That was a mistake on my part. I'll fix. Thanks for noticing it.

cnunciato: https://github.com/pulumi/docs/pull/1245

PR #1238, opened 2019-06-26
Add syntax highlighing to layout templates

cnunciato:
This change adds a Hugo partial for highlighting code snippets in layouts.

Part of #1213.

### DIFF
--- a/layouts/partner/aws.html
+++ b/layouts/partner/aws.html
@@ -1,10 +1,10 @@
 {{ define "hero" }}
     <header class="bg-purple-500 py-24">
-        <div class="container mx-auto text-white md:flex">
+        <div class="container mx-auto text-white md:flex px-4 md:px-0">
             <div class="md:w-1/2">
                 <h1 class="text-white">Modern Infrastructure as Code for AWS</h1>
                 <div class="flex">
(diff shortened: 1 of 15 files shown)


layouts/partner/aws.html
```diff
@@ -76,42 +75,38 @@ <h2>Creating a Serverless REST API</h2>
...
 let counterTable = new cloud.Table("counterTable", "route");
 
-// Create an API endpoint
+// Create an API endpoint.
 let endpoint = new cloud.API("hello-world");
 
 endpoint.get("/{route+}", (req, res) => {
     let route = req.params["route"];
-    console.log(`Getting count for '${route}'`);
+    console.log("Getting count for " + route);
```
justinvp: Bummer. Too bad there isn't a way to easily escape backticks.

cnunciato: Yeah, it’s a huge bummer. I did notice, though, that the backticks in the code samples I put in frontmatter variables _are_ being rendered, so there might be a way to do this somehow. I also really hate having to force all the code to the left just to avoid leading space. So there’s still room for improvement here. I’ll keep messing with it. :)

Issue #1173, opened 2019-06-14
Get Started follow-ups


justinvp:
- [x] (bermudezmt) We have various abbreviated “Get Started” flows throughout the docs (e.g. before each tutorial). Do a pass throughout the entire doc content and ensure we’re consistent about installing Pulumi, prerequisites, configuring providers, etc., and recommending the overall Get Started guides as the single place to get started with Pulumi before anything else.
- [ ] (bermudezmt) Inline streamlined “configuring” steps for each of the clouds instead of only linking to the more detailed pages
- [x] (justinvp) Update the Azure examples to not flow the location, as that's no longer required (Moved to its own issue #1535)
- [ ] (bermudezmt & cnunciato) Intros and Summaries: Provide nicer intros at the start of each cloud's guide (e.g. inline some of the "Why Pulumi?" content for Kubernetes), and nicer summaries in the concluding "Next Steps" pages, including links to API docs, GitHub repos, and links to "Architecture & Concepts". Note from Justin: For Kubernetes in particular, Mike suggested taking the first section of https://www.pulumi.com/docs/reference/clouds/kubernetes/ (intro, diagram, and bullet points) and put that in the Get Started guide for Kubernetes.
- [ ] (bermudezmt & cnunciato) Add overview diagrams in appropriate places, e.g. on the new /docs home page and in /docs/reference. Note from Justin: I removed a couple diagrams that Mike had created/added, as they didn't fit with the new Get Started guide, but we wanted to bring them back in some form, either somewhere on the /docs landing page itself, or in the "Learning Pulumi" selection in /docs/reference. [These](https://github.com/pulumi/docs/pull/1169/files#diff-ef32087fbd8a1d30da698a6e6dd04394L34-L36) were the lines that were removed ([hourglass.svg](https://github.com/pulumi/docs/blob/master/static/images/docs/quickstart/hourglass.svg) and [general-architecture.svg](https://github.com/pulumi/docs/blob/master/static/images/docs/quickstart/general-architecture.svg)). Or you may have better ideas.

- [x] (bermudezmt) Add link to install the CLI somewhere more prominently in the Get Started section, for folks who are already familiar with Pulumi and just want the CLI, and don't want to walk through the wizards. (I believe this is now addressed with #1451 as "Download and Install" is now always shown).
[...]

PR #1239, opened 2019-06-27
Cleanup the /docs page

justinvp:
Cleanup the `/docs` page

- Add left nav
- Make`/docs` page look slightly less like a "homepage"
- Add top-level "Get Started" and "Reference" headings to left nav

I tried several iterations of making it more like a doc article, with some diagrams, but ended up settling on it being pretty similar to what we have currently, just with the left nav and search bar on the right.

I also added a `.btn-secondary` class and used it in a few places. Christian, let me know if you prefer doing this a different way.

Screen Shot 2019-06-26 at 4 58 33 PM

Screen Shot 2019-06-26 at 4 58 42 PM

Screen Shot 2019-06-26 at 4 58 49 PM


layouts/partials/docs/toc.html
```diff
@@ -1,21 +1,50 @@
...
     {{ $menu = .Site.Menus.aws }}
     {{ $has_toc = true }}
+    {{ $isAWS = true }}
 {{ end }}
 
-{{ if $has_toc }}
-    <aside class="toc">
+<aside class="toc">
+    <h3 class="no-anchor">
+        <a href="{{ relref . "/docs" }}">Docs</a>
```
justinvp: I went back and forth on including this. Ultimately, I left it in, similar to how we have a heading for the Pulumi Blog in the left nav. But I'd be OK with getting rid of it as it's a bit redundant with the nav bar.

layouts/partials/docs/toc.html
```diff
@@ -65,8 +94,15 @@
...
-    </aside>
-{{ end }}
+
+    {{ end }}
+
+    {{ if $isGetStarted }}
+        <h4 class="no-anchor">
+            <a href="{{ relref . "/docs/reference" }}">Reference</a>
+        </h4>
+    {{ end }}
```
justinvp: The changes in this file aren't great -- but it was the simplest way to add these top-level nav elements to the left nav without a more involved change.

assets/sass/_buttons.scss
```diff
@@ -16,12 +16,20 @@
...
+    }
+}
+
 .btn-sm {
     @apply text-xs;
 }
 
 .btn-lg {
-    @apply btn py-3 px-5;
+    @apply py-3 px-5;
```
chrsmith: Just for my own understanding, to get the same effect you would need to add `class="btn btn-large"` to the element? What motivated you to make the change? (No opinion on the change, just trying to understand the mystery of Tailwind and `@apply`.)

layouts/partials/docs/toc.html
chrsmith: Earlier when we were talking about a docs redesign, we had a lot of ideas around making the left nav more useful. So if anything, we might want the next step to be adding more stuff to the left-nav. Perhaps quick links like "Clouds" or "Languages"?

I can only guess at what that would look like, but I would just be a little hesitant to removing it.

chrsmith: ... If we do add more if-conditions, we probably want to initialize all those one-of flags like `$isAWS` into a map, and then do lookups. (So we can just use string constants instead of N-variables, in case that is cleaner.) e.g. `if (Get $flags "is-aws")`

justinvp: > Earlier when we were talking about a docs redesign, we had a lot of ideas around making the left nav more useful. So if anything, we might want the next step to be adding more stuff to the left-nav. Perhaps quick links like "Clouds" or "Languages"?

Yeah, as a next step, we should look at reorganizing all the docs content. There's lots of opportunities to make things more visible that are currently all buried under `/reference`. "Clouds" and "Languages" are good candidates, also "Tutorials".

> I can only guess at what that would look like, but I would just be a little hesitant to removing it.

Oh, I didn't mean removing the left nav, I just wasn't sure if we should include a "Docs" header at the top of the nav or not.

assets/sass/_buttons.scss
justinvp: > Just for my own understanding, to get the same effect you would need to add `class="btn btn-large"` to the element? What motivated you to make the change?

@cnunciato can correct me if I'm wrong, but I think the intention is for these to be used like `class="btn btn-lg"` and `class="btn btn-sm"` (and indeed all of our uses are done this way). The Tailwind docs discuss this very situation at https://tailwindcss.com/docs/extracting-components/#keeping-things-composable

When I added the non-primary `btn-secondary` class, I wanted to use it like `class="btn btn-secondary btn-lg"`, but the `btn-lg` was overriding my color styles for `btn-secondary`, so it looked like a regular `btn`.

So I made `btn-lg` be like `btn-sm` and not have it `@apply btn`, which fixed my issue.

justinvp:
> The only thing I'd suggest on the docs page is to make some adjustments to the styling of the Getting Started page:

Mind if I address these separately as part of #1173 (I've added these there)?

assets/sass/_buttons.scss
cnunciato: Yep. 👍 Sorry that messed you up &mdash; I didn't realize I'd applied the `btn` styles to `btn-lg`; I must've done that early on while kicking the tires on Tailwind, and never came back to it.

PR #1241, opened 2019-06-27
Updates for CLI 0.17.21

ellismg:
Some of the older versions had been missing from the table so I added them in as well.  I hope this is not too disruptrive to the fusion stuff, @justinvp.


content/reference/pkg/python/pulumi_datadog/gcp/_index.md
```diff
@@ -0,0 +1,104 @@
+---
```
lukehoban: Have we shipped Datadog publicly yet?

stack72: Datadog package has been published for a few days so I've been asking to get the repo opened to allow this

stack72:
> Looks like we were behind on a lot of docs for a lot of packages here. Have we not been updating docs as part of release process?

I brought this up about a week ago and it was suggested that the docs would be regenerated for the new site and to just wait till then but it seemed to have been a bit longer than expected

lukehoban:
> I brought this up about a week ago and it was suggested that the docs would be regenerated for the new site and to just wait till then but it seemed to have been a bit longer than expected

Hmm - okay.  I don't see any reason to hold off here.  We can easy generate latest in the new site independent of also maintaining a good experience for users of the existing site.

stack72:
@lukehoban I will send a PR now that updates them

lukehoban:
> @lukehoban I will send a PR now that updates them

Doesn’t this PR do that?

PR #1243, opened 2019-06-27
aws.route53.Record.type expects RecordTypes

albestia:


stack72:
I believe the PR I am working on will solve this - I will let you know if it’s the case

ellismg:
Thanks for the contribution, @albestia!

This won't work as is, I think, because the examples are auto-generated by converting some existing code using `tf2pulumi`.  @pgavlin do you know how this is supposed to work in tf2pulumi?  Should we have been typing these things as `RecordType | string` in the TS layer so the above would have continued to work?  /cc @CyrusNajmabadi @stack72

pgavlin:
> Should we have been typing these things as `RecordType | string` in the TS layer so the above would have continued to work?

Yes, this seems like the right thing to do.

CyrusNajmabadi:
Yup.  Seems reasonable.  @stack72 do you want to take this?

stack72:
FYI, I have made a fix in pulumi-aws that changes the behaviour here to what @ellismg / @pgavlin suggested

PR #1248, opened 2019-06-27
Promote master to production

ellismg:


chrsmith:
Closing out as to not get ran over by the GWMO2019.

Issue #1177, opened 2019-06-15
Suggestion: Change Modify step in Getting Started to truly Modify


joeduffy:
Right now in our Getting Started flow, we go from [one example (an S3 bucket)](https://pulumi.io/quickstart/aws/review-project/) to [something entirely different (an EC2 instance running a Python web server)](https://pulumi.io/quickstart/aws/modify-program/). I personally found this confusing to label as "modifying" my program, since I'm actually replacing it altogether. Not only is the EC2 instance example bigger and harder to follow than the nicely simple S3 one we started with, this seems like a missed opportunity to give me a sense of truly updating resources. I would encourage thinking about something else instead, like enabling encryption at rest on our S3 bucket, which is simpler and will give the end user a quicker, more relevant feel for what a typical update entails.

joeduffy:
Another reason to go with the S3 example, rather than EC2. I think if you have an old AWS account created prior to 2013 -- like I do -- it still tries to use EC2-Classic by default when deploying into `us-east-1`, which fails with the default steps specified in our tutorial:

```
    error: Plan apply failed: Error launching source instance: VPCResourceNotSpecified: The specified instance type can only be used in a VPC. A subnet ID or network interface ID is required to carry out the request.
```

I might be mistaken, but I _think_ that's what's going on.

joeduffy:
For instance, this is a nice modification to the base Python example that creates a KMS key and enables encryption on our S3 bucket:

```
import pulumi
from pulumi_aws import kms, s3

# Create a KMS Key for S3 server-side encryption;
key = kms.Key('my-key')

# Create an AWS resource (S3 Bucket)
bucket = s3.Bucket('my-bucket',
    server_side_encryption_configuration={
        'rule': {
            'apply_server_side_encryption_by_default': {
                'sse_algorithm': 'aws:kms',
                'kms_master_key_id': key.id
            }
        }
    })

# Export the name of the bucket
pulumi.export('bucket_name',  bucket.id)
```

It can be done with an in place update, and is pretty easy to follow:

```
Updating (dev):

     Type                 Name            Status      Info
     pulumi:pulumi:Stack  quickstart-dev
 +   ├─ aws:kms:Key       my-key          created
 ~   └─ aws:s3:Bucket     my-bucket       updated     [diff: ~serverSideEncryptionConfiguration]

Outputs:
    bucket_name: "my-bucket-127bd18"

Resources:
    + 1 created
    ~ 1 updated
    2 changes. 1 unchanged

[...]

PR #1253, opened 2019-06-28
Change Modify step in AWS Get Started to truly modify

justinvp:
Updates the AWS Get Started modify step to truly modify -- adding encryption to the S3 bucket via KMS key.

This also addreses an issue that prevented the modified program from working if you had an older AWS account (created prior to 2013) that uses EC2-Classic by default.

Part of #1177


justinvp:
Replaced by #1290

PR #1266, opened 2019-07-02
Add hover menu

cnunciato:
This change adds hover menus for non-touch clients to the Product and Why Pulumi main navigation items.

Part of #1213.

### DIFF
--- a/assets/config/tailwind.config.js
+++ b/assets/config/tailwind.config.js
@@ -71,4 +71,11 @@ module.exports = {
             },
         },
     },
+
+    variants: {
+        margin: [
+            "responsive",
+            "hover",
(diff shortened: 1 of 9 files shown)


assets/config/tailwind.config.js
```diff
@@ -71,4 +71,11 @@ module.exports = {
             },
         },
     },
+
+    variants: {
```
cnunciato: https://tailwindcss.com/docs/margin/#responsive-and-pseudo-class-variants

chrsmith:
Will pull locally and rebase on top of the latest `fusion` bits, which has the right CI/CD hookup.

PR #1273, opened 2019-07-03
Update SAML docs

praneetloke:
Here's a screenshot of the new doc page for SAML. The other changes are updates to the existing doc pages.

Screenshot_2019-07-03 Overview Of SAML SSO Configuration(1)


content/docs/reference/service/saml-aad.md
```diff
@@ -41,10 +41,32 @@ to replace `acmecorp` with whatever your specific Pulumi organization's name is.
...
+
+![User Attributes & Claims](/images/docs/reference/service/saml-aad/attributes-and-claims.png)
+
+1. In the **Manage User Claims** panel, expand the **Choose name identifier format** and select **Email address**.
+
+![Manage Name Identifier Format](/images/docs/reference/service/saml-aad/name-identifier-format.png)
+
+1. Finally, click **Save** at the bottom of the **Manage User Claims** panel.
+
+**IMPORTANT:** Do not change the value of Name ID Format value once your users have started using Pulumi; not even switching its value between Email or Persistent.
```
chrsmith: Nit: "not even switching its value between Email or Persistent" is redundant. Same for other SAML org types.

praneetloke: I wanted to emphasize, that one shouldn't even change the value from one accepted value to another accepted value. That is, from email address to persistent or vice-versa, once their users start using the system. As that also changes the subject value.

PR #1274, opened 2019-07-03
Update the color palette

cnunciato:
This change updates the Tailwind configuration and Sass color maps with a palette based on Pulumi brand colors. (The source file containing these definitions [is now in Figma](https://www.figma.com/file/RZvzusO5se1hZkNcAInj5iLE/Pulumi-Colors?node-id=0%3A1).)

image

Part of #1213.


cnunciato:
☝️ Actually, one quick change to make to the Sass file...

cnunciato:
All right, should be good now. 😄

cnunciato:
@chrsmith Exactly, yeah. There are a few here that are still Tailwind's (the grays, yellows and reds are just fine for how we're using them), but the others were hand chosen by me based on the four colors in the brand set. 

These colors can also be propagated into the Console app to help reduce a bunch of duplication and variations we have there.

PR #1282, opened 2019-07-04
Change base URL to avoid sending telemetry on debug builds

chrsmith:
Hey guys, this is probably the wrong way to go about fixing the problem I'd like to solve. So I'd love your input if there is a better way to go about it.

The problem is that when running the site locally or in a staging environment, we want to disable sending telemetry data like Google Analytics page views.

We have a check in the `robots.txt` file and `head.html` partial that compare the `.Site.BaseURL` value. So if the site's base URL is not "www.pulumi.com", we don't send telemetry and instruct search engines not to index the site.
https://github.com/pulumi/docs/blob/master/layouts/robots.txt
https://github.com/pulumi/docs/blob/master/layouts/partials/head.html#L84

... but that actually doesn't work for the local development case, since the default base URL is www.pulumi.com. So without resorting to adding a new `IsProdSoTotallyEnableThings` field/value/property, it seems like changing the default base URL is the simplest thing to do... but that definitely seems error prone.

Have you seen better ways to address this type of thing?


chrsmith:
@justinvp PTAL. Switching to use `--environment` definitely makes things a lot cleaner. (Sorry I had failed to make this environment change earlier, I skipped it because we weren't parameterizing the signing URL anymore, and didn't realize we could use the same technique for this "is prod or not" check too.)

config/_default/config.toml
```diff
@@ -0,0 +1,29 @@
...
+languageCode = "en-us"
+title = "Pulumi"
+
+disableKinds = ["category", "taxonomyTerm"]
+sectionPagesMenu = "main"
+pygmentsCodeFences = true
+pygmentsUseClasses = true
+
+# Don't crawl non-prod instances.
+enableRobotsTXT = false
```
cnunciato: This has the effect of not _generating_ a robots.txt, which effectively tells the bots it's OK to be indexed. We should enable robots.txt in all environments, and then unless the Hugo env is `production`, `Disallow: *`.

layouts/robots.txt
```diff
@@ -1,5 +1,4 @@
 User-agent: *
-{{- /* baseURL is a template.URL, so we need to convert it before we compare it.  */ -}}
-{{- if ne (string .Site.BaseURL) "https://www.pulumi.com/" }}
+{{- if eq hugo.Environment "production" }}
```
cnunciato: This should be  `ne`, since we want _not_ to be crawled/indexed _unless_ production.

config/_default/config.toml
chrsmith: Ah, good catch.

layouts/robots.txt
chrsmith: Fixed, thanks.

PR #1284, opened 2019-07-06
Update ecs.md

markdingram:


bermudezmt:
Hi @markdingram - thanks for contributing. Sorry I'm only seeing this PR now.

The same change is covered in #1506, so I'm closing this in favor of that one.

PR #1290, opened 2019-07-08
Change Modify step in AWS Get Started to truly modify

justinvp:
(Replaces #1253, which was closed when the `fusion` branch was deleted).

Updates the AWS Get Started modify step to truly modify -- adding encryption to the S3 bucket via KMS key.

This also addreses an issue that prevented the modified program from working if you had an older AWS account (created prior to 2013) that uses EC2-Classic by default.

Part of #1177

### DIFF
--- a/content/docs/quickstart/aws/deploy-changes.md
+++ b/content/docs/quickstart/aws/deploy-changes.md
@@ -18,15 +18,14 @@ Pulumi computes the minimally disruptive change to achieve the desired state des
 ```
 Previewing update (dev):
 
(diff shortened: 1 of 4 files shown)


joeduffy:
I like it!

Do we plan on doing similar things to our other getting started flows?

P. S.  I missed this -- sorry -- suggest asking @bermudezmt to review these going forward!

justinvp:
> Do we plan on doing similar things to our other getting started flows?

Yes. The issue has tasks for the other clouds.

PR #1291, opened 2019-07-08
Use Purgecss to reduce the size of our CSS

justinvp:
Not ready for review yet. Will mark as ready for review when ready, including before/after gzip sizes.

FYI @cnunciato, @chrsmith

Follow-up from #1213

### DIFF
--- a/assets/config/postcss.config.js
+++ b/assets/config/postcss.config.js
@@ -5,20 +5,38 @@
  */
 module.exports = {
     plugins: [
-        // TailwindCSS
+        // TailwindCSS.
         require("tailwindcss")("./assets/config/tailwind.config.js"),
 
-        // Apply vendor prefixes for CSS features that aren't
-        // fully supported yet.
+        // Apply vendor prefixes for CSS features that aren't fully supported yet.
         require("autoprefixer")({
             overrideBrowserslist: [
                 "last 2 versions"
             ]
         }),
 
-        // Minify the CSS even further. (It works!)
-        require('cssnano')({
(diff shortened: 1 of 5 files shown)


assets/config/postcss.config.js
```diff
@@ -5,20 +5,38 @@
  */
 module.exports = {
     plugins: [
-        // TailwindCSS
+        // TailwindCSS.
```
chrsmith: Nit: If a comment is a complete sentence, like on line 11, we'll want proper grammar. (Begin with a capital letter, end with a period.) But for a trivial comment like this, no need to add a period at the end.

assets/config/postcss.config.js
```diff
@@ -5,20 +5,38 @@
...
                 "last 2 versions"
             ]
         }),
 
-        // Minify the CSS even further. (It works!)
-        require('cssnano')({
-            preset: 'default',
-        })
+        // Use the following plugins when not building for "development":
+        ...process.env.NODE_ENV !== "development" ? [
```
chrsmith: I am I reading this correctly in that the `process.env.NODE_ENV !== "development" : [ ... ] : []` is evaluated first, and _then_ the `...` operator expands all of the elements? If so, that's super-subtle and confusing.

Would we instead just want to move this into a package level variable? Like:

```
// Include css-purgecss on non-dev builds to reduce CSS file size.
let plugins = process.env.NODE_ENV === "development" ? [] : [
];
```

justinvp:
Replaced by #2444

PR #1292, opened 2019-07-08
Delete package-lock.json

justinvp:
`make ensure` uses `yarn install` and we already have a `yarn.lock`, so there's no need for a redundant `package-lock.json` for `npm`.

However, if you run `make serve` without having run `make ensure`, Hugo will fail with the following error:

> Building sites … ERROR 2019/07/08 08:16:02 error: failed to transform resource: POSTCSS: failed to transform "css/styles.css" (text/css): PostCSS not found; install with "npm install postcss-cli". See https://gohugo.io/hugo-pipes/postcss/

Running `npm install postcss-cli` will reintroduce `package-lock.json`, so I could see why we might want to consider switching this repo to using `npm` over `yarn` to avoid this. But our use of `yarn` is pretty ubiquitous.


praneetloke:
Should we add `package-lock.json` to the `.gitignore` file then?

PR #1303, opened 2019-07-09
Test out removing limited parallelism

lukehoban:
DO NOT MERGE.  

Testing out how much performance benefit we see if we allow unbounded parallelism.  

This change should be roughly the same as a change that took 31 minutes to run in Travis. https://travis-ci.com/pulumi/docs/builds/118217263.

### DIFF
--- a/content/about/_index.md
+++ b/content/about/_index.md
@@ -2,8 +2,8 @@
 title: About
 type: page
 layout: about
-menu:
-    header:
-        weight: 6
-        identifier: about
+# menu:
+#     header:
+#         weight: 6
+#         identifier: about
 ---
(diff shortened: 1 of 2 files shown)


chrsmith:
Note that for a `pull_request` job we'll only run a `pulumi preview`. So this actually won't trigger a `pulumi up` until you merge this PR. (Which from the looks of it, isn't what you were intending.)

The major difference between a `preview` and an `update` is that we aren't uploading snapshots. So perhaps an alternate test would be to see how _adding_ `--parallel 10` to the `pulumi preview` step impacts the performance.

lukehoban:
Yeah - noticed  that after opening this.  

We'll need to test out a real update - the actual traffic to AWS is totally different for update vs. preview.  I guess there's no way to do that without a dev stack - I'll see about doing it offline.

Alternatively I'd suggest we just up the concurrency to 100 in master and wait to see if that causes any issues.

PR #1306, opened 2019-07-09
Update Google Cloud Build docs

### LINKED ISSUE #1304: Clarify BUILD_TYPE for Google Cloud Build docs
BUILD_TYPE comes from a build trigger, we should clarify and add more explanation about this in the docs.

https://www.pulumi.com/docs/reference/cd-google-cloud-build/

praneetloke:
Fixes #1304.

Here's how it looks now:

Screenshot_2019-07-09 Google Cloud Build


content/docs/reference/cd-google-cloud-build.md
```diff
@@ -68,6 +70,16 @@ steps:
   - 'BUILD_TYPE=$_BUILD_TYPE'
 ```
 
+In the above configuration, we use a substitution `_BUILD_TYPE` to indicate to the build environment about the type of build it is running. This allows us to run a `preview` or an `update` based on the build type. To configure the substitution, you need to setup build triggers for each type of branch builds you intend to support for your project.
```
clstokes: I think we should make it _even_ more clear that both substitutions are provided by the build trigger. Something to the effect of:

> In the configuration above, both `_BUILD_TYPE` and `_INSECURE_SUBSTITUTION_PULUMI_ACCESS_TOKEN` are configured as a part of the Cloud Build Trigger that runs the Cloud Build. You can read more about triggers at ...

praneetloke: Done.

PR #1310, opened 2019-07-09
Add instructions and make build troubleshooting for Linux

bermudezmt:
- Break down macOS and Linux-specific installation steps
- Add list of npm packages that needed to be installed individually on Ubuntu

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,11 +20,23 @@ just check the resulting files directly into the repo under `./content/reference
 
 ### Prerequisites
 
+- [Node.js and npm](https://www.npmjs.com/get-npm)
+- [Ruby](https://www.ruby-lang.org/en/downloads/)
+- [Yarn](https://yarnpkg.com/en/docs/install) for installing dependencies in package.json
+- [Hugo](#hugo)
+- [Go](https://golang.org/dl/)
+
 #### Hugo
 
 The website is powered by [Hugo](https://gohugo.io).
 
 **IMPORTANT.** Recent versions of Hugo have bugs in the markdown renderer (Blackfriday) that prevents fenced code from rendering correctly in lists when a language is specified. Many of our tutorials are made up of ordered lists of steps, each step containing a code snippet. Until those bugs are fixed, and Hugo has adopted the version of Blackfriday with the fixes, we'll pin to [Hugo v0.55.4](https://github.com/gohugoio/hugo/releases/tag/v0.55.4). Tracking issue: https://github.com/pulumi/docs/issues/1091
+
+#### macOS
+
+The following commands use the package manager, [Homebrew](https://brew.sh/). 
+
+##### Install Hugo 
 
 If you already have Hugo installed, uninstall it:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,12 +20,24 @@ just check the resulting files directly into the repo under `./content/reference
 
 ### Prerequisites
 
+- [Node.js and npm](https://www.npmjs.com/get-npm)
+- [Ruby](https://www.ruby-lang.org/en/downloads/)
```
justinvp: Do we still have a dependency on Ruby?

README.md
```diff
@@ -73,6 +114,14 @@ The following repos must be peers of `docs`, should be checked out to an appropr
 - `pulumi-kubernetes`
 - etc.
 
+#### Make errors
+
+`make build` errors are often due to missing dependencies. The following npm packages need to be installed separately on Ubuntu 19.04:
```
justinvp: Running `make ensure` (i.e. `yarn install`) won't install these on Linux?

README.md
chrsmith: Not to my knowledge. Probably safe to remove.

README.md
chrsmith: For more background, we haven't been super-consistent, but WAAAAY back when we wanted to have the same setup steps for reach repo:
https://github.com/pulumi/home#building-and-testing

`make`, `make ensure`, and `make install`.

README.md
bermudezmt: Unfortunately I failed to add why we had to install Ruby in my notes. @metral do you remember what necessitated the Ruby installation?

justinvp: We used to use Jekyll for the docs site, which uses Ruby. Now that we’ve moved to Hugo, I don’t think it’s necessary anymore.

README.md
bermudezmt: @justinvp Unfortunately not. I removed one of the listed dependencies then ran `make ensure`, and got the "package not found" error.

README.md
metral: IIRC we only installed ruby, and the pkgs in [1] because running `make build` kept yelling at us after having run `make ensure`. This could be tied to our package.json not being completely updated, but i have not dived into this further - we should probably bake and clean this up if we haven't

1 - https://github.com/pulumi/docs/pull/1310/files#diff-04c6e90faac2675aa89e2176d2eec7d8R121

README.md
justinvp: What was the specific error that you got? Can you paste the command line output?

README.md
bermudezmt: Sorry just now circling back to this: I uninstalled Ruby, and `make build` still worked, yet running `make ensure` resulted in the following error:

```
make ensure
yarn install
/usr/bin/env: ‘ruby’: No such file or directory
make: *** [Makefile:19: ensure] Error 127
```

README.md
bermudezmt: Ruby is required to run `make ensure`, but `make ensure` takes care of the dependencies, so I've removed that section.

PR #1313, opened 2019-07-10
Rephrase AWS CLI requirement for getting started

bermudezmt:


content/docs/quickstart/aws/configure.md
```diff
@@ -11,7 +11,7 @@ menu:
 
 <a href="{{< relref "/docs/reference/clouds/aws/setup.md" >}}" target="_blank">Configure AWS</a> so the Pulumi CLI can connect to AWS.
 
-> If your <a href="https://aws.amazon.com/cli/" target="_blank">AWS CLI</a>, `aws`, is already working, you should be good to go as Pulumi respects its configuration.
+> If you have `aws` or the <a href="https://aws.amazon.com/cli/" target="_blank">AWS CLI</a> already working, you should be good to go as Pulumi respects its configuration.
```
bermudezmt: How does Pulumi handle multiple AWS profiles? I tested this and it looks like the resources get created for the `default` profile. Is there a way to override that and specify a different named profile?

lukehoban: Note: I believe that the text was technically more correct before this edit - but definitely the sentence could be structured so this reads easier.  In this case: `aws` *is* "the AWS CLI" (it's the name of the CLI binary). 

> How does Pulumi handle multiple AWS profiles? 

You can set the `AWS_PROFILE` environment variable, or use `pulumi config set aws:profile <profilename>` to use  a different profile.  There are a *lot* of configuration options available on AWS.  We've documented them (not super well yet) across https://www.pulumi.com/docs/reference/clouds/aws/#configuration and https://www.pulumi.com/docs/reference/clouds/aws/setup/.

bermudezmt: Thank you, @lukehoban . I realized the note change wasn't the best after Eric commented. Having been bitten in the past with errors over having multiple AWS profiles configured, I thought it might be important to include your response in that note. Let me know what you think.

bermudezmt:
Merging with follow-up work on taking the important bits from the Setup page and updating the styling on notes.

PR #1315, opened 2019-07-10
updated alb ingress featured image

d-nishi:
removed need summary and added an image.


d-nishi:
@lukehoban @aydrian linked to [customer issue](https://github.com/pulumi/pulumi-eks/issues/185)

PR #1318, opened 2019-07-10
TLS Provider document generation

stack72:


config/_default.toml
```diff
@@ -240,6 +240,11 @@ enableRobotsTXT = true
     name = "terraform"
     url = "/docs/reference/pkg/nodejs/pulumi/terraform/"
 
+[[menu.reference]]
+    parent = "node"
+    name = "tls"
+    url = "/docs/reference/pkg/nodejs/pulumi/tls/"
```
justinvp: @chrsmith, why do we now have to duplicate these menu entries in `config/_default.toml` and `config/_default/menu.toml`?

In any case, I want to get rid of the need to have these entries in these files, and fix the API docs codegen to include the menu info in generated page's front matter, so we don't have to maintain these anymore. Opened #1323.

chrsmith: > why do we now have to duplicate these menu entries in config/_default.toml and config/_default/menu.toml?

This was an oversight on my part, so please feel free to fix/cleanup the `config/` folder as you see fit.

stack72:
@justinvp / @chrsmith I am going to merge this now so I can check it out on staging - we can do something with that duplicate config.toml before production

PR #1334, opened 2019-07-12
Fix sort ordering on the About page

### LINKED ISSUE #1330: Randomize team members on the About page

cnunciato:
* Sort team members randomly with JavaScript.
* Honor the `weight` property for board members.

Fixes #1330.

### DIFF
--- a/layouts/page/about.html
+++ b/layouts/page/about.html
@@ -37,7 +37,8 @@
 
     <section id="team" class="px-4 md:px-0">
         <div class="container md:mx-auto max-w-5xl text-center">
-            <ul class="inline-flex flex-wrap list-none justify-center p-0">
+            <!-- Hide the list initially, in order to avoid any visual flicker when shuffled. -->
+            <ul class="invisible inline-flex flex-wrap list-none justify-center p-0" data-shuffle="true">
                 {{ range $.Site.Data.team.team }}
                     {{ if eq .status "active" }}
                         <li class="text-center w-48 mx-4 mb-8">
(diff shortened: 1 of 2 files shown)


layouts/page/about.html
```diff
@@ -37,7 +37,8 @@ <h1>Program the cloud.</h1>
 
     <section id="team" class="px-4 md:px-0">
         <div class="container md:mx-auto max-w-5xl text-center">
-            <ul class="inline-flex flex-wrap list-none justify-center p-0">
+            <!-- Hide the list initially, in order to avoid any visual flicker when shuffled. -->
+            <ul class="invisible inline-flex flex-wrap list-none justify-center p-0" data-shuffle="true">
                 {{ range $.Site.Data.team.team }}
```
chrsmith: Can we not just range over the team data sorted by "weight" as well? My understanding was that is how we achieved the "randomized layout" before. Since every datum had the same weight value of one, it was the ~random ordering in which the underlying data structure was initialized. Does that no longer work with how we are iterating through the team elements?

cnunciato: The [previous code](https://github.com/pulumi/www.pulumi.com-old/blob/ec16345b673a5b90a75b8a7aa262f2a82a96df1e/src/js/source/hero.js#L60-L72) behaved similarly to the code in this PR (it removed, sorted and reattached the list items; I'm just doing something similar with CSS), but I do slightly prefer the `shuffle` approach Mikhail suggests in the issue. 

So if we're are okay with these changing a few times a week rather that of on every page refresh (which I am), then I'd go with that instead. I've pushed a commit that shows the difference. One thing to note, though, is that since `shuffle` is apparently based on a random seed set once per process, reloading this page won't show a sort difference. But subsequent restarts/builds will.

chrsmith: > So if we're okay with these changing a few times a week rather than on every page load ... then I'd go with that instead

Yeah, let's go with that. While the CSS approach is fairly straight-forward, the `shuffle` approach seems a little simpler IMHO.

joeduffy: I'd prefer the re-sort on refresh approach.

chrsmith: Thanks for chiming in @joeduffy ! As owner for all of our website properties, I was debating whether or not we should schedule an emergency meeting to evaluate the merits the sort ordering on the About page.

After evaluating the various technical tradeoffs for the issue, as well as the UX implications, we were about to make the wrong call. Good thing Christian already spend the time implementing a the "re-sort on refresh" approach, so we can just go back and merge that commit. Sure, it's a little more JavaScript code in our website to maintain, load and execute on startup. But for the legions of people who load a company's About page, and then immediately refresh their browser, they will be so impressed with our attention to detail!

I'm glad we are spending our cycles on this PR than, say, the 85 other open issues in this repository. Or the 336 open issues in the `pulumi/pulumi` repo. All of those can definitely wait until we get the layout on the About page to be randomized every time you refresh the browser.

PR #1335, opened 2019-07-12
WIP: Animated code samples

cnunciato:
I've seen a few sites that do this, and it's nice -- could add some life to featured code blocks and CLI output snippets (like those in hero blocks, for example):

Do folks like this? Or no? You'd just add a `"typing" true` to a given sample, and it'd animate.

### DIFF
--- a/assets/js/typing.js
+++ b/assets/js/typing.js
@@ -1,0 +1,38 @@
+(function($) {
+    var spans = $(".typing span");
+    var offset = 500 /* ms */;
+    var delay = 150 /* ms */;
+
+    // Wrap every character in a span to make it individually selectable.
+    spans.each(function(i, span) {
+        var chars = span.textContent.split("");
+
+        $(span).html(chars.map(function(char) {
+            return "<span class='char'>" + char + "</span>";
+        }));
+    });
+
+    // Create an element to represent the cursor.
+    var cursor = $("<span class='cursor'></span>");
+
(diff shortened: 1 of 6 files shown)


chrsmith:
I don't think we should add this capability right now. But I can see us wanting to add it in the future.

I'd personally start at "-100 points" for something like this. If I am looking at a code block, I don't want to wait N seconds to see that code block actually get populated. So I don't see this being a positive change for the common case.

However, if as part of an improved www.pulumi.com landing page, where we highlight key product features, having an animated "typing" code block would actually be a great thing to add. i.e. if you scroll past "predicable, safe deployments" and the next heading is "use a real programming language", the thing being emphasized isn't the content of that code block, but the code itself. (And typing is a great way to highlight that.)

So I'd be inclined to just close out this PR, but keep the code around in your back pocket. Because as soon as we have a specific use-case (e.g. a specific landing page in-mind) I'd be onboard for adding it then.

cnunciato:
Yeah. This is really only appropriate for those situations where you want to communicate and call attention to the _concept_ of code, more than the actual full content of a code block -- like on a home page or a landing page. I was thinking of it specifically as a replacement for the static images in the  home page carousel, since we didn't seem ready to go with a video there yet.

cnunciato:
Closing this for now, but I'll leave the branch around for a bit.

PR #1341, opened 2019-07-14
Enable CI/CD for PRs from forks

chrsmith:
Tweak the Travis CI/CD configuration so that it won't fail when running on forks of the repo. This is because Travis will run the PRs, but not provide any secret configuration values (as to keep them, well, secret).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_install:
 # Download and install required tools.
 # Limit to one push build per branch.
 - go get github.com/pulumi/travisqueue
-- travisqueue start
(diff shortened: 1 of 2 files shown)


cnunciato:
What do you think about maybe tagging at least one reviewer on PRs before merging them (excepting harmless things like typos, of course)? Just to raise awareness?

chrsmith:
My plan is to enable branch protections to require a code review approval for all PRs, as well as a passing CI/CD run. (That last part however is blocked by this commit.) So yes, we 100% should be doing that. Please be merciless the next time you see me merging code without a reviewer.

PR #1363, opened 2019-07-15
Add "copy to clipboard" button to code snippets

### LINKED ISSUE #1023: "Copy to clipboard" button for code snippets

justinvp:
This PR adds a "Copy to Clipboard" button to code snippets.

copybutton

Right now it adds it to any snippet that has a language specified, which covers the vast majority of uses throughout the site. There are cases that specify code snippets without a language, and right now those won't have a copy button. And I'm deliberately disabling it for the code snippets used for marketing content. We should prob follow-up and enable it across the board, and provide a first-class way to opt-out of the copy button when it isn't desired.

Copying `bash`, `shell`, and `sh`, snippets has some extra functionality. `$` prefixes are stripped, comments and empty lines are stripped, multiple lines are collapsed down to a single line, and any "output" lines that aren't prefixed with `$` are stripped.

Examples:

This:

```bash
$ mkdir hello-fargate && cd hello-fargate
$ pulumi new aws-typescript
```

...will have the this written to the clipboard:

```
mkdir hello-fargate && cd hello-fargate && pulumi new aws-typescript
```

And this:

```bash
$ curl http://$(pulumi stack output url)
<html>
    <head>
        <title>Hello Fargate</title>
    </head>
    <body>
        <p>Hello, containers!</p>
        <p>Made with ❤️ with <a href="https://pulumi.com">Pulumi</a></p>
    </body>
</html>
```

...will have this written to the clipboard:

```
curl http://$(pulumi stack output url)
```

Fixes #1023
Replaces #1204

### DIFF
--- a/assets/sass/_fonts.scss
+++ b/assets/sass/_fonts.scss
@@ -1,3 +1,2 @@
 @import url('https://fonts.googleapis.com/css?family=Open+Sans:400,400i,400,700|Ubuntu:400,700');
 @import url('https://fonts.googleapis.com/css?family=PT+Mono');
-@import url('https://fonts.googleapis.com/css?family=Material+Icons');
--- a/assets/sass/_fonts.scss
+++ b/assets/sass/_fonts.scss
@@ -1,2 +1,3 @@
 @import url('https://fonts.googleapis.com/css?family=Open+Sans:400,400i,400,700|Ubuntu:400,700');
 @import url('https://fonts.googleapis.com/css?family=PT+Mono');
+@import url('https://fonts.googleapis.com/css?family=Material+Icons');
--- a/content/docs/reference/tutorials/kubernetes/tutorial-p8s-rollout.md
+++ b/content/docs/reference/tutorials/kubernetes/tutorial-p8s-rollout.md
@@ -55,16 +55,16 @@ can see `canary-example-app` and `canary-staging-app` created.
 
 1. Now, install dependencies:
 
-   ```sh
-   npm install
-   ```
+    ```sh
+    $ npm install
+    ```
 
 1. Create a new stack:
 
-   ```sh
-   $ pulumi stack init
-   Enter a stack name: staged-rollout
-   ```
+    ```sh
+    $ pulumi stack init
+    Enter a stack name: staged-rollout
+    ```
 
 1. **IMPORTANT NOTE:** The code in `index.ts` is meant to be run out-of-cluster (_e.g._, on your
    local machine). It will thus call `kubectl port-forward` on your behalf so that the Prometheus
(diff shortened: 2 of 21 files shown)


assets/sass/_fonts.scss
```diff
@@ -1,2 +1,3 @@
 @import url('https://fonts.googleapis.com/css?family=Open+Sans:400,400i,400,700|Ubuntu:400,700');
 @import url('https://fonts.googleapis.com/css?family=PT+Mono');
+@import url('https://fonts.googleapis.com/css?family=Material+Icons');
```
justinvp: I wanted to use the same material "copy" button we use on https://app.pulumi.com. As a follow-up, would anyone be opposed to removing our use of FontAwesome and just using the Material icons instead?

layouts/partials/code.html
```diff
@@ -3,7 +3,7 @@
     {{ $lang = "cql" }}
 {{ end }}
 
-<div class="code-mode-{{ default "light" .mode }} {{ if .chrome }} code-chrome {{ end }} code-{{ .lang }}">
+<div class="code-hero code-mode-{{ default "light" .mode }} {{ if .chrome }} code-chrome {{ end }} code-{{ .lang }}">
```
justinvp: I added `code-hero` so I could use that in a selector to exclude the copy button from these. Though, I could see wanting to provide it for some uses. As a follow-up, we could add another option to enable it for this partial. (Would require some minor style updates for dark mode).

content/docs/reference/tutorials/kubernetes/tutorial-p8s-rollout.md
```diff
@@ -73,54 +73,52 @@ can see `canary-example-app` and `canary-staging-app` created.
...
-     + p90ResponseTime: "8221.236"
-
-   info: 2 changes performed:
-       + 2 resources created
-         26 resources unchanged
-   Update duration: 7.362744393s
-
-   Permalink: https://app.pulumi.com/hausdorff/test-p8s/updates/1
-   ```
+        $ pulumi up
```
justinvp: Workaround a bug in Hugo with code fences in lists. (As an aside, I think we can actually work around the bug by using `{{< highlight >}}` in the few places where the bug shows up, allowing us to upgrade to the latest version of Hugo. Will be looking at that separately.)

content/docs/reference/tutorials/kubernetes/tutorial-p8s-rollout.md
```diff
@@ -55,16 +55,16 @@ can see `canary-example-app` and `canary-staging-app` created.
 
 1. Now, install dependencies:
 
-   ```sh
-   npm install
-   ```
+    ```sh
+    $ npm install
```
cnunciato: A few of the blocks on this page don't render the copy buttons, apparently because they're parented by an `li` and not a `div`. Changing the query to `:not(.code-hero) > div.highlight` seems to work, though, and _should_ be safe, I think.

layouts/partials/code.html
cnunciato: Having `hero` in the classname bugs me a teeny bit (since there are lots of non-hero places where these are used), but it's only here, so definitely a nit. Maybe `.no-copy` (like `.no-select`), though? Up to you, just thinking aloud.

And yes, it'd be great to have these under Products as well! Could just surface them on hover, too.

assets/sass/_fonts.scss
cnunciato: Personally I’m open to having a look at how using Material icons affects the site overall, if that means being able to settle on a single icon set — but more generally, I’m honestly not in favor of propagating concrete Material design elements (as opposed to just guidance, which is fine) outside of the console.

justinvp:
@cnunciato, I addressed the feedback and made a few more tweaks. Mind taking another look?

Here's what it looks like using FontAwesome's `copy` icon (I got rid of the green checkmark after the click and went with just a tooltip to keep it simple):

copy-fa1

copy-fa-dark

Alternatively, we could use the `clipboard` icon, but I think I prefer `copy`.

Screen Shot 2019-07-16 at 2 52 46 PM

cnunciato:
I like it! 👍

PR #1365, opened 2019-07-16
Correct broken link

d-nishi:
fixed broken link


d-nishi:
Merged per discussion with @chrsmith

Issue #1336, opened 2019-07-12
Add note on google_storage_bucket_iam_binding


bermudezmt:
Actually, `allUsers` is a [special identifier](https://cloud.google.com/storage/docs/json_api/v1/buckets/setIamPolicy) and does not need any prefix like `group:`.
 ```
const binding = new gcp.storage.BucketIAMBinding("member", {
  bucket: config.bucketName,
  members: ["allUsers"],
  role: "roles/storage.objectViewer",
});
```

Maybe you could consider adding it to the documentation, just [here](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/gcp/storage/#BucketIAMBinding).

I'm closing this issue myself

_Originally posted by @sebght in https://github.com/pulumi/pulumi-gcp/issues/177#issuecomment-510928342_

justinvp:
Details on the `members` property of `BucketIAMBindingArgs` is missing from https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/gcp/storage/#BucketIAMBindingArgs

See https://www.terraform.io/docs/providers/google/r/storage_bucket_iam.html

The docs should include the following:

<img width="843" alt="Screen Shot 2019-07-18 at 9 24 51 AM" src="https://user-images.githubusercontent.com/710598/61474633-e6b17700-a93d-11e9-8ec8-2aa0a5a10216.png">

lukehoban:
Fixed in https://github.com/pulumi/terraform-provider-google-beta/commit/54c1f8e3a8f517ccf08b35a931ae68d0b8d6593f.

PR #1368, opened 2019-07-16
updated gcp storage iam doc

sebght:
I added some missing documentation concerning IAM members, as described in #1336 
I basically used Terraform's doc, which I found was really good


justinvp:
Hi @sebght! Thanks so much for the PR! Unfortunately, this is a generated page, so any change here will just be overwritten when the API docs are regenerated.

Which Terraform doc are you using as reference? These pages are actually generated from the Terraform provider's docs, so it'd be interesting to drill into why these parts aren't being included automatically as part of that.

sebght:
Hi @justinvp. I saw this difference between these two docs :
* https://www.terraform.io/docs/providers/google/r/storage_bucket_iam.html
* https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/gcp/storage/#BucketIAMBinding
This is why I thought there was some need to complete it, though I was wrong.

Do you think there is an issue in the doc generation ?

justinvp:
Thanks, @sebght. It does indeed look like this is a doc generation bug. As such, I'll close this and add more details to #1336. Thanks again for raising the issue!

Issue #1277, opened 2019-07-03
Website tweaks


chrsmith:
Collection of issues post GREAT WEBSITE MIGRATION OF 2019

- [x] We serve a few `.webp` images for blog posts, but it looks like the image format isn't supported on Safari. Example: see the "Running Containers in AWS..." feature image on https://www.pulumi.com/blog/tag/containers/.
- [x] Author avatar images get stretched in odd ways on mobile. Navigate to https://www.pulumi.com/blog/author/chris-smith/ and make the window super-duper teeny-tiny. We should set a min-size of the image and/or constrain how it shrinks?
- [x] Do we still have the short links for `/help/...`? We still have some URLs baked into the `pulumi/pulumi` code that reference things like https://pulumi.io/help/stack-reference. Is there an updated version of that?
- [ ] `h6` seems to be formatter as ALLCAPS, even when styling an inner `code` element?  (see https://www.pulumi.com/docs/reference/programming-model/#dependson)

  <img width="808" alt="Screen Shot 2019-07-10 at 10 20 48 PM" src="https://user-images.githubusercontent.com/710598/61023999-07d3ff80-a361-11e9-9dc8-ac3835fd1a8c.png">

  These are `<h6><code>dependsOn</code></h6>`

- [x] The highlighting of currently selected section in the nav is barely noticeable

- [x] We seem to be generating an rss.xml file for all sections; is there a way to make it only generate for /blog/rss.xml? (#1420)

justinvp:
> * Do we still have the short links for `/help/...`?

We should discuss what we want to do with these. See https://github.com/pulumi/pulumi-service/pull/3716#discussion_r300465434

justinvp:
I checked off the "short links" bullet above per https://github.com/pulumi/pulumi-service/pull/3716#issuecomment-509262571

justinvp:
I also added a couple bullets from #1213 (I'm going to close #1213)

PR #1376, opened 2019-07-17
Add descriptions for new resource options

bermudezmt:
- Presented the options alphabetically (not yet programmatic)
- Add descriptions for `aliases`, `import`, and `CustomTimeouts` resource options

Hey @cnunciato - since these resource options are supposed to be camelcased, I'm curious to hear your take on whether we should keep these formatted as H6 headings with a specific CSS class or do something else. 

### DIFF
--- a/content/docs/reference/programming-model.md
+++ b/content/docs/reference/programming-model.md
@@ -115,26 +115,35 @@ The `args` provided to a resource determine what inputs will be used to initiali
 
 All resource constructors also accept an `options` argument which can provide the following additional resource options controlling how the resource will be managed by Pulumi.
 
-###### `dependsOn`
-Provides a list of explicit resource dependencies to add to the resource. Every resource referenced either directly or indirectly by an `Output` that is passed in to the resource constructor will implicitly be included, so this additional information is only needed when the dependency is on something that is not already an input to the resource. The default is `[]`.
+###### `additionalSecretOutputs`
+Provides a list of output properties which should be treated as secrets. This value augments any values that Pulumi detects itself, based on what secret inputs to the resource has. This is typically used to express that for a specific instance of a resource, some of its output properties should be treated as secrets (when they would not normally be).
 
-###### `protect`
-Marks a resource as protected. A protected resource cannot be deleted directly: first you must set `protect: false` and run `pulumi up`. Then, the resource can be deleted, either by removing the line of code or by running `pulumi destroy`.  The default is to inherit this value from the parent resource, and `false` for resources without a parent.
+###### `aliases`
+Lets you rename a resource. When making a breaking change to the name or type of a resource or component, you can add the old name to the list of `aliases` for a resource to ensure that existing resources will be migrated to the new name instead of being deleted and replaced with the new named resource.
 
-###### `parent`
-A parent for the resource. See [Components](#components).  The default is to parent to the implicitly-created `Stack` resource that is a root resource for all Pulumi stacks.
-
-###### `provider`
-A provider for the resource. See [Providers](#providers).  The default is to inherit this value from the parent resource, and to use the ambient provider specified by Pulumi configuration for resources without a parent.
+###### `customTimeouts`
+Lets you specify a request timeout in string format when creating, updating, or deleting a resource. For example, `customTimeOuts: { create: "1m" }`.
 
 ###### `deleteBeforeReplace`
-Specify that replacements of the resource will delete the existing resource before creating its replacement.  This will lead to downtime during the replacement, but may be necessary for some resources that manage scarce resources behind the scenes.  The default is `false`.
+Set this option to `true` to specify that replacements of the resource will delete the existing resource before creating its replacement.  This will lead to downtime during the replacement, but may be necessary for some resources that manage scarce resources behind the scenes.  The default is `false`.
+
+###### `dependsOn`
+Provides a list of explicit resource dependencies to add to the resource. Every resource referenced either directly or indirectly by an `Output` that is passed in to the resource constructor will implicitly be included, so this additional information is only needed when the dependency is on something that is not already an input to the resource. The default is `[]`.
 
 ###### `ignoreChanges`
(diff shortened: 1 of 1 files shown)


content/docs/reference/programming-model.md
```diff
@@ -115,27 +115,36 @@ The `args` provided to a resource determine what inputs will be used to initiali
...
+
 ###### `dependsOn`
 Provides a list of explicit resource dependencies to add to the resource. Every resource referenced either directly or indirectly by an `Output` that is passed in to the resource constructor will implicitly be included, so this additional information is only needed when the dependency is on something that is not already an input to the resource. The default is `[]`.
 
-###### `protect`
-Marks a resource as protected. A protected resource cannot be deleted directly: first you must set `protect: false` and run `pulumi up`. Then, the resource can be deleted, either by removing the line of code or by running `pulumi destroy`.  The default is to inherit this value from the parent resource, and `false` for resources without a parent.
+###### `ignoreChanges`
+Provides a list of properties which will be ignored as part of updates. The value of the property will be used for newly created resources, but will not be used as part of updates. This is typically used to avoid changes in properties leading to diffs or to change defaults for a property without forcing all existing deployed stacks to update or replace the affected resource.
+
+###### `import`
```
bermudezmt: I specifically changed to `instantiating a Pulumi resource`, but not 100% sure that's accurate. I also removed the `i.e` sections from the change log, but want to make sure this description is still accurate. Thoughts, @pgavlin ?

Also, I was wondering if [this comment](https://github.com/pulumi/pulumi/pull/2893#issuecomment-511468099) needs to be called out.

content/docs/reference/programming-model.md
```diff
@@ -115,27 +115,36 @@ The `args` provided to a resource determine what inputs will be used to initiali
 
 All resource constructors also accept an `options` argument which can provide the following additional resource options controlling how the resource will be managed by Pulumi.
 
+###### `additionalSecretOutputs`
+Provides a list of output properties which should be treated as secrets. This value augments any values that Pulumi detects itself, based on what secret inputs to the resource has. This is typically used to express that for a specific instance of a resource, some of its output properties should be treated as secrets (when they would not normally be).
+
+###### `aliases`
+Lets you rename a resource. When making a breaking change to the name or type of a resource or component, you can add the old name to the list of `aliases` for a resource to ensure that existing resources will be migrated to the new name instead of being deleted and replaced with the new named resource.
+
+###### `customTimeouts`
```
bermudezmt: @stack72 I added an example here, but none of the other resource options have them.

cnunciato:
> I'm curious to hear your take on whether we should keep these formatted as h6

@bermudezmt Ah, yes -- [the CSS here](https://github.com/pulumi/docs/blob/317e6f16767be86e8f35aae371ab4db0ba21a7f0/assets/sass/styles.scss#L58-L60) is too specific. It's captured as an item in #1277, which I'll be picking up shortly, so no need to do anything here for it.

content/docs/reference/programming-model.md
pgavlin: Here's what I had written locally:

> Provides a set of custom timeouts for a resource's create, update, and delete operations. These timeouts are specified as a string like "5m", "40s", or "1d" (5 minutes, 40 seconds, or 1 day, respectively).

I do believe that the formatting examples are particularly important here, though they are also present in the language SDKs.

content/docs/reference/programming-model.md
pgavlin: Here's what I wrote:

> The ID of an existing resource to import for management using Pulumi. If this value is set, Pulumi will read the current state of the resource with the given ID from the backing provider (e.g. AWS, Azure, GCP, Kubernetes). The inputs to the resource's constructor must not differ from this state or the import will fail. Once a resource has been imported, this property should be unset.

content/docs/reference/programming-model.md
bermudezmt: Thanks, @pgavlin . If these are options that a user would set or pass in, I do wonder about the usage of "Provides a ...". Please educate me if I am completely off mark. For now, I think we should try to be consistent so I updated the `aliases` description as well to follow the current description pattern.

content/docs/reference/programming-model.md
```diff
@@ -115,30 +115,39 @@ The `args` provided to a resource determine what inputs will be used to initiali
 
 All resource constructors also accept an `options` argument which can provide the following additional resource options controlling how the resource will be managed by Pulumi.
 
+###### `additionalSecretOutputs`
+Provides a list of output properties which should be treated as secrets. This value augments any values that Pulumi detects itself, based on what secret inputs to the resource has. This is typically used to express that for a specific instance of a resource, some of its output properties should be treated as secrets (when they would not normally be).
+
+###### `aliases`
+Provides a list of aliases for a resource or component. When making a breaking change to the name or type of a resource or component, you can add the old name to the list of `aliases` for a resource to ensure that existing resources will be migrated to the new name instead of being deleted and replaced with the new named resource.
```
lukehoban: We'll want to expand this and others much more in the near future.  But it may help for now to just add one additional sentence:

> For example, a resource can be aliased to a full previous resource URN (e.g `aliases: ["urn:pulumi:stackname::projectname::aws:s3/bucket:Bucket::app-function"]`) or to a relative change to the resource's `name`, `parent` and/or `type` (e.g. `aliases: [{ name: "otherchild", parent: this }]`).

content/docs/reference/programming-model.md
```diff
@@ -115,30 +115,39 @@ The `args` provided to a resource determine what inputs will be used to initiali
...
+
+###### `customTimeouts`
+Provides a set of custom timeouts for create, update, and delete operations on a resource. These timeouts can be specified as a string like "5m", "40s", or "1d" (5 minutes, 40 seconds, or 1 day, respectively).
```
lukehoban: Totally minor - but we may want to use code-style for `create`, `update` and `delete` in this sentence to highlight that these are the keys that can be set - and/or add a final sentence:

> For example, `customTimeouts: { create: "1m" }`.

bermudezmt:
Thank you so much for reviewing, @pgavlin and @lukehoban . I'm merging so we can get these docs updates out. cc: @clstokes

PR #1377, opened 2019-07-17
Clean up Hugo config

justinvp:
- Delete redundant/unnecessary `config/_default.toml` file
- Rename `menu.toml` to `menus.toml` (either works, but `menus` plural is more idiomatic)
- Use `HUGO_ENVIRONMENT` variable, which Hugo will use in place of the `--environment` flag, and use `development` as the default value (as that's the default used when running `hugo server`)
- Specify `baseURL` in `config/production/config.toml` and `config/staging/config.toml`
- Delete references to `fusion` as the branch and stack have been deleted


config/staging/config.toml
```diff
@@ -0,0 +1 @@
+baseURL = "https://www-staging.pulumi.com/"
```
justinvp: I realize staging isn't long for this world, but we can delete this file when we make `master` deploy to `production`. For now, this maintains the status-quo.

justinvp:
> let's get on the same page before you submit.

@chrsmith, Keeping the staging stack around for the time being sounds fine to me. After a while, if we find it's not of use, we can get rid of it.

I'll merge this PR as-is, and follow-up with a subsequent PR that makes the changes for "`master` branch is now `production`" and "`staging` branch is now `staging`" as you've described.

PR #1378, opened 2019-07-17
Provide friendlier error message when `node_modules` doesn't exist

justinvp:
Our use of PostCSS in Hugo requires the `postcss-cli` npm package to be installed. If it isn't, Hugo will print an error:

> PostCSS not found; install with "npm install postcss-cli". See https://gohugo.io/hugo-pipes/postcss/

However, this isn't ideal for us, as we use `yarn` in this repo, and don't use it directly. Instead, `make ensure` should be used.

My recent change to add a copy button to code snippets also depends on an npm package (`clipboard-polyfill`) to be installed. If it isn't, Hugo fails with several error messages about not being able to read the file, and it isn't immediately obvious how to address the issue.

This change makes it very clear how to address the issue. If the relevant directories do not exist in `node_modules`, we fail the Hugo build and print an error that says "Run `make ensure`".

---

Before:

```
$ make serve
-e SERVE:
hugo server --buildDrafts --environment default
Building sites … ERROR 2019/07/16 21:59:36 error: failed to transform resource: POSTCSS: failed to transform "css/styles.css" (text/css): PostCSS not found; install with "npm install postcss-cli". See https://gohugo.io/hugo-pipes/postcss/
ERROR 2019/07/16 21:59:40 render of "section" failed: execute of template failed: template: docs/list.html:16:11: executing "docs/list.html" at <partial "scripts.html" .>: error calling partial: "/Users/justin/go/src/github.com/pulumi/docs/layouts/partials/scripts.html:14:21": execute of template failed: template: partials/scripts.html:14:21: executing "partials/scripts.html" at <readFile "node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js">: error calling readFile: stat /Users/justin/go/src/github.com/pulumi/docs/content/node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js: no such file or directory
ERROR 2019/07/16 21:59:40 render of "page" failed: execute of template failed: template: docs/single.html:16:11: executing "docs/single.html" at <partial "scripts.html" .>: error calling partial: "/Users/justin/go/src/github.com/pulumi/docs/layouts/partials/scripts.html:14:21": execute of template failed: template: partials/scripts.html:14:21: executing "partials/scripts.html" at <readFile "node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js">: error calling readFile: stat /Users/justin/go/src/github.com/pulumi/docs/content/node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js: no such file or directory
ERROR 2019/07/16 21:59:40 render of "page" failed: execute of template failed: template: docs/single.html:16:11: executing "docs/single.html" at <partial "scripts.html" .>: error calling partial: "/Users/justin/go/src/github.com/pulumi/docs/layouts/partials/scripts.html:14:21": execute of template failed: template: partials/scripts.html:14:21: executing "partials/scripts.html" at <readFile "node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js">: error calling readFile: stat /Users/justin/go/src/github.com/pulumi/docs/content/node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js: no such file or directory
ERROR 2019/07/16 21:59:40 render of "section" failed: execute of template failed: template: docs/list.html:16:11: executing "docs/list.html" at <partial "scripts.html" .>: error calling partial: "/Users/justin/go/src/github.com/pulumi/docs/layouts/partials/scripts.html:14:21": execute of template failed: template: partials/scripts.html:14:21: executing "partials/scripts.html" at <readFile "node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js">: error calling readFile: stat /Users/justin/go/src/github.com/pulumi/docs/content/node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js: no such file or directory
Total in 4628 ms
Error: Error building site: failed to render pages: render of "section" failed: execute of template failed: template: docs/list.html:16:11: executing "docs/list.html" at <partial "scripts.html" .>: error calling partial: "/Users/justin/go/src/github.com/pulumi/docs/layouts/partials/scripts.html:14:21": execute of template failed: template: partials/scripts.html:14:21: executing "partials/scripts.html" at <readFile "node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js">: error calling readFile: stat /Users/justin/go/src/github.com/pulumi/docs/content/node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js: no such file or directory
make: *** [serve] Error 255
```

After:

```
$ make serve
-e SERVE:
hugo server --buildDrafts --environment default
Building sites … ERROR 2019/07/16 23:12:53 "node_modules/postcss-cli" does not exist. Run `make ensure` to install required dependencies.
ERROR 2019/07/16 23:12:53 "node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js" does not exist. Run `make ensure` to install required dependencies.
Total in 10711 ms
Error: Error building site: logged 2 error(s)
make: *** [serve] Error 255
```

### DIFF
--- a/layouts/partials/scripts.html
+++ b/layouts/partials/scripts.html
@@ -11,7 +11,15 @@
 {{ $s = $s | append (resources.Get "js/carousel.js") }}
 {{ $s = $s | append (resources.Get "js/langchoose.js") }}
 {{ $s = $s | append (resources.Get "js/noselect.js") }}
-{{ $s = $s | append (readFile "node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js" | resources.FromString "js/node/clipboard-polyfill.js") }}
+
+<!-- We use the `clipboard-polyfill` npm package. If it hasn't been installed, show a friendlier error message. -->
+{{ $clipboardFile := "node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js" }}
+{{ if fileExists $clipboardFile }}
+    {{ $s = $s | append (readFile $clipboardFile | resources.FromString "js/node/clipboard-polyfill.js") }}
+{{ else }}
+    {{ errorf "%q does not exist. Run `make ensure` to install required dependencies." $clipboardFile }}
+{{ end }}
+
 {{ $s = $s | append (resources.Get "js/copybutton.js") }}
 {{ $js := $s | resources.Concat "js/bundle.js" | resources.Minify | resources.Fingerprint }}
 <script src="{{ $js.RelPermalink }}" integrity="{{ $js.Data.Integrity }}"></script>
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -71,13 +71,19 @@
     <link rel="alternate" type="application/rss+xml" href="{{ "/blog/rss.xml" | absURL }}" title="Pulumi Blog" />
 
     <!--
-        Build Sass. Generated file will contain content hash in filename,
-        so no need to add query param to cache bust.
+        Build Sass. Generated file will contain content hash in filename, so no need to add query param to cache bust.
+        We check if `postcss-cli` has been installed, which is required by Hugo to use PostCSS, to provide a more
+        relevant error message.
     -->
-    {{ $toCSS := (dict "targetPath" "css/styles.css" "outputStyle" "compressed" "enableSourceMap" true) }}
-    {{ $postCSS := (dict "config" "assets/config/postcss.config.js") }}
-    {{ $styles := resources.Get "sass/styles.scss" | resources.ToCSS $toCSS | resources.PostCSS $postCSS | resources.Fingerprint }}
-    <link rel="stylesheet" href="{{ $styles.Permalink }}" integrity="{{ $styles.Data.Integrity }}">
+    {{ $postCSSCLIFile := "node_modules/postcss-cli" }}
+    {{ if fileExists $postCSSCLIFile }}
+        {{ $toCSS := (dict "targetPath" "css/styles.css" "outputStyle" "compressed" "enableSourceMap" true) }}
+        {{ $postCSS := (dict "config" "assets/config/postcss.config.js") }}
+        {{ $styles := resources.Get "sass/styles.scss" | resources.ToCSS $toCSS | resources.PostCSS $postCSS | resources.Fingerprint }}
+        <link rel="stylesheet" href="{{ $styles.Permalink }}" integrity="{{ $styles.Data.Integrity }}">
+    {{ else }}
+        {{ errorf "%q does not exist. Run `make ensure` to install required dependencies." $postCSSCLIFile }}
+    {{ end }}
 
     <script async defer src="https://buttons.github.io/buttons.js"></script>
     {{- if eq hugo.Environment "production" }}


layouts/partials/scripts.html
```diff
@@ -6,12 +6,16 @@
...
-{{ $s := slice (resources.Get "js/main.js") }}
-{{ $s = $s | append (resources.Get "js/nav.js") }}
-{{ $s = $s | append (resources.Get "js/carousel.js") }}
-{{ $s = $s | append (resources.Get "js/langchoose.js") }}
-{{ $s = $s | append (resources.Get "js/noselect.js") }}
-{{ $s = $s | append (readFile "node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js" | resources.FromString "js/node/clipboard-polyfill.js") }}
-{{ $s = $s | append (resources.Get "js/copybutton.js") }}
-{{ $js := $s | resources.Concat "js/bundle.js" | resources.Minify | resources.Fingerprint }}
-<script src="{{ $js.RelPermalink }}" integrity="{{ $js.Data.Integrity }}"></script>
+{{ if fileExists "node_modules" }}
```
justinvp: Actually, thinking about it more just now, I should check for the actual `clipboard-polyfill` file here, as it's possible for folks to have `node_modules` currently, but haven't run `make ensure` again to get the new `clipboard-polyfill` dependency. Will fix.

justinvp:
> But I don't like the fact we are duplicating the logic in two places.

I had the exact same thoughts, but when I had only a single check the build would still fail in one of the places. I think it has to do with the way Hugo parallelizes the build (multiple pages are generated in parallel). I didn't try putting it at the top of the first default template, though. I could look into it in more detail if you'd like.

chrsmith:
> I didn't try putting it at the top of the first default template

Yeah, I think that's the key thing. Since if you put that validation in some leaf node partial, it will only get rendered on-demand. Whereas `partials/head.html` should be brought in for nearly every page on the site.

So including that validation logic from `partials/head.html` (or `_layouts/default.html`?) would ensure it would get included nearly everywhere.

And as long as it errors building one page, it is OK if the build happens in parallel right? Worst case scenario you see the same build error more than once?

justinvp:
> since there is value in the "verify X exists before using X".

So if we do consolidate these checks all into one place (e.g. `partials/_validation.html` included at the bottom of `partials/head.html`), you'd be OK with some explicit checks for existence of `node_modules/postcss-cli` and `node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js` in there?

I don't love that we're then duplicating these file paths. I'm not worried so much about the postcss one as that's a Hugo dependency, but the the clipboard one makes me a little uneasy -- if we end up not needing to polyfill the `promise` part (e.g. we change to using `clipboard-polyfill.js` instead of `clipboard-polyfill.promise.js`), or decide to use a different clipboard library, then we need to remember to update the path in `partials/_validation.html` as well.

chrsmith:
@justinvp those issues are exactly what I meant by "since there is value in the "verify X exists before using X". I'll leave it up to you to decide what we want to do here. What you have in this PR is fine and will serve us well.

I didn't know that the logic for node modules would be nontrivial (e.g. regarding polyfills). But really all we care about is that `make ensure` has been run, right? So would it be generally sufficient to only verify that the `node_modules` _directory_ exists? And not confirm specific libraries/tools are there?

e.g. you'd get weird errors if your `node_modules` directory is out of date. But the more likely scenario is that it doesn't exist at all (and thus would direct you to `make ensure`)? Anyways, don't let me slow you down.

justinvp:
I'm going to merge this as-is. Even if I move a single check for just the `node_modules` dir into it's own partial that is included at the top of `partials/head.html`, it still shows the other errors before the build stops:

```
$ make serve
-e SERVE:
hugo server --buildDrafts --environment default
Building sites … ERROR 2019/07/17 20:11:16 "node_modules" does not exist. Please run `make ensure` to install required dependencies.
ERROR 2019/07/17 20:11:16 error: failed to transform resource: POSTCSS: failed to transform "css/styles.css" (text/css): PostCSS not found; install with "npm install postcss-cli". See https://gohugo.io/hugo-pipes/postcss/
ERROR 2019/07/17 20:11:20 render of "page" failed: execute of template failed: template: docs/single.html:16:11: executing "docs/single.html" at <partial "scripts.html" .>: error calling partial: "/Users/justin/go/src/github.com/pulumi/docs/layouts/partials/scripts.html:14:21": execute of template failed: template: partials/scripts.html:14:21: executing "partials/scripts.html" at <readFile "node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js">: error calling readFile: stat /Users/justin/go/src/github.com/pulumi/docs/content/node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js: no such file or directory
ERROR 2019/07/17 20:11:20 render of "section" failed: execute of template failed: template: docs/list.html:16:11: executing "docs/list.html" at <partial "scripts.html" .>: error calling partial: "/Users/justin/go/src/github.com/pulumi/docs/layouts/partials/scripts.html:14:21": execute of template failed: template: partials/scripts.html:14:21: executing "partials/scripts.html" at <readFile "node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js">: error calling readFile: stat /Users/justin/go/src/github.com/pulumi/docs/content/node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js: no such file or directory
ERROR 2019/07/17 20:11:20 render of "section" failed: execute of template failed: template: docs/list.html:16:11: executing "docs/list.html" at <partial "scripts.html" .>: error calling partial: "/Users/justin/go/src/github.com/pulumi/docs/layouts/partials/scripts.html:14:21": execute of template failed: template: partials/scripts.html:14:21: executing "partials/scripts.html" at <readFile "node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js">: error calling readFile: stat /Users/justin/go/src/github.com/pulumi/docs/content/node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js: no such file or directory
ERROR 2019/07/17 20:11:20 render of "page" failed: execute of template failed: template: docs/single.html:16:11: executing "docs/single.html" at <partial "scripts.html" .>: error calling partial: "/Users/justin/go/src/github.com/pulumi/docs/layouts/partials/scripts.html:14:21": execute of template failed: template: partials/scripts.html:14:21: executing "partials/scripts.html" at <readFile "node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js">: error calling readFile: stat /Users/justin/go/src/github.com/pulumi/docs/content/node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js: no such file or directory
Total in 4833 ms
Error: Error building site: failed to render pages: render of "page" failed: execute of template failed: template: docs/single.html:16:11: executing "docs/single.html" at <partial "scripts.html" .>: error calling partial: "/Users/justin/go/src/github.com/pulumi/docs/layouts/partials/scripts.html:14:21": execute of template failed: template: partials/scripts.html:14:21: executing "partials/scripts.html" at <readFile "node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js">: error calling readFile: stat /Users/justin/go/src/github.com/pulumi/docs/content/node_modules/clipboard-polyfill/build/clipboard-polyfill.promise.js: no such file or directory
make: *** [serve] Error 255
```

PR #1381, opened 2019-07-17
Updating code examples for Serializing JavaScript Functions

aydrian:
Updated code examples to use `@pulumi/aws` instead of now deprecated `@pulumi/aws-serverless`.

@CyrusNajmabadi: There are still a couple places in the **Pulumi Execution Order** section where the text should be updated since the function call to `serverless.function.createLambdaFunction` has been replaced with an instance of the `aws.lambda.CallbackFunction` class. Specifically, lines 169 and 198. Could you please suggest changes for these?

### DIFF
--- a/content/docs/reference/serializing-functions.md
+++ b/content/docs/reference/serializing-functions.md
@@ -195,7 +195,7 @@ const lambda = new aws.lambda.CallbackFunction("mylambda", {
 });
 ```
 
-When `pulumi` starts executing `createLambdaFunction` it will analyze the JavaScript function code and will see that it uses the 'obj' value. At that point in time it will use whatever the value is currently to serialize over. So, in the first example, it will serialize the value `{ a: 1, b: 2 }`, even though right after executing createLambdaFunction the program code will update that value to `{ a: 3, b: 4}`. In the second example, the code will see the `{ a: 3, b: 4 }` value and will serialize that into the 'run time' code.
+When `pulumi` starts executing `new aws.lambda.CallbackFunction` it will analyze the JavaScript function code and will see that it uses the 'obj' value. At that point in time it will use whatever the value is currently to serialize over. So, in the first example, it will serialize the value `{ a: 1, b: 2 }`, even though right after executing createLambdaFunction the program code will update that value to `{ a: 3, b: 4}`. In the second example, the code will see the `{ a: 3, b: 4 }` value and will serialize that into the 'run time' code.
 
 Notes:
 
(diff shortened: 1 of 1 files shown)


content/docs/reference/serializing-functions.md
```diff
@@ -6,254 +6,263 @@ The Pulumi Node.js SDK provides a core API for converting a JavaScript function
...
 
 ### Pulumi Execution Order
 
-`pulumi` uses `node` to execute a Pulumi application.  During execution, when a call to `createLambdaFunction` is encountered, the function is converted to a Lambda at that point in execution.  That means that if the function captures any state, then the value that is captured will be whatever it was at the point in time.
+`pulumi` uses `node` to execute a Pulumi application. During execution, when a call to `createLambdaFunction` is encountered, the function is converted to a Lambda at that point in execution. That means that if the function captures any state, then the value that is captured will be whatever it was at the point in time.
```
CyrusNajmabadi:
```suggestion
`pulumi` uses `node` to execute a Pulumi application. During execution, when a call to `new aws.lambda.CallbackFunction` is encountered, the function is converted to a Lambda at that point in execution. That means that if the function captures any state, then the value that is captured will be whatever it was at the point in time.
```

content/docs/reference/serializing-functions.md
```diff
@@ -6,254 +6,263 @@ The Pulumi Node.js SDK provides a core API for converting a JavaScript function
...
+const lambda = new aws.lambda.CallbackFunction("mylambda", {
+  callback: async () => {
+    console.log(obj);
+  }
+});
 ```
 
-When `pulumi` starts executing `createLambdaFunction` it will analyze the JavaScript function code and will see that it uses the 'obj' value.  At that point in time it will use whatever the value is currently to serialize over.  So, in the first example, it will serialize the value `{ a: 1, b: 2 }`, even though right after executing createLambdaFunction the program code will update that value to `{ a: 3, b: 4}`.  In the second example, the code will see the `{ a: 3, b: 4 }` value and will serialize that into the 'run time' code.  
+When `pulumi` starts executing `createLambdaFunction` it will analyze the JavaScript function code and will see that it uses the 'obj' value. At that point in time it will use whatever the value is currently to serialize over. So, in the first example, it will serialize the value `{ a: 1, b: 2 }`, even though right after executing createLambdaFunction the program code will update that value to `{ a: 3, b: 4}`. In the second example, the code will see the `{ a: 3, b: 4 }` value and will serialize that into the 'run time' code.
```
CyrusNajmabadi:
```suggestion
When `pulumi` starts executing `new aws.lambda.CallbackFunction` it will analyze the JavaScript function code and will see that it uses the 'obj' value. At that point in time it will use whatever the value is currently to serialize over. So, in the first example, it will serialize the value `{ a: 1, b: 2 }`, even though right after executing createLambdaFunction the program code will update that value to `{ a: 3, b: 4}`. In the second example, the code will see the `{ a: 3, b: 4 }` value and will serialize that into the 'run time' code.
```

CyrusNajmabadi:
Done!

PR #1383, opened 2019-07-17
Use a canonical URL

cnunciato:
Since we serve this site at multiple URLs (sometimes intentionally, others just by virtue of the way we build and deploy it -- e.g., "www-*", public buckets URLs, etc.), we should specify that the "real" URL is always, ultimately, https://www.pulumi.com. This change makes that clear in the HEAD, sitemap and RSS templates.

### DIFF
--- a/config/_default/config.toml
+++ b/config/_default/config.toml
@@ -26,5 +26,6 @@ enableRobotsTXT = true
     fractions = false
 
 [params]
+    canonicalURL = "https://www.pulumi.com"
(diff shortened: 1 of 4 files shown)


cnunciato:
@chrsmith That's right. Since baseURL is necessarily overridden in various places, it doesn't quite fit for this case. I did look into whether Hugo offered more direct support for this, but their [canonicalization](https://gohugo.io/content-management/urls/#canonicalization) appears to use `baseURL` also, so this seemed like an okay solution in that light.

Issue #1325, opened 2019-07-11
Post-migration website tweaks


justinvp:
Some follow-ups:

- [ ] (justinvp) Hook up Purgecss to remove unused CSS classes (#1291)
- [ ] Fix styles used on AWS Crosswalk images to use Tailwind
- [x] Looks like we have both `/image/pulumi.png` and `/social/pulumi.png` used as meta images. We should standardize on `/image/pulumi.png` and update any redirects as appropriate. (#1374)
- [ ] Implement a two-column layout for the docs landing page. Consider implications on search.
- [ ] Follow-up improvements to "On this Page": generate for "Download and Install", add `sticky` positioning

lukehoban:
@cnunciato Anything pressing here?  I'm going to move out of any milestone for now - as we've punted this a lot and nothing seems critical.  But if you want to pull back in or split into specific work feel free.

lukehoban:
Closing as this is quite old.  Will assume we'll reopen specific issue on anything critical here.

PR #1389, opened 2019-07-17
Hide "Edit this Page" link for generated pages

### LINKED ISSUE #1235: Hide "Edit on GitHub" button on generated pages
Consider hiding the "Edit on GitHub" button on generated pages, such as our API and CLI docs.

justinvp:
We currently invite folks to edit these pages, but they can't be edited like other non-generated content. This change hides the "Edit this Page" link on generated pages. The "Request a Change" link remains visible for all pages.

While making changes here, fixed-up the indentation (looks like tabs were being used instead of our standard 4 spaces) and added `target="_blank"` to the "Request a Change" link.

Fixes #1235

### DIFF
--- a/layouts/partials/docs/edit.html
+++ b/layouts/partials/docs/edit.html
@@ -1,16 +1,23 @@
 {{ with .File }}
     <ul class="p-0 list-none">
+        <!-- Only show "Edit this Page" for pages that are not generated. -->
+        {{ $isCLICommand := hasPrefix .Path "docs/reference/cli/" }}
+        {{ $isNodePackage := hasPrefix .Path "docs/reference/pkg/nodejs/pulumi" }}
+        {{ $isPythonPackage := hasPrefix .Path "docs/reference/pkg/python/pulumi" }}
+        {{ if not (or $isCLICommand $isNodePackage $isPythonPackage) }}
+            <li>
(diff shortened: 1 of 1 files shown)


justinvp:
> The alternate approach would be to link to the upstream repos for editing, but that is probably not worth the hassle especially if we change our doc generation process for the API references.

Yeah, that'd certainly be more ideal, but it's a more involved change (as you allude to). We recently started including a preamble above the docs for Terraform-based providers with links to the upstream repos, so hopefully that will help a little. Let's keep this in mind as we hash out our long-term API doc generation strategy.

> I was wondering if we should disable editing/issue submission for `/docs` as well? Maybe we can have that landing page be a two-column layout instead so the images aren't so pinched?

I like this idea. I think I'll hold off on making any changes as part of this PR. But I could see a two-column layout looking better on the docs landing page. I think we'd still want to put a search box somewhere on the landing page (I want to make our built-in search great). Perhaps the search box could go at the top of the page in the main column of the two-column layout? Should we open a separate issue to consider moving to the two-column layout for the `/docs` landing page (or add it as a task to an existing issue)?

bermudezmt:
I added the two-column layout suggestion as a task on https://github.com/pulumi/docs/issues/1325.

PR #1392, opened 2019-07-18
Fixed createLambdaFunction references on Serializing JavaScript Functions doc

aydrian:
After #1381 was merged, there were still 2 instances of `serverless.function.createLambdaFunction` that needed to be updated in the text.

### DIFF
--- a/content/docs/reference/serializing-functions.md
+++ b/content/docs/reference/serializing-functions.md
@@ -2,13 +2,13 @@
 title: Serializing JavaScript Functions
 ---
 
-The Pulumi Node.js SDK provides a core API for converting a JavaScript function into all the code and files necessary to then have that function be used at runtime within some Cloud, for example in an [AWS Lambda](https://aws.amazon.com/lambda/).
+The Pulumi Node.js SDK provides a core API for converting a JavaScript function into all the code and files necessary to have that function be used at runtime within some cloud, like in an [AWS Lambda](https://aws.amazon.com/lambda/) for example.
 
-There are many cases where a small piece of runtime functionality must be defined as part of a Cloud application, and it may make sense to define that runtime functionality directly inline as part of the Pulumi program. This can augment, or even replace, using runtime code and binaries defined outside of Pulumi in Lambda ZIPs, Docker images, VM images, etc.
+There are many cases where a small piece of runtime functionality must be defined as part of a cloud application, and it makes sense to define that runtime functionality directly inline as part of the Pulumi program. This can augment, or even replace, using runtime code and binaries defined outside of Pulumi in Lambda ZIPs, Docker images, VM images, etc.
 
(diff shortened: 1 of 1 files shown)


bermudezmt:
Headed to lunch - I'd like to preview this locally before I merge it, @aydrian. If all looks good though, you can go ahead and merge it.

content/docs/reference/serializing-functions.md
```diff
@@ -38,11 +38,11 @@ bucket.onObjectCreated("mytrigger", async (eventInfo) => {
 });
 ```
 
-This functionality provides a powerful and convenient way to create your Lambdas, without needing to manually create the index.js file, package up all necessary node_modules directories, specify Roles or RolePolicyAttachments, upload S3 buckets, or do any of the normal work traditionally necessary.
+This functionality provides a powerful and convenient way to create your Lambdas, without needing to manually create the **index.js** file, package up all necessary `node_modules` directories, specify Roles or RolePolicyAttachments, upload S3 buckets, or do any of the traditionally necessary work.
```
CyrusNajmabadi: woudl use `` `index.js` `` not `**index.js**`

content/docs/reference/serializing-functions.md
```diff
@@ -126,9 +125,9 @@ const lambda = new aws.lambda.CallbackFunction("mylambda", {
 });
 ```
 
-Then this would need to serialize the entire object value (because 'bar' itself is used from 'foo'). This process happens conservatively. By default objects are serialized in their entirety, and they are only trimmed down when it can be proved that it is totally safe to do so.
+Then Pulumi would need to serialize the entire object value, since `bar` itself is used from `foo`. This process happens conservatively. By default objects are serialized in their entirety, and they are only trimmed down when it can be proved that it is totally safe to do so.
```
CyrusNajmabadi: maybe "can be proven" instead of "can be proved"?  not sure.

content/docs/reference/serializing-functions.md
```diff
@@ -255,14 +250,12 @@ bucket.onObjectCreated(
...
 
-AWS-Specific Notes:
+> **Note:** `pulumi` will not include `@pulumi/...` packages with the Lambda. These packages exist solely to provide _deployment time_ functionality, and do not contain any code that can work properly at _run time_. They are automatically stripped from a Lambda both to prevent accidental usage and to help reduce the size of the uploaded Lambda.
 
-1. It is optional to specify a reference to "aws-sdk" in your package.json. AWS always includes this package with Lambdas, and so it is not necessary to explicitly include it yourself.
+> ### Referencing `aws-sdk`
+>
+> It is optional to specify a reference to "aws-sdk" in your **package.json**. AWS always includes this package with Lambdas, so it is not necessary to explicitly include it yourself.
```
CyrusNajmabadi: ``` `package.json` ```

content/docs/reference/serializing-functions.md
bermudezmt: Oh I missed this one - thank you!

Issue #1395, opened 2019-07-18
Link to correct upstream azurerm provider


clstokes:
We link to the incorrect upstream azure provider in our docs. The text and link should be to https://github.com/terraform-providers/terraform-provider-azurerm, not the _classic_ provider - note: `azurerm` vs `azure`.

![image](https://user-images.githubusercontent.com/293763/61475491-c387c700-a93f-11e9-8e97-e59ff2692375.png)

justinvp:
The problem is we assume the Pulumi package and Terraform package have the same name, but for Azure that's not the case.

https://github.com/pulumi/pulumi-terraform/blob/f106438be1cfc83d362d591dc216570dfda9a879/pkg/tfgen/docs.go#L168-L171

And for Python:

https://github.com/pulumi/docs/blob/fda97a66dfb880b7c7494d3c89bc89b5c5d81a04/tools/pydocgen/pulumi-docs.json#L21-L23

lukehoban:
Note that `g.info.Name` contains the Terraform Provider name.

```
Name           string                     // the TF provider name (e.g. terraform-provider-XXXX).
```

justinvp:
Yep, I already started on a fix.

stack72:
@justinvp  I believe the PR I merged yesterday #1407 - fixes this right?

justinvp:
#1407 fixes it for the Python docs, but the node docs require a release of Azure (that contains https://github.com/pulumi/pulumi-azure/pull/309) and GCP (that contains https://github.com/pulumi/pulumi-gcp/pull/182), and then regenerating node docs for those packages.

stack72:
ok, I will take care of this as soon as I get to the office - we have some python build issues

stack72:
@justinvp GCP - https://github.com/pulumi/docs/pull/1425

stack72:
@justinvp Azure - https://github.com/pulumi/docs/pull/1427

stack72:
Closed via #1425 and #1427

PR #1407, opened 2019-07-19
Fix TF provider name for Azure and GCP in Python docs

justinvp:
First commit updates the `pydocgen` tool, second commit regenerates the Python docs.

@stack72, Regenerating the Python docs included more changes than just the attribution changes. Looks like we sometimes forget to generate the Python docs (via `./scripts/generate_python_docs.sh`) when updating docs for new package versions.

Part of #1395


stack72:
This is awesome @justinvp - when I am in the office this week, I want to talk with you about a workflow for our provider docs generation that should help to keep these differences down!

PR #1411, opened 2019-07-19
Update build status in README.md

justinvp:
`master` now deploys the production site. I decided to just remove staging as we haven't used it yet, and may soon find we don't need it at all.


justinvp:
> We might have the same widget in the `home` repo

I checked and it's already only showing status for `master` there.

PR #1413, opened 2019-07-19
Exclude verifying flaky youtube links

justinvp:
youtube.com is returning 204 response more often than not, which the broken link checker considers broken, which breaks the build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ test:
 		--exclude "https://www.linkedin.com/" \
 		--exclude "https://marketplace.visualstudio.com/items?itemName=pulumi.build-and-release-task" \
 		--exclude "https://blog.mapbox.com/" \
-		--exclude "https://www.youtube.com/pulumitv"
+		--exclude "https://www.youtube.com/"
 
 .PHONY: validate
 validate:


ekrengel:
Is there a way to update/configure the broken link checker to accept 204s?

justinvp:
> Is there a way to update/configure the broken link checker to accept 204s?

Not that I can tell...

```
$ ./node_modules/.bin/blc --help
Broken Link Checker: Find broken links, missing images, etc in your HTML. (v0.7.8)

Usage
  blc [OPTIONS] [ARGS]
                         
Options                  
  --exclude               A keyword/glob to match links against. Can be used multiple times. 
  --exclude-external, -e  Will not check external links.                                     
  --exclude-internal, -i  Will not check internal links.                                     
  --filter-level          The types of tags and attributes that are considered links.        
                            0: clickable links                                               
                            1: 0 + media, iframes, meta refreshes                            
                            2: 1 + stylesheets, scripts, forms                               
                            3: 2 + metadata                                                  
                            Default: 1                                                       
  --follow, -f            Force-follow robot exclusions.                                     
  --get, -g               Change request method to GET.                                      
  --help, -h, -?          Display this help text.                                            
  --input                 URL to an HTML document.                                           
  --host-requests         Concurrent requests limit per host.                                
  --ordered, -o           Maintain the order of links as they appear in their HTML document. 
  --recursive, -r         Recursively scan ("crawl") the HTML document(s).                   
  --requests              Concurrent requests limit.                                         
  --user-agent            The user agent to use for link checks.                             
  --verbose, -v           Display excluded links.                                            
  --version, -V           Display the app version.                                           
                         
Arguments                
  INPUT                   Alias to --input    
```

We could consider using the `broken-link-checker` JS library to customize the checks.

ekrengel:
Hmm guess that makes sense since why would you link to no content

PR #1420, opened 2019-07-22
Only generate an RSS feed for the blog

justinvp:
We currently generate RSS feeds for all sections/taxonomies, which means we're generating and publishing a large number of `rss.xml` files. This change makes it so we only generate a single RSS feed for the blog at `/blog/rss.xml`.

Part of #1277

### DIFF
--- a/config/_default/config.toml
+++ b/config/_default/config.toml
@@ -13,6 +13,16 @@ enableRobotsTXT = true
 [outputFormats.RSS]
     baseName = "rss"
 
+# We set the outputs explicitly to just "HTML" as otherwise they default to both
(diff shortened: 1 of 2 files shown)


justinvp:
> It's kinda odd that in `config.toml` HTML is capitalized, but in `_index.md` it is in lowercase. I assume this is required? But other that (inconsequential) issue, looks good! Thanks for fixing.

Yeah, I thought that was kind of strange as well. I'm not sure if it's required, but that's how it's shown in the docs, so that's how I did it, e.g. https://gohugo.io/templates/output-formats/#customizing-output-formats

Screen Shot 2019-07-22 at 10 16 22 AM

Issue #1419, opened 2019-07-22
Is Cloud Firestore supported ?


sebght:
What is the current offer concerning Cloud Firestore integration ?
[Here](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/gcp) _firestore_ stands out as a supported module. However, the [doc](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/gcp/firestore/) is clearly empty.

As [TF's doc](https://www.terraform.io/docs/providers/google/r/firestore_index.html) does exist, I suppose that Pulumi also uses this module. Then, it might be good to update the documentation associated 😁

stack72:
Hi @sebght 	

Apologies for this not being in the documentation - on Friday we discovered there was an issue with the Firestore SDK. This has been rectified and we are able to regenerate the SDK for it - which means the documentation will also be available

This new release and it's documentation will be available later today (US time)

If there are any problems then please let me know

Thanks

Paul

stack72:
@sebght https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/gcp/firestore/ :)

PR #1425, opened 2019-07-22
Generate GCP documentation based on pulumi-gcp 0.18.14

stack72:
Fixes: https://github.com/pulumi/docs/issues/1419
Fixes: https://github.com/pulumi/pulumi-gcp/issues/176


stack72:
@justinvp doing so now

Issue #1402, opened 2019-07-19
First first batch of website performance issues


chrsmith:
Running www.pulumi.com through Google's PageSpeed Insights, there are a set of issues we can address to should have an outside impact on our website's performance.

See:
https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fwww.pulumi.com%2F&tab=mobile

I'll let you figure out how much time we want to spend on this in the current milestone, but hopefully just fixing a few key things will get us to a better rating. The things that stand out to me are:

## Tweak the Images we Use

We haven't given much thought to the images we serve from the homepage, but the tool points out that some of them should be resized, and/or be compressed using smaller image files.

We probably don't want to revisit every image we have across the site. But certainly on the homepage this would payoff.

## Fixing our Cache Behavior

The AWS-S3-CloudFront stack we use to serve the website just put some basic defaults in for cache TTLs. But a small tweak would go a long way here.

Our default TTL for all file content is to cache it for 60s. See:
https://github.com/pulumi/docs/blob/master/infrastructure/index.ts#L108

That can probably be upped to maybe even 10s of minutes, depending on how frequently we expect the website's contents to change. (And the risk we'd want to run with the browser having different versions of files based on caches, etc.)

But, I think we can create a different "origin" for the CloudFront distribution, e.g. `www.pulumi.com/static/*` and have that use a _different_ caching policy. e.g. cache those larger image/font files for perhaps a weeks at a time, or even longer.
https://github.com/pulumi/docs/blob/master/infrastructure/index.ts#L88

Part of https://github.com/pulumi/home/issues/514

cnunciato:
> To pass this audit, encode all of these images in WebP. 

🤨 

None of the PNG alternatives here are supported across the browsers we need to support, so I think we should stick with PNGs for now.  (I was able to squish the images referenced in the report, though, to just about match the savings estimates.)

chrsmith:
> To pass this audit, use a non-standard image format that isn't recognized by non-Google browsers.

(ノಠ益ಠ)ノ彡┻━┻

PR #1426, opened 2019-07-22
Wrap long unbroken strings on mobile

cnunciato:
This change adds CSS that instructs the browser to wrap long, unbroken strings where possible on mobile.

Before:

After: 

Part of #1402.

### DIFF
--- a/assets/sass/_api-nodejs.scss
+++ b/assets/sass/_api-nodejs.scss
@@ -48,3 +48,8 @@
         }
     }
 }
+
+.pdoc-module-header,
+.pdoc-module-contents {
+    @apply break-all;
+}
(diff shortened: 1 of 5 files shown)


assets/sass/_docs.scss
```diff
@@ -0,0 +1,14 @@
+// CSS overrides for generated SDK and CLI docs.
+
+.section-docs {
+
+    .section {
```
justinvp: Does this class only show up in the generated Python docs? If so, should these styles go in `_api-python.scss`?

cnunciato: The fix is aimed at an issue that shows up in multiple places (not just Python) &mdash; but now I see there are other areas I've missed here, so will need to augment this.

cnunciato:
@justinvp Ended up adding these overrides to the the `_api-*.scss` files after all. Thanks for reminding me they were there. 😄

PR #1431, opened 2019-07-23
chore(migrate-nodegroups): rm tag, fix hdrs, add time est, add filename

metral:


content/docs/reference/tutorials/kubernetes/tutorial-eks-migrate-nodegroups.md
```diff
@@ -112,7 +112,7 @@ $ pulumi stack output nginxServiceUrl
 
 Example: 
 
-```bash
+```
```
justinvp: Just curious why? Leaving it `bash` allows the `pulumi stack output nginxServiceUrl` to be copied to the clipboard easily.

metral: You're correct - there is a dupe, copyable command right above it for the user. This code block is an example of the output of running that command. 

See both code blocks rendered: https://github.com/pulumi/docs/blob/02f660094ea729ae6763305353fc88760fce3b7b/content/docs/reference/tutorials/kubernetes/tutorial-eks-migrate-nodegroups.md#access-the-workload

Does this make sense?

justinvp: Ah, I missed the block above. OK, that makes sense.

Issue #831, opened 2019-01-31
ToC & Information Architecture


justinvp:
- Language chooser, move that to be global? -> IA and TOC
- Move to having a single, global TOC across all pages. And punt on rejiggering the display or organization of that TOC.
- Add breadcrumbs for docs?

See also the discussion in #814 regarding unified TOC.

chrsmith:
Moving to next milestone. We'll be making some larger infrastructure changes for pulumi.io, and it will fall out of that.

justinvp:
With the website migration behind us, we'll be tackling this next milestone.

justinvp:
- [ ] Consider moving `/docs/quickstart/` to the root, e.g. `/get-started` (and decide where we want tutorials to live long term)

justinvp:
This includes improving the /docs table of contents to be easier to navigate. Right now most docs are buried under Reference.

bermudezmt:
> * [ ]  Consider moving `/docs/quickstart/` to the root, e.g. `/get-started` (and decide where we want tutorials to live long term)

- I'm in favor of doing `/docs/get-started` for consistency if you meant moving it to `pulumi.com/get-started` instead. Does Hugo support folder redirects so we don't have to add aliases to the individual pages?

- I think tutorials can be its own category. 

> This includes improving the /docs table of contents to be easier to navigate. Right now most docs are buried under Reference.

@justinvp I'd like to give this a quick pass this week. @lukehoban had suggested we could either have **Get Started** and **Reference** expanded by default or move up the menu items under **Reference**. What do you think of this proposed new navigation?

```
- Intro to Pulumi / What is Pulumi? (/intro)
  - Architecture & Concepts 
    - Programming Model
    - Projects
    - Stacks
    - Configuration and Secrets
    - Organizing Projects and Stacks
    - How Pulumi Works
    - State and Backends
  -  Pulumi vs. Other
     - Terraform
     - Cloud Templates
     - Cloud SDKs
     - Serverless
     - Kubernetes YAML
     - Chef, Puppet, etc.
     - Custom Solutions
  - Supported Clouds and Languages
     - Cloud Providers
        - AWS
        - Azure
        - Google Cloud
        - Kubernetes
        - OpenStack
     - Languages
        - Node.js
        - Python
- Download and Install
- Changelog (probably should rename to release notes?)
- Get Started 
- Tutorials
  - AWS 
    - ...
  - Examples (eventually we'll add a table like this: https://docs.docker.com/samples/)
[...]

PR #1432, opened 2019-07-23
Add breadcrumb to docs pages

justinvp:
Draft as this is still a work-in-progress and not ready for review. FYI, @bermudezmt.

Initial breadcrumb for second level nav inside the docs section of the site. Still needs some style/design work.

breadcrumb

---

As part of cleaning up the TOC, we should consider getting rid of the top-level "Docs" link in the left nav, as it'd be redundant with the breadcrumbs:

Screen Shot 2019-07-23 at 10 25 03 AM

---

Will need to fix-up the code-gen of the API docs to have `linktitle` front matter that provides a shorter title for the breadcrumb, e.g. the breadcrumb item should probably be the shorter `@pulumi/pulumi` rather than `Package @pulumi/pulumi`:

Screen Shot 2019-07-23 at 10 26 02 AM

---

The breadcrumb is based on the directory structure of the files on disk in `content/docs`... Currently, there are some places where the hierarchy of the pages on disk doesn't match our "fabricated" TOC hierarchy, which makes the breadcrumbs confusing in those areas.

Examples:

  - Many of the pages under `Teams & Collaboration` TOC live on disk under `content/docs/reference/service`, but some pages listed there in the TOC aren't, like `Pulumi GitHub App` which actually lives at `content/docs/reference/cd-github.md`.

  Screen Shot 2019-07-23 at 10 27 07 AM

  - The pages under the `Continuous Delivery` are all just in `content/docs/reference/*` instead of being in something like `content/docs/reference/cd/*`.
  - The `Command Line` TOC page is at  `content/docs/reference/commands.md` but the actual commands are under `content/docs/reference/cli/*`.

We should fix-up the on-disk structure as part of reorganizing the TOC in #831, which will make the breadcrumbs make more sense.

Part of #831

### DIFF
--- a/assets/sass/_breadcrumb.scss
+++ b/assets/sass/_breadcrumb.scss
@@ -1,0 +1,20 @@
+ol.breadcrumb {
+    @apply flex flex-wrap py-0 px-0 mt-2 mb-4 list-none;
+
+    li {
+        @apply my-0;
+
+        &.active a {
+            color: inherit;
+        }
+
+        & + li {
+            @apply pl-2;
+
+            &::before {
+                @apply inline-block pr-2;
+                content: "/";
+            }
+        }
+    }
+}
--- a/assets/sass/styles.scss
+++ b/assets/sass/styles.scss
@@ -7,6 +7,7 @@
 @import "animation";
 @import "api-nodejs";
 @import "api-python";
+@import "breadcrumb";
 @import "buttons";
 @import "code";
 @import "copy-button";
--- a/content/docs/_index.md
+++ b/content/docs/_index.md
@@ -1,9 +1,9 @@
 ---
 title: Pulumi Documentation
+linktitle: Docs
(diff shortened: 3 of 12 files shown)


justinvp:
This is just blocked on moving some pages around into a more logical directory hierarchy. Tracked in #1491.

lukehoban:
This looks great!

> As part of cleaning up the TOC, we should consider getting rid of the top-level "Docs" link in the left nav, as it'd be redundant with the breadcrumbs:

Didn’t we already do this in most recent TOC revamp?  I see something pretty different than your screenshots in current site.

justinvp:
> Didn’t we already do this in most recent TOC revamp? I see something pretty different than your screenshots in current site.

Yes. This was all implemented and screenshots taken before that change.

justinvp:
This is being folded into #1563

PR #1433, opened 2019-07-23
Minor copy button improvements

justinvp:
There are some places where we have comments at the end of commands, so strip those.

Also, specify languages for a few code snippets so they get highlighting and the copy button.


content/docs/reference/clouds/aws/setup.md
```diff
@@ -68,15 +68,15 @@ This will have created the `~/.aws/credentials` file and populated it with the e
 
 It is possible to create this file by hand.  For example:
 
-```
+```ini
```
bermudezmt: @justinvp I know you referenced this in a PR before - but what syntax highlighter are we using again? I haven't seen `ini` as a language class before. :)

justinvp: We use Hugo's default syntax highlighter: Chroma.

https://gohugo.io/content-management/syntax-highlighting/
https://github.com/alecthomas/chroma

The AWS shared credentials file uses the INI file format.

https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#the-shared-credentials-file
https://en.wikipedia.org/wiki/INI_file

PR #1434, opened 2019-07-23
Add description to /why-pulumi

erinhyperplane:


justinvp:
@erinhyperplane, this change looks fine to me, but I am curious: Is it deliberate that this page gets its own custom `meta_desc` that is different from the default one?

https://github.com/pulumi/docs/blob/62e4596ae7c15221b8a769d0696f67672e6227af/layouts/partials/head.html#L51

erinhyperplane:
@justinvp, this is @chrsmith and I testing changing meta description on the site as part of Q3 SEO process. So yes, it is intentional to swap the content. :)

chrsmith:
@justinvp yeah, for SEO'in up the place we probably don't want to have that default description used anywhere... but what's interesting is that in the website today the description is empty. So I suspect there is a bug somewhere in the logic in that partial.

www.pulumi.com/why-pulumi/

```
        <meta name="description" content="">
        <meta property="og:description" content="">
```

justinvp:
> So I suspect there is a bug somewhere in the logic in that partial.

@chrismith, want me to investigate?

chrsmith:
@justinvp If you have the cycles to spare, it'd be nice to see if it is a 5m fix. But it isn't worth spending much more time than that. (Since having a bunch of generic, descriptions is marginally better than a _real_, page-specific one.)

justinvp:
https://github.com/pulumi/docs/pull/1445

PR #1435, opened 2019-07-23
Do not use "latest" GKE engine in examples.

clstokes:
Replicating changes from https://github.com/pulumi/examples/pull/343.


justinvp:
Isn't there also a use on https://github.com/pulumi/docs/blob/master/content/blog/program-kubernetes-with-11-cloud-native-pulumi-pearls/index.md ?

clstokes:
@justinvp thanks, good catch, just pushed another commit fixing more occurrences.

PR #1438, opened 2019-07-23
Update get.pulumi.com to v0.17.25

metral:
### Proposed changes

Update get.pulumi.com to v0.17.25


static/latest-version
```diff
@@ -1 +1 @@
-0.17.24
\ No newline at end of file
+0.17.25
```
justinvp: The new line is causing the broken link checker to fail because of the line feed:

```
Getting links from: http://localhost:1313/docs/reference/install/
...
├─BROKEN─ https://get.pulumi.com/releases/sdk/pulumi-v0.17.25%0A-darwin-x64.tar.gz (HTTP_403)
├─BROKEN─ https://get.pulumi.com/releases/sdk/pulumi-v0.17.25%0A-linux-x64.tar.gz (HTTP_403)
├─BROKEN─ https://get.pulumi.com/releases/sdk/pulumi-v0.17.25%0A-windows-x64.zip (HTTP_403)
```

We could fix up `content/docs/reference/install.md` to be resilient for the new line, but I'm not sure if our install scripts and other things that depend on https://www.pulumi.com/latest-version are resilient, so best to remove the new line for now.

metral: Done.
Just pushed a commit with this change.

metral:
Feedback has been addressed. PTAL @justinvp

Issue #927, opened 2019-03-08
Styling around certain docs pages makes sections hard to distinguish


CyrusNajmabadi:
This came up with a discussion with @lukehoban .  Many or our docs pages look like basically a long flat list which is hard to grok.  For example:

https://pulumi.io/reference/programming-model.html

If you go to the `Outputs` section: https://pulumi.io/reference/programming-model.html#outputs then you'll see it is composed out of 5 subsections: The main explanation, followed by `Apply`, `All`, `Convert Input to Output` and `Working with Outputs and strings`.

Problems here are:
1. It's really hard to tell that these subsections even belong to the general `Output` section.  It looks like these are all just top level concepts.
2. Even if you do get that these are subsections, it's really daunting.  You have a wall of text thrown at you.

@lukehoban and i discussed this and understood that this is basically a bikeshedding topic.  There are several things that could be tried out, but really, we'd need to see the results to get a good feel if it addresses the above.  Things that could help include, but are not necessarily limited to:

1. just changing the sizes of headings/subheadings
2. potentially indenting subsections.
3. including glyphs for top level headings.
4. including line-separators around large groups
5. underlining a heading.
6. making subsections easily collapsible (or starting them out collapsed).
7. breaking subsections into their own pages (though it feels like this has its own set of negatives).

Ideally, to me at least, someone coudl get to the `Outputs` section and read a quick blurb that helps give an intuition about them.  In fact, we already have a good blurb for `Outputs` in the opening paragraph.  At that point though, i think the rest of hte info should "get out of the way" until you need it.  So, for example, you could have this:

> Most of the time, when data needs to be passed from one Resource to another, it is sufficient to just pass the Output for that data 'as is'.  However, there may be cases where the data needs to be manipulated first.  There are many ways to go about doing this which you can read about below:

> <details><summary>Accessing simple properties of the Output</summary> yadda yadda</details>
> <details><summary>Combing data from multiple outputs</summary> yadda yadda</details>

That way things look less daunting, and you can expand and learn about this stuff at your own pace.

--

Note: as mentioned above, this is a bikeshedding topic.  Woudl love to work with someone html/css knowledgable to quickly knock out several options for us to compare/contrast.

CyrusNajmabadi:
Tagging @justinvp for  ideas!

cnunciato:
An "On This Page" menu could make this a lot better; it'd help with both orientation/wayfinding and for navigation, and would remain fixed on the desktop as you scroll -- e.g.: 

[...]

PR #1441, opened 2019-07-23
Move "On this Page" feature to right nav

bermudezmt:

### Proposed changes

  - Disabled for home but will need to be disabled for other content types
  - Needs additional styling, plus consider making the navigation fixed

### Related issues (optional)

Addresses #927

### DIFF
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -22,7 +22,9 @@ function bindToggles(selector) {
 function generateMiniToc() {
     var toc = $(".mini-toc > ul");
     if (toc) {
-        $("h2").each(function () {
+        toc.addClass("sidenav-subsection text-sm");
+        $("h2, h3").each(function () {
+            $("h3").addClass("sidenav-topic pl-2");
             var id = $(this).attr("id");
             var text = $(this).text();
             if (id && text) {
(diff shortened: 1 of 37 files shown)


bermudezmt:
@cnunciato This is what it looks like at the moment. May need your help with the following: Applying indentation to H3 elements, and sticky positioning for the right TOC. I wanted to see if this is a good general direction. 

@justinvp I experimented with Hugo's `TableofContents` feature, but this seems like the cleanest quickest way to implement this change for now.

Screenshot from 2019-07-24 11-45-38

cnunciato:
I'm seeing the menu appear on the right-hand side of pages that don't have the content to support it. Can we suppress it in these situations? Happy to help here if you like.

bermudezmt:
Yes, happy to take the help as I'm back to content focus for the rest of the week. Thanks, @cnunciato

justinvp:
@bermudezmt, I've made some changes to get this into shape and address @cnunciato's feedback. I'll push a commit or two soon.

justinvp:
I pushed a commit with various improvements. I believe this is in a good enough state to merge (modulo any feedback), and we can iterate further subsequently (e.g. making it sticky as you scroll). @bermudezmt, @cnunciato, please take a look.

bermudezmt:
Thank you for the improvements, @justinvp . I noted some follow-up work we have to do on https://github.com/pulumi/docs/issues/1325 but I agree, this is good to merge and will be a helpful addition. :)

PR #1445, opened 2019-07-24
Fix empty meta description tags and check params consistently

justinvp:
The `{{ else if ne .Summary "" }}` check is always evaluating to true (probably because `Summary` is typed as `template.HTML` instead of a `string` in Hugo's [source](https://github.com/gohugoio/hugo/blob/708d4ceebd491c6a89f271311eb8d94d6b5d58bc/resources/page/page.go#L68)), so we go down this path, but when then using `.Summary`, the value is empty. This results in many pages having empty `<meta name="description">`  instead of falling back to using the default description. The fix is to check `{{ else if .Summary }}` which will only evaluate to true if `.Summary` is set and is a non-zero value.

While making changes here, I cleaned up how we access other params throughout. We'll often use `if isset .Params "foo"` to check if the `foo` param is set, but this only checks if the param is set -- it doesn't check if the param actually has a value. The fix is to just use `if .Params.foo`, which will only be true if `foo` is set and is not: `false` (boolean), `0` (integer), or a zero-length array, slice, map, or string.

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -4,7 +4,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
     <meta name="google-site-verification" content="j-Opn93zR4YQcV5bbYN99BmNc48BT5mzd2VpzKaW9YY" />
 
-    {{ if isset .Params "redirect_to" }}
+    {{ if .Params.redirect_to }}
         <meta http-equiv="refresh" content="0; url={{ .Params.redirect_to }}" />
     {{ end }}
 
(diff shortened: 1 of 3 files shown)


layouts/partials/head.html
```diff
@@ -114,17 +114,17 @@
     {{ end }}
 
     {{/* Tests to enforce required frontmatter for certain content types. */}}
-    {{ if eq .Type "blog" }}
-        {{ if not (isset .Page.Params "authors") }}
+    {{ if and (eq .Type "blog") (.IsPage) }}
```
chrsmith: Nit: Do we need parens around `.IsPage`?

justinvp: No, it just looked cleaner to my eyes with the parens to the left. But since it's not necessary and you pointed it out, I've removed.

PR #1451, opened 2019-07-25
Reorganize side navigation

bermudezmt:
## Proposed changes

Moving the pages nested under Reference will take a lot of doing, so I took Luke's suggestion and have these items displayed by default. 

Changed the order around, merged supported clouds & languages, still a WIP.

### Related issues (optional)

Partially addresses #831 

Screenshot from 2019-07-24 20-00-08


content/docs/reference/clouds/_index.md
```diff
@@ -1,8 +1,10 @@
 ---
-title: Supported Clouds
+title: Supported Clouds & Languages
```
lukehoban: I think at least for now - we should keep these separate. They are quite different dimensions of our product. Feels reasonable to keep them distinct at least until we want to do a broader revamp of the information architecture here.

content/docs/reference/concepts.md
```diff
@@ -1,9 +1,9 @@
 ---
-title: Architecture & Concepts
+title: Pulumi Cloud Development
```
lukehoban: I think keep this “Architecture and Concepts” for now - or something similar. 

Pulumi Cloud Development is too vague (could apply to any content in this site!). In fact, instead I think we want to rewrite the content in this page to use more consistent terminology (I am not sure the Pulumi Cloud Development Platfrom is actually an official name we use in general).

PR #1452, opened 2019-07-25
Emit `{ }` instead of `__type` for empty type literals

### LINKED ISSUE #1429: Formatting changes halfway through page
Something is causing the font style to change halfway down this API Reference page. Text is going from "normal" to "bold" for some reason.

https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/aws/s3/#Inventory-bucket

image

justinvp:
If we have a reference type named `__type`, it's an empty type literal that should be emitted as `{ }`.

Fixes #1429

### DIFF
--- a/content/docs/reference/pkg/nodejs/pulumi/aws/s3/_index.md
+++ b/content/docs/reference/pkg/nodejs/pulumi/aws/s3/_index.md
@@ -2032,7 +2032,7 @@ The S3 bucket configuration where inventory results are published (documented be
             sseKms: <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined'>undefined</a></span> | {
                 keyId: <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String'>string</a></span>;
             };
-            sseS3: <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined'>undefined</a></span> | __type;
+            sseS3: <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined'>undefined</a></span> | { };
         };
         format: <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String'>string</a></span>;
         prefix: <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined'>undefined</a></span> | <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String'>string</a></span>;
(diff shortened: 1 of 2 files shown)


justinvp:
> The change seems fine, but for my own understanding, where is `__type` coming from? Is the `tscdocgen` running over the TypeScript compiler's AST? And that's some node that doesn't have any type information? Or is it just how they declare an "empty" type?

`__type` is the name they use for type literals, which by definition do not have a name. I went into more detail in the referenced issue #1429. Here's the associated TypeDoc code: https://github.com/TypeStrong/typedoc/blob/e9b28ee2c336617687a968820515b6a37d44515f/src/lib/converter/types/reference.ts#L127-L128

Note that `__type` is used for any type literal -- it doesn't have to be empty. But I found from running typedoc on all our packages, that the combination of `type` being `reference` and name being `__type` that it's an empty type literal. If it wasn't an empty type literal, it'd be `type` == `reflection` and will contain a declaration (for the properties of the type), which we handle in the case below it:

https://github.com/pulumi/docs/blob/ad8149d751538266d63ef0e2a1f99a0cebd3d168/tools/tscdocgen/main.go#L1011-L1036

> Also, does `&lt;{ }&gt;;` get rendered as `<{ }>;;`? Is that, uh, right? Not sure.

It gets rendered as:

```typescript
destination: pulumi.Input<{
    bucket: pulumi.Input<{
        accountId: pulumi.Input<string>;
        bucketArn: pulumi.Input<string>;
        encryption: pulumi.Input<{
            sseKms: pulumi.Input<{
                keyId: pulumi.Input<string>;
            }>;
            sseS3: pulumi.Input<{ }>;
        }>;
        format: pulumi.Input<string>;
        prefix: pulumi.Input<string>;
    }>;
}>;
```

The actual [source code](https://github.com/pulumi/pulumi-aws/blob/a4e17c3cf20eda684061d667ba19db82920d6ca3/sdk/nodejs/s3/inventory.ts#L220) looks like:

```typescript
readonly destination: pulumi.Input<{ bucket: pulumi.Input<{ accountId?: pulumi.Input<string>, bucketArn: pulumi.Input<string>, encryption?: pulumi.Input<{ sseKms?: pulumi.Input<{ keyId: pulumi.Input<string> }>, sseS3?: pulumi.Input<{  }> }>, format: pulumi.Input<string>, prefix?: pulumi.Input<string> }> }>;
```

According to this [TS playground](http://www.typescriptlang.org/play/#code/C4TwDgpgBAkgdmArsAPAFQHxQLxTVAHygAUAnAewFsBLAZwnQwCgnq5gJSAzAQwGNoAMXLkAjFADeTKDKgATCLWBsey8nABcsBMhRTZBqACNEfANYRgW+ElT7DD-n3KJ2MOdZ2olpNgHMMAG5pB0MTc0sAQVJNbVsUH38gkNDZCDg+UnA1WJtde1SHWnoAaUpaT3iCwtSLEHdK3US4AOCahwBfZPbDYogAZQBmRrsoLrb28ZTCrnJSSlURhOBfFu72sFIILmoADyXm1umDKZPkjpY2Dm5+IREAJklphSUVHKXqg3CLKzj849CThcbg8f28KySABoAQ5vlEYgcIWtoT10plstR1B8YYU+mUKmC9DienUGoTDhgUT1ZF0qdSoH0hh8xpTiTTWT1ZvNFuSkQE6YVNts9ojVgEcad2cELkA), either commas or semicolons are valid.

Issue #1404, opened 2019-07-19
docgen leading to broken links


chrsmith:
Looking through the Moz's page insights, we are serving broken links from our generated documentation.

TWO BUGS FOR THE PRICE OF ONE!

## Issue 1

Class references in the TypeScript docs like to source code, which will always fail.

For example, on page: https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/postgresql/#Database

Clicking “class Database” takes you to https://www.pulumi.com/database.ts#L30. In an ideal world we'd link to https://github.com/pulumi/pulumi-postgresql/branch/master/database.ts or something. But at the very least we shouldn't render the broken link.

## Issue 2

Probably WAI, but hopefully we can fix.

Take a look at the output for this page:
`/content/docs/reference/pkg/nodejs/pulumi/gcp/container/_index.md`

```
The IP address range of the Kubernetes services in this
cluster, in [CIDR](http:en.wikipedia.org/wiki/Classless_Inter-Domain_Routing)
notation (e.g. `1.2.3.4/29`). Service addresses are typically put in the last
`/16` from the container CIDR.
```

Notice that `http:en.wikipedia.org/wiki/Classless_Inter-Domain_Routing` should probably be `http://` and probably even `https://`.

Is this a problem with the underlying docs? Even if so, perhaps its worth adding a quick check to auto fix this on the docgen side? Perhaps there are other instances of `http[s]?:[^/]`?

Part of https://github.com/pulumi/home/issues/514

justinvp:
RE: Issue 2:

This is an issue in `terraform-provider-google`. I've submitted a fix at https://github.com/terraform-providers/terraform-provider-google/pull/4103

justinvp:
Going to close this as Issue 1 is addressed and Issue 2 will be fixed upstream with https://github.com/terraform-providers/terraform-provider-google-beta/pull/977

PR #1453, opened 2019-07-25
Fix broken links in @pulumi/postgresql docs

justinvp:
We weren't including `@pulumi/postgresql` in the `gitHubBaseURLs` map, so links to types in that package in the API docs were broken.

Modified the code to return an error if a package is missing from the map, so we don't silently introduce these broken links again in the future with new packages.

Part of #1404

### DIFF
--- a/content/docs/reference/pkg/nodejs/pulumi/postgresql/_index.md
+++ b/content/docs/reference/pkg/nodejs/pulumi/postgresql/_index.md
@@ -58,7 +58,7 @@ import * as postgresql from "@pulumi/postgresql";
 <li><a href="#SchemaState">interface SchemaState</a></li>
 </ul>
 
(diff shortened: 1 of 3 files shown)


justinvp:
BTW, I didn't get to it as part of fixing this issue, but we have some minor technical debt here -- it'd be better if we only had to specify packages (and associated metadata) in a single place. Right now it's spread out across various .sh, .go, .json, and, .toml files and it's not really documented what needs to be updated when adding a new package, so it's very easy to forget to update one of the places.

justinvp:
> Why wouldn't our broken link checking have found a problem with `<a href="/config/vars.ts">config/vars.ts</a>`? It seems like automation should have caught this.

Because we have `/docs/reference/pkg` excluded from the checker due to the number of broken links. I didn't originally add this exclusion -- I'll have to check to see what the current state is.

https://github.com/pulumi/docs/blob/84df8870bfb1bf71fbe136132acf0eb1800df08a/Makefile#L42-L59

PR #1461, opened 2019-07-27
Make label on blog sidebar an H1

chrsmith:
There are a lot of pages on pulumi.com that Moz complains about with "Pages with Missing or Invalid H1". A significant number of those are pages where we list a series of blog posts, such as:

https://www.pulumi.com/blog/
https://www.pulumi.com/blog/page/6/
https://www.pulumi.com/blog/tag/infrastructure/
https://www.pulumi.com/blog/tag/aws/page/2/

For all of these pages, we list blog entires with an `<h2>`, but nowhere on the page is there actually an `<h1>`.

For `taxonomy/tag.html` I just promoted the existing header to be an `<h1>`, as this just makes a ton of sense. However, for the general "blog entries" listing, there isn't an element in there already.

I didn't want to add one and clutter up the page, so I just changed the "🍹 Pulumi Blog" on the left sidebar to be an `<h1>`. (It will be sized the same as it is today, since otherwise it would look a bit jarring.)

The downside is that for these pages there will be possibly two H1 tags. However, the current advice seems to be that having multiple H1s on a page isn't bad for SEO. (And if anything, seems appropriate in this case, too.)

- "🍹 Pulumi Blog" and "Posts Tagged Azure" (for the tag page)
- "🍹 Pulumi Blog" and "Chris Smith" (for the author's page)
 
If you have a suggestion for a better fix let me know. Part of https://github.com/pulumi/home/issues/514

### DIFF
--- a/layouts/partials/blog/sidebar.html
+++ b/layouts/partials/blog/sidebar.html
@@ -11,9 +11,9 @@
         <a href="{{relref . "/blog" }}">🍹 Pulumi Blog</a>
     </div>
 
-    <h3 class="no-anchor hidden md:flex mt-4">
+    <h1 class="no-anchor hidden md:flex mt-4 text-2xl">
         <a href="{{relref . "/blog" }}">🍹 Pulumi Blog</a>
-    </h3>
+    </h1>
 
     <!-- Sidebar content. Hidden on smaller displays unless toggled. -->
     <div class="blog-sidebar-content hidden md:block">
--- a/layouts/taxonomy/tag.html
+++ b/layouts/taxonomy/tag.html
@@ -11,7 +11,7 @@
                 {{ $paginator := .Paginate .Data.Pages.ByDate.Reverse }}
 
                 <header>
-                    <h3 class="no-anchor">Posts Tagged {{ .Data.Term }}</h3>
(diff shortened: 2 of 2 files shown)


justinvp:
@chrsmith, I was listening to a [podcast episode about semantic HTML](http://www.fullstackradio.com/118) a couple weeks ago (an episode of Full Stack Radio hosted by Adam Wathan, author of Refactoring UI and creator of TailwindCSS), and they were saying that semantically there should only be a single H1 on the page. Even if multiple H1s doesn't affect SEO, I'd prefer to be semantic, if possible.

What do you think about making the sidebar header an H1 only on the blog homepage, otherwise an H3?

I created a branch based on your branch with another commit. See https://github.com/pulumi/docs/commit/838b603453db23c869489010fe7e9553d36680b6

I was also reading that hiding H1 tags might not be good for SEO, and currently we hide hide it for small screens (not sure if this impacts SEO). So my change additionally tweaks the menu toggle so that the header tag is always visible, regardless of screen size.

joeduffy:
FWIW, this is my understanding of H1s vis-a-vis SEO also.

cnunciato:
Sorry @chrsmith -- I totally missed this and took care of it with #2079. I'll close.

PR #1462, opened 2019-07-29
Remove the tour

### LINKED ISSUE #1319: Remove the tour
Now that we have the new Getting Started flow, [the tour](https://pulumi.io/tour) seems unnecessary. Or, if we're going to keep it as a way of teaching you more core Pulumi concepts, it needs to change fairly significantly.

Note that we still link to it from our GitHub repo landing page, which we should also fix:

image

justinvp:
Now that we have the new Get Started guide, the tour is unnecessary. All pages link to relevant content elsewhere on the site.

Fixes #1319


content/docs/tour/basics-programs.md
```diff
@@ -1,70 +1,3 @@
 ---
-title: Programs
-menu:
-  tour:
-    parent: tour
-    weight: 1
+redirect_to: /docs/reference/programming-model/#programs
```
justinvp: FYI, in general, we should use Hugo's built-in [`aliases`](https://gohugo.io/content-management/urls/#aliases) functionality for redirects. However, in this case (and a few others), I wanted to redirect to a page with an anchor, and there's no way to do that with `aliases` currently, so I'm using our `redirect_to` functionality to do the redirect:

https://github.com/pulumi/docs/blob/f497736def7d79a90afdd76de6f1656f98ab5412/layouts/partials/head.html#L7-L9

layouts/partials/get-started.html
```diff
@@ -7,33 +7,22 @@ <h2>Get Started with Pulumi</h2>
...
-                <div class="flex align-middle items-center">
-                    <h5 class="flex items-center whitespace-no-wrap">
-                        <span class="border-2 border-blue-500 rounded-full bg-white text-blue-600 w-8 h-8 text-center mr-2 text-green-500">1</span>
-                        Install Pulumi
-                    </h5>
-                </div>
-                <ul class="list-inside list-none pl-8 ml-2">
-                    <li class="leading-snug"><a class="text-blue-700 text-sm" href="{{ relref . "/docs/reference/install" }}">Setup</a></li>
-                </ul>
-            </div>
```
justinvp: Cleaned this up while making changes here (to remove the `/docs/tour` link). We don't really want to send folks to the install page first, we want them to go to the Get Started flow. Originally I kept it 3 items: Get Started, Try Our Tutorials, and Read the Docs, but then the first two steps only had a single link and it didn't look quite right. So collapsed to 2 items.

content/docs/tour/basics-deploying.md
```diff
@@ -1,32 +1,3 @@
 ---
-title: Deploying code
-menu:
-  tour:
-    parent: tour
-    weight: 3
+redirect_to: /docs/reference/cli/pulumi_up/
```
justinvp: I found that we didn't have great conceptual documentation on "deploying" so linked to the command itself, which does provide a good overview. Applied to a few other pages as well (e.g. previewing and destroying).

content/docs/reference/pkg/python/pulumi/_index.md
```diff
@@ -905,7 +905,7 @@ are resolved asynchronously, but they also contain information about the depende
 program.</p>
 <p>Pulumi does not offer direct access to the values contained within Outputs. Instead, you must use the <cite>apply</cite> function
 on the Output class in order to observe the value of an output. See
-<a class="reference external" href="/tour/programs-properties.html">the documentation</a> for more details on this part of the Pulumi programming model.</p>
+<a class="reference external" href="/docs/reference/programming-model/#outputs">the documentation</a> for more details on this part of the Pulumi programming model.</p>
```
chrsmith: From the header, it doesn't look like this file is auto-generated, but it definitely doesn't look like it was written by a human.

Do we need to make this change somewhere else in source code for if this file ever gets regenerated (if that is the case)?

content/docs/tour/basics-destroying.md
```diff
@@ -1,73 +1,3 @@
 ---
-title: Destroying
-menu:
-  tour:
-    parent: tour
-    weight: 7
+redirect_to: https://www.pulumi.com/docs/reference/cli/pulumi_destroy/
```
chrsmith: Nit: We just need `/docs/reference`, right? That would also mean the redirect would work when running locally, etc.

content/docs/tour/basics-projects.md
```diff
@@ -1,100 +1,3 @@
 ---
-title: Creating a project
-menu:
-  tour:
-    parent: tour
-    weight: 2
+redirect_to: https://www.pulumi.com/docs/reference/project/
```
chrsmith: Can this be a relative URL instead?

content/docs/reference/pkg/python/pulumi/_index.md
justinvp: The template (that this is generated from) was also changed in this PR: https://github.com/pulumi/docs/pull/1462/files#diff-d5bf7c984b191b89c451585290f7d7cc

content/docs/tour/basics-destroying.md
justinvp: This should be a relative link. I missed cleaning this up. Will fix.

content/docs/tour/basics-projects.md
justinvp: Yes. This should be a relative link. I missed cleaning this up. Will fix.

content/docs/reference/tutorials/aws/tutorial-rest-api.md
```diff
@@ -146,4 +146,3 @@ From here, you can dive deeper:
         REST APIs that use a document database
   - [Multi-cloud Serverless plus Containers]({{< relref "/docs/quickstart/cloudfx/tutorial-thumbnailer.md" >}}): Create a multi-cloud video
```
bermudezmt:
```suggestion
  - [Multi-cloud Serverless plus Containers]({{< relref "/docs/quickstart/cloudfx/tutorial-thumbnailer" >}}): Create a multi-cloud video
```

When I sat down with @chrsmith before, we started removing `md` in the `relref` links. Also, is my understanding correct that having a trailing slash for these breaks the build? As in, this would trigger a failed build: `relref "/docs/quickstart/cloudfx/tutorial-thumbnailer/"`

content/docs/tour/programs.md
```diff
@@ -1,27 +1,3 @@
 ---
-title: Beyond the Basics
-expanded_url: /docs/tour/programs/
-menu:
-  tour:
-    identifier: programs
-    weight: 2
+redirect_to: /docs/reference/concepts/
```
bermudezmt: More of a question: Are the tour pages now covered under "Getting Started" and "Architecture & Concepts" pages?

content/docs/tour/_index.md
```diff
@@ -1,34 +1,3 @@
 ---
-title: A Tour of Pulumi
```
bermudezmt: I wonder if we should redirect  this to "/docs/quickstart" or "/docs/get-started" if we're changing it to that? Based on my review, the tour seems to have been the precursor of "Get Started" and "Architecture & Concepts".

content/docs/reference/tutorials/aws/tutorial-rest-api.md
justinvp: According to https://gohugo.io/content-management/cross-references/#use-ref-and-relref:

> The single parameter to `ref` is a string with a content `documentname` (e.g., `about.md`) with or without an appended in-document `anchor` (`#who`) without spaces. Hugo is flexible in how we search for documents, so the file suffix may be omitted.

That last sentence: 

> Hugo is flexible in how we search for documents, **so the file suffix may be omitted**.

So it does indeed look like we could remove the `.md` suffix. But I guess adding a trailing `/` will make it so Hugo can't find the content page and fail the build.

content/docs/tour/programs.md
justinvp: The goal of "Getting Started" is to teach simple Pulumi concepts along the way. I'm sure there's more we could do to improve this. The "Architecture & Concepts" would go into all the details.

The problem with redirecting to the Getting Started pages from the tour pages: The main tour index could link to `/docs/quickstart`, but for individual tour pages, which Get Started cloud would you link too?

content/docs/tour/_index.md
justinvp: I went back and forth on this and ended up picking `/docs/reference/concepts/` because that's where various other pages in the tour were largely going. But I could see linking to the Get Started page instead. I don't have a strong pref either way.

justinvp: BTW, if you think it'd be better to link to Get Started, I'd be happy to make the change. Let me know.

bermudezmt: Perhaps that is best? Since the "tour" was mostly a walkthrough with references to the concept docs.

justinvp: https://github.com/pulumi/docs/pull/1465

PR #1466, opened 2019-07-29
Update @pulumi/pulumi docs

CyrusNajmabadi:


content/docs/reference/pkg/nodejs/pulumi/pulumi/_index.md
```diff
@@ -97,6 +97,8 @@ import * as pulumi from "@pulumi/pulumi";
 <li><a href="#isInstance">function isInstance</a></li>
 <li><a href="#isSecretOutput">function isSecretOutput</a></li>
 <li><a href="#liftProperties">function liftProperties</a></li>
+<li><a href="#merge">function merge</a></li>
```
lukehoban: Not blocking this PR - but this function is marked `/** @internal` so should not be showing up here.

CyrusNajmabadi: Found and fixed this.  The `@internal` tag was in the data the doc-generator was examining, just not in a place it expected to find it.

PR #1469, opened 2019-07-29
 Filter out deprecated members. Update AWS docs 

CyrusNajmabadi:


CyrusNajmabadi:
> The changes to the doc generation code look correct to me. It's harder for me to understand if the actual content that got updated matches what what you expect should happen, 

Yup.  Manually validated that it was gone as i expected.

Issue #1394, opened 2019-07-18
Don't hide the install instructions for alternative platforms


jestrjk:
[docs/quickstart/aws/install-pulumi.md](https://www.pulumi.com/docs/quickstart/aws/install-pulumi/)

When you hide linux instructions for people running Windows Subsystem for Linux (WSL) or WSL2 you are hiding the correct instructions for our platform. 

Your page appears to only display the detected OS instructions, and applies display:none to the rest. Please dont.

justinvp:
Thanks for the issue, @jestrjk! We've heard similar feedback and are planning to add platform selection tabs to the Get Started guide (similar to the language tabs) to allow you to select your platform of choice.

In the meantime, if you click the link to the [detailed installation documentation](https://www.pulumi.com/docs/reference/install/) on [docs/quickstart/aws/install-pulumi.md](https://www.pulumi.com/docs/quickstart/aws/install-pulumi/), you can choose your platform from the drop down:

<img width="803" alt="Screen Shot 2019-07-23 at 9 48 09 AM" src="https://user-images.githubusercontent.com/710598/61730899-32459580-ad2f-11e9-9fd6-076be840ff08.png">

joeduffy:
@justinvp Do you want me to take this one?

justinvp:
@joeduffy, sure! Thanks!

PR #1470, opened 2019-07-30
Get Started Follow-ups

bermudezmt:
### Proposed changes

- Index page cleanup
- First pass at informational note styling
- Highlight Pulumi in headings and text for SEO purposes

Screenshot from 2019-07-29 17-33-19

@cnunciato wanted to get your thoughts on the Get Started callouts, especially as it relates to style. I'm also unsure about adding the heading, since "Get Started" is pretty prominent on the left nav.

@justinvp if you think this is a good general direction, can I get a Hugo tip on the best way to access the "index" page for Tutorials to display the same callout? Thank you so much.

### Related issues (optional)

Partially addresses #1173

### DIFF
--- a/content/docs/_index.md
+++ b/content/docs/_index.md
@@ -111,5 +111,5 @@ Describe your resources in code---VMs, networks, databases, containers, Kubernet
     </div>
 </div>
 
-For questions or feedback, reach us through our [community Slack channel](https://slack.pulumi.io),
+For questions or feedback, reach us through our [community Slack channel](https://slack.pulumi.com),
 [GitHub](https://github.com/pulumi), or email [support@pulumi.com](mailto:support@pulumi.com).
--- a/content/docs/_index.md
+++ b/content/docs/_index.md
@@ -11,8 +11,7 @@ menu:
 Pulumi is an <a href="https://github.com/pulumi/pulumi" target="_blank">open source</a> platform for building and
 deploying cloud infrastructure and applications in your favorite language on any cloud.
 
-Describe your resources in code -- VMs, networks, databases, containers, Kubernetes clusters & workloads, serverless
-functions -- and the CLI safely and reliably manages your cloud resources using an infrastructure-as-code approach.
+Describe your resources in code---VMs, networks, databases, containers, Kubernetes clusters and workloads, serverless functions---and the CLI safely and reliably manages your cloud resources using an infrastructure-as-code approach.
 
 <div class="flex justify-center py-6">
     <a class="btn btn-lg mx-1 my-1" href="{{< relref "/docs/quickstart" >}}">GET STARTED</a>
@@ -31,9 +30,9 @@ functions -- and the CLI safely and reliably manages your cloud resources using 
         <p class="text-sm text-gray-600">
             By using <a href="{{< relref "/docs/reference/languages" >}}">general purpose languages</a>
             for infrastructure as code,
-            you get all the benefits of real languages -- IDEs, abstractions and
+            you get all the benefits of real languages&mdash;IDEs, abstractions, and
             reuse thanks to functions, classes, and packages, debugging, testability,
-            and more. The result is far less copy and paste and greater productivity,
+            and more. The result is greater productivity with far less copy and paste,
             and it works the same way no matter which cloud you're targeting.
         </p>
     </div>
@@ -41,9 +40,9 @@ functions -- and the CLI safely and reliably manages your cloud resources using 
         <h3 class="no-anchor">Alternatives</h3>
         <p class="text-sm text-gray-600">
             <a href="{{< relref "/docs/reference/vs" >}}">Other approaches</a> use YAML,
-            JSON, or bespoke DSLs that you need to
-            master -- and convince your team to use. These "languages" fall short of
-            general purpose languages, lacking abstractions and reuse, and reinvent
+            JSON, or bespoke DSLs (Domain-Specific Languages) that you need to
+            master and convince your team to use. These "languages" fall short of
+            general purpose languages, since they lack abstractions and reuse, and reinvent
             familiar concepts like package managers. Worse, these solutions are usually
             unique to every given cloud that you need to target.
         </p>
@@ -112,5 +111,5 @@ functions -- and the CLI safely and reliably manages your cloud resources using 
     </div>
 </div>
 
-For questions or feedback, reach us in our [community Slack channel](https://slack.pulumi.com),
-on [GitHub](https://github.com/pulumi), or by emailing [support@pulumi.com](mailto:support@pulumi.com).
+For questions or feedback, reach us through our [community Slack channel](https://slack.pulumi.io),
+[GitHub](https://github.com/pulumi), or email [support@pulumi.com](mailto:support@pulumi.com).
--- a/content/docs/reference/install.md
+++ b/content/docs/reference/install.md
@@ -16,14 +16,14 @@ This page contains detailed instructions for [installing Pulumi](#install-pulumi
 
 {{< get-started-note >}}
 
-## Available versions
-
-The current stable version of Pulumi is **{{< latest-version >}}**. For a full history of prior versions, including
-release notes and direct download links, visit the <a href="{{< relref "/docs/reference/changelog" >}}">Changelog</a> page.
-Pulumi is packaged as a tarball for Linux and macOS, and as a zip archive for Windows.
-
-
 ## Install Pulumi
+
+<div class="note note-warning" role="alert">
+    <p>
+        <span>Skip version check.</span> If you're in an environment with no internet access, you may skip the pulumi version update check by setting the environment variable
+        <code>PULUMI_SKIP_UPDATE_CHECK</code> to <code>1</code> or <code>true</code>.
+    </p>
+</div>
 
 <script>
     var oses = [ "unknown", "linux", "macos", "windows" ];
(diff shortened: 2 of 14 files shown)


layouts/docs/single.html
```diff
@@ -16,8 +16,9 @@
                 {{ partial "docs/toc.html" . }}
             </div>
 
-            <div class="md:w-7/12">
+            <div class="md:w-8/12">
                 <h1>{{ .Title }}</h1>
+                {{ partial "docs/gs-note.html" . }}
```
justinvp: Instead of doing it like this, and having to conditionally detect the path in `gs-note.html` to determine if it should be rendered or not, I think it'd be better as a shortcode. https://gohugo.io/templates/shortcode-templates/

Basically, move `gs-note.html` into `layouts/shortcodes` and remove the `{{ if ... }}`

Then at the top of `docs/reference/install.md` and `docs/reference/tutorials/_index.md` add the following:

```
{{< gs-note >}}
```

Consider renaming the shortcode from `gs-note.html` to `get-started-note.html`.

content/docs/reference/install.md
```diff
@@ -12,9 +12,16 @@ NOTE: To update this page with a new binary release, do the following:
...
 
-This page contains detailed instructions on installing Pulumi.
+This page contains detailed instructions for [downloading](#download) and [installing Pulumi](#install-pulumi) on your machine.
 
-For a streamlined walkthrough of getting started with Pulumi, including installing Pulumi and prerequisites, see the [Get Started]({{< relref "/docs/quickstart" >}}) guide.
+## Download
+
+The current stable version of Pulumi is **{{< latest-version >}}**. For a full history of prior versions, including
+release notes and direct download links, visit the <a href="{{< relref "/docs/reference/changelog" >}}">Changelog</a> page.
+Pulumi is packaged as a tarball for Linux and macOS, and as a zip archive for Windows.
```
justinvp: The downside of moving this Download section to the top of the page is that it implies that this is what users should be doing. But most users should not be downloading these and installing manually (download/manual-install is a more advanced scenario). For macOS, the majority of users should be using Homebrew, for Linux the installation script, and for Windows the installation script (though, the Windows installation script experience isn't great and may trigger anti-virus -- I believe Paul was looking into Chocolately, which might become our preferred approach there when it's available).

So I worry a little that by having a Download section at the top, folks are going to first try to download a version for their platform, without having read the rest of this page, and then be left scratching their heads with how to install the binary manually. When instead, they should have used one of the preferred methods for their platform.

content/docs/_index.md
```diff
@@ -112,5 +111,5 @@ functions -- and the CLI safely and reliably manages your cloud resources using
     </div>
 </div>
 
-For questions or feedback, reach us in our [community Slack channel](https://slack.pulumi.com),
-on [GitHub](https://github.com/pulumi), or by emailing [support@pulumi.com](mailto:support@pulumi.com).
+For questions or feedback, reach us through our [community Slack channel](https://slack.pulumi.io),
```
justinvp: The link should remain https://slack.pulumi.com not https://slack.pulumi.io

content/docs/reference/install.md
```diff
@@ -127,7 +136,7 @@ The installer script can be rerun to subsequently install new updates.
 
 <div id="linux_installation">
 {{% md %}}
-## Linux
+### Installation script
```
justinvp: Should "script" be capitalized or not? It's upper case in the macOS section above but lower case in the sections below. Same question about "installation".

layouts/docs/single.html
bermudezmt: Yeah, it occurred to me last night that I should have used a shortcode for this instead. And noted on the rename. :)

content/docs/reference/install.md
bermudezmt: I like referring to [Google's style guide](https://developers.google.com/style/capitalization) and have gotten used to sentence case, but I do notice we have mostly used title case previously. We can have a further chat around this, and I can revert to using title case for now.

cnunciato:
I like it! I might suggest a few minor tweaks:

image

image

If these look okay to you, let me know and I'll push a commit onto your PR.

bermudezmt:
It looks perfect, @cnunciato . Please push a commit. :smile:

content/docs/reference/install.md
justinvp: Instead of this section, what do you think of just adding a sentence to the Manual Installation for each of the platforms?

For macOS, something like:

```diff
- 1. Download [Pulumi {{< latest-version >}} for macOS](https://get.pulumi.com/releases/sdk/pulumi-v{{< latest-version >}}-darwin-x64.tar.gz).
+ 1. Download [Pulumi {{< latest-version >}} for macOS](https://get.pulumi.com/releases/sdk/pulumi-v{{< latest-version >}}-darwin-x64.tar.gz). Prior versions and release notes are available [here]({{< relref "/docs/reference/changelog" >}}).
```

And do something similar for Linux and Windows.

bermudezmt: Ah yes, we can do something like that for now. But as we release more major versions, we might want to consider adding a "Release Notes" link to the left nav, or a "View Release Notes" button at the top of the Download and Install page.

Are we ok with leaving this scenario vague though?
Screenshot from 2019-07-31 09-05-25

Does that get triggered often?

bermudezmt: I have one more question, @justinvp : Did we mean to add `either` to the following instruction?
> 2. Unzip the tarball and either copy the binaries in the `pulumi` directory on your `$PATH`.

We have this sentence under "Manual Installation" for macOs and Linux.

justinvp: > Are we ok with leaving this scenario vague though?

Which scenario?

> Does that get triggered often?

Does what get trigged often? The "(choose one)" option? I doubt it's ever triggered, unless someone selects it manually. Next milestone I am going to change the install step of the Get Started guide to show tabs for Linux, macOS, and Windows (#1394). What do you think of switching this to be tabs as well (instead of the drop down)?

bermudezmt: Yes. Will the tabs have the installation options then verify and uninstall will be outside of the tabbed content? I think tabs might look cleaner.

justinvp: > Did we mean to add `either` to the following instruction?

Probably not. It should probably be something like:

> 2. Unzip the tarball and copy the binaries to a `pulumi` directory on your `$PATH`.

But feel free to rewrite it so it actually makes sense 😉

> Will the tabs have the installation options then verify and uninstall will be outside of the tabbed content?

I think so.

content/docs/reference/install.md
justinvp: Sounds good. I don't have a preference either way, as long as we're consistent on a page (and across the site).

bermudezmt: I reverted for now and will apply the site-wide change later. As per [this article](https://www.herosmyth.com/article/headline-capitalization-ux-title-case-vs-sentence-case), sentence case is apparently more "casual" and is geared towards younger people. ;) So, I think I addressed all your feedback. I'd keep this branch for more "Get Started" follow-ups but  perhaps we can have this PR merged if all the changes look good now?

layouts/docs/single.html
```diff
@@ -16,7 +16,7 @@
                 {{ partial "docs/toc.html" . }}
             </div>
 
-            <div class="md:w-7/12">
+            <div class="md:w-8/12">
```
justinvp: Why'd this change? I think we want this to remain 7.

content/docs/reference/install.md
```diff
@@ -16,15 +16,15 @@ This page contains detailed instructions for [installing Pulumi](#install-pulumi
...
-## Available versions
-
-The current stable version of Pulumi is **{{< latest-version >}}**. For a full history of prior versions, including
-release notes and direct download links, visit the <a href="{{< relref "/docs/reference/changelog" >}}">Changelog</a> page.
-Pulumi is packaged as a tarball for Linux and macOS, and as a zip archive for Windows.
-
-
 ## Install Pulumi
 
+<div class="note note-warning" role="alert">
```
bermudezmt: @justinvp Thought I'd address https://github.com/pulumi/docs/issues/1473 as well while we're working on the installation page. Let me know your thoughts.

layouts/docs/single.html
bermudezmt: Yes, sorry. Was playing around and thought I'd checked out that file but somehow it got committed.

content/docs/reference/install.md
justinvp: If it were me, I would do this in its own PR. It's also not clear to me that we want this note here at the top of the page as this is talking about a version check before even discussing installing Pulumi. Perhaps we need another "Version Checks" section (or better name) after the install instructions sections, which could discuss how Pulumi will emit a warning if the current running version is out-of-date, and in that section add a note about the environment variable. We should also consider finding a place on another page where we could document all environment variables used by the CLI.

bermudezmt: Sounds good. Let's revisit the environment variable doc work at next week's sync meeting. I retained the additional styles but removed the note -- will work on #1473 in a separate PR.

content/docs/reference/install.md
```diff
@@ -104,54 +108,59 @@ Subsequent updates can be installed in the usual way:
...
 
-This will install the `pulumi` CLI to `~/.pulumi/bin` and add it to your path.
+This will install the `pulumi` CLI to `~/.pulumi/bin` and add it to your path. When it can't automatically add `pulumi` to your path, you will be prompted to add it manually.
+See [How to permanently set $PATH on Unix](https://stackoverflow.com/questions/14637979/how-to-permanently-set-path-on-linux-unix) for guidance.
+
 
 The installer script can be rerun to subsequently install new updates.
 
-### (Alternative) Manual Installation
+### Manual Installation
```
justinvp: One minor nitpick I just noticed when looking over everything again:

Now that we're removing "(Alternative)" from the heading (which I agree with), should we add a sentence below all of the "Manual Installation" headings that says something like:

> Alternatively, Pulumi can be installed manually:

?

Similar to the sentence under the "Installation Script" headings?

> Alternatively, our installation script can be run:

Issue #1127, opened 2019-06-06
Page titles for Python API docs are all empty


clstokes:
I'm assuming these are auto-generated for our API docs, but they're empty or missing for Python.

https://pulumi.io/reference/pkg/python/pulumi_azure/containerservice/
![image](https://user-images.githubusercontent.com/293763/59061270-ed859e00-8857-11e9-8152-902523e9440f.png)

https://pulumi.io/reference/pkg/python/pulumi_azuread/
![image](https://user-images.githubusercontent.com/293763/59061291-fa09f680-8857-11e9-9eed-cf4ac0c15a4d.png)

PR #1472, opened 2019-07-30
Fix some titles and meta descriptions

joeduffy:
This change fixes up some titles and meta descriptions, in hopes
that it'll have a positive impact on our SEO (and SERP summary).

- Ensure all pages have titles and meaningful meta descriptions.

- Error out if pages lack a meta description. Also add the ability
  to write `nosummary: true` if it's intentionally omitted.

- I'd like to error out if pages lack a title too. Unfortunately,
  the Python API docs are missing titles (pulumi/docs#1127), so this
  check is not currently enabled (easy to uncomment out later).

I also made the possibly-controversial change to stop emitting an
automatic `- Pulumi` at the end of page titles. This

- Leads to illy things -- like `About Pulumi - Pulumi`.

- Might make it more difficult for Google to extract a good title for
  our SERP summary; is that why "Documentation" is currently
  awkwardly labeled simply "Pulumi" on Google at the moment?

- Probably has a net negative effect on things like blog articles,
  which we would like to index naturally based on the content.

I propose that anywhere we want Pulumi in the title, we manually add it.

### DIFF
--- a/content/about/_index.md
+++ b/content/about/_index.md
@@ -1,6 +1,6 @@
 ---
 title: About
-summary: Pulumi is a Seattle-based company that enables teams to create, deploy, and manage modern cloud applications and infrastructure.
+meta_desc: Pulumi is a Seattle-based company that enables teams to create, deploy, and manage modern cloud applications and infrastructure.
 type: page
 layout: about
 menu:
--- a/content/about/_index.md
+++ b/content/about/_index.md
@@ -1,5 +1,6 @@
 ---
 title: About
+summary: Pulumi is a Seattle-based company that enables teams to create, deploy, and manage modern cloud applications and infrastructure.
 type: page
 layout: about
 menu:
--- a/config/_default/config.toml
+++ b/config/_default/config.toml
@@ -38,4 +38,4 @@ enableRobotsTXT = true
     canonicalURL = "https://www.pulumi.com"
     repositoryURL = "https://github.com/pulumi/docs"
     repositoryBranch = "master"
-    summary = "Pulumi's open source infrastructure as code SDK enables you and your team to create, deploy, and manage infrastructure on any cloud, reliably and productively, using your favorite languages."
+    meta_desc = "Pulumi's open source infrastructure as code SDK enables you and your team to create, deploy, and manage infrastructure on any cloud, reliably and productively, using your favorite languages."
--- a/config/_default/config.toml
+++ b/config/_default/config.toml
@@ -1,5 +1,5 @@
 languageCode = "en-us"
-title = "Pulumi"
+title = "Pulumi - Modern Infrastructure as Code"
 
 disableKinds = ["category", "taxonomyTerm"]
 sectionPagesMenu = "main"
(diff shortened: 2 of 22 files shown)


content/about/_index.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: About
+summary: Pulumi is a Seattle-based company that enables teams to create, deploy, and manage modern cloud applications and infrastructure.
```
justinvp: We should use `meta_desc` for these instead of `summary` and remove our use of `summary` in the small number of places it's used. `summary` is only used in a handful of blog posts and was a holdover from the initial blog migration when Chris was originally trying to collapse the summary and meta description into a single thing -- but we've since gone back to allowing separate meta descriptions and summaries (and the summaries are now obtained via `.Summary`, Hugo's [Content Summaries](https://gohugo.io/content-management/summaries/)).

justinvp:
While we're on the topic of `.Summary`, yesterday I had mentioned to @erinhyperplane that using it for the meta description leads to some unusual meta descriptions for pages. For example:

https://www.pulumi.com/docs/reference/install/

```
        <meta name="description" content="This page contains detailed instructions on installing Pulumi.
For a streamlined walkthrough of getting started with Pulumi, including installing Pulumi and prerequisites, see the Get Started guide.
 var oses = [ &#34;unknown&#34;, &#34;linux&#34;, &#34;macos&#34;, &#34;windows&#34; ]; function showInstall(os) { // Show the div and select it in the dropdown. var e = document.getElementById(os &#43; &#34;_installation&#34;); if (e) { e.style.display = &#34;block&#34;; var s = document.getElementById(&#34;os&#34;); if (s) { for (var i = 0; i Operating system: (choose one) Linux macOS Windows">
```

She was wondering if we should stop using `.Summary` for meta description, and let people write `meta_desc` so they can write whatever they want in the first paragraph (or above `<!--more-->`) and not have to worry about having meta descriptions like this. She was going to open an issue with a proposal this week when she started diving in on improving our meta descriptions.

joeduffy:
Yeah, I was just looking at those summaries. All of them are terrible, with few exceptions. (Setting `summaryLength = 30` in the config seems to make them slightly more appealing, but not by much.) Let me see how bad it is to just plow through and tidy this up.

joeduffy:
> She was going to open an issue with a proposal this week when she started diving in on improving our meta descriptions.

Ok, I am admitting defeat 😉 This sounds like a better course of action. I'll clean up the PR based on comments above and leave it to a deeper exercise to figure out what to do with the rest.

joeduffy:
By the way, if it wasn't evident why I was focused on only a smattering of pages, it began with me wondering about Google's SERP/summary for us:

image

I still wish I knew how and why Google is selecting those pages -- as they aren't what I would choose! -- and it will be interesting to see if these changes help with the nested "Pulumi".

erinhyperplane:
it is a myth to me how google are showing those cards - it looks different on my machine, and it looks different from just a week before (it used to have docs, pricing, why pulumi, etc.)

image

PR #1478, opened 2019-07-30
Address some website performance issues

### LINKED ISSUE #1402: First first batch of website performance issues

cnunciato:
This change compresses a few images that appear on the Pulumi home page, and increases the CloudFront cache TTLs to five minutes by default, one hour for files that will rarely change (like fonts, icons, vendor images), and one week for files that should never change, like the CSS and JS files we bundle (and name with a content-based hash) at build-time.

Fixes #1402.


infrastructure/index.ts
```diff
@@ -106,26 +130,47 @@ const distributionArgs: aws.cloudfront.DistributionArgs = {
...
+    },
 
-        // S3 doesn't need take any of these values into account when serving content.
-        forwardedValues: {
-            cookies: {
-                forward: "none",
-            },
-            queryString: false,
+    orderedCacheBehaviors: [
+        {
```
cnunciato: These `pathPattern` blocks are repeated because Amazon doesn't seem to support expressing multiple paths, e.g., as a real regular expression; they just provide for simple wildcards like `?` and `*`.

PR #1482, opened 2019-07-31
Add body of blog post to RSS feed

### LINKED ISSUE #1349: Add body of blog post to RSS feed.
The current RSS Feed doesn't offer more than the description. It would be nice to have the body of the post included.

cnunciato:
This change adds the full text of blog posts to the RSS feed. (It also removes some invisible characters from a few posts, which prevented them from being parsed by feed readers.) 

It's worth noting that this brings the size of the feed from 152K to a little over 2MB (!), so we might want to consider paging, capping or breaking it up at some point, probably soonish.

Fixes #1349.


cnunciato:
Sounds good -- how about we compromise and try 20. 😄 (That takes us back a couple of months.)

PR #1483, opened 2019-07-31
Change get-started button labels to “Get Started”

cnunciato:


layouts/index.html
```diff
@@ -8,7 +8,7 @@ <h1>Modern Infrastructure as Code</h1>
                     favorite language.
                 </p>
                 <div class="header-hero-actions">
-                    <a class="btn btn-lg" href="{{ relref . "/docs/quickstart" }}">Try Pulumi for Free</a>
```
joeduffy: Should we nix the Why Pulumi button and just have a single CTA? Cc @justinvp

justinvp: Yeah, let's nix it.

PR #1488, opened 2019-07-31
Add preview disclaimer to Pulumi Cloud

bermudezmt:
Partially addresses https://github.com/pulumi/pulumi-cloud/issues/737.

image

Is this along the lines of the change you were looking for, @CyrusNajmabadi ? I haven't removed the original paragraph yet, found at https://www.pulumi.com/serverless/. Once we finalize, I'll apply the final note to the other www and docs pages.

### DIFF
--- a/layouts/topics/serverless.html
+++ b/layouts/topics/serverless.html
@@ -99,14 +99,23 @@
                 Pulumi makes it simple to interact with serverless services available from
                 the major cloud vendors, and with Kubernetes.
             </p>
+            <h3>Pulumi Cloud Framework (Preview)</h3>
             <p>
+                Pulumi Cloud Framework is our multi-cloud framework for building modern container and serverless cloud applications. <code>@pulumi/cloud</code> is a preview library demonstrating how to build Pulumi components that can abstract over and target multiple clouds.
+            </p>
+            <p>
+            For developers targeting a single cloud platform like AWS, Azure or GCP, we recommend using the <a href="{{ relref . "/docs/reference/pkg/nodejs/pulumi/aws" }}"><code>@pulumi/aws</code></a>, <a href="{{ relref . "/docs/reference/pkg/nodejs/pulumi/azure" }}"><code>@pulumi/azure</code></a>, and <a href="{{ relref . "/docs/reference/pkg/nodejs/pulumi/gcp" }}"><code>@pulumi/gcp</code></a> packages respectively. These packages give you full access to the breadth of the platform's capabilities and comes with many abstractions to make developing against that platform easier.
+
+            </p>
+
+            <!--p>
                 The <code>@pulumi/cloud</code> library provides very high level, multi
                 cloud access to building blocks of serverless applications, while the
                 specialized <code>@pulumi/cloud-aws</code>,
                 <code>@pulumi/cloud-azure</code>, <code>@pulumi/cloud-gcp</code>, and
                 <code>@pulumi/kubernetes</code> libraries offer fine grained control of
                 the individual cloud vendor services.
-            </p>
+            </p-->
         </div>
     </section>
     <section id="code" class="py-16 bg-gray-200 px-4 md:px-0">
--- a/layouts/topics/serverless.html
+++ b/layouts/topics/serverless.html
@@ -101,21 +101,20 @@
             </p>
             <h3>Pulumi Cloud Framework (Preview)</h3>
             <p>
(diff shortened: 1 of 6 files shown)


layouts/topics/serverless.html
```diff
@@ -99,14 +99,23 @@ <h2>Building Blocks of Serverless Programming</h2>
                 Pulumi makes it simple to interact with serverless services available from
                 the major cloud vendors, and with Kubernetes.
             </p>
+            <h3>Pulumi Cloud Framework (Preview)</h3>
             <p>
+                Pulumi Cloud Framework is our multi-cloud framework for building modern container and serverless cloud applications. <code>@pulumi/cloud</code> is a preview library demonstrating how to build Pulumi components that can abstract over and target multiple clouds.
+            </p>
+            <p>
+            For developers targeting a single cloud platform like AWS, Azure or GCP, we recommend using the <a href="{{ relref . "/docs/reference/pkg/nodejs/pulumi/aws" }}"><code>@pulumi/aws</code></a>, <a href="{{ relref . "/docs/reference/pkg/nodejs/pulumi/azure" }}"><code>@pulumi/azure</code></a>, and <a href="{{ relref . "/docs/reference/pkg/nodejs/pulumi/gcp" }}"><code>@pulumi/gcp</code></a> packages respectively. These packages give you full access to the breadth of the platform's capabilities and comes with many abstractions to make developing against that platform easier.
```
CyrusNajmabadi: "and come"

CyrusNajmabadi: also, consider wrapping since this is hard  to read online :)

bermudezmt: Oops, I copied and pasted from the README in `pulumi-cloud` -- also fixed it there via https://github.com/pulumi/pulumi-cloud/pull/741/files. To make sure I understand, do you mean line wrapping or adding styling to make the note stand out more? Thanks!

CyrusNajmabadi: I just literally mean wrapping in the html (no actual tags).  On github this is just one very very very long line :)

bermudezmt: Hopefully the line wrap helps now. :) 

Pulumi Cloud Index Page

image

Pulumi Cloud Tutorial

image


I've also added the same update to www.pulumi.com/containers. This doesn't include updates to www.pulumi.com/docs/reference/pkg/nodejs/pulumi/cloud/ and www.pulumi.com/docs/reference/pkg/nodejs/pulumi/cloud-aws/ yet but I can do those in a follow-up PR.

CyrusNajmabadi:
FWIW, and luke to confirm, for 1.0 we probably want to take those `pulumi/cloud` tutorials and at least have the `awsx` equivalent that we can point to.   This shouldn't be too hard to do and i can help you out if you need with getting the right code for the containers examples.

lukehoban:
Are these quickstarts linked from the TOC?  I don't see them as an end-user.  

I think to first approximation these are deprecated content.  Assuming we see very low traffic to these from inbound links - we could consider just fully removing them.

bermudezmt:
IIRC, `/docs/quickstart/cloudfx/` is one of the quickstart directories we intentionally left out of the TOC (left navigation). You could access the landing page at  https://www.pulumi.com/docs/quickstart/cloudfx/.

@lukehoban can you please confirm if we're going to keep these in for now or remove them entirely? I can open an issue around updating the tutorials to their non-cloud equivalent and work with @CyrusNajmabadi on that.

joeduffy:
I think it's fine to keep them as-is, especially now that it's marked preview.

If we deprecated/removed the package, I could see an argument for removing them, but we've explicitly decided to keep the package and mark it as preview (at least based on my understanding).

More than anything, I vote for plowing ahead and making forward progress.

justinvp:
Keeping the content sounds fine, but I don't think it makes sense under `/docs/quickstart` anymore because similar content was already moved out of `/docs/quickstart`. For example, the overviews of the clouds and setup pages were moved to be under "Supported Clouds" and the tutorials for each cloud were moved to be under "Tutorials".

So:
- Move the Cloud FX page to either the  "Supported Clouds" (although, it doesn't make as much sense to me there because it isn't a cloud, it's a library that works with multiple clouds), or perhaps better, move the content into the API docs for the package.
- Move the tutorials to be under Tutorials.

lukehoban:
I believe we can close this out based on the conversation above.  We may still come back and move some of these pages - but they aren't in TOC so that isn't critical - and it would be a different change than what's currently in this PR anyway.

PR #1494, opened 2019-08-01
Special case where the docs fine the readme for pulumi-cloud.

CyrusNajmabadi:
Note: i looked into fixing up the actual script that runs over all the repos to put the special case there.  but it ended up being really complicated to make that work out since there's a lot of intricacies about how we pull in readmes and nested-readmes.  Trying to make that all work out was just getting really painful. So this just seemed like a a simple and easy 1.0 thing to do.


ellismg:
Sorry - Trying to understand the intention here.  What exactly are we trying to accomplish?  Is this just so we don't have to have a README.md here? https://github.com/pulumi/pulumi-cloud/tree/master/api (because we'd rather have that content at top level because it looks nicer for the landing page on GitHub)?

CyrusNajmabadi:
Yup.  That's correct.  I've documented that this is a cheap hack.

PR #1502, opened 2019-08-02
More Python Programming Model Coverage

### LINKED ISSUE #1326: Provide in-language examples for Resource Options
Our resource options are really powerful features, but not well documented. Just like our other programming model examples (e.g. [.apply](https://www.pulumi.com/docs/reference/programming-model/#apply)) we should provide language-specific examples of their usage. https://www.pulumi.com/docs/reference/programming-model/#resourceoptions

### LINKED ISSUE #1375: Dynamic Provider outputs need to be passed with Inputs
[docs/reference/programming-model.md](https://www.pulumi.com/docs/reference/programming-model/)

See https://github.com/pulumi/pulumi/issues/2931 for more information. In [this section](https://www.pulumi.com/docs/reference/programming-model/#dynamic-resource-outputs) the code should be changed to
```ts
export class MyResource extends pulumi.dynamic.Resource {
    public readonly myStringOutput: pulumi.Output<string>;
    public readonly myNumberOutput: pulumi.Output<number>;

    constructor(name: string, props: MyResourceInputs, opts?: pulumi.CustomResourceOptions) {
        super(myprovider, name, {...props, myStringOutput: undefined, myNumberOutput: undefined}, opts); 
    }
}
```

### LINKED ISSUE #1403: Update docs for Python dynamic providers
https://github.com/pulumi/pulumi/pull/2900

### LINKED ISSUE #1480: Add Python example for StackReference
### Problem description

The `StackReference` example at https://www.pulumi.com/docs/reference/organizing-stacks-projects/#inter-stack-dependencies shows only a typescript example. Now that StackReference is supported in Python too (https://github.com/pulumi/pulumi/pull/2786) we should add our language chooser and show examples in each of the supported languages. 

https://github.com/pulumi/pulumi/blob/852a8f9dc9baac934e380ab5b8b0c39b80cdc9de/tests/integration/stack_reference/python/__main__.py can provide the starting point for the Python example.

lukehoban:
Add docs for:
- Python Dynamic Providers
- Python Config.require_object
- Python Output.concat
- Python StackReference
- Code examples for all resource options

Also fix some documentation on dynamic provider outputs.

Fixes #1403.
Fixes #1375.
Fixes #1480.
Fixes #1326.

### DIFF
--- a/content/docs/reference/organizing-stacks-projects.md
+++ b/content/docs/reference/organizing-stacks-projects.md
@@ -96,8 +96,24 @@ and testing. In that case, we'll have six distinct stacks, that pair up together
 The way Pulumi programs communicate information for external consumption is by using stack exports. For instance,
 our infrastructure stack might export the Kubernetes configuration information needed to deploy into a cluster:
 
+{{< langchoose >}}
+
+```javascript
+exports.kubeConfig = ... a cluster's output property ...;
+```
+
 ```typescript
 export const kubeConfig = ... a cluster's output property ...;
+```
+
+```python
+pulumi.export("kubeConfig", ... a cluster's output property ...)
+```
+
+```go
+// StackReference is not supported in Go currently.
+//
+// See https://github.com/pulumi/pulumi/issues/1614.
 ```
 
 The challenge here is now our services project needs to ingest this output during deployment so that it can
@@ -105,13 +121,40 @@ connect to the Kubernetes cluster provisioned in its respective environment.
 
 The Pulumi programming model offers a way to do this with its `StackReference` resource type. For example:
 
+{{< langchoose >}}
+
+```javascript
+const k8s = require("@pulumi/kubernetes");
+const pulumi = require("@pulumi/pulumi");
+const env = pulumi.getStack();
+const infra = new pulumi.StackReference(`acmecorp/infra/${env}`);
+const provider = new k8s.Provider("k8s", { kubeConfig: infra.getOutput("kubeConfig") });
+const service = new k8s.core.v1.Service(..., { provider: provider });
+```
+
 ```typescript
 import * as k8s from "@pulumi/kubernetes";
 import * as pulumi from "@pulumi/pulumi";
 const env = pulumi.getStack();
 const infra = new pulumi.StackReference(`acmecorp/infra/${env}`);
 const provider = new k8s.Provider("k8s", { kubeConfig: infra.getOutput("kubeConfig") });
-const service = new k8s.v1.core.Service(..., { provider: provider });
+const service = new k8s.core.v1.Service(..., { provider: provider });
+```
+
+```python
+from pulumi import get_stack, ResourceOptions, StackReference
+from pulumi_kubernetes import Provider, core
+
+env = get_stack()
+infra = StackReference(f"acmecorp/infra/{env}")
+provider = Provider("k8s", { "kubeconfig": infra.get_output("kubeConfig") })
+service = core.v1.Service(..., ResourceOptions(provider=provider))
+```
+
+```go
+// StackReference is not supported in Go currently.
+//
+// See https://github.com/pulumi/pulumi/issues/1614.
 ```
 
 The `StackReference` constructor takes as input a string of the form `<organization>/<project>/<stack>`, and lets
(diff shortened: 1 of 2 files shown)


content/docs/reference/organizing-stacks-projects.md
```diff
@@ -96,22 +96,65 @@ and testing. In that case, we'll have six distinct stacks, that pair up together
 The way Pulumi programs communicate information for external consumption is by using stack exports. For instance,
```
ericrudder:
```suggestion
The way Pulumi programs communicate information for external consumption is by using stack exports. For example,
```

content/docs/reference/organizing-stacks-projects.md
```diff
@@ -96,22 +96,65 @@ and testing. In that case, we'll have six distinct stacks, that pair up together
...
 ```typescript
 export const kubeConfig = ... a cluster's output property ...;
 ```
 
+```python
+pulumi.export("kubeConfig", ... a cluster's output property ...)
+```
+
+```go
+// StackReference is not supported in Go currently.
```
ericrudder:
```suggestion
// StackReference is currently not supported in Go.
```

content/docs/reference/organizing-stacks-projects.md
```diff
@@ -96,22 +96,65 @@ and testing. In that case, we'll have six distinct stacks, that pair up together
...
+//
+// See https://github.com/pulumi/pulumi/issues/1614.
+```
+
 The challenge here is now our services project needs to ingest this output during deployment so that it can
```
ericrudder:
```suggestion
The challenge here is that our services project needs to ingest this output during deployment so that it can
```

content/docs/reference/programming-model.md
```diff
@@ -1055,14 +1276,20 @@ class MyResource extends pulumi.dynamic.Resource {
...
+We can now create instances of the new `MyResource` resource kind in our program with `new MyResource("name", args)`.  When we do so, if Pulumi determines the resource has not yet been created, it will call the `create` method on the resource provider interface.  If another Pulumi deployment happens and the resource already exists, Pulumi will call the `diff` method to determine whether a change can be made in place or whether a replacement is needed.  If a replacement is needed, Pulumi will call `create` for the new resource and then `delete` for the old resource.  If no repacement is needed, Pulumi will call `update`.  In all cases, before doing anything else, Pulumi will call the `check` method with the resource arguments to give the provider a chance to validate that the arguments are valid.  And finally, if Pulumi needs to read an existing resource without managing it directly, it will call `read`.
```
ericrudder:
```suggestion
We can now create instances of the new `MyResource` resource kind in our program with `new MyResource("name", args)`.  When doing so, if Pulumi determines the resource has not yet been created, it will call the `create` method on the resource provider interface.  If another Pulumi deployment happens and the resource already exists, Pulumi will call the `diff` method to determine whether a change can be made in place or whether a replacement is needed.  If a replacement is needed, Pulumi will call `create` for the new resource and then `delete` for the old resource.  If no repacement is needed, Pulumi will call `update`.  In all cases, Pulumi first call the `check` method with the resource arguments to give the provider a chance to validate that the arguments are valid.  And finally, if Pulumi needs to read an existing resource without managing it directly, it will call `read`.
```

content/docs/reference/organizing-stacks-projects.md
lukehoban: JFWIW - This is copy pasted - I suspect there are lots more places we'll want to fix this up if it's worth changing.

content/docs/reference/organizing-stacks-projects.md
lukehoban: Note that many of the changes you are proposing here are not in text that is part of this PR.  Editorial pass over the full docs definitely appreciated - but might be easiest (and also help with consistency) to just do it on the raw doc content instead of on the bits and pieces you happen to see in PRs :-).

ericrudder:
i think i only did this "once or twice" .. i tried *not* to get into a full
pass of this, since i had other stuff .. however, i did read the change in
the "broader context" so i knew what they heck i was looking at and
probably couldn't resist one or two wording comments.  if i did this more
than that, my apologies; i tried to stick to new stuff

PR #1507, opened 2019-08-04
WIP - Filter out BLC log spam

chrsmith:
Trying to see if this works for filtering out blog log spam from `https://github.com/stevenvachon/broken-link-checker/issues/133`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,7 @@ test:
 	#     - Our LinkedIn page, for some reason, returns an HTTP error (despite being valid)
 	#     - Our Visual Studio Marketplace link for the Azure Pipelines task extension,
 	#       although valid and publicly available, is reported as a broken link.
+	# The odd grepping is to reduce log spam. https://github.com/stevenvachon/broken-link-checker/issues/133
 	./node_modules/.bin/blc http://localhost:1313 --recursive --follow \
 		--exclude "/docs/reference/pkg" \
 		--exclude "/docs/reference/changelog" \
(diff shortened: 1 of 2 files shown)


justinvp:
It looks like the next version of the BLC (v0.8.0) is going to have a `verbosity` flag where you can set it to only show broken links. https://github.com/stevenvachon/broken-link-checker/blob/512431eb48f6bebd837ed008a1df79b74094d76d/lib/cli.js#L505-L511. But it hasn't been released/published yet.

justinvp:
I keep waiting for the BLC v0.8.0 to be released on npm... In the meantime, we could just start using  `v0.8.0-alpha` (which is prob fine for our use; yolo) via `yarn add https://github.com/stevenvachon/broken-link-checker#60ffc06c8d5b790c4ad160eddd8f7059454af0c9`. Then we can add `--verbosity 0` to only output broken links.

I started working on a PR to do that, but v0.8.0 drops support for Node versions below 12.0 and I'm still running Node v8.11.2 on this machine and don't want to mess with that right now.

justinvp:
I want to look into possibly switching us over to https://github.com/wjdp/htmltest

chrsmith:
@justinvp re: switching from `bloc` to `htmltest`, that sounds like a good improvement. Since assuming it does as good of a job for broken link detection, we'd be getting more coverage as well. /cc @zchase , who might be interested as well as our linting/validation relates to SEO.

chrsmith:
As for as keeping the repo tidy and what not, I'll go ahead and close this PR out though.

PR #1526, opened 2019-08-09
serializeFunction was moved to pulumi.runtime.closure

redbaron:

### DIFF
--- a/content/docs/reference/programming-model.md
+++ b/content/docs/reference/programming-model.md
@@ -81,7 +81,7 @@ This package also provides the following helpers:
 
 - [pulumi.getStack] to get information about the current stack
 - [pulumi.log] for logging deployment information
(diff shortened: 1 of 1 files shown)


lukehoban:
This function is indeed available at `pulumi.runtime.serializeFunction` - was there something else that indicated to you that it was not?

lukehoban:
Oh - I see - the API docs are showing up in https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/runtime/closure/#serializeFunction.  That is a bug in the doc generator.

lukehoban:
Opened https://github.com/pulumi/docs/issues/1527.

lukehoban:
Closing this one out - as the proper fix will need to be to address #1526.

PR #1538, opened 2019-08-11
Fix Hugo warnings

justinvp:
Fixes `WARN 2019/08/11 14:56:21 make non-relative ref/relref page reference(s) in page %q absolute, e.g. {{< ref "/blog/my-post.md" >}}`

### DIFF
--- a/content/docs/console/continuous-delivery/_index.md
+++ b/content/docs/console/continuous-delivery/_index.md
@@ -42,7 +42,7 @@ commits to a particular branch. This relates to, but is independent from, contin
 a subset of these code changes after specific gates have been passed (certain tests passing, approval, and so on).
 
 Pulumi can also bridge results from your CI/CD system with GitHub, for example surfacing the results of stack
-updates on GitHub pull requests. See [Pulumi GitHub App]({{< relref "github.md" >}}) for more information.
+updates on GitHub pull requests. See [Pulumi GitHub App]({{< relref "github-app.md" >}}) for more information.
 
 ## Managing Environments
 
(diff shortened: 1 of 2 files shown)


justinvp:
Travis failure:

```
Getting links from: http://localhost:1313/docs/console/accounts/saml/
1728-├─BROKEN─ https://www.samltool.com/prettyprint.php (BLC_UNKNOWN)
1729├─BROKEN─ https://www.samltool.com/prettyprint.php (HTTP_undefined)
```

The certificate for this site expired 2 days ago. I thought about adding an exclusion to the checker for this URL, but if the cert isn't updated, we shouldn't be linking to this site.

Screen Shot 2019-08-12 at 6 34 49 AM

justinvp:
> Is there a flag or something you can find to get a listing of the specific pages that introduced the warning?

I contributed https://github.com/gohugoio/hugo/pull/6061 back in June and used a private build to see the pages. But looks like version v0.56.0 and later includes my fix.

> I don't quite understand why this warning is happening.

The warning happens when the path passed to `rel` or `relref` doesn't start with `/` (i.e. isn't absolute) and the page can't be found relative to the current page, but _is_ found somewhere else on the site. For example, if I had `{{< relref "stack.md" >}}` on a blog post I was writing, the warning will be emitted because `stack.md` isn't an absolute path and can't be found relative to my blog post page, but was found at `/docs/reference/stack.md`. Hugo is warning because the page should be specified as `{{< relref "/docs/reference/stack.md" >}}` (or I'm missing a `stack.md` page next to my blog post page).

> Is it that `relref` will only search for the file within the current directory and its children? And you are encouraged to use `ref` for something that exists outside of that? Similarly, what is the point of using `relref` if you fully-qualify the reference. e.g. why use `relref "/docs/...` instead of just `ref "/docs/...`?

The only difference between `ref` and `relref` is the resulting URL generated by Hugo -- it has nothing to do with the path you pass to these (the path you pass is treated the same for either). `ref` generates a fully qualified URL that includes the baseURL, `relref` generates a URL without the baseURL. It's similar to `Permalink` vs. `RelPermalink`.

For example:

 - `{{< relref "/docs/reference/stack.md" >}}` generates `/docs/reference/stack/`
 - `{{< ref "/docs/reference/stack.md" >}}` generates `https://www.pulumi.com/docs/reference/stack/`

The vast majority of our links do not need to have the baseURL in front of them, hence my preference is to use `relref` across the board unless there's a situation that necessitates the baseURL being there (e.g. if a sitemap or rss feed or something requires the baseURL to be there).

justinvp:
The cert for that site has been fixed.

praneetloke:
> Getting links from: http://localhost:1313/docs/console/accounts/saml/
1728-├─BROKEN─ https://www.samltool.com/prettyprint.php (BLC_UNKNOWN)
1729├─BROKEN─ https://www.samltool.com/prettyprint.php (HTTP_undefined)

These links must be coming from the SAML doc I added couple of months ago. It seems that the site is once again valid.

praneetloke:
> The cert for that site has been fixed.

Ha! I added my comment at the same time.

PR #1544, opened 2019-08-12
Add special handling for ShortText that can contain placeholder content that look like HTML tags

### LINKED ISSUE #1534: Necessary information gets stripped from API documentation
The docs for `core.v1.Service.get(...)` are unclear because information from the code documentation is being stripped as a part of the docs generation. Given the nature of the issue it's likely affecting many other functions too.

See the original docs at https://github.com/pulumi/pulumi-kubernetes/blob/1853784b6b4235553c18872d6e11c3026870356e/sdk/nodejs/core/v1/Service.ts#L49-L53

At https://pulumi.io/reference/pkg/nodejs/pulumi/kubernetes/core/v1/#Service-get this becomes the below description (with wrapping added):
```
Get the state of an existing Service resource, as identified by id. 
Typically this ID is of the form /; if is omitted, then (per 
Kubernetes convention) the ID becomes default/.
```

Note specifically that `<namespace>/<name>` becomes `/` in the formatted docs. The full value should be preserved in the docs.

praneetloke:
Fixes #1534.

I didn't regenerate the API docs since I didn't want to pollute the "Files Changed" yet. Once you approve the changes I made to the `tscdocgen` tool, I will regenerate the API docs. Also, I am wondering if I should do a similar fix in the Python docs as well?

### DIFF
--- a/tools/tscdocgen/main.go
+++ b/tools/tscdocgen/main.go
@@ -267,6 +267,15 @@ func (e *emitter) augmentNode(node *typeDocNode, parent *typeDocNode) {
 	node.Label = createLabel(node, parent)
 	node.CodeDetails = createCodeDetails(node)
 	node.RepoURL = getRepoURL(e.repoURL, node, parent)
+	// ShortText can contain an example snippet that contain placeholders that look like
+	// HTML tags, but aren't. For example, <namespace>/name where <namespace> is the placeholder
+	// for an actual namespace value.
+	if node.Comment.ShortText != "" &&
+		strings.Contains(node.Comment.ShortText, "<") &&
+		strings.Contains(node.Comment.ShortText, ">") {
+		node.Comment.ShortText = strings.ReplaceAll(node.Comment.ShortText, "<", "&lt;")
+		node.Comment.ShortText = strings.ReplaceAll(node.Comment.ShortText, ">", "&gt;")
+	}
 
 	// If this extends or implements other types, render them.
 	if len(node.ExtendedTypes) > 0 {
--- a/tools/tscdocgen/main.go
+++ b/tools/tscdocgen/main.go
@@ -270,7 +270,14 @@ func (e *emitter) augmentNode(node *typeDocNode, parent *typeDocNode) {
 	// ShortText can contain an example snippet that contain placeholders that look like
 	// HTML tags, but aren't. For example, <namespace>/name where <namespace> is the placeholder
 	// for an actual namespace value.
-	node.Comment.ShortText = html.EscapeString(node.Comment.ShortText)
+	if node.Comment.ShortText != "" &&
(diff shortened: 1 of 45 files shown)


tools/tscdocgen/main.go
```diff
@@ -267,6 +267,15 @@ func (e *emitter) augmentNode(node *typeDocNode, parent *typeDocNode) {
 	node.Label = createLabel(node, parent)
 	node.CodeDetails = createCodeDetails(node)
 	node.RepoURL = getRepoURL(e.repoURL, node, parent)
+	// ShortText can contain an example snippet that contain placeholders that look like
+	// HTML tags, but aren't. For example, <namespace>/name where <namespace> is the placeholder
+	// for an actual namespace value.
+	if node.Comment.ShortText != "" &&
+		strings.Contains(node.Comment.ShortText, "<") &&
+		strings.Contains(node.Comment.ShortText, ">") {
+		node.Comment.ShortText = strings.ReplaceAll(node.Comment.ShortText, "<", "&lt;")
```
justinvp: I know we replace `<` and `>` with `&lt;` and `&gt;` respectively elsewhere, but I wonder: should we be using https://golang.org/pkg/html/#EscapeString? Which would cover other characters like `&` as well?

praneetloke: I think that'd be OK. This also got me thinking if I even need to check for the presence of `<` and `>`? It seems like we ought to be able to simply call `EscapeString` regardless? I guess I was trying to protect against occurrences of `<` and `>` where they are not used as a tag, but that doesn't seem like it should matter at all? What do you think?

praneetloke:
@justinvp your idea to checkout the previously released tag to regenerate API docs on top of worked great. The diff is now a lot more simpler. Also, unfortunately, as we discovered the double-encoding, I reverted the use of `EscapeString`.

tools/tscdocgen/main.go
```diff
@@ -267,6 +267,17 @@ func (e *emitter) augmentNode(node *typeDocNode, parent *typeDocNode) {
 	node.Label = createLabel(node, parent)
 	node.CodeDetails = createCodeDetails(node)
 	node.RepoURL = getRepoURL(e.repoURL, node, parent)
+	// ShortText can contain an example snippet that contain placeholders that look like
+	// HTML tags, but aren't. For example, <namespace>/name where <namespace> is the placeholder
+	// for an actual namespace value.
+	if node.Comment.ShortText != "" &&
```
justinvp: The empty check isn't necessary nor are the `Contains` checks, right?

praneetloke: Yeah you are right. I don't need them. Just need to always try and replace the `<` and `>`.

PR #1547, opened 2019-08-13
A very crude implementation of the glossary

bermudezmt:
No real styling has been applied yet, but the idea is you get a list of all the terms with anchor links to the definition and relevant documentation pages. 

### DIFF
--- a/content/docs/reference/glossary.md
+++ b/content/docs/reference/glossary.md
@@ -5,7 +5,4 @@ menu:
     weight: 4
 ---
 
-This page is a collection of brief definitions of technical terms used in the Pulumi documentation.
-
-
-
+{{< glossary >}}
--- a/content/docs/reference/glossary.md
+++ b/content/docs/reference/glossary.md
@@ -1,5 +1,5 @@
 ---
-title: Pulumi Glossary
+title: Glossary
 menu:
   reference:
     weight: 4
--- a/content/docs/reference/glossary.md
+++ b/content/docs/reference/glossary.md
@@ -1,0 +1,11 @@
+---
+title: Pulumi Glossary
+menu:
+  reference:
+    weight: 4
+---
+
+This page is a collection of brief definitions of technical terms used in the Pulumi documentation.
+
+
+
--- a/data/glossary.toml
+++ b/data/glossary.toml
@@ -173,15 +173,3 @@ link = "/docs/intro/concepts/state"
 term = "Secrets Encryption"
 description = "The Pulumi Service automatically manages per-stack encryption keys on your behalf. Anytime you encrypt a value using --secret or by programmatically wrapping it as a secret at runtime, a secure protocol is used between the CLI and Pulumi Service that ensures secret data is encrypted in transit, at rest, and physically anywhere it gets stored."
 link = "/docs/intro/concepts/config#configuring-secrets-encryption"
-
-
-
-
-
-
-
-
-
-
-
-
--- a/data/glossary.toml
+++ b/data/glossary.toml
@@ -173,3 +173,15 @@ link = "/docs/intro/concepts/state"
 term = "Secrets Encryption"
 description = "The Pulumi Service automatically manages per-stack encryption keys on your behalf. Anytime you encrypt a value using --secret or by programmatically wrapping it as a secret at runtime, a secure protocol is used between the CLI and Pulumi Service that ensures secret data is encrypted in transit, at rest, and physically anywhere it gets stored."
 link = "/docs/intro/concepts/config#configuring-secrets-encryption"
+
+
+
+
+
+
+
+
+
+
+
+
--- a/data/glossary.toml
+++ b/data/glossary.toml
@@ -39,7 +39,7 @@ description = "SP stands for Service Provider. A Security Assertion Markup Langu
 link = "https://en.wikipedia.org/wiki/Service_provider_(SAML)"
 
 [[entries]]
-term = "IdP"
+term = "IdP Metadata XML"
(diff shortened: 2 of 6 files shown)


layouts/shortcodes/glossary.html
```diff
@@ -0,0 +1,16 @@
+{{ range sort $.Site.Data.glossary.entries "term" }}
+    <div>
+        <ul>
+            <li><a href="#{{ lower .term }}" class="">{{ .term }}</a></li>
```
justinvp: You probably want the loop to be around just the `<li>` elements, e.g.:

```html
<div>
    <ul>
        {{ range sort $.Site.Data.glossary.entries "term" }}
            <li><a href="#{{ lower .term }}" class="">{{ .term }}</a></li>
        {{ end }}
    </ul>
</div>
```

layouts/shortcodes/glossary.html
```diff
@@ -0,0 +1,16 @@
...
+        </ul>
+    </div>
+{{ end }}
+    
+{{ range sort $.Site.Data.glossary.entries "term" }}
+    <div class="flex-1 mt-8">
+        <h2 class="mt-2 mb-0">{{ .term }} </h2>
+            <p>{{ .description }}</p>
+                <ul class="mb-4">
```
justinvp: The closing `</ul>` tag is missing.

layouts/shortcodes/glossary.html
```diff
@@ -0,0 +1,16 @@
...
+                    <li>Read More: <a href="{{ .link }}" class="">{{ .term }}</a></li>
```
justinvp: I wonder if we should use `relref` here, which will ensure the pages we're linking to exist. (This is helpful when moving pages around / restructuring the site).

```suggestion
                    <li>Read More: <a href="{{ relref . .link }}" class="">{{ .term }}</a></li>
```

content/docs/reference/glossary.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Pulumi Glossary
+menu:
+  reference:
+    weight: 11
+---
+
+This page is a collection of brief definitions of technical terms used in the Pulumi documentation.
+
+{{< glossary >}}
```
cnunciato: Do you expect to re-use this shortcode (say, to display portions of the glossary in various places at some pont)? 

If not, I'd probably suggest we use a regular ol' layout instead.

joeduffy:
Love it!! 🎉

bermudezmt:
@justinvp This is ready for review, but I want to break out the terms list using the same amazing styles you've linked to at https://pulumi.slack.com/archives/C85BS3LJZ/p1567199962042300. :)

bermudezmt:
@justinvp is it possible to render the list items in a three-column layout using Hugo's `range` and CSS or does this need JavaScript? I may need your help again. :P

justinvp:
> is it possible to render the list items in a three-column layout using Hugo's `range` and CSS or does this need JavaScript?

Should be possible with CSS. I'll take a look tomorrow.

layouts/docs/single.html
```diff
@@ -22,6 +22,28 @@
                 {{ if ne .Title "" }}
                     <h1>{{ .Title }}</h1>
                 {{ end }}
+                
+                <!-- Render Glossary Page. -->
```
justinvp: @cnunciato, you were wondering if this should be a regular layout instead of a shortcode if it isn't going to be used anywhere else, but I actually think it should be a shortcode even if it's only used in on place. Otherwise we have to do this if check here (which I am not a fan of doing), or we have to essentially duplicate this `single.html` file and modify it to include the below (also not a huge fan of doing that). Unless you have another idea?

In the meantime, I've added a commit that just uses the shortcode for now.

justinvp:
@bermudezmt, I added a commit that adds 3 columns to the term index. Is this what you had in mind, or were you asking for something else?

I also moved this back to the shortcode for now, as I think it actually is fine for it to be in one even if this is the only place it is used (but am open to other ideas if @cnunciato has one).

I made some small other tweaks (hope you don't mind), but feel free to change things as you see fit!

data/glossary.toml
```diff
@@ -0,0 +1,175 @@
...
+description = "IdP stands for Identity Provider. A Security Assertion Markup Language (SAML) IdP is a service that acts as a user directory."
+link = "https://en.wikipedia.org/wiki/Identity_provider#SAML_identity_provider"
+
+[[entries]]
+term = "SP"
+description = "SP stands for Service Provider. A Security Assertion Markup Language (SAML) service provider relies on an identity provider for authentication."
+link = "https://en.wikipedia.org/wiki/Service_provider_(SAML)"
+
+[[entries]]
+term = "IdP"
```
justinvp: Should this be "IdP Metadata XML"? There's already an "IdP" term on line 32.

content/docs/reference/glossary.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Pulumi Glossary
```
justinvp: Wondering if we could get rid of `Pulumi` and just have it be Glossary?

data/glossary.toml
bermudezmt: Yes, I've changed in my last commit. And the 3-column layout was exactly what I was envisioning. Thanks for adding a commit! And please approve if all looks good, so we can have this merged. :)

content/docs/reference/glossary.md
bermudezmt: I took your suggestion.

PR #1561, opened 2019-08-15
Add a section for StackReference to the programming model

pgavlin:
Just what it says on the tin.


content/docs/reference/programming-model.md
```diff
@@ -839,6 +839,39 @@ If a stack contains any output values which are marked as secrets, their values
...
+
+```python
+from pulumi import StackReference
+
+other = StackReference(f"acmecorp/infra/other")
+other_output = other.get_output("x")
+```
+
+```go
+// StackReference is not supported in Go currently.
```
bermudezmt:
```suggestion
// StackReference is not currently supported in Go.
```

PR #1570, opened 2019-08-16
Use Pulumify for live website previews

joeduffy:
This adopts Pulumify, powered by GitHub Actions, to generate
live previews of our websites in response to Pull Requests.

The primary workhorse here is at
https://github.com/pulumi/actions-pulumify. In a nutshell

- Upon a new PR opening,

    * Create a new Pulumi stack, optionally inside the
      organization specified by `PULUMIFY_ORGANIZATION`. It
      will be named `pulumify_{repo}_{branch}`, with the
      necessary substitutions to make this a valid name.

    * Create an AWS S3 bucket configured to serve static
      content as a website.

- Upon a PR opening, reopening, or changing,

    * Synch the Git repo.

    * Optionally rebuild the content, using the
      `PULUMIFY_BUILD` command.

    * Synch the files at `PULUMIFY_ROOT` as AWS S3
      objects into the aforementioned target S3 bucket.
      Note that this uses an optimized synch mechanism that
      uploads a zipped file to the Amazon data center and
      then runs an `aws s3 sync` to copy the expanded files.
      This brings transfer times down from ~20 minutes (on
      our crappy WiFi) to ~2 minutes.

    * Post a comment to the PR with a link to the newly
      updated website, with a link to the Pulumi job.

- Upon a PR closing (or branch deletion),

    * Delete all of the above and remove the stack.

I've included a workaround to a seemingly new problem that
began as soon as I switched to the new GitHub Actions that
just shipped, which is that merging (and now even closing)
a PR doesn't actually trigger the action to run. For now, we
can key off the branch deletion which, given how good we are
at deleting branches, should be a reasonable approximation.

### DIFF
--- a/.github/workflows/pulumify.yml
+++ b/.github/workflows/pulumify.yml
@@ -1,0 +1,16 @@
+name: Pulumify
+on: [pull_request, delete]
+jobs:
+  updateLivePreview:
+    name: Update Live Preview
+    runs-on: ubuntu-latest
+    steps:
+    - uses: docker://joeduffy/pulumify
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        PULUMI_ACCESS_TOKEN: ${{ secrets.PULUMI_ACCESS_TOKEN }}
+        PULUMIFY_BUILD: make ensure && hugo --buildFuture -e $GITHUB_SHA
+        PULUMIFY_ROOT: public
+        PULUMIFY_ORGANIZATION: pulumi


joeduffy:
**Experimental**. I propose we run this side-by-side with our current solution for a bit to see whether we like it or not. If it proves to cause stability problems, we can easily disable it. However, I'm hopeful this will be a nice dogfood example of a new and powerful scenario we can demo/blog/etc.

Also note, I expect this PR to be red for a little bit, since it's our first GitHub Action in this repo (which I believe now supports them), and I will need to configure the requisite secrets.

Feel free to shout if anybody feels like this is a terrible idea.

joeduffy:
Seems to be working! 👍

ericrudder:
nice!
are you updating a parallel to #docs-notifications type thing too or is the
idea that all notifications are email?

joeduffy:
Notifications go straight into the pull request itself (see above)

joeduffy:
Ping @cnunciato -- would love to land for the weekend, in case there are bugs to iron out during lesser time crunch period.

joeduffy:
> On the stack naming, did you consider incorporating PR number also?

Good idea! I'll make that change (over in the pulumify repo).

joeduffy:
> Good idea! I'll make that change (over in the pulumify repo).

Turns out I couldn't make this change, at least in any way that was obvious to me. Unfortunately, given the `pull_request` bug that doesn't run actions when a PR is closed, this means the branch delete function doesn't actually have a PR number (which it would need in order to select the right Pulumi stack). I've reported this -- https://github.community/t5/GitHub-API-Development-and/pull-request-action-does-not-run-on-merge/td-p/29284 -- and if/when it is addressed, I'll make this good suggested change.

PR #1573, opened 2019-08-16
 Reduce diffs when updating API docs

### LINKED ISSUE #1532: Reduce diffs when updating API docs
The generated API docs includes links to source files on GitHub and these links include the Git commit SHA, which makes reviewing PRs difficult as there are lots of diffs due to the SHA change.

We could make it _much_ easier to review API docs changes by emitting the SHA once in front matter at the top of files and then use `{{< param sha >}}` everywhere in the file that needs the SHA.

justinvp:
Instead of emitting links that contain the Git commit hash, emit the package's repo URL (with Git commit hash) once to a data file, and then use a shortcode in the generated docs to access the URL. This should significantly reduce the size of the diffs when updating API docs, making it easier to see/review the relevant changes.

Fixes #1532


justinvp:
Forgot to regenerate kubernetes. Just added it.

PR #1574, opened 2019-08-16
Upgrade Node dependencies

cnunciato:
This upgrades typedoc and postcss-cli to more recent versions.


cnunciato:
@justinvp We shouldn't need to, because `yarn.lock` contains the actual versions, but I can do that. Will push an update. TypeDoc seems fine, too.

PR #1579, opened 2019-08-19
Update team roster

chrsmith:
Updates the Pulumi team roster to mark a few folk as inactive 😢 .


joeduffy:
😢  Two quick things:

1) Let's delete, not mark inactive.
2) Let's add Nick and Dan!! :tada:

PR #1581, opened 2019-08-19
Fix an incorrect statement in the AWS quickstart

joeduffy:
We reference the "EC2 instance shutting down," which is no
longer correct. The initial quickstart used EC2 instances but
it now uses an S3 bucket. This fix just tweaks the statement.


joeduffy:
I know what the problem is with the action -- will fix after team meeting.

PR #1583, opened 2019-08-19
Update the modify step for Azure Quickstart to truly modify an existing resource instead of completely replacing the existing program.

praneetloke:
### Proposed changes

As the title says, this PR updates the Azure quickstart example's "Modify Program" step to truly modify an existing resource by changing some attributes about it.

N/A

### Related issues (optional)

Related to #1177.

### DIFF
--- a/content/docs/quickstart/azure/modify-program.md
+++ b/content/docs/quickstart/azure/modify-program.md
@@ -7,9 +7,7 @@ menu:
     identifier: azure-modify-program
 ---
 
-Now that we have an instance of our Pulumi program deployed, let's enable:
-- Encryption on our storage account for blobs
-- Enable HTTPS-only communication
+Now that we have an instance of our Pulumi program deployed, let's enforce HTTPS-only communication for our storage account. This means, when making requests to this storage account only HTTPS traffic is allowed.
 
 Replace the entire contents of {{< langfile >}} with the following:
 
@@ -28,7 +26,6 @@ const account = new azure.storage.Account("storage", {
     resourceGroupName: resourceGroup.name,
     accountTier: "Standard",
     accountReplicationType: "LRS",
-    enableBlobEncryption: true,
     enableHttpsTrafficOnly: true,
 });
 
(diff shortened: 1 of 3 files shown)


content/docs/quickstart/azure/modify-program.md
```diff
@@ -19,91 +21,60 @@ const pulumi = require("@pulumi/pulumi");
...
-        }],
-    }],
-    osType: "Linux",
+// Create an Azure resource (Storage Account)
+const account = new azure.storage.Account("storage", {
     resourceGroupName: resourceGroup.name,
-    location: resourceGroup.location,
+    accountTier: "Standard",
+    accountReplicationType: "LRS",
+    enableBlobEncryption: true,
```
praneetloke: The reason I didn't keep this similar to AWS is because creating a KeyVault and adding secrets to it in Azure involves a few extra steps and likely to be a bit more advanced for someone trying to do a "quickstart", so I chose to use the Azure managed encryption and another attribute that enforces HTTPS traffic only.

content/docs/quickstart/azure/modify-program.md
```diff
@@ -7,7 +7,9 @@ menu:
     identifier: azure-modify-program
 ---
 
-Let's update our program to do something a little more interesting.
+Now that we have an instance of our Pulumi program deployed, let's enable:
```
justinvp: Add a blank line after this sentence before the beginning of the bullets to make Hugo happy.

content/docs/quickstart/azure/modify-program.md
justinvp: Sounds good.

content/docs/quickstart/azure/modify-program.md
praneetloke: Yep. Just noticed it too. Will fix it.

praneetloke:
> These changes LGTM, but the content of the subsequent pages will need to be updated based on these changes:

    https://www.pulumi.com/docs/quickstart/azure/deploy-changes/
    https://www.pulumi.com/docs/quickstart/azure/destroy-stack/

Ah yes! I'll fix those as well.

content/docs/quickstart/azure/modify-program.md
praneetloke: So it turned out that the `enableBlobEncryption` is `true` by default, which is good, but that meant I had to remove it from here. Instead, now I am using the HTTPS-only attribute as the update.

praneetloke:
@justinvp Build passed but Pulumify failed. I can merge this once you give me 👍

PR #1589, opened 2019-08-19
Update the feature image for token scanning post

praneetloke:
### Proposed changes

The first header Pulumi Access Token was being shown in the introductory text as a regular text, which looks odd. So I removed the first header. Also, updated the feature image to be 2:1 ratio at the recommended size of ~1200x600.

N/A

### Related issues (optional)

N/A


content/blog/github-token-scanning-service/index.md
```diff
@@ -10,8 +10,6 @@ meta_image: "feature.png"
 
 We are very excited to announce that we have partnered with GitHub to offer our users better protection for their Pulumi Access Tokens.
 
-## Pulumi Access Tokens
-
 By default, Pulumi users manage the state of their cloud infrastructure deployments using https://app.pulumi.com. This service provides state storage, concurrency control, audit history and access controls for both individuals and teams working with Pulumi.  Each user and service account can generate one or more Pulumi Access Tokens to be used to authenticate with this service.  These access tokens can be used on both local development machines, as well as in CI/CD systems for automated infrastructure deployments.  These access tokens are sensitive secrets which should never be shared publicly, and in particular should never be committed to source control.
```
justinvp: Another thing Eric noticed:

Auto-links don't show up in summaries:

Screen Shot 2019-08-19 at 2 12 34 PM

But explicit links in summaries of other posts do appear to show up:

Screen Shot 2019-08-19 at 2 13 50 PM

I think we may need to just explicitly make this a link so that it's both a link in the post itself and the summary:

```suggestion
By default, Pulumi users manage the state of their cloud infrastructure deployments using [https://app.pulumi.com](https://app.pulumi.com). This service provides state storage, concurrency control, audit history and access controls for both individuals and teams working with Pulumi.  Each user and service account can generate one or more Pulumi Access Tokens to be used to authenticate with this service.  These access tokens can be used on both local development machines, as well as in CI/CD systems for automated infrastructure deployments.  These access tokens are sensitive secrets which should never be shared publicly, and in particular should never be committed to source control.
```

praneetloke:
Confirmed the summary looks good now in the internal preview site. Pulumify seems to be showing some cached version that shows the summary where the app.pulumi.com isn't hyperlinked in it.

PR #1593, opened 2019-08-20
Improve auto-naming docs

joeduffy:
This change embellishes our rationale on auto-naming, adds a bit more
explanation, and a few new examples. It also adds a FAQ entry for
increased visibility.

### DIFF
--- a/content/docs/reference/programming-model.md
+++ b/content/docs/reference/programming-model.md
@@ -372,15 +372,104 @@ vpc, _ := ec2.Vpc(ctx, "vpc", &VpcArgs{}, pulumi.ResourceOpt{Provider: provider}
 
 ### Resource names {#names}
 
-Every resource managed by Pulumi has a name.  This name is used to track the identity of a resource across multiple deployments of the same program.  The name specified during resource creation is used in three ways:
+Every resource managed by Pulumi has a **logical name** that you specify as an argument to its constructor. For instance, the logical name of this AWS S3 Bucket is `my-bucket`:
 
-1. It is used as part of constructing the Universal Resource Name (URN) used by the Pulumi engine to track the resource across updates.
-2. Most resource providers will use it as a default prefix for constructing the cloud-provider name of the resource.
-3. Some CLI commands use the URN.
+{{< langchoose >}}
+
+```javascript
+let b = new aws.s3.Bucket("my-bucket");
+```
+
+```typescript
+let b = new aws.s3.Bucket("my-bucket");
+```
+
+```python
+b = Bucket("my-bucket")
+```
+
+```go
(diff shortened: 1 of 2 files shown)


content/docs/reference/programming-model.md
```diff
@@ -372,15 +372,104 @@ vpc, _ := ec2.Vpc(ctx, "vpc", &VpcArgs{}, pulumi.ResourceOpt{Provider: provider}
...
+```typescript
+let b = new aws.s3.Bucket("my-bucket", {
+    bucket: "my-bucket-001",
+});
+```
+
+```python
+b = Bucket('my-bucket', {
+    bucket='my-bucket-001'
+})
```
lukehoban: Bucket is kind of an unfortunate example to use as it’s actually a significant outlier in terms of using `bucket` for its physical name instead of `name`. It’s especially important here to capture the general pattern related to `name` since it’s exactly the confusion between the `name` arg (logical) and `name` property (physical) that tends to be so confusing.

content/docs/reference/programming-model.md
```diff
@@ -372,15 +372,104 @@ vpc, _ := ec2.Vpc(ctx, "vpc", &VpcArgs{}, pulumi.ResourceOpt{Provider: provider}
...
+
+* It allows Pulumi to do zero-downtime resource updates. Certain updates require replacing resources, rather than updating them in place. This is because Pulumi can create replacements first, then update existing references to them, and finally delete the old resources. If it weren't for auto-naming, Pulumi would need to do things in a very different order: namely, it would need to delete resources first, and create new instances afterwards, which is far more impactful and leads to downtime.
+
+We understand that some end users will prefer precise names. Auto-naming can be overridden by manually specifying a physical name on your resource. Most resources offer this option by way of a `name` property that may be specified in the argument object to the constructor. It just so happens that our running example of an S3 Bucket uses `bucket` instead:
```
lukehoban: Very minor - but I don’t think we tend to use phrases like “we understand” in these docs - we tend to be more fact based like “For some use cases it is important to have precise names”.

content/docs/reference/programming-model.md
joeduffy: Agree. I'll change -- bucket is just such a simple resource and we use it in our quickstarts. And yet I 100% agree and it felt awkward to have to show an example with `bucket: ` instead of `name: ` :-)  I'll fix.

content/docs/reference/programming-model.md
joeduffy: Agree, will fix!

content/docs/reference/programming-model.md
joeduffy: Ok, updated to `aws.iam.Role`. It's one of the few resources that can be created without other args, which I felt might distract attention away from the salient point here. Lmk if you can think of a better example!

PR #1594, opened 2019-08-20
Update the GCP quickstart's modify step to truly modify an existing resource

### LINKED ISSUE #1177: Suggestion: Change Modify step in Getting Started to truly Modify

praneetloke:
### Proposed changes

Updates the GCP Quickstart steps to modify an existing resource.

N/A

### Related issues (optional)
This is the last of tasks for the linked issue, so I think it fixes #1177.

### DIFF
--- a/content/docs/quickstart/gcp/modify-program.md
+++ b/content/docs/quickstart/gcp/modify-program.md
@@ -84,8 +84,6 @@ bucket = storage.Bucket('my-bucket',
 pulumi.export('bucket_name',  bucket.url)
 ```
 
-Our program now creates a simple virtual machine running a Python web server.
-
 Next, we'll deploy the changes.
 
 {{< get-started-stepper >}}
--- a/content/docs/quickstart/gcp/modify-program.md
+++ b/content/docs/quickstart/gcp/modify-program.md
@@ -9,7 +9,7 @@ menu:
 
 Now that we have an instance of our Pulumi program deployed, let's update it to use our own encryption key instead of the default Google-managed one.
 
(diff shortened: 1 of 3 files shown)


content/docs/quickstart/gcp/modify-program.md
```diff
@@ -7,7 +7,9 @@ menu:
     identifier: gcp-modify-program
 ---
 
-Now that we have an instance of our Pulumi program deployed, let's update it to do something a little more interesting.
+Now that we have an instance of our Pulumi program deployed, let's update it to use our own encryption key instead of the default Google-managed one.
+
+**Note**: You must enable the Google KMS API for your project on the GCP console before proceeding. You can enable the API by following this link: `https://console.cloud.google.com/security/kms/noaccess?project={your_project_id}`. Be sure to replace the `{your_project_id}` with your actual Google project ID.
```
lukehoban: Should we put in a block quote (`> `) like we have in several other pages in the quickstart for similar "asides"?

praneetloke: Sure thing. I'll change it.

PR #1599, opened 2019-08-20
Handle placeholder text that look like HTML tags in the Kubernetes API docs

### LINKED ISSUE #1534: Necessary information gets stripped from API documentation

praneetloke:
### Proposed changes
I have reduced the scope of the previous attempt at fixing this to only do this for the Kubernetes API docs. This at least fixes the known-issue that affects the readability of our Kubernetes nodejs API docs.

N/A

### Related issues (optional)
Fixes #1534.

### DIFF
--- a/tools/tscdocgen/main.go
+++ b/tools/tscdocgen/main.go
@@ -207,7 +207,7 @@ func emitMarkdownDocs(srcdir, pkgname string, doc *typeDocNode, outdir, outdatad
 	if err != nil {
 		return err
 	}
-	e.augmentNode(doc, nil)
+	e.augmentNode(doc, nil, k8s)
 	return e.emitMarkdownModule(rootModule, root, true)
 }
 
@@ -259,7 +259,7 @@ func (e *emitter) emitPackageDataFile(pkgRepoDir, githash string) error {
 }
 
 // augmentNode recurses throughout a tree AST, adding information that we'll need when translating it to Markdown.
-func (e *emitter) augmentNode(node *typeDocNode, parent *typeDocNode) {
+func (e *emitter) augmentNode(node *typeDocNode, parent *typeDocNode, k8s bool) {
 	// Add some labels.
 	node.AnchorName = node.Name
 	if parent != nil && (parent.Kind == typeDocClassNode || parent.Kind == typeDocInterfaceNode) {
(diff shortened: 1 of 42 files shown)


tools/tscdocgen/main.go
```diff
@@ -268,6 +268,14 @@ func (e *emitter) augmentNode(node *typeDocNode, parent *typeDocNode) {
 	node.Label = createLabel(node, parent)
 	node.CodeDetails = createCodeDetails(node)
 	node.URLPath = getURLPath(node, parent)
+	if k8s && node.Comment.ShortText != "" &&
+		strings.Contains(node.Comment.ShortText, "<") &&
+		strings.Contains(node.Comment.ShortText, ">") {
+		// To avoid double-encoding strings that are already encoded, we make a targeted replacement
+		// of the < and > characters alone.
+		node.Comment.ShortText = strings.ReplaceAll(node.Comment.ShortText, "<", "&lt;")
+		node.Comment.ShortText = strings.ReplaceAll(node.Comment.ShortText, ">", "&gt;")
```
justinvp: We still don't need the `node.Comment.ShortText != ""` and `string.Contains` checks, right?

So it could be simply:

```go
if k8s {
	// To avoid double-encoding strings that are already encoded, we make a targeted replacement
	// of the < and > characters alone.
	node.Comment.ShortText = strings.ReplaceAll(node.Comment.ShortText, "<", "&lt;")
	node.Comment.ShortText = strings.ReplaceAll(node.Comment.ShortText, ">", "&gt;")
}
```

Might also be worth adding a little more detail to the comment about why we only do this for Kubernetes (so that `<namespace>/<name>`, which shows up a lot, is properly escaped), but don't for other providers (because there are cases for other providers we do want`<` and `>` to be emitted unescaped).

praneetloke: I'll make the updates.

praneetloke: However, I'll leave the explicit check for `<` and `>` even though it is restricted to k8s now. Just to avoid any other unintended changes.

justinvp: Oh, I see, the checks are ensuring both `<` and `>` are in the string before doing the replacements. Got it.

PR #1601, opened 2019-08-21
Emit docs for input/output types and hyperlink to them

### LINKED ISSUE #1467: Nested types (inputs and outputs) aren't shown and aren't linked in API docs
### Problem description

- I saw a product documentation issue on the following URL: https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/kubernetes/

None of the input/output types are linked in the k8s Node API reference docs.
For example, https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/kubernetes/apps/v1/#Deployment-constructor shows  that `args` is of the `inputApi.apps.v1.Deployment` type, but that type is not linked.

### Suggestions for a fix

The Node SDK in the k8s provider may need to be updated to reference the exported types for the doc generator to work properly.

### LINKED ISSUE #1518: Types in namespaces are missing from API docs
For example, in awsx:

https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/awsx/cloudwatch/#events

Screen Shot 2019-08-07 at 12 12 58 PM

I would expect to see exported types from that namespace, such as `CloudWatchMetricChange`, `deadLetterInvocations`, `invocations`, etc.

https://github.com/pulumi/pulumi-awsx/blob/5ac4a8c74a3e6b48f7ca0e5e48c85217fb0b88ab/nodejs/awsx/cloudwatch/metrics.ts

Screen Shot 2019-08-07 at 12 14 44 PM

---

The docs are also referring to these as "module", but that's older TypeScript terminology.

https://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html

> **A note about terminology**: It’s important to note that in TypeScript 1.5, the nomenclature has changed. “Internal modules” are now “namespaces”. “External modules” are now simply “modules”, as to align with [ECMAScript 2015](http://www.ecma-international.org/ecma-262/6.0/)’s terminology, (namely that `module X {` is equivalent to the now-preferred `namespace X {`).

lukehoban:
This makes minimal changes necessary to support rendering docs pages containing the input and output types that are part of the Kubernetes provider and all tfbridge-based providers.

This involves three things:
1. Support for nested TS namespaces via recursive mustache templates
2. Special-cased hyperlinking to input and output types 
3. Special-case treatment of the top-level `types` modules

This works, and ensures that all this documentation is actually visible and linked to.  And it maintains the same consistent "model" as the existing docs that each "loadable module" gets its own page and any nested namespaces  are treated as nested content within a page.  That model never really made sense (in particular, the user view of any of our TypeScript libraries is as a namespaced structure within a single loadable module, so presenting the loadable module structure as the primary physical structure of the docs is confusing the physical layout on disk for the logical view presented to used), but these changes add new stress on it - in particular in leading to very large pages for `types/input` in the Kubernetes provider, which now contains type definitions for every input type in the entire provider.  After exploring more special-cased changes - I decided that we should stick with what we have for now, and make a more fundamental fix to the doc generators page structure in the future.

Fixes #1518.
Fixes #1467.


tools/tscdocgen/main.go
```diff
@@ -546,10 +547,11 @@ func (e *emitter) gatherModules(doc *typeDocNode, parentModule string, k8s bool)
 			name := strings.Trim(child.Name, `"`)
 
 			// If it begins with a "./", simplify it to just the parent name.
-			if isModule && strings.Index(name, "./") == 0 {
+			if isModule {
```
lukehoban: I do not understand why the `&& strings.Index(name, "./") == 0` case was here previously.  I haven't encountered any providers that trigger this condition at all yet.  For the changes being made here, it is necessary to do this even for cases where this condition applies, but I have probably broken something for other cases where this did apply previously - I just can't find any to verify and understand what they were trying to accomplish.  Anyone know?

content/docs/reference/pkg/nodejs/pulumi/kubernetes/admissionregistration/v1beta1/_index.md
```diff
@@ -44,7 +44,7 @@ or reject and may change the object.
 <div class="pdoc-member-contents">
 {{% md %}}
 
-<pre class="highlight"><span class='kd'></span><span class='kd'>new</span> MutatingWebhookConfiguration(name: <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String'>string</a></span>, args?: inputApi.admissionregistration.v1beta1.MutatingWebhookConfiguration, opts?: <a href='/docs/reference/pkg/nodejs/pulumi/pulumi/#CustomResourceOptions'>pulumi.CustomResourceOptions</a>)</pre>
+<pre class="highlight"><span class='kd'></span><span class='kd'>new</span> MutatingWebhookConfiguration(name: <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String'>string</a></span>, args?: <a href='/docs/reference/pkg/nodejs/pulumi/@pulumi/kubernetes/types/input/admissionregistration/v1beta1/#MutatingWebhookConfiguration'>inputs.admissionregistration.v1beta1.MutatingWebhookConfiguration</a>, opts?: <a href='/docs/reference/pkg/nodejs/pulumi/pulumi/#CustomResourceOptions'>pulumi.CustomResourceOptions</a>)</pre>
```
justinvp: The URL is including an extra `@pulumi` component that should be removed.

This:

```
/docs/reference/pkg/nodejs/pulumi/@pulumi/kubernetes/types/input/...
```

Should be:

```
/docs/reference/pkg/nodejs/pulumi/kubernetes/types/input/...
```

lukehoban: Ahh - yes - this was actually fixed already but the kubernetes docs weren't regenerated with the fix.  Should be taken care of as part of regenerating all the docs.

tools/tscdocgen/main.go
```diff
@@ -546,10 +547,11 @@ func (e *emitter) gatherModules(doc *typeDocNode, parentModule string, k8s bool)
 			name := strings.Trim(child.Name, `"`)
 
 			// If it begins with a "./", simplify it to just the parent name.
```
lblackstone: This comment doesn't match anymore

tools/tscdocgen/main.go
lukehoban: I found a few places that use this - and so ended up putting it back as is.

content/docs/reference/pkg/nodejs/pulumi/azuread/types/_index.md
```diff
@@ -0,0 +1,26 @@
...
+
+<div class="toggleVisible">
+<div class="collapsed">
+<h2 class="pdoc-module-header toggleButton" title="Click to show Modules">Modules ▹</h2>
+</div>
+<div class="expanded">
+<h2 class="pdoc-module-header toggleButton" title="Click to hide Modules">Modules ▾</h2>
+<div class="pdoc-module-contents">
+<ul>
+<li><a href="index">types/index</a></li>
```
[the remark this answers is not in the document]
lukehoban: Added a special case to remove this.

content/docs/reference/pkg/nodejs/pulumi/kubernetes/types/_index.md
```diff
@@ -5,25 +5,148 @@ title: Module types
...
-* <a href="v1alpha1">types/v1alpha1</a>
-* <a href="v1beta1">types/v1beta1</a>
-* <a href="v1beta2">types/v1beta2</a>
-* <a href="v2alpha1">types/v2alpha1</a>
-* <a href="v2beta1">types/v2beta1</a>
-* <a href="v2beta2">types/v2beta2</a>
-* <a href="version">types/version</a>
+<div class="pdoc-module-contents">
+<ul>
+<li><a href="input/admissionregistration">types/input/admissionregistration</a></li>
```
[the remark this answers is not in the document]
lukehoban: Re-did how this works so this no longer occurs.

PR #1563, opened 2019-08-15
New TOC and bread crumbs

bermudezmt:
### Proposed changes
So far, I have done the following:
- Move `quickstart` to `get-started`
- Move `install.md` as a top-level nav item
- Move `concepts` out of reference and under a new directory called `intro`
- Update the TOC to have some labels like https://tailwindcss.com/docs/ (e.g. "Getting Started", "Intro", etc)

I am unable to preview this however as I just get a bunch of "REF NOT FOUND" errors. :( 

#831

PR #1606, opened 2019-08-22
Use 301 responses for redirects instead of meta refresh redirects

justinvp:
We use [Hugo aliases](https://gohugo.io/content-management/urls/#aliases) to create redirects when moving pages around on the site. We also have our own [`redirect_to`](https://github.com/pulumi/docs/blob/71260ec7283cf22d720f01c55b02994e56ccf8ec/layouts/partials/head.html#L7-L9) param that can be specified in a page's front matter to have it redirect to another URL. In both cases, since this is a statically generated site, a meta refresh redirect is used to implement the redirect (e.g. `<meta http-equiv="refresh" content="0; url=destination" />`).

Unfortunately, browsers don't flow #anchors from the source page to the destination with meta refresh redirects, and we make heavy use of #anchors in links throughout the site. However, browsers _will_ flow an #anchor across to the redirect destination with 301 redirects (as long as the destination URL itself does not include its own #anchor).

This change updates the Pulumi program to implement 301 redirects. When crawling the directory, if the file is an .html file that contains a meta refresh tag, configure the bucket object to be a [webpage redirect](https://docs.aws.amazon.com/AmazonS3/latest/dev/how-to-page-redirect.html) that returns a 301 response when requested. We use a fully qualified URL for the redirect destination, which S3 sets in the `Location` header, so this all works even when served from CloudFront. The implementation is adapted from the redirectron-9000.

I've tested these changes in a test stack and it works as expected.

This is a prerequisite to #1563 as we're moving pages around and want #anchors to flow across to the redirected pages.

---

While making changes here, modernize the Pulumi program (e.g. upgrade dependencies, use built-in TypeScript support, and make the linter happy) and remove some code that's no longer necessary.


infrastructure/index.ts
```diff
@@ -267,20 +261,65 @@ crawlDirectory(
...
                 key: relativeFilePath,
                 bucket: contentBucket,
-                source: new pulumi.asset.FileAsset(filePath),
-                contentType: getMimeType(filePath) || undefined,
+
+                // Use "/dev/null" for an empty object if this is a redirect, otherwise the actual file path.
+                source: new pulumi.asset.FileAsset(redirect ? "/dev/null" : filePath),
+
+                // Add `contentType` if this isn't a redirect.
+                ...!redirect && { contentType: getMimeType(filePath) },
```
chrsmith: Personally this `...!redirect && { obj-literal }` is really hard to understand.

Why don't we just create the `aws.s3.BucketObject` in two separate parts of an if-statement. e.g.

```
// Create a new S3 object for most files, but HTML files that contain a redirect
// just create a stubbed out S3 object with the right metadata so they return a
// 301 redirect (instead of serving the HTML with a meta-redirect). This ensures
// the right HTTP code response is returned for search engines, as well as
// enables better support for URL anchors.
if (!redirect) { 
   const contentFile = new aws.S3.BucketObject(
        ...);
} else {
    const s3Redirect = new aws.s3.BucketObject(
        ...);
}
```

infrastructure/index.ts
```diff
@@ -267,20 +261,65 @@ crawlDirectory(
...
+    if (redirect.startsWith("https://") || redirect.startsWith("http://")) {
+        return redirect;
+    }
+
+    // If the redirect starts with "/", prefix with the redirect domain and return it.
+    if (redirect.startsWith("/")) {
+        return `https://${redirectDomain}${redirect}`;
+    }
+
+    return undefined;
```
chrsmith: This doesn't seem right... the HTML file's redirect could exist, but just be malformed? Like "redirect='404.html'"?

Rather than returning `undefined` in this case, why don't we just return a runtime error? (Refusing to perform the Pulumi update.) That would force us to find/fix the issue, rather than not creating the redirect.

What do you think?

infrastructure/index.ts
```diff
@@ -267,20 +261,65 @@ crawlDirectory(
...
+                // Add `websiteRedirect` if this is a redirect.
+                ...redirect && { websiteRedirect: redirect },
             },
             {
                 parent: contentBucket,
             });
     });
 
+// Returns the redirect URL if filePath is a .html file that contains a meta refresh redirect, otherwise undefined.
+function getMetaRefreshRedirect(filePath: string): string | undefined {
```
chrsmith: As a suggestion, I'm not a fan of us checking the `filePath` for the meta refresh redirect, and then also modifying it before returning. That kinda hides/obscures a really important point.

Instead, could we just have `getMetaRefreshRedirect` return just the value from the HTML file and then have a separate `translateFileRedirect` method to do the modification? That makes the code clearer, IMHO. But it isn't that big of a deal.

infrastructure/index.ts
```diff
@@ -267,20 +261,65 @@ crawlDirectory(
...
+    // Only .html files contain meta refresh redirects.
+    if (path.extname(filePath) !== ".html") {
+        return undefined;
+    }
+
+    // Extract the redirect from the content of the file.
+    const text = fs.readFileSync(filePath, "utf8");
+    const regex = /<meta\s+?http-equiv="refresh"\s+?content="0;\s+?url=(.*?)"/gm;
```
chrsmith: It might be worth-while to change this to be a case-_insensitive_ search, using `/gmi`. While the meta redirects are generated by Hugo and like "safe", it would just make this method a bit more robust.

infrastructure/Pulumi.production.yaml
```diff
@@ -3,3 +3,4 @@ config:
   www.pulumi.com:certificateArn: arn:aws:acm:us-east-1:058607598222:certificate/a3f72a2b-e715-4639-b126-1e4efc0b634b
   www.pulumi.com:pathToWebsiteContents: ../public
   www.pulumi.com:targetDomain: www-production.pulumi.com
+  www.pulumi.com:redirectDomain: www.pulumi.com
```
chrsmith: I assume we'd want to update the `Pulumi.staging.yaml` as well? Even if the repo isn't regularly updated, we should keep it up to date so to avoid confusing the next time we do go an update it.

justinvp: What I really want to do is just use the `targetDomain`. The way the code is written, it will use the `redirectDomain` if set, otherwise use the `targetDomain`.

```typescript
const redirectDomain = config.redirectDomain || config.targetDomain;
```

I had to add this `redirectDomain` config because the `targetDomain` for production is not the domain we actually want to use. (It is correct for staging). For production it is `www-production.pulumi.com` but we want to use `www.pulumi.com`. This is fallout of this being set outside of the Pulumi program.

infrastructure/index.ts
justinvp: OK. Done.

infrastructure/index.ts
justinvp: Yeah, I was kind of putting my head in the sand because if there is a problem with the redirect URL, even without this change, we'd already have the problem and wouldn't know about it. But as long as we're now going through the effort to extract and use it in a different way, you're right, we should fail if it is malformed so we can fix the problem. Fixed.

infrastructure/index.ts
justinvp: Sounds good. Fixed.

infrastructure/index.ts
justinvp: Done.

PR #1617, opened 2019-08-25
Shrink the margin on blockquotes slightly

joeduffy:
The margin on blockquotes, with the orange bar on the left,
looked a bit large to my eyes. This has the effect of making
the blockquote, which is often used as an aside, much more
disruptive to the flow -- particularly for smaller blockquotes
(like single liners). Especially with the smaller font we use
for blockquotes, this "feels" more appropriate.


joeduffy:
For example, before:

image

And after:

image

PR #1642, opened 2019-08-28
WIP: See if this change makes lists render properly

lblackstone:


lblackstone:
Made the relevant change in the k8s repo, so this update will get pulled in next time we update the k8s docs.

Issue #1322, opened 2019-07-11
[RECURRING] Editorial pass on blog posts


bermudezmt:
- Consistent voice
- Grammar / spelling fixes
- Adhere to a sensible style guide (https://developers.google.com/style/ for example) 

Reviewed:
- [x] "7 Ways to Deal with Application Secrets in Azure" (https://github.com/pulumi/docs/pull/1379)
- [x] "July 2019 Pulumi Up Recap" (https://github.com/pulumi/docs/pull/1371)
- Review SAML docs 
  - [x] https://www.pulumi.com/docs/reference/service/saml-okta/?origin_team=T4PBPMA8J
  - [x] https://www.pulumi.com/docs/reference/service/saml-aad/
  - [x] https://www.pulumi.com/docs/reference/service/saml-gsuite/
  - [x] https://www.pulumi.com/docs/reference/service/saml-overview/?origin_team=T4PBPMA8J
- Review Kubernetes PRs 
  - [x] EKS Clusters: Migrating Node Groups with Zero Downtime (https://github.com/pulumi/docs/pull/1327)
  - [x] Day 2 Kubernetes: Migrating EKS Node Groups with Zero Downtime (https://github.com/pulumi/docs/pull/1328/files)
- [x] https://www.pulumi.com/docs/reference/serializing-functions/
- [x] https://github.com/pulumi/docs/pull/1557 - 10 Pearls with Azure Functions in Pulumi
- [x] https://www.pulumi.com/docs/intro/vs/
- [x] https://www.pulumi.com/docs/intro/concepts/
- [x] https://www.pulumi.com/docs/intro/vs/cloud_template_transpilers/
- [x] How To Build Globally Distributed Applications with Azure Cosmos DB and Pulumi (Google doc)

bermudezmt:
This is a recurring editorial task that will apply to each milestone. cc: @erinhyperplane for your reference.

bermudezmt:
Done for this milestone via #1644

bermudezmt:
Last edit done via Google Docs.

PR #1616, opened 2019-08-24
Console Follow-Ups

bermudezmt:
- Rework Console landing page
- Update Account, Editions, Organizations pages
- Add a Project and Stack Management page
- Editorial pass on all pages

Signed-off-by: bermudezmt <bermudez.mt@gmail.com>

### Related issues (optional)

Fixes #1616 and #1612

PR #1644, opened 2019-08-29
SAML updates

### LINKED ISSUE #1608: Make it more obvious when placeholders need to be replaced
### Problem description

Our SSO guides require users to replace placeholders in the _ACS URL_, _Entity ID_, etc. It's not uncommon for users to copy and paste these values directly without noticing they need to customize them with their own organization name. This leads to errors when they try to use their SSO.

- https://www.pulumi.com/docs/console/accounts/saml/okta/
- https://www.pulumi.com/docs/console/accounts/saml/aad/
- https://www.pulumi.com/docs/console/accounts/saml/gsuite/

image

### Suggestions for a fix

We should add an icon or some other visual indicator that makes it more obvious that the user needs to customize the URL so that users have a better chance of getting their SSO setup correctly configured the first time.

Also, we present this information inconsistently across the guides; in tabular format in the okta guide, but a list in the gsuite guide, and with inconsistent styling too. It would be nice if this were consistent so it's easier to update and switch between the guides.

bermudezmt:
- Rename SSO page
- Add SSO page as required reading for integration guides
- Edit guides

### Related issues (optional)

Relates to #1322 and fixes #1608


content/docs/guides/saml/aad.md
```diff
@@ -20,12 +24,12 @@ The first step is to add a new Pulumi application to your Azure AD tenant.
...
 
-1. Click **New application** at the top of the **All applications** blade.
+1. Click **New application**.
 
     ![New application](/images/docs/reference/service/saml-aad/new-application.png)
 
-1. Click the **Non-gallery application** tile and in the **Add your own application** blade, enter
-   _Pulumi Cloud Console_ as the application name and click **Add**.
+1. Select **Non-gallery application** tile and in the **Add your own application** panel, enter
+   _Pulumi Cloud Console_ as the application name then click **Add**.
```
bermudezmt: @chrsmith This is dependent on getting #1616 merged, but can we change the instructions here to say "Pulumi Console"? When I was testing with GitHub, GitLab, and Atlassian recently, the app was still labeled "Pulumi Cloud Console".

chrsmith: Yeah, we can change the text here. The "application name" in Azure AD, Okta, etc. doesn't matter. It's just for the Azure or Okta administrator to know what that SAML SSO application is for. (Read: it won't impact SAML SSO actually working.)

content/docs/guides/saml/aad.md
```diff
@@ -38,22 +42,23 @@ The first step is to add a new Pulumi application to your Azure AD tenant.
...
+
 The values you will want to use are dependent upon your organization's name on Pulumi. Be sure
 to replace `acmecorp` with whatever your specific Pulumi organization's name is.
 
 | SAML Setting | Value |
 | --------------- | ----- |
-| Identifier (Entity ID) | `https://api.pulumi.com/login/**acmecorp**/sso/saml/metadata` |
-| Reply URL | `https://api.pulumi.com/login/**acmecorp**/sso/saml/acs` |
-| Relay State | `https://api.pulumi.com/login/**acmecorp**/sso` |
+| Identifier (Entity ID) | `https://api.pulumi.com/login/<acmecorp>/sso/saml/metadata` |
```
chrsmith: Could put a big, yellow warning sign on the paragraph above?

We've had reports from some users not replacing `acmecorp` when configuring their SAML SSO settings, which leads to failures logging in. (And could you also make the corresponding warning/message in our Okta, G Suite, and Azure AD SAML guides too?)

Ultimately this is evidently a bigger stumbling block than you'd think.

content/docs/guides/saml/aad.md
```diff
@@ -63,20 +68,20 @@ The final step is to configure the **name identifier format**.
...
 
-1. In the **Manage User Claims** panel, expand the **Choose name identifier format** and select **Email address**.
+1. In the **Manage User Claims** panel, expand **Choose name identifier format** and select **Email address**.
 
 ![Manage Name Identifier Format](/images/docs/reference/service/saml-aad/name-identifier-format.png)
 
 1. Finally, click **Save** at the bottom of the **Manage User Claims** panel.
 
-**IMPORTANT:** Do not change the value of Name ID Format value once your users have started using Pulumi; not even switching its value between Email or Persistent.
+> **Important:** Do not change the value of Name ID Format value once your users have started using Pulumi---not even switching its value between Email or Persistent.
```
chrsmith: Shouldn't we be using just two `--` dashes, which would then be replaced by an em dash? Or is the triple-dash somehow different/better?

bermudezmt: I think two dashes render an en dash.

content/docs/guides/saml/aad.md
bermudezmt: What do you think of http://content-bucket-1ddffb8.s3-website-us-east-1.amazonaws.com/docs/guides/saml/aad/#enter-pulumi-configuration-into-your-azure-ad-application? @cnunciato this was the warning class I came up with previously, but we could update this (or revisit later) if you have some other suggestions.

bermudezmt: I also added a `required` badge here: http://content-bucket-1ddffb8.s3-website-us-east-1.amazonaws.com/docs/guides/saml/sso/#default-relay-state

bermudezmt:
@joeduffy @ericrudder would love your eyes on http://content-bucket-1ddffb8.s3-website-us-east-1.amazonaws.com/docs/guides/ as I suspect you can "fluff" the text up. :smile:

bermudezmt:
This should address #1608 as well.

Issue #1648, opened 2019-08-29
Re-enable broken link checker in CI


justinvp:
The `validate` target, which runs the broken link checker, is now consistently failing in CI due to Hugo failing with `fatal error: runtime: out of memory`. See below.

This is blocking deploying updates to the site, so I am going to temporarily disable this step until we can have a solution.

This issue tracks addressing the underlying issue and re-enabling the issue.

This may be related to #1647.

```
make validate
311make[1]: Entering directory '/home/travis/build/pulumi/docs'
312hugo server \
313    --buildDrafts --disableBrowserError --disableLiveReload \
314    &>/dev/null &
315while ! nc -z localhost 1313; do sleep 0.1; done
316Building sites … fatal error: runtime: out of memory
317
318runtime stack:
319runtime.throw(0x1229c8b, 0x16)
320	/usr/local/go/src/runtime/panic.go:617 +0x72
321runtime.sysMap(0xc1c8000000, 0x4000000, 0x1ecacd8)
322	/usr/local/go/src/runtime/mem_linux.go:170 +0xc7
323runtime.(*mheap).sysAlloc(0x1ea3f80, 0x9c000, 0x1ea3f90, 0x4e)
324	/usr/local/go/src/runtime/malloc.go:633 +0x1cd
325runtime.(*mheap).grow(0x1ea3f80, 0x4e, 0x0)
326	/usr/local/go/src/runtime/mheap.go:1232 +0x42
327runtime.(*mheap).allocSpanLocked(0x1ea3f80, 0x4e, 0x1ecace8, 0x0)
328	/usr/local/go/src/runtime/mheap.go:1150 +0x3a7
329runtime.(*mheap).alloc_m(0x1ea3f80, 0x4e, 0x101, 0x0)
330	/usr/local/go/src/runtime/mheap.go:977 +0xc2
331runtime.(*mheap).alloc.func1()
332	/usr/local/go/src/runtime/mheap.go:1048 +0x4c
333runtime.(*mheap).alloc(0x1ea3f80, 0x4e, 0x7f82eb000101, 0x4ae3b0)
334	/usr/local/go/src/runtime/mheap.go:1047 +0x8a
335runtime.largeAlloc(0x9c000, 0x4e0100, 0xc1c7f2a000)
336	/usr/local/go/src/runtime/malloc.go:1055 +0x99
337runtime.mallocgc.func1()
338	/usr/local/go/src/runtime/malloc.go:950 +0x46
[...]

PR #1649, opened 2019-08-29
Temporarily disable broken link checker to unblock deploys from CI

justinvp:
Hugo is now consistently failing in Travis with `fatal error: runtime: out of memory` when running `hugo server` to host a local build of the site to run the broken link checker against. This is holding up deploys of the site.


justinvp:
#1648 tracks finding a solution and re-enabling

PR #1653, opened 2019-08-29
update python version of aws ec2 tutorial

danfhernandez:
### Proposed changes
Running through the python version of the ec2 webserver tutorial in it's current form doesn't work (https://www.pulumi.com/docs/tutorials/aws/ec2-webserver/).  The code is based on this example from https://github.com/pulumi/examples/tree/master/aws-py-webserver but doesn't include instructions to add the ami.py file that would let them get any ami.  The other languages hard-code the ami value so updating the python version to do the same.

### DIFF
--- a/content/docs/tutorials/aws/ec2-webserver.md
+++ b/content/docs/tutorials/aws/ec2-webserver.md
@@ -113,7 +113,7 @@ aws:region: (us-east-1)
     server = ec2.Instance('webserver-www',
         instance_type=size,
         security_groups=[group.name], # reference security group from above
-        ami=get_linux_ami(size))
+        ami='ami-0ff8a91507f77f867')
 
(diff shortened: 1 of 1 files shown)


content/docs/tutorials/aws/ec2-webserver.md
```diff
@@ -113,7 +113,7 @@ aws:region: (us-east-1)
     server = ec2.Instance('webserver-www',
         instance_type=size,
         security_groups=[group.name], # reference security group from above
-        ami=get_linux_ami(size))
+        ami='ami-0ff8a91507f77f867')
```
stack72: I don't think we should hardcode this - what region is this for? I would suggest we use the getAmi datasource to get the value instead? Just my $0.02

PR #1661, opened 2019-08-29
Add an OS chooser

### LINKED ISSUE #1394: Don't hide the install instructions for alternative platforms

joeduffy:
This change adds an OS chooser similar to what we have for languages.
This is then used for various installation pages to offer the end user
the ability to choose something other than the default that we've
detected -- in addition to making that choice sticky using cookies.

Most of the same code for the language chooser is refactored and
reused, although some of it had to remain specific to languages due
to the special case of hiding/showing code blocks.

Fixes #1394.


joeduffy:
For example, in the getting started guide:

image

And on the download and installation page:

image

PR #1664, opened 2019-08-30
Limit the number of overloads we try to render in docs

### LINKED ISSUE #1525: Listing all function overloads is unusable when there are many
[docs/reference/pkg/nodejs/pulumi/kubernetes/yaml/_index.md](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/kubernetes/yaml/)

lukehoban:
We have a few places in the Kubernetes API where we generated >1000 overloads for a function.  To keep these pages readable, put a limit on the number of overloads we attempt to render.

Fixes #1525.


tools/tscdocgen/main.go
```diff
@@ -320,6 +320,14 @@ func (e *emitter) augmentNode(node *typeDocNode, parent *typeDocNode, k8s bool)
 	for _, sig := range node.Signatures {
 		e.augmentNode(sig, node, k8s)
 	}
+	if len(node.Signatures) > 20 {
+		// To avoid places where we have 1000+ overloads from becoming unreadable, limit the lists
+		// of overloads to 20, and show a note that there are more aviailable instead of rendering
```
justinvp: Nit:

```suggestion
		// of overloads to 20, and show a note that there are more available instead of rendering
```

PR #1665, opened 2019-08-31
Edit bits of the new homepage

joeduffy:
This is a work in progress, however makes the following changes:

* Use a different heading/subheading.

* Make some fonts larger.

* Move the CTA above the fold and make it more prominent.

* Move Create/Deploy/Manage above the widget and style it differently.

* Change the text within Create/Deploy/Manage cards, and use
  a bulleted headings style for the major points.

* Add a SVG header to the Manage card for symmetry -- although
  it needs to be revised to include e.g. identity providers.

* Add real quotes.

* Add some new logos.


joeduffy:
@cnunciato this is a work in progress, curious what you think.

A few minor notes:

I am thinking of having an SVG for the Manage card, similar to Create and Deploy. Perhaps it could show four identity providers -- GitHub, GitLab, Atlassian, and SAML. Thoughts?

I am pondering whether we should make the quotes carousel animate/rotate. We currently only show a fixed number of quotes (three), and I think it would be nice to make room for the others. Alternatively, we could randomize which three we show upon refreshing the page.

joeduffy:
Another area I could use @cnunciato's help: I'd like to redo the animated carousels slightly.

- Bump the fonts a teensy bit.
- Use a slightly more compelling example, maybe we can use a for loop to populate the bucket?
- Include the preview with confirmation before updating.
- Also render that more compelling example in the service view (more than two objects -- and probably elide the "AWS providers" one). Include the AWS console links in the render.

content/_index.md
```diff
@@ -1,9 +1,10 @@
 ---
-heading: Program the cloud.
+heading: Modern Infrastructure as Code.
+
 
 subhead: >
-    **Modern infrastructure as code**. Create, deploy, and manage infrastructure on
-    any cloud, using your favorite language.
+    Use your favorite languages and tools. Empower your developers
```
stack72: See I'd really question if we should add this - I mean, what if my favourite language is C# / F#?

Wouldn't it be better to say something like, "Use real programming languages and tooling to empower your developers and operators to work better together" ?

cnunciato:
Hm. @joeduffy In the interest of forward progress, let’s see if we can identify the problems you’re having with what’s already been designed and submitted, and then I’ll follow up with whatever changes we decide to make against my PR. Sound fair?

content/_index.md
cnunciato: I second this. We should not say “your favorite language” until we support any language.

joeduffy:
Merging per our in-person discussion — will tidy up wording in @cnunciato's existing PR/branch (including the very good feedback above from @stack72 )

Issue #1492, opened 2019-08-01
Improved landing pages


bermudezmt:
### Problem description

- Add nicer intros and/or images (e.g. language and cloud provider icons)
- More visually appealing lists/tables
- Link to relevant content / next steps

- [X] Tutorials
- [X] Pulumi Console
- [X] Guides
- [X] Intro

lukehoban:
@bermudezmt What exactly is this issue tracking?  Which pages does it involve replacing/improving exactly?  

> Add a table with languages as the column headings, and package list as the rows

This is now https://github.com/pulumi/docs/issues/1580.

lukehoban:
This is tracking improving the top level nav pages like:
* https://www.pulumi.com/docs/reference/tutorials/

lukehoban:
@bermudezmt Anything more we're targeting here for "1.0" release?  Should we close this out and open new issues on specific things that we want to do next for landing pages?

bermudezmt:
Working on getting #1644 merged and then we can close this out.

lukehoban:
#1644 was merged, so I'll close this one out.  We still have opportunity for more landing page improvements, but at this point we can open issues for more specific improvements we want to make.

PR #1667, opened 2019-09-03
Add intro overview

bermudezmt:
Relates to #1492

### DIFF
--- a/content/docs/intro/_index.md
+++ b/content/docs/intro/_index.md
@@ -1,11 +1,38 @@
 ---
-title: Intro
+title: Intro to Pulumi
+linktitle: Overview
+
+menu:
+  intro:
+    weight: 1
 ---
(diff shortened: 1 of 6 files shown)


joeduffy:
This is amazing work! 🎉 I love it.

One question I wonder about: with this, we have two nav items on the left that say "Overview"; should we think about making the "Getting Started" and "Intro" headers clickable, and making these the landing pages? Or are we okay with the two titles being the same?

Don't block on this, of course, let's land this as-is!

bermudezmt:
@joeduffy I think it's okay as is, since "Reference" has an Overview link as well. Guides and Tutorials currently doesn't have an overview link since we have landing pages for "Guides" (captured in #1644 ) and "Tutorials" (which you've taken care of via #1620).

Issue #1595, opened 2019-08-20
Refine Python API docs styling


justinvp:
The styles for the API docs could use another round of refinement: fixing font sizes, indentation, etc., so it is more obvious where you on the page and the hierarchy of things.

- [x] TypeScript/JavaScript (#1669)
- [ ] Python

justinvp:
I made some very minor improvements in #1690 to Python, but it could use a couple more hours cleaning up the styling across the board. Going to move this out to M27 to finish up.

bgdnlp:
I was wondering about the status of this issue.

The Python docs are hard to follow, at least for someone new to Pulumi. The styling on Typescript pages looks much better. 

It would also be nice if at least the nav bar on the right, the one that says "On this page", wuold be in a separate pane from the center column. That way it could be scrolled independently. Helps with jumping around in long pages.

justinvp:
Closing this as a duplicate of https://github.com/pulumi/docs/issues/4615. We don't need this separate issue. We're tracking the work in https://github.com/pulumi/docs/issues/4615.

PR #1669, opened 2019-09-03
Resource-oriented Node.js API docs

### LINKED ISSUE #1597: Reorganize content within generated API docs pages
Improve the layout of existing API pages by better organizing the content within these pages. For example, grouping types by resources, data sources, etc. on the page (with links to these at the top of the page for easier access). This is a step in the direction of #1421, but without the potentially destabilizing aspects (new pages, redirects).

justinvp:
This reorganizes APIs within existing API doc pages, grouping by Resources, Data Source, and Other APIs.

## Notes

I deliberately removed the index sorted by type. Instead we have groupings of APIs: (Modules, Resources, Data Sources, Other)

I've also tried to ensure the "ON THIS PAGE" section to the right is more relevant with just the top-level listing of members.

Screen Shot 2019-09-03 at 2 57 05 PM

"Other" just lists all members that aren't resources/datasources -- and does so with a case insensitive sort of just the name (no longer includes the _type_ as part of the the sort). This category includes various `Args` and `Results` types -- I'd wanted to pull those in and list them under the associated Resource / Data Source -- but ran out of time. I still think that'd be worth doing and would cut down some of the cruft in this category.

If the only thing on a page ends up being "Other", then it will be displayed as "APIs" instead:

Screen Shot 2019-09-03 at 3 00 25 PM

I also deliberately removed the list of files. I personally don't find this useful, especially since the members themselves link to their location within the files on GitHub. But if we wanted to, they could easily be added back (though, wondering if they should go at the very bottom of the page).

## Remaining TODOS

 - [x] Address any feedback

 - [x] Regenerate _all_ docs. I didn't get a chance to regenerate all docs, but will update this PR with full regeneration of all docs. Right now this only contains changes for:
   - pulumi
   - aws
      - ~~I just noticed the `types` module didn't get generated; I've noticed that sometimes when I generate the docs it doesn't show up -- need to regenerate)~~ Update: I regenerated the docs for this same smaller set of packages and the `types` module is now present. It's really strange. I've noticed recently (even without these changes) that sometimes when I run `tsdocgen`, the `types` module isn't included, and other times it is. This happens even when running the tool on the exact same `aws.docs.json` file from TypeDoc. Investigating why this happens.
   - awsx
   - cloud
   - cloudflare
   - eks
   - kubernetes

- [x] If a class or interface extends/implements something else, we just show the "extends Foo". I want to update this to show the full type declaration, e.g. `class Bucket extends CustomResource`

Screen Shot 2019-09-03 at 3 07 12 PM

- [x] Do a better job listing members within namespaces. Right now it doesn't show up nicely and doesn't "nest" well.

  Screen Shot 2019-09-03 at 3 03 52 PM

Fixes #1597
Part of #1595


lukehoban:
Love the overall improvement - the index looks a **lot** better now!

> Do a better job listing members within namespaces. Right now it doesn't show up nicely and doesn't "nest" well.

FWIW - The [AWS versions of this](http://content-bucket-c36bfa6.s3-website-us-east-1.amazonaws.com/docs/reference/pkg/nodejs/pulumi/aws/types/input/#GetAmiFilter) look even more "different" than the rest of these changes:

Screen Shot 2019-09-03 at 5 05 55 PM

Note sure why these aren't getting the nice three column layout.  The lack of framing around each interface header now also makes the rest of the page a little harder to read for these input types (at least for AWS) - though this aspect isn't a huge deal:

Screen Shot 2019-09-03 at 5 07 28 PM

lukehoban:
Ahh - I just saw this comment which I think explains my feedback above:

> I just noticed the types module didn't get generated; I've noticed that sometimes when I generate the docs it doesn't show up -- need to regenerate)

justinvp:
> Ahh - I just saw this comment which I think explains my feedback above

Yeah, it's really strange. I've noticed recently (even without these changes) that sometimes when I run `tsdocgen`, the `types` module isn't included, and other runs it is. This happens even when running the tool on the exact same `aws.docs.json` file from TypeDoc. I'll look into why we're seeing this non-deterministic behavior.

lukehoban:
>  I've noticed recently (even without these changes) that sometimes when I run tsdocgen, the types module isn't included, and other runs it is. 

In fact, I thought I had seen the same thing once when initially adding these pages, but I convinced myself I must have done something else wrong between runs. Very strange indeed!

justinvp:
All the remaining TODOs are complete, other than addressing any review feedback. I generated the docs for a small subset of packages on top of these changes (`pulumi`, `aws`, `awsx`, `cloud`, `cloudflare`, `eks`, and `kubernetes`) and will regenerate **all** Node.js packages once any feedback is addressed. I can do it either before or after #1674 lands.

cnunciato:
Such an incredible improvement! Things just keep getting better. 🎉 

Little thing: visually, it looks like the items are sorted into columns, but since it's actually horizontally wrapped rows, it's a bit strange when you go looking for something specific -- e.g., I went looking for "cloudwatch", but didn't see it under "cloudformation". Totally something we could come back to, but thought it was worth mentioning.

justinvp:
> Little thing: visually, it looks like the items are sorted into columns, but since it's actually horizontally wrapped rows, it's a bit strange when you go looking for something specific -- e.g., I went looking for "cloudwatch", but didn't see it under "cloudformation".

@cnunciato, I'm glad you pointed this out because I'm definitely aware of this and totally agree. I'll fix because it's been bugging me.

PR #1673, opened 2019-09-04
Bump latest-version to 1.0.0

pgavlin:


ellismg:
This should have also updated the table of all versions.  I folded this in to 1674, @pgavlin

PR #1689, opened 2019-09-05
Apply Console, Service and backend naming changes

cnunciato:
We [recently agreed](https://pulumi.slack.com/archives/C7XHM1GNT/p1566057328001800) to standardize on the way we refer to the various parts of the Pulumi platform: 

* "Pulumi Console" for app.pulumi.com
* "Pulumi Service" for the service generally
* "Pulumi Service API" for the REST API specifically
* "Pulumi Service backend" for the service backend

This PR takes a pass through the site to apply those changes.

### DIFF
--- a/content/docs/reference/cli/pulumi_new.md
+++ b/content/docs/reference/cli/pulumi_new.md
@@ -14,12 +14,12 @@ To create a project from a specific template, pass the template name (such as `a
 or `azure-python`).  If no template name is provided, a list of suggested templates will be presented
 which can be selected interactively.
 
-By default, a stack created using the pulumi.com backend will use the pulumi.com secrets
+By default, a stack created using the Pulumi Service backend will use the Pulumi Service secrets
 provider and a stack created using the local or cloud object storage backend will use the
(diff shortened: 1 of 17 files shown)


content/docs/reference/cli/pulumi_new.md
```diff
@@ -14,12 +14,12 @@ To create a project from a specific template, pass the template name (such as `a
 or `azure-python`).  If no template name is provided, a list of suggested templates will be presented
 which can be selected interactively.
 
-By default, a stack created using the pulumi.com backend will use the pulumi.com secrets
+By default, a stack created using the Pulumi Service backend will use the Pulumi Service secrets
```
bermudezmt: I think the CLI descriptions have to be updated upstream at https://github.com/pulumi/pulumi/blob/master/cmd/new.go for example? Not sure if that has been done.

cnunciato: That’s right, yeah — I bypassed the generated docs for this, but I’m happy to have a look at that next!

PR #1690, opened 2019-09-05
Minor Python styling improvements

justinvp:
- Only show a single H1 per page
- Cleaner breadcrumb titles
- Minor styling adjustments
- Regenerate Python docs

### DIFF
--- a/assets/sass/_api-python.scss
+++ b/assets/sass/_api-python.scss
@@ -58,7 +58,7 @@ div.section[id]:not([id=""]) {
 
         dt {
             // Use the same styling as our `pre` tags for definitions.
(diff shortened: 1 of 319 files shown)


justinvp:
Here's an example of the style change. The Parameters, Returns, and Return Type are no longer styled like they're a code definition (making it easier to find actual definitions), and reduced the size of the code font to be consistent with the size we use elsewhere. Still isn't great, but a little better.

## Before

Screen Shot 2019-09-04 at 8 16 55 PM

## After

Screen Shot 2019-09-04 at 8 17 15 PM

justinvp:
I'm going to merge this TBR, just so it gets deployed now before tomorrow morning. Happy to address any feedback subsequently.

lukehoban:
Definitely an improvement. Looks like there’s still plenty of room to go further on styling (and content structure) here - but glad we can lock in these improvements first.

Issue #1685, opened 2019-09-04
Sometimes the `types` module is not generated for Node.js API docs


justinvp:
I've noticed recently that sometimes when I run` tsdocgen` on a package that should have a `types` module, the `types` module isn't generated. And sometimes it is. It's non-deterministic. This happens even when running the tool on the exact same JSON file from TypeDoc. Need to investigate why we're seeing this non-deterministic behavior.

justinvp:
The root cause of this is we're adding to the map of modules while iterating over it:

https://github.com/pulumi/docs/blob/6621dd78887f76ee181eeb1da6fdb6492e987cca/tools/tscdocgen/main.go#L708-L718

According to the [Go "for statements" spec](https://golang.org/ref/spec#For_statements):

> The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next. If a map entry that has not yet been reached is removed during iteration, the corresponding iteration value will not be produced. **If a map entry is created during iteration, that entry may be produced during the iteration or may be skipped. The choice may vary for each entry created and from one iteration to the next.** If the map is nil, the number of iterations is 0.

For example, we'll come across a module named `"types/input"` or `"types/output"`, and add the parent `"types"` to the map. Sometimes we see the `"types"` module that was added before the end of the iteration. Other times we do not, in which case it is never added to the `root` module's `Modules`.

PR #1694, opened 2019-09-05
Generate documentation for all providers for 1.0 release

stack72:


content/docs/reference/pkg/nodejs/pulumi/packet/_index.md
```diff
@@ -27,7 +27,6 @@ import * as packet from "@pulumi/packet";
 <h3>Modules</h3>
 <ul class="api">
     <li><a href="config/"><span class="symbol module"></span>config</a></li>
-    <li><a href="types/"><span class="symbol module"></span>types</a></li>
```
justinvp: Looks like another case of #1685. @stack72, could you try regenerating the docs for `packet` again until this line isn't removed? It's really strange behavior in the doc generator -- sometimes it happens, sometimes it doesn't. Going to look into this strange bug today.

PR #1701, opened 2019-09-06
tutorials: EC2 TypeScript tutorials were using JavaScript require / exports

stack72:
Also took the opportunity to make our webserver tutorials region agnostic
this is for JavaScript, TypeScript and Python

### DIFF
--- a/content/docs/tutorials/aws/ec2-webserver.md
+++ b/content/docs/tutorials/aws/ec2-webserver.md
@@ -57,7 +57,14 @@ aws:region: (us-east-1)
     const aws = require("@pulumi/aws");
 
     let size = "t2.micro";     // t2.micro is available in the AWS free tier
-    let ami  = "ami-0ff8a91507f77f867"; // AMI for Amazon Linux in us-east-1 (Virginia)
+    let ami = aws.getAmi({
+        filters: [{
+          name: "name",
+          values: ["amzn-ami-hvm-*"],
+        }],
+        owners: ["137112412989"], // This owner ID is Amazon
+        mostRecent: true,
+    });
 
     let group = new aws.ec2.SecurityGroup("webserver-secgrp", {
         ingress: [
@@ -68,7 +75,7 @@ aws:region: (us-east-1)
     let server = new aws.ec2.Instance("webserver-www", {
         instanceType: size,
         securityGroups: [ group.name ], // reference the security group resource above
-        ami: ami,
+        ami: ami.id,
     });
 
     exports.publicIp = server.publicIp;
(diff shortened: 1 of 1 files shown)


content/docs/tutorials/aws/ec2-webserver.md
```diff
@@ -55,66 +55,85 @@ aws:region: (us-east-1)
 
     ```javascript
     const aws = require("@pulumi/aws");
-
+
```
lukehoban: Extra spaces?

Plus looks like there are now two blank lines here - probably can be just one?

content/docs/tutorials/aws/ec2-webserver.md
```diff
@@ -55,66 +55,85 @@ aws:region: (us-east-1)
 
     ```javascript
     const aws = require("@pulumi/aws");
-
+    
     let size = "t2.micro";     // t2.micro is available in the AWS free tier
-    let ami  = "ami-0ff8a91507f77f867"; // AMI for Amazon Linux in us-east-1 (Virginia)
```
lukehoban: Given we no longer require us-east-1, can probably remove the comment above all this in the prose about making sure you are in us-east-1 as well.

content/docs/tutorials/aws/ec2-webserver.md
```diff
@@ -55,66 +55,85 @@ aws:region: (us-east-1)
...
+    
+    ami = aws.get_ami(most_recent="true",
+                      owners=["137112412989"],
+                      filters=[{"name":"name","values":["amzn-ami-hvm-*"]}])
+    
+    group = aws.ec2.SecurityGroup('webserver-secgrp',
+                              description='Enable HTTP access',
+                              ingress=[
+                                  { 'protocol': 'tcp', 'from_port': 22, 'to_port': 22, 'cidr_blocks': ['0.0.0.0/0'] }
+                              ])
```
lukehoban: I suspect this formatting will make this not fit horizontally on the screen given our narrow layout.  Worth reformatting to try to ensure the code is readable in the tutorial?

content/docs/tutorials/aws/ec2-webserver.md
```diff
@@ -55,66 +55,85 @@ aws:region: (us-east-1)
...
+    export const publicIp = server.publicIp;
+    export const publicHostName = server.publicDns;
     ```
 
     ```python
     import pulumi
-    from pulumi_aws import ec2
-    ami = 'ami-0ff8a91507f77f867' # AMI for Amazon Linux in us-east-1 (Virginia)
-
+    import pulumi_aws as aws
```
lukehoban: Honestly not sure - but I'll note that we've neraly universally gone with the existing `from pulumi_aws import ec2` style here in other Python examples.  I'm not sure we have to change things here - or should at least agree on what style guideline we're using and consistently change if we believe we are "wrong".

stack72: So the reason I did this was because I needed the root of the pulumi_aws import. There is no way (AFAICT) to do `from pulumi_aws import ` when it's the root package. So by importing it all, I wanted to add an alias

content/docs/tutorials/aws/ec2-webserver.md
stack72: Done!

content/docs/tutorials/aws/ec2-webserver.md
lukehoban: I'm fine with this as is - but FWIW - you could just do `from pulumi_aws import ec2, get_ami`.

PR #1705, opened 2019-09-09
Minor nav fixes

cnunciato:
We recently changed the behavior or the main nav to be fixed on the desktop, which introduced a little bug preventing the submenu from being displayed. This fixes that, and also adjusts the font size of the alert text on mobile, so it wraps a little less often.

### DIFF
--- a/assets/sass/styles.scss
+++ b/assets/sass/styles.scss
@@ -73,10 +73,6 @@ body {
 }
 
 .nav-main {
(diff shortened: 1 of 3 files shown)


cnunciato:
@justinvp Yeah, I see that. 🤔 A shadow of some sort might make that text look a little less translucent when it's. I'll see what I can do for it.

PR #1711, opened 2019-09-10
Fixup the ec2 webservice tutorial updates section

stack72:


justinvp:
@stack72, this needs to be updated now that the link was removed from the READM.md. Just re-run the `./scripts/mktutorial.sh`

PR #1712, opened 2019-09-10
Emit `types` module consistently

### LINKED ISSUE #1685: Sometimes the `types` module is not generated for Node.js API docs

justinvp:
Sometimes when generating the docs, we'd leave out the `types` module. Other times, it would be included.

This nondeterministic behavior was due to us modifying a map while iterating over it. According to the [Go "for statements" spec](https://golang.org/ref/spec#For_statements):

>  If a map entry is created during iteration, that entry may be produced during the iteration or may be skipped. The choice may vary for each entry created and from one iteration to the next.

This started becoming an issue when we recently started including `types` module. Previously, all other modules were parented under the root "index" module, which is always present. But now we have a `types` module which gets added during processing, and sometimes that entry would be included and other times it would be skipped during the iteration.

To address this, we now copy the entries to an array, which will be processed until it is empty, allowing for items to be added during the iteration.

Fixes #1685

### DIFF
--- a/tools/tscdocgen/main.go
+++ b/tools/tscdocgen/main.go
@@ -705,17 +705,31 @@ func (e *emitter) gatherModules(doc *typeDocNode, parentModule string, k8s bool)
 		root = newModule(rootModule)
 		mods[rootModule] = root
 	}
-	for modname, mod := range mods {
+
+	// Copy the names to an array and then process until empty. This allows us to add new names while iterating.
+	var modnames []string
+	for modname := range mods {
+		modnames = append(modnames, modname)
+	}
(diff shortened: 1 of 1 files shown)


justinvp:
I'm going to merge this to avoid having to re-do any upcoming docs generation PRs that were unlucky. Happy to address any feedback subsequently.

PR #1739, opened 2019-09-13
Improved type linking in awsx API docs

### LINKED ISSUE #1117: AWSX API docs don't link to AWS docs
Many of the AWSX resources use AWS resources, but the docs don't link between the two making it difficult and time-consuming to understand the full options needed for AWSX.

As a specific example, [FargateTaskDefinitionArgs#Container](https://pulumi.io/reference/pkg/nodejs/pulumi/awsx/ecs/#FargateTaskDefinitionArgs-container) for `awsx.ecs.FargateService` uses `aws.ecs.Container` but there's no link to the container docs and no `aws` namespace to provide a hint to readers on where to find it. 

image

The docs should cross-link between packages where needed.

justinvp:
Improved linking within the library as well as now linking to `@pulumi/aws` types.

Fixes #1117

### DIFF
--- a/tools/tscdocgen/main.go
+++ b/tools/tscdocgen/main.go
@@ -1188,7 +1188,7 @@ func (e *emitter) typeHyperlink(t *typeDocType) string {
 					link = "/docs/reference/pkg/nodejs/pulumi/pulumi/"
 				} else if e.pkgname == "awsx" && elements[0] == "aws" {
 					link = "/docs/reference/pkg/nodejs/pulumi/aws/"
-				} else if e.pkgname == "awsx" && elements[0] == "x" {
+				} else if e.pkgname == "awsx" && (elements[0] == "x" || elements[0] == "awsx") {
 					link = "/docs/reference/pkg/nodejs/pulumi/awsx/"
 				} else if elements[0] == "inputs" || elements[0] == "inputApi" {
 					link = "/docs/reference/pkg/nodejs/pulumi/" + e.pkgname + "/types/input/"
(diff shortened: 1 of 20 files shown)


tools/tscdocgen/main.go
```diff
@@ -1115,6 +1115,13 @@ func (e *emitter) createCodeDetails(node *typeDocNode) string {
 	}
 }
 
+// @pulumi/awsx imports types within the same modules using the following.
+// We'll use this mapping to map the local import to the actual top-level exported module.
+var awsxImportModuleMap = map[string]string{
+	"ecs.": "ecs",
+	"mod.": "lb",
```
CyrusNajmabadi: i'm not sure ths is always valid.  awsx may import the same module it is in using 'mod' in many different places.  Note: if there's a module import pattern we should be using instead, lmk and we can update awsx to consistently follow it .

justinvp: > i'm not sure ths is always valid. awsx may import the same module it is in using 'mod' in many different places. 

I checked awsx and it's currently only using `import * as mod from ".";` in `lb`.

> Note: if there's a module import pattern we should be using instead, lmk and we can update awsx to consistently follow it .

@CyrusNajmabadi, what do you think about importing the root, and using the same prototypical naming we'd use as a consumer of the library, when referring to public exported types within the library?

For example, instead of:

```typescript
import * as mod from ".";
import * as x from "..";
```

And referring to types using `mod.Foo` and `x.mod.Foo`, do the following instead:

```typescript
import * as awsx from "..";
```

And then always refer to public export types using the full `awsx.mod.Foo` name?

It does mean we'd have to be slightly more verbose in the implementation of the library when referring to such types, but since TypeDoc only provides the name of the type as written, this would allow us to easily do linking of types within a library without any special casing in the doc generator other than seeing the type name starts with `awsx.`.

PR #1740, opened 2019-09-13
Move first time notice to the page where pulumi is first run

rothgar:
### Proposed changes

Users are told to run `pulumi` on the config page but logins aren't talked about until the next page. I was confused by this so I thought I'd suggest moving it to the config page.

Might also be useful to put more information about features you get from logging in and if it's possible to use pulumi without logging in.

### Related issues (optional)

No issues. Just feedback from my personal experience going through the walkthrough.


bermudezmt:
Thanks for your feedback, @rothgar . @justinvp what do you think about moving this section up? I'd also want to change this to a `.note .note-info` instead of a blockquote. cc: @hausdorff

justinvp:
> I'd also want to change this to a `.note .note-info` instead of a blockquote.

@bermudezmt, SGTM. We can do it as a follow-up change.

PR #1743, opened 2019-09-16
mktutorial template improvements

justinvp:
- Fix left alignment of "Edit this Page" and "Request a Change"
- Remove "Edit this Page" for generated pages
- Add "WARNING: this was generated by a tool" comment to generated pages
- Use Tailwind for styles


justinvp:
> but the "Edit this page" link isn't really available on the pages where `no_edit_this_page` is set. Is that the intention?

Yes, the `no_edit_this_page` front matter can now be set to prevent the "Edit this page" from being displayed.

PR #1791, opened 2019-09-27
Fix issue (maybe, hopefully) with long descriptions for generated API docs pages

praneetloke:
### Proposed changes

As the title says, this PR is to address the issue with search engines pulling the auto generated content as the meta description of the page in search results.

I have regenerated the API docs for Packet to show the result of my changes as an example in a separate commit.

N/A

### Related issues (optional)
N/A


content/docs/reference/pkg/nodejs/pulumi/packet/_index.md
```diff
@@ -992,7 +993,7 @@ const pxe1 = new packet.Device("pxe1", {
     operatingSystem: "customIpxe",
     plan: "t1.small.x86",
     projectId: local_project_id,
-    userData: ignition_config_example.rendered,
+    userData: ignition_config_example.rendered.apply(rendered => rendered),
```
justinvp: I don't think this `apply` is needed anymore. Is your `tf2pulumi` up-to-date? (`tf2pulumi version` outputs `v0.5.1` for me).

praneetloke: Hmm yeah my `tf2pulumi` hasn't been updated in a while. Let me check and update.

praneetloke:
> Instead of listing the various categories in the description, why not something simpler?

Yep. That's fine with me. I just didn't have a strong opinion on what to have in the description. Just thought I'd start with something.

praneetloke:
By the way, I came across [this post](https://webmasters.googleblog.com/2019/09/more-controls-on-search.html), which has some tips that I think are relevant to the problem we are trying to solve with this PR. The new options won't be effective until late October according to the post, but something to keep in mind for later.

content/docs/reference/pkg/nodejs/pulumi/packet/_index.md
praneetloke: Done.

tools/tscdocgen/main.go
```diff
@@ -487,27 +487,41 @@ func (e *emitter) emitMarkdownModule(name string, mod *module, root bool) error
...
+	hasMembers := len(members) > 0
+	hasNamespaces := len(namespaces) > 0
+	hasResources := len(resources) > 0
+	hasDataSources := len(dataSources) > 0
+
+	metaDescription := "Explore "
+	if root {
+		metaDescription += "members of the " + linktitle + " package."
+	} else {
+		metaDescription += "members of the " + linktitle + " module in the " + e.pkgname + " package."
```
justinvp: Should this be `e.pkg` instead of `e.pkgname` so it shows up as `@pulumi/packet` instead of `packet`?

praneetloke: Sure thing. I'll change it.

PR #1807, opened 2019-10-03
Add the pricing page back to top nav

chrsmith:
Add the "Pricing" tab back to the top nav.

image

Part of https://github.com/pulumi/home/issues/572.

### DIFF
--- a/content/pricing/_index.md
+++ b/content/pricing/_index.md
@@ -3,4 +3,7 @@ title: Pricing
 meta_desc: Pulumi's open source infrastructure as code SDK and enterprise SaaS products together provide plans for teams of all sizes.
 type: page
 layout: pricing
+menu:		
+    header:		
+        weight: 3
 ---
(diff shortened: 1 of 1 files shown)


content/pricing/_index.md
```diff
@@ -3,4 +3,7 @@ title: Pricing
 meta_desc: Pulumi's open source infrastructure as code SDK and enterprise SaaS products together provide plans for teams of all sizes.
 type: page
 layout: pricing
+menu:		
+    header:		
+        weight: 3
```
cnunciato: Can we make this a `2`? That'll position it closer to Products, as it was before. (Why Pulumi kind of scooted over just as we launched the home-page changes.)

chrsmith: Done. I just kept the weight that it was before, but it looks like `/docs/_index.md` already had a weight of 3, and no menu item had a weight of 2. Does that sound right to you?

cnunciato: Yep, that sounds right.

PR #1816, opened 2019-10-05
Clarify Team Starter pricing

joeduffy:
I've had a number of inquiries from folks where they thought
Team Starter was $50 *per user*. It is actually $50 for 3 users.
Hopefully this new text will clarify the matter at a glance.


joeduffy:
@cnunciato This is the change we discussed a few weeks back. I got another email asking about this last night, so figured it's worth making the small change to clarify the situation.

PR #1817, opened 2019-10-06
Fix invalid bullet points in `stack init`

kenfdev:
Hi! I'm new to pulumi and started referencing the docs. Just found a part where it seems the bullet points aren't rendered correctly.

Wasn't sure how I can scan for other syntax mistakes, so here's a tiny fix for the `stack init` page.

### Proposed changes

This fixes the issue where the bullet points aren't rendered correctly
by hugo.

**Current(https://www.pulumi.com/docs/reference/cli/pulumi_stack_init/)**
image

**Fixed**
image

### DIFF
--- a/content/docs/reference/cli/pulumi_stack_init.md
+++ b/content/docs/reference/cli/pulumi_stack_init.md
@@ -22,9 +22,11 @@ provider and a stack created using the local or cloud object storage backend wil
 `--secrets-provider` flag.
 
 To use the `passphrase` secrets provider with the pulumi.com backend, use:
+
 * `pulumi stack init --secrets-provider=passphrase`
 
 To use a cloud secrets provider with any backend, use one of the following:
+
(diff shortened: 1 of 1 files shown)


lukehoban:
Thanks @kenfdev!

These docs are auto generated from the CLI help text at https://github.com/pulumi/pulumi/blob/master/cmd/stack_init.go#L36. So direct changes here will be overwritten if this is not also addresses in the original source or in the generator. It would be good to see if we can fix in the original source linked above and regenerate to ensure these changes stick.

kenfdev:
d'oh! Sorry about the wrong fix @lukehoban . So it looks like I'll have to build the CLI to see if the changes are appropriate. Let me give it a try 👍 

So if I succeed in changing the docs, I should PR to the pulumi/pulumi repo, correct? Thank you for your advice!

pgavlin:
This has been fixed with current releases.

PR #1854, opened 2019-10-16
Fix Azure blob docs

justinvp:

### DIFF
--- a/content/docs/intro/concepts/state.md
+++ b/content/docs/intro/concepts/state.md
@@ -238,7 +238,7 @@ pulumi login --cloud-url gs://my-pulumi-state-bucket
 ##### Azure Blob Storage
 
 ```
-pulumi login --cloud-url az://my-pulumi-state-bucket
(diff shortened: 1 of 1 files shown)


mikhailshilkov:
We also need to describe the environment variables involved, e.g. Azure requires `AZURE_STORAGE_ACCOUNT` and `AZURE_STORAGE_KEY`, probably AWS needs ones too?

`--cloud-url` parameter name doesn't seem to be required: `pulumi login azblob://state` works too. Should we remove it from the call?

justinvp:
> We also need to describe the environment variables involved, e.g. Azure requires `AZURE_STORAGE_ACCOUNT` and `AZURE_STORAGE_KEY`, probably AWS needs ones too?

Want to submit a separate PR?

> `--cloud-url` parameter name doesn't seem to be required: `pulumi login azblob://state` works too. Should we remove it from the call?

A paragraph or so below says:

> You may omit `--cloud-url` or `-c` when logging in to a remote backend and just
use `pulumi login s3://my-pulumi-state-bucket`.

PR #1877, opened 2019-10-22
Update config.md

ctison:


stack72:
Thanks for picking up this error @ctison :)

Issue #1879, opened 2019-10-23
Clarify full list of supported cloud providers


clstokes:
### Problem description

The [Cloud Providers](https://www.pulumi.com/docs/intro/cloud-providers/) page only lists a subset of the full list of providers that Pulumi supports - e.g. it lists AWS, Azure, Google Cloud, Kubernetes, OpenStack. The full list of providers that pulumi supports is at [API Reference](https://www.pulumi.com/docs/reference/pkg/) - e.g. AWS, Azure, and many more. This can lead to end users thinking Pulumi doesn't support their providers or services when in fact it does.

### Suggestions for a fix

The [Cloud Providers](https://www.pulumi.com/docs/intro/cloud-providers/) page should contain the full list of providers that Pulumi supports.

stack72:
@clstokes OMG that's funny - yesterday I talked to someone at DevOpsDays about vSphere support and they didn't see it was supported as it wasn't listed on that page :) Nice to see someone else pick this up too

stack72:
@clstokes I just spoke to @lukehoban about this - I will drive this forward and make sure we get the page updated with what we support

clstokes:
@stack72 we should include F5, Vault, etc. on this page too, not just _cloud_ providers.

stack72:
@clstokes my thoughts are that we have cloud providers, networking providers etc.

clstokes:
@stack72 yea, **_all_** providers.

stack72:
Oh @clstokes i mean in different sections - if you think they should all be listed here then I’m fine with that

clstokes:
@stack72 sure, I'm fine with different sections. I just want all of them listed on the page.

danfhernandez:
+1 here, I ran into this when trying to show all the providers we have during demo for MS... I would think it should include the list here: https://www.pulumi.com/docs/reference/pkg/

lukehoban:
What's still remaining here?  

We may want to hand this over to @spara or someone else who could help "spruce up" this page (logos, etc.).

stack72:
Closed via #2298

PR #1893, opened 2019-10-25
Adding Packet.net to the list of CloudProviders supported by Pulumi

stack72:
Related: #1879


stack72:
Added the provider configuration options to the provider README as well https://github.com/pulumi/pulumi-packet/pull/40

PR #1895, opened 2019-10-25
Adding Linode to the list of CloudProviders supported by Pulumi

stack72:
Related: #1879


stack72:
Also added the provider configuration to the provider README https://github.com/pulumi/pulumi-linode/pull/31

PR #1911, opened 2019-10-31
Add documentation for Pulumi 1.2 and 1.3 features

### LINKED ISSUE #1910: Document Pulumi 1.2 and 1.3 additions
1.2:
- [ ] Tracing (it's technically covered in CLI docs, but would be nice to add a note to troubleshooting pages)

1.3: 
- [ ] Transformations

cc @pgavlin in case I missed anything.

lukehoban:
Adds documentation for:
1. The `transformations` resource option
2. The `--trace` flag for capturing performance traces

As part of (2) also expanded the troubleshooting page with more
diagnostics tips.

Fixes #1910.

### DIFF
--- a/content/docs/intro/concepts/programming-model.md
+++ b/content/docs/intro/concepts/programming-model.md
@@ -372,7 +372,7 @@ vpc, _ := ec2.Vpc(ctx, "vpc", &VpcArgs{}, pulumi.ResourceOpt{Provider: provider}
 ```
 
 ###### `transformations`
-A list of transformations to apply to the resource and all it's children. This can be used to override or modify the inputs to child resources of a component, for example to add other resource options (like `ignoreChanges` or `protect`) or to modify an input property (like adding to `tags` or changing a property which is not configurable via the component directly).  Transformations can also be applied to all resources in a stack using `pulumi.runtime.registerStackTransformation`.  Transformations are passed the resource type, name, input properties resource options and the resource instance itself.  They can optionally return a new set of resource input properties and resource options which will be used to construct the resource.
+A list of transformations to apply to the resource and all of its children. This can be used to override or modify the inputs to child resources of a component, for example to add other resource options (like `ignoreChanges` or `protect`) or to modify an input property (like adding to `tags` or changing a property which is not configurable via the component directly).  Transformations can also be applied to all resources in a stack using `pulumi.runtime.registerStackTransformation`.  Transformations are passed the resource type, name, input properties resource options and the resource instance itself.  They can optionally return a new set of resource input properties and resource options which will be used to construct the resource.
 
 {{< langchoose >}}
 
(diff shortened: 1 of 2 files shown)


content/docs/intro/concepts/programming-model.md
```diff
@@ -371,6 +371,57 @@ provider, _ := aws.Provider(ctx, "provider", { region: "us-west-2" });
 vpc, _ := ec2.Vpc(ctx, "vpc", &VpcArgs{}, pulumi.ResourceOpt{Provider: provider});
 ```
 
+###### `transformations`
+A list of transformations to apply to the resource and all it's children. This can be used to override or modify the inputs to child resources of a component, for example to add other resource options (like `ignoreChanges` or `protect`) or to modify an input property (like adding to `tags` or changing a property which is not configurable via the component directly).  Transformations can also be applied to all resources in a stack using `pulumi.runtime.registerStackTransformation`.  Transformations are passed the resource type, name, input properties resource options and the resource instance itself.  They can optionally return a new set of resource input properties and resource options which will be used to construct the resource.
```
ericrudder:
```suggestion
A list of transformations to apply to the resource and all of its children. This can be used to override or modify the inputs to child resources of a component, for example to add other resource options (like `ignoreChanges` or `protect`) or to modify an input property (like adding to `tags` or changing a property which is not configurable via the component directly).  Transformations can also be applied to all resources in a stack using `pulumi.runtime.registerStackTransformation`.  Transformations are passed the resource type, name, input properties resource options and the resource instance itself.  They can optionally return a new set of resource input properties and resource options which will be used to construct the resource.
```

content/docs/troubleshooting/_index.md
```diff
@@ -17,6 +17,64 @@ First thing's first, we are always happy to hear from you and will try to help.
...
 community of users, are there to help out. Any and all questions are welcome!
 
+We also encourage everyone to contribute to the [Pulumi open source
+projects](https://github.com/pulumi) by [opening new
+issues](https://github.com/pulumi/pulumi/issues/new) and upvoting existing issues they care about.
+
+## Diagnosing Issues
+
+There are a few tools available to get additional diagnostics on how your Pulumi program is
+behaving.  These can be useful for self-diagnosing issues, and for sharing details of issues as part
```
ericrudder:
```suggestion
behaving.  These can be useful for self-diagnosing issues, and for sharing details as part
```

content/docs/troubleshooting/_index.md
```diff
@@ -17,6 +17,64 @@ First thing's first, we are always happy to hear from you and will try to help.
...
+
+```
+$ TF_LOG=TRACE pulumi up --logtostderr -v=9 2> out.txt
+```
+
+### Performance
+
+If you are seeing unexpectedly slow performance, you can gather a trace to understand what
+operations are being performed throughout the deployment and what the long poles are for your
+deployment.  In most cases, this will be the provisioning of one or more resources in your cloud
```
ericrudder:
```suggestion
deployment.  In most cases, the most time consuming operations will be the provisioning of one or more resources in your cloud
```

content/docs/troubleshooting/_index.md
```diff
@@ -17,6 +17,64 @@ First thing's first, we are always happy to hear from you and will try to help.
...
+provider.  But there may be cases where Pulumi itself is doing work that is limiting the performance
```
ericrudder:
```suggestion
provider, however, there may be cases where Pulumi itself is doing work that is limiting the performance
```

PR #1917, opened 2019-11-01
Tidy up the 404 page

### LINKED ISSUE #1417: Browser-specific bugs on the 404 page
Our long-standing 404 page with the animated clouds has a couple of problems with it these days. (https://www.pulumi.com/404)

- On Safari, for some reason the top-left cloud flashes for a moment and then disappears. Leading to an odd light ray coming from nowhere.

image

- On Chrome, that "light ray" seems to expand the page document's size. So EPIC scroll bars appear, which definitely doesn't work right.

image

- Finally, while the simple text on the page is nice, we might want to spruce it up ever so slightly if we can think of something that would be an improvement.

cnunciato:
This addresses a few browser-specific issues on the 404 page and brings it into closer visual alignment with what we show in the Console:

image

We could definitely do more to reduce duplication here, and it'd probably be good to lay this page out with Tailwind, and fix the disappearing-cloud-in-Safari bug in the Console as well, so I'll file a couple of follow-up issues for that work.  

Fixes #1417.

### DIFF
--- a/assets/sass/404.scss
+++ b/assets/sass/404.scss
@@ -1,7 +1,8 @@
-@import url('https://fonts.googleapis.com/css?family=Oxygen');
+@import "colors";
+@import "fonts";
 
 body {
-    background-color: #7c4792;
+    background-color: map-get($purple, 500);
 }
 
 $base-header-height: 3.779vw;
@@ -17,9 +18,11 @@ $position-left: 27vw;
 
 // The content is put inside of #clouds-and-beams.
 #clouds-and-beams {
-    color: #fcf7e7;
-    font-family: Oxygen;
+    color: map-get($gray, 100);
     line-height: 1.1;
+    font-family: "Open Sans", Arial, Helvetica, sans-serif;
+    font-size: 1.4em;
+    margin-top: 48px;
 
     // The positioning of the #clouds-and-beams div is absolute, with the actual location being
     // dependent up media queries.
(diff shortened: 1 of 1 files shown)


assets/sass/404.scss
```diff
@@ -17,9 +18,11 @@ $position-left: 27vw;
 
 // The content is put inside of #clouds-and-beams.
 #clouds-and-beams {
-    color: #fcf7e7;
-    font-family: Oxygen;
+    color: map-get($gray, 100);
     line-height: 1.1;
+    font-family: "Open Sans", Arial, Helvetica, sans-serif;
```
chrsmith: Any reason why we define a custom font family here and not the existing font stack we use in `pulumi.scss`? (Is this any different from what we serve normally?)

cnunciato: Yep -- it's only because we define the font stack for the rest of the site [in tailwind.config.css](https://github.com/pulumi/docs/blob/dd6043bad71491342c7309118b8d835786b159d2/assets/config/tailwind.config.js#L80-L83), and this page is currently divorced from the Tailwind stuff, so it doesn't have access to any of those classes. 

I started a WIP branch to refactor this page into a layout built with Tailwind and the rest of our new CSS, but it ended up a little bigger that I wanted to bite off today, so I made the necessary changes for the issue, and figured we could come back later to do the rest.

PR #1922, opened 2019-11-03
Update iam.md

pierskarsenbarg:
Updated arguments in example.

See example in reference docs [here](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/aws/iam/#GroupMembership)

### DIFF
--- a/content/docs/aws/iam.md
+++ b/content/docs/aws/iam.md
@@ -69,8 +69,8 @@ const user = new aws.iam.User("myuser");
 const group = new aws.iam.Group("mygroup");
 
 const policyAttachment = new aws.iam.PolicyAttachment("mypolicyattachment", {
-    users: [user],
-    groups: [group],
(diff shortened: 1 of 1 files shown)


stack72:
Hey @pierskarsenbarg 

Sorry for the delay here - I have not been working all week. So I don't believe the name is needed here. The API looks as follows:

```
/**
     * The role(s) the policy should be applied to
     */
    readonly roles?: pulumi.Input<pulumi.Input<Role>[]>;
    /**
     * The user(s) the policy should be applied to
     */
    readonly users?: pulumi.Input<pulumi.Input<User>[]>;
```

As you notice, it takes the user and role objects not just the name. So passing the name would not be correct and would break the API

Thoughts?

Paul

stack72:
Ping @pierskarsenbarg :)

PR #1931, opened 2019-11-06
Prevent header and carousel ide from overflowing page container

zchase:
The IDE in the home page carousel was causing a white gap to appear on the right side of the site when the screen was smaller than ~1300px in width. The header was also spilling over the page container.

ideGap
whitegap

This also includes some alignment and responsive fixes for the actual carousel elements. Some of the "popups" and text were getting cut off and/or lost their alignment as the screen got smaller. I also flipped the text and media of the carousel section when the screen gets smaller to avoid content jumping around as much when the carousel changes.

### DIFF
--- a/assets/sass/_carousel.scss
+++ b/assets/sass/_carousel.scss
@@ -1,37 +1,24 @@
 .homepage-carousel--container {
-	overflow: hidden;
-	.carousel-text--container {
-		height: 520px;
-		@media (min-width: 769px) and (max-width: 1030px) {
-			height: 365px;
-		}
-		@media (min-width: 600px) and (max-width: 768px) {
-			height: 440px;
-		}
-	}
-	@media (max-width: 1030px) {
-		max-width: 1023px;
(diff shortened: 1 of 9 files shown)


zchase:
@cnunciato I updated the PR to use utility classes when possible. One thing I noticed is that if I want to set a specific height on the element I have to use CSS for that instead of a Tailwind utility class. Same thing with absolute positioning. I also had to leave some custom media queries so the console/ide plays well as the screen changes.

One thing I am unclear on is, should I move the utility classes from the `.html` to a `.scss` file to keep everything in one place?

cnunciato:
This was actually fixed with #2192 (sorry! I didn't realize this was still open). Ultimately all that was needed was to hide the overflow and tweak the sizings of things a bit to prevent quite so much overflow. This can probably be closed, unless I missed something somewhere.

zchase:
@cnunciato Sweet deal. Thanks!

PR #1952, opened 2019-11-08
Embellish the .NET landing page a bit

joeduffy:
This change adds a few things to the .NET landing page:

- Add the .NET Core logo.

- A bit more about "why you might care."

- Add a basic code snippet.

- Add a Getting Started section. We want to drive people here
  when first landing on this page as much as possible, since it's
  the best experience (despite it being C#-only).

- Use the language switcher for C#/F#/VB templates.

- Reorder some content so most relevant is above the fold.

- Add CI/CD coverage of Azure DevOps pipelines and GitHub Actions.


joeduffy:
@lukehoban we have a special place for your video if/when it's ready 😄

joeduffy:
Also:

- Rename all occurrences of "Pulumi .NET" or "Pulumi.NET" with "Pulumi for .NET," per the .NET team's request.

joeduffy:
> at MS, i only used .NET since it seemed the safest ... you seem to really want to use CORE in places, so i didn't erase.

Yeah, Microsoft uses "Core" very aggressively now-a-days. It is also widely recognized in the community as meaning "the modern, open source, and cross platform .NET" -- not the old clunky desktop .NET that is effectively being deprecated. Don't shoot the messenger :-)

mikhailshilkov:
> Yeah, Microsoft uses "Core" very aggressively now-a-days

FWIW, Microsoft is killing "Core" in the next version. It's going to be ".NET 5".

PR #1955, opened 2019-11-11
Fix the C# getting started guides

joeduffy:
These guides didn't actually compiled. I guess the APIs may have
changed since writing them, but there were also syntax issues --
like using single quotes instead of doubles -- and semantic
issues -- like referring to incorrect types given the way the
using statements have been laid out. I've verified these compile.

### DIFF
--- a/content/docs/get-started/aws/modify-program.md
+++ b/content/docs/get-started/aws/modify-program.md
@@ -90,25 +90,25 @@ pulumi.export('bucket_name',  bucket.id)
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Pulumi;
-using Pulumi.Aws.Kms;
-using Pulumi.Aws.S3;
+using Aws = Pulumi.Aws;
 
 class Program
 {
-    static Task Main() =>
-        Deployment.Run(() =>
+    static Task Main()
+    {
+        return Deployment.RunAsync(() =>
         {
             // Create a KMS Key for S3 server-side encryption
-            var key = new Kms.Key('my-key');
+            var key = new Aws.Kms.Key("my-key");
 
             // Create an AWS resource (S3 Bucket)
-            var bucket = new S3.Bucket('my-bucket', new S3.BucketArgs
+            var bucket = new Aws.S3.Bucket("my-bucket", new Aws.S3.BucketArgs
             {
-                ServerSideEncryptionConfiguration = new S3.BucketServerSideEncryptionConfigurationArgs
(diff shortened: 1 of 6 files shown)


joeduffy:
@CyrusNajmabadi @mikhailshilkov @lukehoban I'm unsure what the idiomatic way to import and refer to these types is, especially the inputs. For now, I've used fully qualified references in the code — this is what we do in TypeScript and is approximately how the code was already written.

mikhailshilkov:
@joeduffy I *think* importing each namespace is idiomatic. That's what the tooling will do by default.

`Aws.Kms.Key` reads really nice, while `Aws.S3.Inputs.BucketServerSideEncryptionConfigurationRuleArgs` is ugly. It could be fine to pick the style case-by-case, unless it becomes too inconsistent.

joeduffy:
Thanks @mikhailshilkov, will update these to use the import style.

As an aside, do you think there's ever hope for anonymous types to work here? No matter how we qualify the type names, a lot of those input types are very ugly and verbose.

For example, wish I could write

```csharp
ApplyServerSideEncryptionByDefault = new {
    SseAlgorithm = "aws:kms",
    KmsMasterKeyId = key.Id,
},
```

instead of

```csharp
ApplyServerSideEncryptionByDefault = new Aws.S3.Inputs.BucketServerSideEncryptionConfigurationRuleApplyServerSideEncryptionByDefaultArgs {
    SseAlgorithm = "aws:kms",
    KmsMasterKeyId = key.Id,
},
```

But I get an anonymous type implicit conversion error when I try.

lukehoban:
> As an aside, do you think there's ever hope for anonymous types to work here? 

@cyrusnajmabadi had a proposal for how to simplify this. Cyrus - is there an issue open tracking that?  Or is it part of pulumi/pulumi#3407?

mikhailshilkov:
@CyrusNajmabadi looked at this and the conclusion was that we'd need to generate types like `InputBucketServerSideEncryptionConfigurationRuleApplyServerSideEncryptionByDefaultArgs` instead of `Input<BucketServerSideEncryptionConfigurationRuleApplyServerSideEncryptionByDefaultArgs >` because the import part prevents the short-hard syntax.

> I.e. this won't work trivially:    foo = { bar = 1, baz = 2 }
> that's because foo is an Input<Foo>.  But, of course, Input has no bar or baz props
> so we really need to generate an entirely new InputFoo type
> and have it have those lifted props
> and then ensure that InputFoo and Input<Foo> didn't have horrible problems if you tried to use 
> them together :slightly_smiling_face:
> that sort of tning

PR #1966, opened 2019-11-13
Update docs for PaC API changes

justinvp:
Don't merge until we make release a new version of `@pulumi/policy` tomorrow.

### DIFF
--- a/content/docs/get-started/policy-as-code/authoring-a-policy-pack.md
+++ b/content/docs/get-started/policy-as-code/authoring-a-policy-pack.md
@@ -29,7 +29,7 @@ menu:
     Once you're done editing your Policy Pack, run `pulumi policy publish <org-name>/<policy-pack-name>` to publish the pack.
     ```
 
-1. Tweak the Policy Pack in the `index.ts` file as desired. The existing policy in the template (which is annotated below) mandates that an AWS S3 bucket not have public read or write permissions enabled. A Policy Pack can contain up to 25 Policies. Each Policy must have a unique name, an enforcement level, and a validation function. Here we use `validateTypedResource` that allows us to validate S3 Bucket resources.
+2. Tweak the Policy Pack in the `index.ts` file as desired. The existing policy in the template (which is annotated below) mandates that an AWS S3 bucket not have public read or write permissions enabled. Each Policy must have a unique name, an enforcement level, and a validation function. Here we use `validateTypedResource` that allows us to validate S3 Bucket resources.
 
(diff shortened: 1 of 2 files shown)


content/docs/get-started/policy-as-code/authoring-a-policy-pack.md
```diff
@@ -29,7 +29,7 @@ menu:
     Once you're done editing your Policy Pack, run `pulumi policy publish <org-name>/<policy-pack-name>` to publish the pack.
     ```
 
-1. Tweak the Policy Pack in the `index.ts` file as desired. The existing policy in the template (which is annotated below) mandates that an AWS S3 bucket not have public read or write permissions enabled. A Policy Pack can contain up to 25 Policies. Each Policy must have a unique name, an enforcement level and at least one rule. Here we use a `typeRule` that allows us to create an assertion against all S3 resources.
+1. Tweak the Policy Pack in the `index.ts` file as desired. The existing policy in the template (which is annotated below) mandates that an AWS S3 bucket not have public read or write permissions enabled. A Policy Pack can contain up to 25 Policies. Each Policy must have a unique name, an enforcement level, and a validation function. Here we use `validateTypedResource` that allows us to validate S3 Bucket resources.
```
ekrengel: `A Policy Pack can contain up to 25 Policies.` We can remove this now!

content/docs/get-started/policy-as-code/core-concepts.md
```diff
@@ -13,17 +13,17 @@ menu:
 
 A Policy contains specific logic you would like to enforce. For example, you may want to restrict the creation of public S3 buckets or you may disallow resource provisioning without tags. You can refer to other examples [here](https://github.com/pulumi/examples/tree/master/policy-packs).
 
-Policy rules are written as functions that are evaluated against all resources in your Pulumi stack. If the function throws an `AssertionError` (via the assert package), the rule will consider the associated resource to be in violation of the policy.
+Policies are written as validation functions that are evaluated against all resources in your Pulumi stack. If the validation function calls `reportViolation`, the associated resource will be considered in violation of the policy.
```
ekrengel: Should we state that you can report multiple violations?

justinvp:
> although wondering if we should add anything about stack validations? And if so maybe we add the concept of resource vs stack validations to the "Core Concepts" page?

Will do this as a follow-up. It's tracked in https://github.com/pulumi/pulumi-policy/issues/132

PR #1967, opened 2019-11-13
Kubernetes Product Page formatting

zchase:
This PR refactors our `Kubernetes` product page to include additional CTAs, a small overview of Pulumi's Kubernetes offering, and a form to request a demo. Some elements have also been repositioned on the page or removed entirely. I will explain my reason behind the changes below. To start please take a look at the below image that provides the framework of the page:

Untitled drawing

#### HARD CTA
This is the hero section of the page. In the current version of the product page there is an IDE and some text next to it. In this new version we introduce the same buttons a user would use to register for the Pulumi SaaS application. We offer this "Hard CTA" for users that are immediately ready to convert and start using the product. Not everyone is going to want to go through our "Getting Started" guides. Some people will just want to start experimenting on their own and we should offer a pathway these users.

There is also an additional section on Crosswalk for Kubernetes. I copy/pasted from the Crosswalk docs for this, so any feedback is much appreciated here.

#### SOCIAL PROOF
This section contains our customer logos. This provides a bit of proof and backing that we have a "real" solution that provides actual value. "Other companies are using this product, so it is ok if I do too."

#### Product Overview
This section contains a brief overview of Pulumi's Kubernetes offering. This replaces the old `What is Kubernetes` and `Kubernetes Everywhere` sections. The first section was removed because it was just noise on the page. A brief two sentence overview of Kubernetes really doesn't provide value to anyone. If you are already knowledgable about Kubernetes then you probably don't care about the section, if you are not knowledgable about Kubernetes, the two sentence overview isn't very helpful. The latter section was more or less transformed into the new section. The section attempts to answer the question "What is Pulumi's Kubernetes solution?"

#### Examples
This section was carried over from the current page, but I've reduced the amount of examples significantly. We will now only show we support existing .yaml and we are multi cloud. The danger of including more examples is that if we don't include examples for every use case people might get the sentiment that we don't support their use case. In my experience it is best to keep examples in sales/marketing assets as generalized as possible. We don't want to set any expectations that we do or don't support use cases before a prospect/user engages with us. I've also updated the text of the CTAs to say "Next Step" to give the user the feel they have already "started." The link has also been updated to go directly to the Kubernetes path of getting start to avoid any redundant link clicking a user has to perform.

At the bottom of the examples there is a "How Pulumi Works" section. This is to give the user an idea of what happens after I write the code and provide a high level understanding of how Pulumi works. You'll notice I've eliminated the logos in the section. This was to reduce the amount of noise in the section so the user will focus more on the content and not the cool logos. 

#### Getting Started
This section is pretty straightforward and just provides some links to the "Getting Started" journey for a few of the cloud providers. We might want to consider adding additional cloud providers in the future, but for now this is a good start.

#### Request Demo
The final section of the page is a request demo form. This is essentially a last ditch effort. If you were not ready to create an account or do a "Getting Started" guide, then we need to offer you a way to engage us. The form includes an "Interest" field with values of either "Personal" or "Professional". This helps us sort through the request, where "Personal" requests can be sent automated follow up with documentation links, while "Professional" requests can be handled by the sales team. There are going to be cases where organizations are not able to sign up for our SaaS service and try out the product.

Hopefully the above helps explain my thought process when making these changes. If you have any questions or suggestions please do not hesitate to ask me. Below I've include a few links I used as inspiration for refactoring this page:

- https://segment.com/product/connections/
- https://clearbit.com/marketing

### DIFF
--- a/assets/js/tracking.js
+++ b/assets/js/tracking.js
@@ -1,0 +1,42 @@
+/*
+ * Function for registering tracking to a link. To add tracking to a link you
+ * need to have an "id" attritube on the <a> element and the "track-click" attribute.
+ *
+ *   - ex. <a id="get-started-no-yaml" class="btn" href="/docs/get-started" track-click>GET STARTED</a>
+ */
+$(document).ready(function() {
+
+    // Check if the analytics object and track function are available. If they are
+    // not we do not even want to attempt to track anything.
+    if (window && window.analytics && typeof window.analytics.track === "function") {
+
+        // Find all the links with a "track-click" attribute.
+        const buttons = $("a[track-click]");
+
+        function registerTracker(element) {
+
+            // Create the tracking object.
+            const trackingData = {
+                // The id of the element.
+                id: $(element).attr("id"),
+                // The destination url of the link.
+                destinationPath: $(element).attr("href"),
+                // The current path.
+                url: window.location.pathname,
+            }
+
+            // Register a listener to the link to send data to Segment
+            // when it has been clicked.
+            $(element).on("click", function(e) {
+                window.analytics.track("link-click", trackingData);
+            });
+
+        }
+
+        // Loop over the array of elements to register the click listeners.
+        for (var i = 0; i < buttons.length; i++) {
+            registerTracker(buttons[i]);
+        }
+    }
+
+});
--- a/assets/js/tracking.js
+++ b/assets/js/tracking.js
@@ -5,7 +5,6 @@
  *   - ex. <a id="get-started-no-yaml" class="btn" href="/docs/get-started" track-click>GET STARTED</a>
  */
 $(document).ready(function() {
-
     // Check if the analytics object and track function are available. If they are
     // not we do not even want to attempt to track anything.
     if (window && window.analytics && typeof window.analytics.track === "function") {
@@ -14,7 +13,6 @@ $(document).ready(function() {
         const buttons = $("a[track-click]");
 
         function registerTracker(element) {
-
             // Create the tracking object.
             const trackingData = {
                 // The id of the element.
@@ -30,7 +28,6 @@ $(document).ready(function() {
             $(element).on("click", function(e) {
                 window.analytics.track("link-click", trackingData);
             });
-
         }
 
         // Loop over the array of elements to register the click listeners.
@@ -38,5 +35,4 @@ $(document).ready(function() {
             registerTracker(buttons[i]);
         }
     }
-
 });
--- a/assets/sass/_form.scss
+++ b/assets/sass/_form.scss
@@ -1,0 +1,3 @@
+.contact-form--customer-quote {
+    padding: 35px 40px 0;
+}
--- a/assets/sass/_form.scss
+++ b/assets/sass/_form.scss
@@ -1,3 +1,3 @@
 .contact-form--customer-quote {
-    padding: 35px 40px 0;
+    @apply pt-8 px-10;
 }
--- a/assets/sass/_split-section.scss
+++ b/assets/sass/_split-section.scss
@@ -1,0 +1,15 @@
+.split-section {
+
+    &--container {
+        padding: 30px 0;
+    }
+
+    &--left {
+        padding-right: 35px;
+    }
+
+    &--right {
+        padding-left: 35px;
+    }
+
+}
--- a/assets/sass/_split-section.scss
+++ b/assets/sass/_split-section.scss
@@ -12,4 +12,9 @@
         padding-left: 35px;
     }
 
+    &--cta-btn {
+        margin-top: 35px;
+        @apply text-xl;
+    }
+
 }
--- a/assets/sass/_split-section.scss
+++ b/assets/sass/_split-section.scss
@@ -17,4 +17,27 @@
         @apply text-xl;
     }
 
+    @media (max-width: 768px) {
+        &--left {
+            padding-right: 0;
+        }
+    
+        &--right {
+            padding-left: 0;
+        }
+
+        &--cta-btn {
+            width: 100%;
+            text-align: center;
+        }
+    }
+
+    &--container {
+        @media (max-width: 530px) {
+            .language-js {
+                font-size: .5rem;
+            }
+        }
+    }
+
 }
--- a/assets/sass/_split-section.scss
+++ b/assets/sass/_split-section.scss
@@ -1,34 +1,32 @@
 .split-section {
 
     &--container {
-        padding: 30px 0;
+        @apply pt-8 px-0;
     }
 
     &--left {
-        padding-right: 35px;
+        @apply pr-8;
     }
 
     &--right {
-        padding-left: 35px;
+        @apply pl-8;
     }
 
     &--cta-btn {
-        margin-top: 35px;
-        @apply text-xl;
+        @apply mt-8 text-xl;
     }
 
     @media (max-width: 768px) {
         &--left {
-            padding-right: 0;
+            @apply pr-0;
         }
-    
+
         &--right {
-            padding-left: 0;
+            @apply pl-0;
         }
 
         &--cta-btn {
-            width: 100%;
-            text-align: center;
+            @apply w-full text-center;
         }
     }
 
--- a/content/topics/kubernetes.md
+++ b/content/topics/kubernetes.md
@@ -99,6 +99,7 @@ examples:
               .getResource("v1/Service", "wpdev-wordpress")
               .status.apply(status => status.loadBalancer.ingress[0].ip);
       cta:
+          id: get-started-kubernetes-yaml-helm
           url: /docs/get-started/kubernetes/
           label: TRY IT NOW
 
@@ -131,6 +132,7 @@ examples:
           });
 
       cta:
+          id: get-started-kubernetes-multi-cloud
           url: /docs/get-started/kubernetes/
           label: GET STARTED
 ---
--- a/content/topics/kubernetes.md
+++ b/content/topics/kubernetes.md
@@ -69,8 +69,8 @@ examples:
               .getResource("v1/Service", "wpdev-wordpress")
               .status.apply(status => status.loadBalancer.ingress[0].ip);
       cta:
-          url: /docs/get-started
-          label: GET STARTED
+          url: /docs/get-started/kubernetes/
+          label: NEXT STEP
 
     - title: Provision Kubernetes clusters in any cloud
       body: >
@@ -101,6 +101,6 @@ examples:
           });
 
       cta:
-          url: /docs/get-started
-          label: GET STARTED
+          url: /docs/get-started/kubernetes/
+          label: NEXT STEP
 ---
--- a/content/topics/kubernetes.md
+++ b/content/topics/kubernetes.md
@@ -39,6 +39,36 @@ sections:
     - id: contact
       label: Contact Us
 
+kubernetes_crosswalk:
+    code: |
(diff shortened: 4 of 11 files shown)


layouts/topics/kubernetes.html
```diff
@@ -1,9 +1,35 @@
...
+                            href="https://app.pulumi.com/signup?reason=pass_through&type=gitlab.com">
+                            <img class="h-10 my-auto mr-4" src="/logos/tech/gitlab.svg" alt="GitLab">
+                            <p class="hero--registration-button my-auto">GitLab</p>
+                        </a>
+                    </div>
+                    <div class="flex flex-row max-w-full">
+                        <a class="btn bg-gray-300 border border-gray-300 hover:bg-gray-200 mb-5 flex justify-center w-1/2 mr-4"
+                            href="https://app.pulumi.com/signup?reason=pass_through&type=bitbucket.com">
+                            <img class="h-10 my-auto mr-4" src="/logos/tech/atlassian.svg" alt="BitBucket">
+                            <p class="hero--registration-button my-auto">BitBucket</p>
```
praneetloke: The official Bitbucket brand name is spelled with a lower-case 'b' for 'bucket' unlike GitHub and GitLab.

```suggestion
                            <p class="hero--registration-button my-auto">Bitbucket</p>
```

layouts/topics/kubernetes.html
```diff
@@ -1,9 +1,35 @@
...
+                        </a>
+                        <a class="btn bg-gray-300 border border-gray-300 hover:bg-gray-200 mb-5 flex justify-center w-1/2 mr-4"
+                            href="https://app.pulumi.com/signup?reason=pass_through&type=gitlab.com">
+                            <img class="h-10 my-auto mr-4" src="/logos/tech/gitlab.svg" alt="GitLab">
+                            <p class="hero--registration-button my-auto">GitLab</p>
+                        </a>
+                    </div>
+                    <div class="flex flex-row max-w-full">
+                        <a class="btn bg-gray-300 border border-gray-300 hover:bg-gray-200 mb-5 flex justify-center w-1/2 mr-4"
+                            href="https://app.pulumi.com/signup?reason=pass_through&type=bitbucket.com">
```
praneetloke:
```suggestion
                            href="https://app.pulumi.com/signup?reason=pass_through&type=bitbucket.org">
```

zchase:
@chrsmith @praneetloke @cnunciato I added a tracking framework for tracking link clicks on the page and finalized the content. Please let me know if you see anything funky or have any suggested changes. Thanks!

@chrsmith Is there anyone else I need to run these changes?

assets/js/tracking.js
```diff
@@ -0,0 +1,42 @@
...
+ * Function for registering tracking to a link. To add tracking to a link you
+ * need to have an "id" attritube on the <a> element and the "track-click" attribute.
+ *
+ *   - ex. <a id="get-started-no-yaml" class="btn" href="/docs/get-started" track-click>GET STARTED</a>
+ */
+$(document).ready(function() {
+
+    // Check if the analytics object and track function are available. If they are
+    // not we do not even want to attempt to track anything.
+    if (window && window.analytics && typeof window.analytics.track === "function") {
```
chrsmith: Where does the `analytics` object come from? Segment? I wonder how this interacts with GDPR and what not. HubSpot pops up a cookies content form, is there a way we can read that setting?

It might be worth just filing a new issue in the `pulumi/home` repo to track following up on how the "cookie content" issue is managed, and how we can be consistent in honoring that option on our web properties.

assets/js/tracking.js
```diff
@@ -0,0 +1,42 @@
...
+
+        }
+
+        // Loop over the array of elements to register the click listeners.
+        for (var i = 0; i < buttons.length; i++) {
+            registerTracker(buttons[i]);
+        }
+    }
+
+});
```
chrsmith: Nit: Add newline at end of file. Also, deleting the unnecessary vertical whitespace on lines 41, 33, 17, and 8 would be appreciated :)

assets/js/tracking.js
```diff
@@ -0,0 +1,42 @@
...
+                // The destination url of the link.
+                destinationPath: $(element).attr("href"),
+                // The current path.
+                url: window.location.pathname,
+            }
+
+            // Register a listener to the link to send data to Segment
+            // when it has been clicked.
+            $(element).on("click", function(e) {
+                window.analytics.track("link-click", trackingData);
```
chrsmith: Dumb question: on the click event the browser is navigating away... so will this track event actually go through? I assume it "just works", or else Segment would recommend we use something more complicated. (e.g. blocking the event from propagating, adding a 1ms timer callback, and _then_ navigating away.)

assets/sass/_form.scss
```diff
@@ -0,0 +1,3 @@
+.contact-form--customer-quote {
+    padding: 35px 40px 0;
+}
```
chrsmith: Nit: newline at end of file. Also, would `.contact-form .customer-quote` or `form .customer-quote` be more appropriate selectors?

layouts/partials/create-account-cta.html
```diff
@@ -0,0 +1,26 @@
...
+            <img class="h-10 my-auto mr-4" src="/logos/tech/atlassian.svg" alt="BitBucket">
+            <p class="hero--registration-button my-auto">Bitbucket</p>
+        </a>
+        <a id="create-account-email" class="btn bg-gray-300 border border-gray-300 hover:bg-gray-200 mb-5 flex justify-center w-1/2 mr-4"
+            href="https://app.pulumi.com/signup/email" track-click>
+            <i class="h-10 my-auto mr-4 text-black text-4xl fa fa-envelope"></i>
+            <p class="hero--registration-button my-auto">Email</p>
+        </a>
+    </div>
+</div>
```
chrsmith: Newline at end of file here... and pretty much every file :D

assets/js/tracking.js
zchase: Created an issue here: https://github.com/pulumi/home/issues/608

Assigned it to both of us for tracking. Do we have a lawyer we can reach out to about this?

assets/js/tracking.js
zchase: Jquery is doing the work here for us. The `on` listener will run the code inside the block then do whatever the default functionality of the element click is. We could prevent the link from actually navigating anywhere by calling `e.preventDefault();` in the block (I'm not suggesting we do that, I just think that helps explain how jquery handles this).

cnunciato:
Some initial here --

* The CTA buttons are too large and too high-contrast; they command more attention than they should have here, and sort of drown out everything around them. I'll see if I can come up with a treatment that helps them to work a bit better in this context. (I'm assuming the aim is to call attention to them, but also to balance that with the content above and beside them.) 

* The buttons in the page body are also too big; they should adopt the same styling as elsewhere on the site. The defaults should take care of this.

* The staggered layout doesn't follow a grid, and I'm not sure it's winning us much here, so I'm not inclined to adopt it yet. Let's stick with the 50/50 split we have now, unless there's some reason we feel like that isn't working.

* Why remove the secondary navigation?

* The text inside the code samples is getting scaled down at mobile resolution. I don't think we need to make any changes to the behavior of these from what we have now.

assets/sass/_split-section.scss
```diff
@@ -0,0 +1,41 @@
+.split-section {
```
cnunciato: I suspect we can remove this and just stick with the 50/50 split we have now. I'm all for exploring alternate layouts, but I don't think this is something we _need_ to do in this PR.

content/topics/kubernetes.md
```diff
@@ -32,125 +32,44 @@ hero:
...
+kubernetes_crosswalk:
+    code: |
+        import * as gcp from "@pulumi/gcp";
+        import * as k8s from "@pulumi/kubernetes";
 
-          In this example we take a typical Deployment and Service pattern to create a ServiceDeployment
-          class, simplifying the implementation of the canonical Guestbook app.
-      code: |
-          import * as k8sjs from "./k8sjs";
+        // Create a GKE cluster.
```
cnunciato: We use four-space indenting pretty much everywhere. We should stick with that.

content/topics/kubernetes.md
```diff
@@ -32,125 +32,44 @@ hero:
...
 
-          let frontend = new k8sjs.ServiceDeployment("frontend", {
-              replicas: 3,
-              image: "gcr.io/google-samples/gb-frontend:v4",
-              ports: [ 80 ],
-              loadBalancer: true,
-          });
+        // Create a quota.
+        const quotaAppNamespace = new k8s.core.v1.ResourceQuota("apps", {
+            spec: {hard: {cpu: "200", memory: "1Gi", pods: "10"}},
```
cnunciato:
```suggestion
            spec: { hard: { cpu: "200", memory: "1Gi", pods: "10" } },
```

layouts/partials/how-pulumi-works-lite.html
```diff
@@ -0,0 +1,54 @@
+<section id="how-it-works" class="bg-purple-500 py-16 px-4 md:px-0">
```
cnunciato: Why remove the CTAs here? It'd also be preferable, since this is otherwise identical to the how-pulumi-works partial, to have that partial take an instruction to hide the elements we want to hide, rather than duplicate everything.

layouts/topics/kubernetes.html
```diff
@@ -1,9 +1,11 @@
 {{ define "hero" }}
     <header class="header-hero header-hero-glow">
         <div class="header-hero-items container mx-auto">
-            <div class="header-hero-item">
+            <div class="header-hero-item pr-20">
```
cnunciato: I don't think this pr-20 is being applied (because of the properties of header-hero-item).

layouts/topics/kubernetes.html
```diff
@@ -13,93 +15,192 @@ <h1>{{ .Params.hero.title }}</h1>
...
+        </div>
+    </section>
+
+    <section id="k8s-crosswalk" class="py-8 bg-white px-4 md:px-0">
+        <div class="container mx-auto flex flex-col md:mx-auto md:flex md:flex-row split-section--container">
+            <div class="w1/1 order-last mt-16 md:mt-0 md:w-3/5 md:order-none split-section--left">
+                {{ partial "code" (dict "code" .Params.kubernetes_crosswalk.code "lang" "js" "mode" "dark") }}
+            </div>
+            <div class="w1/1 order-first md:w-2/5 md:order-none split-section--right">
+                <h2>Crosswalk for Kubernetes - Playbooks for Infrastructure</h2>
```
cnunciato: Nit: `Crosswalk for Kubernetes: Playbooks for Infrastructure `

layouts/topics/kubernetes.html
```diff
@@ -13,93 +15,192 @@ <h1>{{ .Params.hero.title }}</h1>
...
                 <p>
-                    Pulumi uses real code which makes it easy, and efficient, for the
-                    development inner loop, at the same time as unlocking more complicated,
-                    real-world DevOps scenarios.
+                    Create, deploy, and manage production-ready infrastructure leveraging hosted Kubernetes offerings
```
cnunciato: Minor, but suggest `leveraging` -> `using` or `with`

ericrudder:
yes, "using" is better here :)

cnunciato:
@zchase Given the recent changes to this page, do you want to close this?

zchase:
Yup, no need for this anymore.

PR #2017, opened 2019-11-19
Restyle pricing form and create template for form with quote

zchase:
Add some styling to the Contact Us form on the pricing page. If we agree it looks good, I will go ahead and apply this style to the remaining forms.

Old Form:
oldForm

New Form:
formux

### DIFF
--- a/assets/config/tailwind.config.js
+++ b/assets/config/tailwind.config.js
@@ -86,10 +86,6 @@ module.exports = {
                     ...defaultTheme.fontFamily.sans,
                 ],
             },
-
-            boxShadow: {
-                "input-outline": "0 0 0 3px rgba(81, 38, 104, 0.5)"
-            },
         },
 
         colors: {
--- a/assets/config/tailwind.config.js
+++ b/assets/config/tailwind.config.js
@@ -86,6 +86,10 @@ module.exports = {
                     ...defaultTheme.fontFamily.sans,
                 ],
             },
+
+            boxShadow: {
+                "input-outline": "0 0 0 3px rgba(81, 38, 104, 0.5)"
+            },
         },
 
         colors: {
--- a/assets/sass/_hubspot.scss
+++ b/assets/sass/_hubspot.scss
@@ -32,6 +32,9 @@
 
         input, textarea {
             @apply border border-gray-300 rounded p-2 bg-white;
+            &:focus {
+                @apply outline-none shadow-outline;
+            }
         }
     }
 
@@ -86,6 +89,10 @@
             &.hs-button {
                 @apply block btn btn-lg btn border-none py-3 px-4 rounded-r rounded-l-none w-auto;
             }
+
+            &:focus {
+                @apply outline-none;
+            }
         }
     }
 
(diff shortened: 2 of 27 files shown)


assets/sass/_hubspot.scss
```diff
@@ -91,7 +91,16 @@
 
     &.pricing {
         label {
-            @apply font-normal;
+            @apply font-normal text-base;
+        }
+        input, textarea {
+            @apply shadow appearance-none border rounded w-full py-2 px-3 text-gray-700 leading-tight text-xl;
+            &:focus {
+                @apply outline-none shadow-outline;
```
cnunciato: I'm not sure I understand the intent with the borders and outlines. Is there some browser inconsistency you're solving for?

assets/sass/_hubspot.scss
```diff
@@ -91,7 +91,16 @@
 
     &.pricing {
         label {
-            @apply font-normal;
+            @apply font-normal text-base;
+        }
+        input, textarea {
+            @apply shadow appearance-none border rounded w-full py-2 px-3 text-gray-700 leading-tight text-xl;
```
cnunciato: Let's remove the `shadow` around the form fields and keep them on the same plane as the form itself.

layouts/partials/form-section.html
```diff
@@ -0,0 +1,35 @@
...
+                <h3 class="text-center">{{ .formHeader }}</h3>
+                <div class="mx-auto flex justify-center">
+                    <div class="hs-form hs-form-fg-dark font-normal">
+                        {{ partial "hubspot-form.html" (dict "hubspotFormID" .hubspotFormID "classNames" .formClassNames) }}
+                    </div>
+                </div>
+            </div>
+        </div>
+    </div>
+</section>
```
cnunciato: Newline. 😄 (If you're using VSCode, you can set this with `"files.insertFinalNewline": true`.)

assets/sass/_hubspot.scss
zchase: Yup I am just making sure the focus outline is consistent across browsers. In my changes I updated the outline color to be purple to fit more in line with out theme. I'm interested to to get your opinion on this.

assets/config/tailwind.config.js
```diff
@@ -86,6 +86,10 @@ module.exports = {
                     ...defaultTheme.fontFamily.sans,
                 ],
             },
+
+            boxShadow: {
+                "input-outline": "0 0 0 3px rgba(81, 38, 104, 0.5)"
```
cnunciato: I'm up for giving this a try, though I'd use a color from the maps defined above in this page, and probably one of the lighter-to-medium grays, for portability.

assets/sass/_hubspot.scss
```diff
@@ -93,5 +93,17 @@
...
             @apply font-normal;
         }
+        input, textarea {
+            @apply appearance-none border rounded w-full py-2 px-3 text-gray-700 leading-tight text-xl;
+            &:focus {
+                @apply outline-none shadow-input-outline;
+            }
+        }
+        .hs-button {
+            @apply text-white btn-orange;
```
cnunciato: Let's have this button stick with the blue and unaltered font size (as with the submit button below), which should be the defaults: 

image

PR #2019, opened 2019-11-20
Including the fastly docs

stack72:


stack72:
Pulumify failure isn't related to this build

PR #2025, opened 2019-11-20
Use the pulumify image from the pulumi org

justinvp:
And specify an explicit version tag.


justinvp:
Fixes Pulumify. Merging. cc @joeduffy, @lukehoban

PR #2024, opened 2019-11-20
Updated Query for Kubernetes blog

spara:

PR #2026, opened 2019-11-20
Title Change for Pulumi Query for Kubernetes

spara:


lukehoban:
Note that this feedback still isn't incorporated: https://github.com/pulumi/docs/pull/2024#discussion_r348747784.

PR #2032, opened 2019-11-22
Typo in headline

ustun:
I believe there is a typo in this headline.


stack72:
Thanks so much @ustun!!

PR #2034, opened 2019-11-22
Add link tracking script and apply it home page links

zchase:
This PR adds click tracking to links on our home page. The `tracking.js` script will look for all `<a>` elements with a `track-link` attribute and send link click data to Segment when those links are clicked. Right now tracking is dependent on the `track-click` attribute being set and the `id` on the element being set. When expand the tracking we will just capture all `<a>` elements, though we will have to determine a good fallback for when `id` isn't present on an element. Maybe use the DOM path instead?

### DIFF
--- a/assets/js/tracking.js
+++ b/assets/js/tracking.js
@@ -1,0 +1,54 @@
+/*
+ * Function for registering tracking to a link. To add tracking to a link you
+ * need to have an "id" attritube on the <a> element and the "track-click" attribute.
+ *
+ *   - ex. <a id="get-started-no-yaml" class="btn" href="/docs/get-started" track-click>GET STARTED</a>
+ */
+$(document).ready(function() {
+    // Check if the analytics object and track function are available. If they are
+    // not we do not even want to attempt to track anything.
+    if (window && window.analytics && typeof window.analytics.track === "function") {
+
+        // Find all the links with a "track-click" attribute.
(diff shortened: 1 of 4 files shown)


assets/js/tracking.js
```diff
@@ -0,0 +1,54 @@
...
+            });
+        }
+
+        // Loop over the array of elements to register the click listeners.
+        for (var i = 0; i < links.length; i++) {
+            registerTracker(links[i]);
+        }
+
+        // Remove the event listeners when we navigate to a new page.
+        $(window).on("unload", function() {
```
cnunciato: This seems fine, but since we're always in full-page-load-land, it probably isn't necessary. (Unless I'm misunderstanding.)

assets/js/tracking.js
```diff
@@ -0,0 +1,54 @@
...
+            // when it has been clicked.
+            $(element).on("click", function(e) {
+                // The value is the time in seconds from page load to user action.
+                trackingData.value = ((new Date().getTime()) - now) / 1000;
+                window.analytics.track("link-click", trackingData);
+            });
+        }
+
+        // Loop over the array of elements to register the click listeners.
+        for (var i = 0; i < links.length; i++) {
```
cnunciato: Since you already have `links` as a jQuery collection, could you not just `.each` over the items in the collection and call `registerTracker` directly with them?

PR #2046, opened 2019-11-26
Add FAQs for crossguard

ekrengel:
This adds an FAQ page for PaC primary motivation here was to answer:

1. the interaction btw local and service policy packs
1. how pac works with import/refresh

Closes https://github.com/pulumi/pulumi-policy/issues/70

Related to https://github.com/pulumi/pulumi-service/issues/3574 (but still hoping to add a test)


content/docs/guides/crossguard/faq.md
```diff
@@ -0,0 +1,44 @@
...
+linktitle: Frequently Asked Questions
+
+menu:
+  userguides:
+    parent: crossguard
+    identifier: crossguard-faq
+---
+{{% crossguard-preview %}}
+
+1. *Is CrossGuard open source?*
```
justinvp: It might be worth using headings fro each question, like we do here: https://www.pulumi.com/docs/troubleshooting/faq/

Then the list of questions will show up on the right-hand nav bar.

PR #2047, opened 2019-11-26
Add stack policy exmaple

ekrengel:
This adds a simple example of a stack policy validation. I did not include the helper function but can if we think it would useful. Ideally we get some sort of helper like this in the `pulumi-policy` repo.


ekrengel:
opted for adding the utility helper here instead of making people hop all over for how to do it!

PR #2048, opened 2019-11-26
Add API docs for @pulumi/policy

justinvp:
Change to add API docs for `@pulumi/policy`. I will regenerate this and mark the PR ready for review after we've published the new release of `@pulumi/policy`.

Pulumify Preview: http://content-bucket-a41cd30.s3-website-us-east-1.amazonaws.com/docs/reference/pkg/nodejs/pulumi/policy/


tools/tscdocgen/main.go
```diff
@@ -441,7 +441,11 @@ func (e *emitter) emitMarkdownModule(name string, mod *module, root bool) error
 		Name string
 		Link string
 	}
-	for modname := range mod.Modules {
+	for modname, module := range mod.Modules {
+		if len(module.Exports) == 0 && len(module.Modules) == 0 {
+			continue
+		}
```
justinvp: This avoids including an empty `tests` module in the generated docs.

tools/tscdocgen/main.go
```diff
@@ -1294,8 +1301,12 @@ func (e *emitter) createTypeLabel(t *typeDocType, indent int) string {
 					label += fmt.Sprintf("%s%s;\n",
 						strings.Repeat(" ", indent*4), e.createSignature(child.Signatures[0], decl, false))
 				} else {
+					var childType string
+					if child.Type != nil {
+						childType = e.createTypeLabel(child.Type, indent)
+					}
 					label += fmt.Sprintf("%s%s: %s;\n",
-						strings.Repeat(" ", indent*4), child.Name, e.createTypeLabel(child.Type, indent))
+						strings.Repeat(" ", indent*4), child.Name, childType)
```
justinvp: We're now hitting a new case that our doc generator hasn't run into before leading to dereferencing a nil pointer.

For now, I'm simply avoiding the nil reference, but it will require some work (which will have to be done subsequently), to generate the right signature. Right now it shows up as:

Screen Shot 2019-11-26 at 1 24 10 PM

```typescript
validateTypedResource<TResource,TArgs>(resourceClass: {
    constructor: ;
}, validate: (props: Unwrap<NonNullable<TArgs>>, args: ResourceValidationArgs, reportViolation: ReportViolation) => Promise<void> | void): ResourceValidation
```

When it should look more like:

```typescript
function validateTypedResource<TResource extends Resource, TArgs>(resourceClass: {
    new(name: string, args: TArgs, ...rest: any[]): TResource
}, validate: (props: Unwrap<NonNullable<TArgs>>, args: ResourceValidationArgs, reportViolation: ReportViolation) => Promise<void> | void): ResourceValidation
```

I'll file a bug to follow-up and fix this. What we have now isn't ideal but not the end of the world.

justinvp:
This is ready for review.

ekrengel:
Can we add the preview badge to this like how we have next to the cloud package?

PR #2055, opened 2019-11-27
Allow for restricting max-height on code blocks

### LINKED ISSUE #1969: Add ability to limit codeblocks to a maxHeight
### Problem description

To improve readability, I'd like to limit `codeblocks` to a given height / length to make them scrollable past a certain point.

### Suggestions for a fix

A markdown chooser like setting that could be applied to a whole file would be great. This way you can limit each file separately to a given max LoC.

cnunciato:
Fixes #1969. Example usage:

```
<div class="code-max-h">
    <div class="max-h-50"> <!-- as % of viewport height: we have -25, -50, -75 and -100) -->
        some beautiful code
    </div>
</div>
```

I'd hoped to be able to do this with a single div (say, `code-max-h max-h-25`), but the need to keep the copy buttons absolutely positioned required a containing div, alas.

### DIFF
--- a/assets/config/tailwind.config.js
+++ b/assets/config/tailwind.config.js
@@ -73,9 +73,7 @@ const green = {
 
 module.exports = {
     theme: {
-
         extend: {
-
             fontFamily: {
                 display: [
                     "Ubuntu",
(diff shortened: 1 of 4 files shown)


cnunciato:
@justinvp Yes, exactly right!

PR #2056, opened 2019-11-28
Add AWSGuard guide and various PaC-related tweaks

justinvp:
Adds a page that guides getting started using AWSGuard. Also updates the commands throughout to show how to set `PULUMI_EXPERIMENTAL` correctly on the various OSes (using the OS chooser tabs). And some minor other tweaks.

PR #2060, opened 2019-11-29
Prepare for 2.0 roadmap announcement

joeduffy:
This change prepares for the 2.0 roadmap announcement by adjusting some of the content, adding a new 2.0 blog entry, and changing the announcement alert to refer to it.


content/_index.md
```diff
@@ -19,7 +19,7 @@ carousel:
       description: modern applications.
       details:
         - title: Real programming languages.
-          description: Define infrastructure in JavaScript, TypeScript, Python, or Go.
+          description: Define infrastructure in JavaScript, TypeScript, Python, Go, or .NET Core.
```
justinvp: Is there a good reason to be using "Core" all over the place? The next version is dropping "Core". My personal preference would be to use ".NET" without "Core" by default in most places, and only mention "Core" on the installation page about the .NET Core 3.0 SDK requirement. This would be similar to what we do for Python, where we mention just "Python" in most places, not "Python 3" (vs. 2). @CyrusNajmabadi, @mikhailshilkov, @lukehoban, do you have a preference?

mikhailshilkov: I agree. The next version is going to be called ".NET 5". I think it was fine to use Core at the launch to build on the newly released .NET Core 3.0, but I'd start cleaning it up.

joeduffy: .NET marketing explicitly asked us to use Core in our blog post, and they seem to do the same, so I think until .NET 5 comes out, this is what we're supposed to be doing.

I agree we want to be consistent as .NET 5 becomes increasingly available, but that's a little ways off, isn't it?

joeduffy: That said, in this particular case, I think it's not quite right anyway. ".NET Core" isn't a language. Updated to "any .NET language, including C#, F#, and VB." LMK what you think.

content/docs/get-started/crossguard/_index.md
```diff
@@ -1,11 +1,13 @@
 ---
 title: Get Started with Policy as Code
-linktitle: Policy as Code
+linktitle: Policy as Code ("CrossGuard")
 menu:
   getstarted:
     identifier: pac
     weight: 3
+aliases: ["/docs/get-started/policy-as-code/"]
```
justinvp: The other moved pages need aliases as well.

justinvp:
@joeduffy, unfortunately, I just merged some [changes](https://github.com/pulumi/docs/pull/2056) that conflict with some of the changes here. Mainly, it introduces a new `content/docs/get-started/policy-as-code/awsguard.md` page that will have to be moved to the new directory (and links updated), and some page weights readjusted.

joeduffy:
Ok, @justinvp, do you mind pushing merge changes to my branch? If you can't, I'll try to merge in the morning before publishing the main announcement.

content/docs/get-started/crossguard/_index.md
justinvp: I added a commit that adds aliases to the other two moved pages.

justinvp:
> do you mind pushing merge changes to my branch

@joeduffy, No problem. I rebased against master, addressed conflicts, and force pushed. I also added aliases to the two other moved pages that needed them.

content/blog/pulumi-2-0-roadmap/index.md
```diff
@@ -0,0 +1,92 @@
+---
+title: "Pulumi 2.0 Roadmap"
```
mikhailshilkov: What I don't quite understand after reading this blog: why do those changed require 2.0? Are they breaking 1.0 and where?

content/blog/pulumi-2-0-roadmap/index.md
```diff
@@ -0,0 +1,92 @@
...
+authors: ["joe-duffy"]
+tags: ["Pulumi-News"]
+meta_desc: "Today we've published Pulumi's 2.0 roadmap, the next major phase in Pulumi's journey. This release will support great productivity, deeper support for enterprise workloads, and a whole lot more. Let us know what you think."
+date: "2019-11-29"
```
joeduffy:
```suggestion
date: "2019-12-02"
```

PR #2099, opened 2019-12-04
Add a blog post about EKS cluster features

joeduffy:


content/blog/aws-eks-managed-nodes-fargate/index.md
```diff
@@ -0,0 +1,310 @@
...
+
+## Using EKS Managed Node Groups
+
+The new EKS feature [Managed Node Groups](https://aws.amazon.com/blogs/containers/eks-managed-node-groups/) simplifies the task of managing explicit pools of worker nodes, at the cost of some amount of control. The `@pulumi/eks` package already had many of these conveniences built-in but this is now an official feature of the AWS platform.
+
+Managed Node Groups will automatically scale the EC2 instances powering your cluster using an Auto Scaling Group (ASG) managed by EKS. This ASG also runs the latest Amazon EKS-optimized Amazon Linux 2 AMI. This is great on one hand &mdash; because updates will be applied automatically for you &mdash; but if you want control over this you will want to manage your own node groups. Finally, security groups, IAM roles, and connecting them together is handled for you.
+
+To opt-in to using Managed Node Groups, you can use the raw [`aws.eks.NodeGroup`](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/aws/eks/#NodeGroup) class. However, if you're using the `eks.Cluster` abstraction, we've made it easy to use them, with the addition of a new `createManagedNodeGroup` function.
```
metral: Though you can create `aws.eks.NodeGroup` and link it to the cluster, the stack will error out because AWS will automatically create the `aws-auth` IAM configmap for us and we get a dupe error when trying to create the pulumi managed configmap.

This creates a race between AWS creating the configmap with only IAM data of the first nodegroup to join, and not any user provided role or user IAM mappings, or other managed nodegroup IAM data for that matter.

To get around this, in `createManagedNodeGroup` we create the configmap first, and then have the managed nodegroup depend on the object to ensure we always create the aws-auth as the user expects, and not AWS.

I'd simplify this statement to only focusing on `createManagedNodeGroup` and not the raw `aws.eks.NodeGroup` class.

content/blog/aws-eks-managed-nodes-fargate/index.md
```diff
@@ -0,0 +1,310 @@
...
+    tags: { "org": "pulumi" },
+}, cluster);
+
+// Export the cluster's kubeconfig.
+export const kubeconfig = cluster.kubeconfig;
+```
+
+This can be provisioned with a single `pulumi up`, just like before, but instead of the automatically created default node group, our cluster will use a Managed Node Group instead.
+
+> This looks more complex than the basic `eks.Cluster` example above simply because that exaple benefitted greatly from the EKS package's internal smarts. This approach gives you more control over the worker nodes, while still leaning on EKS to handle the EC2 node management based on your declarative specifications.
```
metral: I'd suggest we use the simplified example here to reduce verbosity:

https://github.com/pulumi/pulumi-eks/blob/e9d4c42036c56c7fd526ab6c7ddd85ff10d0fb17/nodejs/eks/examples/managed-nodegroups/index.ts#L19-L24

joeduffy: I considered it ... However, that particular example doesn't show any of the control that might drive you to want to use a managed node group, though, does it? (If there's a benefit to it, it's too subtle for me!)

content/blog/aws-eks-managed-nodes-fargate/index.md
joeduffy: Ah, good to know.

So, the tension here is, we really should be telling people that the raw building blocks are available. Otherwise, we look like we're getting in the way. We want everybody to know that every low-level building block feature of AWS is exposed by our SDK.

And then we layer on top simplifications. Like this one.

How does this sound: I will reword it to make it clear that if you're using the low level building blocks, you'd use `aws.eks.NodeGroup` directly -- but, for the reason you cite, you should use `createManagedNodeGroup` if you're using our higher level EKS package?

content/blog/aws-eks-managed-nodes-fargate/index.md
joeduffy: Actually, can you help me with one question? What is the essential difference from an end user's perspective between that example you link to, and this one, for purposes of the blog post summary? https://github.com/pulumi/pulumi-eks/blob/e9d4c42036c56c7fd526ab6c7ddd85ff10d0fb17/nodejs/eks/examples/nodegroup/index.ts#L33-L40

joeduffy: I wanted to show unmanaged node groups => complex, managed node groups => so much simpler, but the two look essentially equivalent to my eyes...

metral: We bake in sane defaults, but still allow complete control possible.

At the very least i think we should not include the note about this being more complex than usual

joeduffy: The intended flow of the post was:

1. Complex (EC2)
2. Easier (Managed Node Groups)
3. Easiest (Fargate)

Do you have a good example of (1) that drives the point home? Every example I can find for going from (1) to (2) actually makes them look more complex.

It could be that we show:

0. eks.Cluster with defaults -- pretty darn easy already
1. Complex (EC2)
2. Easier (Managed Node Groups)
3. Easiest (Fargate)

joeduffy: Ok, I've given this a shot. Basically, lead in with the simple `eks.Cluster` example. _Then_ go into a section about manual EC2 node groups, with a more complex example. Then the managed node group example is simplified similar to the example you pointed out. Feels like it flows better -- lmk what you think!

content/blog/aws-eks-managed-nodes-fargate/index.md
metral: > How does this sound: I will reword it to make it clear that if you're using the low level building blocks, you'd use aws.eks.NodeGroup directly -- but, for the reason you cite, you should use createManagedNodeGroup if you're using our higher level EKS package

SGTM 👍

content/blog/aws-eks-managed-nodes-fargate/index.md
```diff
@@ -0,0 +1,356 @@
...
+    instanceProfile: instanceProfile,
+});
+
+// Export the cluster's kubeconfig.
+export const kubeconfig = cluster.kubeconfig;
+```
+
+There is even more you can do here, including digging deep into the individual EC2 instances and Auto Scaling Groups powering the node group. Needing to manage workers at this level, and connect them to the right IAM and networking services, can be tedious and difficult to get right. That's where the new AWS features come into play.
+
+## Automatically Managed Node Groups
```
metral:
```suggestion
## AWS Managed Node Groups
```

content/blog/aws-eks-managed-nodes-fargate/index.md
metral: @joeduffy The updates flow way better 👍

content/blog/aws-eks-managed-nodes-fargate/index.md
```diff
@@ -0,0 +1,356 @@
...
+let counter = 0;
+for (const policyArn of [
+    "arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy",
+    "arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy",
+    "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly",
+]) {
+    new aws.iam.RolePolicyAttachment(`my-cluster-ng-role-policy-${counter++}`,
+        { policyArn, role },
+    );
+}
```
lukehoban: All this instanceprofile stuff isn’t needed is it?  @metral

content/blog/aws-eks-managed-nodes-fargate/index.md
```diff
@@ -0,0 +1,356 @@
...
+    skipDefaultNodeGroup: true,
+    vpcId: vpc.vpcId,
+    publicSubnetIds: vpc.publicSubnetIds,
+    privateSubnetIds: vpc.privateSubnetIds,
+    deployDashboard: false,
+    instanceRoles: [ role ],
+});
+
+// Create a simple AWS managed node group using a cluster as input.
+const managedNodeGroup = eks.createManagedNodeGroup("my-cluster-ng", {
```
joeduffy: @metral @lukehoban Do you know why `createManagedNodeGroup` is so stylistically different than `createNodeGroup`? Note that I need to pass the `cluster` as an argument (whereas with `createNodeGroup`, it's a member function on the cluster itself) and, as a result, I need to pass it twice (to get the `parent` wired up).

content/blog/aws-eks-managed-nodes-fargate/index.md
metral: This is for the original, self-managed node groups, not the AWS managed ones.

content/blog/aws-eks-managed-nodes-fargate/index.md
metral: > Do you know why createManagedNodeGroup is so stylistically different than createNodeGroup?

We tried to set some defaults where possible, while also inferring settings so the user doesn't have to be so explicit. But optional enough that if they want to provide full configuration they still can. Here is the defaulting we do:

https://github.com/pulumi/pulumi-eks/blob/e9d4c42036c56c7fd526ab6c7ddd85ff10d0fb17/nodejs/eks/nodegroup.ts#L638-L672

Do note that we aren't radically changing the spec's here, just simplifying it for folks where it makes sense. `aws.eks.NodeGroup` has a spec that closely resembles our `eks.NodeGroup` class, so there isn't much more reduction we can do that we haven't already incorporated.

> Note that I need to pass the cluster as an argument (whereas with createNodeGroup, it's a member function on the cluster itself) and, as a result, I need to pass it twice (to get the parent wired up).

This is a simple fix we can consolidate so we can auto-set the parent w/o having to pass it in. I opened an issue to track this. If you have any other ideas to correct the API style here, please drop a note in the issue:
https://github.com/pulumi/pulumi-eks/issues/285

PR #2112, opened 2019-12-06
Adjust tags styling for consistency

cnunciato:
Just a little thing I've noticed that I figured I'd fix up. (In the sidebar, they're lowercased, so this does the same, and reduces the text size slightly as well.)


cnunciato:
Yeah, I agree -- there was a bit of difference with the font size here, but everything else should be the same. I'll do that.

PR #2114, opened 2019-12-07
Use latest version of TypeDoc

### LINKED ISSUE #1654: Upgrade TypeDoc and TypeScript
We should upgrade the version of TypeDoc and TypeScript used to generate the Node.js docs, but we need to also update the `tsdocgen` tool to handle hyperlinking properly due to some changes to the JSON output from TypeDoc. Basically avoiding #1650.

### LINKED ISSUE #2103: tsdocgen: unhandled type - predicate
### Problem description

- Other: The `tsdocgen` utility doesn't properly handle the `predicate` type. I added a no-op case to allow doc generation to continue, but this should be implemented properly.

Related: #2102 

Here are the warning messages that are currently generated for docs on kubernetesx:
```
2019/12/05 09:55:55 ignoring predicate casepredicate
2019/12/05 09:55:55 ignoring predicate casepredicate
2019/12/05 09:55:55 ignoring predicate casepredicate
2019/12/05 09:55:55 ignoring predicate casepredicate
2019/12/05 09:55:55 ignoring predicate casepredicate
2019/12/05 09:55:55 ignoring predicate casepredicate
2019/12/05 09:55:55 ignoring predicate casepredicate
2019/12/05 09:55:55 ignoring predicate casepredicate
```

Here was the previous error message:
```
2019/12/05 09:37:35 unrecognized type node type: predicate
exit status 1
```

justinvp:
**TL;DR**: This change upgrades TypeDoc to the latest version which uses TypeScript `3.7.3`, so we can now use it to generate docs for libraries that are using more recent versions of TypeScript. Due to some changes to TypeDoc's emitted JSON output, we now get better linking of types within packages (more linking and more accurate), imported type names no longer have the `[pkg.]mod` prefix in front of `[pkg.]mod.Type` in most cases (cleaner looking docs), and I added support for predicate types, so type guards now show a return value of `obj is Type` rather than `boolean`.

---

This change upgrades to the latest version of TypeDoc, which uses TypeScript `3.7.3`, which has some changes to the JSON that it outputs that we need to handle:

## First commit:

- TypeDoc now specifies an `id` for types where it didn't before. This mostly happens with the `types/input` and `types/output` types. So when there is an `id` for the type, instead of emitting a hyperlink to that type as an anchor, actually look through all the modules to find the type, and generate a hyperlink to the type in the right module. This has a side benefit that we're now more correct with these links. Previously, it was possible we'd generate an anchor link for a type that didn't actually exist on the page. Now we get the link correct. We also now get much more linking within packages.

- TypeDoc is no longer includes the `pkg.module` prefix in type names (in most cases; I found there were still times that it did include these; and I'm not sure why). This actually results in cleaner looking API docs, because you no longer see `types.input.Foo` but rather just `Foo` the vast majority of the time (the hyperlinking still works as expected). However, there are some cases, such as in `awsx` where we were using the `aws.<mod>.Type` to link to types in the `aws` package. Since we no longer have the prefix in the type name, we have to look at just the type name itself to generate these.

- I went ahead and added similar lookups for `eks`.

## Second commit:

- TypeDoc now has a new "predicate" type, which are typically the return signature of type guards. Previously, these were typed as `boolean`, but with this upgrade, we now emit the type guard. This mostly shows up in our `isInstance` functions, e.g. `isInstance(obj: any): obj is Bucket`

## Third commit:

- Regenerate all the Node.js API docs.

Fixes #1654
Fixes #2103


content/docs/reference/pkg/nodejs/pulumi/pulumi/_index.md
```diff
@@ -813,7 +813,7 @@ fully qualified names, such as `pulumi:foo:a`, ..., and `pulumi:bar:a`, respecti
 </h4>
 
 
-<pre class="highlight"><code><span class='kd'>public </span>get&lt;K&gt;(key: <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String'>string</a></span>, opts?: <a href='#StringConfigOptions'>StringConfigOptions</a>&lt;K&gt;): K | <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined'>undefined</a></span></code></pre>
+<pre class="highlight"><code><span class='kd'>public </span>get&lt;K&gt;(key: <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String'>string</a></span>, opts?: StringConfigOptions&lt;K&gt;): K | <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined'>undefined</a></span></code></pre>
```
justinvp: `StringConfigOptions` and `NumberConfigOptions` aren't exported (which is really a bug in `@pulumi/pulumi`), so they don't show up in our docs. Previously, we were generating an anchor link, but those links didn't work because the anchor doesn't exist on the page.

justinvp: https://github.com/pulumi/pulumi/pull/3615

justinvp:
Going to merge this. Happy to address any feedback subsequently.

PR #2128, opened 2019-12-09
WIP: Add link tracking across majority of website

zchase:
This PR will add link click tracking across the manually defined links. This PR also includes a link linter `link_linter.js` for ensuring our links are properly tracked.

I made more progress and adding the proper configs to each link. But we are going to need to somehow track the links autogenerated by: https://github.com/pulumi/docs/tree/master/tools/mktutorial. I haven't taken a look yet but we will need to figure out some sort of strategy for assigning those link unique IDs and tracking their clicks.


layouts/index.html
```diff
@@ -108,33 +108,33 @@ <h1 class="font-sans font-body font-normal">
     <section class="mb-16 px-4 md:px-0">
         <div class="container mx-auto">
             <div class="lg:flex items-center justify-center my-4">
-                <a id="tableau-logo" href="https://www.tableau.com/" target="_blank" track-click>
+                <a id="logo-home-tableau" href="https://www.tableau.com/" target="_blank" track-click>
```
cnunciato: What's the benefit of tracking things like these?

layouts/shortcodes/changelog-table-row.html
```diff
@@ -13,13 +13,13 @@
 {{ end }}
 
 <tr>
-    <td><a href="https://github.com/pulumi/pulumi/blob/master/CHANGELOG.md#{{ $flatVersion }}-{{ $date }}">{{ $version }}</a></td>
+    <td><a id="cta-changelog-pulumi-sdk-version" href="https://github.com/pulumi/pulumi/blob/master/CHANGELOG.md#{{ $flatVersion }}-{{ $date }}" track-click>{{ $version }}</a></td>
     <td>{{ $date }}</td>
     <td>
-        <a href="https://get.pulumi.com/releases/sdk/pulumi-v{{ $version }}-linux-x64.tar.gz">Linux</a>
+        <a id="cta-changelog-pulumi-sdk-linux" href="https://get.pulumi.com/releases/sdk/pulumi-v{{ $version }}-linux-x64.tar.gz" track-click>Linux</a>
```
cnunciato: So, I think we can definitely be in a place where we don't need _both_ a handcrafted `id` _and_ a `track-link` attribute. I think if we can define some sort of a schema that makes sense (and maybe this is better handled a `data-*` kind of thing), we can do one or the other, and have our script understand how to interpret that.

layouts/index.html
zchase: Just in case really. If we start seeing high exit rates on certain it's good to have all links tracked so we can narrow down what is going. It is also good to know how many people are actually using these links in general. My personal assumption is that not a ton of people click these links, but that is all it is an assumption. I have no data to prove or disprove my assumption. I think it is best in this case to track all links so we have the data on everything, that way we can always make data driven decisions.

Interested to hear if you have a differing opinion however. I can see how this tracking can be view as overkill.

layouts/shortcodes/changelog-table-row.html
zchase: Yup that is a much cleaner idea. Something like `data-tracking-id` would make sense to me. I've tried to maintain a schema of `${type}-${page}-${section}-${uniqueIdentifier}` with `uniqueIdentifier` being something to differentiate the element. Such as the button text or what the link does.

These short codes are a bit different because they're shared elements. So it is probably best to set the page value of the "id" in the javascript.

I am having a hard time trying to figure out a way where we don't have add some sort of attribute to all of our links. Maybe one of use will have an epiphany over the weekend :)

zchase:
Closing this PR as we are going to change the `id` and `track-click` attributes to just be `data-tracking-id`. It is easier to just throw away this PR and start fresh on a new one.

PR #2129, opened 2019-12-10
Add C# to Kubernetes getting started

lukehoban:


content/docs/get-started/kubernetes/modify-program.md
```diff
@@ -135,6 +135,109 @@ else:
...
+                    {
+                        MatchLabels = appLabels,
+                    },
+                    Replicas = 1,
+                    Template = new PodTemplateSpecArgs
+                    {
+                        Metadata = new ObjectMetaArgs
+                        {
+                            Labels = appLabels,
+                        },
```
mikhailshilkov: One-liners for single-property definition should fit the string and make it shorter in height

lukehoban: Going to leave as-is for now - as horizontal space is pretty tight already for these examples - and the current format is consistent with many other examples we have (and with other parts of the same example).  Though open to adopting this pattern for single property definitions broadly in follow-up if that's what you think is best to standardize on.

PR #2135, opened 2019-12-11
Add markdown linting

zchase:
This PR adds linting for our markdown files. The reasoning behind adding linting is to not only catch silly mistakes in our markdown, but to check that we are not making any SEO mistakes. The linter will catch duplicate H1 tags, missing alt text on an image, missing page titles, and long meta descriptions.

Remaining:
- [x] Fix remaining lint errors.
- [x] Add support for checking page title being defined.
- [x] Add support for checking meta description length.

Markdown Lint Rules: https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md


scripts/markdown_seo_lint.js
```diff
@@ -0,0 +1,219 @@
+const fs = require("fs");
+const yaml = require("js-yaml");
+const markdownlint = require("markdownlint");
+const path = require("path");
+
+const FRONT_MATTER_REGEX = /((^---\s*$[^]*?^---\s*$)|(^\+\+\+\s*$[^]*?^(\+\+\+|\.\.\.)\s*$))(\r\n|\r|\n|$)/m;
```
chrsmith: YES.

scripts/markdown_seo_lint.js
```diff
@@ -0,0 +1,219 @@
...
+
+function checkPageTitle(title) {
```
chrsmith: Since we aren't using a strongly typed language, we need to be super clear in comments what the expected semantics of each function is. So I'd strongly urge you to put a doc comment on top of every function explaining what it does and what its return type is.

e.g. does `checkPageTitle(title)` take an object? A string? Does it throw an error if it encounters a problem? Or just return a user-friendly error message or `null`?

scripts/markdown_seo_lint.js
```diff
@@ -0,0 +1,219 @@
...
+    let error = null;
+    if (!title) {
+        error = "Missing page title";
+    }else if (title.trim().length === 0) {
```
chrsmith: Nit: Rather than call `title.trim().length` twice, we can just rewrite this as the following. Also, IMHO it seems a little cleaner to just return the string literal rather than assign it to `error` and return that.

```
let titleLength = title.trim().length;
if (titleLength === 0) {
    return "Page title is empty";
}
if (titleLength > 60) {
    return "Page title exceeds 60 characters";
}
return null;
```

scripts/markdown_seo_lint.js
```diff
@@ -0,0 +1,219 @@
...
+        error = "Page title exceeds 60 characters";
+    }
+    return error;
+}
+
+function checkPageMetaDescription(meta) {
+    let error = null;
+    if (!meta) {
+        error = "Missing meta description";
+    }else if (meta.trim().length === 0) {
```
chrsmith: Assigning `.trim().length` to a local variable, and just returning the string directly seems like it would clean up this function too.

scripts/markdown_seo_lint.js
```diff
@@ -0,0 +1,219 @@
...
+    }else if (meta.trim().length < 50) {
+        error = "Meta description is too short. Must be at least 50 characters";
+    }else if (meta.trim().length > 160) {
+        error = "Meta descripiton is too long. Must be shorter than 160 characters";
+    }
+    return error;
+}
+
+// Recursively generate a list of markdown files from a given
+// list of paths.
```
chrsmith: To be more precise, is "paths" expected to be a string array?

Also, what does this function return? What is `result` supposed to be? It seems like `options` would be a clearer name for the variable. What do you think?

scripts/markdown_seo_lint.js
```diff
@@ -0,0 +1,219 @@
...
+    const fileSuffix = fileParts[fileParts.length - 1];
+
+    // Ignore auto generated docs.
+    if (file.indexOf("/content/docs/reference/pkg") > -1) {
+        const remaining = paths.slice(1, paths.length);
+        return searchForMarkdown(remaining, result);
+
+    // If the path is a directory we want to add the contents of the directory
+    // to the list.
+    }else if (isDirectory) {
```
chrsmith: Nit: Since the previous code block (`if (file.indexOf(....)`) returns, we can omit the `else` and make this a bit more readable.

```
// ...
// ...
if (isDirectory) {
     ...
}

File: scripts/markdown_seo_lint.js [hunk 7]
```diff
@@ -0,0 +1,219 @@
...
+        });
+        paths[0] = contents;
+
+        // Flatten the array.
+        const newPaths = [].concat.apply([], paths);
+        return searchForMarkdown(newPaths, result);
+
+    // Else check if the file suffix is a markdown
+    // and add it the resulting file list.
+    }else {
```
chrsmith: And remove the `} else {` since it is unnecessary. (As the previous code block returns.)

scripts/markdown_seo_lint.js
```diff
@@ -0,0 +1,219 @@
...
+        if (remaining.length > 0) {
+            return searchForMarkdown(remaining, result);
+        }
+    }
+
+    return result;
+}
+
+// Get a list of markdown files.
+function getMarkdownFileList(parentPath) {
```
chrsmith: Would `getMarkdownFiles` be a more natural name for the function?

scripts/markdown_seo_lint.js
```diff
@@ -0,0 +1,219 @@
...
+    const fullParentPath = path.resolve(__dirname, parentPath)
+    const dirs = fs.readdirSync(fullParentPath).map(function(dir) {
+        return parentPath + "/" + dir;
```
chrsmith: There are all sorts of problems (security and logical) when joining paths, so typically you should always call `path.Join(parentDir, dir)` instead of just concatenating things yourself.

scripts/markdown_seo_lint.js
```diff
@@ -0,0 +1,219 @@
...
+const errors = Object.keys(result).map(function(key) {
+    const lintErrors = result[key];
+    const frontMatterErrors = filesToLint.frontMatter[key];
+
+    if (frontMatterErrors.error) {
+        lintErrors.push({
+            lineNumber: "File Header",
+            ruleDescription: frontMatterErrors.error,
+        });
+    }else {
```
chrsmith: Nit: Space after `}` so `} else {`.

scripts/markdown_seo_lint.js
```diff
@@ -0,0 +1,219 @@
...
+            });
+        }
+    }
+
+    if (lintErrors.length > 0) {
+        return { path: key, errors: lintErrors };
+    }else {
+        return null;
+    }
+}).filter(function(item) { return item !== null; });
```
chrsmith: As much as I love functional programming, I'd suggest you break this up into multiple expressions for readability. As-is it's a bit too dense.
```
const errorsCount = [].concat.apply([], Object.values(result)).length;
const errors = Object.keys(result).map(function(key) {
    ...
}).filter(function(item) { return item !== null; });
```

Can we split this up into multiple local variables? Or use a function defined somewhere else instead of an anonymous function/lambda?

Ultimately if the "main" part of the script were shorter, things would be a lot easier to follow. e.g.:

```
let errors = groupLintErrorOutput(...);
```

scripts/markdown_seo_lint.js
```diff
@@ -0,0 +1,219 @@
...
+                const error = err.errors[i];
+                msg += "Line " + error.lineNumber + ": " + error.ruleDescription;
+                msg += error.errorDetail ? " [" + error.errorDetail + "].\n" : ".\n";
+            }
+            return msg;
+        }).join("\n")
+    }
+    `);
+    process.exit(1);
+}else {
```
chrsmith: Nit: No need for `} else {` here, as the if block above terminates.

scripts/markdown_seo_lint.js
zchase: Sounds good. Though to write it like this we will first need to check the title is not `null` or `undefined` otherwise the script will throw an error.

```
    if (!title) {
        return "Missing page title";
    }else if (typeof title === "string") {
        const titleLength = title.trim().length;
        if (titleLength === 0) {
            return "Page title is empty";
        }else if (titleLength > 60) {
            return "Page title exceeds 60 characters";
        }
    }
    return null;
```

PR #2136, opened 2019-12-11
Document azure:location prompt duing create project

bjorncoltof:
### Proposed changes

During the Create Project step for Azure you will be prompted for the location, this was missing from the docs


lukehoban:
Sorry for the delay responding on this - the good news is this was addressed some time ago!

PR #2159, opened 2019-12-14
Attempt to clarify open source vs. freemium vs. paid

joeduffy:
This change:

- Add the Community Edition to the top-level edition tabs on
  the pricing page, replacing the separate section we already had.

- Adjust the feature differentiation waterfall in the edition tabs.

- Add a section to the top of the pricing page that clarifies that
  Pulumi is available as open source, freemium, or paid editions.

- Update the FAQ to call out the ability to use the open source
  independent of the service.

- Updates some text so that it doesn't sound like we're limiting
  the Community Edition to "individuals" -- in reality, it really
  just means "single user scenarios", which is subtly different.

This is certainly not perfect, but is a set of incremental changes
that I hope will clarify things at least a little bit.

The only thing I don't love is that it pushes the pricing tabs
closer to the fold, but I've adjusted some margins and kept the
new text fairly short, in an attempt to mitigate this.

### DIFF
--- a/layouts/page/pricing.html
+++ b/layouts/page/pricing.html
@@ -436,7 +436,7 @@
             <div class="pricing-item">
                 <span class="pricing-item-label"><a href="{{ relref . "/docs/intro/console/collaboration/organization-roles" }}" class="link">Role Based Access Control</a></span>
                 <span class="pricing-item-value"><i class="fas fa-times text-grey"></i></span>
-                <span class="pricing-item-value"><i class="fas fa-times text-grey"></i></span>
+                <span class="pricing-item-value"><i class="fas fa-check-square text-green-500"></i></span>
                 <span class="pricing-item-value"><i class="fas fa-check-square text-green-500"></i></span>
             </div>
             <div class="pricing-item">
--- a/layouts/page/pricing.html
+++ b/layouts/page/pricing.html
@@ -1,29 +1,34 @@
 {{ define "hero" }}
     <header class="header-hero">
-        <div class="container mx-auto text-center flex-col max-w-2xl">
+        <div class="container mx-auto text-center flex-col max-w-5xl">
             <h1>Pulumi for every team and individual.</h1>
             <div class="container mx-auto lg:flex justify-center mt-8">
-                <div class="px-8 lg:w-1/2 text-center flex flex-col text-gray-300">
-                    <div class="text-lg uppercase rounded text-blue-300 font-bold">Open source</div>
+                <div class="px-8 lg:w-1/3 text-center flex flex-col text-gray-300">
+                    <div class="text-lg uppercase rounded text-blue-300 font-bold">Open source.</div>
                     <p>
-                        Pulumi is an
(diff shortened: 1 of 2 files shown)


layouts/page/pricing.html
```diff
@@ -1,55 +1,114 @@
...
+                </div>
+            </div>
         </div>
     </header>
 {{ end }}
 
 {{ define "main" }}
-    <section class="bg-purple-800 pt-24 text-white px-4 lg:px-0">
+    <section class="bg-purple-800 pt-24 text-white px-4 lg:px-0 justify-center text-center">
+        <h2 class="text-gray-300 mb-8">Pulumi Editions</h2>
```
cnunciato: You could win back a little vertical space by dropping this, since it's not really necessary.

joeduffy: Sounds great. Done!

mikhailshilkov:
This is a great change! Here are some ideas to bring it even further.

My initial screen at the laptop looks like this:
image

There's no pricing table yet, I only see two blocks of text. There seems to be enough space to add more info here.

The transition to the pricing tiers isn't fully. If it's open-source, why is it not free for teams?

I think "you can run Pulumi offline and manage state yourself" might not be clear enough.

I would consider making three statements:

- Pulumi CLI is free
- Pulumi backend is not free but useful for teams
- Pulumi backend is free for individuals

The first FAQ answer is great!

That's my impression, but overall I'm super happy with the direction of the change.

joeduffy:
Great feedback, @mikhailshilkov!

@cnunciato I agree with a lot of the feedback, but honestly struggled to get the layout close to what I had originally envisioned, especially as I wanted to be consistent with our overall hero styling (which, frankly, probably needs to be revisited overall given how much wasted sea of purple space it has). Completely random example of another website, https://unearthlabs.com/pricing/, but notice how much information is instantly visible above the fold.

I'll keep pondering -- I am thinking of maybe merging it close to as-is, since it's a good step forward, and then perhaps we can file an issue for a deeper overhaul during M31?

cnunciato:
I’m +1 to merge even as-is because it addresses the important bit by clarifying the existence of a first-class free plan. I’m happy to work on incorporating this feedback in a follow-up pass.

joeduffy:
Here's an alternative I was just toying with. Better? Worse?

image

joeduffy:
One thing I dislike is the "run it offline" link may distract from the CTAs on the page.

joeduffy:
@cnunciato I propose we keep this change simple for now, esp. given the potentially significant impact to user behavior on this page. This effectively now comes down to not burying Community Edition at the bottom of the page. Notably, this doesn't really clarify the relationship between Pulumi's SDK open source and the fact you can use the service (or not), which is still a problem.

image

As a bonus, it actually saves a dozen or so vertical pixels compared to our current page :smile:

cnunciato:
Love it. I was actually going to suggest we minimize the changes to the hero section just to keep from switching up too many things at once. This is great. :shipit:

PR #2184, opened 2019-12-20
fix  #1764

### LINKED ISSUE #1764: Getting started with Python on Windows - wrong activate command
### Problem description

Getting Started guides, while running Python on Windows, give an incorrect hint about how to activate the virtual environment ([example](https://www.pulumi.com/docs/get-started/azure/review-project/)):

```
source venv/bin/activate
```

In addition, nowhere in the guide is it said to install `virtualenv`.

### Suggestions for a fix

CLI gives the correct hint:

```
2. venv\Scripts\activate.bat
```

So the guide should give it too.

Also suggest installing `virtualenv` with `pip3 install virtualenv`.

spara:
Updated python sections for Getting Started

Fixes #1764.


content/docs/get-started/aws/review-project.md
```diff
@@ -82,18 +82,33 @@ class Program
...
+
 ```bash
-$ source venv/bin/activate
+$ source aws-env/bin/activate
 ```
 
+Activate the environment in Windows:
+
+```bash
+> aws-env\Scripts\activate.bat
```
justinvp: It'd be best to not include the `>` in front of `bat` snippets right now, because currently when our copy button for code snippets is clicked, it won't strip the `>` from the text that is added to the clipboard like we do for `$` in `bash`/`shell`/`sh` snippets.

https://github.com/pulumi/docs/blob/a3a71548eee00c5fddaee66e9154aee5b364e24b/assets/js/copybutton.js#L17-L22

content/docs/get-started/aws/review-project.md
```diff
@@ -82,18 +82,33 @@ class Program
...
+
+> *Note*: As a prerequisite, [install virtualenv](https://virtualenv.pypa.io/en/latest/installation/) to manage project requirements
+
 For Python, before we deploy the stack, the following commands need to be run to create a virtual environment, activate it, and install dependencies:
 
+Create a virtual environment:
+
 ```bash
-$ virtualenv -p python3 venv
+$ python -m venv aws-env
```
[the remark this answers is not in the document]
justinvp: Ah, I see in https://github.com/pulumi/pulumi/issues/3398#issuecomment-546138407 that `python3.exe` doesn't appear to be a thing on Windows.

content/docs/get-started/aws/review-project.md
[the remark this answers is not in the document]
spara: accepted

content/docs/get-started/aws/review-project.md
```diff
@@ -82,18 +82,33 @@ class Program
...
+
+```bash
+> aws-env\Scripts\activate.bat
+```
+
+Install dependencies:
+
 ```bash
-$ pip3 install -r requirements.txt
+$ pip install -r requirements.txt
```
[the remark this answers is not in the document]
spara: Agree it's better to leave in pip3

content/docs/get-started/aws/review-project.md
```diff
@@ -82,18 +82,33 @@ class Program
...
 For Python, before we deploy the stack, the following commands need to be run to create a virtual environment, activate it, and install dependencies:
 
+Create a virtual environment:
+
 ```bash
-$ virtualenv -p python3 venv
+$ python -m venv aws-env
 ```
 
+Activate the environment in Linux and MacOS:
```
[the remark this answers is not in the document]
spara: I don't think we use the os chooser in the other languages. We should rethink the docs organization before making this change.

content/docs/get-started/aws/review-project.md
spara: accepted

content/docs/get-started/aws/review-project.md
```diff
@@ -82,16 +82,31 @@ class Program
...
+
 ```bash
-$ source venv/bin/activate
+$ source aws-env/bin/activate
 ```
 
+Activate the environment in Windows:
+
+```bat
+ aws-env\Scripts\activate
```
lukehoban: Should we be consistent in whether we include the shell prompt?  It’s weird to have it for Linux but not for Windows.

content/docs/get-started/azure/review-project.md
```diff
@@ -110,16 +110,31 @@ This Pulumi program creates an Azure resource group and storage account and expo
...
+
+> *Note*: As a prerequisite, [install virtualenv](https://virtualenv.pypa.io/en/latest/installation/) to manage project requirements
+
 For Python, before we deploy the stack, the following commands need to be run to create a virtual environment, activate it, and install dependencies:
 
+Create a virtual environment:
+
 ```bash
-$ virtualenv -p python3 venv
+$ python -m venv azure-env
```
lukehoban: I believe our offline discussion was that we in general cannot expect `Python` to be Python 3?

Cc also @jen20.

content/docs/get-started/aws/review-project.md
```diff
@@ -82,16 +82,31 @@ class Program
 This Pulumi program creates an S3 bucket and exports the name of the bucket.
 
 {{% lang python %}}
+
+> *Note*: As a prerequisite, [install virtualenv](https://virtualenv.pypa.io/en/latest/installation/) to manage project requirements
```
lukehoban: Can we put this on the install python page of the getting started instead?  We may as well just tell folks to install this upfront. 

We should also make sure this requirement is clearly called out on the Python language page.

content/docs/get-started/azure/review-project.md
```diff
@@ -110,16 +110,31 @@ This Pulumi program creates an Azure resource group and storage account and expo
...
 ```
 
+Activate the environment in Linux and MacOS:
+
 ```bash
-$ source venv/bin/activate
+$ source azure-env/bin/activate
```
lukehoban: Do we really want to use names like `azure-env` here vs a more generic `venv`?  I don’t know the namespacing of these names - but I can’t come up with a reason why `azure-env` feels right for this (either it’s too specific or not specific enough).

content/docs/reference/pkg/nodejs/pulumi/vault/lDAP/_index.md
```diff
@@ -1,1282 +0,0 @@
----
```
lukehoban: I assume this is getting deleted bashed we changed the capitalization of its name or something?  I hope we are not actually deleting any content?

content/docs/get-started/aws/review-project.md
justinvp: We'll have to update the copy button script to ignore the shell prompt for `bat`, so it isn't included in the clipboard (like we do for `bash`/`shell`).

content/docs/get-started/azure/review-project.md
justinvp: We should use `venv` for the reasons I commented earlier in https://github.com/pulumi/docs/pull/2184#discussion_r360544638

content/docs/reference/pkg/nodejs/pulumi/vault/lDAP/_index.md
justinvp: Yeah, this is fallout from a capitalization change in the provider. https://github.com/pulumi/docs/pull/2224 cleans it up.

content/docs/get-started/aws/review-project.md
spara: I'll add it back in, should I create an issue for the copy script button?

justinvp: Or just leave it off, and I’ll take care of adding a prompt for all the windows snippets as part of updating the copy butt script. Please open an issue, and assign to me.

content/docs/get-started/azure/review-project.md
spara: For Linux, CentOS doesn't have a default python, Ubuntu/Debian includes 2 and 3. macOS before Catalina has 2.7 as default and python3 has to be user installed. In those cases, `python3` would be correct. Let's go with `python3`  for now until version checking for python is implemented.

content/docs/get-started/aws/review-project.md
spara: I've updated the Python language page and removed it from the getting started pages

content/docs/get-started/azure/review-project.md
spara: venv is very generic and also the name of the executable. Because python can be sensitive to dependencies, it's generally good practice to create a virtual environment for each project. Here's the example from venv

`c:\>python -m venv c:\path\to\myenv`

We can go with just `myenv` to reduce confusion,

justinvp: Whatever we choose, we'll need to update the `.gitignore` file for all of our Python templates to exclude the virtual environment directory.

Is `venv` really so bad? It seems like it's commonly used based on GitHub's default `.gitignore` for Python:

https://github.com/justinvp/test123/blob/ab56476cad18c989f67d573033d80f16967fd2a5/.gitignore#L104-L111

```
# Environments
.env
.venv
env/
venv/
ENV/
env.bak/
venv.bak/
```

spara: OK, not my hill. `python3 -m venv venv` it is

dstine:
The docs as currently published on the site:
* No indication that `virtualenv` is a pre-requisite
* Installation snippets use `virtualenv` to create virtual environment

This PR at the moment:
* Indicates that `virtualenv` is a pre-requisite
* Installation snippets use `venv` to create virtual environment

My comments:

`virtualenv` is indeed a standalone tool so if its use is recommended, it should be mentioned. But then `venv` has been part of Python standard library since 3.3 ([link](https://docs.python.org/3/library/venv.html)].  And I agree that one should favor `venv` as in this PR.

If Pulumi supports Python 2.7 (I couldn't quickly figure this out) then you might want to document both approaches. But if Pulumi supports only Python 3+, then no one should be using `3.0 <= version < 3.3` and documentation unconditionally utilizing `venv` would be my recommendation.

dstine:
Addendum to prior comment:

I found the statement that only Python 3.6+ is supported. So no need to mention or document `virtualenv` at all.

https://www.pulumi.com/docs/intro/languages/python/

content/docs/get-started/aws/review-project.md
```diff
@@ -105,16 +105,29 @@ class Program
...
 
 {{% lang python %}}
+
 For Python, before we deploy the stack, the following commands need to be run to create a virtual environment, activate it, and install dependencies:
 
+Create a virtual environment:
+
 ```bash
-$ virtualenv -p python3 venv
+$ python3 -m venv venv
```
lukehoban: We’ve seen many issues related to Windows not installing a `python3` binary by default. Anything we can do to make sure these docs are usable on Windows?  Perhaps just a note early on in the flow that Windows users may need to use `python` instead?

content/docs/get-started/aws/review-project.md
```diff
@@ -105,16 +105,29 @@ class Program
...
+Activate the environment in Windows:
+
+```bat
+> venv\Scripts\activate
+```
+
+Install dependencies:
+
 ```bash
 $ pip3 install -r requirements.txt
```
lukehoban: If we are going the `Python -m` route - should this be `python3 -m pip`?

content/docs/get-started/kubernetes/review-project.md
```diff
@@ -145,18 +145,33 @@ class Program
 This Pulumi program creates an NGINX deployment and exports the name of the deployment.
 
 {{% lang python %}}
+
+> *Note*: As a prerequisite, [install virtualenv](https://virtualenv.pypa.io/en/latest/installation/) to manage project requirements
+
```
lukehoban: I think we decided to remove this?

content/docs/get-started/kubernetes/review-project.md
```diff
@@ -145,18 +145,33 @@ class Program
...
 For Python, before we deploy the stack, the following commands need to be run to create a virtual environment, activate it, and install dependencies:
 
+Create a virtual environment:
+
 ```bash
-$ virtualenv -p python3 venv
+$ python -m venv k-env
```
lukehoban: Intentional that this is different than others?  (Python va Python3 and k-env vs venv)

content/docs/intro/languages/python.md
```diff
@@ -40,7 +44,7 @@ This will leave behind a `Pulumi.yaml` file, containing some minimal metadata ab
 
 ## Using Pulumi PyPI Packages {#pypi-packages}
 
-It is not required, but we recommend using [`pipenv`](https://pipenv.readthedocs.io/en/latest/) for repeatable installations that work no matter your operating system.
+It is not required, but we recommend using virtual environments such as [`venv`](https://docs.python.org/3/library/venv.html) of [`pipenv`](https://pipenv.readthedocs.io/en/latest/) for repeatable installations that work no matter your operating system.
 
 ```
 $ pipenv --python 3
```
lukehoban: This expands the scope a bit - but I think we are still very inconsistent in recommendations across venv, virtualenv and pipenv in various places in docs and product. If we’re sure what’s proposed in these getting started docs is right, we should open an issue to track standardizing on this through everything else.

content/docs/get-started/aws/review-project.md
dstine: I believe it should be `pip install ...` since you have already instructed the user to activate the virtual environment, and `pip` is bootstrapped in by default.

https://docs.python.org/3/library/venv.html (see discussion of `—without-pip`)

Related: looks like the activate command for Windows is missing in the site preview.

content/docs/intro/languages/python.md
dstine: This is definitely an area in which it is possible to get bogged down. There are a variety of tooling options and opinions. If you all have experience-driven and/or Pulumi-specific reasons to recommend `pipenv` that is one thing; if not, I suggest keeping it simple. Something like:

a) we recommend virtual environments for various reasons (should be uncontroversial)

b) you should create and manage your virtual environments in whatever way works best for you, and if you don’t happen to already have a preferred approach then we show you how with the `venv` module of the standard library which you already have

I don’t recall seeing this written down explicitly, but I expect a key reason you are investing in supporting the Pulumi embedded DSL in many host languages is to “meet people where they are”. So most new Pulumi Python users would not be new to the Python ecosystem, and would arrive with a preferred approach already.

(I personally used `pipenv` for awhile, became disillusioned for both technical and other reasons and moved to a combination of `venv` and `pip-tools`. But that is just for context of my comment, not a one-size-fits-all recommendation.)

lukehoban: Thanks for all the thoughts and background - definitely helpful!  

We’ve seen folks use a wide variety of tools in this area already with Pulumi, and the good thing of course is that Pulumi itself isn’t really tied to any one of these - it fits well into the existing ecosystem of tools. 

The challenge we’ve faced is more in picking the right “golden path” tooling to use for docs, examples, getting started, installers, etc. We’ve been indecisive on this and ended up suggested different tools at different stages of the user journey. The most important thing at this point is just to be consistent so that users don’t get confused - not really to have any strongly held preference for one tools vs another.  

It is true that many/most users coming to Python + Pulumi already know Python, but I’ve actually been a little surprised at how many folks don’t have a ton of Python experience, and really do rely on these guides to walk though not just the Pulumi aspect, but also handhold through the Python (and cloud) configuration. So we do want to have something simple and clear here. 

Sounds like for now just standardizing on venv is a good simple unopionated built-in option we should just adopt across the board.

content/docs/get-started/aws/review-project.md
spara: We can be more explicit and just add instructions for Windows.

Create a virtual environment in Windows:

```
python -m venv venv
``

File: content/docs/get-started/kubernetes/review-project.md (hunk 12 above)
Comment (spara) [reply content/docs/get-started/kubernetes/review-project.md]: yes

File: content/docs/intro/languages/python.md (hunk 14 above)
Comment (spara) [reply content/docs/intro/languages/python.md]: Will remove.

File: content/docs/get-started/kubernetes/review-project.md (hunk 13 above)
Comment (spara) [reply content/docs/get-started/kubernetes/review-project.md]: fixed

File: content/docs/get-started/kubernetes/review-project.md [hunk 15]
```diff
@@ -145,18 +145,38 @@ class Program
...
 
+Create a virtual environment in Windows:
+
+```bat
+$ python -m venv venv
+
+Activate the environment in Windows:
+
+```bat
+> venv\Scripts\activate.bat
```
justinvp: Everywhere else we leave off the `.bat`, so leave it off here.

```suggestion
> venv\Scripts\activate
```

content/docs/get-started/kubernetes/review-project.md
```diff
@@ -145,18 +145,38 @@ class Program
...
+```
+
+Install dependencies:
+
 ```bash
-$ pip3 install -r requirements.txt
+$ pip install -r requirements.txt
```
justinvp: Is this supposed to be `pip` or `pip3`? Elsewhere we still have it as `pip3`.

joeduffy:
@spara @justinvp I just hit this myself---when do we expect this to get merged?

justinvp:
@spara, did you see my most recent comments?

content/docs/get-started/kubernetes/review-project.md
leezen: It's probably better to leave it as pip3 to remove any ambiguity about if we're actually running in a Python3 environment.

leezen:
Couple comments on my recent changes:

* Since we're just repeating the text across the pages, I consolidated it as a shortcode.
* You'll note that it uses "bash" as the language with a `$` prompt. The `$` was more for consistency with everything else and so that copy works correctly. The reason for `bash` is because on those pages, for whatever reason, `language-batch` is not class we can use. I don't think this functionally affects the highlighting despite it being a bit misleading in naming.

layouts/shortcodes/python-venv.html
```diff
@@ -0,0 +1,24 @@
...
+</div>
+
+<p>Activate the environment in Linux and macOS:</p>
+<div class="highlight">
+    <pre class="chroma"><code class="language-bash" data-lang="bash">$ source venv/bin/activate</code></pre>
+</div>
+
+<p>Activate the environment in Windows:</p>
+<div class="highlight">
+    <pre class="chroma"><code class="language-bash" data-lang="bash">$ venv\Scripts\activate</code></pre>
```
justinvp: > * The reason for `bash` is because on those pages, for whatever reason, `language-batch` is not class we can use.

The reason `language-batch` (or `language-bat`) isn't displayed here is due to a bug in how we hide/show languages when there is a langchooser on the page. We need to fix the `chooser.js` script to ignore more "shell languages" when showing/hiding languages (right now it's only ignoring `bash` and `sh`):

https://github.com/pulumi/docs/blob/4101d892b121505f1900b401a1a1e5d0799a53e0/assets/js/chooser.js#L59

Probably easiest to use the same list of languages we use for the copy button and noselect:

https://github.com/pulumi/docs/blob/4101d892b121505f1900b401a1a1e5d0799a53e0/assets/js/copybutton.js#L27-L55

https://github.com/pulumi/docs/blob/4101d892b121505f1900b401a1a1e5d0799a53e0/assets/js/noselect.js#L15-L16

Once that's done, this could be changed to the following:

```suggestion
    <pre class="chroma"><code class="language-bat" data-lang="bat"><span class="p">&gt;</span> venv\Scripts\activate</code></pre>
```

layouts/shortcodes/python-venv.html
```diff
@@ -0,0 +1,24 @@
+<p>
+    For Python, before we deploy the stack, the following commands need to be
+    run to create a virtual environment, activate it, and install dependencies:
+</p>
+
+<p>Create a virtual environment:</p>
+<div class="highlight">
+    <pre class="chroma"><code class="language-bash" data-lang="bash">$ python3 -m venv venv</code></pre>
```
justinvp: On Windows, this should be `python` instead of `python3`. The default installation of Python on Windows doesn't include a `python3` binary.

layouts/shortcodes/python-venv.html
```diff
@@ -0,0 +1,24 @@
...
+</div>
+
+<p>Activate the environment in Linux and macOS:</p>
+<div class="highlight">
+    <pre class="chroma"><code class="language-bash" data-lang="bash">$ source venv/bin/activate</code></pre>
+</div>
+
+<p>Activate the environment in Windows:</p>
```
justinvp: Alternatively, we could use the OS chooser here...

I realize there's duplication for mac/linux, but this allows using the right `python` vs `python3` command on Windows, and allows using the right `>` vs `$` prompts that we've started using elsewhere.

```html
{{/*
This table is effectively the same as oschoose.md, however it
needs to be embedded inline because shortcodes cannot reference
other shortcodes.
*/}}
        <a data-track="oschoose-macOs" class="mdl-tabs__tab os-tab is-active">macOS</a>
        <a data-track="oschoose-linux" class="mdl-tabs__tab os-tab">Linux</a>
        <a data-track="oschoose-windows" class="mdl-tabs__tab os-tab">Windows</a>
    </div>
</div>

{{/*
.mdl-tabs__tab uses floats, so we need to use clearfix to avoid
wrapping/indenting the subsequent element containing the code snippet.
*/}}
<div class="clearfix"></div>

<div class="os-prologue-macos"></div>
<div>
    <p>Create a virtual environment:</p>
    <div class="highlight">
        <pre class="chroma"><code class="language-bash" data-lang="bash">$ python3 -m venv venv</code></pre>
    </div>

    <p>Activate the environment:</p>
    <div class="highlight">
        <pre class="chroma"><code class="language-bash" data-lang="bash">$ source venv/bin/activate</code></pre>
    </div>

    <p>Install dependencies:</p>
    <div class="highlight">
        <pre class="chroma"><code class="language-bash" data-lang="bash">$ pip3 install -r requirements.txt</code></pre>
    </div>
</div>

<div class="os-prologue-linux"></div>
<div>
    <p>Create a virtual environment:</p>
    <div class="highlight">
        <pre class="chroma"><code class="language-bash" data-lang="bash">$ python3 -m venv venv</code></pre>
    </div>

    <p>Activate the environment:</p>
    <div class="highlight">
        <pre class="chroma"><code class="language-bash" data-lang="bash">$ source venv/bin/activate</code></pre>
    </div>

    <p>Install dependencies:</p>
    <div class="highlight">
        <pre class="chroma"><code class="language-bash" data-lang="bash">$ pip3 install -r requirements.txt</code></pre>
    </div>
</div>

<div class="os-prologue-windows"></div>
<div>
    <p>Create a virtual environment:</p>
    <div class="highlight">
        <pre class="chroma"><code class="language-bat" data-lang="bat"><span class="p">&gt;</span> python -m venv venv</code></pre>
    </div>

    <p>Activate the environment:</p>
    <div class="highlight">
        <pre class="chroma"><code class="language-bat" data-lang="bat"><span class="p">&gt;</span> venv\Scripts\activate</code></pre>
    </div>

    <p>Install dependencies:</p>
    <div class="highlight">
        <pre class="chroma"><code class="language-bat" data-lang="bat"><span class="p">&gt;</span> pip3 install -r requirements.txt</code></pre>
    </div>
</div>
```

layouts/shortcodes/install-pulumi.html
```diff
@@ -35,13 +35,13 @@
...
+    <p>Install Pulumi on Windows through the <a href="https://chocolatey.org" target="_blank">Chocolatey package manager</a>:</p>
     <div class="highlight">
-        <pre class="chroma"><code class="language-batch" data-lang="batch">@"%SystemRoot%\System32\WindowsPowerShell\v1.0\powershell.exe" -NoProfile -InputFormat None -ExecutionPolicy Bypass -Command "[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; iex ((New-Object System.Net.WebClient).DownloadString('https://get.pulumi.com/install.ps1'))" && SET "PATH=%PATH%;%USERPROFILE%\.pulumi\bin"</code></pre>
+        <pre class="chroma"><code class="language-batch" data-lang="batch">choco install pulumi</code></pre>
     </div>
 </div>
 
 <p>
-    <i>For manual installation instructions, other platforms, or troubleshooting, see
+    <i>For alternative installation instructions (e.g. script-based installation, Windows installer, etc.) or troubleshooting, see
```
justinvp: Nit: We have a Windows installer?

layouts/shortcodes/python-venv.html
leezen: Thanks for letting me know where that is!

content/docs/get-started/install/_index.md
```diff
@@ -131,6 +119,20 @@ Subsequent updates can be installed in the usual way:
 > choco upgrade pulumi
 ```
 
+### Installation Script
+
+Windows 8 and 10 are supported.
```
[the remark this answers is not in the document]
leezen: This is under Chocolatey? Unless I'm not understanding the feedback?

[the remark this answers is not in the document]
justinvp: I'm suggesting we move the text "Windows 8 and 10 are supported." to be under the first heading for Windows, which is now `### Chocolatey`.

layouts/shortcodes/install-pulumi.html
```diff
@@ -35,13 +35,13 @@
 
 <div class="os-prologue-windows"></div>
 <div class="install_windows">
-    <p>Install Pulumi on Windows by running the installation script from a <code>cmd.exe</code> window:</p>
+    <p>Install Pulumi on Windows through the <a href="https://chocolatey.org" target="_blank">Chocolatey package manager</a>:</p>
```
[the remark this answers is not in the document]
leezen: I'm going to fix it for this particular shortcode, but we have `target="_blank"` in a few hundred places throughout the docs. I've opened a separate issue to specifically address that: https://github.com/pulumi/docs/issues/2575

content/docs/get-started/install/_index.md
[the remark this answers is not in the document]
leezen: Ah, now I understand. I think if that's the case, if anything, why wouldn't we have this above the first heading to make it clear it applies to everything?

content/docs/get-started/install/_index.md
```diff
@@ -131,6 +119,20 @@ Subsequent updates can be installed in the usual way:
...
+
+1. Open a new command prompt window (**WIN+R**: `cmd.exe`):
+
+1. Run our installation script:
+
+   ```bat
+   @"%SystemRoot%\System32\WindowsPowerShell\v1.0\powershell.exe" -NoProfile -InputFormat None -ExecutionPolicy Bypass -Command "[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; iex ((New-Object System.Net.WebClient).DownloadString('https://get.pulumi.com/install.ps1'))" && SET "PATH=%PATH%;%USERPROFILE%\.pulumi\bin"
+   ```
+
+   This will install the `pulumi.exe` CLI to `%USERPROFILE%\.pulumi\bin` and add it to your path.
```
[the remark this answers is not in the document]
leezen: Fixed

layouts/shortcodes/install-pulumi.html
leezen: Yeah... I guess I shouldn't call it that. It's really just copy binaries from a .zip.

layouts/shortcodes/python-venv.html
leezen: Great suggestion!

assets/js/chooser.js
```diff
@@ -52,11 +52,16 @@ function selectChoice(kind, choice, extra) {
...
+            function(l) {
+                return "language-" + l;
+            }
+        );
         // In addition to the basic showing/hiding, highlight the right code blocks:
         $("code").each(function (i, e) {
             var classes = getElemClasses(e);
             for (var i = 0; i < classes.length; i++) {
-                if (classes[i].startsWith("language-") && classes[i] !== "language-bash" && classes[i] !== "language-sh") {
+                if (classes[i].startsWith("language-") && !shellLanguages.includes(classes[i])) {
```
justinvp: [IE11 doesn't support `includes`](https://caniuse.com/#feat=array-includes). While I don't think we need to concern ourselves much with IE these days (I'm not even sure how well the site renders in IE11 in general), switching to using `indexOf` instead would allow this to at least work in IE11 without any script errors.

I know we see errors on https://app.pulumi.com every once and a while due to someone browsing the site with IE and usage of `includes` there.

```suggestion
                if (classes[i].startsWith("language-") && !shellLanguages.indexOf(classes[i]) !== -1) {
```

content/docs/get-started/install/_index.md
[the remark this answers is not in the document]
justinvp: > why wouldn't we have this above the first heading to make it clear it applies to everything?

Makes sense. Mind doing the same for macOS while you're at it?

assets/js/chooser.js
leezen: Wow, yeah, it's definitely been a while since I've written JS destined straight for the browser. I totally forgot that that kind of stuff still happens. Thanks for the suggestion!

content/docs/get-started/install/_index.md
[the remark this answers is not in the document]
leezen: Will do

layouts/shortcodes/install-pulumi.html
```diff
@@ -35,13 +35,13 @@
 
 <div class="os-prologue-windows"></div>
 <div class="install_windows">
-    <p>Install Pulumi on Windows by running the installation script from a <code>cmd.exe</code> window:</p>
+    <p>Install Pulumi on Windows through the <a href="https://chocolatey.org" target="_blank" rel="noopener">Chocolatey package manager</a>:</p>
     <div class="highlight">
-        <pre class="chroma"><code class="language-batch" data-lang="batch">@"%SystemRoot%\System32\WindowsPowerShell\v1.0\powershell.exe" -NoProfile -InputFormat None -ExecutionPolicy Bypass -Command "[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; iex ((New-Object System.Net.WebClient).DownloadString('https://get.pulumi.com/install.ps1'))" && SET "PATH=%PATH%;%USERPROFILE%\.pulumi\bin"</code></pre>
+        <pre class="chroma"><code class="language-batch" data-lang="batch">choco install pulumi</code></pre>
```
justinvp: Nit:

```suggestion
        <pre class="chroma"><code class="language-bat" data-lang="bat"><span class="p">&gt;</span> choco install pulumi</code></pre>
```

assets/js/chooser.js
```diff
@@ -52,11 +52,16 @@ function selectChoice(kind, choice, extra) {
...
+            function(l) {
+                return "language-" + l;
+            }
+        );
         // In addition to the basic showing/hiding, highlight the right code blocks:
         $("code").each(function (i, e) {
             var classes = getElemClasses(e);
             for (var i = 0; i < classes.length; i++) {
-                if (classes[i].startsWith("language-") && classes[i] !== "language-bash" && classes[i] !== "language-sh") {
+                if (classes[i].startsWith("language-") && !shellLanguages.indexOf(classes[i]) !== -1) {
```
justinvp: Oops, this needs to be the following:

```suggestion
                if (classes[i].startsWith("language-") && shellLanguages.indexOf(classes[i]) === -1) {
```

layouts/shortcodes/python-venv.html
```diff
@@ -0,0 +1,66 @@
+{{/*
```
justinvp: The following text was in all the places where this shortcode is now used, but wasn't carried over to this file. Was that intentional, or do we need to add it at the top of this file?

```html
<p>
    For Python, before we deploy the stack, the following commands need to be run to create a virtual environment, activate it, and install dependencies:
</p>
```

leezen: Not intentional. I think I just missed it when copy and pasting into the shortcode.

PR #2193, opened 2019-12-23
Remove CE notes in Product and Why Pulumi pages

cnunciato:
These lines were added in an attempt to make it clear there's a Community Edition and that it's free, but given the recent changes to the Pricing page, they're arguably no longer be necessary (since the VIEW PRICING buttons now bring you to a very that does make that clear).

### DIFF
--- a/layouts/page/product.html
+++ b/layouts/page/product.html
@@ -8,11 +8,6 @@
                     Easily go to production with modern cloud architectures, and break down
                     silos inhibiting innovation within your organization.
                 </p>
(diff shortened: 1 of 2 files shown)


cnunciato:
@joeduffy just for consideration, wondering what you think.

cnunciato:
Thanks, I appreciate that. I'm quite confident in how clear the Pricing page makes this now, and I'd like to remove elements that distract from the messaging in these areas, along with anything that makes Community Edition look like it might be something _else_.

PR #2199, opened 2019-12-27
Add click tracking to all html links

zchase:
This PR adds link tracking to the across all the HTML links we manually write ourselves. This doesn't include support for links defined in markdown files. There is also a script for lint the HTML to ensure our links are tracked properly.

The naming convention for the Ids is generated in two ways, following the same common pattern. The first way is autogenerated based on the current path, the target path, and the index of the link on the page. The second way replaces the target path with the `data-track` attribute if it exists on the link. 

All the data is being pushed into Google Analytics where it is sorted by page. For now, we are going to rely on GA to sort the events out by the page.


scripts/link_linter.js
```diff
@@ -0,0 +1,191 @@
+const parse = require("node-html-parser");
+const fs = require('fs');
+const path = require('path');
+
+/**
+ * Recursively search for html files and add them to the result list.
```
chrsmith: NIT: Since it is an acronym, HTML should be all caps, here and throughout.

scripts/link_linter.js
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * @param {string[]} files A list of files to search through.
+ * @param {string[]} result A list of complete file paths to html files.
+ */
+function fetchFilePaths(files, result = []) {
+    // Grab the first file in the list and build its full path.
+    const file = files[0];
```
chrsmith: What if `files` is undefined/empty? Rather than just throwing with an index out of bounds error here, perhaps we should just return `if (!files || files.length === 0)`?

scripts/link_linter.js
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    // Grab everything leading up the start of the specified string.
+    const shortenedBlock = blockText.substr(0, stringIndex);
+
+    // Split the shortened block by '\n' and return the length of the array.
+    return shortenedBlock.split("\n").length;
+}
+
+const DATA_TRACKING_ID_REGEX = /data-tracking-id=("|')([a-zA-Z0-9]+)([-{ ()$.a-zA-Z0-9}]+)("|')/g;
```
chrsmith: Is there any reason why we don't want to only allow `[a-zA-Z0-9_-.]+`? Allowing `{ ()` seems like it's just asking for trouble, since we'd be at the mercy of how whatever 3rd party system interprets/honors the data. (e.g. GA might URL encode the value, replacing " " with "%20" automatically.

It's a pain to have a stricter format for these tracking IDs, but since we control the IDs put in our markup, it's probably better to start strict now since it would be harder to change later.

scripts/link_linter.js
```diff
@@ -0,0 +1,191 @@
+const parse = require("node-html-parser");
```
chrsmith: Adding a lint rule to enforce that we are setting the newly added data links on all `<a>` tags? I wear a single emoji tear of joy. 😢 😀

assets/js/tracking.js
```diff
@@ -19,15 +19,15 @@ $(document).ready(function() {
             // Create the tracking object.
             const trackingData = {
                 // The id of the element.
-                element_id: $(element).attr("id"),
+                element_id: $(element).attr("data-tracking-id"),
```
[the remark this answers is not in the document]
zchase: A bit paranoid :), but I'll add some code to handle if the element doesn't exist. The way the script works is that only links found on the page will be fed into the function. But this is the world of browser javascript so it's probably best to err on the side of caution.

If the attribute isn't present it won't cause a "problem" for the user. We will just have useless data to look at. The id should always be present though because the I query for links with `data-tracking-id` and then that list is looped over and fed into this function.

layouts/blog/single.html
```diff
@@ -28,7 +28,7 @@ <h1 class="no-anchor"><a href="{{ .Permalink }}">{{ .Title }}</a></h1>
                         <ul class="m-0 p-0">
                             {{ range $tag := .Params.tags }}
                             <li class="mt-0 inline-flex">
-                                <a class="tag tag-blog" href="/blog/tag/{{ $tag | urlize }}/">
+                                <a data-tracking-id="blog-single-tag-{{ $tag }}" class="tag tag-blog" href="/blog/tag/{{ $tag | urlize }}/">
```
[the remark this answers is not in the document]
zchase: Added `urlize` to dynamic ids.

scripts/link_linter.js
zchase: Added a check at the top of the function for this.

scripts/link_linter.js
zchase: In the current state of the linter, the files are linted pre-compile. So before `hugo` does its magic. This means there will be ids like `shortcodes-glossary-{{ .term | urlize }}`, `shortcodes-changlelog-row-{{ $version | urlize }}`, `home-article-{{ (index .Params.articles 1).title | urlize }}`, etc...

By using `hugo` variables we could get rid of most of the symbols, and in that case we just have to include `{}` and `$` as symbols in the regex. 

So I think the best solution is just to add rules to validate the IDs are in the correct format.

layouts/partials/contact-us.html
```diff
@@ -7,17 +7,17 @@ <h3>Ask Us About</h3>
...
-                        <a href="{{ relref . "/why-pulumi/delivering-cloud-native-infrastructure-as-code" }}">
+                        <a data-tracking-id="partials-contact-us-why-pulumi" href="{{ relref . "/why-pulumi/delivering-cloud-native-infrastructure-as-code" }}">
                             cloud native infrastructure as code</a>, with general purpose programming languages.
                     </li>
                     <li>
                         How Pulumi supports infrastructure, managed services,
-                        <a href="{{ relref . "/topics/containers.md" }}">containers</a>,
-                        <a href="{{ relref . "/topics/kubernetes.md" }}">Kubernetes</a>, and
-                        <a href="{{ relref . "/topics/serverless.md" }}">serverless</a> infrastructure and apps.
+                        <a data-tracking-id="partials-contact-us-containers" href="{{ relref . "/topics/containers.md" }}">containers</a>,
```
cnunciato: Not a big deal, but just as a head's up, Justin's got a PR (#2196) to remove all `.md` extensions, so you might want to rebase on that, and replace-out the ones in the files you're touching as well, in case Git gets confused.

cnunciato:
Nice! Much bigger fan of using a `data-*` attribute for these -- thanks.

A few comments to echo what we discussed offline a bit:

Currently, the linter doesn't track duplicates, which should be a simple thing to add -- for example, this passes: 

```
<a data-tracking-id="stuff" href="https://somewhere.com">Go here.</a>
<a data-tracking-id="stuff" href="https://somewhere.com">No, here.</a>
```

Perhaps we could build up a map of them, and check to make sure someone hasn't copy-pasted a link from elsewhere, as will likely be the case, and easy to miss on a code review. We should also provide some sort of guide as to how tracking IDs are crafted, so we don't end up with a bunch of link-tracking soup down the road somehow. 

And I agree with Chrisit'd be good to make these IDs as URL/machine-friendly as we can, though, so maybe a helper of some sort would be called for here, too, just to keep things tidy from the get-go.

More broadly, the requirement to put a unique, hand-crafted `data-tracking-id` attribute on _every single link_ isn't sitting well with me. It feels burdensome, especially considering most of these IDs can be inferred/constructed from their files, DOM paths, and so on. I suspect we might be able to get away with  attributing some sort of  "category", maybe, to denote the kind of thing we care about (CTAs, outbound links, etc.) and then programmatically (say, with some JS to capture and report the clicks before page unload) to identify the which actual thing in that category of things was clicked on, which could reduce the burden on the team and probably result in cleaner data over time.

That said, this definitely feels like it's heading in the right direction! Thanks again for putting so much thought into it.

layouts/partials/contact-us.html
justinvp: Ah, sorry about the merge conflicts @zchase :(

zchase:
@cnunciato Here is the new and improved link tracking based on our conversations. Let me know if this has landed in a good spot or there is still some more work to hammer out here.

layouts/docs/glossary.html
```diff
@@ -5,12 +5,12 @@
...
 {{ end }}
-    
+
 {{ range sort $.Site.Data.glossary.entries "term" }}
     <div class="flex-1 mt-8">
         <h2 class="mt-2 mb-0">{{ .term }} </h2>
             <p>{{ .description }}</p>
                 <ul class="mb-4">
-                    <li>Read More: <a href="{{ .link }}" class="">{{ .term }}</a></li>
+                    <li>Read More: <a data-track="read-more-{{ .term | urlize }}-read-more" href="{{ .link }}" class="">{{ .term }}</a></li>
```
justinvp: Just curious: Why is there both a `read-more-` prefix and `-read-more` suffix in the `data-track` value?

layouts/partials/docs/breadcrumb.html
```diff
@@ -14,6 +14,8 @@
     {{ end }}
     <!-- If the page is a redirect page, don't include it in the breadcrumb. -->
     {{ if not (isset .p1.Params "redirect_to") }}
-        <li {{ if eq .p1 .p2 }}class="active"{{ end }}><a href="{{ .p1.Permalink }}">{{ .p1.LinkTitle }}</a></li>
+        <li {{ if eq .p1 .p2 }}class="active"{{ end }}>
+            <a data-track="breadcrumb" href="{{ .p1.Permalink }}">{{ .p1.LinkTitle }}</a>
+        </li>
```
justinvp: It looks like this change is causing a space to show up in front of bread crumb items for some reason. Not sure why the added whitespace here is causing that to happen.

http://content-bucket-a1a482d.s3-website-us-east-1.amazonaws.com/docs/intro/concepts/

Screen Shot 2020-01-09 at 2 38 10 PM

https://www.pulumi.com/docs/intro/concepts/

Screen Shot 2020-01-09 at 2 40 59 PM

layouts/shortcodes/install-pulumi.html
```diff
@@ -5,9 +5,9 @@
 */}}
 <div class="mdl-tabs os-tab-tabs">
     <div class="mdl-tabs__tab-dir os-tab-tabstrip">
-        <a class="mdl-tabs__tab os-tab is-active">macOS</a>
-        <a class="mdl-tabs__tab os-tab">Linux</a>
-        <a class="mdl-tabs__tab os-tab">Windows</a>
+        <a data-track="oschoose-macOs" class="mdl-tabs__tab os-tab is-active">macOS</a>
```
justinvp: Nit:

```suggestion
        <a data-track="oschoose-macos" class="mdl-tabs__tab os-tab is-active">macOS</a>
```

layouts/shortcodes/oschoose.html
```diff
@@ -1,8 +1,8 @@
 <div class="mdl-tabs os-tab-tabs">
     <div class="mdl-tabs__tab-dir os-tab-tabstrip">
-        <a class="mdl-tabs__tab os-tab is-active">macOS</a>
-        <a class="mdl-tabs__tab os-tab">Linux</a>
-        <a class="mdl-tabs__tab os-tab">Windows</a>
+        <a data-track="oschoose-macOS" class="mdl-tabs__tab os-tab is-active">macOS</a>
```
justinvp: Nit:

```suggestion
        <a data-track="oschoose-macos" class="mdl-tabs__tab os-tab is-active">macOS</a>
```

package.json
```diff
@@ -2,11 +2,13 @@
...
   "version": "0.0.1",
   "scripts": {
-    "lint-markdown": "node ./scripts/markdown_seo_lint.js"
+    "lint-markdown": "node ./scripts/markdown_seo_lint.js",
+    "lint-links": "node ./scripts/link_linter.js"
   },
   "dependencies": {
     "broken-link-checker": "^0.7.8",
     "clipboard-polyfill": "^2.8.1",
+    "node-html-parser": "^1.1.16",
```
cnunciato: Is this still needed? I don't see it being used anywhere.

layouts/index.html
```diff
@@ -11,13 +11,13 @@ <h1 class="font-sans font-body font-normal">
                 {{ .Params.subhead | markdownify }}
             </p>
             <div class="mt-6">
-                <a id="home-hero-cta" href="{{ relref . "/docs/get-started" }}" class="btn bg-purple-300 hover:bg-purple-200 border-none px-8 py-4 transition-all text-sm rounded-full uppercase shadow-lg" title="Create your first project with Pulumi" track-click>
+                <a href="{{ relref . "/docs/get-started" }}" class="btn bg-purple-300 hover:bg-purple-200 border-none px-8 py-4 transition-all text-sm rounded-full uppercase shadow-lg" title="Create your first project with Pulumi">
```
cnunciato: Why not track these?

layouts/index.html
```diff
@@ -11,13 +11,13 @@ <h1 class="font-sans font-body font-normal">
...
                     Get Started
                 </a>
             </div>
             <div class="mt-8 text-sm text-purple-100">
-                Pulumi is <a id="home-hero-pricing" class="text-white hover:underline" href="{{ relref . "/pricing#community-edition" }}" track-click>free</a>
-                and <a id="home-hero-github" class="text-white hover:underline" href="https://github.com/pulumi/pulumi" target="_blank" track-click>open source</a>.
+                Pulumi is <a class="text-white hover:underline" href="{{ relref . "/pricing#community-edition" }}">free</a>
```
cnunciato: Same question here (why not track), and below, in this file.

cnunciato:
We should include a README or section somewhere explaining what the naming conventions are now, and update the PR description if necessary, to make this clear. A couple of examples of the intended output might help, too, just to make it clear what the inputs and outputs are.

layouts/legal/single.html
```diff
@@ -5,7 +5,7 @@
                 <aside>
                     <ul class="md:w-8/12 p-0 list-none">
                         {{ range .Site.Menus.legal }}
-                            <li><a href="{{ .URL }}">{{ .Title }}</a></li>
+                            <li><a data-track="{{ .Title | urlize }}" href="{{ .URL }}">{{ .Title }}</a></li>
```
cnunciato: Why track these with a title and not a category like the others (since the href will make it clear what the destination is)?

layouts/page/about.html
```diff
@@ -119,7 +119,7 @@ <h2 class="text-center md:text-left">Recent News</h2>
                             <div class="pl-4 flex-1">
                                 <div class="text-sm text-gray-500">{{ .date }}</div>
                                 <h5 class="mt-2 mb-0">{{ .title }}</h5>
-                                <a class="block my-2" href="{{ .url }}">Read more &rarr;</a>
+                                <a data-track="news-{{ .url | urlize }}" class="block my-2" href="{{ .url }}">Read more &rarr;</a>
```
cnunciato: I'd expect this just to be `news`, or `news-read-on`, or something, rather than `news-https//www.businessinsider.com/18-hottest-enterprise-startups-from-seattles-booming-tech-scene-2019-12`.

layouts/page/about.html
```diff
@@ -132,7 +132,7 @@ <h2 class="text-center mb-8">Awards &amp; Recognitions</h2>
             <ul class="list-none block p-0 inline-flex flex-wrap justify-center">
                 {{ range first 6 (where $.Site.Data.awards.awards "featured" true) }}
                     <li class="mx-4 bg-white">
-                        <a href="{{ .url }}">
+                        <a data-track="award-{{ .url | urlize }}" href="{{ .url }}">
```
cnunciato: Same here -- `award` is probably what you want, or suchlike.

layouts/page/awards.html
```diff
@@ -16,15 +16,15 @@ <h1>{{ .Title }}</h1>
...
                                     <img class="text-center inline-block rounded shadow p-1" src="{{ .img }}">
                                 </a>
                             </div>
                             <div class="md:w-2/3 p-4 text-left">
                                 <h3 class="leading-tight">
-                                    <a class="text-gray-700" href="{{ .url }}">{{ .title }}</a>
+                                    <a data-track="{{ .title | urlize }}" class="text-gray-700" href="{{ .url }}">{{ .title }}</a>
                                 </h3>
-                                <a href="{{ .url }}" class="text-blue-500 text-base">Read more &rarr;</a>
+                                <a data-track="{{ .title | urlize }}-read-more" href="{{ .url }}" class="text-blue-500 text-base">Read more &rarr;</a>
```
cnunciato:
```suggestion
                                <a data-track="award-read-more" href="{{ .url }}" class="text-blue-500 text-base">Read more &rarr;</a>
```

layouts/page/awards.html
```diff
@@ -16,15 +16,15 @@ <h1>{{ .Title }}</h1>
                     <li class="my-0 md:w-1/2 p-4 w-full">
                         <div class="text-center md:text-left md:flex rounded shadow-md bg-white border border-gray-200 p-4">
                             <div class="md:w-1/3 p-4">
-                                <a class="block text-center" href="{{ .url }}">
+                                <a data-track="{{ .title | urlize }}-img" class="block text-center" href="{{ .url }}">
```
cnunciato:
```suggestion
                                <a data-track="award-img" class="block text-center" href="{{ .url }}">
```

layouts/page/awards.html
```diff
@@ -16,15 +16,15 @@ <h1>{{ .Title }}</h1>
...
                                     <img class="text-center inline-block rounded shadow p-1" src="{{ .img }}">
                                 </a>
                             </div>
                             <div class="md:w-2/3 p-4 text-left">
                                 <h3 class="leading-tight">
-                                    <a class="text-gray-700" href="{{ .url }}">{{ .title }}</a>
+                                    <a data-track="{{ .title | urlize }}" class="text-gray-700" href="{{ .url }}">{{ .title }}</a>
```
cnunciato:
```suggestion
                                    <a data-track="award-title" class="text-gray-700" href="{{ .url }}">{{ .title }}</a>
```

layouts/page/pricing.html
```diff
@@ -434,20 +434,20 @@ <h2>Plan details</h2>
...
+                <span class="pricing-item-label"><a data-track="grid-saml" href="{{ relref . "/docs/guides/saml" }}" class="link">SAML/SSO</a></span>
                 <span class="pricing-item-value"><i class="fas fa-times text-grey"></i></span>
                 <span class="pricing-item-value"><i class="fas fa-times text-grey"></i></span>
                 <span class="pricing-item-value">
-                    <a href="{{ relref . "/docs/guides/saml/aad" }}" class="link">Azure Active Directory</a>,
-                    <a href="{{ relref . "/docs/guides/saml/gsuite" }}" class="link">G Suite</a>,
-                    <a href="{{ relref . "/docs/guides/saml/okta" }}" class="link">Okta</a>,
-                    <a href="{{ relref . "/docs/guides/saml/sso" }}" class="link">any SAML SSO 2.0 identity provider</a>
+                    <a data-track="grid-saml-azure" href="{{ relref . "/docs/guides/saml/aad" }}" class="link">Azure Active Directory</a>,
+                    <a data-track="grid-saml-g-suite" href="{{ relref . "/docs/guides/saml/gsuite" }}" class="link">G Suite</a>,
```
cnunciato:
```suggestion
                    <a data-track="grid-saml-gsuite" href="{{ relref . "/docs/guides/saml/gsuite" }}" class="link">G Suite</a>,
```

layouts/page/pricing.html
```diff
@@ -514,11 +514,10 @@ <h3>Can I use Pulumi for free? </h3>
             </p>
             <h3>How do I get started? </h3>
             <p>
-                To get started with a team-based plan, simply <a
-                href="https://app.pulumi.com/signup">sign up for free</a> using your
-                identity provider of choice (ideally the same one your organization will
-                be backed by). From there, <a href="https://app.pulumi.com/site/organizations/add">create
-                an organization</a> that your team members will use.
+                To get started with a team-based plan, simply <a href="https://app.pulumi.com/signup">
```
cnunciato: No track?

layouts/page/pricing.html
```diff
@@ -514,11 +514,10 @@ <h3>Can I use Pulumi for free? </h3>
...
+                sign up for free</a> using your identity provider of choice (ideally the same one your organization will
+                be backed by). From there, <a href="https://app.pulumi.com/site/organizations/add">
```
cnunciato: No track?

layouts/page/pricing.html
```diff
@@ -547,13 +546,14 @@ <h3>What is an organization? </h3>
             </p>
             <h3>What are projects and stacks? </h3>
             <p>
-                Pulumi <a href="{{ relref . "/docs/intro/concepts/project" }}">projects</a> and <a href="{{ relref . "/docs/intro/concepts/stack" }}">stacks</a> are a way to organize
+                Pulumi <a href="{{ relref . "/docs/intro/concepts/project" }}">projects</a>
```
cnunciato: If not tracking these is intentional, we should definitely provide some sort of documentation explaining when and how to use `data-track` attributes. I'll stop commenting on missing attributes for now.

layouts/page/whitepaper.html
```diff
@@ -27,7 +27,7 @@ <h1>
         <div class="container mx-auto">
             <ul>
                 {{ range .Params.sections }}
-                    <li><a href="#{{ .anchor }}">{{ .label }}</a></li>
+                    <li><a data-track="{{ .label | urlize }}" href="#{{ .anchor }}">{{ .label }}</a></li>
```
cnunciato:
```suggestion
                    <li><a data-track="subnav" href="#{{ .anchor }}">{{ .label }}</a></li>
```

layouts/partials/blog/authors.html
```diff
@@ -5,10 +5,10 @@
 
         <span class="flex items-center mr-2">
             {{ if $authorData }}
-                <a class="rounded-full p-1 bg-gray-300 mr-2" href="{{ $authorPage.Permalink }}">
+                <a data-track="{{ $authorData.name | urlize }}-img" class="rounded-full p-1 bg-gray-300 mr-2" href="{{ $authorPage.Permalink }}">
```
cnunciato:
```suggestion
                <a data-track="author-img" class="rounded-full p-1 bg-gray-300 mr-2" href="{{ $authorPage.Permalink }}">
```

layouts/partials/blog/authors.html
```diff
@@ -5,10 +5,10 @@
...
                     <img class="w-8 rounded-full" src="/images/team/{{ $authorData.id }}.jpg" alt="{{ $authorData.name }}" title="{{ $authorData.name }}">
                 </a>
-                <a class="" href="{{ $authorPage.Permalink }}">{{ $authorData.name }}</a>
+                <a data-track="{{ $authorData.name | urlize }}-name" class="" href="{{ $authorPage.Permalink }}">{{ $authorData.name }}</a>
```
cnunciato:
```suggestion
                <a data-track="author-name" class="" href="{{ $authorPage.Permalink }}">{{ $authorData.name }}</a>
```

layouts/partials/blog/sidebar.html
```diff
@@ -27,15 +27,15 @@ <h2 class="no-anchor hidden md:flex mt-4 text">
...
+                <a data-track="sidebar" href="{{ relref . "/docs/get-started" }}">Get started &rarr;</a>
             </p>
         </div>
 
         <ul class="list-none p-0 mb-4 inline-flex flex-wrap text-xs">
             {{ range $tag, $_ := .Site.Taxonomies.tags }}
                 {{ if $tag }}
                     <li>
-                        <a class="tag tag-blog text-xs {{ if eq (lower $tag) (urlize ((lower $.Title))) }} active {{ end }}" href="{{ .Page.Permalink }}">
+                        <a data-track="blog-tag-{{ $tag | urlize }}" class="tag tag-blog text-xs {{ if eq (lower $tag) (urlize ((lower $.Title))) }} active {{ end }}" href="{{ .Page.Permalink }}">
```
cnunciato:
```suggestion
                        <a data-track="blog-tag" class="tag tag-blog text-xs {{ if eq (lower $tag) (urlize ((lower $.Title))) }} active {{ end }}" href="{{ .Page.Permalink }}">
```

layouts/partials/blog/summary.html
```diff
@@ -6,7 +6,7 @@
     {{ .Summary }}
 </p>
 <p>
-    <a href="{{ .Permalink }}">
+    <a data-track="read-more" href="{{ .Permalink }}">
```
cnunciato: Similarly, should the convention be `read-more` or `[kind-of-thing]-read-more`? Is just `read more` enough, given we aren't relying on DOM positioning? It may be, since with the href, we'll probably be able to pinpoint the source click, but this should be clarified somewhere.

layouts/partials/docs/breadcrumb.html
cnunciato: I'm seeing this locally also.

layouts/partials/docs/right-nav.html
```diff
@@ -6,14 +6,14 @@
         {{ $isPythonPackage := hasPrefix .Path "docs/reference/pkg/python/pulumi" }}
         {{ if not (or $isCLICommand $isNodePackage $isPythonPackage $.Page.Params.no_edit_this_page) }}
             <li>
-                <a class="text-gray-600 hover:text-blue-700 text-xs" href="{{ $.Site.Params.repositoryURL }}/edit/{{ $.Site.Params.repositoryBranch }}/content/{{ .Path }}" target="_blank">
+                <a data-track="edit-page" class="text-gray-600 hover:text-blue-700 text-xs" href="{{ $.Site.Params.repositoryURL }}/edit/{{ $.Site.Params.repositoryBranch }}/content/{{ .Path }}" target="_blank">
```
cnunciato: I'd like to know more about why we care about things like this, along with some others in this PR. What questions are we asking by tracking clicks to Edit Page, Request a Change, etc.?

layouts/partials/header.html
```diff
@@ -31,21 +31,23 @@
                     {{ range $.Site.Menus.header }}
                         <li class="whitespace-no-wrap">
                             {{ $active := hasPrefix $.RelPermalink .URL }}
-                            <a href="{{ .URL }}" data-submenu="{{ .Identifier }}" class="block px-4 py-2 my-2 md:px-2 md:my-0 lg:mx-4 rounded {{ if $active }}bg-purple-400 hover:bg-purple-400{{ else }}hover:bg-purple-500{{ end }}">
+                            <a data-track="header-nav-{{ .Name | urlize }}" href="{{ .URL }}" data-submenu="{{ .Identifier }}" class="block px-4 py-2 my-2 md:px-2 md:my-0 lg:mx-4 rounded {{ if $active }}bg-purple-400 hover:bg-purple-400{{ else }}hover:bg-purple-500{{ end }}">
```
cnunciato:
```suggestion
                            <a data-track="header-nav" href="{{ .URL }}" data-submenu="{{ .Identifier }}" class="block px-4 py-2 my-2 md:px-2 md:my-0 lg:mx-4 rounded {{ if $active }}bg-purple-400 hover:bg-purple-400{{ else }}hover:bg-purple-500{{ end }}">
```

layouts/partials/section-nav.html
```diff
@@ -2,7 +2,7 @@
     <div class="container mx-auto uppercase">
         <ul>
             {{ range .sections }}
-                <li><a href="#{{ .id }}">{{ .label }}</a></li>
+                <li><a data-track="nav-{{ .label | urlize }}" href="#{{ .id }}">{{ .label }}</a></li>
```
cnunciato:
```suggestion
                <li><a data-track="nav" href="#{{ .id }}">{{ .label }}</a></li>
```

... since the href contains the target, and that should be enough to identify what was clicked. 

We should also set down some conventions for naming these, since they vary a bit between pages.

layouts/partials/widgets/social-icons.html
```diff
@@ -11,7 +11,7 @@
         {{ range $idx, $platform := $platforms }}
             {{ if (index $socialIdentities $platform) }}
                 <li class="mx-2">
-                    <a href="{{ index $platformUrlPrefixes $idx }}/{{ index $socialIdentities $platform }}">
+                    <a data-track="{{ $platform | urlize }}" href="{{ index $platformUrlPrefixes $idx }}/{{ index $socialIdentities $platform }}">
```
cnunciato:
```suggestion
                    <a data-track="social-icon" href="{{ index $platformUrlPrefixes $idx }}/{{ index $socialIdentities $platform }}">
```

layouts/shortcodes/cloud-intro.html
```diff
@@ -2,7 +2,8 @@
     <p>
         Pulumi's infrastructure-as-code SDK helps you create, deploy, and manage {{ .Get 0 }} containers,
         serverless functions, and infrastructure using real programming languages. The Pulumi {{ .Get 0 }} provider packages
-        and <a href="{{ relref . "/docs/reference/cli" }}">CLI</a> help you accomplish all these within minutes.
+        and <a href="{{ relref . "/docs/reference/cli" }}">CLI</a>
+        help you accomplish all these within minutes.
```
cnunciato: Outside this PR, but since we're in there:

```suggestion
        help you accomplish all of these within minutes.
```

layouts/shortcodes/cloudchoose.html
```diff
@@ -1,8 +1,8 @@
 <div class="mdl-tabs cloud-tab-tabs">
     <div class="mdl-tabs__tab-dir cloud-tab-tabstrip">
-        <a class="mdl-tabs__tab cloud-tab is-active">AWS</a>
-        <a class="mdl-tabs__tab cloud-tab">Azure</a>
-        <a class="mdl-tabs__tab cloud-tab">GCP</a>
+        <a data-track="cloudchoose-aws" class="mdl-tabs__tab cloud-tab is-active">AWS</a>
```
cnunciato: These raise the question of how we'll use these on the back end, which I think this PR should address somehow as well, since there are so many additions here. For example, if we wanted to ask ourselves the question, "Which cloud do docs users choose most often?", I'd imagine the ideal (pseudo-)query we'd want to write would be something like, "Select all clicks of cloudchoose type, grouped by provider". Would this naming scheme support that kind of query?

layouts/shortcodes/glossary.html
```diff
@@ -12,6 +12,6 @@ <h2 id="{{ anchorize .term }}">{{ .term }}</h2>
     {{ if hasPrefix $link "/" }}
         {{ $link = relref $.Page $link }}
     {{ end }}
-    <a href="{{ $link }}"><i class="fab fa-readme mr-2"></i>{{ .term }}</a>
+    <a data-track="{{ .term | urlize }}" href="{{ $link }}"><i class="fab fa-readme mr-2"></i>{{ .term }}</a>
```
cnunciato:
```suggestion
    <a data-track="glossary-term" href="{{ $link }}"><i class="fab fa-readme mr-2"></i>{{ .term }}</a>
```

layouts/shortcodes/install-python.html
```diff
@@ -12,4 +12,3 @@
         various operating systems and distributions.
     </p>
 </div>
-
```
cnunciato: We probably want to keep this. (VSCode should have a setting for automatically inserting end-of-file newlines.)

layouts/taxonomy/author.html
```diff
@@ -34,7 +34,7 @@ <h1 class="no-anchor">{{ $authorData.name }}</h1>
 
                 {{ range $paginator.Pages }}
                     <article class="my-8 pb-8 border-b border-gray-200">
-                        <h2 class="no-anchor"><a href="{{ .Permalink }}">{{ .Title }}</a></h2>
+                        <h2 class="no-anchor"><a data-track="{{ .Title | urlize }}" href="{{ .Permalink }}">{{ .Title }}</a></h2>
```
cnunciato:
```suggestion
                        <h2 class="no-anchor"><a data-track="author-post" href="{{ .Permalink }}">{{ .Title }}</a></h2>
```

layouts/taxonomy/tag.html
```diff
@@ -16,7 +16,7 @@ <h1 class="no-anchor">Posts Tagged {{ .Data.Term }}</h1>
 
                 {{ range $paginator.Pages }}
                     <article class="my-8 pb-8 border-b border-gray-200">
-                        <h2><a href="{{ .Permalink }}">{{ .Title }}</a></h2>
+                        <h2><a data-track="{{ .Title | urlize }}" href="{{ .Permalink }}">{{ .Title }}</a></h2>
```
cnunciato:
```suggestion
                        <h2><a data-track="tag-post" href="{{ .Permalink }}">{{ .Title }}</a></h2>
```

layouts/docs/glossary.html
zchase: Because I'm an idiot and didn't realize I put it on both sides lol. Fixing.

layouts/partials/docs/breadcrumb.html
zchase: Moving the html back to one line.

package.json
zchase: Removed

layouts/index.html
zchase: They are tracked, but they are auto-generated ids from the current URL path, the target URL, and what number link they are on the page. So, for example, this link gets an id of `home-docs-get-started-12`

layouts/index.html
zchase: See above. All links get auto-generated ids. I used specific Ids in cases where I felt it would be needed.

layouts/page/about.html
zchase: My thought here was from a historical perspective if the links change then we would want to record the target URL. Just so our ids to reflect changes in content.

layouts/page/about.html
zchase: See above. The intention is that changes in the content will change the ids so the data is a bit easier for us to consume (more human-readable).

layouts/shortcodes/cloudchoose.html
zchase: In a way yes. We can filter out IDs in GA to start so we would only see clicks with "cloudchoose" in the ID and then we can look at the list of events. If we only look at a single page at a time, the above query becomes a bit easier. So yes we would be able to, but it will take a bit of leg work outside of GA in its current state. I think this is a good place to start, and we might have to adjust based on the specific questions we want to have answered.

PR #2200, opened 2019-12-27
Fix duplicate h1 tags on blog and contact us

zchase:
This PR fixes duplicate H1 tags on the blog (this sidebar had duplicate h1 tag). This also fixes a duplicate H1 tag from the contact us partial. I ended up having to add a different file for the main blog page sidebar because I couldn't get the partial to accept variables from a dict and still keep the context.


zchase:
@chrsmith Update the blog sidebar to key off the `Title` of the page and removed the `main-sidebar` partial. I think it is a bit more fragile and could break if the name ever changes from "Blog". It is not a big concern of mine as long as we are continually auditing our site (something I do pretty periodically). And fixing the H1 issue would be as easy as just updating the `if... else` to key off the new title.

chrsmith:
> I think it is a bit more fragile and could break if the name ever changes from "Blog". 

I hear you. I still think it's more likely we'll want to make a change to the blog sidebar and have forgotten to update the mostly-identical partial, than change the title of the blog page. ✊

PR #2211, opened 2020-01-02
Fixed broken Route 53 code examples

### LINKED ISSUE #1923: Broken Route 53 Code Examples
[docs/reference/pkg/nodejs/pulumi/aws/acm/_index.md](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/aws/acm/)

I believe the code examples for DNS Validation with Route 53 are incorrect and should be corrected to the following.

```diff
const certValidation = new aws.route53.Record("certValidation", {
+     name: certCertificate.domainValidationOptions[0].resourceRecordName,
      records: [certCertificate.domainValidationOptions[0].resourceRecordValue],
      ttl: 60,
      type: certCertificate.domainValidationOptions[0].resourceRecordType,
-     zoneId: zone.id,
+     zoneId: zone.zoneId,
});
```

This is presumably also incorrect for the Alternative Domains DNS Validation with Route 53 example.

spenserpothier:
### Proposed changes

The examples in [docs/reference/pkg/nodejs/pulumi/aws/acm/_index.md](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/aws/acm/) for using DNS Validation with Route 53 were incorrect as they were using the wrong name for the record.

### Related issues (optional)

Fixes #1923

### DIFF
--- a/content/docs/reference/pkg/nodejs/pulumi/aws/acm/_index.md
+++ b/content/docs/reference/pkg/nodejs/pulumi/aws/acm/_index.md
@@ -314,6 +314,7 @@ const zone = aws.route53.getZone({
     privateZone: false,
 });
 const certValidation = new aws.route53.Record("certValidation", {
+    name: certCertificate.domainValidationOptions[0].resourceRecordName,
     records: [certCertificate.domainValidationOptions[0].resourceRecordValue],
     ttl: 60,
     type: certCertificate.domainValidationOptions[0].resourceRecordType,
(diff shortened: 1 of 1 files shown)


lukehoban:
Thanks @spenserpothier!

As it turns out, these examples are auto-generated, so to permanently fix this we need to fix the source of the problem.  That happens to be the fact that the `pulumi-aws` was erroneously marking the `name` property on `aws.route53.Record` as auto-nameable, when in practice it should never be auto-named.  

I've opened a PR at https://github.com/pulumi/pulumi-aws/pull/842 with the fix for that, which you can see also fixes the API documentation for the case you hit here.  Once that is merged, the docs here will be automatically updated to incorporate that API doc fix.

spenserpothier:
Ah I thought that might be the case but didn't have time to dig further. Glad to hear it's being fixed though. Thanks @lukehoban !

PR #2213, opened 2020-01-03
Adding the HashiCorp providers to the Cloud Providers page

stack72:
Related: #1879


content/docs/intro/cloud-providers/consul/setup.md
```diff
@@ -0,0 +1,28 @@
...
+---
+
+The [Pulumi Consul provider]({{< relref "./" >}}) uses the Consul SDK to manage resources.
+
+> Pulumi relies on the Consul SDK to authenticate requests from your computer to HashiCorp Consul. Your credentials are never sent
+> to pulumi.com.
+
+### Configuring The Provider
+
+Once obtained, there are two ways to communicate your configuration tokens to Pulumi:
```
justinvp: Nit: This says "tokens" plural, but below only shows a single env variable / config value. Should this be singular or the word "tokens" removed from the sentence? Or should it show setting the token env variable / config value?

content/docs/intro/cloud-providers/consul/setup.md
```diff
@@ -0,0 +1,28 @@
...
+
+1. Set the environment variable `CONSUL_HTTP_ADDR`:
+
+    ```bash
+    $ export CONSUL_HTTP_ADDR=XXXXXX
+    ```
+
+2. Set them using configuration, if you prefer that they be stored alongside your Pulumi stack for easy multi-user access:
```
justinvp: Similarly "them" is plural, but we're only showing a single config value.

content/docs/intro/cloud-providers/consul/setup.md
```diff
@@ -0,0 +1,28 @@
...
+
+    ```bash
+    $ pulumi config set consul:address XXXXXX
+    ```
+
+A full set of configuration parameters can be found listed on the [Project README](https://github.com/pulumi/pulumi-consul/blob/master/README.md).
+Remember to pass `--secret` when setting `token` so that it is properly encrypted.
```
justinvp: Again, this mentions `token`, but doesn't actually show it above.

content/docs/intro/cloud-providers/consul/setup.md
stack72: Changed to `communicate your configuration parameters`

content/docs/intro/cloud-providers/consul/setup.md
stack72: Fixed

content/docs/intro/cloud-providers/_index.md
```diff
@@ -25,8 +25,10 @@ aliases: ["/docs/reference/clouds/"]
 
 ## Infrastructure Providers
 
+* [Consul]({{< relref "./consul" >}})
 * [Docker]({{< relref "./docker" >}})
 * [Kubernetes]({{< relref "./kubernetes" >}})
+* [Vault]({{< relref "./vault" >}})
```
lukehoban: I feel like these are in a very different category than Docker and Kubernetes.

And in particular, this ends up *really* burying Kubernetes - which is one of our most important providers.

I think we should probably move Kubernetes back into the "Cloud" section, or create a new section for the "big 4" providers (AWS, Azure, GCP, Kubernetes) to make clear these are the core of what Pulumi is about.

stack72: @lukehoban I am following the same as per the HC docs - we can restructure these as much as you want - I just need to know what way people want them :)

stack72:
@lukehoban I updated the PR to include a new `Core Providers` section - that's where we can add our big 4 :)

Is this better for you?

PR #2214, opened 2020-01-04
Integrate "overview" pages with the nav

joeduffy:
This changes the structure of the nav slightly: instead of showing
"Overview" pages for each sub-section as an independent entry, which
consumes vertical space, hyperlink the section header itself. This
feels more intuitive to me, matches the way the breadcrumbs work,
and saves us on a bit of vertical real estate (which is getting fairly
scarce on our site now-a-days, as we continue to add more sections).

### DIFF
--- a/assets/sass/_toc.scss
+++ b/assets/sass/_toc.scss
@@ -16,7 +16,7 @@
     }
 
     h5 {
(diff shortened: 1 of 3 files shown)


joeduffy:
Before:

image

After:

image

PR #2224, opened 2020-01-07
Rename lDAP to ldap

cnunciato:
I'm assuming it should be `ldap` and not `lDAP`. (Git was insisting I'd made a local change until I realized it was recognizing the case-only rename.)


cnunciato:
I think this is what @spara ran into the other day. 😄

PR #2238, opened 2020-01-10
Fixes #2235.

### LINKED ISSUE #2235: crosswalk doc error
user reported 

on https://www.pulumi.com/docs/guides/crosswalk/aws/iam/#iam-users

 new aws.iam.AccessKey("webmasterKey", { user });

Should be
 
new aws.iam.AccessKey("webmasterKey", { user: user.name });

lukehoban:


lukehoban:
Looks like `pulumify` fails on branches that are numbers (like `2235`)?

```
+ git checkout --progress --force 2235
fatal: reference is not a tree: 2235
```

PR #2239, opened 2020-01-10
Add digital version of event one pager

zchase:
This PR adds a digital version of the event one-pager so we can link to it via QR code.


isaac-pulumi:
Key Requirements:
1. meetup / Happy Hour info should be posted
CTA: "Meet the Pulumi team for a beer on Wednesday 1/22"
- Location: Create33 999 3rd Ave Seattle
 - 5:00-6pm (Happy Hour)
- 6:30-8:30pm (SEA k8s meetup)
2. Link to meetup registration

content/events/one-pager.md
```diff
@@ -0,0 +1,92 @@
+---
+title: Modern Infrastructure as Code
+type: page
+layout: event-one-pager
+url: /show
```
cnunciato: Should this have a more specific URL, like /events/[date] or the like?

We've also talked about creating an Events listing, so perhaps driving this these with Hugo data somehow might better enable figure events to slot in more automatically?

content/events/one-pager.md
```diff
@@ -0,0 +1,92 @@
...
+    title: Meet the Pulumi team for tech talks and beer on Wednesday 1/22
+    body: >
+        Join us at Create33 for talks that focus on the development process on K8s.
+        From 5:00pm to 6:00pm we will be hosting a happy hour with the Meetup starting at 6:30pm.
+
+        The happy hour and Meetup are at Create33, which is located on the 33rd floor of the Wells Fargo
+        Center at <a class="underline text-blue-500" href="https://www.google.com/maps/place/Create33/@47.6050632,-122.3339756,15z/data=!4m2!3m1!1s0x0:0xdfce9739a948e539?sa=X&ved=2ahUKEwj5n43yxobnAhWUFzQIHaqaAbYQ_BIwDHoECA0QCA" target="_blank">999 3rd Ave Seattle, WA</a>.
+        The entry doors may be locked when you arrive but we’ll have somebody
+        nearby to let you in. Once inside, please take the elevator to the 33rd floor.
+    code: |
```
cnunciato: I don't see this block in the layout, so perhaps it can be removed?

content/events/one-pager.md
```diff
@@ -0,0 +1,92 @@
...
+
+          // Expose the Deployment using a load balanced Kubernetes Service.
+          const service = deployment.createService({
+              type: kx.types.ServiceType.LoadBalancer,
+          });
+    cta:
+        url: https://www.meetup.com/Seattle-Kubernetes-Meetup/events/267073230/
+        label: Register Here
+
+carousel:
```
cnunciato: It seems a shame to duplicate all of this. Could we re-use the existing component somehow?

layouts/page/event-one-pager.html
```diff
@@ -0,0 +1,30 @@
...
+            <div class="header-hero-item">
+                <h1>{{ .Params.hero.title }}</h1>
+                <div class="text-base md:my-5">{{ .Params.hero.body | markdownify }}</div>
+                <div class="header-hero-actions w-full md:w-1/3">
+                    <a class="btn btn-lg" href="{{ .Params.hero.cta.url }}">
+                        {{ .Params.hero.cta.label }}
+                    </a>
+                </div>
+            </div>
+            <!-- <div class="hidden md:inline-block header-hero-item">
```
cnunciato: Ah, I see. We should probably just remove this if we aren't going to use it.

layouts/partials/head.html
```diff
@@ -8,6 +8,10 @@
         <meta http-equiv="refresh" content="0; url={{ .Params.redirect_to }}" />
     {{ end }}
 
+    {{ if .Params.noindex }}
+        <meta name="robots" content="noindex">
```
justinvp: So, I've already used `noindex` elsewhere to mean "exclude this page in our (lame) Pulumi search index". This is repurposing `noindex` to mean `<meta name="robots" content="noindex">`.

I'd be fine with renaming existing uses of `noindex` to something that better explains what it means, e.g. `exclude_from_pulumi_search_index`, and repurposing `noindex` for `<meta name="robots" content="noindex">`.

content/events/one-pager.md
isaac-pulumi: I believe that we're attributing traffic and reporting based on utm codes used in campaigns.

zchase: The URL is general because this specific page is going to linked to from a QR code on our printed banner and slide deck. So we want to maintain the ability to change the page content without having to reprint the QR code.

layouts/page/event-one-pager.html
zchase: Yup good catch.

content/events/one-pager.md
zchase: Yup, it is unused. Good catch

layouts/partials/head.html
zchase: Ah, that is good to know. I renamed both of the variables to be more specific and added descriptions on how/when to use them in the contributing doc.

content/events/one-pager.md
zchase: I can't find a way to reuse the YAML definition from the homepage content front matter. We reuse the HTML for it but can't reuse the content definitions. The first solution I can think is to write some sort of script that parses the front matter for something like `includeContent: ["carousel"]`, then rewrite the front matter to include the given content.

content/docs/_index.md
```diff
@@ -2,7 +2,8 @@
 title: Documentation
 linktitle: Docs
 meta_desc: Learn how to create, deploy, and manage infrastructure on any cloud using Pulumi's open source infrastructure as code SDK.
-noindex: true
+exclude_from_pulumi_search_index: true
+block_external_search_index: true
```
justinvp: Let’s remove `block_external_search_index` here. I think it’s ok for Google et. al. to index this page. But keep `exclude_from_pulumi_search_index` as this isn’t really that relevant of a page when searching within the site.

zchase: Fixed

content/events/one-pager.md
cnunciato: Ah, okay -- that makes sense. Thanks for the clarification.

cnunciato: It does seem like maybe this URL could be slightly less detached from anything concrete, though. Maybe I'm overthinking it, but I can imagine, say, having two banners or something -- do you want to know which one was interacted with? 

If not, just tell me to shut up. It's just bugging me but again, I'm probably overthinking it.

cnunciato: It's also sort of bugging me that the name is /events/one-pager, and the layout is /page/event-one-pager, but the url is /show. Are there going to be "event-two-pager"s? The naming just isn't lining up for me.

Like, if everything were named "event" -- /event, /event.md, /event.html -- that'd make sense. Or even /show (although event seems much clearer). Do you see what I mean? I'm just not understanding why these names are what they are.

content/events/one-pager.md
cnunciato: Zack and I chatted out of band and I'm going to circle back after this and submit a PR that pulls the YAML into the partial, since there's no need to keep them separate. (We did this originally just to avoid git-stomping on one another while finishing up the new home page.)

content/events/one-pager.md
```diff
@@ -0,0 +1,77 @@
+---
+title: Modern Infrastructure as Code
+type: page
+layout: event-one-pager
```
cnunciato: We chatted here also and agreed we'd just name these all `show` for now, just for consistency, and then going forward consider how we'll handle multiple "show"s happening at the same time, etc.

content/events/one-pager.md
justinvp: > Like, if everything were named "event" -- /event, /event.md, /event.html -- that'd make sense. Or even /show (although event seems much clearer). Do you see what I mean? I'm just not understanding why these names are what they are.

I agree.

content/events/one-pager.md
justinvp: That sounds good to me.

The alternative would be to define the shared values in `config.toml` (it'd have to be converted from YAML to TOML, or switch from `config.toml` to `config.yaml` and use `{{ $.Param "carousel" }}` to get them where used. Individual pages could then override the values in font matter.

But if we don't need to customize these on individual pages, then I'd just move them into the partials themselves, yeah.

content/events/one-pager.md
justinvp: Ah, missed this comment when I was reviewing a minute ago. Naming all `show` for now SGTM.

PR #2240, opened 2020-01-11
Change the TeamCity post to ensure that it doesn't sound AWS specific

stack72:


stack72:
@justinvp this PR removes the incorrect aliases

PR #2241, opened 2020-01-12
Do an editorial pass over programming model docs

joeduffy:
This change includes an editorial pass over the programming model section and a few entrypoints into them. This is some of our most helpful content but over time it's grown a bit unwieldy. And a lot is JavaScript/TypeScript-specific, reinforcing the notion that we give preferential treatment to them.

For the most part, the changes are intended to simplify the presentation of materials, grammar (including removing passive voice), improve the flow (restructuring of Hx's, including simplifying the right-hand nav, fewer forward references), and embellish and expand upon important concepts (including examples). Additionally, I've tried to rewrite anything language-specific; this includes the introduction of short-codes for various names/links that need to be customized per language.

As part of trying to make languages supported "on equal footing," I noticed we were missing a language page for Go, in addition to Getting Started guides, so I added those too.


joeduffy:
Unless I've overlooked something, this is ready for review now.

content/docs/intro/concepts/programming-model.md
```diff
@@ -522,172 +1003,333 @@ var vpc = new MyVpcComponent("vpc", new ResourceOptions
...
+                InstanceType = "t2.micro",
+                SecurityGroups = { group.Name }
+            });
+        });
+    }
+}
+```
 
-The name specified during resource creation is used in two key ways:
+Importantly, Pulumi will never attempt to modify the Security Group in this example. It simply reads back its state from your currently configured cloud account, and then uses it as inputs to the EC2 Instance.
```
cnunciato:
```suggestion
Importantly, Pulumi will never attempt to modify the Security Group in this example. It simply reads back its state from your currently configured cloud account, and then uses it as input for the EC2 Instance.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -706,93 +1348,106 @@ url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error)
...
 ```csharp
-// Helpers for accessing properties are not yet available in .NET.
+// In .NET 'Output.Tuple' is used so that each unwrapped value will preserve their distinct type.
+// 'Output.All' can be used when all input values have the same type (i.e. all are Output<string>)
+var connectionString = Output.Tuple(sqlServer.name, database.name)
+    .Apply(t => `Server=tcp:${t.Item1}.database.windows.net;initial catalog=${t.Item2}...`);
 ```
 
-To make this kind of property and array element access more simple, an `Output` _lifts_ the properties of the value that is wrapped, allowing you to access them directly off of the `Output` itself.  This allows the previous example to be written more simply as:
+Notice that {{< pulumi-all >}} works by itself returning an output that represents the combination of all of the input outputs, so that within the callback, the raw values are available inside of [a tuple](https://en.wikipedia.org/wiki/Tuple).
```
cnunciato: `input outputs` reads a bit oddly -- maybe something like this?

```suggestion
Notice that {{< pulumi-all >}} works by itself returning an output that represents the combination of multiple outputs, so that within the callback, the raw values are available inside of [a tuple](https://en.wikipedia.org/wiki/Tuple).
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -706,93 +1348,106 @@ url := virtualmachine.DnsName().Apply(func(dnsName string) (interface{}, error)
...
+
+##### Accessing Properties of an Output {#lifting}
+
+If you just need to access a property off of an {{< pulumi-output >}} value, in order to pass in that property's value as ann argument to another resource's constructor, you can often just directly access it.
```
cnunciato:
```suggestion
If you just need to access a property of an {{< pulumi-output >}} value, in order to pass that property's value as an argument to another resource's constructor, you can often just directly access it.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -802,138 +1457,110 @@ record = aws.route53.Record("validation",
 ```
 
 ```csharp
-// Helpers for accessing properties are not yet available in .NET
+// Helpers for accessing properties are not yet available in .NET.
 ```
 
-This helps the clarity of the final code, while not losing any important dependency information that is needed for properly creating and maintaining the stack.
+To ease simple property and array element access, an {{< pulumi-output >}} _lifts_ the properties of the underlying value, behaving very much like an instance of it. This allows you to access properties and elements directly off of the {{< pulumi-output >}} itself without needing {{< pulumi-apply >}}. If we return to the above example, we can now simplify it:
```
cnunciato:
```suggestion
To ease simple property and array element access, an {{< pulumi-output >}} _lifts_ the properties of the underlying value, behaving very much like an instance of it. This allows you to access properties and elements directly from the {{< pulumi-output >}} itself without needing {{< pulumi-apply >}}. If we return to the above example, we can now simplify it:
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -802,138 +1457,110 @@ record = aws.route53.Record("validation",
...
 //
 // See https://github.com/pulumi/pulumi/issues/1614.
 ```
 
 ```csharp
-// In .NET 'Output.Tuple' is used so that each unwrapped value will preserve their distinct type.
-// 'Output.All' can be used when all input values have the same type (i.e. all are Output<string>)
-var connectionString = Output.Tuple(sqlServer.name, database.name)
-                             .Apply(t => `Server=tcp:${t.Item1}.database.windows.net;initial catalog=${t.Item2}...`);
+// Helpers for accessing properties are not yet available in .NET
```
cnunciato:
```suggestion
// Helpers for accessing properties are not yet available in .NET.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -802,138 +1457,110 @@ record = aws.route53.Record("validation",
...
+let certValidation = new aws.route53.Record("cert_validation", {
+  records: [certCertificate.apply(cc => cc ? cc.domainValidationOptions : undefined)
+                           .apply(dvo => dvo ? dvo[0] : undefined)
+                           .apply(o => o ? o.resourceRecordValue : undefined)],
 ```
 
-##### Working with Outputs and strings {#outputs-and-strings}
+##### Working with Outputs and Strings {#outputs-and-strings}
+
+Outputs containing strings cannot be used directly in operations like string concatenation. _String interpolation_ lets you more easily build a string out of various output values, without needing {{< pulumi-apply >}} or {{< pulumi-all >}}. This can be used to export a stack output, provide a dynamically computed string as a new resource argument, or even just for diagnostics purposes.
```
cnunciato:
```suggestion
Outputs containing strings cannot be used directly in operations like string concatenation. _String interpolation_ lets you more easily build a string out of various output values, without needing {{< pulumi-apply >}} or {{< pulumi-all >}}. This can be used to export a stack output, provide a dynamically computed string as a new resource argument, or even just for diagnostic purposes.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -1135,29 +1779,29 @@ var param = new Aws.Ssm.Parameter("a-secret-param", new Aws.Ssm.ParameterArgs
...
 
-An `apply`'s callback is given the plaintext value of the underlying secret. Although Pulumi ensures that the value returned from an `apply` on a secret is also marked as secret, Pulumi cannot guarantee that the `apply` callback itself will not expose the secret value. For instance, by explicitly printing the value to the console or saving it to a file. Be careful that you do not pass this plaintext value to code that might intentionally or accidentally expose it.
+An {{< pulumi-apply >}}'s callback is given the plaintext value of the underlying secret. Although Pulumi ensures that the value returned from an {{< pulumi-apply >}} on a secret is also marked as secret, Pulumi cannot guarantee that the {{< pulumi-apply >}} callback itself will not expose the secret value. For instance, by explicitly printing the value to the console or saving it to a file. Be careful that you do not pass this plaintext value to code that might intentionally or accidentally expose it.
```
cnunciato:
```suggestion
An {{< pulumi-apply >}}'s callback is given the plaintext value of the underlying secret. Although Pulumi ensures that the value returned from an {{< pulumi-apply >}} on a secret is also marked as secret, Pulumi cannot guarantee that the {{< pulumi-apply >}} callback itself will not expose the secret value---for instance, by explicitly printing the value to the console or saving it to a file. Be careful that you do not pass this plaintext value to code that might intentionally or accidentally expose it.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -1135,29 +1779,29 @@ var param = new Aws.Ssm.Parameter("a-secret-param", new Aws.Ssm.ParameterArgs
...
+#### Explicitly Marking Resource Outputs as Secrets
 
 It is possible to mark resource outputs as containg secrets. In this case, Pulumi will automatically treat those outputs
 as secrets and encrypt them in the state file and anywhere they flow to. To do so,
 [use the "additional secret outputs" option, as described above]({{< relref "#additionalsecretoutputs" >}}).
 
-## Stack outputs {#stack-outputs}
+### Stack Outputs {#stack-outputs}
+
+A stack may export values as [stack outputs]({{< relref "stack#outputs" >}}). These outputs are shown during an update, can be easily retrieved from the Pulumi CLI, and are displayed on pulumi.com. They can be used for important values like resource IDs and computed IP addresses and DNS names. They can also be used for [inter-stack dependencies](#stack-references), such as when a lower layer of infrastructure needs to export values for consumption elsewhere.
```
cnunciato:
```suggestion
A stack may export values as [stack outputs]({{< relref "stack#outputs" >}}). These outputs are shown during an update, can be easily retrieved from the Pulumi CLI, and are displayed in the Pulumi Console. They can be used for important values like resource IDs and computed IP addresses and DNS names. They can also be used for [inter-stack dependencies](#stack-references), such as when a lower layer of infrastructure needs to export values for consumption elsewhere.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -1289,49 +1941,75 @@ other_output = other.get_output("x")
...
 ```csharp
 var config = new Pulumi.Config();
 var name = config.Require("name");
-Console.WriteLine($"Hello, {name}!");
+var lucky = config.GetInt32("lucky") ?? 42;
+Console.WriteLine($"Hello, {name} -- I see your lucky number is {lucky}!");
 ```
 
-Configuration values can be retrieved using [config.get] or [config.require].  Using `get` will return `undefined` if the configuration value was not provided, and `require` will raise an exception with a helpful error message to prevent the deployment from continuing.
+In this example, we have created an instance of the {{< pulumi-config >}} class, which is a bag of key/variable pairs. On that instance, a number of getter functions allow us to read the currently-set values.
```
cnunciato:
```suggestion
In this example, we have created an instance of the {{< pulumi-config >}} class, which is a bag of key/value pairs. On that instance, a number of getter functions allow us to read the currently-set values.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -1390,159 +2068,29 @@ var data = config.RequireObject<JsonElement>("data");
...
-{
-    { "bucketDnsName", bucket.BucketDomainName }
-});
-```
+Any packages that create custom resources---classes that derive from the {{< pulumi-customresource >}} base class---will cause Pulumi to load an associated "resource provider plugin" at runtime, which is a binary that implements the Create, Read, Update, and Delete resources defined by the package. Normally plugins are installed automatically when you install the package, but you can also [manage plugins explicitly using the CLI]({{< relref "/docs/reference/cli/pulumi_plugin" >}}).
 
-In addition to the usual resource options, components accept a set of [providers](#providers) to use for their child resources. If a component is itself a child of another component, its set of providers is inherited from its parent by default.
+This is in contrast to a component resource---classes that serive from the {{< pulumi-componentresource >}} base class--whose logic is written entirely within that library itself, without any external plugin required. A component resource does not manage any external infrastructure state; instead, it simply aggregates existing resources into a larger abstraction.
```
cnunciato:
```suggestion
This is in contrast to a component resource---classes that derive from the {{< pulumi-componentresource >}} base class--whose logic is written entirely within that library itself, without any external plugin required. A component resource does not manage any external infrastructure state; instead, it simply aggregates existing resources into a larger abstraction.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -1799,34 +2345,33 @@ class Program
 }
 ```
 
-## Dynamic Providers <span class="badge badge-preview">PREVIEW</span> {#dynamicproviders}
+#### Dynamic Providers <span class="badge badge-preview">PREVIEW</span> {#dynamicproviders}
 
-Every `CustomResource` has a provider associated with it which knows how to `create`, `read`, `update`, and `delete` instances of the custom resource in the backing cloud provider.  This provider can be defined by implementing the Pulumi Resource Provider gRPC interface.  There are generally two approaches to implementing this provider interface:
+Every {{< pulumi-customresource >}} has a provider associated with it which knows how to `create`, `read`, `update`, and `delete` instances of the custom resource in the backing cloud provider. This provider plugin is defined by implementing the Pulumi Resource Provider gRPC interface. Most resources implement this without you need to know how.
```
cnunciato:
```suggestion
Every {{< pulumi-customresource >}} has a provider associated with it which knows how to `create`, `read`, `update`, and `delete` instances of the custom resource in the backing cloud provider. This provider plugin is defined by implementing the Pulumi Resource Provider gRPC interface. Most resources implement this without you needing to know how.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -1799,34 +2345,33 @@ class Program
...
 
-1. Create a provider binary with the appropriate name and put it on the path, such that it will be loaded to handle CRUD operations from the Pulumi engine on resources from the package it is defined to handle.  For example, the `pulumi-resource-aws` binary will handle resources from the `aws` package.  This binary can be authored in any language, but must be authored and distributed out of band of a Pulumi program.
-2. Define an implementation of the Provider interface directly within your Pulumi program, just for use within that program.
+This plugin model is flexible and allows multi-language consumption of a shared set of resource providers. It does mean, however, that you need to implement a gRPC interface, and create and distribute a plugin. In some cases, it's useful to more rapidly create new resource types. For instance, if a feature is missing from your favorite provider, you'd like to create a resource model over some REST APIs, or you need to implement some custom CRUD logic.
```
cnunciato:
```suggestion
This plugin model is flexible and allows multi-language consumption of a shared set of resource providers. It does mean, however, that you need to implement a gRPC interface, and create and distribute a plugin. In some cases, it's useful to more rapidly create new resource types---for instance, if a feature is missing from your favorite provider, you'd like to create a resource model over some REST APIs, or you need to implement some custom CRUD logic.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -1799,34 +2345,33 @@ class Program
...
 
-The former is used for most common Pulumi providers like AWS and Kubernetes.  The latter is a concept called _Dynamic Providers_, which provide a flexible approach to defining custom resource types directly within the source code of your Pulumi program.
+Pulumi supports a concept called **dynamic providers** for this. Dynamic providers provide a flexible approach to defining custom resource types directly within the source code of your Pulumi program. They aren't as easy to share with other programs, and cannot be used across languages, but are as easy to create as implementing an interface.
 
 You should consider implementing a dynamic provider if:
 
-1. You need to manage a cloud resource for which there is not yet a published Pulumi Provider, but you expect to only use it from within one program. If you expect to use it from many programs, and in many languages, implementing a full provider is preferrable.
+1. You need to manage a cloud resource that doesn't have provider support, and you expect to only use it from within one program. If you expect to use it from many programs, and in many languages, implementing a full provider is preferrable.
```
cnunciato:
```suggestion
1. You need to manage a cloud resource that doesn't have provider support, and you expect to only use it from within one program. If you expect to use it from many programs, and in many languages, implementing a full provider is preferable.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -1887,15 +2432,23 @@ class MyResource(Resource):
...
+
+##### How Dynamic Providers Work
+
+Dynamic providers are a flexible and low-level mechanism to plug arbitrary code directly into the deployment process. Whereas most code in a Pulumi program runs as part of constructing the desired state of resources (the "resource graph"), the code inside the dynamic provider resource provider interface implementations, such as `create` or `update`, runs instead during resource provisioning (while the resource graph is being turned into a set of CRUD operations scheduled against the cloud providers).
```
cnunciato: I stumbled over this wording. Is this what you mean?

```suggestion
Dynamic providers are a flexible and low-level mechanism to plug arbitrary code directly into the deployment process. Whereas most code in a Pulumi program runs as part of constructing the desired state of resources (the "resource graph"), the code inside a dynamic provider's interface implementation, such as `create` or `update`, runs instead during resource provisioning (while the resource graph is being turned into a set of CRUD operations scheduled against the cloud providers).
```
```suggestion
Dynamic providers are a flexible and low-level mechanism to plug arbitrary code directly into the deployment process. Whereas most code in a Pulumi program runs as part of constructing the desired state of resources (the "resource graph"), the code inside a dynamic provider's interface implementation, such as `create` or `update`, runs instead during resource provisioning (while the resource graph is being turned into a set of CRUD operations scheduled against the cloud providers).
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -1996,41 +2549,41 @@ class MyResource extends pulumi.dynamic.Resource {
...
+`check` is invoked before any other methods, and is passed the resolved input properties that were originally provided to the resource constructor by the user. It is passed both the old input properties that were stored in the _state file_ after the previous update to the resource, as well as the new inputs from the current deployment. It has two jobs: (1) Verify that the inputs (particularly the news) are valid and return useful error messages if they are not, and (2) Return a set of checked inputs. The inputs returned from the call to `check` will be the inputs that the Pulumi engine uses for all further processing of the resource, including the values that will be passed back in to `diff`, `create`, `update`, or other operations. In many cases, the `news` can be returned directly as the checked inputs. But in cases where the provider need to populate defaults, or do some normalization on values, it may want to do that in the `check` method so that this data is complete and normalized prior to being passed in to other methods.
```
cnunciato:
```suggestion
`check` is invoked before any other methods, and is passed the resolved input properties that were originally provided to the resource constructor by the user. It is passed both the old input properties that were stored in the _state file_ after the previous update to the resource, as well as the new inputs from the current deployment. It has two jobs: (1) Verify that the inputs (particularly the news) are valid and return useful error messages if they are not, and (2) Return a set of checked inputs. The inputs returned from the call to `check` will be the inputs that the Pulumi engine uses for all further processing of the resource, including the values that will be passed back in to `diff`, `create`, `update`, or other operations. In many cases, the `news` can be returned directly as the checked inputs. But in cases where the provider needs to populate defaults, or do some normalization on values, it may want to do that in the `check` method so that this data is complete and normalized prior to being passed in to other methods.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -1996,41 +2549,41 @@ class MyResource extends pulumi.dynamic.Resource {
...
+`create` is invoked when the URN of the resource created by the user is not found in the existing state of the deployment.  The engine passes the provider the checked inputs returned from the call to `check`. The `create` method is expected to do the work in the backing cloud provider to create the requested resource. It then returns two pieces of data: (1) an `id` that can uniquely identify the resource in the backing provider for later lookups, and (2) a set of `outputs` from the backing provider that should be returned to the user code as properties on the {{< pulumi-customresource >}} object, and stored into the _checkpoint_ file. If an error occurs, an exception can be thrown from the `create` method to return this error to the user.
```
cnunciato:
```suggestion
`create` is invoked when the URN of the resource created by the user is not found in the existing state of the deployment.  The engine passes the provider the checked inputs returned from the call to `check`. The `create` method is expected to do the work in the backing cloud provider to create the requested resource. It then returns two pieces of data: (1) an `id` that can uniquely identify the resource in the backing provider for later lookups, and (2) a set of `outputs` from the backing provider that should be returned to the user code as properties on the {{< pulumi-customresource >}} object, and stored in the _checkpoint_ file. If an error occurs, an exception can be thrown from the `create` method to return this error to the user.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -1996,41 +2549,41 @@ class MyResource extends pulumi.dynamic.Resource {
...
-- `deleteBeforeReplace`: `true` if the proposed replacements require deleteing the existing resource before creating the new one.  By default Pulumi will try to create the new resource before deleting the old one to avoid downtime.
+* `changes`: `true` if the provider believes there is a difference between the `olds` and `news` and wants to do an `update` or `replace` to affect this change.
+* `replaces`: An array of property names that have changed that should force a replacement. Returning a non-zero length array here will tell the Pulumi engine to schedule a replacement instead of an update, which might involve downtime, so this should only be used when a `diff` requested by the user cannot be implemented as an in-place update on the backing cloud provider.
+* `stables`: An array of property names that are known to not change between updates. Pulumi will use this information to allow some {{< pulumi-apply >}} calls on {{< pulumi-output >}} to be processed during `previews` because it knows that the values of these will stay the same during an update.
```
cnunciato:
```suggestion
* `stables`: An array of property names that are known not to change between updates. Pulumi will use this information to allow some {{< pulumi-apply >}} calls on {{< pulumi-output >}} to be processed during `previews` because it knows that the values of these will stay the same during an update.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -1996,41 +2549,41 @@ class MyResource extends pulumi.dynamic.Resource {
...
+* `deleteBeforeReplace`: `true` if the proposed replacements require deleteing the existing resource before creating the new one. By default Pulumi will try to create the new resource before deleting the old one to avoid downtime.
```
cnunciato:
```suggestion
* `deleteBeforeReplace`: `true` if the proposed replacements require deleting the existing resource before creating the new one. By default Pulumi will try to create the new resource before deleting the old one to avoid downtime.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -1996,41 +2549,41 @@ class MyResource extends pulumi.dynamic.Resource {
...
 
-#### `read(id, props)`
+###### `read(id, props)`
 
-Read is invoked when the Pulumi engine needs to get data about a resource that is not managed by Pulumi. It is passed the the `id` of the resource as tracked in the backing cloud provider, and an optional bag of additional properties that can be used to disambiguate the request if needed. The `read` method is expected to look up the requested resource, and return the canonical `id` and output properties of this resource if found.  If an error occurs, an exception can be thrown from the `read` method to return this error to the user.
+Read is invoked when the Pulumi engine needs to get data about a resource that is not managed by Pulumi. It is passed the the `id` of the resource as tracked in the backing cloud provider, and an optional bag of additional properties that can be used to disambiguate the request if needed. The `read` method is expected to look up the requested resource, and return the canonical `id` and output properties of this resource if found. If an error occurs, an exception can be thrown from the `read` method to return this error to the user.
```
cnunciato:
```suggestion
Read is invoked when the Pulumi engine needs to get data about a resource that is not managed by Pulumi. It is passed the `id` of the resource as tracked in the backing cloud provider, and an optional bag of additional properties that can be used to disambiguate the request if needed. The `read` method is expected to look up the requested resource, and return the canonical `id` and output properties of this resource if found. If an error occurs, an exception can be thrown from the `read` method to return this error to the user.
```

content/docs/intro/concepts/programming-model.md
```diff
@@ -2302,13 +2859,23 @@ export("label_url", label.url)
...
+#### Getting the Current Project or Stack {#current-project-stack}
 
-## Runtime code {#runtime}
+The {{< pulumi-getproject >}} and {{< pulumi-getstack >}} functions give you the currently deploying project and stack, respectively. This can be useful for naming or tagging resources.
+
+#### Logging
+
+The {{< pulumi-log >}} collection of functions allows you to log diagnostics, warnings, or errors with the Pulumi engine. These will be displayed alongside all other Pulumi output in the CLI and on app.pulumi.com. They will also be logged and kept for historical purposes if you ever want to use them to audit or diagnose what has transpired.
```
cnunciato:
```suggestion
The {{< pulumi-log >}} collection of functions allows you to log diagnostics, warnings, or errors with the Pulumi engine. These will be displayed alongside all other Pulumi output in the CLI and in the Pulumi Console. They will also be logged and kept for historical purposes if you ever want to use them to audit or diagnose what has transpired.
```

PR #2243, opened 2020-01-13
Adding the cloudflare provider to the providers page

stack72:


content/docs/intro/cloud-providers/cloudflare/_index.md
```diff
@@ -0,0 +1,106 @@
...
+    parent: cloud-providers
+    identifier: clouds-cloudflare
+    weight: 5
+---
+
+<img src="/logos/tech/cloudflare.png" align="right" class="h-16 px-8 pb-4">
+
+The Cloudflare provider for Pulumi can be used to provision any of the resources available in [Cloudflare](https://www.cloudflare.com/).
+
+See the [full API documentation]({{< relref "/docs/reference/pkg/nodejs/pulumi/cloudflare" >}}) for complete details of the available Cloudflare provider APIs.
```
justinvp: It occurs to me that we're linking to only the NodeJS docs on these pages. Should we also be linking to Python, since we have those available (and in the future the rest of the language API docs once they're published)?

stack72: @justinvp I will add an issue to circle back on all of these to add the other sets of docs in M32

I just wanna get the first pass done

PR #2253, opened 2020-01-15
Fix broken link in dotnet preview blog

zchase:
### Proposed changes

This fixes a broken link in the dotnet preview blog. It looks like Microsoft has recently restructured some of their documentation urls. Please double check that it is indeed the right place to linking to.

Old link: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-overview#resource-groups
New Link: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/overview#resource-groups


justinvp:
> Please double check that it is indeed the right place to linking to.

It is. In fact, the old URL is now 301 redirecting to that new URL.

PR #2261, opened 2020-01-16
match release dates to CHANGELOG

dstine:
These two relatively recent releases have different dates in the CHANGELOG and in the Available Versions page. This PR fixes that, assuming that the CHANGELOG is the source of truth.


stack72:
Thanks so much for this @dstine

PR #2269, opened 2020-01-18
Adding fastly to the cloud providers page

stack72:


content/docs/intro/cloud-providers/fastly/_index.md
```diff
@@ -0,0 +1,138 @@
+---
+title: Fastly
+meta_desc: This page provides an overview of the Fastly Provider for Pulumi.
+menu:
+  intro:
+    parent: cloud-providers
+    identifier: clouds-fastly
+    weight: 5
+
+aliases: ["/docs/reference/clouds/fastly/"]
```
justinvp: Remove.

@stack72, in general, any **_new_** page you add should not have any aliases. We generally only need to use an alias when moving an existing page to a different URL path within the site.

content/docs/intro/cloud-providers/fastly/_index.md
```diff
@@ -0,0 +1,138 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Pulumi;
+using Pulumi.Fastly;
+
+class Program
+{
+    static Task Main() =>
+        Deployment.Run(() => {
+            var service = new Fastly.Servicev1("test", new Fastly.Servicev1Args
```
justinvp: I haven't tried it out, so could be wrong, but just from reading it, should this be:

```suggestion
            var service = new Servicev1("test", new Servicev1Args
```

?

content/docs/intro/cloud-providers/fastly/setup.md
```diff
@@ -0,0 +1,32 @@
+---
+title: Fastly Setup
+meta_desc: This page provides an overview on how to configure credentials for the Pulumi Fastly Provider.
+aliases: ["/docs/reference/clouds/fastly/setup/"]
```
justinvp: Remove

content/docs/intro/cloud-providers/fastly/setup.md
```diff
@@ -0,0 +1,32 @@
...
+    $ export FASTLY_API_KEY=XXXXXXXXXXXXXX
+    ```
+
+2. Set them using configuration, if you prefer that they be stored alongside your Pulumi stack for easy multi-user access:
+
+    ```bash
+    $ pulumi config set fastly:apiKey XXXXXXXXXXXXXX --secret
+    ```
+
+Remember to pass `--secret` when setting `token` so that it is properly encrypted. A full set of configuration parameters
```
justinvp:
```suggestion
Remember to pass `--secret` when setting `fastly:apiKey` so that it is properly encrypted. A full set of configuration parameters
```

content/docs/intro/cloud-providers/fastly/_index.md
stack72: You are correct - and confirmed by @mikhailshilkov

PR #2280, opened 2020-01-22
Add language-batch, -bat, -powershell to list of shell languages

joeduffy:
This change adds language-batch to the list of shell languages,
which are meant to be skipped when applying the language chooser.


joeduffy:
@justinvp something like this is probably "right" given the intention of the current JavaScript, but I don't know what sort of secondary effects this could have. So until we find that it's required -- or until you are comfortable because you understand the potential impact -- I will just close out.

Issue #2223, opened 2020-01-07
Provide way for readers to provide feedback/suggestions


chrsmith:
Something we've heard from Pulumi users/customers is that our documentation is very important for their success, and at times isn't quite what is needed. (In addition to the "Edit on GitHub" link.)

We should provide a way for readers to push feedback to us to highlight any content issues and/or provide suggestions for how to improve various pages.

Looking at Kubernetes.io, you can see a "Feedback" widget at the bottoms of pages. This looks like it just sends data to Google Tags Manager.
- https://kubernetes.io/docs/reference/kubectl/overview/
- [Hugo template](https://github.com/kubernetes/website/blob/master/layouts/partials/feedback.html)
- [Hugo template continued](https://github.com/kubernetes/website/blob/88d0f7825190694ed21be2229839cef9e75f14cf/i18n/en.toml#L107)

@zchase could you look into what the best tool for this sort of thing is? e.g. is Google Tags Manager the right tool here? Is there a capability of HubSpot/Segment that we can use to track 👍 / 👎  ratings? Do we need to roll our own?

Once we have an idea for how to go about this, I'm happy to do the work.

zchase:
@chrsmith Yup, we would use Segment and Google Analytics to track this. All we need to do is fire an event when the button is clicked, with the code looking something like this:
```
window.analytics.track(
    category: "User Documentation | Helpful",
    label: "Yes",
);
```

Then in Google Analytics we can bucket the events by page and see how many people clicked each of the button. The `category` above can be named whatever we want, as long as we can identify these events are for user feedback from the docs.

justinvp:
@zchase, @chrsmith Is it possible to include an optional form field so users can type in a message on why they're clicking thumbs up or down?

zchase:
@justinvp Yes and no. We can but we will have to rethink how we will store the feedback. We won't be able to store longer feedback in Google Analytics, it would be incredibly difficult to maintain this data in HubSpot, so I think we would have to store it in a table.

justinvp:
@zchase, ok, I was just curious. I don't think we need the form field... We could just do something similar to the Kubernetes site that Chris linked to. When I click "No" it provides links to report a problem or suggest an improvement (by filing GitHub issues). A similar approach should work for us.

<img width="716" alt="Screen Shot 2020-01-09 at 9 30 14 AM" src="https://user-images.githubusercontent.com/710598/72076049-e7124580-32c2-11ea-8811-5a28a54ab79f.png">

[...]

Issue #1900, opened 2019-10-28
Consider making "On This Page" always visible


justinvp:
When reading longer pages like https://www.pulumi.com/docs/intro/concepts/programming-model/, it's easy to lose track of where you are on the page.

To help, we should consider "anchoring" the right-nav "ON THIS PAGE" so it's always visible (similar to the [Tailwind CSS docs](https://tailwindcss.com/docs/installation)) and bold the current section of the page (similar to [Docker's docs](https://docs.docker.com)).

joeduffy:
@justinvp Ironically, I was just about to file this same issue... for the programming model page. This seems like a pretty minor change that would have a potentially large improvement to UX. Should we try to tackle this soon?

Note that we may want to consider the same for the left-hand nav, though it's probably less urgent.

justinvp:
> Note that we may want to consider the same for the left-hand nav, though it's probably less urgent.

I kind of like how the Tailwind docs handle this: The left nav is scrollable and the right nav is sticky and scrollable. If that sounds good:

- [ ] Right-nav sticky
- [ ] Right-nav vertically scrollable
- [ ] Right-nav current section bolded as you scroll through the page
- [ ] Left-nav vertically scrollable

@cnunciato, here's one worth considering this milestone.

PR #2284, opened 2020-01-22
Add a way for users to provide feedback on a docs page.

zchase:
# Proposed changes

This PR adds a section to the end of our documentation pages that allows for a user to provide feedback on that specific document. The data is sent to Segment and then stored in Redshift which is connected to Metabase. The data will also be viewable in Google Analytics, but to see specific comments we will have to generate a dashboard in Metabase.

feedbackBlock

The user is able to select if the document is helpful or not and then they are shown an additional comment section.

feedbackComment

The form fields are not required. Once the submits the form they are shown this thank you section.

feedbackThankYou

If the user does not submit the form but does press either the "Yes" or "No" button, that answer is sent when the user navigates away from the page. 

@cnunciato Let me know if there are any UX/design things I can improve

### Related issues (optional)

https://github.com/pulumi/docs/issues/2223

### DIFF
--- a/assets/js/docs-feedback.js
+++ b/assets/js/docs-feedback.js
@@ -12,7 +12,6 @@ $(function(){
         ["#docsFeedbackYes", "#docsFeedbackNo"].forEach(function(key) {
             var answer = key === "#docsFeedbackYes" ? "Yes" : "No";
             $(key).on("click", function() {
-                $("#feedbackButtons").addClass("hidden");
                 $("#feedbackLongForm").removeClass("hidden");
 
                 // Show the additional comment section.
@@ -56,6 +55,7 @@ $(function(){
                 sendFeedbackToSegement(answer, comments, email);
 
                 // Show the thank you section.
+                $("#feedbackButtons").addClass("hidden");
                 $("#feedbackLongForm").addClass("hidden");
                 $("#feedbackThankYou").removeClass("hidden");
             });
--- a/assets/js/docs-feedback.js
+++ b/assets/js/docs-feedback.js
@@ -60,11 +60,21 @@ $(function(){
                 $("#feedbackThankYou").removeClass("hidden");
             });
 
+            $("#docsCloseFeedbackLongForm").on("click", function() {
+                // Send to Segment.
+                sendFeedbackToSegement(answer, "", "");
+
+                // Show the thank you section.
+                $("#feedbackButtons").addClass("hidden");
+                $("#feedbackLongForm").addClass("hidden");
+                $("#feedbackThankYou").removeClass("hidden");
+            });
+
             $(window).on("beforeunload", function() {
                 // When page unloads send the answer if it has not already been sent.
                 var feedbackSent = $("#feedbackLongForm").hasClass("hidden");
                 if (!feedbackSent) {
-                    sendFeedbackToSegement(answer);
+                    sendFeedbackToSegement(answer, "", "");
                 }
             });
         }
--- a/assets/js/docs-feedback.js
+++ b/assets/js/docs-feedback.js
@@ -1,0 +1,75 @@
+/**
+ *
+ */
+$(document).ready(function(){
+    // Check that the analytics track function is available.
+    var analyticsAvailable = window.analytics && window.analytics.track &&
+                             (typeof window.analytics.track === "function");
+
+    if (analyticsAvailable) {
+        // Show the docs feedback container.
+        $("#docsFeedbackContainer").removeClass("hidden");
+
+        // For each of the button ids attach a click hanlder that shows
+        // the additional comment section.
+        ["#docsFeedbackYes", "#docsFeedbackNo"].forEach(function(key) {
+            var answer = key === "#docsFeedbackYes" ? "Yes" : "No";
+            $(key).on("click", function() {
+                $("#feedbackButtons").addClass("hidden");
+                $("#feedbackLongForm").removeClass("hidden");
+
+                // Show the additional comment section.
+                showAdditionalCommentSection(answer);
+            });
+        });
+
+        /**
+         * Submit the documentation feedback to Segment.
+         *
+         * @param {string} answer The value of the button the user clicked when giving feedback.
+         * @param {string} comments The value of the text area in the comment section. Defaults to an empty string
+         * @param {string} email The value of the email input in the comment section. Defaults to an empty string
+         */
+        function sendFeedbackToSegement(answer, comments = "", email = "") {
+            var trackingObj = {
+                answer,
+                comments,
+                email,
+                url: window.location.pathname,
+                category: "Documentation Feedback",
+                label: answer,
+            };
+            window.analytics.track("docs-feedback", trackingObj);
+        }
+
+        /**
+         * Show the additional comments section and inject the user's feedback
+         * answer into the submission click handler and form abandonment handler.
+         *
+         * @param {string} answer The value of the button the user clicked when giving feedback.
+         */
+        function showAdditionalCommentSection(answer) {
+            // Add a click handler to the submit button
+            $("#docsSubmitFeedback").on("click", function() {
+                // Grab the values of the comments and email inputs.
+                var comments = $("#feedbackAdditionalComments").val().trim();
+                var email = $("#feedbackEmail").val().trim();
+
+                // Send to Segment.
+                sendFeedbackToSegement(answer, comments, email);
+
+                // Show the thank you section.
+                $("#feedbackLongForm").addClass("hidden");
+                $("#feedbackThankYou").removeClass("hidden");
+            });
+
+            $(window).on("unload", function() {
+                // When page unloads send the answer if it has not already been sent.
+                var shouldSendFeedback = $("#feedbackLongForm").hasClass("hidden");
+                if (!shouldSendFeedback) {
+                    sendFeedbackToSegement(answer);
+                }
+            });
+        }
+    }
+});
--- a/assets/js/docs-feedback.js
+++ b/assets/js/docs-feedback.js
@@ -1,7 +1,4 @@
-/**
- *
- */
-$(document).ready(function(){
+$(function(){
     // Check that the analytics track function is available.
     var analyticsAvailable = window.analytics && window.analytics.track &&
                              (typeof window.analytics.track === "function");
@@ -10,7 +7,7 @@ $(document).ready(function(){
         // Show the docs feedback container.
         $("#docsFeedbackContainer").removeClass("hidden");
 
-        // For each of the button ids attach a click hanlder that shows
+        // For each of the button ids attach a click handler that shows
         // the additional comment section.
         ["#docsFeedbackYes", "#docsFeedbackNo"].forEach(function(key) {
             var answer = key === "#docsFeedbackYes" ? "Yes" : "No";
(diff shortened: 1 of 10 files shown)


lukehoban:
> adds a section to the end of our documentation pages

I wonder if it makes sense to put this in this piece of UX instead - it's prominent, and also somewhat similar in purpose.

Screen Shot 2020-01-22 at 2 47 01 PM

lukehoban:
> http://content-bucket-e18f8f6.s3-website-us-east-1.amazonaws.com

I don't see these changes showing up in the preview anywhere - is that expected?

zchase:
@lukehoban Yes it only shows if `window.analytics.track` is available. Otherwise, the section is hidden because we won't be able to record the feedback. You can run it locally be editing `partials/head.html` to load the Segment script. Though if you do that you will need to provide a write key for Segment that isn't our production key. I have a dev Segment account set up so just ping me on Slack and I can send it over.

assets/js/docs-feedback.js
```diff
@@ -0,0 +1,75 @@
...
+        });
+
+        /**
+         * Submit the documentation feedback to Segment.
+         *
+         * @param {string} answer The value of the button the user clicked when giving feedback.
+         * @param {string} comments The value of the text area in the comment section. Defaults to an empty string
+         * @param {string} email The value of the email input in the comment section. Defaults to an empty string
+         */
+        function sendFeedbackToSegement(answer, comments = "", email = "") {
```
justinvp: Default parameters are an ES2015 feature, which won't work in IE. I don't think we really care much about IE at this point, but it'd still be nice if the site worked (for the most part) there.

As an aside, if we continue to invest in scripting on the site, I'd love to setup a more modern pipeline, so we can use modern language features, TypeScript, webpack, etc.

assets/js/docs-feedback.js
```diff
@@ -0,0 +1,75 @@
...
+            var trackingObj = {
+                answer,
+                comments,
+                email,
```
justinvp: Shorthand property names is a ES2015 feature that isn't supported in IE. Again, we probably don't care, but it's also easy enough to use the older syntax:

```
answer: answer,
comments: comments,
email: email,
```

assets/js/docs-feedback.js
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        /**
+         * Show the additional comments section and inject the user's feedback
+         * answer into the submission click handler and form abandonment handler.
+         *
+         * @param {string} answer The value of the button the user clicked when giving feedback.
+         */
+        function showAdditionalCommentSection(answer) {
+            // Add a click handler to the submit button
```
justinvp: Nit: \*channelling Chris\*
```suggestion
            // Add a click handler to the submit button.
```

assets/js/docs-feedback.js
```diff
@@ -0,0 +1,75 @@
...
+                $("#feedbackLongForm").addClass("hidden");
+                $("#feedbackThankYou").removeClass("hidden");
+            });
+
+            $(window).on("beforeunload", function() {
+                // When page unloads send the answer if it has not already been sent.
+                var shouldSendFeedback = $("#feedbackLongForm").hasClass("hidden");
+                if (!shouldSendFeedback) {
+                    sendFeedbackToSegement(answer);
+                }
```
justinvp: I was confused by the name of the variable here. Should it be something like:

```javascript
var feedbackSent = $("#feedbackLongForm").hasClass("hidden");
if (!feedbackSent) {
    sendFeedbackToSegement(answer);
}
```

?

layouts/partials/docs/feedback.html
```diff
@@ -0,0 +1,33 @@
...
+        </div>
+    </div>
+    <div id="feedbackLongForm" class="hidden">
+        <p>Additional Comments?</p>
+        <textarea id="feedbackAdditionalComments" rows="5" class="mb-3 w-full md:w-3/4 p-2 bg-white rounded border border-gray-500 outline-none"></textarea>
+        <input id="feedbackEmail" type="text" class="block mb-4 w-full md:w-3/4 p-2 rounded border border-gray-500 outline-none" placeholder="Email (optional)" />
+        <a id="docsSubmitFeedback" class="btn md:w-1/4 text-center block">Submit</a>
+    </div>
+    <div id="feedbackThankYou" class="hidden">
+        {{ $problemTitle := printf "Issue with %s" .Permalink}}
```
justinvp: Nit
```suggestion
        {{ $problemTitle := printf "Issue with %s" .Permalink }}
```

layouts/partials/docs/feedback.html
```diff
@@ -0,0 +1,33 @@
...
+        {{ $improvementTitle := printf "Improvement for %s" .Permalink}}
```
justinvp: Nit
```suggestion
        {{ $improvementTitle := printf "Improvement for %s" .Permalink }}
```

layouts/partials/docs/feedback.html
```diff
@@ -0,0 +1,33 @@
...
+            <a id="docsFeedbackNo" data-track="feedback-no" class="btn flex mx-5">
+                <i class="fas fa-thumbs-down mr-2 text-lg"></i>
+                No
+            </a>
+        </div>
+    </div>
+    <div id="feedbackLongForm" class="hidden">
+        <p>Additional Comments?</p>
+        <textarea id="feedbackAdditionalComments" rows="5" class="mb-3 w-full md:w-3/4 p-2 bg-white rounded border border-gray-500 outline-none"></textarea>
+        <input id="feedbackEmail" type="text" class="block mb-4 w-full md:w-3/4 p-2 rounded border border-gray-500 outline-none" placeholder="Email (optional)" />
```
justinvp:
```suggestion
        <input id="feedbackEmail" type="email" class="block mb-4 w-full md:w-3/4 p-2 rounded border border-gray-500 outline-none" placeholder="Email (optional)" />
```

joeduffy:
I agree with Luke that having this on the right side nav would make this more prominent and more likely to be used.

If the page missed the mark, many users won't bother scrolling to the end (especially for some of our very long pages, like API docs).

@justinvp we are also making that sticky/floating as you scroll, right? Soon?

If yes, that probably increases likelihood even more ...

justinvp:
> we are also making that sticky/floating as you scroll, right? Soon?

Yep, https://github.com/pulumi/docs/issues/1900 is assigned to @cnunciato to address this milestone. 😄

zchase:
@justinvp Updated the PR per your feedback.

@joeduffy @lukehoban I pushed an update to move the feedback section to the right-hand side. Also in this update, I have my personal dev Segment key applied, so that we can view the changes in the preview. I'll revert back to the production key once we are ready to merge in.

layouts/partials/head.html
```diff
@@ -104,15 +104,13 @@
     {{ end }}
 
     <script async defer src="https://buttons.github.io/buttons.js"></script>
-    {{- if eq hugo.Environment "production" }}
```
zchase: I removed this logic and added my own site key for the site preview. This will make it easier for people to provide feedback. I will change this back to the production set up when we are ready to merge this in.

layouts/partials/docs/feedback.html
```diff
@@ -0,0 +1,42 @@
+<div id="docsFeedbackContainer" class="mt-8 md:mt-0 hidden">
+    <h4 class="no-anchor text-gray-600 uppercase tracking-wide font-bold text-sm">Feedback</h4>
+    <div id="feedbackButtons">
+        <p class="mt-3 mb-5 text-sm">Was this page helpful?</p>
+        <div class="flex">
+            <a id="docsFeedbackYes" data-track="feedback-yes" class="btn flex">
```
cnunciato: The buttons are too visually prominent, IMO (on part with CTAs) -- lemme see if I can come up with a quick treatment that makes them easily discoverable but not quite so demanding of attention.

joeduffy: I also wonder if they ought to go just above the "ON THIS PAGE" (when smaller), again so they aren't lost -- e.g., on the programming model page, they're still below the fold.

Might also need to adjust a few margins here and there.

layouts/partials/docs/feedback.html
```diff
@@ -0,0 +1,42 @@
...
+        <p>
+            Thank you for your feedback. If you have a specific, answerable question about how
+            to use Pulumi, ask it in our <a href="https://slack.pulumi.com/">Community Slack</a>.
+            Open an issue in the GitHub repo if you want to
+            <a href="https://github.com/pulumi/docs/issues/new?title={{ $problemTitle }}">report a problem</a> or
+            <a href="https://github.com/pulumi/docs/issues/new?title={{ $improvementTitle }}">suggest an improvement</a>.
+        </p>
+    </div>
+    <div id="feedbackLongForm" class="hidden">
+        <div class="popup-overlay">
```
cnunciato: It looks like there's only a SUBMIT button but not a cancel button, and blur/click-outside doesn't seem to dismiss it. We should handle being able to dismiss the dialog without submitting feedback.

It's also unclear when I click a thumb whether the sentiment was captured or whther I have to take the additional step of clicking the SUBMIT button also. I'd expect to be able to just click +1/-1 and then optionally be invited to submit additional feedback after that. Does that make sense?

layouts/partials/scripts.html
```diff
@@ -12,6 +12,7 @@
 {{ $s = $s | append (resources.Get "js/chooser.js") }}
 {{ $s = $s | append (resources.Get "js/noselect.js") }}
 {{ $s = $s | append (resources.Get "js/tracking.js") }}
+{{ $s = $s | append (resources.Get "js/docs-feedback.js") }}
```
cnunciato: +1 to adding something like Stencil.js for things like these going forward. (Happy to look into that myself.)

assets/js/docs-feedback.js
```diff
@@ -0,0 +1,72 @@
...
+                $("#feedbackButtons").addClass("hidden");
+                $("#feedbackLongForm").addClass("hidden");
+                $("#feedbackThankYou").removeClass("hidden");
+            });
+
+            $(window).on("beforeunload", function() {
+                // When page unloads send the answer if it has not already been sent.
+                var feedbackSent = $("#feedbackLongForm").hasClass("hidden");
+                if (!feedbackSent) {
+                    sendFeedbackToSegement(answer);
```
cnunciato:
```suggestion
                    sendFeedbackToSegment(answer);
```

assets/js/docs-feedback.js
```diff
@@ -0,0 +1,72 @@
...
+
+    if (analyticsAvailable) {
+        // Show the docs feedback container.
+        $("#docsFeedbackContainer").removeClass("hidden");
+
+        // For each of the button ids attach a click handler that shows
+        // the additional comment section.
+        ["#docsFeedbackYes", "#docsFeedbackNo"].forEach(function(key) {
+            var answer = key === "#docsFeedbackYes" ? "Yes" : "No";
+            $(key).on("click", function() {
```
cnunciato: Noticed after submitting that when I clicked a thumb again on the same page, the fields were still there. We should probably clear the fields on successive clicks to open this form (and probably clear them on successful submission as well) -- not a huge deal of course, since users will be navigating around, but still, keeps things just a smidge tidier that way.

layouts/partials/docs/feedback.html
```diff
@@ -0,0 +1,51 @@
...
+        {{ $improvementTitle := printf "Improvement for %s" .Permalink }}
+        <p>
+            Thank you for your feedback!
+        </p>
+        <p>
+            If you have a specific, answerable question about how
+            to use Pulumi, ask it in our <a href="https://slack.pulumi.com/">Community Slack</a>.
+        </p>
+        <p>
+            Open an issue on Github to
```
justinvp:
```suggestion
            Open an issue on GitHub to
```

PR #2290, opened 2020-01-23
Declare assets as dependencies of web pages

### LINKED ISSUE #1687: Functionality on the site that relies on scripts can break during deploys
We just saw aspects of the site that rely on scripts become broken during a deployment. In this case, the language and platform chooser tabs stopped working altogether.

Here's what happened:
 - The particular deploy updated some JS scripts, so a new `/js/bundle.min.<hash>.js` was generated with a unique hash.
 - All pages were updated to refer to the new bundle script file.
 - During the deploy, many of the updated pages were uploaded to S3 _before_ the new `/js/bundle.min.<hash>.js` was deployed.
 - Those pages that relied on scripts in the bundle were broken (lang/platform chooser tabs).
 - The deploy took > 30 min because all pages were being updated, so there was a rather large window when things could be broken.

Suggestions:

- One way to improve this would be to ensure files that other pages refer to get uploaded to S3 before any of the pages (namely `/js/bundle.min.<hash>.js` and `/css/styles.<hash>.css`, though there may be others).

- We could also look into reducing the amount of time it takes to deploy the site (reducing the window where things could be broken), potentially using an approach like Pulumify.

 - A more robust solution would be to do more of a blue/green approach to deploying the site. It'd also serve as a good example and is something we could blog about.

cnunciato:
This change adds a bit of logic to the infrastructure app to declare assets (like CSS and JS files and others) as dependencies on HTML files, so the site doesn't break momentarily during deployment.

Fixes #1687.

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -264,17 +264,24 @@ const assetTypes = [
 
 // Keep track of the asset paths we encounter and the resources we create from them, so we
 // can declare the necessary dependency relationships.
-const assetPaths: string[] = [];
+const assetPaths = new Set<string>();
 const assetResources: aws.s3.BucketObject[] = [];
+
+// Find all files by pattern (e.g., "/path/to/begin/from/**/*.jpg"), matching those with
+// dot-prefixed folders or filenames (".net", for example) and excluding directories.
+// https://github.com/isaacs/node-glob#options
+function filePathsByPattern(pattern: string): string[] {
+    return glob.sync(pattern, { dot: true, nodir: true });
+}
 
 // Find all files matching the patterns above, creating a new BucketObject for each one.
 // These objects are declared first in order to pass them conditionally as dependencies
 // below.
 assetTypes.forEach(pattern => {
-    glob.sync(`${webContentsRootPath}/**/*.${pattern}`).forEach(assetPath => {
+    filePathsByPattern(`${webContentsRootPath}/**/*.${pattern}`).forEach(assetPath => {
 
         // Remember the asset path, so we can ignore it later.
-        assetPaths.push(assetPath);
+        assetPaths.add(assetPath);
 
         // Turn the full (local) path into a relative path, so we can use it for an S3 object key.
         const relativeAssetPath = assetPath.replace(webContentsRootPath + "/", "");
@@ -300,15 +307,10 @@ assetTypes.forEach(pattern => {
 });
 
 // Now create bucket objects for everything else, bypassing anything created already.
-glob.sync(`${webContentsRootPath}/**/*`).forEach(filePath => {
-
-    // Ignore directories.
-    if (fs.lstatSync(filePath).isDirectory()) {
-        return;
-    }
+filePathsByPattern(`${webContentsRootPath}/**/*`).forEach(filePath => {
 
     // Ignore previously encountered asset paths, since they'll already have been declared.
-    if (assetPaths.includes(filePath)) {
+    if (assetPaths.has(filePath)) {
         return;
     }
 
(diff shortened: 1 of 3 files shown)


cnunciato:
Looks from the preview like my logic missed a couple of files that have been marked for deletion — I’ll have a look and get a fix pushed up.

infrastructure/index.ts
```diff
@@ -257,47 +242,121 @@ function getMimeType(filePath: string): string | undefined {
...
+// Now create bucket objects for everything else, bypassing anything created already.
+glob.sync(`${webContentsRootPath}/**/*`).forEach(filePath => {
+
+    // Ignore directories.
+    if (fs.lstatSync(filePath).isDirectory()) {
+        return;
+    }
+
+    // Ignore previously encountered asset paths, since they'll already have been declared.
+    if (assetPaths.includes(filePath)) {
```
justinvp: I don't know if it'd make a measurable improvement, but I wonder if it'd be more performant to store these in a `Set<string>` for speeder lookups, instead of iterating over the (potentially entire) array doing string comparisons for each file.

```typescript
const assetPaths = new Set<string>();
```

```typescript
// Remember the asset path, so we can ignore it later.
assetPaths.add(assetPath);
```

```typescript
// Ignore previously encountered asset paths, since they'll already have been declared.
if (assetPaths.has(filePath)) {
    return;
}
```

infrastructure/index.ts
```diff
@@ -257,47 +242,121 @@ function getMimeType(filePath: string): string | undefined {
...
+
+    // Create a new S3 object for most files, but HTML files that contain a redirect
+    // just create a stubbed out S3 object with the right metadata so they return a
+    // 301 redirect (instead of serving the HTML with a meta-redirect). This ensures
+    // the right HTTP code response is returned for search engines, as well as
+    // enables better support for URL anchors.
+    const redirect = getMetaRefreshRedirect(filePath);
+    if (!redirect) {
+        const contentType = getMimeType(filePath) || undefined;
+        console.log(contentType);
```
justinvp: Nit: I assume this was leftover from debugging that you meant to remove?

infrastructure/index.ts
```diff
@@ -257,47 +242,121 @@ function getMimeType(filePath: string): string | undefined {
...
     });
+});
+
+// Now create bucket objects for everything else, bypassing anything created already.
+glob.sync(`${webContentsRootPath}/**/*`).forEach(filePath => {
+
+    // Ignore directories.
+    if (fs.lstatSync(filePath).isDirectory()) {
+        return;
+    }
```
justinvp: Instead of this check, wondering if we should let `glob` handle it via the `nodir` option, e.g.:

```typescript
glob.sync(`${webContentsRootPath}/**/*`, { nodir: true }).forEach(filePath => {
```

https://github.com/isaacs/node-glob#options

> * `nodir` Do not match directories, only files. (Note: to match only directories, simply put a `/` at the end of the pattern.)

infrastructure/package.json
```diff
@@ -11,6 +11,8 @@
     "dependencies": {
         "@pulumi/aws": "^0.18.27",
         "@pulumi/pulumi": "^0.17.28",
+        "@types/glob": "^7.1.1",
```
justinvp: Nit: This could be a dev dependency.

infrastructure/index.ts
cnunciato: Ah, good suggestion!

infrastructure/index.ts
cnunciato: That works well -- and there's a `dot` boolean that handles the `.net` matching issue as well. 🎉

PR #2302, opened 2020-01-25
Revert website-file dependency commits

cnunciato:
😢 

These are being reverted because they introduced additional latency that pushed us beyond the two-hour Travis job limit.


justinvp:
The two deletions in the Preview were pending deletes that I removed from the `pending_operations`, so those are expected. And https://travis-ci.com/pulumi/docs/builds/146031126 just took care of deleting them.

PR #2310, opened 2020-01-27
Deploy pulumi.com with s3 sync

### LINKED ISSUE #1687: Functionality on the site that relies on scripts can break during deploys

### LINKED ISSUE #2309: Update the website with s3 sync
In order to reduce the amount of time it takes to deploy pulumi.com (which is currently taking up to an hour, largely because the site itself contains several thousand files), we should adopt an approach more like the one used by Pulumify, which zips up the build and uploads it in one go, unpacks it on the receiving end, and synchronizes within the AWS network using `aws s3 sync`.

cnunciato:
This PR deploys pulumi.com using an approach more like the one we use with Pulumify, which `tar`s up the website and syncs it with `aws s3 sync`, rather than have Pulumi manage files individually. Drops the time to deploy from roughly an hour down to a few minutes.

Also adds a bit of logic to upload bundle assets (e.g., CSS and JS files) before HTML files, to prevent the site from breaking during an update.

Note that as currently written this change will _delete_ the website momentarily (because the code defining file resources has been removed 😱) before the process to upload, unpack and sync its replacement contents is initiated.

Fixes #2309.
Fixes #1687.

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -47,8 +47,9 @@ const contentBucket = new aws.s3.Bucket(
     },
 );
 
+// Create a bucket for handling uploads of website archive files.
 const cluster = awsx.ecs.Cluster.getDefault();
-const archiveBucket = new aws.s3.Bucket("archive-bucket", { forceDestroy: true });
+const archiveBucket = new aws.s3.Bucket("archive-bucket");
 const archiveHandler = new awsx.ecs.FargateTaskDefinition("archiveHandler", {
     container: {
         image: awsx.ecs.Image.fromPath("archiveHandlerImage", "./"),
@@ -57,6 +58,8 @@ const archiveHandler = new awsx.ecs.FargateTaskDefinition("archiveHandler", {
     },
 });
 
+// Handle archive uploads by running container task that downloads and unpacks the file
+// and synchronizes its contents with S3.
 archiveBucket.onObjectCreated("onArchiveUploaded", new aws.lambda.CallbackFunction<aws.s3.BucketEvent, void>("onUploadHandler", {
     runtime: "nodejs10.x",
     policies: [
@@ -301,9 +304,10 @@ function translateRedirect(filePath: string, redirect: string): string {
     throw new Error(`The redirect "${redirect}" in "${filePath}" is not in an expected format.`);
 }
 
+// Find all Hugo-generated meta-refresh redirects and convert them to proper
+// (Pulumi-managed) AWS web-page redirects.
+// https://docs.aws.amazon.com/AmazonS3/latest/dev/how-to-page-redirect.html
 const redirectPaths = new Set<string>();
-
-// Convert
 glob.sync(`${webContentsRootPath}/**/*.html`).map(filePath => {
     const relativeFilePath = filePath.replace(webContentsRootPath + "/", "");
     const redirect = getMetaRefreshRedirect(filePath);
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -47,9 +47,10 @@ const contentBucket = new aws.s3.Bucket(
     },
 );
 
-// Create a bucket for handling uploads of website archive files.
-const cluster = awsx.ecs.Cluster.getDefault();
+// Create a bucket for receiving uploaded tarballs of website builds.
 const archiveBucket = new aws.s3.Bucket("archive-bucket");
+
+// Create a container task for handling tarball uploads.
 const archiveHandler = new awsx.ecs.FargateTaskDefinition("archiveHandler", {
     container: {
         image: awsx.ecs.Image.fromPath("archiveHandlerImage", "./"),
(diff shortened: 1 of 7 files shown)


cnunciato:
> It looks like we are having the Pulumi program, e.g. what runs when you do pulumi up, spawn a Fargate ask, that then executes aws sync. But that seems like an unnecessary step.

The program doesn't execute the task, it just defines the upload handler. When the program runs, it does build the archive and upload it at the end of the run, but only because that's a _lot_ nicer to do in TypeScript, and in the same program, than it would be to do it externally in Bash (e.g., filtering based on which files have redirects, which runs some functions that'd already been written, etc.).

> The tradeoff there is that we can avoid the complexity of needing to tar up the output, upload it to the cloud, and then start a Fargate task to download, unzip, and finally sync it... but instead have our Travis CI worker just run pulumi up (which we don't expect to ~ever have changes) and then run sync.sh on the same machine.

We could totally do that, yes -- I just went a step further to work with an archive just out of wanting to make it faster (and since that's what Pulumify does today). Note that it may not even be necessary to define an external Dockerfile and sync.sh (which I didn't love, TBH), as `awsx.ecs.Image.fromFunction` should theoretically be able to let us define everything inline. (cc @CyrusNajmabadi)

infrastructure/index.ts
```diff
@@ -41,26 +44,56 @@ const contentBucket = new aws.s3.Bucket(
...
+});
+
+// Handle archive uploads by running container task that downloads and unpacks the file
+// and synchronizes its contents with S3.
+archiveBucket.onObjectCreated("onArchiveUploaded", new aws.lambda.CallbackFunction<aws.s3.BucketEvent, void>("onUploadHandler", {
+    runtime: "nodejs10.x",
+    policies: [
+        aws.iam.ManagedPolicies.AWSLambdaFullAccess,
+        aws.iam.ManagedPolicies.AmazonEC2ContainerServiceFullAccess,
+    ],
```
CyrusNajmabadi: can simplify to `policies: awsx.ecs.TaskDefinition.defaultTaskRolePolicyARNs()`

CyrusNajmabadi:
> as awsx.ecs.Image.fromFunction should theoretically be able to let us define everything inline. (cc @CyrusNajmabadi)

Yes. This should be possible. :)

I've validated at least some simply cases of having some event run a fargate task definition.  Note that loggings that the task-def does inside hte lambda (i.e. through console.log) can be found in cloudwatch.  Look for the loggroup in your stack with the name of your container.  i.e.:

image

infrastructure/index.ts
```diff
@@ -341,6 +304,67 @@ function translateRedirect(filePath: string, redirect: string): string {
...
+                filter: (filePath, stat) => !redirectPaths.has(filePath),
+                file: archivePath,
+                C: config.pathToWebsiteContents,
+                portable: true,
+            },
+            fs.readdirSync(config.pathToWebsiteContents),
+        );
+
+        // Upload the archive.
+        const result = s3.putObject({
```
justinvp: Do we need to `await` the promise to prevent it from being unresolved?

```suggestion
        const result = await s3.putObject({
```

chrsmith:
> > Note that as currently written this change will delete the website momentarily (because the code defining file resources has been removed 😱) before the process to upload, unpack and sync its replacement contents is initiated.
> Do we need to mitigate this in some way?

Yeah, we definitely should. Even if it will just be a blip in the grand scheme of things, we should prevent the outage on www.pulumi.com since we can.

Here's a proposal for how we can deploy this PR without any issues:

1. Disable the Travis CI for the repository. (Just to make sure no more updates are going to happen while we make the change.)
2. Run `pulumi stack export` to get the checkpoint, and edit it to delete every `s3:Object` resource in it.
3. Run `pulumi stack import`. So at this point Pulumi is only tracking the CDN, S3 bucket, etc.
4. Run the `pulumi preview` and `pulumi up` manually from a developer machine. Just to confirm that the preview shows creating the new Fargate task, etc.
5. When we run the update the "aws s3 cp" should "just work", but I believe there is a `--dry-run` flag that might be good to add first... so when we look at the output for the docker container, we can confirm that there isn't a zillion deletes.

That should work, does that sound right to everybody else?

infrastructure/index.ts
```diff
@@ -224,80 +253,11 @@ const cdn = new aws.cloudfront.Distribution(
...
-        }
-    }
-}
-
-// Some files do not get the correct mime/type inferred from the mime package, and we
-// need to set our own.
-function getMimeType(filePath: string): string | undefined {
-    // Ensure that latest-version's mime type is always text/plain. Otherwise it
-    // will end up being set to binary/octet-stream, which is not what we want.
-    if (path.basename(filePath) === "latest-version") {
```
cnunciato: TODO: Make sure this is handled, too.

cnunciato:
@chrsmith That sounds good to me, yep, and I'm happy to take that on.

I think this should be good to go, then. Thoughts?

cnunciato:
Closing this in favor of #2345, which handles the process in two steps.

Issue #1364, opened 2019-07-15
Add an Events listing


cnunciato:
To help raise the visibility of Pulumi-related events, we should create an Events listing on the website. Entries would be similar to blog posts, but be located at a different path (e.g, `/events`, or `/blog/events`), and with some additional properties, such as

* `event_date` and time, so announcements can be published in advance
* `registration_url`, for linking to outside websites, like Meetup
* `location`, e.g., "Seattle, WA", and maybe with an address and/or maps link

... and probably some others as well. (cc @aydrian.)

We might also consider publishing these events in a separate RSS feed, since it's easy and offers one more way to subscribe to them.

aydrian:
I vote for not under blog. We should have a start and end date with logic smart enough to know if it's a single day. It would be nice to tag them as sponsorship or presenting, and categorize them as conference, meetup, hackathon, webinar, etc. Maybe have a short description as well. Have location and some kind of url for linking is also good.

joeduffy:
@cnunciato Adding this to M31 for consideration. As part of figuring out our overall 2020 roadmap of conferences and workshops (first 2 weeks of Q1), this would be a good thing to have.

joeduffy:
As part of this, it would also be good to have a place to aggregate webinars, past talks (with transcriptions), etc.

zchase:
Work being done in this PR: https://github.com/pulumi/docs/pull/2318

lukehoban:
Is this done now that #2318 has landed?  If not, what is remaining - and should we open a new issue to track any follow ups?

zchase:
Yup this is done -- Created another issue for follow up items: https://github.com/pulumi/docs/issues/2416

PR #2318, opened 2020-01-29
Add events pages

zchase:
### Proposed changes

This PR adds an event list page and event-specific landing pages.

Right now I have framework for the page in place, but it needs some tender love and care. I added a CTA to the right hand side of the listing page, but it is just a placeholder for now. I think we can create a much more appealing CTA, but I will need some help designing out the CTA.

There is functionality for filtering the event list by "tag". Right now I have started with a few generic tags but we should finalize exactly what is going to be on the page and how we would like to filter the results for users.

eventList

Any and all feedback is appreciated!

### Related issues (optional)

https://github.com/pulumi/docs/issues/1364


assets/js/event-filtering.js
```diff
@@ -0,0 +1,67 @@
...
+    var eventFilterParent = $("#eventFilter");
+    var tags = [];
+
+    for (var i = 0; i < eventsElements.length; i++) {
+        var elem = eventsElements[i];
+        var text = $(elem).text();
+        if (tags.indexOf(text) === -1) {
+            tags.push(text);
+            eventFilterParent.append(
+                '<input type="checkbox" class="mx-2" value="' + text + '">' + text + '</p>'
```
chrsmith: What if the value of text is somehow `<script>alert("lol, y'all been hacked")</script>`? Is there a jQuery method for sanitizing input in situations like this? (Just to err on the side of belt+suspenders.)

assets/js/event-filtering.js
```diff
@@ -0,0 +1,67 @@
+$(function() {
+    var eventsElements = $(".event-tags span");
```
chrsmith: What does jQuery return if no matches are found? Does it return `[]` or `undefined`? Similarly, what about `$("#eventFilter")` below? (In that case, I assume it's a single object?)

In either case, it seems like we should add some basic checks to ensure the script doesn't fail at runtime. Right?

assets/js/event-filtering.js
```diff
@@ -0,0 +1,67 @@
...
+                '<input type="checkbox" class="mx-2" value="' + text + '">' + text + '</p>'
+            );
+        }
+    }
+
+    $("input[type='checkbox']").on("click", function() {
+        var inputs = $("input[type='checkbox']");
+        var chosenInputs = [];
+
+        for (var z = 0; z < inputs.length; z++) {
```
chrsmith: Nit: Any reason why you didn't want to use `var i` like in the `for`-loop above?

assets/js/event-filtering.js
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     *
+     * @param {*[]} arr1 An array
+     * @param {*[]} arr2 An array
+     */
+    function checkForIntersection(arr1, arr2) {
+        for (var i = 0; i < arr1.length; i++) {
+            var val = arr1[i];
+            var hasIntersection = arr2.indexOf(val) > -1;
```
chrsmith: Nit: IMHO just writing this as `if (arr2.indexOf(val) !== -1)` seems a little cleaner.

But `indexOf` will only work using referential equality, right?

```javascript
var arr1 = [ { foo: "alpha" } ];
var arr2 = [ { foo: "alpha" } ];

// Returns false, since the objects are different references.
checkForIntersection(arr1, arr2)
// Returns true, since the objects would have the same reference.
checkForIntersection(arr1, arr1)
```

Right? Probably worth clarifying in the comment above, IMHO.

assets/js/event-filtering.js
```diff
@@ -0,0 +1,67 @@
...
+        var events = $("#eventList li");
+
+        for (var i = 0; i < events.length; i++) {
+            var event = $(events[i]);
+            var dataEventType = event.attr("data-event-type").split(",");
+            var shouldBeVisible = tags.length === 0 ? true : checkForIntersection(tags, dataEventType);
+
+            if (shouldBeVisible) {
+                event.removeClass("hidden");
+            }else {
```
chrsmith: Mega-pedantic nit: Space after `}`, so `} else {`.

content/events/workshop-amsterdam-2020-03-30.md
```diff
@@ -0,0 +1,55 @@
+---
+title: KubeCon EU  Pre-day Workshop
+url: /events/workshop-amsterdam-2020-03-30
+type: page
+layout: event-single
+calendly_url: "https://calendly.com/pulumi/kubecon?month=2020-03&hide_event_type_details=1&primary_color=512668"
```
chrsmith: Should we just put the `&hide_event_type_details=1&primary_color=512668` part of the URL in the template itself? It seems kinda error prone if we expect the `calendly_url` in the page's front mater to have the same formatting, etc.

content/events/workshop-amsterdam-2020-03-30.md
```diff
@@ -0,0 +1,55 @@
+---
+title: KubeCon EU  Pre-day Workshop
+url: /events/workshop-amsterdam-2020-03-30
```
chrsmith: Is this necessary? If we are just going to match the ~filename, shouldn't we just use whatever built-in support Hugo has for referring to the current page's filename?

content/events/workshop-amsterdam-2020-03-30.md
```diff
@@ -0,0 +1,55 @@
...
+
+11:30am - Lunch will be served in the meeting room.
+
+12:00-3:00pm - Workshop with hands-on labs.
+
+3:00-4:00pm - Happy hour with drinks/snacks.
+
+**Instructors**
+
+Joe Duffy - CEO, Pulumi <a href="https://twitter.com/@funcofjoe" target="_blank">@funcofjoe</a>
```
chrsmith: Don't worry about it for this PR, but it might be nice to link the persons name to the "authors" section on the blog, e.g. https://www.pulumi.com/blog/author/joe-duffy/. Maybe useful?

data/events.toml
```diff
@@ -0,0 +1,56 @@
+[[events]]
+type = ["workshop"]
```
chrsmith: I assume you already looked into it, but I presume there is no way for us to treat the "collection" of event posts as data? It just seems unfortunate we need to duplicate the metadata here as well as in the `/events/xxx.md` file. But I'll totally believe you if you say that isn't possible and/or prohibitively a PITA.

layouts/page/event-single.html
```diff
@@ -0,0 +1,27 @@
...
+
+{{ define "main" }}
+    <div class="flex flex-col md:flex-row p-10 md:px-20">
+        <div class="w-full md:w-6/12 lg:w-8/12 mx-auto md:py-12">
+            {{ .Content }}
+        </div>
+        <div class="w-full md:w-6/12 lg:w-4/12 mx-auto">
+            <!-- Calendly inline widget begin -->
+
+            <div class="calendly-inline-widget" data-url="{{ .Params.calendly_url }}" style="min-width:320px;height:750px;"></div>
```
chrsmith: Minor nit: Rather than having the inline-style here, can we just add `id="calendly-widget"` and put the styles somewhere in our .CSS file? it just seems like that might be easier to maintain.

scripts/generate_events.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+go run ./tools/eventgen/*.go
```
chrsmith: Could you please expand on the doc comment explaining what this script does, and what its output is? It isn't immediately obvious.

test.md
```diff
@@ -0,0 +1,52 @@
+
+---
+title: Event 1
```
chrsmith: Just as a reminder: probably something to delete before you check in?

Or would `draft: true` in the front matter exclude this from most builds?

data/events.toml
chrsmith: ... I see `tools/eventgen/main.go ` later in the PR, which makes this much more palatable. Though if it is valid, putting a header like the following would be much appreciated:

```
# DO NOT MODIFY
# This file is automatically generated by the ./tools/xxx. It is built from
# ...
```

tools/eventgen/main.go
```diff
@@ -0,0 +1,174 @@
...
+import(
+	"fmt"
+	"io/ioutil"
+	"os"
+	"time"
+
+	"github.com/BurntSushi/toml"
+)
+
+// Function for generating the front matter of an event file
```
chrsmith: This isn't quite accurate, as it also includes the first part of the page too, with a standard header with the event's details. I'd suggest updating the comment.

tools/eventgen/main.go
```diff
@@ -0,0 +1,174 @@
...
+	var date string
+	if formattedStartDate == formattedEndDate {
+		date = fmt.Sprintf("%s, %s", formattedStartDate, startYear)
+	} else if startYear != endYear {
+		date = fmt.Sprintf("%s, %s - %s, %s", formattedStartDate, startYear, formattedEndDate, endYear)
+	} else {
+		date = fmt.Sprintf("%s - %s, %s", formattedStartDate, formattedEndDate, endYear)
+	}
+
+	frontMatter := fmt.Sprintf(`---
```
chrsmith: This is fine for now, but in the future, this format string is perhaps complicated enough that you'd want to just break down and use Go's built-in text templating stack. As it would make the rendering a little less error prone. (No need to remember the ordering of the `%s` and instead just use `{{ title }}` and `{{ url }}`, etc.)
https://golang.org/pkg/text/template/

tools/eventgen/main.go
```diff
@@ -0,0 +1,174 @@
...
+		if event.ContentType != "" {
+			filteredEvents = append(filteredEvents, event)
+		}
+	}
+
+	return filteredEvents
+}
+
+func writeEventFiles(events []Event) string {
+	event := events[0]
```
chrsmith: Any reason why you didn't just write this as:

```go
for _, event := range events {
    writeEventFile(event)
}
return "Event pages have been generated."
```

That seems much more straight forward.

assets/js/event-filtering.js
zchase: They return `[]` if it does not exist. I added checks to make sure the element returns before proceeding, to err on the side of caution. They are not really needed because nothing will happen if `eventsElements` comes back empty.

https://learn.jquery.com/using-jquery-core/faq/how-do-i-test-whether-an-element-exists/

assets/js/event-filtering.js
zchase: No reason in particular. Probably a side-effect from something I was doing earlier that had two for loops in the same function. Good catch.

Side note: `z`'s are much cooler than `i`'s 😎

assets/js/event-filtering.js
zchase: Yup, you're right. Clarified in the comment

assets/js/event-filtering.js
zchase: One day I will submit a PR with correct spacing in `if... else...` blocks...

content/events/workshop-amsterdam-2020-03-30.md
zchase: 👍

content/events/workshop-amsterdam-2020-03-30.md
zchase: Yup, you're right, good catch.

content/events/workshop-amsterdam-2020-03-30.md
zchase: Yeah, at some point in the future we will want to add pictures of the instructors and maybe more contact info.

test.md
zchase: Oh snap, didn't realize this made it in... Removing this..

tools/eventgen/main.go
zchase: Personal bias against for-loops from my background in javascript. I figured it was best to do each file one by one than to loop through. But I guess it does it one by one?

tools/eventgen/main.go
zchase: done

scripts/generate_events.sh
zchase: 👍

data/events.toml
zchase: 👍

assets/js/event-filtering.js
zchase: I adjusted the code to generate the elements via the document API and user `innerText` to set the value of the text. So that should cover some malicious situation where the `tag` value is a script like you mentioned above. Everything will just get rendered as a string .

package.json
```diff
@@ -8,6 +8,7 @@
   "dependencies": {
     "broken-link-checker": "^0.7.8",
     "clipboard-polyfill": "^2.8.1",
+    "toml": "^3.0.0",
```
cnunciato: Are we using this?

cnunciato:
Why not just make a Markdown file for every event? A [Hugo archetype](https://gohugo.io/content-management/archetypes/) would make that dead simple and give you the fields you need with placeholder data, and would eliminate the need for having an additional Go program to perform what's essentially the same functionality. Unless I'm missing something we wouldn't get from that approach, I'd really rather we kept it simple and leaned on Hugo here.

cnunciato:
Oops. 😇

cnunciato:
We use an archetype for the blog, too, incidentally:

https://github.com/pulumi/docs/blob/master/BLOGGING.md#make-a-new-post
https://github.com/pulumi/docs/tree/master/archetypes/blog-post

zchase:
@cnunciato I think archetypes are close to what we want but not quite exactly what we want. The way I understand archetypes is that we would have a common front matter template that we use to generate the event files. We also need to populate the content of the page, which will be duplicated across common events. To me it doesn't look like Hugo will populate the content of the markdown so we would still have to copy/paste the description into the file. 

The go tool will auto-generate the entire page so we don't have to intervene manually. So all a person has to do to add an event is to add it to `data/events.toml` and the tool will handle everything else. With archetypes, I imagine you would have to add the event to the data file, run `hugo new` to generate the front matter, fill in relevant info to the front matter, then add in the event description to the markdown file.

cnunciato:
That's right -- you'd run `hugo new`, get a new file at `events/some-event.md`, and then have to fill in the fields you'd otherwise have to fill into events.toml, anyway. So events.toml would go away and just be replaced by iterating over the list of event pages, as data, instead.

The one difference is coping with common elements among event types, which I hear, but that's also what templates and partials are for, too. I'm still not seeing what having this additional layer of complexity buys us in exchange for the flexibility we get from just using the tools we already have.

cnunciato:
Could we constrain this horizontally using `container` somehow?

image

assets/js/event-filtering.js
```diff
@@ -0,0 +1,119 @@
...
+        label.className = "cursor-pointer";
+        label.setAttribute("for", checkbox.id);
+
+        container.appendChild(checkbox);
+        container.appendChild(label);
+
+        return container;
+    }
+
+    // Loop through the tags and creat a unique array of tag names and
```
cnunciato:
```suggestion
    // Loop through the tags and create a unique array of tag names and
```

archetypes/workshop/index.md
```diff
@@ -0,0 +1,33 @@
...
+    start_date: ""
+    # The end date of an event. Format YYYY-MM-DD
+    end_date: ""
+    # The start and end time of an event with the time zone.
+    # i.e. 5:30PM - 8:30PM (PT)
+    # This is only displayed on event specific pages.
+    time: ""
+    # The event description shown on the event list page.
+    description: ""
+    # The calendly registrtion url for event specific pages.
```
cnunciato:
```suggestion
    # The Calendly registrtion url for event specific pages.
```

PR #2345, opened 2020-01-30
Add support for deploying with S3 sync

cnunciato:
This change represents the first of two steps in migrating to using`aws s3 sync` to manage deployments of the website. It is completely additive (new bucket, event handler and associated infrastructure) so it can be merged safely without disrupting the current deployment.

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -1,11 +1,15 @@
 // Copyright 2016-2019, Pulumi Corporation.  All rights reserved.
 
 import * as aws from "@pulumi/aws";
+import * as awsx from "@pulumi/awsx";
 import * as pulumi from "@pulumi/pulumi";
 
 import * as fs from "fs";
+import * as glob from "glob";
 import * as mime from "mime";
 import * as path from "path";
+import * as tar from "tar";
+import * as tmp from "tmp";
 
 const stackConfig = new pulumi.Config();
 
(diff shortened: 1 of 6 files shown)


infrastructure/index.ts
```diff
@@ -41,26 +47,77 @@ const contentBucket = new aws.s3.Bucket(
...
     {
         protect: false,
-    });
+    },
+);
 
-// contentBucket needs to have the "public-read" ACL so its contents can be ready by CloudFront and
-// served. But we deny the s3:ListBucket permission to prevent unintended disclosure of the bucket's
-// contents.
-const denyListPolicyState: aws.s3.BucketPolicyArgs = {
```
cnunciato: Removing this to allow for `s3 ls`.

justinvp: You wouldn't happen to know of a policy that would deny it from being listed by any rando, but would still allow us to list it internally in the AWS console?

cnunciato: I do not, but I'm not yet an expert here, either. I can imagine there might be some way to deny all, but allow some assumed role, perhaps?

cnunciato: We should be able to apply an exception to that `Deny` rule, so I've restored this block and will submit a separate PR with that alone. Thanks for the nudge on it. 😄

PR #2351, opened 2020-01-30
Cut over to s3 sync-based deployment

cnunciato:
This change modifies the website's CloudFront distribution to point to the bucket now being populated by `aws s3 sync`.

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -29,7 +29,7 @@ const config = {
 };
 
 // redirectDomain is the domain to use when redirecting.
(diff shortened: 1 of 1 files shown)


cnunciato:
Looks like [the build that added the A record](https://travis-ci.com/pulumi/docs/builds/146837934) for `www-prod` was canceled because another was in progress, which is why this PR shows that record being created.

cnunciato:
⚠️ Don't merge before https://github.com/pulumi/docs/pull/2358.

PR #2358, opened 2020-01-31
Process redirects in the sync step

cnunciato:
This PR changes the behavior of the new sync-based deployment to process redirects with `s3api` rather than with Pulumi, because the `s3 sync` command (specifically with the `--delete` option, which we want) is removing those redirect objects from the S3 bucket after the Pulumi update completes. Takes an extra minute or two to do it this way (😢), but is necessary.

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -473,41 +473,29 @@ async function createAliasRecord(
         });
 }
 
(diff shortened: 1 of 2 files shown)


cnunciato:
Proactively merging this in order to verify bucket contents match up before merging #2351.

PR #2359, opened 2020-01-31
Specify the correct source directory

cnunciato:
This fixes an oversight in #2358 that looked for the redirect source file in the wrong place.


cnunciato:
One more proactive merge 😇, again just to make sure the old and new buckets match up before cutting over.

PR #2363, opened 2020-01-31
Add ppc landing pages

zchase:
This PR adds the framework and first ppc landing page for our site. You will notice the header and footer have been hidden for these pages. We want to keep any who clicks on an ad focused on the goal of the page. Including additional navigation can confuse users and lead them to become distracted. The other important thing to note are the logos and customer quotes. They provide social proof for us, which is helpful when trying to convert users. They can see that actual companies use our product.


zchase:
@joeduffy @lukehoban Pushed up an initial desktop version of the new landing page. It would be awesome to get some feedback on it. I tried to show how Pulumi reduces lines of code in the hero as clear as possible and am interested in what y'all think about it.

PR #2364, opened 2020-01-31
Define CDN aliases explicitly

cnunciato:
The brief website outage we experienced this morning occurred because the CloudFront distribution aliases in use with www.pulumi.com had not been declared in the Pulumi program that defines that distribution. (They'd been keyed into the AWS Console by hand as part of the GREAT WEBSITE MIGRATION last year.) Since some of the properties of the distribution were modified today in #2351, when that commit's update ran, it cleared out the list of aliases in AWS, rendering the website temporarily inaccessible. This PR adds those aliases to the program explicitly to prevent that from occurring in the future.

⚠️ ⚠️ ⚠️ ⚠️ ⚠️ 

Note that as of now, because the most immediate fix for today's outage was to specify those values by hand _again_, the CloudFront distribution's [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) and aliases list are once again out of sync, **so another manual `pulumi refresh` will need to be performed before this change is merged** (or another outage will occur).

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -300,6 +300,19 @@ const distributionArgs: aws.cloudfront.DistributionArgs = {
     },
 };
 
+// domainAliases is a list of CNAMEs that accompany the CloudFront distribution. Any
+// domain name to be used to access the website must be listed here.
+const domainAliases = [];
+// websiteDomain is the A record for the website bucket associated with the website.
+domainAliases.push(config.websiteDomain);
+// targetDomain is the A record associated with the bucket populated by Pulumi. It may be
+// removed that bucket is removed.
+domainAliases.push(config.targetDomain);
+// redirectDomain is the domain to use for fully-qualified 301 redirects.
+if (config.redirectDomain) {
+     domainAliases.push(config.redirectDomain);
+}
+
 // NOTE: Sometimes updating the CloudFront distribution will fail with:
 // "PreconditionFailed: The request failed because it didn't meet the preconditions in one or more
 // request-header fields."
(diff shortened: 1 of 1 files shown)


infrastructure/index.ts
```diff
@@ -312,6 +312,11 @@ const cdn = new aws.cloudfront.Distribution(
     {
         protect: true,
         dependsOn: [ websiteBucket, websiteLogsBucket ],
+        aliases:  [
+            config.websiteDomain,
+            config.targetDomain,
+            config.redirectDomain || "",    // redirectDomain may be undefined.
+        ].filter(domain => domain !== ""),
```
chrsmith: The "array literal" immediately passed to `.filter(` seems kinda lame, IMHO.

Can we just declare a variable elsewhere? That would also give us the opportunity to better explain the (also lame) reason why we have up to 3 different aliases. (* I can call this all lame, since it's kinda-sorta my fault we have those three 😭 )

```
const domainAliases = [];
// website domain does X. It is important because of Y
domainAliases.push(config.websiteDomain);
//  target domain does X. This is a lame hack, but necessary because of Y.
domainAliases.push(config.targetDomain);
// Optionally set the redirect domain, you know, because of X.
if (config.redirectDomain) {
     domainAliases.push(config.redirectDomain);
}
```

What do you think?

cnunciato: Sure, that works. 👍

infrastructure/index.ts
```diff
@@ -300,6 +300,19 @@ const distributionArgs: aws.cloudfront.DistributionArgs = {
     },
 };
 
+// domainAliases is a list of CNAMEs that accompany the CloudFront distribution. Any
+// domain name to be used to access the website must be listed here.
+const domainAliases = [];
+// websiteDomain is the A record for the website bucket associated with the website.
+domainAliases.push(config.websiteDomain);
+// targetDomain is the A record associated with the bucket populated by Pulumi. It may be
```
chrsmith: Nit: "It may be removed that bucket is removed." seems like it is missing a word.

PR #2366, opened 2020-01-31
Fix error in AWS IoT TopicRule example.

MichaelStott:
AWS IoT TopicRule example in documentation does not currently work. `name` property must be set. Otherwise, it will fail with the following:

```
Diagnostics:
  aws:iot:TopicRule (rule):
    error: aws:iot/topicRule:TopicRule resource 'rule' has a problem: Name must match the pattern ^[a-zA-Z0-9_]+$
```

### Proposed changes

Add the name field :upside_down_face:


stack72:
Thank you so much!!

PR #2367, opened 2020-02-01
Add confirmation page for workshops

zchase:
This PR adds a confirmation page for workshop sign-ups. We are going to use Calendly to handle the registrations, and after a person registers on Calendly they will be redirected to this page. Eventually, we will want to have confirmation pages for other events, and using this layout we should just need to create another markdown file and define the front matter values.

### DIFF
--- a/content/confirmation/_index.md
+++ b/content/confirmation/_index.md
@@ -1,6 +1,6 @@
 ---
 private: true
 title: Confirmation
-meta_desc: This page is provides confirmation about register for a Pulumi event, webinar, or workshop.
+meta_desc: This page provides confirmation after registering for a Pulumi event, webinar, or workshop.
 block_external_search_index: true
 ---
--- a/content/confirmation/_index.md
+++ b/content/confirmation/_index.md
@@ -1,0 +1,6 @@
+---
+private: true
+title: Confirmation
+meta_desc: This page is provides confirmation about register for a Pulumi event, webinar, or workshop.
+block_external_search_index: true
+---
--- a/content/confirmation/_index.md
+++ b/content/confirmation/_index.md
@@ -1,6 +1,3 @@
 ---
-private: true
-title: Confirmation
-meta_desc: This page provides confirmation after registering for a Pulumi event, webinar, or workshop.
-block_external_search_index: true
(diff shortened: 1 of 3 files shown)


content/confirmation/_index.md
```diff
@@ -0,0 +1,6 @@
+---
+private: true
+title: Confirmation
+meta_desc: This page is provides confirmation about register for a Pulumi event, webinar, or workshop.
```
justinvp:
```suggestion
meta_desc: This page provides confirmation after registering for a Pulumi event, webinar, or workshop.
```

content/confirmation/workshop.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Thank You | Workshop
+meta_desc: This page is provides confirmation about register for a Pulumi workshop.
```
justinvp:
```suggestion
meta_desc: This page provides confirmation after registering for a Pulumi workshop.
```

content/confirmation/_index.md
```diff
@@ -0,0 +1,6 @@
...
+block_external_search_index: true
+---
```
justinvp: Do we ever expect to use this index page? Instead of having it as effectively empty page, we could consider just redirecting to the homepage, e.g. changing the content of this file to:

```markdown
---
redirect_to: /
---
```

content/confirmation/workshop.md
```diff
@@ -0,0 +1,22 @@
...
+type: page
+layout: confirmation
+url: /confirmation/workshop
```
justinvp: Is the `url` parameter necessary? I think we might be able to delete this line.

```suggestion
```

content/confirmation/workshop.md
zchase: Good catch, thanks!

content/confirmation/_index.md
zchase: Yup, that worked. I removed all the extra stuff and just have the redirect now.

content/confirmation/workshop.md
zchase: I guess it's not. I removed it.

PR #2368, opened 2020-02-01
Stop managing bucket objects with Pulumi

cnunciato:
Now that we're serving content from the `www-prod` bucket, which is managed by a sync process external to the Pulumi program, we can remove the code that's been managing bucket objects up to now.

This PR will delete most of the files in that bucket, but leave the bucket itself, along with the bucket used for access logs prior to cutover. I'll remove these last few items in a follow-up PR. 

This is the change that should significantly reduce deployment times going forward.

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -330,22 +330,6 @@ const cdn = new aws.cloudfront.Distribution(
     }
 );
 
-// crawlDirectory recursive crawls the provided directory, applying the provided function
-// to every file it contains. Doesn't handle cycles from symlinks.
-function crawlDirectory(dir: string, f: (_: string) => void) {
-    const files = fs.readdirSync(dir);
-    for (const file of files) {
-        const filePath = `${dir}/${file}`;
-        const stat = fs.statSync(filePath);
-        if (stat.isDirectory()) {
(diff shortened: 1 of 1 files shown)


cnunciato:
☝️ The [Travis run](https://travis-ci.com/pulumi/docs/builds/147010715) makes it a bit clearer what's changing here:

```
 ++ aws:ecs:TaskDefinition archive-handler create replacement [diff: ~containerDefinitions]
 +- aws:ecs:TaskDefinition archive-handler replace [diff: ~containerDefinitions]
 ~  aws:lambda:Function archive-bucket-callback update [diff: +lastModified~code]
    aws:lambda:Permission archive-bucket-subscription  
 ~  aws:s3:BucketNotification archive-bucket-subscription update [diff: ~lambdaFunctions]
 -- aws:ecs:TaskDefinition archive-handler delete original [diff: ~containerDefinitions]
```

cnunciato:
> It would be a shame to lose that data if we ever wanted to get a rough estimate of the historic performance of blog posts, etc.

Absolutely, yeah -- I wasn't planning on deleting the actual logs, of course! 😄 Figured we'd chat and come up with a good place to put them before deleting the code that manages the bucket.

PR #2373, opened 2020-02-04
Allow s3:ListBucket access for account users

cnunciato:
This PR removes the (now empty) S3 bucket that served the website prior to #2368, and applies a `ListBucket` policy to the current bucket that denies public list access, but grants it to members of the AWS account that manages the stack.

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -52,7 +52,7 @@ const websiteBucket = new aws.s3.Bucket(
     },
 );
 
-// The website bucket needs to have the "public-read" ACL so their contents can be read by
+// The website bucket needs to have the "public-read" ACL so its contents can be read by
 // CloudFront and served. But we deny the s3:ListBucket permission to anyone but account
 // users to prevent unintended disclosure of the bucket's contents.
 const policy = new aws.s3.BucketPolicy("website-bucket-policy", {
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -26,48 +26,12 @@ const config = {
     certificateArn: stackConfig.require("certificateArn"),
     // redirectDomain is the domain to use for any redirects.
     redirectDomain: stackConfig.get("redirectDomain") || undefined,
+    // principalAccount is the AWS account used for managing this stack.
+    principalAccount: stackConfig.require("principalAccount"),
 };
 
 // redirectDomain is the domain to use when redirecting.
 const redirectDomain = config.redirectDomain || config.websiteDomain;
-
-// contentBucket stores the static content to be served via the CDN.
-const contentBucket = new aws.s3.Bucket(
-    "contentBucket",
-    {
-        bucket: config.targetDomain,
-        acl: "public-read",
-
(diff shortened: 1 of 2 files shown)


infrastructure/index.ts
```diff
@@ -50,43 +52,28 @@ const contentBucket = new aws.s3.Bucket(
     },
 );
 
-// contentBucket needs to have the "public-read" ACL so its contents can be ready by CloudFront and
-// served. But we deny the s3:ListBucket permission to prevent unintended disclosure of the bucket's
-// contents.
-const denyListPolicyState: aws.s3.BucketPolicyArgs = {
-    bucket: contentBucket.bucket,
-    policy: contentBucket.arn.apply((arn: string) => JSON.stringify({
+// The website bucket needs to have the "public-read" ACL so their contents can be read by
```
justinvp: Super nitpick: "its" still reads better to me, but happy to go with "their" if it is more correct.

```suggestion
// The website bucket needs to have the "public-read" ACL so its contents can be read by
```

cnunciato: Ah yes, you're right -- this was written when there were two buckets to manage, but now we only have one.

infrastructure/index.ts
```diff
@@ -50,43 +52,28 @@ const contentBucket = new aws.s3.Bucket(
...
+                    },
+                },
             },
         ],
     })),
-};
-const denyListPolicy = new aws.s3.BucketPolicy("deny-list", denyListPolicyState);
-
-// websiteBucket stores the static content to be served via the CDN.
-const websiteBucket = new aws.s3.Bucket(
```
chrsmith: Just confirming: The `websiteBucket` resource is still around. It's just the `contentBucket` that was removed, because we now push the data directly into `websiteBucket`. (Since we switched over to the new setup, right?)

infrastructure/index.ts
```diff
@@ -50,43 +52,28 @@ const contentBucket = new aws.s3.Bucket(
...
         Version: "2008-10-17",
         Statement: [
             {
                 Effect: "Deny",
                 Principal: "*",
                 Action: "s3:ListBucket",
                 Resource: arn,
+                Condition: {
+                    StringNotEquals: {
+                        "aws:PrincipalAccount": config.principalAccount,
```
chrsmith: Why don't we pull the account ID at runtime? So that it always matches the AWS account used, and "owns" the bucket? e.g.
https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/aws/#getCallerIdentity

Hooking it up via config just means that it's something that we could get wrong. (Configure to use account X, but actually deploy to account Y.) By pulling this at runtime, it will always "just work". Right?

cnunciato: That sounds great! How do o do that? 😉

infrastructure/index.ts
cnunciato: That’s right — the diff just looks terrifying, presumably because of the way I duplicated `contentBucket`’s declaration during the cutover.

PR #2376, opened 2020-02-05
97th Floor SEO optimization

97thFloor:
@cnunciato this was optimized for KW: CloudFormation. I realized I made the changes, but not a pull request.

### Proposed changes

Instances of keyword added, as well as relevant language to aid ranking, based on semantic analysis of SERP.

I'm not sure how to create headers, but "What is Pulumi?", "What is CloudFormation?" and "The Pulumi Difference" were intended to be subheadings. If that's not possible they're also fine as they are.

### DIFF
--- a/content/docs/intro/vs/cloud_templates.md
+++ b/content/docs/intro/vs/cloud_templates.md
@@ -1,7 +1,6 @@
 ---
-title: Pulumi vs. Cloud Templates (AWS CloudFormation, etc.)
-meta_desc: This page gives an overview of the major differences between Pulumi
-           and Cloud Templates (AWS CloudFormation, etc.).
+title: Pulumi vs. AWS CloudFormation and Other Templates | Pulumi
+meta_desc: This page gives an overview of the major differences between Pulumi and Cloud Templates such as AWS CloudFormation.
 linktitle: Cloud Templates
 menu:
   intro:
@@ -11,18 +10,26 @@ menu:
 aliases: ["/docs/reference/vs/cloud_templates/"]
 ---
 
+What is Pulumi?
+
+Pulumi offers a multi-cloud alternative to typical, more restrictive infrastructure-as-code template solutions, such as AWS CloudFormation. Pulumi allows you to streamline processes through managing resources from different platforms all in one place, using real code, in your favorite language.
+
+What is CloudFormation?
+
 All major cloud providers offer their own form of infrastructure-as-"code" solution, typically by way of JSON or
 YAML-based templating solutions. This includes AWS CloudFormation and Azure Resource Manager (ARM) templates.
 
 These markup-based configuration files are often uploaded to a hosted service in the target cloud, where a hosted
 service will then process the files to create, update, or delete resources as necessary.
 
+The Pulumi Difference
+
 Pulumi's model shares a lot with these systems. Although Pulumi programs are written in imperative, familiar languages,
 they are ultimately evaluated to produce a similar set of create, update, or delete operations for your cloud resources.
 
-This is where Pulumi begins to diverge from these other solutions, however.
+This is where Pulumi begins to diverge from these other solutions such as AWS CloudFormation, however.
 
-Pulumi lets you use your favorite languages, instead of bespoke templating solutions. These typically use syntaxes that
+Pulumi lets you use your favorite languages — complete control rather than bespoke templating solutions such as CloudFormation. These templates typically use syntaxes that
 are awkward, hard to learn, and even harder to remember. Pulumi programs are just real code. Thanks to basic things
 like functions and classes, we've seen 25,000 line CloudFormation templates shrink to just a few hundred lines of code.
 This approach also leads to far less copy and pasting between projects because you can share packages.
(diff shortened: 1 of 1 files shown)


content/docs/intro/vs/cloud_templates.md
```diff
@@ -1,7 +1,6 @@
 ---
-title: Pulumi vs. Cloud Templates (AWS CloudFormation, etc.)
-meta_desc: This page gives an overview of the major differences between Pulumi
-           and Cloud Templates (AWS CloudFormation, etc.).
+title: Pulumi vs. AWS CloudFormation and Other Templates | Pulumi
```
joeduffy: A while back, we used to auto-append "| Pulumi" to all page titles. Is that worth bringing back?

content/docs/intro/vs/cloud_templates.md
```diff
@@ -42,5 +49,5 @@ There are many community-led projects which allow you to write code in a real la
 configuration templates as a sort of "compilation" step. These offer nice syntactic sugar on top of the raw
 templates, but the underlying model leaks through and carries forward all of the other abovementioned shortcomings.
 
-Finally, Pulumi is [open source](https://github.com/pulumi/pulumi) and community-driven. All of the cloud systems are
+Finally, Pulumi is [open source](https://github.com/pulumi/pulumi) and community-driven. All of the other cloud systems — such as Azure Automation or AWS CloudFormation — are
```
joeduffy:
```suggestion
Finally, Pulumi is [open source](https://github.com/pulumi/pulumi) and community-driven. All of the other cloud systems — such as Azure Resource Manager or AWS CloudFormation — are
```

content/docs/intro/vs/cloud_templates.md
97thFloor: There's room for adjusting to suit your brand, as it's not a super important SEO issue, but Moz does recommend the "| Brand Name" format for title tags in Google; it looks good and keeps things uniform, so that's what we typically recommend.

cnunciato: In most cases (including this page), we do indeed append this automatically, yep:

image

https://github.com/pulumi/docs/blob/master/layouts/partials/head.html#L66-L71

(So there's no need to specify these manually in frontmatter titles, @97thFloor.)

content/docs/intro/vs/cloud_templates.md
```diff
@@ -11,18 +10,26 @@ menu:
...
+The Pulumi Difference
+
 Pulumi's model shares a lot with these systems. Although Pulumi programs are written in imperative, familiar languages,
 they are ultimately evaluated to produce a similar set of create, update, or delete operations for your cloud resources.
 
-This is where Pulumi begins to diverge from these other solutions, however.
+This is where Pulumi begins to diverge from these other solutions such as AWS CloudFormation, however.
 
-Pulumi lets you use your favorite languages, instead of bespoke templating solutions. These typically use syntaxes that
+Pulumi lets you use your favorite languages — complete control rather than bespoke templating solutions such as CloudFormation. These templates typically use syntaxes that
```
cnunciato: This sentence is no longer clear. Should the dash be a colon? And "complete control" and "bespoke templating solutions" are not directly comparable enough to be separated with a "rather than". (In other words, languages and templating solutions are comparable, as they're both technologies. Control is not a technology.)

content/docs/intro/vs/cloud_templates.md
```diff
@@ -1,7 +1,6 @@
 ---
-title: Pulumi vs. Cloud Templates (AWS CloudFormation, etc.)
-meta_desc: This page gives an overview of the major differences between Pulumi
-           and Cloud Templates (AWS CloudFormation, etc.).
+title: Pulumi vs. AWS CloudFormation and Other Templates | Pulumi
+meta_desc: This page gives an overview of the major differences between Pulumi and Cloud Templates such as AWS CloudFormation.
```
cnunciato: Caps are unnecessary, as cloud templates isn't a proper noun.

```suggestion
meta_desc: This page gives an overview of the major differences between Pulumi and cloud templates such as AWS CloudFormation.
```

content/docs/intro/vs/cloud_templates.md
```diff
@@ -11,18 +10,26 @@ menu:
 aliases: ["/docs/reference/vs/cloud_templates/"]
 ---
 
+What is Pulumi?
```
cnunciato: Leading off with what sounds like an ad for Pulumi doesn't seem like the right thing, here, since these pages are aimed at delivering comparative info, and they'd seem to be working relatively well in that regard. I'd remove these headings and this introductory section and leave these pages structurally as they are today, unless breaking them up is solving a problem that can't be solved otherwise.

97thFloor: It was mostly to get the "What is CloudFormation?" part in there, as that is the first snippet question on the SERP, and CloudFormation is the primary KW. The "What is Pulumi?" part was more to balance it out and provide a clearer structure, since this is a comparison page. Would it be ok to switch the order and lead with CloudFormation?

content/docs/intro/vs/cloud_templates.md
```diff
@@ -11,18 +10,26 @@ menu:
...
 These markup-based configuration files are often uploaded to a hosted service in the target cloud, where a hosted
 service will then process the files to create, update, or delete resources as necessary.
 
+The Pulumi Difference
+
 Pulumi's model shares a lot with these systems. Although Pulumi programs are written in imperative, familiar languages,
 they are ultimately evaluated to produce a similar set of create, update, or delete operations for your cloud resources.
 
-This is where Pulumi begins to diverge from these other solutions, however.
+This is where Pulumi begins to diverge from these other solutions such as AWS CloudFormation, however.
```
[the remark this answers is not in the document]
97thFloor: Maybe "This is where Pulumi begins to diverge from solutions like AWS CloudFormation, however"?

content/docs/intro/vs/cloud_templates.md
97thFloor: It's probably fine to leave as is then, as long as we can still get the "CloudFormation" part in there.

zchase:
This PR is now being tracked in https://github.com/pulumi/docs/pull/2515. Closing.

PR #2377, opened 2020-02-05
Make the docs sidebars sticky

### LINKED ISSUE #1900: Consider making "On This Page" always visible

cnunciato:
This change applies sticky positioning and independent scrolling to the left- and right-hand sidebars in docs layouts, and highlights the on-the-page item that represents where you are in the overall page hierarchy.

Fixes #1900.

### DIFF
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -23,21 +23,50 @@ function generateOnThisPage() {
     var $ul = $(".on-this-page > ul");
     if ($ul) {
         var found = false;
+        var headings = [];
+
         $("h2, h3").each(function () {
             var $el = $(this);
             var id = $el.attr("id");
             var text = $el.text();
             var linkTitle = $el.data("link-title");
             var tag = $el.prop("tagName").toLowerCase();
+
             if (id && text) {
                 found = true;
-                $ul.append("<li class='" + tag + "'><a href='#" + id + "'>" + (linkTitle || text) + "</a></li>");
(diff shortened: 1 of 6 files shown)


justinvp:
Minor bug I noticed trying it out locally...

If I go to http://localhost:1313/docs/reference/pkg/nodejs/pulumi/aws/s3

And click on `AccountPublicAccessBlock` in the right nav (first item under `Resources`), it doesn't become bold. Instead `Resources` is bold. Same if I scroll through the page, `AccountPublicAccessBlock` never becomes bold. It goes from `Resources` being bold to `Bucket` being bold, skipping `AccountPublicAccessBlock`.

Screen Shot 2020-02-05 at 1 56 39 PM

cnunciato:
Oh, interesting! This is happening because of the height/margin trickery I'm doing with anchors that get the `:target` pseudoclass (e.g., those that are clicked). Going to have to think a bit on this one. 🤔

justinvp:
> Going to have to think a bit on this one.

Let's just go ahead and merge this. It's such a big improvement as-is, particularly on our larger pages. And then fix this minor issue subsequently.

PR #2381, opened 2020-02-05
Regen docs for pulumi@v1.10.0

pgavlin:


static/latest-version
```diff
@@ -1 +1 @@
-1.9.1
\ No newline at end of file
+v1.10.0
```
justinvp: I don't think we typically include the `v` prefix here.

Hopefully this GitHub suggestion won't add new line if you apply it through GH.

```suggestion
1.10.0
```

pgavlin: Fixed

PR #2390, opened 2020-02-06
Regen docs for pulumi@v1.10.1

pgavlin:


content/docs/get-started/install/versions.md
```diff
@@ -17,6 +17,8 @@ The current stable version of Pulumi is **{{< latest-version >}}**.
         </tr>
     </thead>
     <tbody>
+        {{< changelog-table-row version="1.10.1" date="2020-02-06" >}}
+        {{< changelog-table-row version="1.10.0" date="2020-02-05" >}}
```
pgavlin: Note that this will probably trip the broken link checker, as no Windows build was published for this release.

justinvp: How do we want to handle this?

- Drop `1.10.0` from the table? (Downside being you can't access that version for Linux/macOS)
- Leave it in the table, but live with the broken link?
  - Exclude https://get.pulumi.com/releases/sdk/pulumi-v1.10.0-windows-x64.zip from the broken link checker
- Enhance `changelog-table-row` to be able to hide the link for certain OSes?

pgavlin: I think option (3) is best long-term, but option (2) sounds like the right thing to do for now. What do I need to change to make that happen?

pgavlin: nvm, I think I got it

scripts/check-links.sh
```diff
@@ -28,7 +28,8 @@ check_links() {
         --exclude "https://ksonnet.io/" \
         --exclude "https://www.latlong.net/" \
         --exclude "https://media.amazonwebservices.com/architecturecenter/AWS_ac_ra_web_01.pdf" \
-        --exclude "https://www.packet.com/"
+        --exclude "https://www.packet.com/" \
+		--exclude "https://get.pulumi.com/releases/sdk/pulumi-v1.10.0-windows-x64.zip"
```
justinvp: Nit: tabs => spaces

```suggestion
        --exclude "https://get.pulumi.com/releases/sdk/pulumi-v1.10.0-windows-x64.zip"
```

PR #2394, opened 2020-02-06
Add a Brand page

### LINKED ISSUE #2188: Add a brand/logo page to the website
We need a page to allow those (e.g., conference organizers, press) who need logos and related brand assets to obtain them directly. This page should provide access to:

* Official Pulumi logos suitable for digital and print
* Brand colors (hex, HSL/RGB, and Pantone)

cnunciato:
This change adds a Brand page, linked from the footer, with download links for logos and brand color reference.

We should probably have some sort of a usage-guidelines section on this page as well, but I don't believe we have publicly useful content for such a thing yet, so we could probably circle back and add it later, once we do.

Fixes #2188.


content/brand/_index.md
```diff
@@ -0,0 +1,62 @@
+---
+title: Pulumi Brand
```
zchase: NIT: I believe this will render to `Pulumi Brand | Pulumi`, which is a little redundant. If we just use Brand, then we `Brand | Pulumi` which is a bit cleaner IMO.

layouts/page/brand.html
```diff
@@ -0,0 +1,84 @@
...
+        <div class="container mx-auto">
+            <ul>
+                {{ range .Params.sections }}
+                    <li><a href="#{{ .anchor }}" class="uppercase">{{ .label }}</a></li>
+                {{ end }}
+            </ul>
+        </div>
+    </nav>
+
+    <section id="logo" class="mt-8 px-4 md:px-0">
```
zchase: I think this needs to be `id="logos"`. The anchor tag in the front matter has a value of `logos` so the menu is not currently working.

PR #2406, opened 2020-02-07
Add some pre-deployment sanity checks

cnunciato:
This change adds a couple of quick checks to the deployment script before it runs s3 sync, just to ensure we don't do anything too destructive by accident. It looks first for an index.html file at the docroot, then for at least a thousand HTML files in the archive payload, exiting if either of those checks fails.

### DIFF
--- a/infrastructure/sync.sh
+++ b/infrastructure/sync.sh
@@ -22,14 +22,13 @@ aws s3 cp $site_archive .
 mkdir site_contents
 tar -xzvf $(basename "$site_archive") -C site_contents
 
-# Perform a couple of sanity checks to make sure the unpacked archive looks like a proper
-# website docroot (e.g., one composed of at least a thousand index.html files, including
-# one at the root).
-if [ -z "$(ls -al site_contents/index.html)" ]; then
+# Verify index.html exists in the root.
+if [ ! -f "site_contents/index.html" ]; then
     echo "Missing root index.html. See the archive extraction log above for details."
     exit 1
 fi
 
+# Verify we have at least 1000 index.html files in total across the site.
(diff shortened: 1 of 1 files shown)


infrastructure/sync.sh
```diff
@@ -6,12 +6,35 @@ site_archive="$1"
...
 # Download the archive from the archive bucket and unpack it into a local folder.
 echo "Downloading $site_archive..."
 aws s3 cp $site_archive .
 mkdir site_contents
 tar -xzvf $(basename "$site_archive") -C site_contents
 
+# Perform a couple of sanity checks to make sure the unpacked archive looks like a proper
+# website docroot (e.g., one composed of at least a thousand index.html files, including
+# one at the root).
+if [ -z "$(ls -al site_contents/index.html)" ]; then
```
chrsmith: Rather than checking for the output of `ls` (which, IDK, maybe could just print out with "file not found"), wouldn't it make more sense to use the `-f` flag here? Which I believe texts specifically if a file exists or not?

`if [ ! -f "./site_contents/index.html" ]; then` seems like a better approach if I have that right.

infrastructure/sync.sh
```diff
@@ -6,12 +6,35 @@ site_archive="$1"
...
+    echo "Missing root index.html. See the archive extraction log above for details."
+    exit 1
+fi
+
```
chrsmith: Pedantic and annoying nit: I'd simplify the comment on line 25, to just be "Verify index.html exists in the root" and then "Verify we have at least 1000 index.html files in total across the site". (As a suggestion to just keep the comments closer to the specific thing they do. Not that this code is difficult to follow or anything.)

PR #2414, opened 2020-02-09
Update _index.md

Ray-Gill:
### Proposed changes

Update ChartOpts link to corresponding v3 link from v2


justinvp:
Thanks for the PR, @Ray-Gill! This file is actually auto-generated from the Python package, so your changes will unfortunately be overwritten the next time we regenerate the docs. I'm guessing the reason it is mentioning `v2` is because `v3` is an alias for `v2` per https://github.com/pulumi/pulumi-kubernetes/pull/970

cc: @lblackstone

PR #2419, opened 2020-02-10
UX improvements to events pages

zchase:
This PR moves the filter to the right hand side of screen, as to expand the event listing so it feels less cramped. This also adds a heading to the top of the list showing how many events are available when filtering.

@cnunciato Let's use this PR to get in any UX improvements.


assets/js/event-filtering.js
```diff
@@ -41,40 +72,26 @@ $(function() {
...
-            const input = createCheckbox(text);
-            eventFilterParent.append(input);
-        }
-    }
-
-    // This click handler will determine which checkboxes are selected
-    // and then provide them to the filter event function to filter the events.
-    $("#eventFilter input[type='checkbox']").on("click", function() {
+    // This function grabs the filter checkboxes, loops through them to
+    // determin what is checked, and then filters the events.
```
justinvp: Nit:
```suggestion
    // determine what is checked, and then filters the events.
```

assets/js/event-filtering.js
```diff
@@ -1,18 +1,35 @@
 $(function() {
-    var eventsElements = $(".event-tags span");
-    var eventFilterParent = $("#eventFilter");
+    var queryParams = window.location;
```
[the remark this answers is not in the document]
zchase: Nope, looks it was leftover from some early work I did. Removing.

assets/js/event-filtering.js
```diff
@@ -1,18 +1,35 @@
...
-    if (!eventFilterParent.length) {
-        // Stop the script because the event filter element does not exist.
-        return;
-    }
+    /**
+     * This function returns an object of the query params key/value
+     * from the current URL.
+     *
+     */
+    function getQueryParams() {
```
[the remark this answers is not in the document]
zchase: Ah thanks, I will rework to use the shared function.

assets/js/event-filtering.js
zchase: Fixed

assets/js/event-filtering.js
```diff
@@ -1,37 +1,36 @@
...
-        checkbox.className = "mx-2 cursor-pointer";
-        checkbox.value = text;
+        checkbox.className = "mr-2 cursor-pointer";
+        checkbox.value = text.toLowerCase();
+
+        // If the filter query parameter is available use it to determine
+        // what checkboxes should initially be checked. If the query param
+        // is not set, check all the boxes.
+        if (filterValue !== undefined) {
+            var shouldBeChecked = filterValue.split(",").indexOf(checkbox.value) > -1;
```
justinvp: Nit: `getQueryVariable` doesn't lowercase the value, which you were doing previously, so if you want to maintain that behavior, you'll have to do something like:

```suggestion
            var shouldBeChecked = filterValue.toLowerCase().split(",").indexOf(checkbox.value) > -1;
```

zchase:
@justinvp The `filter` param will be used for external links. I'm not exactly certain on the details, this was a request from Isaac (who isn't on GitHub?). Interesting point on having the URL manage the "state" of the filter, I hadn't thought of that. Might be something worth considering as this page gets more events and becomes more crowded.

PR #2424, opened 2020-02-11
Add the Keycloak documentation

stack72:


content/docs/reference/pkg/nodejs/pulumi/keycloak/Ldap/_index.md
```diff
@@ -0,0 +1,2112 @@
...
+<h2 id="resources">Resources</h2>
+<h3 class="pdoc-module-header" id="FullNameMapper" data-link-title="FullNameMapper">
+    <a href="https://github.com/pulumi/pulumi-keycloak/blob/{{< param git_sha >}}/sdk/nodejs/Ldap/fullNameMapper.ts#L60">
+        Resource <strong>FullNameMapper</strong>
+    </a>
+</h3>
+
+<pre class="highlight"><code><span class='kr'>class</span> <span class='nx'>FullNameMapper</span> <span class='kr'>extends</span> <a href='/docs/reference/pkg/nodejs/pulumi/pulumi/#CustomResource'>CustomResource</a></code></pre>
+
+#### # keycloak.Ldap.FullNameMapper
```
justinvp: This is probably going to look odd. Seems like another case where we could improve the doc extraction.

stack72: 👍 I don't think many eyes will be on this set of docs for a while - so we have time :)

PR #2437, opened 2020-02-12
Apply cache-control settings for asset bundles

### LINKED ISSUE #2436: Set client-side cache headers for bundled assets
We cache our CSS and JS bundles at the CDN level, but don't explicitly set client-side cache headers to instruct browsers how to long to hold onto them. (Instead, we let browsers use their default settings -- so _some_ caching happens, we just don't dictate how.)

We can cache these bundles for a good while (e.g., at least a month), since we cache them all at the edges for a year.

cnunciato:
This change sets cache-control headers for our asset bundles to instruct browsers to cache them for one year, which matches our CDN distribution settings.

image

Fixes #2436.

### DIFF
--- a/infrastructure/sync.sh
+++ b/infrastructure/sync.sh
@@ -34,15 +34,16 @@ if [ ! "$(find site_contents -type f | grep index.html | wc -l)" -ge 1000 ]; the
     exit 1
 fi
 
-# Upload the JS and CSS bundles first.
+# Upload the JS and CSS bundles first, applying cache-control headers that match our
+# CloudFront settings (to cache these objects for one year).
 cd site_contents
 
 for file in $(find css -name "styles.*.css") ; do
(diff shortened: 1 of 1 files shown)


zchase:
@cnunciato Any chance we could do this same thing for the images in the static directory?

cnunciato:
@zchase Technically yes, but I'd probably suggest we do that as a separate PR, because it's a little trickier for a couple of reasons -- one being that we currently upload the whole site with `s3 sync`, which doesn't give you file or "folder" level control over this sort of thing (which is why we do these two files individually; if we ran `s3 cp` for every image, the build would go back to taking forever again). another is that we probably wouldn't want to cache all static images quite so aggressively, since they do sometimes change. 

So in short, yes, but we'd probably want to think about it a bit more first.

PR #2441, opened 2020-02-13
Intro to AWS step functions

spara:
Builds on the python step function example and explains how to add another step using the Pulumi programming model


content/blog/intro-to-step-functions/index.md
```diff
@@ -0,0 +1,178 @@
+---
+title: "Intro to AWS Serverless Step Functions"
+date: 2020-02-13
+meta_desc: "An Introduction to building serverless apps with AWS Step Functions."
+meta_image: meta-step-function.png
+authors:
+    - sophia-parafina
+tags:
+    - serverless
+    - "step functions"
```
chrsmith: Not 100% sure if this needs to be "step-functions" or not.

On a related not, it looks like we have `f`, `c`, and `cd` as blog tags. Which I assume are from other posts where the tags didn't get captured/interpreted correctly. Would you mind fixing that? Maybe we need to add some checks in the Hugo build to report an error if we have unexpected tags in the future?

content/blog/intro-to-step-functions/index.md
```diff
@@ -0,0 +1,178 @@
...
+
+Let’s start with a hello world function in an AWS Lambda. You can start the [example from Github](https://github.com/pulumi/examples/tree/master/aws-py-stepfunctions") by clicking on the `Deploy with Pulumi` button. A new project page is created automatically in Pulumi console, click on `Create Project` to get started. Follow the steps to copy the project locally. As with any Python project, we recommend creating a virtual environment to store project-specific dependencies.
+
+```bash
+$ python3 -m venv venv
+$ source venv/bin/activate
+$ pip3 install -r requirements.txt
+```
+
+Before we run the example, let’s look at the code. A step function is made up of a lambda function and a state machine that runs the function. Both the lambda and the state machine require an [IAM execution role](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles.html) and [policy](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies.html) to access AWS services and resources. The `iam.py` module establishes the roles and policies and the AWS region, which are imported in `__main__,py`.
```
chrsmith: The mention of the IAM permissions says what we need to do, but not _why_. Perhaps we want to take a step back and phrase this like:

> We are defining a function that will execute some code. So first we need to define what permissions that code should run as. In AWS, that is determined by an [IAM execution role]. Similarly, we need to figure out who can actually trigger/invoke our Lambda. That's done by an [IAM access policy].

Nit: Do we want it to be `[IAM access policy]` as well, to more closely match `[IAM execution role]` earlier in the paragraph?

content/blog/intro-to-step-functions/index.md
```diff
@@ -0,0 +1,178 @@
...
+
+The lambda returns “Hello world!”  when called.  Note that the handler is the name of the module, `hello`, and the name of the function, `handler`.
+
+```python
+hello_world_fn = lambda_.Function('helloWorldFunction',
+   role=iam.lambda_role.arn,
+   runtime="python3.7",
+   handler="hello.handler",
+   code=pulumi.AssetArchive({
+       '.': pulumi.FileArchive('./step_hello')
```
chrsmith: Is this right? The file is named `step_hello`? Should it be `step_hello.py` or something?

content/blog/intro-to-step-functions/index.md
```diff
@@ -0,0 +1,178 @@
...
+    + 6 to create
+    1 unchanged
+
+Do you want to perform this update?
+  yes
+> no
+  details
+```
+
+Select `yes` to run the example, but note that the function doesn’t specify where to return `Hello world!`. We can execute the function from the AWS cli,  but to see how it works, we’ll need to launch the  [AWS Conslole](https://console.aws.amazon.com/states/home#/statemachines/).
```
chrsmith: Nit: AWS "Console"

content/blog/intro-to-step-functions/index.md
```diff
@@ -0,0 +1,178 @@
...
+
+```
+
+We will also need to import `Output` from pulumi along with the other imports. We’ll use it later.
+
+```python
+from pulumi import Output
+```
+
+Next, add the new lambda function. Note that we can reuse the same IAM role and policy we use in the `hello_world_fn` function.
```
chrsmith: Nit: Does it make sense to mention the name? e.g. "... new lambda function, `nice_fn`. Note that we can ..."

content/blog/intro-to-step-functions/index.md
```diff
@@ -0,0 +1,178 @@
...
+meta_desc: "An Introduction to building serverless apps with AWS Step Functions."
+meta_image: meta-step-function.png
+authors:
+    - sophia-parafina
+tags:
+    - serverless
+    - "step functions"
+---
+
+AWS Step Functions let you build applications by connecting AWS services. We'll go into the details of creating a lambda function, IAM roles and policies, and creating a workflow. Once we have the example deployed, we'll walk through the process of adding another function and step to the workflow. Included in the walkthrough is a discussion of one of the aspects of the Pulumi programming model. The goal of this article is to provide a foundation for building your application using serverless workflows.
```
chrsmith: After reading the post I'm still not sure what a "Step function" is. Is it a separate AWS product offering specifically for making it easier to build/manage these workflows? And we use AWS Lambda as a simple example of a "step"?

It would be nice if this was called out a little more explicitly. Since later in the text we just have "The state machine controls the workflow.", and IMHO we could make it a little clearer by elaborating on what we mean by "workflow" and how later, we'll write code to manage the progress through that workflow using a "state machine".

content/blog/intro-to-step-functions/index.md
spara: AWS has it as 'Step Functions' so I went with that.

We've discussed cleaning up the tags, but are afraid that it will break links. It's on the radar but on the back burner.

content/blog/intro-to-step-functions/index.md
spara: It's pointing to the directory

content/blog/intro-to-step-functions/index.md
spara: fixed

content/blog/intro-to-step-functions/index.md
spara: added

content/blog/intro-to-step-functions/index.md
spara: Reworked as per suggestion.

content/blog/intro-to-step-functions/index.md
spara: I updated the intro paragraph:

[AWS Step Functions](https://aws.amazon.com/step-functions/) lets you build applications by connecting AWS services. Daisy chaining steps into a workflow simplifies application development by creating a state machine diagram which shows how services are connected to each other in your application. We'll go into the details of creating a lambda function, IAM roles and policies, and creating a workflow. Once we have the example deployed, we'll walk through the process of adding another function and step to the workflow. Included in the walkthrough is a discussion of one of the aspects of the Pulumi programming model. The goal of this article is to provide a foundation for building your application using serverless workflows.

content/blog/intro-to-step-functions/index.md
```diff
@@ -0,0 +1,178 @@
...
+
+Then check the AWS console to see if the lambda was executed.
+
+![AWS Severless Console](sf-console-1.png)
+
+Clicking on the `Name` displays the step details, which include the input and out, i.e., ‘Hello world!’
+
+![AWS Severless Console  with 1 step function](sf-console-2.png)
+
+The AWS console provides a graph of your workflow and the details of each step. Clicking on `Output` shows “Hello world” meaning that that the call to the lambda function was successful.
```
EvanBoyle: nit: duplicate that in `that that the call`

content/blog/intro-to-step-functions/index.md
chrsmith: At least for me, that's a lot clearer in framing what the post is all about. Thanks.

Issue #2378, opened 2020-02-05
Improve mobile performance


joeduffy:
I didn't seen an issue already tracking this, so filing to make sure we don't lose track. We currently have relatively poor performance on mobile, [as scored by Google](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fpulumi.com%2Fterraform%2F&tab=mobile):

![image](https://user-images.githubusercontent.com/3953235/73806700-12336c00-477f-11ea-8af5-27f7121728ad.png)

This is hurting our SEO and ads performance.

Some specific ideas from @zchase to help move the needle here are:

* Lazy load images
* Take advantage of hugo's image resize api to properly size images when served on mobile
* Figure out what to do with the unused css issue. We use tailwindcss library and I am assuming we have to load the entire library on each page

I'm assigning to both @cnunciato and @zchase since I'm unsure who technically owns this.

cnunciato:
Also note HTML minification was just merged (with https://github.com/pulumi/docs/pull/2456), which reduces page sizes by ~20%.

zchase:
PurgeCSS work has landed (https://github.com/pulumi/docs/pull/2444). We saw a small increase in page speed but not the same increase as I was seeing locally. I will re-run the page speed test tomorrow AM and update the issue with the new metrics.

zchase:
Updated page speed score after @cnunciato & my changes. Progress 💪 

<img width="1675" alt="pagespeed" src="https://user-images.githubusercontent.com/15146337/74857532-a6154400-52f8-11ea-8053-01e499d23747.png">

lukehoban:
FWIW - this appears to have gone back down again.

<img width="1145" alt="Screen Shot 2020-11-13 at 10 15 48 PM" src="https://user-images.githubusercontent.com/223467/99141198-cced7600-25fd-11eb-90e0-734d0a3b8ba5.png">

susanev:
slightly up again
<img width="829" alt="performance" src="https://user-images.githubusercontent.com/5489125/155400608-c8951c0a-93d6-418f-8b27-12f6f50a8d1e.png">

susanev:
@kimberleyamackenzie would you have interest in investigating performance on mobile?

PR #2444, opened 2020-02-13
Add PurgeCSS to remove unused CSS classes from page loads

zchase:
This PR adds PurgeCSS as a plugin to our already existing PostCSS process. PurgeCSS removed unused CSS classes, which cuts the CSS file load in about half. A lot of third party classes and some of our custom CSS are excluded by default so I had to add custom whitelisting rules to include them.

Relevant issue: https://github.com/pulumi/docs/issues/2378

### DIFF
--- a/assets/config/postcss.config.js
+++ b/assets/config/postcss.config.js
@@ -21,9 +21,9 @@ module.exports = {
             // Specify the paths to all of the template files in your project
             content: [ "./layouts/**/*.html" ],
             // Whitelist HubSpot specific classes so they don't get removed.
-            whitelist: ["supported-cicd-platforms"],
-            whitelistPatterns: [/^hs-/, /^highlight$/],
-            whitelistPatternsChildren: [/^hs-/, /^highlight$/],
+            whitelist: ["supported-cicd-platforms", ":not"],
+            whitelistPatterns: [/^fa-/, /^hs-/, /^highlight$/, /^pagination$/, /^code-/,  /^copy-/],
+            whitelistPatternsChildren: [/^hs-/, /^highlight$/, /^pagination$/, /^code-/,  /^copy-/],
             // Extract the default tailwind classes.
             defaultExtractor: content => content.match(/[\w-/:]+(?<!:)/g) || [],
         }),
--- a/assets/config/postcss.config.js
+++ b/assets/config/postcss.config.js
@@ -20,7 +20,19 @@ module.exports = {
         // Docs: https://purgecss.com/plugins/postcss.html
         require("@fullhuman/postcss-purgecss")({
             // Specify the paths to all of the template files in your project
-            content: [ "./layouts/**/*.html" ],
+            content: [
+                "./layouts/**/*.html",
+                // Some of our scripts reference CSS classes.
+                "./assets/js/**/*.js",
+                // Look for CSS classes in our markdown content. We use `glob` explicitly here so we can ignore the
+                // files for the API docs in ./content/docs/reference/pkg/**/* because it includes a large number of
+                // files that significantly impacts build time (~25 seconds vs. many minutes). Instead, we'll only look
+                // for CSS classes in a single package (the Pulumi SDK package) for nodejs and python, which should
+                // include all classes used in the docs for any other package.
+                ...require("glob").sync("./content/**/*.md", { ignore: "./content/docs/reference/pkg/**/*" }),
+                "./content/docs/reference/pkg/nodejs/pulumi/pulumi/**/*.md",
+                "./content/docs/reference/pkg/python/pulumi/**/*.md",
+            ],
             // Whitelist HubSpot specific classes so they don't get removed.
             whitelist: ["supported-cicd-platforms", ":not", "md:max-w-lg", "blink", "typing", "char"],
             whitelistPatterns: [/^fa-/, /^hs-/, /^highlight$/, /^pagination$/, /^code-/, /^copy-/, /^carousel/],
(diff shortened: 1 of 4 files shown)


cnunciato:
Nice! Seeing a couple things after running a `make build`, but it's looking like it's mostly there!

* The home page animation doesn't seem to be working now.
* There's a minor but noteworthy oddity with the cloud-provider buttons on the Getting Started home page that we should make sure didn't clobber something we might be using elsewhere, too:

assets/config/postcss.config.js
```diff
@@ -16,6 +16,18 @@ module.exports = {
...
 
+        // PurgeCSS to remove unused classes for better page speed.
+        require("@fullhuman/postcss-purgecss")({
+            // Specify the paths to all of the template files in your project
+            content: [ "./layouts/**/*.html" ],
+            // Whitelist HubSpot specific classes so they don't get removed.
+            whitelist: ["supported-cicd-platforms", ":not"],
+            whitelistPatterns: [/^fa-/, /^hs-/, /^highlight$/, /^pagination$/, /^code-/,  /^copy-/],
+            whitelistPatternsChildren: [/^hs-/, /^highlight$/, /^pagination$/, /^code-/,  /^copy-/],
+            // Extract the default tailwind classes.
```
cnunciato:
```suggestion
            // Extract the default Tailwind classes.
```

assets/config/postcss.config.js
```diff
@@ -16,6 +16,18 @@ module.exports = {
...
+            defaultExtractor: content => content.match(/[\w-/:]+(?<!:)/g) || [],
```
cnunciato: Could you add a more descriptive comment explaining what this regex is doing?

assets/config/postcss.config.js
```diff
@@ -16,6 +16,18 @@ module.exports = {
...
         }),
 
+        // PurgeCSS to remove unused classes for better page speed.
+        require("@fullhuman/postcss-purgecss")({
+            // Specify the paths to all of the template files in your project
+            content: [ "./layouts/**/*.html" ],
+            // Whitelist HubSpot specific classes so they don't get removed.
+            whitelist: ["supported-cicd-platforms", ":not"],
+            whitelistPatterns: [/^fa-/, /^hs-/, /^highlight$/, /^pagination$/, /^code-/,  /^copy-/],
+            whitelistPatternsChildren: [/^hs-/, /^highlight$/, /^pagination$/, /^code-/,  /^copy-/],
```
cnunciato:
```suggestion
            whitelistPatternsChildren: [/^hs-/, /^highlight$/, /^pagination$/, /^code-/, /^copy-/],
```

assets/config/postcss.config.js
```diff
@@ -16,6 +16,18 @@ module.exports = {
             ]
         }),
 
+        // PurgeCSS to remove unused classes for better page speed.
+        require("@fullhuman/postcss-purgecss")({
+            // Specify the paths to all of the template files in your project
+            content: [ "./layouts/**/*.html" ],
+            // Whitelist HubSpot specific classes so they don't get removed.
+            whitelist: ["supported-cicd-platforms", ":not"],
+            whitelistPatterns: [/^fa-/, /^hs-/, /^highlight$/, /^pagination$/, /^code-/,  /^copy-/],
```
cnunciato:
```suggestion
            whitelistPatterns: [/^fa-/, /^hs-/, /^highlight$/, /^pagination$/, /^code-/, /^copy-/],
```

assets/config/postcss.config.js
```diff
@@ -16,6 +16,18 @@ module.exports = {
             ]
         }),
 
+        // PurgeCSS to remove unused classes for better page speed.
+        require("@fullhuman/postcss-purgecss")({
```
cnunciato: It might be good to drop a link to the documentation for this config just as a quick reference.

assets/config/postcss.config.js
```diff
@@ -16,6 +16,18 @@ module.exports = {
...
+            // Specify the paths to all of the template files in your project
+            content: [ "./layouts/**/*.html" ],
```
justinvp: We also use selectors in some of our scripts, such as `search.js` to add HTML dynamically, and I think a couple other scripts. Might be worth just including them:

```javascript
// Some of our scripts reference CSS classes.
"./assets/js/**/*.js",
```

assets/config/postcss.config.js
zchase: 👍

assets/config/postcss.config.js
zchase: 👍

assets/config/postcss.config.js
zchase: 👍

assets/config/postcss.config.js
```diff
@@ -16,6 +16,34 @@ module.exports = {
...
+                // Look for CSS classes in our markdown content. We use `glob` explicitly here so we can ignore the
+                // files for the API docs in ./content/docs/reference/pkg/**/* because it includes a large number of
+                // files that significantly impacts build time (~25 seconds vs. many minutes). Instead, we'll only look
+                // for CSS classes in a single package (the Pulumi SDK package) for nodejs and python, which should
+                // include all classes used in the docs for any other package.
+                ...require("glob").sync("./content/**/*.md", { ignore: "./content/docs/reference/pkg/**/*" }),
+                "./content/docs/reference/pkg/nodejs/pulumi/pulumi/**/*.md",
+                "./content/docs/reference/pkg/python/pulumi/**/*.md",
+            ],
+            // Whitelist HubSpot specific classes so they don't get removed.
```
justinvp: This is more than just HubSpot now, right?

justinvp:
> I haven't spotted anything obviously broken clicking around the site locally (`make serve`).

Ahh, actually... I was looking at the wrong branch. Let me double check again.

assets/config/postcss.config.js
zchase: Yup, good catch. I updated the comments!

Issue #2416, opened 2020-02-10
Expose the events page via header and footer navigation links


zchase:
Now that we have an events page, https://github.com/pulumi/docs/pull/2318, we will need to expose the page via our header and footer navigation. Here are the following items that need to happen before adding to the header/footer:

- [x] Ensure all events are listed
- [x] Final UI/UX sweep to clean things up

zchase:
PR: https://github.com/pulumi/docs/pull/2454

zchase:
PR has been merged in. We held off adding an "Events" link to the top nav because in our opinion it made the nav a bit too crowded. We will revisit how to add the events link to the top nav.

PR #2454, opened 2020-02-18
Add events page links to header/footer navs

zchase:
As the title says, this PR adds links to the event list page in the header and footer.

https://github.com/pulumi/docs/issues/2416

### DIFF
--- a/content/events/_index.md
+++ b/content/events/_index.md
@@ -1,9 +1,14 @@
 ---
-title: Upcoming Events
+title: Events
 type: events
 layout: events
 description: Talks, workshops, and other gatherings to help you connect with the Pulumi community.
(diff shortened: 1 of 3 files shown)


cnunciato:
We do need an easy way to find the events listing, but I'm not sure these links really belong where they are here right now. 🤔 

A footer link seems fine for the time being just to get the job done, but I wouldn't mind if we thought about this a bit more and came up with something a little more integrated. On other projects I've seen Community pages used for things like this (to collect events, links to Slack, etc.), and maybe that could work for us, too. I would suggest that we remove the Events link from the main nav for now, though, just 'cause it doesn't quite line up with the others in terms of top-level relevance/importance.

PR #2456, opened 2020-02-18
Pass the minify option to the Hugo build

### LINKED ISSUE #2344: Fix Pulumify
Pulumify is now consistently failing due to `No space left on device`. We should disable Pulumify until we have a fix/workaround available.

- [x] Temporarily disable Pulumify (#2372)
- [ ] Fix Pulumify and re-enable

Simple fixes may be:

 - Leverage lambda EFS if/when we have access
 - Go back to syncing contents from the build machine, rather than inside a lambda (though time-to-pulumify-preview will probably be more like 10 mins than 2)

cnunciato:
This change adds the `--minify` option to the Hugo build command, which trims superfluous whitespace and markup (using https://github.com/tdewolff/minify), reducing the size of the build by about 20% (and individual page sizes by about the same): 

```
$du -sh public_master public_minified
544M	public_master
433M	public_minified
```

This also lets us re-enable Pulumify, since the resulting build is now well below the 512MB limit.

We should do some manual browser testing to make sure this doesn't break anything important, and cc @zchase as well, in case this could affect anything SEO-related.

Fixes #2344 (if not necessarily permanently).

### DIFF
--- a/.github/workflows/pulumify.yml
+++ b/.github/workflows/pulumify.yml
@@ -1,0 +1,16 @@
+name: Pulumify
+on: [pull_request, delete]
+jobs:
+  updateLivePreview:
+    name: Update Live Preview
+    runs-on: ubuntu-latest
+    steps:
+    - uses: docker://pulumi/pulumify:v0.1.2
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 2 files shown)


zchase:
LGTM from the SEO side.

cnunciato:
Not seeing any issues with manual browser testing.

PR #2464, opened 2020-02-19
OLD: Add terraform ppc landing page

zchase:
This PR adds the framework and first ppc landing page for our site. You will notice the header and footer have been hidden for these pages. We want to keep any who clicks on an ad focused on the goal of the page. Including additional navigation can confuse users and lead them to become distracted. The other important thing to note are the logos and customer quotes. They provide social proof for us, which is helpful when trying to convert users. They can see that actual companies use our product.


zchase:
Duplicate of https://github.com/pulumi/docs/pull/2667. Had to create a new PR to fix Pulumify. Closing

PR #2476, opened 2020-02-19
Use definite assignment assertions with dynamic provider example

justinvp:
As of TypeScript 2.7, [strict property initialization](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#strict-class-initialization) is enabled by default (typically), which means TypeScript won't realize properties of dynamic resources will be initialized. We use a definite assignment assertion to relay to TypeScript that these will indeed be initialized.

### DIFF
--- a/content/docs/intro/concepts/programming-model.md
+++ b/content/docs/intro/concepts/programming-model.md
@@ -2647,8 +2647,8 @@ class MyResourceProvider extends pulumi.dynamic.ResourceProvider {
 }
 
 export class MyResource extends pulumi.dynamic.Resource {
-    public readonly myStringOutput: pulumi.Output<string>;
-    public readonly myNumberOutput: pulumi.Output<number>;
+    public readonly myStringOutput!: pulumi.Output<string>;
+    public readonly myNumberOutput!: pulumi.Output<number>;
 
     constructor(name: string, props: MyResourceInputs, opts?: pulumi.CustomResourceOptions) {
         super(myprovider, name, { myStringOutput: undefined, myNumberOutput: undefined, ...props }, opts);


content/docs/intro/concepts/programming-model.md
```diff
@@ -2624,7 +2624,7 @@ Any outputs can be returned by your `create` function in the `outs` property of
...
+If you need to access the outputs of your custom resource outside it with strong typing support, declare each output property returned in the `outs` property by your `create` function as a class member of the `pulumi.dynamic.Resource` itself. For example, in TypeScript, these must be declared as `public readonly` class members in your `pulumi.dynamic.Resource` class. These class members must also have the type `pulumi.Output<T>`. In TypeScript, if `--strictPropertyInitialization` is enabled (it typically is enabled by default along with other `--strict` mode flags), [definite assignment assertion modifiers](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#strict-class-initialization) (an `!` exclamation point) must be used on each class member to relay to TypeScript that these will be initialized indirectly.
```
praneetloke: What do you think of this?

```suggestion
If you need to access the outputs of your custom resource outside it with strong typing support, declare each output property returned in the `outs` property by your `create` function as a class member of the `pulumi.dynamic.Resource` itself.

For example, in TypeScript, these must be declared as `public readonly` class members in your `pulumi.dynamic.Resource` class. These class members must also have the type `pulumi.Output<T>`.

> If `--strictPropertyInitialization` is enabled (it typically is enabled by default along with other `--strict` mode flags), [definite assignment assertion modifiers](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#strict-class-initialization) (an `!` exclamation point) must be used on each class member to relay to TypeScript that these will be initialized indirectly.
```

justinvp: I considered doing it like this, but then the blockquote gets combined with the one directly below it, which doesn't look quite right:

Screen Shot 2020-02-19 at 2 02 22 PM

praneetloke: Ah gotcha. Could we then remove the blockquote for the "Note"?

justinvp: How about this: I moved the existing note into the first paragraph, and added this new blockquote as it's own note. What do you think?

Screen Shot 2020-02-19 at 2 32 36 PM

content/docs/intro/concepts/programming-model.md
```diff
@@ -2624,9 +2624,11 @@ Any outputs can be returned by your `create` function in the `outs` property of
...
+If you need to access the outputs of your custom resource outside it with strong typing support, declare each output property returned in the `outs` property by your `create` function as a class member of the `pulumi.dynamic.Resource` itself. The name of the class member must match the names of the output properties as returned by the `create` function.
 
-> **Note:** The name of the class member must match the names of the output properties as returned by the `create` function.
+For example, in TypeScript, these must be declared as `public readonly` class members in your `pulumi.dynamic.Resource` class. These class members must also have the type `pulumi.Output<T>`.
+
+> **Note:** If `--strictPropertyInitialization` is enabled (it typically is enabled by default along with other `--strict` mode flags), [definite assignment assertion modifiers](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#strict-class-initialization) (an `!` exclamation point) must be used on each class member to relay to TypeScript that these will be initialized indirectly.
```
lukehoban: Do you think we need this?  I worry about having anything overly TypeScript specific in the prose - as it distracts from others.  If we have the code sample show the right thing - is that enough?

justinvp: OK, I'll cut it. We already have a lot of overly TypeScript specific prose, so erring on the side of not adding more, and reducing it in general, sounds good to me.

content/docs/intro/concepts/programming-model.md
praneetloke: That looks good! Thank you, @justinvp.

Issue #2036, opened 2019-11-22
Intended service reference text incorrectly omitted from rendered text description 


danvanderboom:
### Problem description

- I saw a product documentation issue on the following URL:
https://www.pulumi.com/blog/managing-secrets-with-pulumi/

“This means when storing state with , we use a key managed by the service, specific to your stack”

The missing text, in two places, is a reference to the Pulumi backend service. It appears correctly in the github.com view, but not as rendered by Safari on iOS (at least). 

### Suggestions for a fix

Format the service url differently. Test in Safari on iOS.

leezen:
Thanks for pointing this out! I've submitted a PR to address this.

PR #2479, opened 2020-02-19
Fix app.pulumi.com link in the managing secrets blog post

leezen:
https://github.com/pulumi/docs/issues/2036

### DIFF
--- a/content/blog/managing-secrets-with-pulumi/index.md
+++ b/content/blog/managing-secrets-with-pulumi/index.md
@@ -204,7 +204,7 @@ passwords or access credentials.
 You might be wondering how these values are actually encrypted. We use
 the same encryption that we have always used for our configuration
 system. This means when storing state with
-&lt;app.pulumi.com&gt;, we use a key managed by the &lt;app.pulumi.com&gt; service, specific to your stack, to
+<https://app.pulumi.com>, we use a key managed by the <https://app.pulumi.com> service, specific to your stack, to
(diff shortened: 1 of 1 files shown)


content/blog/managing-secrets-with-pulumi/index.md
```diff
@@ -204,7 +204,7 @@ passwords or access credentials.
 You might be wondering how these values are actually encrypted. We use
 the same encryption that we have always used for our configuration
 system. This means when storing state with
-<app.pulumi.com>, we use a key managed by the <app.pulumi.com> service, specific to your stack, to
+&lt;app.pulumi.com&gt;, we use a key managed by the &lt;app.pulumi.com&gt; service, specific to your stack, to
```
justinvp: I'm guessing the intention was for these to be [automatically linked](https://github.com/russross/blackfriday/blob/41c5fccfd6f60cb5fd077118835b1d7122498df6/testdata/Markdown%20Documentation%20-%20Syntax.text#L833-L839), but perhaps the `https://` prefix is needed for that to happen?

```suggestion
<https://app.pulumi.com>, we use a key managed by the <https://app.pulumi.com> service, specific to your stack, to
```

Wondering if we make the above suggestion, if we'd see these as links as intended in the Pulumify preview.

leezen: Ah, that makes more sense in terms of the original intent. Sure, let me swap to that.

Issue #2311, opened 2020-01-27
GA Go SDK documentation sweep


EvanBoyle:
We should make sure our docs are up to date and polished for the updated Go programming model:

- [x] Getting Started
- [x] API docs (https://github.com/pulumi/docs/pull/2536)
- [x] Language page
- [x] Programming model page (https://github.com/pulumi/docs/pull/2529)
- [ ] Tutorials
- [x] All `lang-choose` elements across the site

We should ensure we have done the following:

1. Remove preview labels
2. Fill in embedded examples for currently unsupported features.

EvanBoyle:
@lukehoban looking at the tutorials page, most of our content is exclusively in node. Is adding a go choose to all of these pages in scope here? We could consider having something organized by language instead if we're not going to aim for 100% coverage. https://www.pulumi.com/docs/tutorials/

EvanBoyle:
We decided that tutorials need a major overhaul as they currently aren't structured to be multi-lang. That work is out of scope for this issue.

PR #2480, opened 2020-02-19
update go aws getting started docs

EvanBoyle:
Ran through the getting started and found some issues. With these changes, users should be able to successfully run through the getting started in go. Fixes:

1. Add installation of `dep` to the setup guide. Hopefully, this goes away if we can find a solution to https://github.com/pulumi/pulumi/issues/3817
2. Make sure we run `dep ensure` any time we add a new dependency. This really shouldn't be necessary and is a crappy user experience. During the 'modify program` step, if you don't run dep ensure after adding the KMS import you get an error: 
```
main.go:4:2: cannot find package "github.com/pulumi/pulumi-aws/sdk/go/aws/kms" in any of:
    	/Users/evanboyle/go/src/go_qs_2/vendor/github.com/pulumi/pulumi-aws/sdk/go/aws/kms (vendor tree)
    	/usr/local/Cellar/go/1.13.8/libexec/src/github.com/pulumi/pulumi-aws/sdk/go/aws/kms (from $GOROOT)
    	/Users/evanboyle/go/src/github.com/pulumi/pulumi-aws/sdk/go/aws/kms (from $GOPATH)
```
3. Remove any references to building artifacts. This is no longer necessary with the `go run` work in https://github.com/pulumi/pulumi/issues/3486
4. Fix the modify program code.
5. Add instructions for plugin installation

contributes to https://github.com/pulumi/docs/issues/2311

### DIFF
--- a/content/docs/get-started/aws/create-project.md
+++ b/content/docs/get-started/aws/create-project.md
@@ -40,6 +40,7 @@ $ pulumi new aws-python
 <div class="language-prologue-go"></div>
 
 ```bash
+# from within your $GOPATH
 $ mkdir quickstart && cd quickstart
 $ pulumi new aws-go
 ```
--- a/content/docs/get-started/aws/create-project.md
+++ b/content/docs/get-started/aws/create-project.md
@@ -42,7 +42,7 @@ $ pulumi new aws-python
 ```bash
 # install the pulumi aws plugin
 # check for the release version here https://github.com/pulumi/pulumi-aws/releases
-$ pulumi plugin install resource aws 1.22.0 
+$ pulumi plugin install resource aws 1.22.0
 # from within your $GOPATH
 $ mkdir quickstart && cd quickstart
 $ pulumi new aws-go
--- a/content/docs/get-started/aws/create-project.md
+++ b/content/docs/get-started/aws/create-project.md
@@ -40,6 +40,9 @@ $ pulumi new aws-python
 <div class="language-prologue-go"></div>
 
 ```bash
+# install the pulumi aws plugin
+# check for the release version here https://github.com/pulumi/pulumi-aws/releases
+$ pulumi plugin install resource aws 1.22.0 
 # from within your $GOPATH
 $ mkdir quickstart && cd quickstart
 $ pulumi new aws-go
--- a/content/docs/get-started/aws/modify-program.md
+++ b/content/docs/get-started/aws/modify-program.md
@@ -93,38 +93,38 @@ pulumi.export('bucket_name',  bucket.id)
 package main
 
 import (
-    "github.com/pulumi/pulumi-aws/sdk/go/aws/kms"
-    "github.com/pulumi/pulumi-aws/sdk/go/aws/s3"
-    "github.com/pulumi/pulumi/sdk/go/pulumi"
+	"github.com/pulumi/pulumi-aws/sdk/go/aws/kms"
+	"github.com/pulumi/pulumi-aws/sdk/go/aws/s3"
+	"github.com/pulumi/pulumi/sdk/go/pulumi"
 )
 
 func main() {
-    pulumi.Run(func(ctx *pulumi.Context) error {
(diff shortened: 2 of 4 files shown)


content/docs/get-started/aws/create-project.md
```diff
@@ -40,6 +40,10 @@ $ pulumi new aws-python
 <div class="language-prologue-go"></div>
 
 ```bash
+# install the pulumi aws plugin
+# check for the release version here https://github.com/pulumi/pulumi-aws/releases
+$ pulumi plugin install resource aws 1.22.0
```
joeduffy: Bummer that we need to hard-code this. Does `pulumi plugin install` automatically pick the latest is a version is unspecified? Or is there any way to inject the version using a template, similar to how the `{{< latest-version >}}` shortcode works for the CLI/SDK version?

content/docs/get-started/aws/modify-program.md
```diff
@@ -93,38 +93,38 @@ pulumi.export('bucket_name',  bucket.id)
 package main
 
 import (
-    "github.com/pulumi/pulumi-aws/sdk/go/aws/kms"
-    "github.com/pulumi/pulumi-aws/sdk/go/aws/s3"
-    "github.com/pulumi/pulumi/sdk/go/pulumi"
+	"github.com/pulumi/pulumi-aws/sdk/go/aws/kms"
```
joeduffy: FWIW, I think @cnunciato intentionally asked to remove the tabs, since we generally don't use them in our code snippets. But I get that in Go this is idiomatic. Just calling it out since he had commented on this previously.

content/docs/get-started/aws/modify-program.md
```diff
@@ -171,10 +171,10 @@ class Program
 Our program now creates a KMS key and enables server-side encryption on the S3 bucket using the KMS key.
 
 {{% lang go %}}
-Recompile your project so the changes are picked up:
+We'll need to run `dep ensure` to pick up the new dependencies:
 
 ```bash
-$ go build $(basename $(pwd))
+$ dep ensure
```
joeduffy: Don't we still need to build the binary?

joeduffy: Oh, nix that, nice! Hadn't seen https://github.com/pulumi/pulumi/issues/3486.

content/docs/get-started/aws/modify-program.md
cnunciato: Spaces will behave much more reliably than tabs, yeah. 🙏

pgavlin: Tabs really are the right choice here if we want these examples to be copy-pastable and idiomatically formatted (i.e. `gofmt` clean). Code samples in Go docs always use tabs (see e.g. the examples in https://golang.org/pkg/os/).

cnunciato: I do see that point, re: copy-paste, yeah. 🤔 Let's try to make it work with tabs, then -- I do suspect we're going to have formatting issues, but we can address them when the time comes.

content/docs/get-started/aws/create-project.md
EvanBoyle: I agree this is a subpar experience. Two reasons I think this is ok for now:

1. This is strictly better than what we have today, which is a dead end.
2. @pgavlin has a P1 scheduled for m32 that will make this obsolete https://github.com/pulumi/pulumi/issues/1549. So this should be a temporary change. 

With that said I think we should just ship this now and clean it up as a part of go plugin discovery.

joeduffy: This is fine -- but _please_ let's make sure to update the docs when we do this!!

EvanBoyle: 💯 I've added a comment into the issue to make sure this doesn't get lost. 

https://github.com/pulumi/pulumi/issues/1549#issuecomment-588607084

PR #2484, opened 2020-02-20
Add an Adopting Pulumi user guide

### LINKED ISSUE #1454: User Guide: Adopting Pulumi
### Problem description

We need a more conceptual doc which can cover the workflows for getting, importing, RemoteStateReference, etc. from a more "workflow" based perspective than raw facts. 

### Suggestions for a fix

YT video and notes

joeduffy:
This adds an "Adopting Pulumi" user guide which will talk through the nuts and bolts of converting to Pulumi from an existing solution, including coexisting with infrastructure, importing existing infrastructure into Pulumi, as well as conversion, across Terraform, AWS CloudFormation, Azure Resource Manager (ARM) templates, and Kubernetes and Helm YAML configuration.

This fixes #1454.


joeduffy:
@lukehoban I'm getting close to ready to check this in. The two things remaining are to finish the Kubernetes and Helm YAML guide and, afterwards, to do a general editorial pass over the content. Any early eyeballs and feedback would be great.

content/docs/guides/adopting/from_aws.md
```diff
@@ -0,0 +1,567 @@
...
+    parent: adopting
+    weight: 3
+---
+
+<img src="/logos/tech/aws_cloudformation.png" align="right" class="h-32 px-8 pb-4">
+
+If your team has already provisioned infrastructure using AWS CloudFormation, and you'd like to adopt Pulumi, you have three primary strategies you can take:
+
+* [**Coexist**](#referencing-stack-outputs) with resources provisioned by CloudFormation by referencing stack outputs.
+* [**Import**]({{< relref "import" >}}) existing resources into Pulumi in the usual way.
```
lukehoban:
```suggestion
* [**Import**]({{< relref "import" >}}) existing resources into Pulumi [in the usual way]({{< relref "import" >}}).
```

content/docs/guides/adopting/from_aws.md
```diff
@@ -0,0 +1,567 @@
...
+            return new Dictionary<string, object?>
+            {
+                { "vpcId", vpc.Id },
+            };
+        });
+    }
+}
+```
+
+After running `pulumi up`, your VPC will become under the control of Pulumi without any disruption, and you can then delete the import directives from your code. All subsequent infrastructure changes you'd like to be made can happen within Pulumi instead of AWS CloudFormation.
```
lukehoban: There's an implied, but not explicit as far as I can tell, step here of removing the cloudformation template from the `.ts` code so that it gets deleted, but the resource doesn't experience any disruption.  Might be worth showing the `pulumi up` that does the transition?  (imports the resource, then destroys the cloudformation stack).

content/docs/guides/adopting/from_azure.md
```diff
@@ -0,0 +1,703 @@
...
+// Create a blob for our own deployment.
+let blob = new azure.storage.ZipBlob("myBlob", {
+    storageAccountName: storageAccountName,
+    storageContainerName: new azure.storage.Container("myStorageContainer", {
+        storageAccountName: storageAccountName,
+        containerAccessType: "private",
+    }).name,
+    type: "block",
+    content: new pulumi.asset.FileArchive("wwwroot"),
+});
```
lukehoban: I think we're just about to remove `ZipBlob` as part of the 2.0 Azure provider (it was a hack to work around Archive/Asset issues).

Unfortunately, thinking about this example, I'm not sure there's code we can write that works both before and after those changes.  @mikhailshilkov is that right? 

https://github.com/pulumi/pulumi-azure/pull/479

content/docs/guides/adopting/from_kubernetes.md
```diff
@@ -0,0 +1,389 @@
...
+This Pulumi program uses `ConfigGroup` to read these YAML files, provision the resources inside of them, and export the resulting IP addresses:
+
+
+{{< langchoose csharp >}}
+
+```javascript
+let k8s = require("@pulumi/kubernetes");
+
+// Create resources from standard Kubernetes guestbook YAML example.
+let guestbook = new k8s.yaml.ConfigFile("guestbook", {
```
lukehoban: These examples are all still using `ConfigGroup`.  Looks like possibly this file is still WIP so will hold off on looking more closely at this one.

content/docs/guides/adopting/from_azure.md
joeduffy: Ahh, shucks. I was trying to find a simple pair of resources that would go nicely together -- similar to the CloudFormation pair of VPC/Subnet and WebServer. Does anything else come to mind?

content/docs/guides/adopting/_index.md
```diff
@@ -0,0 +1,66 @@
...
+
+Pulumi can adopt existing infrastructure so that, going forward, it is under the control of Pulumi. This is called _importing_, and it is a great way to adopt Pulumi fully without needing to perturb or recreate existing infrastructure.
+
+In the above coexistence scenarios, Pulumi simply _reads_ your existing infrastructure, but won't assume control of managing it. As such, Pulumi will never modify or delete infrastructure that is still managed outside of Pulumi. In the case of importing infrastructure, on the other hand, that won't be true. After the import process, all resources are entirely managed by Pulumi.
+
+Furthermore, Pulumi doesn't care where the infrastructure originally came from. You could have manually provisioned it in your cloud's console UI, from the CLI, using Terraform, your cloud's built-in templating mechanism, and so on. In all cases, after the import process, you'll be left with a working Pulumi program, and all subequent infrastructure updates can be made with Pulumi. You can then retire the old way of managing your infrastructure. This works even if you've lost the original scripts or templates that created the infrastructure.
+
```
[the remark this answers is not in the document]
joeduffy: :+1:

content/docs/guides/adopting/from_aws.md
joeduffy: The idea here was that because of the retain-upon-deletion annotation on the CloudFormation resource, these don't need to be split into two distinct steps. Do you think that would help with clarity here?

content/docs/guides/adopting/from_azure.md
joeduffy: I'll go ahead and merge as-is (especially as it took an hour or so to create and run all of these variants for different languages), but if you guys have a better replacement, please do let me know.

content/docs/guides/adopting/from_terraform.md
```diff
@@ -0,0 +1,303 @@
...
+
+Next, run `tf2pulumi`. It will conver the entire project whose directory you are in and print the resulting code to `stdout`. You'll probably want to redirect its output, for instance to a file named `index.ts` in the directory that contains the Pulumi project you just created:
+
+```bash
+$ tf2pulumi >my-stack/index.ts
+```
+
+This will generate a Pulumi TypeScript program in index.ts that when run with pulumi update will deploy the infrastructure originally described by the Terraform project. Note that if your infrastructure references files or directories with paths relative to the location of the Terraform project, you will most likely need to update these paths such that they are relative to the generated index.ts file.
+
+> If `tf2pulumi` complains about missing Terraform resource plugins, install those plugins as per the instructions in the error message and re-run the command above.
```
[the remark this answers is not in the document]
joeduffy: :+1:

content/docs/guides/adopting/import.md
```diff
@@ -0,0 +1,215 @@
...
+No matter how you've provisioned these resources &mdash; manually in your cloud provider's console or CLI, using an infrastructure as code tool like Terraform or AWS CloudFormation &mdash; Pulumi enables you to adopt and manage your resources.
+
+<!--more-->
+
+When working with existing resources, there are two primary scenarios:
+
+* You need to reference existing resources to use as inputs to new resources in Pulumi
+* You need to adopt existing resources under management so they can be managed by Pulumi
+
+We'll review referencing existing resources and then dive deeper to see how to adopt existing resources to be entirely under the control of Pulumi.
```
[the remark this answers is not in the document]
joeduffy: Definitely agree. I've done a pass to minimize this duplication.

content/docs/guides/adopting/from_aws.md
lukehoban: My point was just to make the one step more clear.  That the one step both imports the resource and destroys the CloudFormation stack (without interrupting the actual cloud resources you care about).

PR #2485, opened 2020-02-20
Tweak PurgeCSS to include some missing CSS selectors

justinvp:
I noticed the CSS for styling a data source "symbol" was missing in the `@pulumi/aws` NodeJS API docs: https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/aws/#data-sources-1

The reason it's missing is because we only look or classes in the `@pulumi/pulumi` package, and that package doesn't have any data sources. To fix, just swap looking at `@pulumi/aws` rather than `@pulumi/pulumi`.

## Before

Screen Shot 2020-02-19 at 8 30 44 PM

## After

Screen Shot 2020-02-19 at 8 30 52 PM


justinvp:
> Re the Pulumify failure, I think @stack72 said the cleanup scripts wouldn't start ignoring these resources until tomorrow, so I'm assuming that's what happened here.

Ah, ok.

PR #2494, opened 2020-02-20
Only run autoprefixer, cssnano and purgecss for production builds

cnunciato:
This change limits the use of Autoprefixer, CSSNano and PurgeCSS to full/production builds only, to make the development loop (significantly!) quicker. 

Note that for the Pulumify action, we need to set NODE_ENV at the individual build-command level because [Pulumify excludes ambient environment variables](https://github.com/pulumi/actions-pulumify/blob/0781e889546d5cce5efcb3eec289894fb043e841/infra/pulumify#L82-L85).

### DIFF
--- a/.github/workflows/pulumify.yml
+++ b/.github/workflows/pulumify.yml
@@ -11,6 +11,7 @@ jobs:
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         PULUMI_ACCESS_TOKEN: ${{ secrets.PULUMI_ACCESS_TOKEN }}
-        PULUMIFY_BUILD: make ensure && hugo --minify --buildFuture -e $GITHUB_SHA
+        NODE_ENV: production
+        PULUMIFY_BUILD: NODE_ENV=production make ensure && NODE_ENV=production hugo --minify --buildFuture -e $GITHUB_SHA
         PULUMIFY_ROOT: public
         PULUMIFY_ORGANIZATION: pulumi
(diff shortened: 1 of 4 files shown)


.github/workflows/pulumify.yml
```diff
@@ -11,6 +11,6 @@ jobs:
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         PULUMI_ACCESS_TOKEN: ${{ secrets.PULUMI_ACCESS_TOKEN }}
-        PULUMIFY_BUILD: make ensure && hugo --minify --buildFuture -e $GITHUB_SHA
+        PULUMIFY_BUILD: NODE_ENV=production make ensure && NODE_ENV=production hugo --minify --buildFuture -e $GITHUB_SHA
```
cnunciato: Not sure why this isn't just settable at the action's `env` level, but it doesn't seem to be getting applied. Could have something to do with the way the command is being executed by Pulumify, too. 🤔

cnunciato: Yep, [it does](https://github.com/pulumi/actions-pulumify/blob/0781e889546d5cce5efcb3eec289894fb043e841/infra/pulumify#L82-L85).

Makefile
```diff
@@ -62,7 +62,7 @@ travis_push::
 	$(MAKE) banner
 	$(MAKE) ensure
 ifeq ($(TRAVIS_BRANCH),master)
-	HUGO_ENVIRONMENT=production $(MAKE) build
```
justinvp: There are places where we make use of `{{- if ne hugo.Environment "production" }}` and `{{- if eq hugo.Environment "production" }}` checks. Do those still work now that we're no longer setting `HUGO_ENVIRONMENT=production` explicitly? Or does running `hugo` use `production` by default (whereas `hugo server` uses `development` by default)? (I can't recall off the top of my head.)

PR #2505, opened 2020-02-21
Add support for eventbrite registration modal

zchase:
@isaac-pulumi requested we move the Amsterdam workshop to register through Eventbrite instead of Calendly. This PR adds support for using an Eventbrite embedded modal for registration.

Unfortunately for the modal to show up on the site https needs to be enabled. If you use `ngrok` you can test this locally, otherwise, you will get redirected to an eventbrite page from the live preview.

https://jsonhilder.github.io/blog/ngrok-hugo/

### DIFF
--- a/content/events/workshop-amsterdam-2020-03-30/index.md
+++ b/content/events/workshop-amsterdam-2020-03-30/index.md
@@ -25,8 +25,9 @@ event:
     time: "11:30AM - 4:00pm (CET)"
     # The event description shown on the event list page.
     description: "Join us as we walkthrough Infrastructure as Code concepts via a series of hands-on labs. Topics covered include IaC fundamentals, in addition to application architectures and how to use IaC to create, update, and manage them."
-    # The calendly registrtion url for event specific pages.
-    calendly_url: "https://calendly.com/pulumi/kubecon?month=2020-03"
+    # The eventbrite registrtion url for event specific pages.
+    eventbrite_url: "https://www.eventbrite.com/e/infrastructure-as-code-with-pulumi-tickets-95941386269"
+    eventbrite_id: "95941386269"
     # The cost of an event.
     cost: "€99"
 ---
--- a/content/events/workshop-amsterdam-2020-03-30/index.md
+++ b/content/events/workshop-amsterdam-2020-03-30/index.md
@@ -25,7 +25,7 @@ event:
     time: "11:30AM - 4:00pm (CET)"
     # The event description shown on the event list page.
     description: "Join us as we walkthrough Infrastructure as Code concepts via a series of hands-on labs. Topics covered include IaC fundamentals, in addition to application architectures and how to use IaC to create, update, and manage them."
-    # The eventbrite registrtion url for event specific pages.
+    # The Eventbrite registration URL for the event.
     eventbrite_url: "https://www.eventbrite.com/e/infrastructure-as-code-with-pulumi-tickets-95941386269"
     eventbrite_id: "95941386269"
     # The cost of an event.
--- a/layouts/events/single.html
+++ b/layouts/events/single.html
@@ -55,7 +55,7 @@
                             <button id="eventbrite-widget-modal-trigger-{{ .Params.event.eventbrite_id }}" type="button" class="btn bg-orange-500 border-2 border-orange-600 hover:bg-orange-600 hover:border-orange-600">Buy Tickets</button>
                             <noscript></a>Buy Tickets on Eventbrite</noscript>
                             <script src="https://www.eventbrite.com/static/widgets/eb_widgets.js"></script>
-                            <script type="text/javascript">
+                            <script>
                                 if (window && window.EBWidgets && (typeof window.EBWidgets.createWidget === "function")) {
                                     window.EBWidgets.createWidget({
                                         widgetType: "checkout",
--- a/layouts/events/single.html
+++ b/layouts/events/single.html
@@ -44,7 +44,7 @@
 
                             <!-- Calendly link widget begin -->
                             <link href="https://assets.calendly.com/assets/external/widget.css" rel="stylesheet">
-                            <script src="https://assets.calendly.com/assets/external/widget.js" type="text/javascript"></script>
+                            <script src="https://assets.calendly.com/assets/external/widget.js"></script>
                             <a href="#" class="btn bg-orange-500 border-2 border-orange-600 hover:bg-orange-600 hover:border-orange-600" onclick="Calendly.initPopupWidget({url: '{{ .Params.event.calendly_url }}&hide_event_type_details=1'});return false;">Register Now</a>
                             <!-- Calendly link widget end -->
 
--- a/layouts/events/single.html
+++ b/layouts/events/single.html
@@ -63,7 +63,7 @@
                                         modal: true,
                                         modalTriggerElementId: "eventbrite-widget-modal-trigger-{{ .Params.event.eventbrite_id }}",
                                         onOrderComplete: function() {
-                                            window.location = "/confirmation/workshop/"
+                                            window.location = "{{ relref . "/confirmation/workshop" }}";
                                         }
                                     });
                                 }
--- a/layouts/events/single.html
+++ b/layouts/events/single.html
@@ -65,7 +65,7 @@
                                     eventId: "{{ .Params.event.eventbrite_id }}",
                                     modal: true,
                                     modalTriggerElementId: "eventbrite-widget-modal-trigger-{{ .Params.event.eventbrite_id }}",
-                                    onOrderComplete: exampleCallback
+                                    onOrderComplete: goToThankYouPage
                                 });
                             </script>
 
--- a/layouts/events/single.html
+++ b/layouts/events/single.html
@@ -40,11 +40,36 @@
                         <span>{{ .Params.event.cost }}</span>
                     </div>
                     <div class="my-8">
-                        <!-- Calendly link widget begin -->
-                        <link href="https://assets.calendly.com/assets/external/widget.css" rel="stylesheet">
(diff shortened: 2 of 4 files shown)


layouts/events/single.html
```diff
@@ -40,11 +40,36 @@ <h3>Event Details</h3>
...
+                            <!-- Calendly link widget begin -->
+                            <link href="https://assets.calendly.com/assets/external/widget.css" rel="stylesheet">
+                            <script src="https://assets.calendly.com/assets/external/widget.js" type="text/javascript"></script>
+                            <a href="#" class="btn bg-orange-500 border-2 border-orange-600 hover:bg-orange-600 hover:border-orange-600" onclick="Calendly.initPopupWidget({url: '{{ .Params.event.calendly_url }}&hide_event_type_details=1'});return false;">Register Now</a>
+                            <!-- Calendly link widget end -->
+
+                        {{ else if .Params.event.eventbrite_url }}
+
+                            <!-- Noscript content for added SEO -->
+                            <noscript><a href="{{ .Params.event.eventbrite_url }}" rel="noopener noreferrer" target="_blank"></noscript>
```
cnunciato: Should there be something else in here? I also don't see a closing `a` tag.

layouts/events/single.html
```diff
@@ -40,11 +40,36 @@ <h3>Event Details</h3>
...
+                            <!-- You can customize this button any way you like -->
```
cnunciato: Assuming this came from EventBrite, we could probably just remove it.

layouts/events/single.html
```diff
@@ -40,11 +40,36 @@ <h3>Event Details</h3>
...
+                            <button id="eventbrite-widget-modal-trigger-{{ .Params.event.eventbrite_id }}" type="button" class="btn bg-orange-500 border-2 border-orange-600 hover:bg-orange-600 hover:border-orange-600">Buy Tickets</button>
+                            <noscript></a>Buy Tickets on Eventbrite</noscript>
```
cnunciato: Oh, I see -- hm. Why not put these noscript blocks together? There's probably a good reason; it just seems odd, so worth asking.

layouts/events/single.html
```diff
@@ -40,11 +40,36 @@ <h3>Event Details</h3>
...
+                            <script src="https://www.eventbrite.com/static/widgets/eb_widgets.js"></script>
+                            <script type="text/javascript">
+                                var goToThankYouPage = function() {
+                                    window.location = "/confirmation/workshop/"
+                                };
+                                window.EBWidgets.createWidget({
```
cnunciato: We should probably check for the presence of `EBWidgets` and `EBWidgets.createWidget` first.

layouts/events/single.html
```diff
@@ -40,11 +40,36 @@ <h3>Event Details</h3>
...
+                                    widgetType: "checkout",
+                                    eventId: "{{ .Params.event.eventbrite_id }}",
+                                    modal: true,
+                                    modalTriggerElementId: "eventbrite-widget-modal-trigger-{{ .Params.event.eventbrite_id }}",
+                                    onOrderComplete: goToThankYouPage
```
cnunciato: Nit: Maybe just put the function right inline, here, since doesn't seem to be needed otherwise.

layouts/events/single.html
zchase: It is on line 57. The code snippet is directly from Eventbrite so it is a bit wonky

layouts/events/single.html
zchase: 👍

layouts/events/single.html
zchase: I agree it seems odd, but this is what Eventbrite tells us to do ..

layouts/events/single.html
zchase: 👍

layouts/events/single.html
zchase: 👍

layouts/events/single.html
```diff
@@ -40,11 +40,36 @@ <h3>Event Details</h3>
...
+                            <!-- Calendly link widget end -->
+
+                        {{ else if .Params.event.eventbrite_url }}
+
+                            <!-- Noscript content for added SEO -->
+                            <noscript><a href="{{ .Params.event.eventbrite_url }}" rel="noopener noreferrer" target="_blank"></noscript>
+                            <button id="eventbrite-widget-modal-trigger-{{ .Params.event.eventbrite_id }}" type="button" class="btn bg-orange-500 border-2 border-orange-600 hover:bg-orange-600 hover:border-orange-600">Buy Tickets</button>
+                            <noscript></a>Buy Tickets on Eventbrite</noscript>
+                            <script src="https://www.eventbrite.com/static/widgets/eb_widgets.js"></script>
+                            <script type="text/javascript">
```
justinvp: Nit:

```suggestion
                            <script>
```

layouts/events/single.html
```diff
@@ -40,11 +40,36 @@ <h3>Event Details</h3>
...
-                        <!-- Calendly link widget begin -->
-                        <link href="https://assets.calendly.com/assets/external/widget.css" rel="stylesheet">
-                        <script src="https://assets.calendly.com/assets/external/widget.js" type="text/javascript"></script>
-                        <a href="#" class="btn bg-orange-500 border-2 border-orange-600 hover:bg-orange-600 hover:border-orange-600" onclick="Calendly.initPopupWidget({url: '{{ .Params.event.calendly_url }}&hide_event_type_details=1'});return false;">Register Now</a>
-                        <!-- Calendly link widget end -->
+                        {{ if .Params.event.calendly_url }}
+
+                            <!-- Calendly link widget begin -->
+                            <link href="https://assets.calendly.com/assets/external/widget.css" rel="stylesheet">
+                            <script src="https://assets.calendly.com/assets/external/widget.js" type="text/javascript"></script>
```
justinvp: Nit:
```suggestion
                            <script src="https://assets.calendly.com/assets/external/widget.js"></script>
```

content/events/workshop-amsterdam-2020-03-30/index.md
```diff
@@ -25,8 +25,9 @@ event:
     time: "11:30AM - 4:00pm (CET)"
     # The event description shown on the event list page.
     description: "Join us as we walkthrough Infrastructure as Code concepts via a series of hands-on labs. Topics covered include IaC fundamentals, in addition to application architectures and how to use IaC to create, update, and manage them."
-    # The calendly registrtion url for event specific pages.
-    calendly_url: "https://calendly.com/pulumi/kubecon?month=2020-03"
+    # The eventbrite registrtion url for event specific pages.
```
cnunciato:
```suggestion
    # The Eventbrite registration URL for the event.
```

Also, should we add this to the archetype? Should it _replace_ Calendly?
Should we add this to the archetype as well, as an option? Are we _switching_ to using EventBrite.

layouts/events/single.html
cnunciato: Ok. If it does only work this way, we should probably leave a comment similar to the Calendly example that makes it clear this is a pasted snip from Eventbrite.

layouts/events/single.html
```diff
@@ -40,11 +40,36 @@ <h3>Event Details</h3>
...
+                            <script src="https://www.eventbrite.com/static/widgets/eb_widgets.js"></script>
+                            <script>
+                                if (window && window.EBWidgets && (typeof window.EBWidgets.createWidget === "function")) {
+                                    window.EBWidgets.createWidget({
+                                        widgetType: "checkout",
+                                        eventId: "{{ .Params.event.eventbrite_id }}",
+                                        modal: true,
+                                        modalTriggerElementId: "eventbrite-widget-modal-trigger-{{ .Params.event.eventbrite_id }}",
+                                        onOrderComplete: function() {
+                                            window.location = "/confirmation/workshop/"
```
cnunciato:
```suggestion
                                            window.location = "/confirmation/workshop/";
```

We should probably use a `relref` here, too.

content/events/workshop-amsterdam-2020-03-30/index.md
zchase: So it is unclear if we are switching to Calendly. Eventbrite is being used in this specific case because we needed to be able to use promo codes. For now, this is a one-off.

layouts/events/single.html
```diff
@@ -40,11 +40,36 @@ <h3>Event Details</h3>
...
+
+                        {{ else if .Params.event.eventbrite_url }}
+
+                            <!-- Noscript content for added SEO -->
+                            <noscript><a href="{{ .Params.event.eventbrite_url }}" rel="noopener noreferrer" target="_blank"></noscript>
+                            <button id="eventbrite-widget-modal-trigger-{{ .Params.event.eventbrite_id }}" type="button" class="btn bg-orange-500 border-2 border-orange-600 hover:bg-orange-600 hover:border-orange-600">Buy Tickets</button>
+                            <noscript></a>Buy Tickets on Eventbrite</noscript>
+                            <script src="https://www.eventbrite.com/static/widgets/eb_widgets.js"></script>
+                            <script>
+                                if (window && window.EBWidgets && (typeof window.EBWidgets.createWidget === "function")) {
```
cnunciato:
```suggestion
                                if (window.EBWidgets && (typeof window.EBWidgets.createWidget === "function")) {
```

(We can safely assume window, just not `EBWidgets`.)

layouts/events/single.html
zchase: 👍

layouts/events/single.html
zchase: 👍

zchase:
@cnunciato This PR also now includes custom subtitle support because we need to get the change in before MSFT tweets about the Redmond workshop Monday morning.

PR #2506, opened 2020-02-21
Do not show old events on the event list

zchase:
This PR adds support for hiding old events from the event list. I also removed the old event pages. 

The best way to handle old events is for us to delete the pages after the event ends but this PR adds an `if...` statement to check if the event end date is after yesterday. If we check against `now` our events will be unlisted a day early. So I made the decision to check against the previous day to ensure the event will be listed through the entirety of the event duration.

### DIFF
--- a/layouts/events/list.html
+++ b/layouts/events/list.html
@@ -19,7 +19,7 @@
                 <ul id="event-list" class="list-none md:flex md:flex-wrap pl-0">
                     {{ $events := (where $.Pages "Type" "events") }}
                     {{ range $index, $event := (sort $events ".Params.event.start_date") }}
-                        {{ if eq ((time .Params.event.end_date).After (now.AddDate 0 0 -1)) true}}
+                        {{ if (time .Params.event.end_date).After (now.AddDate 0 0 -1)}}
                             <li class="w-full m-0 p-0" data-event-type='{{ delimit .Params.event.type "," }}'>
                                 <article class="{{ if lt (add $index 1) (len $events) }} pb-16 mb-12 border-b border-gray-300 {{ end }}">
                                     <h3>
--- a/layouts/events/list.html
+++ b/layouts/events/list.html
@@ -18,8 +18,7 @@
                 <h2 id="event-list-heading" class="mb-8">All Upcoming Events</h2>
                 <ul id="event-list" class="list-none md:flex md:flex-wrap pl-0">
                     {{ $events := (where $.Pages "Type" "events") }}
-                    {{ $currentEvents := (where $events "{{ (time .Params.event.start_date).After (now.AddDate 0 0 -1) }}" "ne" true ) }}
-                    {{ range $index, $event := (sort $currentEvents ".Params.event.start_date") }}
+                    {{ range $index, $event := (sort $events ".Params.event.start_date") }}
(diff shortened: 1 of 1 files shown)


stack72:
I think an enhancement here would be to search for the old events as well - I am almost sure we will want to track (at the end of the year) how many events we have been to but YMMV of course

zchase:
@stack72 So it is probably best to keep the old event files around instead of deleting them?

cnunciato:
I wouldn't delete them either, yeah. Let's just let them accumulate for now.

It does make me wonder whether we could do a better job of keeping these directories tidy, though -- like grouping them by year, month, date. We should probably be doing the same for the blog as well. But in general yeah, don't delete, yet.

stack72:
I’m going to defer this to @cnunciato et al

layouts/events/list.html
```diff
@@ -18,60 +18,63 @@ <h1>Upcoming Events</h1>
...
-                                            <i class="text-sm ml-2 fas fa-external-link-alt"></i>
-                                        </a>
-                                    {{ end }}
-                                </h3>
-                                <div>
-                                    <i class="text-gray-800 far fa-calendar inline-block"></i>
-                                    <span class="ml-1 text-sm uppercase">
-                                        {{ if eq .Params.event.start_date .Params.event.end_date }}
-                                            {{ dateFormat "Monday, Jan 2" .Params.event.start_date }}
+                    {{ $currentEvents := (where $events "{{ (time .Params.event.start_date).After (now.AddDate 0 0 -1) }}" "ne" true ) }}
```
cnunciato: Since we aren't passing this as a list to a paginator, we can detangle this a bit by just doing the filtering as a separate if:

```
{{ range $index, $event ... }}
    {{ if (time .Params.event.start_date).After (now.AddDate 0 0 -1) }}
```

zchase: `$currentEvents` was leftover from when I was trying to filter using `where` so I would need the `if..` statement. I went ahead and removed the old code to simplify it all. Thanks for catching this!

layouts/events/list.html
```diff
@@ -19,59 +19,61 @@ <h2 id="event-list-heading" class="mb-8">All Upcoming Events</h2>
...
-                                            <i class="text-sm ml-2 fas fa-external-link-alt"></i>
-                                        </a>
-                                    {{ end }}
-                                </h3>
-                                <div>
-                                    <i class="text-gray-800 far fa-calendar inline-block"></i>
-                                    <span class="ml-1 text-sm uppercase">
-                                        {{ if eq .Params.event.start_date .Params.event.end_date }}
-                                            {{ dateFormat "Monday, Jan 2" .Params.event.start_date }}
+                        {{ if eq ((time .Params.event.end_date).After (now.AddDate 0 0 -1)) true}}
```
cnunciato: Not to drag this out forever, but the snip I posted actually works as-is, without the `eq ... true` (since because the expression evaluates boolean, the check for additional trueness isn't necessary):

```suggestion
                        {{ if (time .Params.event.start_date).After (now.AddDate 0 0 -1) }}
```

zchase: I removed the `eq..` stuff but left the end date because I think the event should be listed until it is over rather than when it starts.

PR #2515, opened 2020-02-24
SEO optimizations for "Pulumi vs Cloud Templates"

zchase:
This PR is a continuation of https://github.com/pulumi/docs/pull/2376. This applies the suggested changes after 97th Floor did an initial optimization.

### DIFF
--- a/content/docs/intro/vs/cloud_templates.md
+++ b/content/docs/intro/vs/cloud_templates.md
@@ -10,20 +10,13 @@ menu:
 aliases: ["/docs/reference/vs/cloud_templates/"]
 ---
 
-## What is CloudFormation?
-
 All major cloud providers offer their own form of infrastructure-as-"code" solution, typically by way of JSON or
 YAML-based templating solutions. This includes AWS CloudFormation and Azure Resource Manager (ARM) templates.
 
 These markup-based configuration files are often uploaded to a hosted service in the target cloud, where a hosted
 service will then process the files to create, update, or delete resources as necessary.
 
-## The Pulumi Difference
-
-Pulumi offers a multi-cloud alternative to more restrictive infrastructure-as-code template solutions, such as AWS CloudFormation. Pulumi allows you to streamline processes through managing resources from different platforms all in one place, using real code, in your favorite language.
-
-Pulumi's model shares a lot with these systems. Although Pulumi programs are written in imperative, familiar languages,
-they are ultimately evaluated to produce a similar set of create, update, or delete operations for your cloud resources. This is where Pulumi begins to diverge from these other solutions such as AWS CloudFormation, however.
+Pulumi offers a multi-cloud alternative to more restrictive infrastructure-as-code template solutions, like AWS CloudFormation. Pulumi allows you to streamline processes by managing resources from different platforms all in one place, using real code, in your favorite language.
 
 Pulumi lets you use your favorite languages, instead of bespoke templating solutions such as CloudFormation. These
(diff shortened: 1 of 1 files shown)


content/docs/intro/vs/cloud_templates.md
```diff
@@ -11,21 +10,25 @@ menu:
 aliases: ["/docs/reference/vs/cloud_templates/"]
 ---
 
+## What is CloudFormation?
```
cnunciato: This heading doesn't work, as the content addresses ARM also. IMO we should just remove it.

content/docs/intro/vs/cloud_templates.md
```diff
@@ -11,21 +10,25 @@ menu:
...
+
 All major cloud providers offer their own form of infrastructure-as-"code" solution, typically by way of JSON or
 YAML-based templating solutions. This includes AWS CloudFormation and Azure Resource Manager (ARM) templates.
 
 These markup-based configuration files are often uploaded to a hosted service in the target cloud, where a hosted
 service will then process the files to create, update, or delete resources as necessary.
 
-Pulumi's model shares a lot with these systems. Although Pulumi programs are written in imperative, familiar languages,
-they are ultimately evaluated to produce a similar set of create, update, or delete operations for your cloud resources.
+## The Pulumi Difference
```
cnunciato: Same with this one -- the phrase has an advertising-y ring to it and doesn't add much value, and since already seem to rank #1 for "difference between pulumi and cloudformation", presumably because of the "vs" in the title, so I'd suggest dropping this heading, too.

content/docs/intro/vs/cloud_templates.md
```diff
@@ -11,21 +10,25 @@ menu:
...
 
-This is where Pulumi begins to diverge from these other solutions, however.
+Pulumi offers a multi-cloud alternative to more restrictive infrastructure-as-code template solutions, such as AWS CloudFormation. Pulumi allows you to streamline processes through managing resources from different platforms all in one place, using real code, in your favorite language.
```
cnunciato:
```suggestion
Pulumi offers a multi-cloud alternative to more restrictive infrastructure-as-code template solutions like AWS CloudFormation. Pulumi allows you to streamline processes by managing resources from different platforms all in one place, using real code, in your favorite language.
```

content/docs/intro/vs/cloud_templates.md
```diff
@@ -11,21 +10,25 @@ menu:
...
+
+Pulumi's model shares a lot with these systems. Although Pulumi programs are written in imperative, familiar languages,
```
cnunciato: This paragraph doesn't make sense wedged in where it is. It interrupts the flow between the second and fourth paragraphs, and is essentially saying the same thing as the fifth. We should reposition it, make it work, or remove it.

content/docs/intro/vs/cloud_templates.md
zchase: 👍

content/docs/intro/vs/cloud_templates.md
zchase: 👍

content/docs/intro/vs/cloud_templates.md
zchase: 👍

content/docs/intro/vs/cloud_templates.md
zchase: Since it is redundant I opted to remove it. No reason to have redundant content IMO.

PR #2536, opened 2020-02-26
Add Go API docs

lukehoban:
Part of #2311.


praneetloke:
I see this on godoc.org. Should we link to that?

>  Pkg.go.dev is a new destination for Go discovery & docs. Check it out at pkg.go.dev/github.com/pulumi/pulumi/sdk/go/pulumi and share your feedback.

lukehoban:
> Pkg.go.dev is a new destination for Go discovery & docs. 

Indeed - looks like that is now a much better place to link to - will update!

PR #2540, opened 2020-02-27
SEO optimization for KW "OpenStack Providers"

97thFloor:
### Optimized for SEO

Optimized metadata and some new language based on semantic analysis.
Primary KW: OpenStack Providers
Language from semantic analysis: private, public, infrastructure, cloud provider, software, data set

### Proposed changes

Just from a content standpoint, I think a single sentence intro/explanation before the example would benefit this page, but I don't have the information to write that myself. I also assumed "packager" was a typo, but feel free to reject that if I'm wrong there.

@cnunciato @zchase

### DIFF
--- a/content/docs/intro/cloud-providers/openstack/_index.md
+++ b/content/docs/intro/cloud-providers/openstack/_index.md
@@ -1,5 +1,5 @@
 ---
-title: OpenStack Provider
+title: OpenStack
 meta_desc: The OpenStack provider for Pulumi can be used to provision any of the cloud resources available in OpenStack.
 menu:
   intro:
--- a/content/docs/intro/cloud-providers/openstack/_index.md
+++ b/content/docs/intro/cloud-providers/openstack/_index.md
@@ -1,6 +1,6 @@
 ---
-title: OpenStack
-meta_desc: This page provides an overview of the OpenStack Provider for Pulumi.
+title: OpenStack Provider
+meta_desc: The OpenStack provider for Pulumi can be used to provision any of the cloud resources available in OpenStack.
 menu:
   intro:
     parent: cloud-providers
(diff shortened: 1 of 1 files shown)


content/docs/intro/cloud-providers/openstack/_index.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: OpenStack
-meta_desc: This page provides an overview of the OpenStack Provider for Pulumi.
+title: OpenStack Provider
```
zchase:
```suggestion
title: OpenStack
```

@97thFloor How important is it to add provider into the title here? None of the other "providers" have "Provider" in their title so it makes this page look a bit out of place in our nav. If it is a good idea to add provider to the title's we should do it across all the pages and add a field for link text.

97thFloor: Most of the pages ranking on page one for this keyword have "provider" in the title, aside from OpenStack themselves, so I'd say keep it if possible. For the other pages I can't say definitely without knowing what keywords they are ranking for, if any. However, since they're all the same format to this page, I feel confident saying it would probably be a good move. From a conversions standpoint, it's a bit more specific, so people looking at the SERP would know that this is a useful tool rather than just an informational page.

zchase: @cnunciato ☝️ WDYT?

cnunciato: I think this is pretty visually noisy, TBH:

image

So if we could serve the need by having the page title (i.e., the H1) maybe contain the word Provider instead, keeping the navigation clean and scannable, that'd be a better way to go.

zchase:
@cnunciato I commited removing the word "Provider" from the title for now. I filed a follow up issue here:https://github.com/pulumi/docs/issues/2673

I think we are good to merge now?

content/docs/intro/cloud-providers/openstack/_index.md
```diff
@@ -80,11 +80,11 @@ class Program
...
-The following packages are available in packager managers:
+The following packages are available in package managers:
 
 * JavaScript/TypeScript: [`@pulumi/openstack`](https://www.npmjs.com/package/@pulumi/openstack)
 * Python: [`pulumi-openstack`](https://pypi.org/project/pulumi-openstack/)
 * Go: [`github.com/pulumi/pulumi-openstack/sdk/go/openstack`](https://github.com/pulumi/pulumi-openstack)
 * .NET: [`Pulumi.Openstack`](https://www.nuget.org/packages/Pulumi.Openstack)
 
-The OpenStack provider is open source and available in the [pulumi/pulumi-openstack](https://github.com/pulumi/pulumi-openstack) repo.
+As with all Pulumi software solutions, the OpenStack provider is open source and available in the [pulumi/pulumi-openstack](https://github.com/pulumi/pulumi-openstack) repo. You can see all Pulumi infrastructure and cloud providers [here] (https://www.pulumi.com/docs/intro/cloud-providers/).
```
zchase, line 90:
```suggestion
As with all Pulumi software solutions, the OpenStack provider is open source and available in the [pulumi/pulumi-openstack](https://github.com/pulumi/pulumi-openstack) repo. You can see all Pulumi infrastructure and cloud providers [here] ({{< relref "/docs/intro/cloud-providers" >}}).
```

PR #2551, opened 2020-02-27
Initial resource model docs for AWS

justinvp:
This is not intended to be checked-in. It just demonstrates an early look at the resource model docs for AWS. Many TODOs and follow-ups remain.


justinvp:
@cnunciato, looks like we're going to need to fix Pulumify for real sooner than later. Adding all these new resource model docs pushes us over the limit.

content/docs/reference/providers/pulumi/aws/s3/bucket.md
```diff
@@ -0,0 +1,1827 @@
...
+<style>
+  table td p { margin-top: 0; margin-bottom: 0; }
+</style>
+
+Provides a S3 bucket resource.
+
+> This content is derived from https://github.com/terraform-providers/terraform-provider-aws/blob/master/website/docs/r/s3_bucket.html.markdown.
+
+
+## Example Usage
```
joeduffy: I wonder if we can collapse this entire section by default, with a little ▶️ collapser icon. I actually think it's right to include this at the top, however it tends to be lengthy and it can obscure the instructions on how to create new instances (which may be what most folks are wanting to see when they get here). Making it collapsed seems fine because if you want to see examples, you're going to see "Example Usage ▶️" and immediately know to click to expand.

The only hitch I see here is if you've linked to an anchor -- e.g. .../docs/reference/providers/pulumi/aws/s3/bucket/#private-bucket-w-tags -- if it's collapsed the right thing may not happen.

I actually think this is a small change that could make a huge positive usability impact, and lessen the feeling that our pages are so long and require so much scrolling to get to the meat.

content/docs/reference/providers/pulumi/aws/s3/bucket.md
```diff
@@ -0,0 +1,1827 @@
...
+const bucket = new aws.s3.Bucket("b", {
+	acl: "private",
+	tags: {
+		Environment: "Dev",
+		Name: "My bucket",
+	},
+});
+```
+
+```python
```
joeduffy: Should we add other languages and say coming soon for them also?

content/docs/reference/providers/pulumi/aws/s3/bucket.md
```diff
@@ -0,0 +1,1827 @@
...
+    <thead>
+        <tr>
+            <th>Argument</th>
+            <th>Type</th>
+            <th>Description</th>
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td class="align-top">acceleration<wbr>Status</td>
```
joeduffy: Short we sort the required properties to come first? I'd also love to make them pop better visually, either an icon or color the rows differently.

content/docs/reference/providers/pulumi/aws/s3/bucket.md
```diff
@@ -0,0 +1,1827 @@
...
+            <td class="align-top"><code>string</code></td>
+            <td class="align-top">{{% md %}}
+(Optional) The website endpoint, if the bucket is configured with a website. If not, this will be an empty string.
+
+{{% /md %}}</td>
+        </tr>
+    </tbody>
+</table>
+
+## Import an Existing Bucket Resource
```
joeduffy: It doesn't feel like we need an import section. Unlike Terraform, Pulumi's support for import works universally across all resource types, and isn't any different than just creating the new resource with an import tag.

content/docs/reference/providers/pulumi/aws/s3/bucket.md
```diff
@@ -0,0 +1,1827 @@
...
+            <td class="align-top"><code>string</code></td>
+            <td class="align-top">{{% md %}}
+(Optional) The website endpoint, if the bucket is configured with a website. If not, this will be an empty string.
+
+{{% /md %}}</td>
+        </tr>
+    </tbody>
+</table>
+
+## Bucket Output Properties
```
joeduffy: This section would be so much more useful if it filtered out the properties that are also input properties. The docs aren't any different and if I'm explicitly looking for output properties, I probably only expect to see the outputs -- that is, the properties computed by the resource provider. Will make this way more streamlined and useful.

content/docs/reference/providers/pulumi/aws/s3/bucket.md
```diff
@@ -0,0 +1,1827 @@
...
+{{% lang csharp %}}
+<ul class="pl-10">
+    <li><strong>name</strong> &ndash; (Required) The unique name of the resulting resource.</li>
+    <li><strong>id</strong> &ndash; (Required) The _unique_ provider ID of the resource to lookup.</li>
+    <li><strong>state</strong> &ndash; (Optional) Any extra arguments used during the lookup.</li>
+    <li><strong>opts</strong> &ndash; (Optional) A bag of options that control this resource's behavior.</li>
+</ul>
+{{% /lang %}}
+
+The following state arguments are supported:
```
joeduffy: I'll admit, it bums me out that this uses up so much space when it's actually so rare to use. 90% of the time, you're just using the ID. Maybe we can link to the input arguments above and say it's option?

content/docs/reference/providers/pulumi/aws/s3/bucket.md
```diff
@@ -0,0 +1,1827 @@
...
+	"Redirect": {
+		"ReplaceKeyPrefixWith": "documents/"
+	}
+}]
+`,
+	},
+});
+```
+
+```python
```
joeduffy: I'd add the other languages here too, even if empty, with a link to the work item so folks know they're on the way.

content/docs/reference/providers/pulumi/aws/s3/bucket.md
```diff
@@ -0,0 +1,1827 @@
...
+
+{{% lang csharp %}}
+<ul class="pl-10">
+    <li><strong>name</strong> &ndash; (Required) The unique name of the resulting resource.</li>
+    <li><strong>args</strong> &ndash; (Optional) The arguments to use to populate this resource's properties.</li>
+    <li><strong>opts</strong> &ndash; (Optional) A bag of options that control this resource's behavior.</li>
+</ul>
+{{% /lang %}}
+
+The following arguments are supported:
```
joeduffy: I wish these tables could be more densely packed. For prose, I find our current layout relatively pleasant, although the limited horizontal width means that I often feel like our docs require a lot of scrolling. Should we consider using a more condensed layout and smaller font for the information like properties which should be more dense?

For instance, check out https://devcenter.heroku.com/articles/platform-api-reference#account-attributes.

cnunciato: We chatted about this and I have some ideas.

content/docs/reference/providers/pulumi/aws/s3/bucket.md
justinvp: > The only hitch I see here is if you've linked to an anchor -- e.g. .../docs/reference/providers/pulumi/aws/s3/bucket/#private-bucket-w-tags -- if it's collapsed the right thing may not happen.

We could write some JavaScript that would make sure the right thing happens, e.g. if there current page's URL contains an anchor inside the collapsed example section, to automatically expand it.

justinvp:
Great feedback, thanks! cc: @praneetloke

content/docs/reference/providers/pulumi/aws/s3/bucket.md
cnunciato: It does need to be collapsed and expandable, I agree. This was also on my list of things to try.

joeduffy:
🎉 These changes are incredible, I can't wait to see them land. Even if they land with some warts and get better over time, I'd encourage that path -- people are going to love these 😄

Issue #2509, opened 2020-02-21
Create Infrastructure As Code landing page


zchase:
### Problem description

We are currently ranking on the second page of Google for the term "infrastructure as code" which is pointing to our homepage. We should create an a landing page for the term, i.e. `https://pulumi.com/infrastructure-as-code`.

cc// @joeduffy

PR #2554, opened 2020-02-27
Add infra as code landing page

zchase:
## Problem description
We are currently ranking on the second page of Google for the term "infrastructure as code" which is pointing to our homepage. We should create an a landing page for the term, i.e. https://pulumi.com/infrastructure-as-code.

Relevant issue: https://github.com/pulumi/docs/issues/2509

### DIFF
--- a/content/topics/infrastructure-as-code.md
+++ b/content/topics/infrastructure-as-code.md
@@ -1,0 +1,98 @@
+---
+title: What is Infrastrucutre as Code?
+layout: infrastructure-as-code
+url: /infrastructure-as-code
+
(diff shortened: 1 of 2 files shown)


joeduffy:
This warrants a bit of a deeper discussion. I'm fine starting with what we've got here, however a major thing we want to do with this page is describe what _modern_ IaC means. And for that we need to articulate what that means to us and why we think our approach is differentiated as a result.

cnunciato:
@zchase do you want to close this or revisit it?

zchase:
@cnunciato We want this because if ever change the home page messaging to be something other than "Infrastructure as Code" then we would lose the SEO ranking. For example, if we end up moving to a more "Cloud Engineering" based message.

cnunciato:
@zchase OK, thanks. Do you plan on picking this back up and resubmitting, then? I was mainly asking in light of its age, unmerged status and Joe's question.

cnunciato:
Still keep, or close and move to pulumi-hugo?

susanev:
@zchase do we still need to keep this pr around?

PR #2558, opened 2020-02-28
Regen docs for pulumi@v1.11.1

pgavlin:


static/latest-version
```diff
@@ -1 +1 @@
-1.11.0
\ No newline at end of file
+v1.11.1
```
justinvp: Remove prefix.

pgavlin: Argh! Thanks.

PR #2572, opened 2020-03-03
SEO optimization for 'aws api gateway'

zchase:
We are currently ranking on the second page for the term "aws api gateway". We are currently ranking 15th and would like to get to the first page for this term and related terms.

apigatewayseorank

I tried to get the keyword into places that made sense. If you don't think a keyword makes sense were it is, we can remove it.

### DIFF
--- a/content/docs/guides/crosswalk/aws/api-gateway.md
+++ b/content/docs/guides/crosswalk/aws/api-gateway.md
@@ -540,7 +540,7 @@ https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-control-
 ### Authorizing Requests using Cognito Authorizers
 
 [Cognito Authorizers](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-integrate-with-cognito.html)
-allow you to use [AWS Cognito User Pools](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-identity-pools.html)
+allow you to use [Amazon Cognito User Pools](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-identity-pools.html)
 as an Authorizer for API Gateway. With a user pool, your users can sign into your web or mobile app through
 Amazon Cognito directly, or through social identity providers like Facebook or Amazon, or even through SAML
 identity providers. This enables your API Gateway to offload the difficult work of security to Cognito entirely.
--- a/content/docs/guides/crosswalk/aws/api-gateway.md
+++ b/content/docs/guides/crosswalk/aws/api-gateway.md
@@ -527,7 +527,7 @@ and Amazon Web Application Firewall (WAF) -- and limiting access to authorized c
 
 The `awsx.apigateway.API` class supports three specific methods of controlling access to your APIs:
 
-* *AWS Cognito user pools* let you create customizable authentication and authorization for your APIs.
+* *Amazon Cognito user pools* let you create customizable authentication and authorization for your APIs.
 * *Lambda authorizers* are Lambda Functions that control access to your APIs based on HTTP information
   available in headers, paths, query strings, or other request information, including bearer tokens.
 * *Usage plans* let you provide *API keys* to customers, and then track and limit usage of your APIs.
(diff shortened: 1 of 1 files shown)


content/docs/guides/crosswalk/aws/api-gateway.md
```diff
@@ -14,20 +14,20 @@ aliases: ["/docs/reference/crosswalk/aws/api-gateway/"]
...
-## Overview
-
-Pulumi Crosswalk for AWS provides significantly easier ways of programming API Gateway. This includes using
-infrastructure as code techniques for simple, declarative APIs, including easy Lambda-based handlers.
+## Defining an AWS API Gateway Endpoint and Routes
 
-## Defining an API Gateway Endpoint and Routes
-
-API Gateway creates REST APIs that:
+An AWS API Gateway creates REST APIs that:
```
cnunciato: I think this is okay without the additional article, since that's its name.

```suggestion
AWS API Gateway creates REST APIs that:
```

content/docs/guides/crosswalk/aws/api-gateway.md
```diff
@@ -520,27 +520,27 @@ https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-method-
...
-API Gateway supports several mechanisms for controlling and managing access to your APIs. This includes authentication
+AWS API Gateway supports several mechanisms for controlling and managing access to your APIs. This includes authentication
 and authorization -- e.g., resource policies, standard AWS IAM roles and policies, Cognito user pools, and Lambda
 authorizers -- other access control tasks -- e.g., cross-origin resource sharing (CORS), client-side SSL certificates,
 and Amazon Web Application Firewall (WAF) -- and limiting access to authorized clients through usage plans and API keys.
 
 The `awsx.apigateway.API` class supports three specific methods of controlling access to your APIs:
 
-* *Amazon Cognito user pools* let you create customizable authentication and authorization for your APIs.
+* *AWS Cognito user pools* let you create customizable authentication and authorization for your APIs.
```
cnunciato: I believe for this one, the product name actually is Amazon Cognito, not AWS Cognito.
```suggestion
* *Amazon Cognito user pools* let you create customizable authentication and authorization for your APIs.
```

content/docs/guides/crosswalk/aws/api-gateway.md
```diff
@@ -520,27 +520,27 @@ https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-method-
...
 [AWS package]({{< relref "/docs/reference/pkg/nodejs/pulumi/aws" >}}), and are described in depth in the article
-[Controlling and Managing Access to a REST API in API Gateway](
+[Controlling and Managing Access to a REST API in AWS API Gateway](
 https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-control-access-to-api.html).
 
 ### Authorizing Requests using Cognito Authorizers
 
 [Cognito Authorizers](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-integrate-with-cognito.html)
-allow you to use [Amazon Cognito User Pools](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-identity-pools.html)
+allow you to use [AWS Cognito User Pools](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-identity-pools.html)
```
cnunciato:
```suggestion
allow you to use [Amazon Cognito User Pools](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-identity-pools.html)
```

cnunciato:
```suggestion
allow you to use [Amazon Cognito User Pools](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-identity-pools.html)
```

content/docs/guides/crosswalk/aws/api-gateway.md
```diff
@@ -14,20 +14,20 @@ aliases: ["/docs/reference/crosswalk/aws/api-gateway/"]
     <img src="/images/docs/reference/crosswalk/aws/logo.svg" align="right" width="280" style="margin: 0 0 32px 16px;">
 </a>
 
-[Amazon API Gateway](https://aws.amazon.com/api-gateway/) is a fully managed service for creating, monitoring, and
+Pulumi Crosswalk for AWS provides significantly easier ways of programming an AWS API Gateway. This includes using
```
cnunciato: I actually think this works well, but it's inconsistent with the rest of the pages in this section. So if we do apply this change here, we should probably do so on its siblings as well.

zchase:
@cnunciato I took your suggestions and reverted the "What is.." part I added to how it originally was. I think it would be a nice addition SEO wise, but agree that we should be consistent across all pages. Just adding the keywords should be fine for now, and if we feel we still need to optimize the page further we can look at that format change

PR #2576, opened 2020-03-03
Added Go example for transformations

tasiah:
Added an example for Go transformations. 

I wasn't sure how best to express the `pulumi.runtime.registerStackTransformation` equivalent for Go, so I tacked it onto the end of the paragraph. Another option would be to add `...then applies the same transformation to all future resources in the stack:` to the previous paragraph and include each language's `registerStackTransformation` with its example.

### DIFF
--- a/content/docs/intro/concepts/programming-model.md
+++ b/content/docs/intro/concepts/programming-model.md
@@ -1013,7 +1013,7 @@ var vpc = new MyVpcComponent("vpc", new ResourceOptions
 });
 ```
 
-Transformations can also be applied in bulk to many resources in a stack by using the `pulumi.runtime.registerStackTransformation` or `ctx.RegisterStackTransformation` function.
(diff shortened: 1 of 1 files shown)


content/docs/intro/concepts/programming-model.md
```diff
@@ -1005,7 +1013,7 @@ var vpc = new MyVpcComponent("vpc", new ResourceOptions
 });
 ```
 
-Transformations can also be applied in bulk to many resources in a stack by using the `pulumi.runtime.registerStackTransformation` function.
+Transformations can also be applied in bulk to many resources in a stack by using the `pulumi.runtime.registerStackTransformation` or `ctx.RegisterStackTransformation` function.
```
EvanBoyle: This might be confusing the users. We should either have a language choose here, or just refer to the name of the method without the namespacing.

Issue #2137, opened 2019-12-11
API docs for .NET SDK


mikhailshilkov:
Add .NET to the API docs at https://www.pulumi.com/docs/reference/pkg/

lukehoban:
@mikhailshilkov @praneetloke Our plan of record right now is that we will have resource-level docs as our primary entrypoint from the docs site - but that these will link into language-specific docs for details of individual language projections.  That means we will still need .NET API docs to be available.  Is the plan to publish and host those on `pulumi.com/docs` (as we currently do for TypeScript and Python) or to host somewhere externally (as we do for Go)?

praneetloke:
My understanding is these will need to be hosted by us. I haven't looked into this much yet, but from what I know there isn't a central place for dotnet docs similar to pkg.go.dev.

lukehoban:
I think @mikhailshilkov had a WIP on this - so I'll assign it to him.  I believe he was just blocked on: 
1. Us deciding the plan for resoruce-model and/or API model docs (we've now decided on "both")
2. Issues with breaking Pulumify when adding thousands of new files (we should not block on this - if Pulumify can't yet handle the content we need to host on this site - we should stop using Pulumify until it can be fixed to support it)

praneetloke:
> Issues with breaking Pulumify when adding thousands of new files (we should not block on this - if Pulumify can't yet handle the content we need to host on this site - we should stop using Pulumify until it can be fixed to support it)

Yes, this is a problem without adding the .NET docs as well. Just with the new resource-level docs, Pulumify does not work in its current form. See this [comment](https://github.com/pulumi/docs/pull/2551#issuecomment-592175142) in this prototype PR opened by Justin.

mikhailshilkov:
Another challenge is related to styling these pages. How much time do we want to spend on making them look and feel as the rest of the site? If we want ~100% flexibility, it might make sense to write our own tool akin to ones for TS/Python instead of trying to fit the default HTML.

Is there somebody with better HTML/CSS skills than me that I could discuss this with?

praneetloke:
@mikhailshilkov let me connect with you offline.

mikhailshilkov:
An early PR: https://github.com/pulumi/docs/pull/2172
that shows the preview:
- http://content-bucket-a3491d9.s3-website-us-east-1.amazonaws.com/docs/reference/pkg/dotnet/Pulumi/Pulumi.StackReference.html
[...]

PR #2580, opened 2020-03-04
Introduce DocFX for .NET API reference

mikhailshilkov:
Part of #2137, this PR introduces DocFX to generate .NET API Reference docs for our NuGet packages. For now, only the core "Pulumi" docs are generated.


praneetloke:
@cnunciato wondering if we should merge this without anything linking to these auto-generated docs page yet?

@mikhailshilkov clicking on the "Improve this Doc" links navigates to some weird URL. Should we do something to enable that functionality? Or perhaps disable it?

Screenshot_2020-03-04 Class ComponentResource

Screenshot_2020-03-04 Class Output

cnunciato:
@praneetloke I'm not sure I'd merge it without something linking to it, but this is probably pretty close. (Thanks, @mikhailshilkov!) I'll take a closer look at it tomorrow.

praneetloke:
To be clear, this PR does have the appropriate linking to the generated docs. I was just wondering if we should not link to it until we have #2585 in place as well. But I'll defer that to you.

mikhailshilkov:
> clicking on the "Improve this Doc" links navigates to some weird URL. Should we do something to enable that functionality? Or perhaps disable it?

Nice catch! I looked at what it's supposed to do and it seems that it won't have any value for us. I did not find a way to disable it with configuration while keeping `View Source` button, so I modified the HTML template to remove `Improve this Doc`.

praneetloke:
Ping.

docfx/pulumi-template/ManagedReference.common.js
```diff
@@ -0,0 +1,258 @@
+// Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
lukehoban: It's a bit of a shame we have to copy in all these files - are they not available somewhere we can reference instead?  If vendoring them like this recommended?

Assuming we haven't made any changes - perhaps add a README in the folder with details about where this all comes from?

docfx/pulumi-template/layout/_master.tmpl
```diff
@@ -0,0 +1,55 @@
+{{!Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE file in the project root for full license information.}}
```
lukehoban: If this is for our code - should we add our Copyright instead?

docfx/pulumi-template/ManagedReference.common.js
mikhailshilkov: Great point! I found out that we can do overrides with themes, so I removed all files that were untouched and left the ones I changed.

docfx/pulumi-template/layout/_master.tmpl
mikhailshilkov: It looks like we don't have copyright in our templates/styles in other parts of docs, so I just removed those top lines. Let me know if it's a wrong thing to do.

PR #2592, opened 2020-03-05
Update eks blog intro to be less time constrained

zchase:
This blog was ranking well for "aws eks" but has recently taken a hit in rankings. The working theory is that it is due to a high bounce rate. Our first experiment in bringing down the bounce rate is to adjust the intro paragraph of the blog post to be less time-constrained (tied to re:Invent 2019).

Current bounce rate: ~88%

### DIFF
--- a/content/blog/aws-eks-managed-nodes-fargate/index.md
+++ b/content/blog/aws-eks-managed-nodes-fargate/index.md
@@ -8,7 +8,7 @@ date: "2019-12-05"
 meta_image: "pulumi-eks-fargate.png"
 ---
 
-At AWS re:Invent this week, Amazon highlighted two new features that simplify scaling your Elastic Kubernetes Service (EKS) clusters: Managed Node Groups and Fargate. We're happy to announce that we've integrated support for both, making our existing EKS support even easier to use than before. The result is a great spectrum of options for managing your cluster's compute &mdash; offering productivity, flexibility, and control, based on your needs.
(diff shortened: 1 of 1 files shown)


zchase:
@joeduffy I don't think a deploy button will help bounce because if the person is a "bouncer" then it is likely they don't find the content relevant to their search. If they don't find it relevant I don't think they will want to deploy an EKS cluster.

Created a new issue here to run bounce reducing experiments on our top 3 blogs of 2020: https://github.com/pulumi/docs/issues/2619

joeduffy:
My thinking was they _did_ want to create an EKS cluster, but they landed on a page that just told them "blah blah blah new EKS features blah blah blah," without a crisp CTA that told them how to provision one. I may very well be wrong, and any improved understanding of bounce reasons would be great.

zchase:
@joeduffy Hmm that is an interesting point. I think I was a little too quick to the gun with my last comment and based it on a personal assumption of what a user is looking for. Maybe they are ready to deploy a cluster and informational blog isn't what they are looking for. I think this something we should test and get some results on.

I'll open a PR with CTA about deploying a cluster. 

The copy change and the CTA can be tested in parallel. Lower bounce rate + increased pages/session means the CTA is working, lower bounce rate + increased avg time on page means the copy change is working.

Thanks for bringing this up, it definitely helped me think about our approach to the bounce issue a bit differently 🙌

PR #2593, opened 2020-03-05
SEO optimization for KW: "Kubernetes multi cloud"

97thFloor:
@cnunciato @zchase since this one was so technical, I've included below the main keyword, but also the language that we drew from our semantic analysis. I tried to include this where possible, but I'm including the rest in case you see any easy opportunities to include it that I missed. 

The main topics that I think we're missing on this page appear to be regarding data security and containers. If you have any information on those topics and how they relate to this one, feel free to send it my way and I can write up a paragraph or couple of sentences — or if you think it's just easier to add it yourself, by all means!

**Primary Keyword:** Kubernetes Multi Cloud
Language from **Semantic Analysis**: platform, container, multicloud, data, security, public, application, clouds, service, multiple, tools, containership

### DIFF
--- a/content/blog/multicloud-app/index.md
+++ b/content/blog/multicloud-app/index.md
@@ -8,9 +8,9 @@ date: "2019-08-14"
 meta_image: "multicloud.png"
 ---
 
-Kubernetes clusters from the managed offerings of AWS EKS,
-Azure AKS, and GCP GKE all vary in configuration, management, and resource
-properties. This variance creates unnecessary complexity in cluster provisioning and app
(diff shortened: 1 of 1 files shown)


zchase:
@metral Our SEO firm, 97thFloor, provided us with some optimizations for the multi-cloud blog you wrote. Would you mind giving it a quick review and make sure everything still makes sense? The goal of the optimization is to get more keywords into the blog post so it will rank higher and for more search terms.

cnunciato:
🎉 Thanks @metral!

PR #2598, opened 2020-03-06
Removed canceled events

zchase:
This PR adds an `unlisted` flag to events. This is so we can hide events that have been canceled.

### DIFF
--- a/archetypes/event/index.md
+++ b/archetypes/event/index.md
@@ -2,6 +2,9 @@
 # Name of the event.
 title: "Name of the Event"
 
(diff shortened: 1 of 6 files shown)


isaac-pulumi:
> LGTM.
> 
> Should the pages for these events also have content updates mentioning that the event is now canceled, ideally with a link to the actual event page with explanation for the cancellation?

I'm okay with just pulling these without explanation.  It avoids 'burying' the events that are still planned with events that don't exist anymore.

Issue #2517, opened 2020-02-24
"Copy to clipboard" button doesn't work on multi-line command with backslash


clstokes:
### Problem description

If a snippet in our docs contains a backslash (`\`) (or the line ends with one?), only the first line is copied, not the full snippet.

For the snippet at https://www.pulumi.com/docs/intro/concepts/config/#aws-key-management-service-kms:
```
$ pulumi stack init my-stack \
    --secrets-provider="awskms://1234abcd-12ab-34cd-56ef-1234567890ab?region=us-east-1"
```

This is what gets copied:
```
pulumi stack init my-stack \
```

I'm not 100% sure the backslash is the issue, but I've only noticed the issue on multi-line commands when a backslash is used.

### Examples

- https://www.pulumi.com/docs/intro/concepts/config/#aws-key-management-service-kms
- https://www.pulumi.com/docs/intro/concepts/config/#azure-key-vault
- https://www.pulumi.com/docs/intro/concepts/config/#google-cloud-key-management-service-kms
- https://www.pulumi.com/docs/intro/concepts/config/#hashicorp-vault-transit-secrets-engine

### Suggestions for a fix

Copy the whole snippet.

joeduffy:
Ahh, just hit this one too:

```
$ mkdir yaml
$pushd yaml
$ curl -L --remote-name \
    https://raw.githubusercontent.com/kubernetes/examples/master/guestbook/{frontend-deployment,frontend-service,redis-master-deployment,redis-master-service,redis-slave-deployment,redis-slave-service}.yaml
$ popd yaml
```

This gets copied as:

`mkdir yaml && pushd yaml && curl -L --remote-name \`

PR #2606, opened 2020-03-08
Support shell line continuations in copy button

joeduffy:
Right now if you have shell continuations, the copy button doesn't do the right thing. E.g.

```
$ mkdir yaml
$pushd yaml
$ curl -L --remote-name \
    https://raw.githubusercontent.com/kubernetes/examples/master/guestbook/{frontend-deployment,frontend-service,redis-master-deployment,redis-master-service,redis-slave-deployment,redis-slave-service}.yaml
$ popd yaml
```

Gets copied as `mkdir yaml && pushd yaml && curl -L --remote-name \`.

 I hit this during some user guide development so I spent a couple minutes trying to fix. Turned out to be not too crazy. Fixes https://github.com/pulumi/docs/issues/2517.

### DIFF
--- a/assets/js/copybutton.js
+++ b/assets/js/copybutton.js
@@ -16,13 +16,15 @@ $(function() {
 
         // For command line code snippets:
         //   1. Strip the prompt from any lines that are prefixed with it
-        //   2. Discard subsequent lines that don't start with the prompt (i.e. output)
+        //   2. Discard subsequent lines that don't start with the prompt (i.e. output),
+        //      making sure to respect line continuation charcters.
         //   3. Combine multiple lines that start with a prompt into a single line, e.g.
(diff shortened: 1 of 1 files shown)


assets/js/copybutton.js
```diff
@@ -16,13 +16,15 @@ $(function() {
 
         // For command line code snippets:
         //   1. Strip the prompt from any lines that are prefixed with it
-        //   2. Discard subsequent lines that don't start with the prompt (i.e. output)
+        //   2. Discard subsequent lines that don't start with the prompt (i.e. output),
+        //      making sure to respect line continuation charcters.
```
justinvp: Nit:

```suggestion
        //      making sure to respect line continuation characters.
```

joeduffy: Oops, overlooked this, will fix in a separate commit.

PR #2607, opened 2020-03-08
Unscramble the cloud providers ToC

joeduffy:
I noticed that the ordering on the left-hand navigation for the
cloud providers section got scrambled at some point. I suspect we
just weren't watching the "weight" specifiers closely as we added
some of them (if it is intentional, it wasn't apparent to me).

I've changed this to have two clumps of cloud providers:

* First, the "primary four" -- AWS, Azure, GCP, and Kubernetes --
  that sorts at the very top (in alphabetical order).

* Second, the rest.

I gave them the same weights and let the ToC rendering logic
alphabetize rather than needing to maintain them by hand, which
would get quite tedious.

### DIFF
--- a/content/docs/intro/cloud-providers/aiven/_index.md
+++ b/content/docs/intro/cloud-providers/aiven/_index.md
@@ -5,7 +5,7 @@ menu:
   intro:
     parent: cloud-providers
     identifier: clouds-aiven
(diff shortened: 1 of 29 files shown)


joeduffy:
Before:

image

After:

image

PR #2613, opened 2020-03-10
WIP: Managing AWS Credentials on CI/CD

spara:


content/blog/managing-aws-credentials-on-cicd/index.md
```diff
@@ -0,0 +1,231 @@
+---
+title: "Managing Aws Credentials on CI/CD"
+date: 2020-03-10
+meta_desc:
+meta_image: meta.png
+authors:
+    - sophia-parafina
+tags:
+    - change-me
```
chrsmith: Just flagging this in case we miss it later...

content/blog/managing-aws-credentials-on-cicd/index.md
```diff
@@ -0,0 +1,231 @@
...
+
+Take a deep breath. It’s going to be OK. You can securely provide AWS credentials to your CI/CD system and practice continuous delivery. This post covers all the things you need to know to do it.
+
+The goal of this article is to have a clear understanding of how to properly use Pulumi to update AWS infrastructure within a hosted CI/CD service like [CircleCI](https://circleci.com), [GitLab CI](https://about.gitlab.com/product/continuous-integration/), or [Travis CI](https://travis-ci.org).
+
+> **NOTE:** These recommendations do not apply ff you are running your own CI/CD system within your AWS account, e.g., running a Jenkins server on an EC2 instance or using [AWS CodeDeploy](https://aws.amazon.com/codedeploy/). There are established methods to [assume IAM Roles from applications running on an EC2 instance](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2.html). In that case, refer to the documentation specific to your CI/CD system for how to handle AWS credentials.
```
chrsmith:
```suggestion
> **NOTE:** These recommendations do not apply if you are running your own CI/CD system within your AWS account, e.g., running a Jenkins server on an EC2 instance or using [AWS CodeDeploy](https://aws.amazon.com/codedeploy/). There are established methods to [assume IAM Roles from applications running on an EC2 instance](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2.html). In that case, refer to the documentation specific to your CI/CD system for how to handle AWS credentials.
```

chrsmith: "do not apply **if** you are"

content/blog/managing-aws-credentials-on-cicd/index.md
```diff
@@ -0,0 +1,231 @@
...
+
+Depending on your specific environment, needs, and constraints, there may be a better approach for you to take. These recommendations describe a general “one-size fits most” approach that requires a minimal amount of work to configure and maintain.
+
+The approach includes the following:
+
+- Create a dedicated IAM User for your CI/CD
```
chrsmith: Nit: "CI/CD _system_". It is odd to just say "CI/CD" since the next bullet says "CI/CD system".

content/blog/managing-aws-credentials-on-cicd/index.md
```diff
@@ -0,0 +1,231 @@
...
+- Automate rotating and revoking that user’s credentials
+- During deployments, exchange the IAM User credentials for more privileged IAM Role credentials
+
+The process may sound complicated, but we’ll cover each step in detail.  We’ll point out additional ways to enhance your security, such as using Pulumi’s support for client-side encryption to limit access to sensitive information.
+
+## Create a dedicated IAM User for your CI/CD
+
+The first step for securely automating CI/CD on AWS is to create a dedicated IAM User for use in your CI/CD pipelines.
+
+By default, a new IAM User does not have any permissions associated with it, meaning it can only authenticate itself with its access key and secret. Let’s walk through how to grant this IAM User the permissions necessary to update your cloud infrastructure later. For now, we limit this user to have as little access as possible.
```
chrsmith: This paragraph is a little confusing. Perhaps rephrase it? Or drop the part about "an IAM User without any permissions cannot do anything"?

I think the main things to emphasize are:

- You obviously want to limit the permissions as narrowly as possible.
- This IAM User will _later_ assume an IAM Role that will have more broad permissions, when it actually goes and performs a deployment.

So in the following code snippet, we're just creating an IAM User that will be used in CI/CD jobs.

content/blog/managing-aws-credentials-on-cicd/index.md
```diff
@@ -0,0 +1,231 @@
...
+
+The following code snippet shows how to create an IAM User specifically for your CI/CD jobs. 
+
+```ts
+const aws = require("@pulumi/aws");
```
chrsmith: This snippet doesn't quite look right. (So apologies if I was the one that wrote this in the first place!)

Unless I am mistaken, you just need the following:

```ts
// First create the user
// ... probably want to give it a name and tag, right?
const user = new aws.iam.User("cicdUser", {
    name: "cicd-bot",
    tags: {
        "purpose": "Account for performing deployments for jobs running on <your CI/CD service>",
    },
);
```

Then, the role policy you want to attach to that IAM user -- i.e. the set of permissions that are granted to that user -- is _just_ the `sts:AssumeRole` permission. (Although if there is an `sts:GetCallerIdentity` that might be useful too, so that user can confirm they are who they think thy are!)

AWS's documentation recommends however that role policies are attached to _groups_ and not users. I'd suggest calling that out in a code comment, and just updating the code as well. Something like:

```ts
const group = new aws.iam.Group("cicdBots", {
    name: "cicd-bots",

   // Just put the new IAM User in this group
    members: [ iamUser ],

    tags: {
      "purpose": "IAM Group containing all of the IAM Users who perform deployments in the CI/CD system".
    },
);
```

... and then (if I am not mistaken), you would attach the IAM RolePolicy to the IAM _Group_ and not the IAM User. Then the user would be given access to `sts:AssumeRole` by virtue of its membership in the IAM Group.

content/blog/managing-aws-credentials-on-cicd/index.md
```diff
@@ -0,0 +1,231 @@
...
+    users: [user],
+    groups: [group],
+    roles: [role],
+    policyArn: policy.arn
+});
+```
+
+### Isolating with Different Accounts
+
+A common and **recommended** practice is to use multiple AWS accounts to isolate resources and users from one another. For example, if your testing and production environments are in different AWS accounts, it would be difficult for a severe bug in the testing environment to harm your customer data.
```
chrsmith: Nit: The emphasis on recommended seems out of place. Is there a particular white paper or something we should be linking to?

content/blog/managing-aws-credentials-on-cicd/index.md
```diff
@@ -0,0 +1,231 @@
...
+
+Keep this in mind as we configure this IAM User and later grant them IAM Roles to perform cloud deployments. Creating this specific IAM User to CI/CD jobs with its AWS account means that there are no resources to modify, in addition to not having permissions to do anything.
```
chrsmith: "jobs with its AWS account" doesn't quite read right.

Are we trying to say: "You should separate your testing and production instances in separate AWS accounts. This separation ensures that a bug in your testing environment cannot harm customer data."?

Are we trying to make an additional claim about the AWS account where the IAM User we are using for CI/CD should be created? We could make that claim, but the details are complicated and it's a pain. So I'd just mention it in a link. e.g. "It isn't a requirement that the IAM User you will use to deploy your infrastructure is in the same AWS account that the cloud resources are located in. For example, it is possible for the IAM User in your "testing" account to perform a deployment of the application housed in your "production" environment.

For the rest of this post, we'll assume that the IAM user is in the same AWS account. But for more information about IAM and cross-account access, see [the documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_common-scenarios_third-party.html). (The same approach described here, assuming an IAM Role, is used.)

content/blog/managing-aws-credentials-on-cicd/index.md
```diff
@@ -0,0 +1,231 @@
...
+
+Providing AWS access keys to any system is dangerous because of the risk that a bad actor could acquire those keys to do something nefarious. Even if you do trust your CI/CD service provider to store these credentials properly, the keys could be inadvertently exposed in debugging output, system logs, or some other place they could be disclosed.
+
+However, if you want to automate changes to your AWS infrastructure from a 3rd party CI/CD service, you must provide those credentials. Here is where following security best practices come into play.
+
+If we do need to provide credentials to a 3rd party, rather than hoping it will be 100% secure forever (which is impossible), we can instead make those credentials volatile. If we regularly invalidate and rotate the credentials given supplied to CI/CD system, we can dramatically reduce the impact of any accidental disclosure. Since even if the IAM User’s credentials were leaked, by the time they were discovered and used, they would no longer be valid. We’ll go into how to manage key rotation shortly.
```
chrsmith: Just to tie this section to the overall theme of "this is how you setup using Pulumi to deploy apps on AWS via CI/CD", you might want a final paragraph making it super-clear what the action is:

> So the next step is to take the credentials for the IAM user we created earlier, and enter those into your CI/CD system so that they are available during the next build.

content/blog/managing-aws-credentials-on-cicd/index.md
```diff
@@ -0,0 +1,231 @@
...
+“Just create a new IAM User credential, log into the CI/CD system, and then update the job’s settings” sounds simple enough. But it’s likely going to be tedious, and if you need to repeat the process for dozens if not hundreds of CI/CD pipelines, you need a better solution. Unless you start with an automated approach for key rotation, you are unlikely to add one later.
+
+The AWS security blog describes how to [rotate access keys for IAM Users](https://aws.amazon.com/blogs/security/how-to-rotate-access-keys-for-iam-users/).
+
+1. Create a second access key in addition to the one in use.
+1. Update all your applications to use the new access key and validate that the applications are working.
+1. Change the state of the previous access key to inactive.
+1. Validate that your applications are still working as expected.
+1. Delete the inactive access key.
+
```
chrsmith: It seems like I first need to dust off the code (collecting dust over at https://github.com/pulumi/autobots/pull/7) and publish it somewhere. (Maybe in the examples repo?) And include snippets or a link in the text here.

Ultimately this step needs to be a "turn key" type experience to be useful.

content/blog/managing-aws-credentials-on-cicd/index.md
```diff
@@ -0,0 +1,231 @@
...
+## During deployments, exchange the IAM user credentials for a more privileged IAM role
+
+Now that we have a regularly rotated, low-privileged credential to use during deployments, the next step is to use “assume role” to exchange those IAM User credentials for an IAM Role credential. This sounds confusing at first. Why exchange one credential for another? Why not just use the same credentials? In other words, what is an IAM Role, and why is this so much better than an IAM User? The credentials used to assume an IAM Role are only valid for a short period (minutes to hours). After a fixed duration, there is no threat if disclosed.
```
chrsmith: I'd suggest putting a paragraph break before "The credentials used to assume..." since it marks the transition to explaining the difference between IAM Users and IAM Roles.

content/blog/managing-aws-credentials-on-cicd/index.md
```diff
@@ -0,0 +1,231 @@
...
+
+Earlier, we mitigate potentially exposing IAM User credentials by using our CI/CD provider’s ability to store secret values and automatically rotating credentials. Using the IAM Role credentials adds another layer of security, but within the context of the machine itself during the deployment.
+
+So how do you use these credentials? The AWS Secure Token Service can do this, but in practice, you’ll want to use a lightweight tool to do this.
+
+### Configuring AWS Security Group
```
chrsmith: Perhaps this was added by mistake? I don't think this ("AWS security groups") applies to this post.

Instead we'd probably want to explain how the AWS CLI tool can be used to assume a role, (e.g. https://aws.amazon.com/premiumsupport/knowledge-center/iam-assume-role-cli/). (Another tool that is perhaps easier to the task is the https://github.com/remind101/assume-role tool.)

So that when you go to deploy your Pulumi application, you are doing so as the "higher-privledge IAM Role" rather than the "no access IAM User account".

content/blog/managing-aws-credentials-on-cicd/index.md
```diff
@@ -0,0 +1,231 @@
...
+1. Change the state of the previous access key to inactive.
+1. Validate that your applications are still working as expected.
+1. Delete the inactive access key.
+
+## During deployments, exchange the IAM user credentials for a more privileged IAM role
+
+Now that we have a regularly rotated, low-privileged credential to use during deployments, the next step is to use “assume role” to exchange those IAM User credentials for an IAM Role credential. This sounds confusing at first. Why exchange one credential for another? Why not just use the same credentials? In other words, what is an IAM Role, and why is this so much better than an IAM User? The credentials used to assume an IAM Role are only valid for a short period (minutes to hours). After a fixed duration, there is no threat if disclosed.
+
+Earlier, we mitigate potentially exposing IAM User credentials by using our CI/CD provider’s ability to store secret values and automatically rotating credentials. Using the IAM Role credentials adds another layer of security, but within the context of the machine itself during the deployment.
+
```
chrsmith: We probably want to add a note explaining that it is the IAM Role that needs the permissions to actually deploy your Pulumi program. (e.g. access to the required APIs in your production account.)

However, we still want to be careful in the permissions granted to the IAM Role policy. For example, explicitly disallowing the `iam:*` permission is probably a good start. (Link to this [blog post](https://know.bishopfox.com/research/privilege-escalation-in-aws) that goes into the details of how having access to some IAM APIs would allow a compromised credential to fully take over an AWS account!)

content/blog/managing-aws-credentials-on-cicd/index.md
```diff
@@ -0,0 +1,231 @@
...
+So far, we’ve written extensively about securing AWS access tokens, but aren’t Pulumi credentials sensitive as well? And wouldn’t they need to be secure?
+
+Yes. Though as much as we like to think our product provides information that eases deployments, understanding the history, state, and topology of your cloud resources is not as important as having access to your cloud resources themselves.
+
+You’ll want to follow the same approach we described for AWS credentials with your Pulumi Access tokens. We support revoking those credentials in the Pulumi UI, and you'll want to recreate and rotate them regularly as well.
+
+It’s best practice to store them using AWS SecretsManager to [load and store those credentials](https://aws.amazon.com/secrets-manager/).
+
+Here is an example for reading/writing a Pulumi Access Token from AWS Secrets Manager. This, of course, assumes that the current AWS credentials (e.g., the result of assume-role) have access. Create an access token using the Pulumi console and copy it to a file.
```
chrsmith: Maybe call out that you need to have the IAM User assume role into the more privileged IAM Role account in order to read the secret from secrets manager? (Since it isn't obviously "who" should have access to this secret.)

content/blog/managing-aws-credentials-on-cicd/index.md
spara: I've made the changes as suggested, please check for correctness.

content/blog/managing-aws-credentials-on-cicd/index.md
spara: I want to split the post into two parts. This part discusses what needs to happen and the second part would be an implementation example.

spara: let's save this for a followup post

content/blog/managing-aws-credentials-on-cicd/index.md
spara: deleted

chrsmith:
@spara , please take a look at https://github.com/pulumi/docs/pull/2613/commits/1c6c92323ff03f18d9b5076c623e342a0a0ebe79. I pulled the post locally and tried to polish up the very first section (just covering the creation of the IAM User and Group).

That should be a reasonable (if not small) first post in the series. I'll start work on the second post right now, which will cover:

- Provide the IAM User’s credentials to your CI/CD system
- Automating IAM credential rotation using Pulumi

And then, the last part would be:

- Assuming IAM Roles for performing updates
- Securing sensitive data using Pulumi
- Comparison with using hosted sercret managers

Does that sound like a plan to you?

PR #2626, opened 2020-03-10
Add Stencil.js for web components

cnunciato:
This change adds [Stencil.js](https://stenciljs.com/), a web-component compiler, to the website, along with a couple of new components that we'll use for the language chooser. (It does not yet replace the language chooser, though; I'll submit that in a separate PR after this one merges, just to limit the scope of this first one.)

The goal here is to:

* Move away from writing jQuery and inline scripts, and toward writing proper [web components](https://developer.mozilla.org/en-US/docs/Web/Web_Components) for functional elements like the language chooser, on-page TOCs, event filtering, and so on.

* Move away from reading from, and writing to, cookies and localStorage for persistence, and instead use a Redux store for persistence (serializing to and from localStorage as appropriate).

Components are written in TypeScript and are unit- and functionally testable, although I haven't yet wired the tests for these first components up to CI. Will also do that in a follow-up PR.

See the README in `components/` for more details.


components/src/components/chooser/chooser.e2e.ts
```diff
@@ -0,0 +1,15 @@
+import { newE2EPage } from "@stencil/core/testing";
+
+describe("pulumi-chooser", () => {
+
+    it("renders a list items", async () => {
```
cnunciato:
```suggestion
    it("renders a list of items", async () => {
```

components/src/components/choosable/choosable.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { Choosable } from './choosable';
+
+describe("pulumi-chooser", () => {
```
cnunciato:
```suggestion
describe("pulumi-choosable", () => {
```

layouts/shortcodes/cloudchoose.html
```diff
@@ -1,8 +1,8 @@
 <div class="mdl-tabs cloud-tab-tabs">
     <div class="mdl-tabs__tab-dir cloud-tab-tabstrip">
-        <a data-track="cloudchoose-aws" class="mdl-tabs__tab cloud-tab is-active">AWS</a>
-        <a data-track="cloudchoose-azure" class="mdl-tabs__tab cloud-tab">Azure</a>
-        <a data-track="cloudchoose-gcp" class="mdl-tabs__tab cloud-tab">GCP</a>
+        <a data-choice="aws" data-track="cloudchoose-aws" class="mdl-tabs__tab cloud-tab">AWS</a>
```
cnunciato: I just figured I'd go ahead and add these, because I started to noticed inconsistent behavior once I'd integrated the new code.

layouts/shortcodes/install-pulumi.html
```diff
@@ -5,9 +5,9 @@
 */}}
 <div class="mdl-tabs os-tab-tabs">
     <div class="mdl-tabs__tab-dir os-tab-tabstrip">
-        <a data-track="oschoose-macOs" class="mdl-tabs__tab os-tab is-active">macOS</a>
```
cnunciato: These are causing a brief blink on reload, so since we seem to be setting them in all cases, it seemed fine to remove them.

layouts/partials/scripts.html
```diff
@@ -26,3 +26,13 @@
...
 <script src="{{ $js.RelPermalink }}" integrity="{{ $js.Data.Integrity }}"></script>
+
+<!--
+    These two scripts are provided separately in order to allow browsers to decide which
+    component bundle to load. Modern browsers that support web components will load the
+    smaller bundle. Others will load the (much larger) polyfill. Because they're ultimately
+    responsible for loading child components, whose filenames will contain content-based
+    hashes, they must not be cached.
+-->
+<script type="module" src="/js/components/components.esm.js" defer></script>
```
cnunciato: Note that in modern browsers, the payload here will be _very_ small: non-gzipped, the request for `components.esm.js` is 252 bytes, and the subsequent component download (which can be forever-cached) is ~11K.

components/package-lock.json
```diff
@@ -0,0 +1,5334 @@
+{
```
justinvp: This is `npm`'s lock file, right? I think we can delete it since we're using `yarn` in this repo. If so, let's add it to the `.gitignore`.

components/src/index.html
```diff
@@ -0,0 +1,71 @@
+<!DOCTYPE html>
```
justinvp: Just asking for my own understanding: This file is served by `make serve-components`? And if I wanted to test out a new component, I'd add it to this file?

components/package-lock.json
cnunciato: Ah yes, good spot, thanks.

components/src/index.html
cnunciato: Yes, that's right -- and now I see that I omitted this slightly relevant detail from the README. 😂 I'll add it.

PR #2633, opened 2020-03-11
Add CTA to top of EKS blog post as a bounce reducing experiment

zchase:
This is a continuation of https://github.com/pulumi/docs/pull/2592. This PR adds a callout at the top of the blog post directing users to our EKS doc with code examples. 

deployEKSCTA

### DIFF
--- a/content/blog/aws-eks-managed-nodes-fargate/index.md
+++ b/content/blog/aws-eks-managed-nodes-fargate/index.md
@@ -17,7 +17,7 @@ summary: |
 
 {{% notes %}}
     Looking for information on how to deploy an EKS cluster? Checkout our
-    <a href="{{< relref "/docs/guides/crosswalk/aws/eks" >}}">AWS Elastic Kubernetes Service</a>.
+    <a href="{{< relref "/docs/guides/crosswalk/aws/eks" >}}">AWS Elastic Kubernetes Service</a> documentation.
 {{% /notes %}}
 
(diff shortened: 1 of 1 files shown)


content/blog/aws-eks-managed-nodes-fargate/index.md
```diff
@@ -8,6 +8,17 @@ date: "2019-12-05"
 meta_image: "pulumi-eks-fargate.png"
 ---
 
+<div class="bg-blue-100 border-t-4 border-blue-500 rounded-b text-blue-900 px-4 py-3 shadow-md my-5" role="alert">
```
cnunciato: Couple lil' things:

* In the list view, this shows up as:

    image

    ...which I don't think we want. But if you set an explicit `summary` param to the post's frontmatter with the text you want to appear in the summary, and then remove the `<!--more-->` token, [Hugo will use the frontmatter summary instead](https://gohugo.io/content-management/summaries/#summary-selection-order):

        ....
        summary: |
            AWS Elastic Kubernetes Service (EKS) provides a range of performance and control for
            dynamically scaling your Kubernetes clusters, including Managed Node Groups, Fargate, and
            Manually-Managed Node Groups in EC2. In this post, we'll see how to use each of these
            compute options, and when to prefer one over the other in order to maximize productivity,
            flexibility, and control, based on your needs.
        ...

* And then also, for the note, we already have a shortcode for this sort of thing, so we should use that instead. Something like this should work:

        {{% notes %}}
        Looking for information on how to deploy an EKS cluster? Checkout our
        [AWS Elastic Kubernetes Service]({{< relref "/docs/guides/crosswalk/aws/eks" >}}) documentation.
        {{% /notes %}}

    ... although it looks like the `notes` shortcode isn't rendering markdown (🤔), so if you wouldn't mind changing the content in that shortcode from `{{ .Inner }}` to `{{ .Inner | markdownify }}`, that'd make it work as I think it's intended to.

Issue #2555, opened 2020-02-27
Generate TF-based provider docs using the schema-based doc generation process


praneetloke:
* [x] Update doc generation scripts to run new doc gen for resource providers (PR #2641)
* [x] AWS
* [x] Azure (PR https://github.com/pulumi/docs/pull/2746)
* [x] Azure AD (PR https://github.com/pulumi/docs/pull/2746)
* [x] GCP (PR https://github.com/pulumi/docs/pull/2746)
...

Issue #2587, opened 2020-03-04
Layout improvements for the new auto-generated resource-level doc


praneetloke:
Some early thoughts that were discussed for this were:

* [ ] Sticky breadcrumbs
* [x] Improving the side-nav
* [ ] Reducing the repetition of the language chooser
* [ ] Fill in appropriate front matter on each generated page
  * The `header.tmpl` that is injected into every page contains just the `title` in the front-matter right now.

Note that the templates that generate the markdown files are in the `pulumi/pulumi` repo under `pkg/codegen/docs`.

cnunciato:
Closing this in favor of what's captured in https://github.com/pulumi/docs/issues/2816.

Issue #2586, opened 2020-03-04
Update the current API reference index page with resource docs information


praneetloke:
~~We would like to have a sibling category under the Reference heading on our docs that will be the main index page, similar to our current index page for our language-specific docs at `docs/reference/pkg/`. So something like `docs/reference/providers/` (the word `Providers` is just used for testing and not necessarily suggestive of the final name of such an index)?~~

~~Here's a screenshot of what I have been using for my testing:~~

![Screenshot_2020-03-04 Providers](https://user-images.githubusercontent.com/1466314/75910949-9c650380-5e03-11ea-90bf-a87d189bfcd8.png)

--------------

:point_up:  that was the original issue description. Please see this [comment](https://github.com/pulumi/docs/issues/2586#issuecomment-599835166) for the updated plan of action for this work item.

praneetloke:
Adding some notes as a follow-up to a discussion Justin, Christian and I had on 3/13/2020.

* The new resource docs will be placed under the existing `content/docs/reference/pkg` directory, directly as siblings to the already existing `nodejs` and `python` docs and soon the dotnet API docs as well.
* The existing top-level API reference index page (`content/docs/reference/pkg/_index.md`) will be updated to make the resource-centric docs the more prominent and recommended docs for a user to navigate. Although the links to per-language API docs will be maintained on the same page in some form.

So in summary, users can expect to find resource docs for a given cloud provider at `https://www.pulumi.com/docs/reference/pkg/<cloud_provider>`. For example, take AWS, a user would be able to visit the new AWS resource-centric docs by going to `https://www.pulumi.com/docs/reference/pkg/aws` (or going to `https://www.pulumi.com/docs/reference/pkg` and clicking on AWS).

#### Versioning

At this time Pulumi docs always show the docs of the latest version of a Pulumi provider that is released. This trend will continue for the time being. In the future, should versioned-docs be made available, the URL structure could easily be adjusted to add a version number _after_ the cloud_provider name in the URL. For example, `https://www.pulumi.com/docs/reference/pkg/<cloud_provider>/<version>`.

PR #2641, opened 2020-03-12
Add a tool to generate resource docs using the Pulumi schema

praneetloke:
### Proposed changes
Add a resource doc generator tool. This tool generates a schema using the provider's tfgen and then calls on the new schema-based docs code generator to generate the docs.

~~Although this tool is currently generating the schema from the provider, note that there is some pending work to actually update all of the downstream providers to check-in a `schema.json` file, which this tool could simply import, therefore skipping the part about generating the schema, which is where bulk of the time is spent. See this [comment](https://github.com/pulumi/pulumi-terraform-bridge/pull/128#issuecomment-597178553) for relevant info.~~

UPDATE: This PR has been updated to read the `schema.json` file in the provider repo and use that to generate the docs.

N/A

### Related issues (optional)
Part of #2555.


content/docs/reference/pkg/aws/inspector/getrulespackages.md
```diff
@@ -0,0 +1,260 @@
...
+            <th>Description</th>
+        </tr>
+    </thead>
+    <tbody>
+    
+        <tr>
+            <td class="align-top">arns</td>
+            <td class="align-top">
+                
+                <code>list[string]</code>
```
justinvp: Also, when specifying a typehint for a list in Python, the `List` is generally upper-case, e.g. `List[str]`.

content/docs/reference/pkg/aws/s3/bucket.md
```diff
@@ -0,0 +1,10931 @@
...
+
+            
+            </td>
+        </tr>
+    
+        <tr>
+            <td class="align-top">object_<wbr>lock_<wbr>configuration</td>
+            <td class="align-top">
+                
+                <code><a href="#bucketobjectlockconfiguration">dict{bucket_<wbr>object_<wbr>lock_<wbr>configuration}</a></code>
```
justinvp: This syntax for dictionaries doesn't feel very Pythony, but I don't have a great alternative to suggest. I guess technically: `Dict[str, Any]`.

If we bumped our min version of Python from 3.6 to 3.8, we could consider introducing `TypedDict` for these, which means, the nested type could be emitted for Python as something like:

```python
class BucketObjectLockConfiguration(TypedDict):
    objectLockEnabled: str
    rule: Optional[BucketObjectLockConfigurationRule]
```

or

```python
class BucketObjectLockConfiguration(TypedDict, total=False):
    objectLockEnabled: str
    rule: BucketObjectLockConfigurationRule
```

And then we could just refer to it here as `BucketObjectLockConfiguration`.

content/docs/reference/pkg/aws/s3/bucket.md
```diff
@@ -0,0 +1,10931 @@
...
+        </tr>
+    
+        <tr>
+            <td class="align-top">Tags</td>
+            <td class="align-top">
+                
+                <code>Dictionary&lt;string, object&gt;?</code>
+            </td>
+            <td class="align-top">{{% md %}} A mapping of tags that identifies subset of objects to which the rule applies.
+The rule applies only to objects having all t
```
justinvp:
```suggestion
                <code>Dict[str, Any]</code>
```

justinvp: Reminder: Pretty sure for these nested types, we'll need to change these from snake_case to camelCase. @pgavlin to confirm that's still the case.

content/docs/reference/pkg/aws/s3/bucket.md
```diff
@@ -0,0 +1,10931 @@
...
+
+<div class="highlight"><pre class="chroma"><code class="language-python" data-lang="python"><span class="k">def </span><span class="nf">Bucket</span><span class="p">(resource_name, opts=None, </span>acceleration_status=None<span class="p">, </span>acl=None<span class="p">, </span>arn=None<span class="p">, </span>bucket=None<span class="p">, </span>bucket_prefix=None<span class="p">, </span>cors_rules=None<span class="p">, </span>force_destroy=None<span class="p">, </span>grants=None<span class="p">, </span>hosted_zone_id=None<span class="p">, </span>lifecycle_rules=None<span class="p">, </span>loggings=None<span class="p">, </span>object_lock_configuration=None<span class="p">, </span>policy=None<span class="p">, </span>region=None<span class="p">, </span>replication_configuration=None<span class="p">, </span>request_payer=None<span class="p">, </span>server_side_encryption_configuration=None<span class="p">, </span>tags=None<span class="p">, </span>versioning=None<span class="p">, </span>website=None<span class="p">, </span>website_domain=None<span class="p">, </span>website_endpoint=None<span class="p">, __props__=None);</span></code></pre></div>
```
justinvp: Can we make `Bucket` link to https://www.pulumi.com/docs/reference/pkg/python/pulumi_aws/s3/#pulumi_aws.s3.Bucket ?

content/docs/reference/pkg/aws/s3/bucket.md
```diff
@@ -0,0 +1,10931 @@
...
+
+<div class="highlight"><pre class="chroma"><code class="language-go" data-lang="go"><span class="k">func </span>NewBucket<span class="p">(</span><span class="nx">ctx</span> *<span class="nx"><a href="https://pkg.go.dev/github.com/pulumi/pulumi/sdk/go/pulumi?tab=doc#Context">pulumi.Context</a></span><span class="p">, </span><span class="nx">name</span> <span class="nx"><a href="https://golang.org/pkg/builtin/#string">string</a></span><span class="p">, </span><span class="nx">args</span> *<span class="nx"><a href="https://pkg.go.dev/github.com/pulumi/pulumi-aws/sdk/go/aws/s3?tab=doc#BucketArgs">BucketArgs</a></span><span class="p">, </span><span class="nx">opts</span> ...<span class="nx"><a href="https://pkg.go.dev/github.com/pulumi/pulumi/sdk/go/pulumi?tab=doc#ResourceOption">pulumi.ResourceOption</a></span><span class="p">) (*<span class="nx"><a href="https://pkg.go.dev/github.com/pulumi/pulumi-aws/sdk/go/aws/s3?tab=doc#Bucket">Bucket</a></span>, error)</span></code></pre></div>
```
justinvp: Similarly, can we make `NewBucket` link to https://godoc.org/github.com/pulumi/pulumi-aws/sdk/go/aws/s3#NewBucket ?

content/docs/reference/pkg/aws/s3/bucket.md
```diff
@@ -0,0 +1,10931 @@
...
+
+            
+            </td>
+        </tr>
+    
+        <tr>
+            <td class="align-top">website</td>
+            <td class="align-top">
+                
+                <code><a href="#bucketwebsite">s3.<wbr>Bucket<wbr>Website?</a></code>
```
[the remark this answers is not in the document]
cnunciato: I think we can do without this, yeah, especially since it's on this page, and the whole page is "in" the `s3` namespace as well, so it seems implicit and okay to omit.

tools/resourcedocsgen/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Generating docs
+
+Run `make resource_docs`.
+
+### Using the latest templates in `pulumi/pulumi`
+
+This tool indirectly depends on the `pulumi/pulumi` repo, namely the `pkg/codegen/docs` generator and the templates it uses, which is the underlying tool responsible for generating docs from the Pulumi schema.
+
+To make changes to the templates or to use the latest ones from `master`, you will need to override the module path for `pulumi/pulumi` in your local copy of `pulumi-terraform-bridge` using:
```
cnunciato: Is this right? It looks like `gen_resource_docs.sh` actually looks for `../pulumi`, and when I ran it, I got the changes I'd made locally.

praneetloke: Ah yes. Good catch. I need to update this instruction. The script regenerates the templates bundle each time. So local modifications can be picked-up easily. I'll update the README. Thank you, @cnunciato, for spotting it!

joeduffy:
I am so excited to see this landing!! 🎉🎉🎉 

Really great work, just a few comments from me ([many discussed previously here FWIW](https://github.com/pulumi/docs/pull/2551))

* I was surprised [the language chooser remains on the API docs landing page](http://content-bucket-de682d9.s3-website-us-east-1.amazonaws.com/docs/reference/pkg/) and that there are still differences in the package lists. If we need it for the subset of packages that are still language-specific (like AWSX), it would be good to put those in a separate section on the page.

* In fact, all links on that page seem to go to the old docs still? Personally, I think the new docs are so much better already, we should just change those links and get people seeing them and providing feedback, even as we add the finishing touches.

* I do question whether we should be using the same breadcrumbs for these as the API docs. Both say Docs > Reference > API > ..., which makes it slightly confusing that we now have two different kinds of AWS packages there. Maybe we should have Docs > Reference > API > (Language) > (Package), and Docs > Reference > Providers > (Provider) as two parallel breadcrumbs.

* If I have a non-TypeScript language selected, the examples have titles but are empty. Can we add a "Coming Soon" with a link to the GitHub work item, like we do in other areas where we have a language gap?

* We seem to be hiding the JavaScript language chooser element. I understand why we did it, but it's confusing to me, because it's inconsistent with what we do everywhere else in the site. Should we just show it even though it's technically the same as TypeScript today, for consistency? Or are we planning on removing the JavaScript tab elsewhere in the site? (My vote is let's not do that FWIW.)

* Minor thing, it would be good to have @zchase look at the titles we're creating programmatically. I suspect something like "AWS S3 Bucket Resource Guide | Pulumi" would be better than the current title of "Bucket | Pulumi" for e.g. [this page](http://content-bucket-de682d9.s3-website-us-east-1.amazonaws.com/docs/reference/pkg/aws/s3/bucket/) (in terms of SEO).

* Do we plan to make the "Example Usage" section collapsible? It consumes a lot of vertical space. See https://github.com/pulumi/docs/pull/2551#discussion_r385440905 for some prior discussion.

* Nit, the alignment of the args table isn't aligned with the language chooser tabs just before it.

* Do we plan on making the font for that table smaller? A lot of things wrap and it makes it a bit difficult to read for my eyes (especially as our left/right navs consume so much space).

* Can we sort required argument properties to the top of the table?

* Can we use smaller indicators than "(Required) " and "(Optional) " for these designations? Maybe one of the chiclets (not orange, as that'd be jarring, but maybe light grey). In fact, if we sort them as suggested above, then I'd simply leave off "(Required) " and make it implied.

* Can we omit the output properties that _also_ appear as input properties? The table would be 10x more valuable because then I'd be able to get a sense of what properties are actually computed for me, which I suspect is what people want 99% of the time when looking at this table. It doesn't seem helpful to just mirror the input properties which are always available as an output also.

Nothing here blocking, I'd land as-is and just have follow ups for these, honestly. Again great work!

* I noticed some markdown isn't getting rendered; e.g., for the "Look up Existing Resource" table, the id property seems to say \_unique\_ when I think it should render as _unique_.

* Can we link from the "Look up Existing Resource" section to [the programming model docs](https://www.pulumi.com/docs/intro/concepts/programming-model/#resource-get)? One thing that's not clear is what ID to use for a given resource as it varies by resource type (e.g., sometimes the ARN, sometimes the raw name, in Azure the fully qualified subscription name, and so on) -- is that available in the docs somewhere?

* Minor suggestion for the long-term (not now, but if you have a laundry list would be good to track), would be nice to have a little description of the module when I hover over it (e.g., hovering over `s3` would say "Amazon S3 provides object storage to build and retrieve data from anywhere.")

cnunciato:
This is a great list, thanks @joeduffy! This PR doesn't contain any UX treatment from me yet, and most of the items you mention are definitely on my list. Once this merges, you'll see it start to land.

praneetloke:
As Christian pointed out, this PR is only for adding docs generation tool to this repo to produce docs. There are separate issues tracking the UX work. Please see #2586 and #2587.

Also, FYI to see all of the work that is remaining for the resource docs, please see this GH project: https://github.com/orgs/pulumi/projects/10.

content/docs/reference/pkg/aws/s3/bucket.md
praneetloke: I created https://github.com/pulumi/pulumi/issues/4095 to track this. I will confirm with Pat offline and fix it in a separate PR in the pulumi/pulumi repo.

content/docs/reference/pkg/aws/inspector/getrulespackages.md
praneetloke: I fixed this in the latest commit in https://github.com/pulumi/pulumi/pull/4085.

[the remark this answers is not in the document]
praneetloke: Fixed this in the latest commit in https://github.com/pulumi/pulumi/pull/4085.

content/docs/reference/pkg/aws/s3/bucket.md
praneetloke: Fixed this in the latest commit in https://github.com/pulumi/pulumi/pull/4085.

praneetloke: Once the new preview is available, could you please take a look at it?

content/docs/reference/pkg/aws/s3/bucket.md
lukehoban: Agreed - we should use `PascalCase` names for Python types.

lukehoban:
I see `(Optional) The type of volume. Can be &#34;standard&#34;, &#34;gp2&#34;, or &#34;io1&#34;. (Default: &#34;gp2&#34;)`. on http://content-bucket-de682d9.s3-website-us-east-1.amazonaws.com/docs/reference/pkg/aws/ec2/instance/, but that awkward escaping of the quotes doesn't show up on (e.g.) https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/aws/ec2/#block-devices.  Is this a more general issue with the way we are embedding the content here?

praneetloke:
>  Is this a more general issue with the way we are embedding the content here?

Yes, I believe so. I think this is an issue with rendering something that has double-quotes in it. I wrote-up a [note](https://github.com/orgs/pulumi/projects/10#card-34741390) in the docs project to investigate/fix separately.

PR #2643, opened 2020-03-12
Add logging example for Go

tasiah:
I'm not sure if or where we should describe what the other 3 fields are, since all of the other languages besides Go support optional parameters.

### DIFF
--- a/content/docs/intro/concepts/programming-model.md
+++ b/content/docs/intro/concepts/programming-model.md
@@ -3068,11 +3068,18 @@ log.error("fatal error")
 ```
 
 ```go
-ctx.Log.Info("message", nil, 0, false)
-ctx.Log.Info("message", resource, 0, false)
-ctx.Log.Debug("hidden by default", nil, 0, false)
-ctx.Log.Warn("warning", nil, 0, false)
-ctx.Log.Error("fatal error", nil, 0, false)
+// Optional arguments for logging.
+args := &pulumi.LogArgs{
+    Resource: resource,
+    StreamID: 0,
+    Ephemeral: false,
+}
+
+ctx.Log.Info("message", nil)
+ctx.Log.Info("message", args)
+ctx.Log.Debug("hidden by default", nil)
+ctx.Log.Warn("warning", nil)
+ctx.Log.Error("fatal error", nil)
 ```
 
 ```csharp
--- a/content/docs/intro/concepts/programming-model.md
+++ b/content/docs/intro/concepts/programming-model.md
@@ -3068,9 +3068,11 @@ log.error("fatal error")
 ```
 
 ```go
-// The Pulumi logging functions are not yet available in Go.
-//
-// See https://github.com/pulumi/pulumi/issues/3801.
+ctx.Log.Info("message", nil, 0, false)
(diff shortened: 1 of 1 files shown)


content/docs/intro/concepts/programming-model.md
```diff
@@ -3068,9 +3068,11 @@ log.error("fatal error")
 ```
 
 ```go
-// The Pulumi logging functions are not yet available in Go.
-//
-// See https://github.com/pulumi/pulumi/issues/3801.
+ctx.Log.Info("message", nil, 0, false)
```
EvanBoyle: You could do something along these lines to highlight that these params are optional:

```go
// "optional" config for logging
var resource Resource
var streamID int32
var ephemeral bool

ctx.Log.Info("message", resource, streamID, ephemeral)
...
```

Alternatively, you could redesign the API that we expose to be in the form:

```go
type LogConfig struct {
 Resource Resource
 StreamID int32
 Ephemeral bool
}

func (log *logState) Info(msg string, logConfig *LogConfig) error {}```

Comment (lukehoban) [reply content/docs/intro/concepts/programming-model.md]: Note that we do need to keep consistency in this example across languages. If we want to change/improve it - we can do that uniformly across all 5 examples.

Comment (tasiah) [reply content/docs/intro/concepts/programming-model.md]: @lukehoban @pgavlin Thoughts about using a `LogConfig`? I kind of like it because it removes the onus on the user to "make up" streamIDs and/or resources, but I'm not sure if it's too inconsistent with the other languages.

Comment (lukehoban) [reply content/docs/intro/concepts/programming-model.md]: It's a good point.  I don't feel too strongly either way - but on balance I do think this is still close enough to the API design in other languages that it would be fine, and it does feel a little nicer from a Go API usage experience.

Comment (chrsmith) [reply content/docs/intro/concepts/programming-model.md]: +1 for making this a little less onerous to use, since for most cases the user won't know what those values even mean.

We could go with several approaches here, from having multiple variants like `Info(msg)` and `InfoWithContext(msg, logCtx)`. But perhaps the more idiomatic thing would just be to expose two different versions of the `log.Logger` [interface](https://golang.org/pkg/log/#Logger)?

Something like:

```go
// A default Logger is initialized and on the `Log` field, which associates
// data with the stack resource, and StreamID 0. Non-ephemeral.
ctx.Log.Infof("Creating sprocket using widget configuration %+v", widgetCfg)

// You can optionally create a logger with a provided log config, to customize
// how/where the log lines are written.
//
// The `LogWithContext` method returns an instance of the `log.Logger` interface,
// but exposing the Pulumi log functionality that will use the provided log configuration.
ctx.LogWithContext(LogConfig{ Resource: doodad.URN() }).Infof("Dodad created")
```

What do you all think?

lukehoban: I do want to make sure we stay close to the api design across other languages. For that reason I prefer the current checked in design or what @evanboyle and @tasiah suggest - as both are pretty 1:1 mappings of existing api design into Go. 

Also worth looking in practice at what @tasiah has done in the Docker PR, which shows common usage style.

PR #2652, opened 2020-03-13
Add case study for Credijusto

cnunciato:


cnunciato:
Merging on an OK from @isaac-pulumi.

PR #2667, opened 2020-03-17
Add Terraform PPC page 

zchase:
https://github.com/pulumi/home/issues/660


cnunciato:
@zchase what's needed to finish this off, assuming it's still relevant? Can I help get it finished up?

zchase:
@cnunciato A review was all this needed, but now that `stencil.js` is a thing we probably want to implement the JS in `stencil` instead of `jQuery`, right?

cnunciato:
Probably, but I'd also like to talk through this a bit more so I can help out with the overall UX.

cnunciato:
@zchase Can we close this PR and revisit collaboratively on a replacement?

zchase:
@cnunciato Yeah let's mark it as closed and come back. There are no near-term plans to use the page currently, though we should revisit this and `/terraform`.

PR #2689, opened 2020-03-20
Revert "1.13.0 failed to publish"

pgavlin:
Reverts pulumi/docs#2683


pgavlin:
1.13.0 for OSX is now available.

PR #2714, opened 2020-03-23
Run the broken-link checker using a local website build

### LINKED ISSUE #2709: Travis CRON jobs are failing often
Many Travis CRON jobs have recently started failing due to `hugo server` OOMing:

https://travis-ci.com/github/pulumi/docs/builds/154411399

```
371make validate
372make[1]: Entering directory '/home/travis/build/pulumi/docs'
373hugo server \
374    --buildDrafts --disableBrowserError --disableLiveReload \
375    &>/dev/null &
376while ! nc -z localhost 1313; do sleep 0.1; done
377Building sites … fatal error: runtime: out of memory
```

Since we’re no longer running the broken link checker against PRs, should we consider making it so that the CRON job checks for broken links against the actual deployed www.pulumi.com itself, rather than using Hugo's built-in webserver?

Or are there any other knobs to `hugo server` we could add to help prevent it from OOMing?

Another option would be to explore upgrading Hugo to a more recent version, which might not OOM as much?

Or can we get more memory to use during Travis jobs?

@cnunciato, do you have any cycles to look into this?

cnunciato:
Fixes #2709.

### DIFF
--- a/scripts/check-links.sh
+++ b/scripts/check-links.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -o nounset -o errexit -o pipefail
 
+HTTP_SERVER_PORT=8888
+
 check_links() {
     # We exclude some links:
 	#     - Our generated API docs have lots of broken links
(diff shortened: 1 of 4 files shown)


scripts/check-links.sh
```diff
@@ -49,4 +51,14 @@ retry() {
...
+yarn run http-server public --port $HTTP_SERVER_PORT &>/dev/null &
+HTTP_SERVER_PID=$!
+
+# Run the link checker once the HTTP server is listening.
+while ! nc -z localhost $HTTP_SERVER_PORT; do sleep 0.1; done
+# Re-run the link checker up to three times, to handle the occasional transient failure.
 retry check_links
+
+# Kill the server process.
+kill -f $HTTP_SERVER_PID
```
justinvp: Is the `-f` flag needed for `kill`?

cnunciato: Ah, no, my oversight -- it isn't. I'll remove it.

PR #2724, opened 2020-03-23
Update language around Node.js version support

### LINKED ISSUE #2722: Update supported Node versions
I believe we should decouple this from `2.0` and just adjust our [messaging](https://www.pulumi.com/docs/intro/languages/javascript/) from "Node 8 and later" to "Maintenance, Current and Active LTS versions of Node" with a link to https://nodejs.org/en/about/releases/.  We should drop Node 8 from our CI testing across the board, and add Node 13 now, and Node 14 in April when it is released.

_Originally posted by @lukehoban in https://github.com/pulumi/pulumi/issues/3736#issuecomment-602759805_

lukehoban:
Fixes #2722

### DIFF
--- a/content/docs/intro/languages/javascript.md
+++ b/content/docs/intro/languages/javascript.md
@@ -12,7 +12,7 @@ aliases: ["/docs/reference/javascript/"]
 
 <img src="/logos/tech/logo-nodejs.png" align="right" width="150" style="padding:8px; margin-top: -64px">
 
-Pulumi supports JavaScript programs running on Node.js version 8 and later.
+Pulumi supports JavaScript programs running on Node.js using any of the [Current, Active and Maintenance LTS versions](https://nodejs.org/en/about/releases/).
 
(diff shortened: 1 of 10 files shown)


justinvp:
I did a regex search for `node.*8` and found some other places that could be updated.

content/docs/intro/languages/javascript.md
```diff
@@ -12,7 +12,7 @@ aliases: ["/docs/reference/javascript/"]
 
 <img src="/logos/tech/logo-nodejs.png" align="right" width="150" style="padding:8px; margin-top: -64px">
 
-Pulumi supports JavaScript programs running on Node.js version 8 and later.
+Pulumi supports JavaScript programs running on Node.js using any of the [Current, Active and Maintenance LTS versions](https://nodejs.org/en/about/releases/).
```
stack72: Love this! Removing the hardcoding FTW!!

lukehoban:
> LGTM, but do a regex search for node.*8 -- there are some other places that could be updated as well.

Thanks - yes - fixed up a few more.

PR #2729, opened 2020-03-23
Add examples shortcodes

cnunciato:
These don’t do anything yet — they just render whatever’s passed into them.

### DIFF
--- a/components/src/components.d.ts
+++ b/components/src/components.d.ts
@@ -34,6 +34,8 @@ export namespace Components {
     'selection': ChooserKey;
     'type': ChooserType;
   }
+  interface PulumiExample {}
+  interface PulumiExamples {}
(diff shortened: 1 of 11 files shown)


components/src/components.d.ts
```diff
@@ -60,6 +74,8 @@ declare global {
   interface HTMLElementTagNameMap {
     'pulumi-choosable': HTMLPulumiChoosableElement;
     'pulumi-chooser': HTMLPulumiChooserElement;
+    'pulumi-example': HTMLPulumiExampleElement;
```
praneetloke: Just so I understand, the prefix `pulumi-` here is just for the component and not for the actual shortcode, correct?

cnunciato: Yep, that's right.

PR #2734, opened 2020-03-24
Close unclosed shortcodes

cnunciato:
This fixes a JavaScript error [on the Kubernetes Control Plane page](https://www.pulumi.com/docs/guides/crosswalk/kubernetes/control-plane/) (which was also causing some content to be hidden).


cnunciato:
Merging this proactively as it fixes a bug -- happy to address any subsequent feedback, though.

PR #2736, opened 2020-03-24
Close an unclosed shortcode

cnunciato:
Fixes a rendering issue on a tutorial page.


cnunciato:
Proactively merging this one as well, again just because it is safe and address a live-site problem.

PR #2738, opened 2020-03-24
Testing guide

### LINKED ISSUE #992: User Guide: Testing Infrastructure
There are multiple forms of testing you might want to do with Pulumi:

- I want to unit test code in my Pulumi app
- I want to test that my Pulumi app created resources as expected
- I want to test that Pulumi itself provisioned the resources correctly
- I want to test the runtime behavior of apps provisioned by my Pulumi app

We have answers for all of these, but we don't have a centralize, one stop shopping doc, about them.

Let's write one.

### LINKED ISSUE #2541: Guide on using PaC for in-deployment testing

### LINKED ISSUE #2542: Guide on using Pitfall/ProgramTest for integration testing

mikhailshilkov:
Add a testing guide to the docs, with sub-guides per testing type.

Resolves #992 
Resolves #2542
Resolves #2541


content/docs/guides/testing/_index.md
```diff
@@ -0,0 +1,82 @@
...
+
+Pulumi uses general-purpose programming languages to define cloud infrastructure. This capability opens up all the doors to **automated testing** of the infrastructure resources.
+
+The software industry has come up with multiple styles of automated testing. Unit testing of individual pieces, integration testing of system components, end-to-end testing of the complete application&mdash;each style has its advantages and challenges.
+
+Pulumi provides a similar choice of testing options for cloud programs. Currently, we identify three types of tests:
+
+- **Unit Tests** are fast in-memory tests that mock all external calls.
+- **Tests based on Policy as Code** run resource-level assertions *while* infrastructure is being deployed.
+- **Integration Tests** deploy ephemeral infrastructure and run external tests against it.
```
justinvp: Should these bullets be in the same order as the columns in the table and headers below?

content/docs/guides/testing/_index.md
```diff
@@ -0,0 +1,82 @@
...
+
+The following table summarizes the differences between the three approaches:
+
+|                                | Unit Tests  | Policy as Code  | Integration Tests  |
+|--------------------------------|-----|------|------|
+| Provision real infrastructure  | No  | Yes  | Yes  |
+| Require the Pulumi CLI         | No  | Yes  | Yes  |
+| Time to execute                | Milliseconds  | Seconds to minutes  | Seconds to minutes  |
```
lukehoban: Perhaps simplify this to "Milliseconds", "Seconds", "Minutes"?  

I guess in part it depends what you think "Policy-as-code" based tests "cost" in terms of time - they are sort of "free" in that they run in amongst your deployment, but you can't get results for them without actually doing a deployment.

Nevertheless, in normal usage, these tests run quickly on the back of incremental changes to infrastructure, whereas integration tests stand up full copies of infrastructure from scratch and tear it down - so being clear these are more heavy-weight feels appropriate.

content/docs/guides/testing/_index.md
```diff
@@ -0,0 +1,82 @@
...
+| Language                       | Same as Pulumi program  | Node.js or Python  | Any language  |
```
lukehoban: We don't yet have bespoke libraries for "any language" for Integration testing, but depending on how you frame it in the details, this seems fine as is (that is, if you position as a workflow of doing `pulumi up`, "run validation" and `pulumi destroy` that can be scripted in any language).

content/docs/guides/testing/_index.md
```diff
@@ -0,0 +1,82 @@
...
+| Validation target              | Resource inputs  | Resource inputs and outputs | External endpoints |
+
+We encourage you to try all three styles of testing and use the ones that are most suitable for your quality targets, development practices, and application style.
```
lukehoban: Perhaps also something like - "for many teams, a combination of these approaches may make sense - unit tests that can run quickly and validate program logic, functional tests to validate key correctness invariants, and integration tests to fully test end-to-end correctness of infrastructure components" or something like that?

content/docs/guides/testing/_index.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Unit Testing
+
+If you ever created unit tests for your application code, you can directly transfer your skills to testing Pulumi programs. Unit tests evaluate the behavior of your code in isolation, while all external dependencies are replaced by **test doubles** (mocks).
```
lukehoban: Any reason to introduce the term "test doubles" here - vs. just using "mocks" directly?

content/docs/guides/testing/_index.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Integration Testing
+
+Integration testing takes the opposite approach: the tests deploy real cloud resources and validate their actual **behavior**.
+
+The flow goes like this:
+
+1. A test invokes the Pulumi command-line interface (CLI) to deploy infrastructure to an ephemeral environment.
+2. When deployment succeeds, the test queries stack outputs to retrieve endpoints of the infrastructure: usually, a URL or a public IP address.
+3. The test verifies that the infrastructure behaves as expected: for example, it expects a valid HTML document from a health-check endpoint.
```
lukehoban: ", or runs a suite of application-level tests against the endpoint"?

content/docs/guides/testing/_index.md
```diff
@@ -0,0 +1,82 @@
...
+
+[Learn more and get started with Integration Testing]({{< relref "./integration" >}}).
+
+## Testing with Policy as Code
+
+[Policy as Code]({{< relref "../crossguard" >}}) (also known as "CrossGuard") is Pulumi's offering to set guardrails and enforce compliance for cloud resources. In addition to authoring company-wide policies, CrossGuard enables another type of infrastructure testing.
+
+Policy-based tests run inside the Pulumi CLI before and after infrastructure provisioning. In contrast to "black-box" integration testing, policy rules have access to all input and output values of all cloud resources in the stack. As opposed to unit testing, policy rules can evaluate real values returned from the cloud provider instead of test doubles.
+
+Policy-based tests run against an ephemeral cloud environment, so the considerations of speed and cost still apply.
```
lukehoban: This isn't quite true - these tests typically would run against a "real" long-lived cloud environment, like a staging or production environment.  The key differences from "integration tests" is that these tests *don't* stand up an ephemeral environment to test, and instead run "inside" an existing deployment.

lukehoban: In fact, this point may be worth pulling up into the chart:

"Test environment": "In-memory", "long-lived stack (e.g. production)", "ephemeral stack"

?

content/docs/guides/testing/_index.md
mikhailshilkov: This is a more broad term, but I'm okay to simplify to mocks.

content/docs/guides/testing/_index.md
mikhailshilkov: Ah, this means we have different views on the suggested workflow. Your statement is true for "proper" policy as code but I thought that here we repurpose the same functionality to run them on ephemeral environments similar to integration tests.

In your view, if I'm a developer and want to have a suit of policy tests -- to make sure the quality of my code overall -- how and when will I run them? Would I have my personal long-lived development environment? How do I run them in CI/CD, for pull requests etc?

@justinvp @pgavlin @joeduffy What are your views on the workflow for PaC testing?

lukehoban: > In your view, if I'm a developer and want to have a suit of policy tests -- to make sure the quality of my code overall -- how and when will I run them? Would I have my personal long-lived development environment?

Yes, or a staging environment you open PRs against which fail on previews if you break a test (and fail deployments if you merge).

> How do I run them in CI/CD, for pull requests etc?

You either enable global policies for a set of environments, or provide a local pack of project-specific tests (policies) along with the project and set up `--policy-pack` parameters to the `pulumi up` you run inside your CI and PRs.

>  I thought that here we repurpose the same functionality to run them on ephemeral environments similar to integration tests.

I suppose you could run these policies inside deployment of ephemeral environments too, but we already have good support for validation of ephemeral environments with integration tests.  The thing that policies uniquely provide is tests that run intertwined with previews and updates against actual existing infrastructure.

joeduffy: Isn't it important to be able to validate policies against ephemeral environments to make sure that, as you roll out to your *real* environments, you don't encounter unexpected violations?

/cc @clstokes

mikhailshilkov: @lukehoban What prevents me from using the same workflow for integration tests? I could have the same staging environment and run my integration tests against it?

> fail on previews

Previews aren't always enough: if I make a non-trivial change, I may not have enough outputs on a preview to validate. E.g. I can't check vnets of a cluster that isn't deployed because I just introduced it in my PR.

I would say that a unique thing about policies is access to all the resources and their interconnections, so it's a different level of validation.

content/docs/guides/testing/_index.md
mikhailshilkov: Makes sense 👍

content/docs/guides/testing/_index.md
mikhailshilkov: Yes, I'll describe this in the details

content/docs/guides/testing/_index.md
```diff
@@ -0,0 +1,82 @@
...
+---
+
+Pulumi uses general-purpose programming languages to define cloud infrastructure. This capability opens up all the doors to **automated testing** of the infrastructure resources.
+
+The software industry has come up with multiple styles of automated testing. Unit testing of individual pieces, integration testing of system components, end-to-end testing of the complete application&mdash;each style has its advantages and challenges.
+
+Pulumi provides a similar choice of testing options for cloud programs. Currently, we identify three types of tests:
+
+- **Unit Tests** are fast in-memory tests that mock all external calls.
+- **Tests based on Policy as Code** run resource-level assertions *while* infrastructure is being deployed.
```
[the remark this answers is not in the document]
mikhailshilkov: Agreed about a better name, let's find one.

> functional tests

Wikipedia says it's "a type of black-box testing that bases its test cases on the specifications of the software component under test." Doesn't quite fit, IMO...

content/docs/guides/testing/_index.md
clstokes: > Isn't it important to be able to validate policies against ephemeral environments to make sure that, as you roll out to your _real_ environments, you don't encounter unexpected violations?
> 
> /cc @clstokes

Yes, I think we would recommend validating policies against _all_ environments - ephemeral, long-lived testing environments, and production environments.

mikhailshilkov: @clstokes

> validating policies

and also to @ericrudder 's email remark

> compliance (or policy) verification?

This guide "repurposes" policies for developers (or DevOps) to run sort of more powerful unit tests. While they use exactly the same tech, their goals and workflows are different. They don't write company-wide restrictions, but they verify that a specific program works as expected in some policy-like aspects.

Does that make sense? Would this be clear to people? Does this change anything in the terminology and workflow?

ericrudder:
i still like my original suggestion (close to cam) ... however, my last
attempt is "functional verification" ... i don't think anything is
"confusing" per se, i only replied since you asked!

i actually think your previous reply (this guide ... ) is the best
summation!  (and so you could make it prominent)

content/docs/guides/testing/_index.md
mikhailshilkov: Re-ordered

content/docs/guides/testing/_index.md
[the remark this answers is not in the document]
mikhailshilkov: I went with "property tests". LMKWYT.

content/docs/guides/testing/_index.md
mikhailshilkov: Changed to "seconds"

content/docs/guides/testing/_index.md
mikhailshilkov: I mentioned both usage scenarios now

content/docs/guides/testing/unit.md
```diff
@@ -0,0 +1,964 @@
...
+```
+
+{{% /choosable %}}
+
+All the tests passed!
+
+## Full Example
+
+{{% choosable language "typescript" %}}
+The full code for this guide is available in the examples repository: [Unit Tests in TypeScript](https://github.com/pulumi/examples/tree/master/testing-unit-ts).
```
leezen, line 954: For whatever reason, these links don't look styled correctly (they don't have a different color), so it's not clear that they're links. @cnunciato is this a bug in the chooser?

content/docs/guides/testing/unit.md
```diff
@@ -0,0 +1,964 @@
...
+dotnet add package Moq
+dotnet add package FluentAssertions
+```
+
+{{% /choosable %}}
+
+## Add Mocks
+
+Let's add the following code to mock the external calls to the Pulumi CLI.
+
```
leezen, line 234: I think it would be good to add a pointer here to https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/runtime/#Mocks so that end users know where these come from and what the interface looks like.

content/docs/guides/testing/_index.md
```diff
@@ -0,0 +1,82 @@
...
+[Learn more and get started with Property Testing]({{< relref "./property-testing" >}}).
+
+## Integration Testing
+
+Integration testing takes the opposite approach: the tests deploy real cloud resources and validate their actual **behavior**.
+
+The flow goes like this:
+
+1. A test invokes the Pulumi command-line interface (CLI) to deploy infrastructure to an ephemeral environment.
+2. When deployment succeeds, the test queries stack outputs to retrieve endpoints of the infrastructure: usually, a URL or a public IP address.
```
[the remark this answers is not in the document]
mikhailshilkov: Changed to `the test retrieves endpoints of the infrastructure from the stack outputs`

content/docs/guides/testing/property-testing.md
```diff
@@ -0,0 +1,233 @@
...
+- Running Kubernetes version `1.13`.
+- Provisioned inside a private VPC, rather than the default one.
+
+## Blank Project
+
+Our setup consists of two directories. The parent folder contains a typical Pulumi program written in TypeScript and bootstrapped with `pulumi new aws-typescript` command.
+
+The sub-directory `tests` contains a policy pack with our future property tests and is created with `pulumi policy new aws-typescript` (notice the `policy` argument).
+
+You can see the full layout in [the examples repository](https://github.com/pulumi/examples/tree/31056c3480cc445e5d4d3a8a0a86977adce2bc5e/testing-pac-ts).
```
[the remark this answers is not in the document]
mikhailshilkov: Makes sense! https://github.com/pulumi/examples/issues/622

content/docs/guides/testing/integration.md
```diff
@@ -0,0 +1,156 @@
...
+            if res.Type == "aws:s3/bucket:Bucket" {
+                foundBuckets++
+            }
+        }
+        assert.Equal(t, 1, foundBuckets, "Expected to find a single AWS S3 Bucket")
+    },
+})
+```
+
+Now when we run `go test`, it will not only run through the battery of lifecycle tests but will also, after successfully standing up the stack, perform the extra validation against the resulting state.
```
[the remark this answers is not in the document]
mikhailshilkov: Great point... I don't think I have a crystal-clear answer here. Do you have a take on that?

content/docs/guides/testing/unit.md
mikhailshilkov: Added `The definition of the mocks interface is available at the runtime API reference page.` down below

content/docs/guides/testing/unit.md
mikhailshilkov: Yeah, that's weird. I could figure out the reason and "fixed" it by adding some random `&nbsp;` at the end.

mikhailshilkov:
@leezen @spara Thanks for the great suggestions, I applied them to the guides. Please have another look.

PR #2742, opened 2020-03-24
Update resource docs script for generating Azure, Azure AD, and GCP resource docs

praneetloke:
### Proposed changes

As the title reads, this PR adds the resource docs generated using the Pulumi schema for Azure, Azure AD, and GCP cloud providers.

This [commit](https://github.com/pulumi/docs/pull/2742/commits/eecf7166ed4b2ad4cfc87acf892dca5d227a9990) contains changes to the `update_repos.sh` script as well as to the `gen_resource_docs.sh` script. The latter was changed to call `update_repos` to check local git repos to the latest released version; similar to what we do when we generate the tsdoc and pydoc API docs. The change to the `update_repos.sh` itself is to support accepting an optional argument so the caller can selectively update a repo rather than update all provider repos.

N/A

### Related issues (optional)

Related to #2555.


praneetloke:
> Any particular reason this is a draft PR?

Oh it was just so I could evaluate the preview before making it ready for review.

praneetloke:
Restarted the build one more time, but I have a feeling it'll fail again due to OOM. :(

cc @cnunciato

cnunciato:
Yeah, even when I build this locally, it quickly reaches just shy of 7GB. I think we're currently [capped at 7.5](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system) on Travis, too. Commenting the PostCSS processing doesn't seem to make a difference. (I thought it might.) I wonder if we can bump that memory restriction somehow.

praneetloke:
I rebased this PR with the latest `master` and also updated it to no longer have the Azure, GCP etc. resource docs, since the PR #2746 is doing that. Instead, this PR now simply introduces the changes to the scripts necessary to generate all the docs. I will leave this in a draft state for Christian to be able to copy the script changes to that PR and eventually will close this out.

praneetloke:
#2746 is now merged. Closing this out.

PR #2746, opened 2020-03-25
Resource docs and UX for AWS, Azure and GCP

### LINKED ISSUE #2557: Prototype look and feel of the new docs page layout
There are some pieces as part of the docs prototype that we couldn't complete, like the language chooser, table width, etc. This issue is to track the work to finish the design work on that to be applied to the new pages when they are ready.

### LINKED ISSUE #2586: Update the current API reference index page with resource docs information

cnunciato:
This change adds resource documentation for AWS, Azure, AzureAD and GCP, including links in the TOC and tiles to the API Reference home page. 

Also makes some slight adjustments to both the left and right columns for consistency and better use of space.

Fixes #2557.
Fixes #2586.
Part of #2587.


assets/config/tailwind.config.js
```diff
@@ -83,6 +83,10 @@ module.exports = {
                     "Open Sans",
                     ...defaultTheme.fontFamily.sans,
                 ],
+                mono: [
+                    "Source Code Pro",
+                    ...defaultTheme.fontFamily.mono,
+                ]
```
justinvp: Just curious: Why the change to use Source Code Pro?

assets/sass/components/_choosable.scss
```diff
@@ -12,6 +12,10 @@ pulumi-choosable {
     display: block;
 
     pre.chroma {
-        @apply mt-0;
+        @apply mt-0 bg-gray-100;
+    }
+
+    .highlight, .chroma {
+        // @apply border-transparent;
```
justinvp: Delete?

cnunciato: Oh yes, thanks -- missed it.

assets/config/tailwind.config.js
cnunciato: Ah, I should've called this out in the PR. Mainly I just wanted to try a code font that was a tiny bit smaller at the same relative size (since the text in these new docs is a bit smaller, too), and our current font's always felt slightly to me. Source Code Pro is just one that I've used before and happen to like. Totally open to alternatives, though.

PR #2765, opened 2020-03-26
Add pac config docs & remove PULUMI_EXPERIMENTAL

ekrengel:
Two commits:

1. Remove PULUMI_EXPERIMENTAL from commands since its no longer required & remove chooser since we dont need it anymore!
1. Add documentation on configuration

I still need to update the AWSGuard docs about the new config (unless @jkisk was planning on doing that?) and I need to add docs on some of the newer commands / optional flags. I plan to do that in a different PR.


content/docs/guides/crossguard/configuration.md
```diff
@@ -0,0 +1,132 @@
...
+    parent: crossguard
+---
+<!-- markdownlint-disable ul code -->
+{{% crossguard-preview %}}
+
+## Overview
+
+Configuration allows you to author flexible Policy Packs that can be reused across your entire organization. As an organization administrator, you can use a single Policy Pack yet vary configuration (e.g. enforcement level, the allowed instance types, cost allowances, etc.) from one Policy Group to the next. For example, you may have a Policy Group for your non-production stacks that allow smaller instance types, while your Production Policy Group allows for use of large instance types.
+
+Configuration schema defined using per policy, and then the actual configuration can be set via a form in the Pulumi Console or using a JSON file.
```
sean1588: Configuration schema "is defined per policy, and then the actual....."  ?

ekrengel: ... what is the question?

ekrengel: oh i see.. will fix!

content/docs/guides/crossguard/configuration.md
```diff
@@ -0,0 +1,132 @@
...
+                type: "string",
+                minLength: 5,
+                maxLength: 50,
+                default: "This is a configurable message."
+            }
+        },
+        required: ["message"],
+    },
+    validateResource: validateTypedResource(aws.s3.Bucket, (_, args, reportViolation) => {
+        const config = args.getConfig<{ message?: string }>();
```
justinvp: Since you've specified `required: ["message"]`, this can be typed as:

```suggestion
        const config = args.getConfig<{ message: string }>();
```

...without the `?` after `message`. Message will always be set since it was specified as being required.


However, I wonder if we should lead with an example that doesn't use a required property, and have some verbiage about how properties are optional by default. And show the use of the `?` in the type specified with the `getConfig<T>` call, and then inside the validation handler, show how to check if it is set, but otherwise still have the policy be usable without the config flag being set. Because we generally recommend not using required config properties -- that just adds friction in enabling/adopting the policy. It's generally better to provide policies that can run without any config, but can change behavior or opt-in to more specific checking if certain config properties _are_ set.

For example, a policy that checks encryption is enabled (that can do that check without any config), that has an optional config value for an "id of the encryption key" that, when specified, _further_ enforces that that specific key is being used for the encryption.

And then have a separate paragraph that mentions, how in some cases a policy by it's very nature necessitates a required property, in which case, here's how you'd do it... How the `?` in the `getConfig<T>` type isn't necessary because the system will enforce it being set, etc.

content/docs/guides/crossguard/configuration.md
```diff
@@ -0,0 +1,132 @@
...
+```json
+{
+    "all": {
+        "enforcementLevel": "disabled",
+    },
+    "a-policy": {
+         "enforcementLevel": "mandatory",
+    }
+}
+```
```
justinvp: JSON can be more strict about trailing commas. I don't know off the top of my head if Go's JSON unmarshalling is lenient in this regard, but I'd personally leave them off.

````suggestion
```json
{
    "all": {
        "enforcementLevel": "disabled"
    },
    "a-policy": {
         "enforcementLevel": "mandatory"
    }
}
```
````

Comment (justinvp) [inline content/docs/guides/crossguard/configuration.md]: Also, I did make it possible to specify the the enforcement level value directly when that's the only thing being configured for a policy (similar to how [ESLint allows specifying just a string for a rule, or an array of values](https://eslint.org/docs/user-guide/configuring#using-configuration-files)). Do we want to document this?

After that example, perhaps something like:

````markdown
As a convenience, when only configuring an enforcement level for a policy, its value can be specified directly. The above example could therefore also be written as:

```json
{
    "all": "disabled",
    "a-policy": "mandatory"
}
```
````

File: content/docs/guides/crossguard/configuration.md [hunk 4]
```diff
@@ -0,0 +1,132 @@
...
+
+At the top level, the `configSchema` for each Policy has only the `properties` and `required` fields. Within the `properties` field, you may use the entire breadth of [validations keywords provided by the JSON Schema](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6). Validation keywords are fields like `minLength` and `maxLength` in the above example.
+
+## Configuring a Policy Pack
+
+You can configure a Policy Pack via the `pulumi` CLI or the Pulumi Console.
+
+### Using the Pulumi CLI
```
justinvp: I find the flow of these headings confusing:

- Using the Pulumi CLI
  - Running Policy Packs Locally
  - Validating Configuration
  - Enabling the Policy Pack
- Using the Pulumi Console

It starts off talking about running policy packs locally, but then gets into validating configuration and enabling the policy pack, which are about using policy packs with the Pulumi Console (via CLI commands).

I wonder if it's make more sense to organize this around the two ways of using Policy Packs:

- Running Policy Packs Locally
- Using the Pulumi Console

And then under "Using the Pulumi Console", first show how the UI itself can be used, but then also have sections about how the CLI can be used to validate and enable policy packs published to the service.

content/docs/guides/crossguard/configuration.md
```diff
@@ -0,0 +1,132 @@
...
+
+If the Policy Pack has already been [published to the Pulumi Console]({{< relref "/docs/get-started/crossguard/enforcing-a-policy-pack" >}}), you can validate the configuration using the `pulumi policy validate-config` command.
+
+```bash
+$ pulumi policy validate-config <org-name>/<policy-pack-name> <version> --config <path-to-policy-pack-config-file>
+```
+
+#### Enabling the Policy Pack
+
+Once you are satisfied with the configuration, (if Policy as Code is enabled for your organization) you can enable the Policy Pack and configuration for your organization's defualt Policy Group by running:
```
justinvp:
```suggestion
Once you are satisfied with the configuration, (if Policy as Code is enabled for your organization) you can enable the Policy Pack and configuration for your organization's default Policy Group by running:
```

content/docs/guides/crossguard/configuration.md
ekrengel: - [ ] lead with non-required example
- [ ] discuss properties being optional by default
- [ ] generally recommend not having required PPs bc it adds friction

content/docs/guides/crossguard/configuration.md
```diff
@@ -0,0 +1,169 @@
...
+
+Configuration schema is defined per policy, and then the actual configuration can be set via a form in the Pulumi Console or using a JSON file.
+
+Configuration is currently supported for the Node.js Policy SDK (TypeScript/JavaScript). Python support for configuration is [coming soon](https://github.com/pulumi/pulumi-policy/issues/210).
+
+## Writing Configurable Policy Packs
+
+### Enforcement Level
+
+The Policy SDK will automatically make the enforcement levels of all policies configurable. You can configure the enforcement level for all policies in a Policy Pack or configure it for each individual policy level. The top level enforcement level under `all` will override the existing enforcement levels of all policies in the Policy Pack. A configured enforcement level for a specific policy will override the policy's existing enforcement level as well as the top level configured enforcement level.
```
justinvp:
```suggestion
The Policy SDK will automatically make the enforcement levels of all policies configurable. You can configure the enforcement level for all policies in a Policy Pack or configure it for each individual policy. The top-level enforcement level under `all` will override the existing enforcement levels of all policies in the Policy Pack. A configured enforcement level for a specific policy will override the policy's existing enforcement level as well as the top-level configured enforcement level.
```

content/docs/guides/crossguard/configuration.md
```diff
@@ -0,0 +1,169 @@
...
+        const config = args.getConfig<{ message?: string }>();
+        if (!config.message) {
+            config.message = "Setting reasonable defaults is recommended!";
+        }
+        reportViolation("Here is the configurable message: " + config.message);
+    }),
+}
+```
+
+The schema in the above example includes an optional property. If the property value is not set, a reasonable message is set from within the `validateResource` function. It's generally better to provide policies that has reasonable defaults and can run without setting any configuration values. Configuration should be used to change behavior or opt-in to more specific checks in the case that a specific configuration property is set. For example, a policy that checks encryption is enabled can be done without any configuration, but an optional configuration property for an "id of the encryption key" can be specified to further enforce that a specific key is being used for the encryption.
```
justinvp:
```suggestion
The schema in the above example includes an optional property. If the property value is not set, a reasonable message is set from within the `validateResource` function. It's generally better to provide policies that have reasonable defaults and can run without setting any configuration values. Configuration should be used to change behavior or opt-in to more specific checks in the case that a specific configuration property is set. For example, a policy that checks encryption is enabled can be done without any configuration, but an optional configuration property for an "id of the encryption key" can be specified to further enforce that a specific key is being used for the encryption.
```

content/docs/guides/crossguard/configuration.md
```diff
@@ -0,0 +1,169 @@
...
+
+```json
+{
+    "all": {
+        "enforcementLevel": "mandatory"
+    },
+    "example-policy-with-schema":{
+        "message": "Here is my message."
+    }
+}
```
justinvp:
```suggestion
{
    "all": "mandatory",
    "example-policy-with-schema": {
        "message": "Here is my message."
    }
}
```

content/docs/guides/crossguard/configuration.md
```diff
@@ -0,0 +1,169 @@
...
+            message: {
+                type: "string",
+                minLength: 5,
+                maxLength: 50,
+            }
+        },
+        required: ["message"],
+    },
+    validateResource: validateTypedResource(aws.s3.Bucket, (_, args, reportViolation) => {
+        // We no longer need to add the ? in getConfig function, since the message field is required.
```
justinvp:
```suggestion
        // The `?` after the `message` property in the type specified to the `getConfig` function is no
        // longer necessary since `message` is required and will always have a value.
```

PR #2768, opened 2020-03-27
Migrate to GitHub Actions

praneetloke:
### Proposed changes
Migrate from Travis CI to GitHub Actions.

The reason for this is, we have been hitting OOM errors on Travis with Hugo when we add new resource docs for additional providers. Hugo seems to be hitting some sort of Travis CI limitations. Though the failure is because of OOM, we have 7.5GB on our Travis runner, and with GitHub-hosted runners, we get 7GB. So my suspicion is that we are hitting some other sort of limit or just the way that Hugo works is exhausting the Travis VM in a way that is unexpected. Moving to GitHub Actions posed as an option as we are already using Pulumify and also the move was mostly 1:1 without much changes in our core build files.

I have simply erased all of the contents of `.travis.yml` and replaced it with a deprecation message, mostly because I am not sure what removing the Travis yml outright will do the other PRs. But I would like to handle disconnecting Travis for this repo separately, and can remove the file at that time. Let me know what you think.

N/A

### Related issues (optional)
This currently blocks PR #2742.


praneetloke:
Here are some links: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources

Travis CI: https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system. See the column under `Ubuntu` for max Memory.

Makefile
```diff
@@ -60,30 +60,22 @@ build:
 test:
 	./scripts/check-links.sh
 
-.PHONY: travis_push
-travis_push::
+.PHONY: ci_push
+ci_push::
 	$(MAKE) banner
 	$(MAKE) ensure
-ifeq ($(TRAVIS_BRANCH),master)
```
praneetloke: I removed these checks because the GitHub actions workflow is set to only run for events against the `master` branches.

praneetloke:
I will merge this first thing on Monday.

PR #2791, opened 2020-03-31
2.0 migration doc

leezen:


content/docs/get-started/install/migrating-2.0.md
```diff
@@ -0,0 +1,113 @@
...
+no_on_this_page: true
+---
+
+## Migrating to 2.0
+
+Upgrading to 2.0 is simple. First, you will download the 2.0 CLI, which is required to use the 2.0 SDK. Then, update each of your Pulumi programs to utilize the new SDK. If you're using Javascript or Typescript, you'll also need to ensure you're using invokes asynchronously. We provide detailed instructions on each of these steps below.
+
+## Install Pulumi 2.0 CLI
+
+If you are upgrading to a beta release, obtain the latest binaries from the [versions]({{< relref "versions" >}}) page. Then, follow the [manual installation]({{< relref "./#manual-installation" >}}) steps for your operating system.
```
lukehoban: Perhaps add a `> Note` here that once the final `2.0.0` release is available, users will also be able to install from all currently supported package managers (HomeBrew, Choclatey).

content/docs/get-started/install/migrating-2.0.md
```diff
@@ -0,0 +1,113 @@
+---
+title: Migrating to Pulumi 2.0
+meta_desc: This page provides instructions for migrating to Pulumi 2.0
+no_on_this_page: true
+---
+
+## Migrating to 2.0
+
+Upgrading to 2.0 is simple. First, you will download the 2.0 CLI, which is required to use the 2.0 SDK. Then, update each of your Pulumi programs to utilize the new SDK. If you're using Javascript or Typescript, you'll also need to ensure you're using invokes asynchronously. We provide detailed instructions on each of these steps below.
```
lukehoban:
```suggestion
Upgrading to 2.0 is simple. First, you will download the 2.0 CLI, which is required to use the 2.0 SDK. Then, update each of your Pulumi programs to utilize the new SDK. If you're using JavaScript or TypeScript, you'll also need to ensure you're using invokes asynchronously. We provide detailed instructions on each of these steps below.
```

content/docs/get-started/install/migrating-2.0.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Update Dependencies
+
+```bash
+npm install @pulumi/pulumi@^2.0.0
+```
+
+### Remove synchronous invokes
+
+Pulumi 2.0 no longer supports synchronous invokes. As a result, using the return values of any resource `get` operations will need to be made asynchronous. Similarly, the use of `getOutputSync` and `requireOutputSync` are no longer supported and you should move to using the `Output`-based versions of `getOutput` and `requireOutput`.
```
lukehoban:
```suggestion
Pulumi 2.0 no longer supports synchronous invokes. As a result, the return value of all `getSomething` operations are now `Promise<Something>` values, instead of `Something` values, and should be processed using `.then`, `async/await` or by passing the results to `pulumi.output()`. Similarly, the the `StackReference` operations `getOutputSync` and `requireOutputSync` are no longer supported and you should move to using the `Output`-based versions of `getOutput` and `requireOutput`.
```

content/docs/get-started/install/migrating-2.0.md
```diff
@@ -0,0 +1,113 @@
...
+
+```javascript
+const engineVersion = gcp.container.getEngineVersions().latestMasterVersion;
+```
+
+becomes
+
+```javascript
+const engineVersion = gcp.container.getEngineVersions().then(v => v.latestMasterVersion);
+```
```
lukehoban: Perhaps these three:

```typescript
//Before:
const version: string = gcp.container.getEngineVersions().latestMasterVersion;
//After w/ Promise:
const version: Promise<string> = gcp.container.getEngineVersions().then(v => v.latestMasterVersion);
//After w/ Output:
const version: Output<string> = pulumi.output(gcp.container.getEngineVersions()).latestMasterVersion;
```

The types maybe help make clear what has to change in surrounding code, and the two options highlight that often just wrapping in `pulumi.output` is the easiest way to accomplish very few other changes being needed (due to the lifting of properties onto `Output`s.

content/docs/get-started/install/migrating-2.0.md
```diff
@@ -0,0 +1,113 @@
...
+{{% choosable language python %}}
+
+### Update Dependencies
+
+Modify your `requirements.txt` file:
+
+```
+pulumi>=2.0.0
+```
+
```
lukehoban: Perhaps add a note about potentially needing to update other Pulumi libraries to newer versions (not quite just `2.0` as some will be `3.0`, and others will be <1.0 still)?  (Or just include AWS like is included in the Go example below?)

content/docs/get-started/install/migrating-2.0.md
leezen: Makes sense -- thanks for the suggestion. I also like inlining the before/after w/in the same code block.

PR #2796, opened 2020-03-31
Minor correction to how-pulumi-works

ddgenome:
### Proposed changes

Make a single correction to what appears to be an incorrect work, replacing "exists" with "exits".


lukehoban:
Thanks!

PR #2815, opened 2020-04-01
Handle inner content in the example shortcode

cnunciato:
This shortcode will have content passed into it, so we'll need to handle it [as we do the examples shortcode](https://github.com/pulumi/docs/blob/master/layouts/shortcodes/examples.html#L2).


cnunciato:
Closing as we'll incorporate this into an in-flight change.

PR #2809, opened 2020-04-01
Support for package details sections in resource docs

lukehoban:
Builds on pulumi/pulumi#4256 and https://github.com/pulumi/pulumi-terraform-bridge/pull/148.

Part of pulumi/home#734.

PR #2820, opened 2020-04-02
Regenerate AWS, Azure (AD), GCP resource docs

praneetloke:
### Proposed changes

This is a companion PR for some fixes being made to the resource docs generator and as such regenerated docs for the currently released docs.

Note that some of the providers now have a release tag of 2.0.0. (I have locked the `gen_resource_docs.sh` to use the `master` branch and temporarily not call `update_repos.sh`. The `master` branch currently represents the `1.x` release channel still.) ~~and so `update_repos.sh` updated my local copy of those repos to that tag, which brought in the changes to include the examples short code, which is mostly what you'll see as changed.~~ 

I also updated the `gen_resource_docs.sh` script to accept a provider name override to generate docs for a single provider, which is required for CI scenarios when we release a new provider. This [commit](https://github.com/pulumi/docs/pull/2820/commits/24037853eebd8518e5f3dde626b2a257a94f4251) has those changes. We are not yet ready to make the upstream script changes to make use of this (though I expect it to work). My main motivation for doing this is for my local development reasons.

I have split-up the commits by generating docs for each provider.

N/A

### Related issues (optional)

pulumi/pulumi#4269
https://github.com/pulumi/scripts/issues/104


praneetloke:
I rebased my `pulumi/pulumi` PR branch with the latest `master` which has some changes to the templates, specifically with adding some package details at the bottom of every resource page. I will update this repo with regenerated docs that will contain those changes as well. We will need to merge https://github.com/pulumi/docs/pull/2809 first, though, which contains the CSS used by that.

content/docs/reference/pkg/azure/_index.md
```diff
@@ -101,3 +101,13 @@ A Pulumi package for creating and managing Microsoft Azure cloud resources.
...
 </ul>
 
+<h3>Package Details</h3>
+<dl class="package-details">
+	<dt>Repository</dt>
+	<dd>https://github.com/pulumi/pulumi-azure</dd>
+	<dt>License</dt>
+	<dd>Apache-2.0</dd>
+	<dt>Notes</dt>
+	<dd></dd>
```
praneetloke, line 111: It seems like there is some issue with generating the package details in the doc. I'll take a look tomorrow.

praneetloke: My suspicion is that it is likely because the pulumi-terraform-bridge dep might need to be updated in the provider repos since the change that adds it to the schema was only merged today.

lukehoban: Yes - we'll need to update the providers to pull in the latest `pulumi-terraform-bridge` (and `pulumi`).  Alternatively we could suppress this `Notes` row if the data for it is not present in the schema for now.

lukehoban: Opened https://github.com/pulumi/pulumi/pull/4275 for the latter to unblock things here while we wait on getting all the providers updated.

praneetloke:
I am going to update `update_repos.sh` to limit to 1.x release tags for now until 2.0 tag is considered to be GA. Will regenerate the docs here with that.

praneetloke:
Alright folks. I cleaned-up the commits in this PR. The latest commits are the correct ones to review. I have also updated the PR description.

PR #2821, opened 2020-04-02
Reorganize webinar listings

zchase:
This PR adjusts the layout of the webinar page, adds a control for showing different lists of webinars, and adds "gated" functionality for on-demand videos.

The images on the webinar listings are probably going to change, but I thought it would be good to start getting everything else reviewed. So we can merge when we finalize what images we are going to use.


assets/js/webinar.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @param {string} date The string value of the date to localize.
+ */
+function localizeDateTime(date) {
+    var localeDate = new Date(date.getAttribute("data-datetime"));
+    var options = { timeZoneName: "short" };
+    return localeDate.toLocaleString(undefined, options);
+}
+
+$(function() {
```
justinvp, line 11: It's fine to keep this, but going forward, now that @cnunciato has hooked up Stencil.js builds in this repo, for all new JavaScript written, we should really consider leveraging that and phasing out our usage of JQuery. See https://github.com/pulumi/docs/tree/master/components. This way we can use modern language features in TypeScript, real components, etc.

layouts/partials/webinar-list-item.html
```diff
@@ -1,27 +1,43 @@
...
                     </a>
                 {{ end }}
-            </h3>
+                <p class="mb-5">
+                    {{ .Params.meta_desc }}
+                </p>
+                {{ if eq .Params.pre_recorded false}}
+                    <p class="webinar-datetime my-0 text-lg">
+                        <i class="fas fa-calendar mr-2"></i>
+                        <span data-datetime="{{ .Params.main.sortable_date }}">{{ .Params.main.datetime }}</span>
```
justinvp, line 33: These display as "Invalid Date" for me in Safari 13.1:

Screen Shot 2020-04-06 at 9 23 07 AM

layouts/partials/webinar-list-item.html
```diff
@@ -1,27 +1,43 @@
-<li class="w-full md:w-1/2 lg:w-1/3 m-0 p-2">
-    <article class="rounded shadow-md bg-white border border-gray-200 mb-10 flex flex-col mx-auto" style="max-width: 320px;">
-        <div style="height: 239px;">
+<li class="w-full m-0 p-2{{ cond (eq .Params.featured true) " order-first" "" }}">
+    {{ if eq .Params.featured true }}
+        <p class="hidden text-2xl font-bold lg:block">FEATURED</p>
+    {{ end }}
+    <article class="rounded shadow-md bg-white border border-gray-200 mb-10 flex flex-col lg:flex-row mx-auto">
+        <div class="w-full {{ cond (eq .Params.featured true) "lg:w-5/12" "lg:w-4/12" }}">
```
justinvp, line 6: Having the Featured one larger is OK, but it also sort of makes the ones following it look like they aren't sized/aligned correctly. (When I look at it I wonder "was this by design or a bug?").

Screen Shot 2020-04-06 at 9 26 24 AM

Also, the FEATURED heading above the featured item, almost makes it seem like it's the heading for all subsequent items. Wondering if we need another heading for the non-featured items or some kind of separator to make it more clear which one is the single featured item, vs the regular items.

assets/js/webinar.js
zchase: 👍  Sounds good.

layouts/partials/webinar-list-item.html
zchase: 👍

Issue #2654, opened 2020-03-14
Update docs to reflect areas where 2.0 is different from 1.0


leezen:

leezen:
PR is ready, but pushing to 0.34 to track that we haven't merged the PR and will only do so after 2.0 itself is out.

PR #2834, opened 2020-04-03
Update docs to reflect 2.0-ness

leezen:
Fixes https://github.com/pulumi/docs/issues/2654


content/docs/get-started/install/migrating-2.0.md
```diff
@@ -113,7 +87,7 @@ Then run `go mod download`
 Update your package reference to the latest version of the SDK:
 
 ```csharp
-<PackageReference Include="Pulumi" Version="2.0.0-beta.2" />
+<PackageReference Include="Pulumi" Version="2.0.0" />
```
stack72, line 90: so this will pin them to 2.0.0 - don't we want to use `2.*` as it will grown as the series grows?

content/docs/get-started/install/migrating-2.0.md
```diff
@@ -135,3 +109,14 @@ var clientConfig = await Pulumi.Azure.Core.GetClientConfig.InvokeAsync();
...
 {{< /chooser >}}
+
+## Remaining on Pulumi 1.0
+
+We recommend switching to Pulumi 2.0 if possible. We will only push critical security and bug fixes into the `1.x` branch. Other fixes, feature enhancements, and
+new functionality will not be supported in the `1.x` branch. In addition, provider updates will only be built against Pulumi 2.0.
+
+If you wish to remain on the `1.x` CLI, you can continue to download the CLI by referring to [the manual installation instructions]({{< relref "/docs/get-started/install/#manual-installation" >}}) and [choosing a specific version]({{< relref "/docs/get-started/install/versions" >}}).
+
+`pulumi new` will attempt to use the latest versions of the templates, which pull in the `2.0` SDK. You can continue to use the `1.x` templates by running `pulumi new https://github.com/pulumi/templates/tree/1.x`.
```
stack72, line 120: <3

content/docs/get-started/install/migrating-2.0.md
```diff
@@ -135,3 +109,14 @@ var clientConfig = await Pulumi.Azure.Core.GetClientConfig.InvokeAsync();
...
+
+Note that previous `1.x` templates for Python had `>=1.0.0` in `requirements.txt`. As a result, you may accidentally end up on the `2.x` SDK when updating dependencies from the `1.x` template.
```
lukehoban, line 122:
```suggestion
Note that previous `1.x` templates for Python had `>=1.0.0` in `requirements.txt`. As a result, you may accidentally end up on the `2.x` SDK when updating dependencies from the `1.x` template.  We encourage updating to `>=1.0.0,<2.0.0` if you intend to remain on the `1.x` releases.
```

stack72, line 122: this isn't needed - it has been ficed

content/docs/get-started/install/migrating-2.0.md
leezen: Ah, yes, of course. Let me fix that.

content/docs/get-started/install/migrating-2.0.md
leezen: I left this in b/c while it's been fixed going forward, any users who previously used 1.x templates might see this issue, so I left the note here in case people ran into that.

PR #2848, opened 2020-04-05
Data Sources => Functions

justinvp:
In the resource-oriented docs, these are Functions. We should be consistent in the Node.js API docs.

### DIFF
--- a/assets/sass/_api-nodejs.scss
+++ b/assets/sass/_api-nodejs.scss
@@ -76,9 +76,10 @@ ul.api {
                 }
             }
 
-            &.datasource {
+            &.datasource, // TODO: Remove `datasource` when it's no longer referenced anywhere.
+            &.function {
                 &:before {
-                    content: "D";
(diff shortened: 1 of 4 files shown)


assets/sass/_api-nodejs.scss
```diff
@@ -76,9 +76,10 @@ ul.api {
                 }
             }
 
-            &.datasource {
+            &.datasource, // TODO: Remove `datasource` when it's no longer referenced anywhere.
```
praneetloke, line 79: Ha I did this in a PR branch I am working on as well.

justinvp: Ah, sorry! Thought this would be a tiny thing I could do, but you're already on top of it! 😬

praneetloke: No I absolutely appreciate it! Thank you so much of thinking to do this. :)

PR #2849, opened 2020-04-06
fix missing go stack reference docs

EvanBoyle:
Looks like we missed two language choosers (stack references). 

I did a global search for `https://github.com/pulumi/pulumi/issues/1614` and the only other references were for dynamic providers and the 2.0 roadmap blog post so this should be it.

### DIFF
--- a/content/docs/intro/concepts/organizing-stacks-projects.md
+++ b/content/docs/intro/concepts/organizing-stacks-projects.md
@@ -124,9 +124,7 @@ pulumi.export("kubeConfig", ... a cluster's output property ...)
 {{% choosable language go %}}
 
 ```go
-// StackReference is not supported in Go currently.
-//
-// See https://github.com/pulumi/pulumi/issues/1614.
+ctx.Export("kubeConfig", /*...a cluster's output property...*/)
 ```
 
 {{% /choosable %}}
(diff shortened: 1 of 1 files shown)


content/docs/intro/concepts/organizing-stacks-projects.md
```diff
@@ -197,9 +195,21 @@ service = core.v1.Service(..., ResourceOptions(provider=provider))
...
-// See https://github.com/pulumi/pulumi/issues/1614.
+import (
+	"fmt"
+
+	"github.com/pulumi/pulumi/sdk/go/pulumi"
+)
+
+func main() {
+	pulumi.Run(func(ctx *pulumi.Context) error {
+    org := cfg.Require("org")
```
lukehoban, line 206: Where does cfg come from?  Also is indentation off here?

content/docs/intro/concepts/organizing-stacks-projects.md
```diff
@@ -197,9 +195,21 @@ service = core.v1.Service(..., ResourceOptions(provider=provider))
...
+		slug := fmt.Sprintf("%v/%v/%v", org, ctx.Project(), ctx.Stack())
+		stackRef, err := pulumi.NewStackReference(ctx, slug, nil)
+  }
+
+  kubeConfig := stackRef.GetOutput(pulumi.String("kubeConfig"))
```
lukehoban, line 211: This looks like it’s not inside the Run call?

content/docs/intro/concepts/organizing-stacks-projects.md
EvanBoyle: Oops, meant to remove the reference to config.

EvanBoyle: Yeah, I think this is a tabs vs spaces thing. It only shows up in the diff, must be due to my editor settings. The rendering on the page looks fine, but I fixed it for cleanliness.

content/docs/intro/concepts/organizing-stacks-projects.md
```diff
@@ -197,9 +195,22 @@ service = core.v1.Service(..., ResourceOptions(provider=provider))
...
-//
-// See https://github.com/pulumi/pulumi/issues/1614.
+import (
+  "fmt"
+
+  "github.com/pulumi/pulumi/sdk/go/pulumi"
+)
+
+func main() {
+  pulumi.Run(func(ctx *pulumi.Context) error {
```
justinvp, line 205: Should tabs be used for indentation so if someone copies/pastes, it will already formatted in the way one would expect for Go?

EvanBoyle: This is consistent with the other code examples in other languages in this file. Not sure if we have some sort of editor config checked in, but this is the behavior of `tab` in vs code. 

`go fmt` will sort this out when users take the snipped anyway, so not too concerned about it.

PR #2853, opened 2020-04-06
Generate all resource docs

praneetloke:
### Proposed changes

As the title states, this PR includes the resource docs for all of the providers that support generating a schema. I noticed that 3 providers, `pulumi-aiven`, `pulumi-gitlab` and `pulumi-mailgun` don't have the `Makefile` target to generate a schema. We can add docs for those once they are updated.

I made some minor adjustments to the css of anchor tags inside list items. These are used in the index pages. Please see this [commit](https://github.com/pulumi/docs/commit/34662791ac9ccf7d251f473202cbc8b952d2b8cd).

N/A

### Related issues (optional)

These docs were generated with changes from https://github.com/pulumi/pulumi/pull/4308.


scripts/gen_all_resource_docs.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
praneetloke: I used this script to generate _all_ of these docs, so I figured I would add this to the repo as well. I don't anticipate this being in service for too long, though. Once we switch to automatically generating the docs from upstream when a provider release is made, this would not be needed.

assets/sass/_api-nodejs.scss
```diff
@@ -42,7 +42,9 @@ ul.api {
         @apply truncate mt-0;
 
         a {
-            @apply btn bg-transparent inline-block p-1;
+            @apply btn truncate bg-transparent inline-block p-1;
```
cnunciato, line 45: If we're going to truncate these, we should make sure they have `title` tags also, so you can see what they are. (I'm usually not a fan of truncating because it obscures important info, but when it's necessary, it should always be possible to figure out what's been lopped off.)

cnunciato:
Most of these seem to be causing an unclosed blockquote somewhere:

image

assets/sass/_api-nodejs.scss
praneetloke: > If we're going to truncate these, we should make sure they have title tags also, so you can see what they are

Yep I did this.

> I'm usually not a fan of truncating because it obscures important info, but when it's necessary, it should always be possible to figure out what's been lopped off.

They get cut-off today anyway if the name is too long. This just ensures that we show an ellipsis, as well as show a tooltip using the `title` attribute. Admittedly, latter is a change I have made only to the resource docs templates and not to the API docs ones.

praneetloke:
> Most of these seem to be causing an unclosed blockquote somewhere:

I bet it is because of a missing new-line between the blockquote on the previous and the start of the `h3` tag. I'll look into it.

content/docs/reference/pkg/alicloud/_index.md
```diff
@@ -0,0 +1,85 @@
...
+<!-- WARNING: this file was generated by Pulumi Docs Generator. -->
+<!-- Do not edit by hand unless you're certain you know what you are doing! -->
+
+A Pulumi package for creating and managing AliCloud resources.
+
+> This provider is a derived work of the [Terraform Provider](https://github.com/terraform-providers/terraform-provider-alicloud)
+> distributed under [MPL 2.0](https://www.mozilla.org/en-US/MPL/2.0/). If you encounter a bug or missing feature,
+> first check the [`pulumi/pulumi-alicloud` repo](https://github.com/pulumi/pulumi-alicloud/issues); however, if that doesn't turn up anything,
+> please consult the source [`terraform-providers/terraform-provider-alicloud` repo](https://github.com/terraform-providers/terraform-provider-alicloud/issues).
+<h3>Modules</h3>
```
cnunciato, line 18: Looks like an empty line needs to be added to the template in order to create some space between the `>` and the `<h3`>.

praneetloke: This is fixed with the latest set of commits.

praneetloke:
Ok this PR has been updated with improvements to the `gen_resource_docs.sh` script which will install the required provider plugin so that the examples are properly converted and included in the docs.

lukehoban:
:tada:

PR #2871, opened 2020-04-07
Update dates for london event; add note to top of page w/ link to webinars

zchase:
This PR updates the dates for DevOpsCon London and adds a note to the top of the events page linking to our new webinar pages.

### DIFF
--- a/layouts/events/list.html
+++ b/layouts/events/list.html
@@ -13,14 +13,12 @@
 
 {{ define "main" }}
     <div class="container mx-auto">
-        <div class="w-10/12 mx-auto mt-10 note note-info" role="alert">
-            <p class="text-lg">
+        <div class="note note-info w-11/12 mx-auto mt-10" role="alert">
+            <p class="text-base">
                 <i class="fas fa-info-circle pr-2"></i>
-                The demand for Pulumi content and education continues to grow and, even in
-                the face of restricted travel, we will ramp it up further. Although many events
-                worldwide we’d planned to sponsor or otherwise participate in have been canceled,
-                we have shifted to working with customers through digital channels wherever we can.
-                Checkout our <a href="{{ relref . "/webinars" }}"><b>upcoming webinars</b></a>.
+                In light of recent events and restricted travel, we’ve had to cancel many of our events planned for
+                the first half of this year. To keep up with demand, we’ve scheduled a series of free remote tutorials, tech talks
+                and workshops — check out our <a href="{{ relref . "/webinars" }}"><strong>upcoming sessions and on-demand content</strong></a>.
             </p>
         </div>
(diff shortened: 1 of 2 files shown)


layouts/events/list.html
```diff
@@ -13,7 +13,17 @@ <h1>Upcoming Events</h1>
...
     <div class="container mx-auto">
-        <div class="flex justify-between py-10 md:py-24 px-5 md:px-20">
+        <div class="w-10/12 mx-auto mt-10 note note-info" role="alert">
+            <p class="text-lg">
+                <i class="fas fa-info-circle pr-2"></i>
+                The demand for Pulumi content and education continues to grow and, even in
+                the face of restricted travel, we will ramp it up further. Although many events
+                worldwide we’d planned to sponsor or otherwise participate in have been canceled,
+                we have shifted to working with customers through digital channels wherever we can.
+                Checkout our <a href="{{ relref . "/webinars" }}"><b>upcoming webinars</b></a>.
```
cnunciato, line 23:
```suggestion
                Check out our <a href="{{ relref . "/webinars" }}"><b>upcoming webinars</b></a>.
```

layouts/events/list.html
```diff
@@ -13,7 +13,17 @@ <h1>Upcoming Events</h1>
 
 {{ define "main" }}
     <div class="container mx-auto">
-        <div class="flex justify-between py-10 md:py-24 px-5 md:px-20">
+        <div class="w-10/12 mx-auto mt-10 note note-info" role="alert">
+            <p class="text-lg">
```
cnunciato, line 17:
```suggestion
            <p class="text-base">
```

layouts/events/list.html
```diff
@@ -13,7 +13,17 @@ <h1>Upcoming Events</h1>
 
 {{ define "main" }}
     <div class="container mx-auto">
-        <div class="flex justify-between py-10 md:py-24 px-5 md:px-20">
+        <div class="w-10/12 mx-auto mt-10 note note-info" role="alert">
```
cnunciato, line 16: This doesn't quite seem to line up with the sections below it:

image

And needs padding to line up properly on mobile (we apply the same padding pretty much everywhere):

image

layouts/events/list.html
cnunciato, line 23: I'd prefer without the bold, but if we want to make this stand out, we should use `<strong>` instead, 'cause semantic.

layouts/events/list.html
zchase: 👍

layouts/events/list.html
zchase: 👍

zchase: 👍

PR #2876, opened 2020-04-07
further cleanup of concepts

spara:
- minor edits, e.g., replaced reference to "Kubernetes the Prod Way" to "Crosswalk for Kubernetes"


content/docs/intro/concepts/project.md
```diff
@@ -9,7 +9,7 @@ menu:
 aliases: ["/docs/reference/project/"]
 ---
 
-A Pulumi project is any folder which contains a `Pulumi.yaml` file.  When in a subfolder, the closest enclosing folder with a `Pulumi.yaml` file determines the current project.  A new project can be created with `pulumi new`.  A project specifies which runtime to use, which determines where to look for the program that should be executed during deployments.  Currently, `nodejs` and `python` are supported runtimes.
+A Pulumi project is any folder which contains a `Pulumi.yaml` file.  When in a subfolder, the closest enclosing folder with a `Pulumi.yaml` file determines the current project.  A new project can be created with `pulumi new`.  A project specifies which runtime to use, which determines where to look for the program that should be executed during deployments.  Supported runtimes are `nodejs`, `python`, `.NET`, and `Go`.
```
justinvp, line 12: These are values that go inside `Pulumi.yaml`:

```suggestion
A Pulumi project is any folder which contains a `Pulumi.yaml` file.  When in a subfolder, the closest enclosing folder with a `Pulumi.yaml` file determines the current project.  A new project can be created with `pulumi new`.  A project specifies which runtime to use, which determines where to look for the program that should be executed during deployments.  Supported runtimes are `nodejs`, `python`, `dotnet`, and `go`.
```

PR #2877, opened 2020-04-07
Update the pricing page

joeduffy:
This change clarifies a few aspects of our pricing page:

* Add a "START TRIAL" button at the very top and make it
   clear that we offer 14-day free trials of all editions.

* Add a billing cycle chooser to switch between Monthly and
  Annually (the default) prices.

* Move the user count to the top of each box, and bold the
  numbers, to make them stand out as clearly different.

* Add tooltips to explain the concept of $225/270 min buy-in
   for Team Pro. Also add better explanation to the details below.

* Trim the bulleted list -- it's gotten quite lengthy. That
  includes removing the "free" and "trial" points, since this
  information is already represented in other areas (headers,
  buttons, etc).

### DIFF
--- a/layouts/page/pricing.html
+++ b/layouts/page/pricing.html
@@ -1,8 +1,8 @@
 {{ define "hero" }}
     <header class="header-hero">
         <div class="container mx-auto text-center flex-col max-w-2xl">
-            <h1>Pulumi for every team and individual.</h1>
-            <p>
+            <h1>Plans for every team and individual.</h1>
+            <p class="text-gray-400">
                 Whether you're using open source, our free Community Edition, or using
                 Pulumi in your team or entire organization, we have a solution for you.
             </p>
@@ -12,7 +12,7 @@
 
 {{ define "main" }}
     <section class="bg-purple-800 pt-12 text-white px-4 justify-center text-center" id="editions">
-        <div class="container mx-auto lg:flex justify-center pb-8 hidden lg:block">
+        <div class="container mx-auto lg:flex justify-center pb-4 hidden lg:block">
             <div class="lg:w-3/4 pr-1">
                 <div class="container" style="border-bottom: 2px solid #a48bb3">
                     <p class="text-purple-200 text-sm font-bold">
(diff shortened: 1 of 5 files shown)


joeduffy:
Before:

image

After:

image

lukehoban:
The After looks like it implies that 25 Team Pro users costs $225/mo. There is no longer any clear indication that this is priced per-user. (I suspect the info widget next to users says this if I hover it - but that might be too subtle?).  Perhaps that line should say “3 Team Member included - additional users $75/mo”?

joeduffy:
I appreciate the feedback. It's a little tricky to do this in a way that's not confusing, due to our "minimum of 3"; however, here are two reference comparisons which effectively do this same thing:

[LaunchDarkly](https://launchdarkly.com/pricing/):

image

[Netlify](https://www.netlify.com/pricing/):

image

Indeed, as you say, the tooltip explains this in more detail:

image

Also note that if you scroll down, the fine print is spelled out:

image

Ultimately, this feels more genuine than saying something like "$17/user/mo" for Team Starter and "$75/user/mo" for Team Pro, since then you'd be surprised that the minimum is 3x that.

All that said, still discussing and experimenting, so will see what we can do to make this clearer.

joeduffy:
@lukehoban Appreciate the feedback, PTAL. This feels cleaner to me and does a better job of describing how the pricing works.

@zchase @cnunciato and I are discussing whether and if so how to A/B test the new trial button.

layouts/page/pricing.html
```diff
@@ -2,17 +2,19 @@
     <header class="header-hero">
         <div class="container mx-auto text-center flex-col max-w-2xl">
             <h1>Pulumi for every team and individual.</h1>
-            <p>
-                Whether you're using open source, our free Community Edition, or using
-                Pulumi in your team or entire organization, we have a solution for you.
+            <p class="text-gray-400 mb-8">
+                Try any edition risk-free for 14-days. No credit card required.
             </p>
+            <a data-track="team-starter" class="text-center btn btn-lg btn-blue mx-auto mt-8 py-4 px-6" href="https://app.pulumi.com/site/organizations/add">
```
zchase, line 8:
```suggestion
            <a data-track="hero-free-trial" class="text-center btn btn-lg btn-blue mx-auto mt-8 py-4 px-6" href="https://app.pulumi.com/site/organizations/add">
```

The `data-track` for this should describe what it is so we can call it out in Google Analytics a bit more.

cnunciato:
@joeduffy What's motivating the content and CTA changes to the hero area specifically? 

TBH this part concerns me, because the placement and messaging suggest they apply to all plans, since they're center-positioned above all of them, like an umbrella. As I look at this, here's how I read it:

image

Given the work we did to dispel the impression that Pulumi was a paid-only service, and the goal being to encourage new users to try Pulumi Community, I'm not sure it's a good idea to do this.

joeduffy:
> What's motivating the content and CTA changes to the hero area specifically?

The motivation for the hero changes is to drive more trials and to measure whether it does that. I would prefer to A/B test this and would like to discuss the best way to go about that.

Note that the free trial _does_ apply to all editions, except for community where it's vacuously "true" since it is free forever. This is a flaw in the current page where it looks like you don't get to trial Enterprise without first speaking to a salesperson, which isn't correct (we let you trial and then contact you). FWIW, this is the same approach GitLab has taken: https://about.gitlab.com/pricing/. Also figure if someone is on this page, they are looking for pricing information, so expect the commercial twist.

For what it's worth, I don't think this page does a very good job explaining the open source difference _as it stands_ other than one subtle mention in the paragraph. One idea I had to improve on this was to mimic the subtext from the homepage CTA: say something like

> Pulumi open source and Community Editions are free forever.

And then hyperlink "open source" to a page that describes the difference between open source and Community Edition can hyperlink to the CE pricing element. This feels like it both improves the open source message as well as having a more prominent trial CTA. What do you think?

joeduffy:
>  I think we should display the $225/mo price 

This ended up feeling like a bit too significant a change for now (based also on feedback from sales team). We are going to keep evaluating this going forward but will probably want to A/B test that. It's unfortunately not super clear what metrics to use to validate success. Note that I took inspiration from GitHub's pricing page which has our same conundrum -- see https://github.com/pricing.

PR #2889, opened 2020-04-08
Update Go API links

lukehoban:
Add Kubernetes and link to `v2` of Azure and GCP.


stack72:
Oh nice - we need to add this to the list for the 2.0 release when I do this to all providers!

lukehoban:
Note that the links from the resource-based docs are also broken in the same way - but that will need to be fixed in the code generators in `pulumi/pulumi`.  I'll open a separate PR on that.

PR #2891, opened 2020-04-08
Add a go module for resourcedocsgen

praneetloke:
This PR adds a go.mod file for `tools/resourcedocsgen`, which seems to be required as the dependencies in `pulumi/pulumi` are not vendored.


tools/resourcedocsgen/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/pulumi/docs/tools/resourcedocsgen
```
praneetloke, line 1: @EvanBoyle this is what I added. Moreover, trying to run this tool from any other path other than this folder always seems to fail. So I had to adjust the `gen_resource_docs.sh` script accordingly.

tools/go.mod
```diff
@@ -1,3 +0,0 @@
-module github.com/pulumi/docs/tools
```
praneetloke, line 1: @lukehoban looks like you added this module file recently. In this PR, I added a module directly for `resourcedocsgen`, so I removed this one. Is there a preference to keep this over what I have done?

PR #2902, opened 2020-04-09
Generate the latest resource docs for all enabled providers

praneetloke:
### Proposed changes

Just as the title says. I also enabled GitLab as well since it has the appropriate make target to be able to generate the schema. This includes the new display name change for all of the providers in the index front-matter. (There will be other changes too since the providers themselves may have been updated as well.)

N/A

### Related issues (optional)
N/A


praneetloke:
FYI @sean1588 @jkisk @ekrengel

content/docs/reference/pkg/gitlab/_index.md
```diff
@@ -0,0 +1,63 @@
+---
+title: "gitlab"
```
praneetloke, line 2: Looks like we need to add gitlab to the lookup map. I'll add it in a PR I have going for something else.

PR #2907, opened 2020-04-10
Change Support link in footer to send to Zendesk.

clstokes:
### Proposed changes

Changes the *_Support_* link in the footer to send users to https://support.pulumi.com/. 

This does not change or remove the existing support page at https://www.pulumi.com/support/, so any existing links to that will continue to work. Will address that in a future PR.

image


clstokes:
@zchase this is a really minimal change, but tagged you for review in case there's any SEO concern here.

PR #2920, opened 2020-04-13
Add check for required Hugo version

chrsmith:
Add a check that you are running Hugo v0.55.4, as required to properly render the site's HTML. If you are on a newer version of Hugo, the build will complete successfully but fail to render HTML in the correct format.

We just so happen to install the right version of Hugo in CI, but adding a check will save developer time by failing fast. (It's likely this will break some developers local flow, since depending on what parts of the site you are editing, you might not notice any problems running a newer version of Hugo.) But it's probably better to just add this check.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ lint_markdown:
 .PHONY: serve
 serve:
 	@echo -e "\033[0;32mSERVE:\033[0m"
+	./scripts/check-hugo-version.sh
 	yarn lint-markdown --no-error
 	yarn --cwd components run build
 	$(MAKE) copy_static_prebuilt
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -18,6 +18,7 @@ banner:
 ensure:
 	yarn install
 	yarn --cwd components install
+	./scripts/check-hugo-version.sh
```
cnunciato, line 21: Any reason not to add this to `serve` as well?

cnunciato:
Closing as outdated, but we did add this functionality in pulumi-hugo: https://github.com/pulumi/pulumi-hugo/blob/master/scripts/ensure.sh#L19-L23

PR #2922, opened 2020-04-13
Specify the options in the langfile shortcode

### LINKED ISSUE #2909: `Uncaught TypeError` on some pages with `pulumi-choosable`
For example, on this page: https://www.pulumi.com/docs/get-started/aws/review-project/

Screen Shot 2020-04-10 at 2 48 37 PM

```
p-uxr2padp.entry.js:1 Uncaught TypeError: Cannot read property 'key' of undefined
    at i (p-uxr2padp.entry.js:1)
    at p-uxr2padp.entry.js:1
    at o (p-7ce3a4d1.js:1)
    at Object.mapStateToProps (p-7ce3a4d1.js:1)
    at h.onRendered (p-uxr2padp.entry.js:1)
    at HTMLDocument.<anonymous> (p-7ce3a4d1.js:1)
    at Object.emit (p-7ce3a4d1.js:1)
    at N.componentDidRender (p-uxr2padp.entry.js:1)
    at it (p-7ce3a4d1.js:1)
    at rt (p-7ce3a4d1.js:1)
```

I haven't debugged to closely, but looks like it's happening around this line:

https://github.com/pulumi/docs/blob/fa8080aae0f05ca7126054418be7ecff4554b72a/components/src/components/chooser/chooser.tsx#L157

cnunciato:
These [should be specified](https://github.com/pulumi/docs/tree/master/components#usage), but it looks like I never did that. 🤦

Also removes an unused CSS class.

Fixes #2909.

---

Updated to apply this change to all shortcodes, and to throw when either `type` or `options` is missing.

### DIFF
--- a/components/src/components/chooser/chooser.tsx
+++ b/components/src/components/chooser/chooser.tsx
@@ -219,14 +219,21 @@ export class Chooser {
     private parseOptions() {
         this.currentOptions = [];
 
-        if (this.type && this.options) {
-            try {
-                const keys: string[] = this.options.split(",").map(s => s.trim());
-                this.mapOptions(this.type, keys as ChooserKey[]);
-            }
-            catch (err) {
-                console.error(`Error parsing chooser options "${this.options}"`, err);
-            }
+        if (!this.type) {
+            throw new Error("Chooser attribute `type` is required.");
+        }
+
+        if (!this.options) {
+            throw new Error("Chooser attribute `options` is required.");
+        }
+
+        try {
+            const keys: string[] = this.options.split(",").map(s => s.trim());
+            this.mapOptions(this.type, keys as ChooserKey[]);
+        }
+        catch (err) {
+            console.error(`Error parsing chooser options "${this.options}"`);
+            throw(err);
         }
     }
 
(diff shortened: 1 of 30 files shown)


justinvp:
There are several other shortcodes that follow this pattern. Do they need to be updated similarly?

justinvp:
I also just merged a change that has a handful of new `policy-*.html` shortcodes.

cnunciato:
These really should all have both `type` and `options` specified, yeah -- my oversight, sorry about that. They work because they happen to be responding to what's set globally (and the error is due to the presence of `type` but absence of `options`), but I can't think of a reason why these properties shouldn't both be required, other than just to make components a little "smarter", perhaps.

I'll go through and add them all to this PR, unless you feel otherwise?

justinvp:
> I'll go through and add them all to this PR, unless you feel otherwise?

Sounds good. Thanks!

Should it throw an explicit error if `type` and `options` aren't specified?

cnunciato:
> Should it throw an explicit error if type and options aren't specified?

I was thinking the same!

components/src/components/chooser/chooser.tsx
```diff
@@ -219,6 +219,14 @@ export class Chooser {
...
 
+        if (!this.type) {
+            throw new Error("Chooser attribute `type` is required.");
+        }
+
+        if (!this.options) {
+            throw new Error("Chooser attribute `options` is required.");
+        }
+
         if (this.type && this.options) {
```
justinvp, line 230: Does this check need to remain?

cnunciato: Nope, removed!

PR #2954, opened 2020-04-14
Regenerate all resource docs

praneetloke:
### Proposed changes

As the title states, I have regenerated resource docs for all of the providers. This contains all changes in the generator made in the last few days which includes the following fixes:

https://github.com/pulumi/pulumi/pull/4374
https://github.com/pulumi/pulumi/pull/4360
https://github.com/pulumi/pulumi/pull/4351

In addition, I have also updated the `gen_resource_docs.sh` script to checkout a provider repo to the latest released v1 tag if the script is not invoked with a version. I needed to do this because `master` branch now contains 2.0 features which is not yet GA.

N/A

### Related issues (optional)

N/A


content/docs/reference/pkg/alicloud/adb/getclusters.md
```diff
@@ -366,11 +358,19 @@ The following output properties are available:
     <dd>{{% md %}}A list of ADB cluster IDs. 
 {{% /md %}}</dd>
 
+    <dt class="property-"
```
cnunciato, line 361: Not this PR I assume,but this doesn't look right; presumably something amiss here: https://github.com/pulumi/pulumi/blob/729ae08c9d4bf0099e6382ac5d176540ac16f98a/pkg/codegen/docs/templates/properties.tmpl#L7

content/docs/reference/pkg/alicloud/actiontrail/gettrails.md
```diff
@@ -291,26 +291,26 @@ The following output properties are available:
...
+    <dd>{{% md %}}A list of trail names.
+{{% /md %}}</dd>
 
     <dt class="property-"
             title="">
-        <span>names</span>
+        <span>name<wbr>Regex</span>
         <span class="property-indicator"></span>
-        <span class="property-type">string[]</span>
+        <span class="property-type">string</span>
```
cnunciato, line 305: Is this change right?

cnunciato: Ah, maybe it's just that the diff rendering is making this look weird.

praneetloke: Just adding a comment for others wondering why this diff is ok to ignore. This is because the properties are sorted by whether or not they are required. It's from https://github.com/pulumi/pulumi/pull/4351.

PR #2955, opened 2020-04-15
Add UTM param tracking to form submissions

zchase:
Currently, our tracking for form submissions is missing UTM parameter tracking in HubSpot. This PR adds an analytics `track` event so we can track the UTM parameters in Metabase. I played around with a few solutions for this such as having our Lambda as an endpoint, adding hidden form fields and trying to associate the parameters per submission in HubSpot, but this implementation seemed like the best solution for us right now. I know this adds a bit of jQuery and we are trying to move over to `stencil.js`. Hopefully, you can forgive me for this one.

This type of tracking is really important as we venture into paid advertising.

Fixes: https://github.com/pulumi/home/issues/758

### DIFF
--- a/assets/js/webinar.js
+++ b/assets/js/webinar.js
@@ -8,6 +8,40 @@ function localizeDateTime(date) {
     return localeDate.toLocaleString(undefined, options);
 }
 
+/**
+ * This function parses the cookie string and returns an object of the current cookies.
+ */
+function parseCookie() {
+    var cookieObject = {};
+    var cookieString = document.cookie;
+    var parsedCookie = cookieString.split(";").map(function(cookie) {
+        var parts = cookie.split("=");
+        cookieObject[parts[0]] = parts[1];
+        return cookie;
+    });
+
+    return cookieObject
+}
+
+/**
+ * This function parses the UTM cookie and returns and object of the cookie.
+ *
+ * @param {string} utmCookieString The value of the '__utmzz' cookie. Values are separated by '|'.
+ */
+function parseUTMCookie(utmCookieString) {
+    var cookie = utmCookieString ? utmCookieString : "";
+    var parts = cookie.split("|");
+    var utmObj = {};
+
+    for (let i = 0; i < parts.length; i++) {
+        var value = parts[i];
+        var valueParts = value.split("=");
+        utmObj[valueParts[0]] = utmObj[1];
+    }
+
+    return parseUTMCookie;
+}
+
 $(function() {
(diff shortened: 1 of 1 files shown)


assets/js/webinar.js
```diff
@@ -8,6 +8,40 @@ function localizeDateTime(date) {
...
+    var parts = cookie.split("|");
+    var utmObj = {};
+
+    for (let i = 0; i < parts.length; i++) {
+        var value = parts[i];
+        var valueParts = value.split("=");
+        utmObj[valueParts[0]] = utmObj[1];
+    }
+
+    return parseUTMCookie;
```
justinvp, line 42: `utmObj[valueParts[0]] = utmObj[1];` doesn't look right.

How about something like this:

```suggestion
    return (utmCookieString || "").split("|").reduce(function(obj, utm) {
        var parts = utm.split("=");
        obj[parts[0]] = parts[1];
        return obj;
    }, {});
```

assets/js/webinar.js
```diff
@@ -8,6 +8,40 @@ function localizeDateTime(date) {
...
+function parseCookie() {
+    var cookieObject = {};
+    var cookieString = document.cookie;
+    var parsedCookie = cookieString.split(";").map(function(cookie) {
+        var parts = cookie.split("=");
+        cookieObject[parts[0]] = parts[1];
+        return cookie;
+    });
+
+    return cookieObject
```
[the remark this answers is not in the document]
zchase: 👍

assets/js/webinar.js
zchase: 👍

assets/js/webinar.js
```diff
@@ -45,4 +71,35 @@ $(function() {
...
+    if (webinarDetails && analyticsAvailable) {
+        // Parse the cookie and grab the UTM cookie values.
+        var cookies = parseCookie();
+        var utmCookie = parseUTMCookie(cookies["__utmzz"]);
+
+        // Listen for the HubSpot form to be loaded.
+        window.addEventListener('message', event => {
+            if(event.data.type === 'hsFormCallback' && event.data.eventName === 'onFormReady') {
+                var form = $(".hbspt-form form");
+                console.log(form);
```
justinvp, line 88: Was it intentional to keep the `console.log`?

zchase: Nope, just a mistake after a long day.. Thanks for pointing this out.

Issue #2866, opened 2020-04-07
Unify the styling of deprecation notices in API and resource docs


cnunciato:
We're applying different styling to deprecation notices in the resource docs and API docs:

https://www.pulumi.com/docs/reference/pkg/aws/cognito/userpool/#userpooladmincreateuserconfig

![image](https://user-images.githubusercontent.com/274700/78696810-c69f5a80-78b4-11ea-9d07-eec3f7c61f81.png)

https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/cloud/#HttpEndpoint

![image](https://user-images.githubusercontent.com/274700/78696788-bedfb600-78b4-11ea-9fdf-fa0d1bcfac57.png)

Both are using different CSS classnames. We should apply the same classname in both places to ensure they look the same.

jkisk:
@cnunciato which of the two styles do you prefer?

cnunciato:
@jkisk I think the second one, with the orange, is pretty nice. What do you think?

jkisk:
@cnunciato I agree, so I will use that one!

jkisk:
@cnunciato , with a fairly small change we can go from: 
![image](https://user-images.githubusercontent.com/41454626/79392783-ce9c7180-7f28-11ea-93af-3fb3aad6830e.png)
to: 
![image](https://user-images.githubusercontent.com/41454626/79392819-dd832400-7f28-11ea-9725-b553da0dab29.png)
and with another small change in another repo, we could make the word "DEPRECATED" appear in all caps... rather than open two PRs, just inquiring here.  The original style is a collection of css that is layered, so a simple switch of classnames won't work.

jkisk:
without code block highlighting: 
![image](https://user-images.githubusercontent.com/41454626/79394077-7e72de80-7f2b-11ea-895a-9c516bb41242.png)

PR #2973, opened 2020-04-16
Adjust styling of deprecation notice.

jkisk:
This change makes the deprecation notices in the resource docs closer to the ones in the API docs.  See discussion in issue https://github.com/pulumi/docs/issues/2866.

Before:
image

After:
image


jkisk:
Thank you!

PR #3023, opened 2020-04-19
Regenerate all resource docs with the latest generator code

praneetloke:


content/docs/reference/pkg/alicloud/actiontrail/getinstances.md
```diff
@@ -21,9 +21,9 @@ import * as alicloud from "@pulumi/alicloud";
 const config = new pulumi.Config();
 const instanceName = config.get("instanceName") || "alikafkaInstanceName";
 
-const defaultZones = alicloud.getZones({
+const defaultZones = pulumi.output(alicloud.getZones({
     availableResourceCreation: "VSwitch",
-});
+}, { async: true }));
```
joeduffy, line 26: Isn't async now the default as part of 2.0? Can we eliminate this?

Also, why the addition of `pulumi.output` here?

praneetloke: For the `async: true` -- it was added as part of this [PR](https://github.com/pulumi/tf2pulumi/pull/155/files) in `tf2pulumi`.

As for the `pulumi.output`...I am not sure of the history here with that. cc @pgavlin  to see if he has any ideas.

praneetloke: Also note that I regenerated these docs with the `master` branch of `tf2pulumi`. I notice that there is a `v2` branch, but I don't think that's ready for use yet?

PR #3024, opened 2020-04-19
Update target version for typescript config

rtkelly13:
### Proposed changes

When creating a new project using pulumi new aws-typescript the typescript file generated target es2016.
This can also be seen here https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/tsconfig.json


content/docs/intro/languages/javascript.md
```diff
@@ -112,7 +112,7 @@ Tools like VS Code will give you completion lists, live error reporting and inli
 
 When using the built in TypeScript support, Pulumi sets the following compiler settings, which may not be overridden:
 
-- target: "es6",
+- target: "es2016",
```
justinvp, line 115: Note: Pulumi actually still defaults to `es6`: https://github.com/pulumi/pulumi/blob/dc2d91e294916a3d74be7d98f1b67e674182b0ee/sdk/nodejs/cmd/run/run.ts#L158

```suggestion
- target: "es6",
```

content/docs/intro/languages/javascript.md
```diff
@@ -78,9 +78,9 @@ When using Pulumi's built in TypeScript support, a `tsconfig.json` file is optio
 {
     "compilerOptions": {
         "outDir": "bin",
-        "target": "es6",
+        "target": "es2016",
         "lib": [
-            "es6"
+            "es2016"
```
justinvp, line 83: This whole `lib` setting could be deleted.

justinvp: Or just replace this whole snippet with the `tsconfig.json` from the TypeScript templates: https://github.com/pulumi/templates/blob/master/typescript/tsconfig.json

Issue #2958, opened 2020-04-15
Change GCP "Getting Started" example to not require CryptoKey


komalali:
### Problem description

- The GCP [getting started guide](https://www.pulumi.com/docs/get-started/gcp/modify-program/) involves creating a CryptoKey, which is then impossible to destroy without "owner" roles. Our users may not have the required credentials to destroy this resource.

### Suggestions for a fix

- We might want to change the getting started example to avoid touching resources that might have restricted permissions.

PR #2964, opened 2020-04-15
Update Azure getting started

leezen:
* Fix Azure getting started to use tags instead of https to force a change.
* Make create project match actual outputs.

PR #3043, opened 2020-04-21
Update gcp getting started to be smooth like butter

EvanBoyle:
fixes https://github.com/pulumi/docs/issues/2958

Right now GCP getting started requires updating account privileges to be successful. This should not be necessary. Updating this example to sidestep this problem in a similar way to Azure https://github.com/pulumi/docs/pull/2964

Tested all of these changes manually.

### DIFF
--- a/content/docs/get-started/gcp/deploy-stack.md
+++ b/content/docs/get-started/gcp/deploy-stack.md
@@ -24,9 +24,9 @@ This command instructs Pulumi to determine the resources needed to create the st
 $ pulumi up
 Previewing update (dev):
      Type                   Name               Plan       Info
-     pulumi:pulumi:Stack    quickstart-dev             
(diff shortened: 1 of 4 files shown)


content/docs/get-started/gcp/deploy-stack.md
```diff
@@ -21,16 +21,20 @@ $ pulumi up
 This command instructs Pulumi to determine the resources needed to create the stack. First, a preview is shown of the changes that will be made:
 
 ```
+$ pulumi up
 Previewing update (dev):
+     Type                   Name               Plan       Info
```
komalali, line 26: This looks like the code to modify (from deploy-changes.md), not the initial preview. Did you mean to make this change here?

EvanBoyle: Good catch!

Issue #3055, opened 2020-04-21
Document ability to use an exported `async` function as an entrypoint


lukehoban:
I don't believe we've documented the support added in https://github.com/pulumi/pulumi/pull/3321.  This is more important now in 2.0 as a workaround for the lack of `getOutputSync` on `StackReference` per discussion in e.g. https://github.com/pulumi/pulumi/issues/4462.

PR #3058, opened 2020-04-21
Include details on async entrypoint

leezen:
Fixes https://github.com/pulumi/docs/issues/3055

### DIFF
--- a/content/docs/get-started/install/migrating-2.0.md
+++ b/content/docs/get-started/install/migrating-2.0.md
@@ -48,6 +48,8 @@ const version: Output<string> = pulumi.output(gcp.container.getEngineVersions())
 
 {{% choosable language python %}}
 
+Optionally, you can use an `async` function [entrypoint]({{< relref "/docs/intro/languages/javascript#optional-async-entrypoint" >}}) to more easily write `async/await` code throughout the rest of your program.
+
 ### Update Dependencies
 
 Modify your `requirements.txt` file to update the Pulumi SDK and related providers as below:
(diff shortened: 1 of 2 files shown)


content/docs/intro/languages/javascript.md
```diff
@@ -31,6 +31,36 @@ $ pulumi new javascript
 
 This will create a `Pulumi.yaml` [project file]({{< relref "../concepts/project" >}}), a `package.json` file for dependencies, and an `index.js` file, containing your program. The name of the directory is used as the project name in `Pulumi.yaml`.
 
+## Optional `async` Entrypoint
```
lukehoban, line 34: I think if we are going to document this here - we should probably broaden to describe the entrypoint more generally.

* The entrypoint file is specified in your package.json (`index.ts` by default)
* The entrypoint file is loaded as a Node.js module by Pulumi  (`require("index.ts")`)
* That module can do either:
   * Return a module object with properties for each stack output (`export const out = myResource.output`)
   * Set it's export to an async function whose return value is an object with members for each stack output (`export = async () => { }`)
* Most Pulumi programs use the first option, but programs that need to do `async` work at top level may find they want to use the second.

content/docs/get-started/install/migrating-2.0.md
```diff
@@ -48,6 +48,8 @@ const version: Output<string> = pulumi.output(gcp.container.getEngineVersions())
 
 {{% choosable language python %}}
 
+Optionally, you can use an `async` function [entrypoint]({{< relref "/docs/intro/languages/javascript#optional-async-entrypoint" >}}) to more easily write `async/await` code throughout the rest of your program.
```
cnunciato, line 51: This may be in the wrong place. Does it belong in the JavaScript section?

content/docs/intro/languages/javascript.md
leezen: Makes sense -- will do

content/docs/get-started/install/migrating-2.0.md
leezen: Yeah... not sure how my eyes saw that as within the JS/TS block. Thanks for catching that!

Issue #3065, opened 2020-04-22
Add social sharing buttons to blog


joeduffy:
To encourage sharing blog articles on social media, let's add Twitter and LinkedIn sharing buttons. We should optimize placement for clickability (top and bottom?)

You can see I do this here: http://joeduffyblog.com/2019/09/05/journey-to-pulumi-1-0/ -- scroll to the bottom:

![image](https://user-images.githubusercontent.com/3953235/79998046-106c7100-846f-11ea-8bb2-1e43e294580d.png)

The source code is [here](https://github.com/joeduffy/joeduffy.github.io/blob/master/_layouts/post.html#L20) -- Twitter makes this very easy to do.

PR #3079, opened 2020-04-23
Add a "share this post" section to blogs

joeduffy:
This change adds a Twitter and LinkedIn sharing button to the right
hand nav for blogs. It's sticky so we don't need to choose between
top/bottom and I tried to keep the font sizes small and colors subdued
so as not to distract from the actual post content itself.

Fixes pulumi/docs#3065.

### DIFF
--- a/layouts/partials/blog/right-nav.html
+++ b/layouts/partials/blog/right-nav.html
@@ -6,12 +6,12 @@
                     class="mr-2"
                     href="https://twitter.com/intent/tweet?text={{ $.Page.Title }}&tw_p=tweetbutton&url={{ $.Page.Permalink }}"
                     target="_blank">
-                <i class="fab fa-twitter text-gray-500 hover:text-blue-700 text-xl"></i>
+                <i class="fab fa-twitter text-blue-500 hover:text-blue-700 text-xl"></i>
             </a>
             <a data-track="share-linkedin"
                     href="https://www.linkedin.com/sharing/share-offsite/?url={{ $.Page.Permalink }}"
                     target="_blank">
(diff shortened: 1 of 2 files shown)


joeduffy:
Example screenshot:

image

Here is an alternative darker color so it's more prominent -- felt a little heavy handed, but definitely more visible -- do folks prefer this?

image

layouts/partials/blog/right-nav.html
```diff
@@ -0,0 +1,17 @@
+{{ with .File }}
+    <p class="text-gray-500 text-xs">Share this post</p>
+    <ul class="p-0 list-none">
+        <li>
+            <a data-track="share-twitter"
+                    href="https://twitter.com/intent/tweet?text={{ $.Page.Title }}&tw_p=tweetbutton&url={{ $.Page.Permalink }}"
```
cnunciato, line 6: Just to add a bit of breathing room between the icons:

```suggestion
                    class="mr-2"
                    href="https://twitter.com/intent/tweet?text={{ $.Page.Title }}&tw_p=tweetbutton&url={{ $.Page.Permalink }}"
```

joeduffy:
@cnunciato one more color possibility. I think I like this best -- thoughts?

image

cnunciato:
@joeduffy I like that too, yeah. Is there a little hover on it as well (doesn't look like the code's been updated yet)? Like if that's a -500, maybe a -600 on hover? Otherwise yes, a little color there is nicer, I agree.

joeduffy:
Yeah, exactly, great -- will go with that! Thanks.

PR #3094, opened 2020-04-24
Make a CSS selector more specific

### LINKED ISSUE #3091: CSS Issue selecting property names
Looks like we have a CSS bug where property names are getting selected within the text.

image

[example doc](https://www.pulumi.com/docs/reference/pkg/aws/emr/cluster/)

[Slack convo](https://pulumi.slack.com/archives/C85BS3LJZ/p1587748027051100)

cnunciato:
We want this to match within `property-message` elements, but leave others alone.

Fixes #3091.

### DIFF
--- a/assets/sass/_lists.scss
+++ b/assets/sass/_lists.scss
@@ -72,9 +72,10 @@ dl {
                 ~ dd {
                     .property-message {
                         @apply bg-orange-200 text-orange-600 my-4 px-5 py-4 rounded text-sm;
-                    }
-                    code {
-                        @apply bg-orange-200;
+
+                        code {
+                            @apply bg-orange-200;
+                        }
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


justinvp:
Possibly related: @cnunciato, is it by-design that `<code></code>` blocks in the resource docs don't have a gray background like they do in the rest of the site?

justinvp:
Just as an example:

https://www.pulumi.com/docs/get-started/aws/configure/

Screen Shot 2020-04-24 at 1 05 50 PM

vs

https://www.pulumi.com/docs/reference/pkg/aws/s3/bucket/#inputs

Screen Shot 2020-04-24 at 1 07 56 PM

cnunciato:
@justinvp It was at first (I thought it might make the pages look blotchy with the gray backgrounds of the property names as well), but it actually looks fine IMO:

image

So I'll remove that override as well. Thanks for noticing it!

PR #3103, opened 2020-04-27
Adding GitHub docs to the site

stack72:


stack72:
@praneetloke I am not quite sure how to add resource docs - can you add some information on how I scaffold that section of the site for a new provider?

praneetloke:
@stack72 you shouldn't have to manually build the docs for a new provider release. The current CI setup for `pulumi-github` indicates that it would have run `build-package-docs.sh`, which would have generated resource docs for it with this PR.

But if you are looking to generate them manually on your machine and push them to this PR branch, you can do so by running [`./scripts/gen_resource_docs.sh github true`](https://github.com/pulumi/docs#updating-resource-docs). However, I recommend waiting until https://github.com/pulumi/pulumi/pull/4490 is merged. It adds `github` to the title lookup map, so that the generated resource docs have the title `GitHub` instead of the raw package name `github`.

stack72:
@praneetloke we need to add the docs to the actual site though - even though we build the docs, we don't add the links / sidebar

praneetloke:
They will show up automatically in the left nav as long as the resource docs are available under `content/docs/reference/pkg`. As for listing the link to the GitHub language-specific API docs you would need to do that here: https://github.com/pulumi/docs/blob/master/content/docs/reference/pkg/_index.md. Does that help?

stack72:
I will close this out and open a more specific PR

Issue #2864, opened 2020-04-07
Add a “back to top” link to make long pages more usable.


cnunciato:
Some of our docs pages are extremely long. We should provide a 🔝 link of some sort, perhaps fixed-positioned and visible when scrolled beyond a certain threshold, to help users get around a little more easily.

PR #3152, opened 2020-04-29
Add scroll to top button

sean1588:
Issue: #2864
Adds css and js for scroll to top button.

This PR doesn't make any visual changes. It just adds the needed JS and CSS for the scroll button to work. After this gets in then we can add the scroll to top button with the "top-button" class to the templates we want to add it to. 

I tried to make the button fairly suttle, this is the best I could land on after toying with the css a bit. Please give suggestions for improvement if you have any....
Screen Shot 2020-04-29 at 10 00 14 AM


assets/sass/_top-button.scss
```diff
@@ -0,0 +1,26 @@
...
+    right: 27%;
+    bottom: 40px;
+    opacity: 30%;
+    z-index: 1000001;
+    cursor: pointer;
+    text-align: center;
+}
+
+.top-button::after {
+    content: "🔝";
```
praneetloke, line 19: I am wondering if a simple "chevron" pointing up would be better?

assets/sass/_top-button.scss
```diff
@@ -0,0 +1,26 @@
+.top-button {
```
praneetloke, line 1: I am not a fan of the current style of the button per the screenshot, but I don't have a recommendation either. I am sure Christian would have some ideas, though.

assets/js/topbutton.js
```diff
@@ -0,0 +1,20 @@
+// Scroll to top button
+(function ($) {
+    $(document).ready(function () {
+        $(window).scroll(function () {
+          if ($(this).scrollTop() > 2500) {
+            $('.top-button').fadeIn();
```
praneetloke, line 6: Ha I didn't realize we used jQuery on the site. Could you please use double-quotes for these selector arguments? I see a mix of single/double-quotes in here.

praneetloke: I don't know if there is a plan to get rid of jQuery at some point, but most of what you have here could be done natively with plain JS as well these days including the part where you are [animating](https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollTo) to the top.

sean1588: yea, it could be done with plain js as well. would have prefered to not have needed any JS at all, tried to see if there was a pure CSS way to do this, but I couldn't find one.

assets/sass/_top-button.scss
sean1588: neither am I, tbh.

assets/sass/_top-button.scss
sean1588: Yes, I wanted to use a chevron and tried, but couldn't find one. I also tried using a font-awesome one, but it wouldn't render. It seems it is not available in our current font awesome set.

cnunciato: We do use Font Awesome Free for icons on this site, but there are some arrows in here that might suffice: https://fontawesome.com/icons?d=gallery&q=arrow%20up

assets/js/topbutton.js
cnunciato: We are moving away from jQuery, yeah, to Stencil.js. ([Here's a README](https://github.com/pulumi/docs/tree/master/components) -- it's like React + Angular => Web Components, with TypeScript. It's good stuff. 😄 )

I do think this could be simplified a bit by just using `a href="#"` (boring as that may be); animating the scroll is okay too, but since we don't really have anything else using it yet, and we're not animating the anchor-js clicks either, I'm kind of 🤷‍♂️ on it. 

Placement is tricky too, and hard to gauge without being able to interact with it (the Pulumify preview doesn't seem to be working, so I tried adding a DOM node manually with a `top-button` class, and I can see the box, but without the label, and it's floating mid-column, probably because we're on different-sized monitors):

image

It also gets complicated on pages with right-hand navs that require scrolling (there are many of these, particularly in the SDK docs). I suspect we could maybe tuck something away pinned to the bottom somehow, like absolutely positioned in relation to the right-hand side, with a solid background, or just below that div so that the scrolling weren't impacted. Not quite sure, but I do think your instinct of putting it at the bottom-right makes sense.

It might also be worth considering just making the ON THIS PAGE heading clickable, and/or conditionally showing an up-arrow beside it, since that's always an option, too, and never too far away:

Screen Recording 2020-04-29 at 6 25 19 PM mov

sean1588: I orginally just did an href and an anchor at the top of the page, but the reason I did the jquery was more to control the showing/hiding of the icon depending on how far you scrolled down the page. If not it will be ever present even when you are scrolled to the top which looked kind of silly when I tried it. Not sure how important this is to you. 

So I went ahead and was able to get the chevron icon that Praneet was suggesting to work. Looks better IMO. The reason it wasn;t working before is because I was trying to use it as a pseudo icon in the css by just adding the unicode code to the css content property. Apparantly this only works for some font awesome icons and not all. So in order to use this icon you will need to add the element to the templates or nav-bars as follows:
```<a class="top-button fas fa-chevron-up"></a>
```
This is how it looks now:

Screen Shot 2020-04-30 at 10 51 11 PM

sean1588: Also made "On This Page" clickable to scroll to the top. I think this is an improvement.  Also thanks for the info on Stenciljs. Not that I want to add more dependency on jquery, and would love to convert to something like that, but for timesake I am inclined to just roll with the jquery for now as it is already so heavily ingrained in what we have. Let me know what you think @cnunciato. I orginally just did try an href and an anchor at the top of the page, but the reason I did the jquery was more to control the showing/hiding of the icon depending on how far you scrolled down the page. If not the scroll to top button would be ever-present on the page.

assets/sass/_top-button.scss
```diff
@@ -0,0 +1,17 @@
...
+    border-radius: 3px;
```
cnunciato, line 2: Nearly all of these should be replaceable with Tailwind classes, and [we have a `transition` mixin](https://github.com/pulumi/docs/blob/1afa9b04ca14fec1e7aea10222194ee27bd82567/assets/sass/_mixins.scss#L1-L3) that you can use for that particular bit. We should only write CSS when we can't use Tailwind.

components/src/components/top-button/top-button.tsx
```diff
@@ -0,0 +1,31 @@
...
+@Component({
+  tag: 'pulumi-top-button',
+  styleUrl: 'top-button.scss',
+  shadow: false
+})
+
+export class TopButton {
+
+  @Prop({ mutable: true })
+  scrollPostion: number;
```
cnunciato, line 13:
```suggestion
  scrollPosition: number;
```

components/src/components/top-button/top-button.tsx
```diff
@@ -0,0 +1,31 @@
+import { Component, Listen, Prop, h } from '@stencil/core';
+
+// Scroll to top button.
+@Component({
+  tag: 'pulumi-top-button',
```
cnunciato, line 5: Nit -- four spaces and double-quotes, and should be applied throughout. (This irritates me, too; Stencil generates these stubs with two spaces and single quotes, and there doesn't seem to be a config setting to override that, sadly. We should probably add some linting rules for it, though.)

components/src/components/top-button/top-button.tsx
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  componentWillRender() {
+    this.scrollPostion = window.scrollY;
+  }
+
+  render() {
+    return (
+      <div>
+          <a class="top-button fas fa-chevron-up" title="Scroll to top" style={{display: this.scrollPostion < 2500 ? "none" : ""}}></a>
```
cnunciato, line 27: What tells this to actually scroll to top? (I'd expect to see some kind of a click handler or `href="#"`).

layouts/_default/baseof.html
```diff
@@ -11,6 +11,7 @@
             {{ block "main" . }}
             {{ end }}
         </main>
+        <pulumi-top-button></pulumi-top-button>
```
cnunciato, line 14: I _think_ we only want this to appear in the docs, so this should probably be placed in one those layouts. You might put it in one of the partials; [right-nav](https://github.com/pulumi/docs/blob/master/layouts/partials/docs/right-nav.html) might be an okay place for it.

components/src/components/top-button/top-button.tsx
cnunciato, line 27: Rather than explicitly set a style within the code of the component, you might consider applying a CSS class instead, and put the definition of that class in the Sass along with the other visual things.

components/src/components/top-button/top-button.tsx
```diff
@@ -0,0 +1,31 @@
...
+// Scroll to top button.
+@Component({
+  tag: 'pulumi-top-button',
+  styleUrl: 'top-button.scss',
+  shadow: false
+})
+
+export class TopButton {
+
+  @Prop({ mutable: true })
```
cnunciato, line 12: So this is a Stencil thing, but I think what you want here is [`State`](https://stenciljs.com/docs/state) rather than `Prop`, since `scroll-position` isn't exposed as an API on this component, and in any case is not settable.

components/src/components/top-button/top-button.tsx
cnunciato, line 27:
```
style={{display: this.[...]
```

TIL -- I had no idea Stencil/JSX had types defined at the style-tag level. 🎓

components/src/components/top-button/top-button.tsx
```diff
@@ -0,0 +1,31 @@
+import { Component, Listen, Prop, h } from '@stencil/core';
+
+// Scroll to top button.
+@Component({
+  tag: 'pulumi-top-button',
+  styleUrl: 'top-button.scss',
+  shadow: false
+})
+
```
cnunciato, line 9:
```suggestion
```

components/src/components/top-button/top-button.tsx
sean1588: My approach, for better or worse, has just been pretending stencil is react, until stencil screams at me. lol.

assets/js/main.js
```diff
@@ -49,6 +49,11 @@ function generateOnThisPage() {
         // It's hidden by default. If we added links to the list, show it.
         if (found) {
             $(".on-this-page").show();
+            $(".on-this-page").click(function(){
```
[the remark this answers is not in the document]
sean1588: Would you be ok with just still using jquery for this, but just remove the animation? This portion of the app is 100% generated using jquery. Maybe my jquery skills are just not up to par here, but at least for me I am having a hard time trying to decouple this small piece from using it at this point.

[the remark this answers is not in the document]
cnunciato: Oh absolutely -- I didn't mean to suggest that you should rewrite this whole right-hand nav. 😄 Just removing the animation and replacing with an href seems fine to me.

sean1588:
@cnunciato, not sure if the CSS is exactly how you want now. I tried moving everything to using tailwind but could not get all styles to work, so I converted what I could to using tailwind. I am probably doing something wrong...... This should all be working now, not sure why the preview link is showing something different as far as styling goes, but functionally should be fine.

assets/sass/_top-button.scss
cnunciato: Incidentally, the preview clearly shows that the button is not being styled:


Localhost:

image

Preview:

image

This is almost surely happening because PostCSS is removing the styles you've defined here, as they aren't being referenced in the markup. (We do this intentionally to keep our CSS from becoming bloated -- I believe this has been mentioned before.) [The PostCSS config is here](https://github.com/pulumi/docs/blob/6de64d4e4b0a8ae91c61ac2647472def1b8800c9/assets/config/postcss.config.js). You probably need to add an exception to the whitelist to make sure your styles are retained.

assets/js/main.js
```diff
@@ -49,6 +49,9 @@ function generateOnThisPage() {
...
+                $(window).scrollTop(0); 
+            });
```
cnunciato, line 54: You can literally delete this:

```suggestion
```

.. and make [this line](https://github.com/pulumi/docs/blob/master/layouts/partials/docs/right-nav.html#L4) be this instead:

```
<h4 class="no-anchor">
    <a href="#">On This Page</a>
</h4>
```

There's no need for any script here, and [semantically](https://developer.mozilla.org/en-US/docs/Glossary/Semantics), it should be an `a` tag with an `href` as well.

layouts/_default/baseof.html
```diff
@@ -2,6 +2,7 @@
 <html lang="en-US" prefix="og: http://ogp.me/ns#">
     {{ partial "head.html" . }}
     <body class="section-{{ .Section }}">
+        <div id="top"></div>
```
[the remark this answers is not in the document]
sean1588: OK, thank you. Learned something new. I was not aware of that.

assets/js/main.js
sean1588: >This is almost surely happening because PostCSS is removing the styles you've defined here, as they aren't being referenced in the markup. (We do this intentionally to keep our CSS from becoming bloated -- I believe this has been mentioned before.) 

Yes, it was mentioned before. For whatever reason, I had interpreted that as it having something to do specifically with the way we were generating the preview and that you were just calling it out as to why the preview wasn't working. I didn't realize something needed to be done to like whitelist the styles. I will go ahead and do that.

cnunciato: >For whatever reason, I had interpreted that as it having something to do specifically with the way we were generating the preview

The preview is a prod build, so what you see in preview is what you'll see in prod, yeah. We should treat these as what the world will look like post-merge.

cnunciato: BTW, you can get the same behavior by running `make build`, `cd`ing into `public`, and starting a local HTTP server. So you don't ave to actually push a commit to GitHub to see what'd happen in a preview.

sean1588:
@cnunciato this should be all good to go now! Preview is also showing the correct styling as well! 😄

assets/sass/_buttons.scss
```diff
@@ -83,3 +83,8 @@
         @apply border-gray-700 text-gray-700 bg-transparent;
     }
 }
+
+.btn-scroll-top {
+    @apply text-white text-center cursor-pointer opacity-50 rounded-sm mt-8 w-8 h-4;
+    @include bg-gradient(#90a4ae,#78909c, 30em, 0, 0);
```
cnunciato, line 89: Any reason not to use one of the color-palette colors here? We really should, wherever possible. 😉 🙏

assets/sass/_buttons.scss
```diff
@@ -83,3 +83,8 @@
         @apply border-gray-700 text-gray-700 bg-transparent;
     }
 }
+
+.btn-scroll-top {
+    @apply text-white text-center cursor-pointer opacity-50 rounded-sm mt-8 w-8 h-4;
```
cnunciato, line 88: Is the cursor-pointer needed? I only ask because it's an href now, so it should be a pointer automatically.

sean1588: good point. i'm guessing we will get it for free now.

assets/sass/_buttons.scss
sean1588: Will do! Thanks.

PR #3164, opened 2020-04-29
Allow specifying a package override for pydocgen

praneetloke:
### Proposed changes

As the title states, this PR contains changes to the `generate_python_docs.sh` script to allow passing an optional argument to override the package name. I have also updated `pydocgen` to accept the same optional argument as well and build docs for just that package.

I have also added some missing package names to the following scripts so that they can be run to generate all docs. I kind of went down the rabbit hole with this one:
* `run_typedoc.sh`
* `generate_python_docs.sh`
* `generate_all_resource_docs.sh`
* `docfx/docfx.json`

Along with that I also updated our README with instructions on how to enlist a new provider for docs since that information seemed to be missing.

EDIT: Oh and once this is merged, we'll need to update https://github.com/pulumi/scripts/blob/master/ci/build-package-docs.sh#L77 to pass the provider name.

N/A

### Related issues (optional)

N/A


docfx/docfx.json
```diff
@@ -208,6 +226,15 @@
...
       "dest": "metadata/Pulumi.Mailgun"
     },
+    {
+      "src": [
+        {
+          "src": "../../pulumi-mongodbatlas/sdk/dotnet",
+          "files": ["Pulumi.*.csproj"]
+        }
+      ],
+      "dest": "metadata/Pulumi.Mongodbatlas"
```
praneetloke, line 236: I got this name by looking at our .NET SDK for this provider and checked the namespace for one of the C# classes.

docfx/docfx.json
```diff
@@ -163,6 +163,15 @@
...
       "dest": "metadata/Pulumi.F5BigIP"
     },
+    {
+      "src": [
+        {
+          "src": "../../pulumi-github/sdk/dotnet",
+          "files": ["Pulumi.*.csproj"]
+        }
+      ],
+      "dest": "metadata/Pulumi.Github"
```
praneetloke, line 173: I know this way of spelling GitHub is not the right way, but alas, that's the name we used in our .NET SDK per one of the C# classes that I looked at.

README.md
```diff
@@ -157,6 +157,26 @@ We currently have CI in-place to call the above script whenever a new provider v
...
+Since Pulumi uses CI to generate docs, there isn't a need to keep scripts up-to-date when docs for a new provider
+ need to be generated. Every provider repo uses similar `Makefile`  that contains calls to the
+  relevant scripts to generate docs for that provider when a new version is released.
+
+However, there may be times when you need to generate docs for all providers out-of-band and submit a PR for it. To
+ do that, ensure that the following files have all of the providers listed:
+* For NodeJS API docs: `scripts/run_typedoc.sh`
+* For Python API docs: `scripts/generate_python_docs.sh` and `tools/pydocgen/pulumi-docs.json`
+* For .NET API docs: `docfx/docfx.json`
+* For Resource docs: `scripts/generate_all_resource_docs.sh`
```
justinvp, line 172: Lower priority, but we really should consolidate these into a single place. Should we open an issue?

tools/pydocgen/pulumi-docs.json
```diff
@@ -115,6 +115,12 @@
       "pulumi_provider_name": "keycloak",
       "terraform_provider_name": ""
     },
+    {
+      "name": "Pulumi Kong",
+      "package_name": "pulumi_kong",
+      "pulumi_provider_name": "kong",
+      "terraform_provider_name": "kong"
```
justinvp, line 122: We leave this blank when the TF provider isn't in the https://github.com/terraform-providers GH org.

```suggestion
      "terraform_provider_name": ""
```

README.md
praneetloke: Yes, definitely! I was thinking the exact thing while I was typing this out. But yeah this is the current situation. And yes, I'll open an issue for it.

tools/pydocgen/pulumi-docs.json
praneetloke: Ah ok. I'll update it.

README.md
praneetloke: Created https://github.com/pulumi/docs/issues/3168 for this.

PR #3175, opened 2020-04-30
Regenerate Python docs for all packages

praneetloke:
I have Python 3.8 on my macOS and the version of Sphinx that we have pinned for `tools/pydocgen` does not work unless we upgrade Sphinx to v2.3.1 or later. See https://github.com/sphinx-doc/sphinx/issues/6803. The fix was to disable parallel build in Sphinx on macOS for Python 3.8 https://github.com/sphinx-doc/sphinx/pull/6879. I am not sure if we ever upgraded Sphinx before and if there are things to "watch out" for?

Looking at the diffs, though, I see that the every property, class, method now gets the `py` css class as well. I also see that output properties have a type associated with it. Not sure if that is Sphinx doing that, or if we recently added those types. Those are the two things I noticed. If this is too much, I can switch to Python 3.7 and regenerate them.

Lastly, you might be wondering what's up with the commit messages saying "NodeJS". Sigh. I wrote a little script to commit the files per-package and I wrote it for the other PR for the NodeJS docs, and forgot to change the commit message while generating the Python docs. I'll fix the commit messages when I "squash and merge", so please ignore the "typo" for now. :)


Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
```
justinvp, line 1: Should this file be under `tools/pydocgen`?

Pipfile
```diff
@@ -0,0 +1,11 @@
...
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
+
+[requires]
+python_version = "3.8"
```
justinvp, line 11: Does this line mean 3.8 is required now to generate the docs?

The min version of Python we currently support is 3.6, which means _I think_ we want to continue to build our packages using 3.6 (according to https://docs.travis-ci.com/user/languages/python/#specifying-python-versions, 3.6 is the default version of Python on Travis CI).

If we're now requiring 3.8 to generate the docs, won't we run into issues when releasing a new version of a package, since it'll be using Python 3.6 to regen the docs?

Pipfile
praneetloke: Yikes. This file was accidentally added when I ran `pipenv update sphinx` from the root of the repo instead of the `tools/pydocgen` folder. I have removed this now.

Pipfile
praneetloke: Yeah this whole file was an unintentional addition. I have removed it.

PR #3176, opened 2020-04-30
fix: make typescript test example spec compliant

SimenB:
### Proposed changes

See https://github.com/pulumi/examples/pull/608#discussion_r400767835

### DIFF
--- a/content/docs/guides/testing/unit.md
+++ b/content/docs/guides/testing/unit.md
@@ -331,11 +331,11 @@ pulumi.runtime.setMocks({
 });
 
 describe("Infrastructure", function() {
-    let infra: typeof import('../index')
+    let infra: typeof import('../index');
 
     before(async function() {
         // It's important to import the program _after_ the mocks are defined.
-        infra = await import('../index')
+        infra = await import('../index');
     })
 
     describe("#server", function() {
--- a/content/docs/guides/testing/unit.md
+++ b/content/docs/guides/testing/unit.md
@@ -330,18 +330,22 @@ pulumi.runtime.setMocks({
     // ... mocks as shown above
 });
 
-// It's important to import the program _after_ the mocks are defined.
-import * as infra from "./index";
-
 describe("Infrastructure", function() {
-    const server = infra.server;
+    let infra: typeof import('../index')
+
+    before(async function() {
+        // It's important to import the program _after_ the mocks are defined.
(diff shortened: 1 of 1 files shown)


stack72:
@komalali thoughts? I know you are making some improvements in some areas for this

komalali:
I haven't touched anything specific to this particular case, but the reasoning looks sound to me following the conversation at https://github.com/pulumi/examples/pull/608#discussion_r400767835

@mikhailshilkov was involved in the original convo so tagging him here, but looks good to me and thanks for the improvement @SimenB !

content/docs/guides/testing/unit.md
```diff
@@ -330,18 +330,22 @@ pulumi.runtime.setMocks({
     // ... mocks as shown above
 });
 
-// It's important to import the program _after_ the mocks are defined.
-import * as infra from "./index";
-
 describe("Infrastructure", function() {
-    const server = infra.server;
+    let infra: typeof import('../index')
```
mikhailshilkov, line 334:
```suggestion
    let infra: typeof import('../index');
```

content/docs/guides/testing/unit.md
```diff
@@ -330,18 +330,22 @@ pulumi.runtime.setMocks({
...
+
+    before(async function() {
+        // It's important to import the program _after_ the mocks are defined.
+        infra = await import('../index')
```
mikhailshilkov, line 338:
```suggestion
        infra = await import('../index');
```

content/docs/guides/testing/unit.md
```diff
@@ -330,18 +330,22 @@ pulumi.runtime.setMocks({
...
-    const server = infra.server;
+    let infra: typeof import('../index');
+
+    before(async function() {
+        // It's important to import the program _after_ the mocks are defined.
+        infra = await import('../index');
+    })
+
     describe("#server", function() {
+        const server = infra.server;
```
SimenB, line 342: looking back, this is broken. We cannot destructure inside `describe` as that runs before any hooks (like `before`) and tests. It needs to be accessed either after the `await` or inside a test

SimenB: (at least that's the way Jest and Jasmine works, I don't think I've every used Mocha. But I assume it also works that way)

PR #3187, opened 2020-05-01
Use next page title as the button in Getting Started

joeduffy:
This changes the "NEXT STEP >" button to use the title of the
next page in the flow, like "INSTALL PULUMI >", etc. The theory
is that this is more enticing to click as well as clearly
communicating what comes next.

As part of this, I changed the "CONFIGURE >" step to be a little
clearer; for now, I chose "CONFIGURE CLOUD ACCESS >" for AWS,
Azure, and GCP, and "CONFIGURE CLUSTER ACCESS >" for Kubernetes.

### DIFF
--- a/content/docs/get-started/aws/configure.md
+++ b/content/docs/get-started/aws/configure.md
@@ -1,7 +1,7 @@
 ---
-title: Configure | AWS
-h1: Configure
-linktitle: Configure
+title: Configure Cloud Access | AWS
+h1: Configure Cloud Access
+linktitle: Configure Cloud Access
 meta_desc: This page provides an overview of how to configure an AWS project.
 weight: 4
 menu:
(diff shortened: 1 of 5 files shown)


joeduffy:
For example:

image

joeduffy:
Agree on the awkwardness. I originally used:

> Configure AWS

But it seems odd also -- I mean, you're not actually configuring _AWS_, you're configuring _access_ to AWS. Am I being too pedantic?

cnunciato:
> But it seems odd also -- I mean, you're not actually configuring _AWS_, you're configuring _access_ to AWS.

I agree with you there, too, yeah. I figured since we're already using that phrasing elsewhere, it'd at least be consistent: page reads, "Next, we’ll configure AWS", followed by a button that says "Configure AWS" that when you click it, shows a page whose heading is also "Configure AWS". 

And I guess in a way, you sort of are configuring AWS in that it's the same configuration you'd use to configure the AWS SDK, as explained by the AWS Setup page, which is linked from this page (with a label of "Configure AWS").

joeduffy:
SGTM :+1:

PR #3198, opened 2020-05-01
scott lowe guest blog - redo images/tags

spara:


lblackstone:
The third image in that post is blurry as well

lblackstone:
Revisions LGTM

PR #3200, opened 2020-05-01
DO NOT MERGE: Google Optimize A/B Testing

zchase:
Playground for testing Google Optimize. Do not merge.

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -110,15 +110,12 @@
     {{ end }}
 
     <script async defer src="https://buttons.github.io/buttons.js"></script>
-    {{- if eq hugo.Environment "production" }}
-    <!--Segment tracking-->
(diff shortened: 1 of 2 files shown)


zchase:
@cnunciato Google Optimize can cause a flicker to happen when showing a variant during an experiment. They recommend adding an anti-flicker snippet to the page but it can cause page performance issues. I put it after all of our other scripts because if I understand the documentation correctly the snippet should come after anything we might want to potentially capture within Optimize. What are your thoughts here? 

Docs: https://support.google.com/optimize/answer/7100284
Anti Flicker Script Explanation: https://developers.google.com/optimize/devguides/antiflicker

cnunciato:
Closing this but leaving the branch for reference (please remove if it's no longer needed).

PR #3208, opened 2020-05-05
Update Step 2 C# Example

efleming18:
I tried following this example and ran in to multiple potential issues. 
- The initial tutorial puts all of the code in to the `MyStack.cs` class, and the run is only using that file for all of the updates. The description in Step 2, describes putting the code in to `MyStack.cs` but the code snippet is putting everything in `Program.cs`.
- Even if using `Program.cs`, it would need to be `Deployment.RunAsync`, there's a compiler error with a straight copy/paste
- There's a syntax issue when specifying the `SecurityGroupIngressArgs`, there's an extra set of curly brackets that I've removed
- The line for specifying the `ami.Id` will not work unless you get the `.Result` of the `GetAmi.InvokeAsync`. Without, `ami.Id` just returns `1`. Is there a better way to fix this? I always try to avoid doing `.Result`, but not sure how you want to approach that in this case.
- I added the two `[Output]` properties for `PublicIp` and `PublicDns`
- I also changed the use of `SecurityGroups` which is apparently deprecated to use `VpcSecurityGroupIds`. This appears to have worked like normal, but would definitely like for you to confirm.

Please feel free to take another approach with this, this is just what I had to do in order to get the tutorial working as I believe it was intended to. I'm also happy to make any adjustments to this PR if that would help you out. 

If I did something wrong and the tutorial _is_ working as intended the way it is on the site right now, please let me know so I can adjust for future reference.

Thanks!

### DIFF
--- a/content/docs/tutorials/aws/ec2-webserver.md
+++ b/content/docs/tutorials/aws/ec2-webserver.md
@@ -169,15 +169,16 @@ pulumi.export('publicHostName', server.public_dns)
 
 ```csharp
 using Pulumi;
+using Pulumi.Aws;
+using Pulumi.Aws.Ec2;
+using Pulumi.Aws.Ec2.Inputs;
 using Pulumi.Aws.Inputs;
 
 class MyStack : Stack
 {
     public MyStack()
     {
-        var size = "t2.micro";     // t2.micro is available in the AWS free tier
-
-        var ami = Pulumi.Aws.GetAmi.InvokeAsync(new Pulumi.Aws.GetAmiArgs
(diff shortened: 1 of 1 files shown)


mikhailshilkov:
@efleming18 Thanks for the PR! These tutorials are unfortunately outdated and didn't get any love after the recent changes. I believe we can copy the code from https://github.com/pulumi/examples/blob/master/aws-cs-webserver/WebServerStack.cs - this would avoid using `.Result`, for example. Could you please update your PR to match? Thank you!

efleming18:
I added some updates based on your feedback, @mikhailshilkov - if there's anything else I need to do for it, please let me know.

PR #3224, opened 2020-05-06
Parse lang from anchor and query params

sean1588:
This PR enables language selection from tilde delimited anchor tags and query param specifying `?language=python`. This will enable anchoring to properties on the resource docs pages.

```/docs/reference/pkg/aws/s3/bucket/#region~python```
```/docs/reference/pkg/aws/s3/bucket/?language=python#region~python```

Issue: #2854

### DIFF
--- a/assets/js/chooser.js
+++ b/assets/js/chooser.js
@@ -170,12 +170,12 @@ $(document).on("rendered", function() {
     });
 
     // Parse and set language from anchor tag if present.
-    var url = window.location.href
+    var url = window.location.href;
     var anchorTag = url.includes("#") ? url.substring(url.indexOf("#")+1) : "";
     if (anchorTag) {
         var matches = /[^~]*$/gm.exec(anchorTag);
         if (matches) {
-            var language = matches[0]
+            var language = matches[0];
             selectChoice("language", language);
         }
     }
--- a/assets/js/chooser.js
+++ b/assets/js/chooser.js
@@ -170,8 +170,7 @@ $(document).on("rendered", function() {
     });
 
     // Parse and set language from anchor tag if present.
-    var url = window.location.href;
-    var anchorTag = url.includes("#") ? url.substring(url.indexOf("#")+1) : "";
+    var anchorTag = window.location.hash;
     if (anchorTag) {
         var matches = /[^~]*$/gm.exec(anchorTag);
         if (matches) {
--- a/assets/js/chooser.js
+++ b/assets/js/chooser.js
@@ -169,6 +169,18 @@ $(document).on("rendered", function() {
         }
     });
 
+    // Parse and set language from anchor tag if present.
+    var url = window.location.href
+    var anchorTag = url.includes("#") ? url.substring(url.indexOf("#")+1) : "";
+    if (anchorTag) {
+        var matches = /[^~]*$/gm.exec(anchorTag);
+        if (matches) {
+            var language = matches[0]
+            selectChoice("language", language);
+        }
+    }
+    
+
     // If no language is chosen yet, default to TypeScript.
     hideShowChoices("language", selectLanguage, "typescript");
 
--- a/assets/js/chooser.js
+++ b/assets/js/chooser.js
@@ -169,17 +169,6 @@ $(document).on("rendered", function() {
         }
     });
 
-    // Parse and set language from anchor tag if present.
-    var anchorTag = window.location.hash;
-    if (anchorTag) {
-        var matches = /[^~]*$/gm.exec(anchorTag);
-        if (matches) {
-            var language = matches[0];
-            selectChoice("language", language);
-        }
-    }
-    
-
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: assets/js/chooser.js [hunk 1]
```diff
@@ -169,6 +169,18 @@ $(document).on("rendered", function() {
         }
     });
 
+    // Parse and set language from anchor tag if present.
+    var url = window.location.href;
```
praneetloke, line 173: Could we use [`window.location.hash`](https://developer.mozilla.org/en-US/docs/Web/API/Location/hash) instead here? It would allow you to get rid of the next line where you are extracting the hash fragment from the `href`.

sean1588: Sure. Didn't even know that was a thing. Thanks.

sean1588:
OK. I went and moved everything from using the cookies to using redux after chatting with @cnunciato. Should be good for review.

components/src/components/root/root.tsx
```diff
@@ -15,21 +18,47 @@ export class Root {
...
+        // Map internal methods to actions defined on the store.
+        this.store.mapDispatchToProps(this, { setLanguage });
     }
 
     componentDidRender() {
 
         // Since this element initializes the store, dispatch a DOM event letting
         // listeners know when rendering is complete.
         this.rendered.emit();
+        this.setSelectedLanguage();
```
cnunciato, line 37: Did you try dispatching this before emitting the event? I only ask because that's where I would've put it first myself, since the chooser components are written to respond to that event.

components/src/components/root/root.tsx
```diff
@@ -15,21 +18,47 @@ export class Root {
...
     render() {
         return <div>
         </div>;
     }
+
+    private setSelectedLanguage() {
+
+        // Check if language is specified in the query params and set language if present.
+        const queryParamLanguage = getQueryVariable("language");
+        if(queryParamLanguage){
```
cnunciato, line 49:
```suggestion
        if (queryParamLanguage){
```

sean1588:
> * I think we also discussed translating `?language=nodejs` and `acl~nodejs` to `typescript`. Did you want to do that in this PR as well?

This should be working fine. I did not write any code to explicitly handle this, because if the language is not found it defaults back to typescript anyway. At least it works for me locally. 

> * This doesn't seem to be working for me, but based on the code, it looks like it should: http://content-bucket-b0837eb.s3-website-us-east-1.amazonaws.com/docs/reference/pkg/aws/s3/bucket/index.html?language=nodejs

I can say for certain it works locally when I build. Not sure why it doesnt work with the preview.

> * It'd be nice to handle hashchange events also here, but not a huge deal I guess.

I tried to add an @Listen annotation with a handler, but it is not picking up on it.... maybe I have it configured wrong..... but don't think this is a huge deal personally.

components/src/components/root/root.tsx
sean1588: good point. makes more sense that way.

sean1588:
@cnunciato, @praneetloke the preview should be fine now. Seemed I made a small error when refactoring the query param function that I took from chooser.js to be more es6-y and when I tested it I think I must have had an anchor in the url, so it appeared to have continued working. Anyways... should work now. Then once this is merged, we can go and merge up this anchoring PR in pulumi/pulumi. https://github.com/pulumi/pulumi/pull/4488

sean1588:
> .. which is maybe less elegant than a regex, but a bit easier to read, more explicit about positioning, and adds a check to make sure the passed-in language is also a valid one.
> 
> I'll leave the implementation up to you, but I do think we should make this work with real, valid hashes as well.

Agreed. I think it is definitely clearer if anything. Changed to use the built in js functions instead of the regex. The extra check is also useful. Used your code snippet for the most part. Thanks.

components/src/components/root/root.tsx
```diff
@@ -15,14 +18,26 @@ export class Root {
     @Event()
     rendered: EventEmitter;
 
+    @Listen('hashchange')
+    hashChangeEventHandler(event: CustomEvent) {
```
cnunciato, line 22: Oh, wait -- is this needed?

sean1588: hmmmm..... yeah let me see. Maybe it was left over, when I was trying to get it working. Not sure if it does. Good catch.

sean1588: ok. removed the event handler. will merge up once everything passes.

joeduffy:
@sean1588 @cnunciato quick question: Will this work for _any_ page? I frequently find myself wanting to send a link to a specific page in our docs rendered for a given language, not just the resource docs.

cnunciato:
@joeduffy This is intended to work with any page using the `pulumi-chooser` component, yep.

joeduffy:
Huzzah! Great 🎉

PR #3234, opened 2020-05-07
Make the examples in the resource docs collapsible

### LINKED ISSUE #2858: Make the examples in the resource docs collapsible
For example, here:
https://www.pulumi.com/docs/reference/pkg/aws/s3/bucket/

Expand the first one by default. Ensure this works for multiple languages as well.

cnunciato:
This change adds the ability to expand or collapse the examples we present in the resource docs. The first example in the selected language is expanded by default, and choosing another language maintains the same set of expanded/collapsed examples. 

PR also includes new resource docs across all providers.

Fixes #2858.


praneetloke:
When there are no examples, I see a horizontal line, which I guess is because the component didn't have anything to render. But not sure if that is expected?

cnunciato:
> So perhaps this is intentional to demarcate the examples from the rest of the page?

It was, but since it's not behaving uniformly, we can leave it out. I've removed it for now.

PR #3235, opened 2020-05-08
Add a CrossGuard landing page

joeduffy:


joeduffy:
@clstokes PTAL -- curious what you think.

@zchase PTAL -- especially feedback on placement of CTAs, webinar, etc.

clstokes:
> @clstokes PTAL -- curious what you think.

It's perfect! :shipit: 

> Also, we don't have any inbound links (other than the pricing page). Any thoughts on where to advertise? I thought about putting it in the "Products" drop down, but wasn't sure.

A "Policy as Code" link and text in the Products drop down makes sense to me and maybe a "For Security Leaders" under the "Why Pulumi" drop down?

joeduffy:
Sounds great. Done!

Issue #3225, opened 2020-05-06
Update text on the "start" page of each cloud's Getting Started guide


zchase:
The current text on the first page of each cloud's Getting Started guide does not do a good job of communicating what the user is going to do. We should be more verbose in our copy and explain to the user what they are going to do and what to expect. Some suggestions:

- Tell the user what exactly they are going to do. For example in the AWS flow, explain that they will deploy an S3 bucket and then modify it to have a KMS key.
- Explain if there is any potential cost to the user. 

**Pages:**
- [ ] `/docs/get-started/aws`
- [ ] `/docs/get-started/azure`
- [ ] `/docs/get-started/gcp`
- [ ] `/docs/get-started/kubernetes`

zchase:
Working doc w/ copy: https://docs.google.com/document/d/1TbVGzXvwqZX1Oq2X8isdUwC3qhieDP_adyNnwMRNFhc/edit?usp=sharing

joeduffy:
It seemed like we had universal consensus to switch the tutorials to something more "exciting." Specifically, to serve a website out of the bucket rather than enabling encryption, for AWS, which would then allow us to make the initial text more exciting. Do we plan on doing that?

zchase:
Filed a new issue for that here: https://github.com/pulumi/docs/issues/3243

It's worth an A/B test (and continued testing) so we can figure out exactly what is "best" content to show the user.

PR #3245, opened 2020-05-08
Make the copy on index page of getting started flows more verbose.

zchase:
This PR adds more verbose copy about what the user is going to exactly do in each Getting Started guide.

Fixes: https://github.com/pulumi/docs/issues/3225

### DIFF
--- a/content/docs/get-started/aws/_index.md
+++ b/content/docs/get-started/aws/_index.md
@@ -13,6 +13,14 @@ aliases: ["/docs/quickstart/aws/"]
 
 {{< cloud-intro "AWS" >}}
 
-Let's start with installing Pulumi.
+- Configure a Pulumi Community Edition account
+- Deploy a S3 Bucket
+- Modify the bucket to include a KMS key
+
+Estimated AWS Cost: <strong>Free</strong>
+<br />
+Estimated Time: <strong>10 minutes</strong>
+
+Click the button below to get started with the first step.
 
 {{< get-started-stepper >}}
--- a/content/docs/get-started/aws/_index.md
+++ b/content/docs/get-started/aws/_index.md
@@ -13,9 +13,8 @@ aliases: ["/docs/quickstart/aws/"]
 
 {{< cloud-intro "AWS" >}}
 
-- Configure a Pulumi Community Edition account
 - Deploy a S3 Bucket
-- Modify the bucket to include a KMS key
+- Modify the bucket to protect the bucket's contents with server-side encryption
 
 Estimated AWS Cost: <strong>Free</strong>
 <br />
--- a/content/docs/get-started/azure/_index.md
+++ b/content/docs/get-started/azure/_index.md
@@ -13,7 +13,6 @@ aliases: ["/docs/quickstart/azure/"]
 
 {{< cloud-intro "Azure" >}}
 
-- Configure a Pulumi Community Edition account
 - Deploy an Azure Resource Group and Storage Account
 - Modify the program to export the storage account’s connection string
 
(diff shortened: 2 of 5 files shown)


content/docs/get-started/aws/_index.md
```diff
@@ -13,6 +13,14 @@ aliases: ["/docs/quickstart/aws/"]
 
 {{< cloud-intro "AWS" >}}
 
-Let's start with installing Pulumi.
+- Configure a Pulumi Community Edition account
```
cnunciato, line 16: It wouldn't be one of my goals to create an account. I'd leave this out, as it could easily turn people away who actually might not mind creating an account as part of the process.

content/docs/get-started/aws/_index.md
```diff
@@ -13,6 +13,14 @@ aliases: ["/docs/quickstart/aws/"]
...
+- Deploy a S3 Bucket
+- Modify the bucket to include a KMS key
```
cnunciato, line 18: I didn't know what a KMS key was when I started working here, and I suspect many/most won't, either. If we're going to keep leading with KMS key, we should explain it more generally, like "create a storage bucket on Amazon S3" and  "protect the bucket's contents with server-side encryption".

content/docs/get-started/aws/_index.md
zchase: 👍

content/docs/get-started/aws/_index.md
zchase: 👍

joeduffy:
We are A/B testing this, right?

I think the intro paragraph can be tightened up quite a bit. Instead of:

> Pulumi's infrastructure-as-code SDK helps you create, deploy, and manage AWS containers, serverless functions, and infrastructure using real programming languages. The Pulumi AWS provider packages and CLI help you accomplish all these within minutes.

Maybe the whole thing can just be simplified down to:

> Pulumi lets you declare cloud applications and infrastructure using real programming languages. In this tutorial, you will create a new project, deploy and update an AWS S3 bucket, and then clean up, all in under 10 minutes using the AWS free tier.

A "Learn More" may be useful somewhere in here, but we can A/B test that separately?

zchase:
@joeduffy I was under the opinion that this was something we are just "going to do" but I can easily set it up to A/B testable. 

To @cnunciato point above, we are trying to keep this change very narrow so we can measure the effectiveness of the change. I originally had edited the intro paragraph in the PR but as Christian rightly pointed out, that could skew the results.

For the intro paragraph changes, we can run a separate A/B test to see the effect of those specific changes.

joeduffy:
To be honest, I think the page as-is is _more_ risky than what I suggested! I'd just optimize to make this page as good as it can be, whether we A/B test or not...

cnunciato:
@joeduffy's suggestion for replacing the intro paragraph makes sense to me as well -- it's focused and conveys what we want to convey. 👍

zchase:
@joeduffy @cnunciato I created an A/B test issue to test Joe's proposed intro paragraph here: https://github.com/pulumi/home/issues/827

Currently, the A/B test for combined steps is running and I had to make changes to `/docs/get-started/aws/` for routing and to correctly name the button. So to avoid clashing tests it is in a "proposed" state. I agree that the new paragraph is more than likely "better", but I'd like to run the A/B test to determine what exactly "better" means.

cnunciato:
@zchase Ok. Do you want any additional feedback on this PR, or do you just want to run the test when the time comes?

lukehoban:
@zchase Still working on this?  Or should we close out?

zchase:
@lukehoban Going to close this out. Will revisit this copy once we get the copy for the main Getting Started page done.

PR #3246, opened 2020-05-09
Fix anchor tag language logic.

sean1588:
Fixing my *EXTREME* carelessness. Had changed from `filter` to `find` so no need for index.... Could have sworn this was working.... but obviously not.


sean1588:
@cnunciato, you are so fast. 🥇

PR #3257, opened 2020-05-11
added newline

spara:


cnunciato:
Pulumify doesn't seem happy, but merging this anyway in order to unblock master.

Issue #3256, opened 2020-05-11
Come up with a plan for tutorials


leezen:
Decide if we should regenerate or trim some (and have a plan for the ones we cull) or something else

Two parts
1. Stop the bleeding
2. How do we get to a "good situation"

This issues focuses on (1)

For (2), there's a separate overarching topic covered by another issue

lukehoban:
I've re-generated the tutorials from current `examples` repo in https://github.com/pulumi/docs/pull/3260 which addresses the immediate need here.

I'll open new issues on (a) automating this process and (b) deciding what we want to do about examples/tutorials content overall.

lukehoban:
https://github.com/pulumi/docs/issues/1671 actually already tracks keeping content for tutorials up to date.

PR #3260, opened 2020-05-11
Update tutorials

lukehoban:
Re-run mktutorial.sh to regenerate tutorial content from the Examples repo.

It's not immediately clear to me we *want* to keep all the newly added tutorials here, as the examples are of a varying level of quality/completeness.  We could alternatively only accept the "modifications" here and leave out the new tutorials for now.  But broadly speaking the additions help flesh out multi-language coverage for a number of areas.  I'm inclined to just keep everything here for now, until we decide on a different strategy for tutorials overall.

Part of #3256


cnunciato:
It looks like the JavaScript examples have gone missing. E.g., compare:

http://content-bucket-d9bdbd0.s3-website-us-east-1.amazonaws.com/docs/tutorials/aws/?language=javascript

https://www.pulumi.com/docs/tutorials/aws/?language=javascript

lukehoban:
> It looks like the JavaScript examples have gone missing. E.g., compare:

Yes - I left a comment in the code about that.  I collapsed both JS and TS into a single "TypeScript" header, because we largely don't try to duplicate examples across those two languages, we treat them largely as being a single pool of audience.  (that may be "wrong" in terms of examples - I don't think it is - but it could be - but in terms of what it ends up implying for the turorials here - it just feels a little weird to separate these and then have "there are no JavaScript examples for this cloud" messages, or only show 2 examples when there are actually 20).

https://github.com/pulumi/docs/pull/3260/files#diff-404aabb61dca6f4b7ae30627612587abR347-R348

lukehoban:
> It looks like the JavaScript examples have gone missing. E.g., compare:

Ahh - but now that I see the preview - I see the issue - the headers are still the old headers - looks like I missed updating some part of the widget there...

cnunciato:
@lukehoban Yep, looks like you got 'em. 😄

cnunciato:
BTW,  👍  on "current and consistent but not perfect" is better than "neither current nor consistent." 😄

tools/mktutorial/main.go
```diff
@@ -369,9 +375,12 @@ func emitCloudIndexShortcode(root, cloud string, tutorials []tutorial) error {
 
 	// Render the template using the tutorials data.
 	if err = cloudIndexTemplate.FRender(f, map[string]interface{}{
```
leezen, line 377: Is there any reason this map would be different from the previous map? In which case, can we just declare this once?

tools/mktutorial/main.go
```diff
@@ -186,9 +186,9 @@ func gatherTutorials(root string) ([]tutorial, error) {
 
 		// Great! We have a new tutorial. Append it and let's move on to the next one.
 		tutorials = append(tutorials, tutorial{
-			Name:              name,
-			Title:             title,
-			H1:                h1,
+			Name:  name,
```
leezen, line 189: Nit: whitespace change -- something up with your IDE perhaps?

tools/mktutorial/main.go
```diff
@@ -271,6 +271,9 @@ func makeLangMap(tutorials []tutorial, include []string, exclude []string) (map[
 		"js": nil,
 		"ts": nil,
 		"py": nil,
+		"go": nil,
+		"cs": nil,
+		"fs": nil,
```
leezen, line 276: I don't see what I would expect structurally in the rest of this PR to support F#? For example, `cloud_index.tmpl`  only has a C# block and similarly, we don't include "fs" in our map for rendering the templates. If you're intentionally leaving it out, I don't think there's any need to account for it in this map.

tools/mktutorial/main.go
lukehoban: This is actually correct `gofmt` formatting - not sure why it wasn't formatted originally.

tools/mktutorial/main.go
lukehoban: I can remove the `fs`.  Included here just so that it's easier to add support in the future if we want (this list has all the languages we have examples for - the other lists use a subset of those to present in UI).  But can leave this out.

tools/mktutorial/main.go
lukehoban: I left things the way they already were.  Happy to do some additional refactoring.

lukehoban: Actually - going to leave this as-is.  These are parameters to the templates, which could well intentionally want to diverge in the future.  No particularly great need to DRY this.

PR #3271, opened 2020-05-11
[tools/pydocgen] Update generate_python_docs script to not prefix the core Pulumi SDK

praneetloke:
### Proposed changes

The `generate_python_docs.sh` script allows one to override the repo/package for which we want to build Python docs. As part of that it installs the latest Python package for that Pulumi package from `pip`, however, the script was incorrectly prefix the core `pulumi` package with a `pulumi_`. So this PR fixes that.

The issue was reported to me by @stack72.

N/A

### Related issues (optional)

N/A

### DIFF
--- a/scripts/generate_python_docs.sh
+++ b/scripts/generate_python_docs.sh
@@ -67,7 +67,11 @@ else
     # Install the Python package and run the pydocgen tool for just the provider that
     # was requested.
     echo "Building Python docs for ${REPO_OVERRIDE}..."
-    PACKAGE="pulumi_${REPO_OVERRIDE}"
+    if [ "${REPO_OVERRIDE}" = "pulumi" ]; then
+      PACKAGE="${REPO_OVERRIDE}"
+    else
+      PACKAGE="pulumi_${REPO_OVERRIDE}"
+    fi
     pipenv run pip install "${PACKAGE}"
     pipenv run python -m pydocgen "../../content/docs/reference/pkg" "${PACKAGE}"
 fi
--- a/scripts/generate_python_docs.sh
+++ b/scripts/generate_python_docs.sh
@@ -67,10 +67,9 @@ else
     # Install the Python package and run the pydocgen tool for just the provider that
     # was requested.
     echo "Building Python docs for ${REPO_OVERRIDE}..."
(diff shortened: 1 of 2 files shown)


praneetloke:
@stack72 if you could pull these changes locally and test it out, that'd be great! I tried to regenerate the docs for `pulumi` specifically with these changes, and while it does not fail, I do not see any changes. Not sure if that is correct, though.

tools/pydocgen/pydocgen/__init__.py
```diff
@@ -106,7 +106,8 @@ def read_input(input_file: str, package_override: str = "") -> Input:
         providers = []
         if package_override != "":
             provider = [obj for obj in input_dict.get("providers") if obj["package_name"] == package_override]
-            providers.append(Provider(**provider[0]))
+            if len(provider) > 0:
```
justinvp, line 109: Nit: I believe it's more "pythonic" to check for non-empty like:

```suggestion
            if provider:
```

scripts/generate_python_docs.sh
```diff
@@ -67,7 +67,11 @@ else
     # Install the Python package and run the pydocgen tool for just the provider that
     # was requested.
     echo "Building Python docs for ${REPO_OVERRIDE}..."
-    PACKAGE="pulumi_${REPO_OVERRIDE}"
+    if [ "${REPO_OVERRIDE}" = "pulumi" ]; then
+      PACKAGE="${REPO_OVERRIDE}"
+    else
+      PACKAGE="pulumi_${REPO_OVERRIDE}"
+    fi
```
justinvp, line 74: Nit: Missing extra 2 spaces of indentation inside the if/else. Could also be simplified as the following:

```suggestion
    PACKAGE="pulumi"
    if [ "${REPO_OVERRIDE}" != "pulumi" ]; then
        PACKAGE="pulumi_${REPO_OVERRIDE}"
    fi
```

tools/pydocgen/pydocgen/__init__.py
praneetloke: Ha! I was literally reading a thread about that, but felt that checking for a list's length would be more explicit about what we want this to be. But I don't regularly write Python, so I don't know what is preferred. I can change it per your suggestion if you still think I should?

PR #3273, opened 2020-05-12
Regenerate resource docs for AWS, Azure AD, GCP and Random

praneetloke:
### Proposed changes

This PR replaces #3270.

A handful of provider releases were already in-flight when the PR to restore generation of resource docs as part of provider releases was merged. So this PR contains the docs regenerated for those providers/packages with the latest version.

N/A

### Related issues (optional)

N/A


praneetloke:
So the diffs largely look good to me. Lots of new additions due to Python examples becoming available for resources. In some cases, though, it seems that we have lost an example. For example, [here's](https://github.com/pulumi/docs/pull/3273/files#diff-64f45de83e17e6c85e12b8f87a8aa6ae) one case in AWS. Perhaps a conversion failure?

stack72:
I’m going to investigate some of these failures separately

PR #3316, opened 2020-05-13
architecture as code kubernetes

spara:
fix relref paths


cnunciato:
Would you mind rebasing with master and force-pushing? I'm wondering what's causing that Pulumify build to fail.

cnunciato:
BTW, you may need to actually push a commit that's rebased in order to merge up with master; Pulumify doesn't do this by default.

PR #3339, opened 2020-05-15
Bump pulumify version

justinvp:


justinvp:
> error: --yes must be passed in to proceed when running in non-interactive mode

🤦 Oops, forgot we have to explicitly pass `--yes` now in Pulumi 2.0.

PR to do that in `pulumify`: https://github.com/pulumi/actions-pulumify/pull/16

justinvp:
It looks like there's an edge case bug in pulumify: if pulumify had previously created a stack, but failed to update, subsequent runs of pulumify will attempt to run `pulumi config refresh -f` (since the stack exists), but this will fail because there is no previous deployment.

```
+ PULUMI_STACK_NAME=pulumi/pulumify-pulumi-docs-justin-pulumify-upgrade
+ echo '# PR stack name: pulumi/pulumify-pulumi-docs-justin-pulumify-upgrade'
+ pulumi stack select pulumi/pulumify-pulumi-docs-justin-pulumify-upgrade
# PR stack name: pulumi/pulumify-pulumi-docs-justin-pulumify-upgrade
+ PULUMI_STACK_EXISTS=0
+ '[' 0 -eq 0 ']'
+ pulumi config refresh -f
error: no previous deployment
```

I'm just going to close this PR and open a new one with a different branch to work around the issue.

PR #3342, opened 2020-05-16
Docs for changing secrets provider

### LINKED ISSUE #3309: Add docs for stack migration workflows
As part of delivering https://github.com/pulumi/pulumi/pull/4046 in Pulumi 2.2.0, we should add docs that describe how to use this command to migrate stacks between backends and organizations.

These docs should likely be added to one or both of:
* https://www.pulumi.com/docs/intro/concepts/stack/
* https://www.pulumi.com/docs/intro/concepts/state/#going-back-to-the-pulumi-service-backend

(most likely cross referencing the other as appropriate to avoid duplication).

Note that in particular https://www.pulumi.com/docs/intro/concepts/state/#going-back-to-the-pulumi-service-backend already covers one of these scenarios, though the example should be improved now that `--show-secrets` is supported (which will in practice be needed - what's in the current doc won't work as written for most users).

Ideally at least:
1. From the local backend to Pulumi Service backend
2. From one organization/project in the Pulumi Service backend to another
3. From the default Pulumi Service backend secrets-provider to a custom secrets provider

lukehoban:
This content is unfortunately not "simple", though it is at least now "possible" so feels worth documenting.

Fixes #3309.

### DIFF
--- a/content/docs/intro/concepts/config.md
+++ b/content/docs/intro/concepts/config.md
@@ -430,3 +430,61 @@ For example, this configures a stack to use a HashiCorp Vault transit key named 
 $ pulumi stack init my-stack \
     --secrets-provider="hashivault://payroll"
 ```
+
+### Changing the Secrets Provider for a Stack
+
+> Note: There is not (yet) built-in `pulumi` CLI support for changing the secrets provider on an existing stack, but for cases where this is necessary, it is possible with some manual steps, as outlined in this section. 
+
+> Note: These steps can also be used to change the passphrase that is used for an existing stack.
+
+There are two places where secrets are stored in an encrypted form: (1) the configuration stored in the stack file (`Pulumi.<stackname>.yaml`) and (2) the state stored in your state backend.  To change the secrets provider for a stack, both of these will need to be manually updated.
+
+First, you will need to get copies of the two assets above with plaintext copies of your secret values from your current stack.  This is a sensitive operation, and you should be careful to do this in a safe environment and be sure to remove these files when you are done.
+
+```
+$ pulumi config --show-secrets --json > config.json
+$ pulumi stack export --show-secrets > stack.json
+```
+
+Next, you will need to create a temporary new stack using the desired new secrets provider.  If using the `passphrase` secrets provider, this can also be used to create a new passphrase to use for encryption.
+
+```
+$ pulumi stack init temp --secrets-provider <provider>
+```
+
+If you chose `passphrase` you will see something like this:
+
+```
+$ cat Pulumi.temp.yaml
+encryptionsalt: v1:h8kauiCOXZc=:v1:0Tjy4uHvwgw/b0Hp:gKAqRJ74vyIkc5BeEUlCFXynwRTPFQ==
+```
+
+Else if you chose one of the cloud-provider secrets providers you will see something like this:
+
+```
+$ cat Pulumi.temp.yaml
+secretsprovider: awskms://alias/my-kms-key
+encryptedkey: AQICAHhivZtfYgNwNJDVWDd++88uo2qgy5MJqqzsPfTFv7uzDgF+1kaw9/wSKgGh5QpT/cwtAAAAfjB8BgkqhkiG9w0BBwagbzBtAgEAMGgGCSqGSIb3DQEHATAeBglghkgBZQMEAS4wEQQMc2EiH7Mli5ohzB1aAgEQgDsOszkuuC7wffN3FdtKCNAcRoJq3krWmeA/ucnSMn/F2OYVIDKxDZkjJYZBAjEmpWL49Fcly/n8NMNHhw==
+```
+
+And if you chose the Pulumi Service secrets provider, there will be no stack file.
+
+You will use the values above to replace those in use in your existing stack.  
+
+First, your config.  
+
+1. You will need to remove all encypted secrets from your current config using `pulumi config rm` or by removing them by hand from the `.yaml` file (you saved them to `config.json` above).  
+2. Then copy the lines above (`encyptionsalt` or `secretsprovider`+`encryptedkey`) from the `temp` stack into your primary stack (replacing the existing corresponding lines). 
+3. Save that, and then run `pulumi config set --secret <key> <value>` for each of the secret config values you removed, using the values from `config.json`.  
+
+Next, your state.  
+
+1. Open `stack.json` and find the `secrets_providers` section.  
+2. Based on the output of your `Pulumi.temp.yaml` pick from these options:
+    * For `encyrptionsalt` set `"type": "passphrase",` and replace the contents of the `state` with `"salt": "<value>"`.
+    * For `secretsprovider`+`encryptedkey` set `"type": "cloud"` and replace the contents of the `state` with `"url": "<value>"` and `"encryptedkey": "<value>"` respectively.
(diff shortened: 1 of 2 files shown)


content/docs/intro/concepts/config.md
```diff
@@ -430,3 +430,61 @@ For example, this configures a stack to use a HashiCorp Vault transit key named
...
+
+1. You will need to remove all encypted secrets from your current config using `pulumi config rm` or by removing them by hand from the `.yaml` file (you saved them to `config.json` above).  
+2. Then copy the lines above (`encyptionsalt` or `secretsprovider`+`encryptedkey`) from the `temp` stack into your primary stack (replacing the existing corresponding lines). 
+3. Save that, and then run `pulumi config set --secret <key> <value>` for each of the secret config values you removed, using the values from `config.json`.  
+
+Next, your state.  
+
+1. Open `stack.json` and find the `secrets_providers` section.  
+2. Based on the output of your `Pulumi.temp.yaml` pick from these options:
+    * For `encyrptionsalt` set `"type": "passphrase",` and replace the contents of the `state` with `"salt": "<value>"`.
```
clstokes, line 484: typo

```suggestion
    * For `encryptionsalt` set `"type": "passphrase",` and replace the contents of the `state` with `"salt": "<value>"`.
```

content/docs/intro/concepts/config.md
```diff
@@ -430,3 +430,61 @@ For example, this configures a stack to use a HashiCorp Vault transit key named
...
+```
+
+And if you chose the Pulumi Service secrets provider, there will be no stack file.
+
+You will use the values above to replace those in use in your existing stack.  
+
+First, your config.  
+
+1. You will need to remove all encypted secrets from your current config using `pulumi config rm` or by removing them by hand from the `.yaml` file (you saved them to `config.json` above).  
+2. Then copy the lines above (`encyptionsalt` or `secretsprovider`+`encryptedkey`) from the `temp` stack into your primary stack (replacing the existing corresponding lines).
```
clstokes, line 477: typo 
```suggestion
2. Then copy the lines above (`encryptionsalt` or `secretsprovider`+`encryptedkey`) from the `temp` stack into your primary stack (replacing the existing corresponding lines). 
```

content/docs/intro/concepts/config.md
```diff
@@ -430,3 +430,61 @@ For example, this configures a stack to use a HashiCorp Vault transit key named
...
+
+1. Open `stack.json` and find the `secrets_providers` section.  
+2. Based on the output of your `Pulumi.temp.yaml` pick from these options:
+    * For `encyrptionsalt` set `"type": "passphrase",` and replace the contents of the `state` with `"salt": "<value>"`.
+    * For `secretsprovider`+`encryptedkey` set `"type": "cloud"` and replace the contents of the `state` with `"url": "<value>"` and `"encryptedkey": "<value>"` respectively.
+    * For Pulumi Service secrets-provider, set `"type": "service"` and replace the contents of the `state` with `"url"`, `"owner"`, `"project"` and `"stack"` as appropriate for your existing stack.
+
+Save your `stack.json` with those changes, then run `pulumi stack import < stack.json` to import it into your Pulumi stack. 
+
+You should now be able to run `pulumi preview` and see no proposed changes.  Your configuration secrets and state files are now encyrpted using the new secrets provider.  You can remove the `stack.json` and `config.json` files to ensure the plain text of the secrets is no longer available, as well as remove the temporarary stack (`pulumi stack rm temp`).
```
clstokes, line 490: typo 
```suggestion
You should now be able to run `pulumi preview` and see no proposed changes.  Your configuration secrets and state files are now encrypted using the new secrets provider.  You can remove the `stack.json` and `config.json` files to ensure the plain text of the secrets is no longer available, as well as remove the temporary stack (`pulumi stack rm temp`).
```

content/docs/intro/concepts/config.md
```diff
@@ -430,3 +430,61 @@ For example, this configures a stack to use a HashiCorp Vault transit key named
...
+
+Next, your state.  
+
+1. Open `stack.json` and find the `secrets_providers` section.  
+2. Based on the output of your `Pulumi.temp.yaml` pick from these options:
+    * For `encyrptionsalt` set `"type": "passphrase",` and replace the contents of the `state` with `"salt": "<value>"`.
+    * For `secretsprovider`+`encryptedkey` set `"type": "cloud"` and replace the contents of the `state` with `"url": "<value>"` and `"encryptedkey": "<value>"` respectively.
+    * For Pulumi Service secrets-provider, set `"type": "service"` and replace the contents of the `state` with `"url"`, `"owner"`, `"project"` and `"stack"` as appropriate for your existing stack.
+
+Save your `stack.json` with those changes, then run `pulumi stack import < stack.json` to import it into your Pulumi stack.
```
clstokes, line 488: The `pulumi stack init ...` before getting to this point will change the actively selected stack, so I think we need to do a `pulumi stack select` prior to the `import` to go back to the "real" stack, right?

content/docs/intro/concepts/config.md
```diff
@@ -430,3 +430,61 @@ For example, this configures a stack to use a HashiCorp Vault transit key named
...
+$ cat Pulumi.temp.yaml
+secretsprovider: awskms://alias/my-kms-key
+encryptedkey: AQICAHhivZtfYgNwNJDVWDd++88uo2qgy5MJqqzsPfTFv7uzDgF+1kaw9/wSKgGh5QpT/cwtAAAAfjB8BgkqhkiG9w0BBwagbzBtAgEAMGgGCSqGSIb3DQEHATAeBglghkgBZQMEAS4wEQQMc2EiH7Mli5ohzB1aAgEQgDsOszkuuC7wffN3FdtKCNAcRoJq3krWmeA/ucnSMn/F2OYVIDKxDZkjJYZBAjEmpWL49Fcly/n8NMNHhw==
+```
+
+And if you chose the Pulumi Service secrets provider, there will be no stack file.
+
+You will use the values above to replace those in use in your existing stack.  
+
+First, your config.
```
clstokes, line 474: Roughly here is where I think we need the `pulumi stack select ...` back to the original stack as I mentioned in https://github.com/pulumi/docs/pull/3342/files#r428351515.

content/docs/intro/concepts/config.md
```diff
@@ -430,3 +430,61 @@ For example, this configures a stack to use a HashiCorp Vault transit key named
 $ pulumi stack init my-stack \
     --secrets-provider="hashivault://payroll"
 ```
+
+### Changing the Secrets Provider for a Stack
+
+> Note: There is not (yet) built-in `pulumi` CLI support for changing the secrets provider on an existing stack, but for cases where this is necessary, it is possible with some manual steps, as outlined in this section. 
+
+> Note: These steps can also be used to change the passphrase that is used for an existing stack.
+
```
clstokes, line 439: I think it would be helpful to list the high-level steps up front to make the detailed steps easier to follow along. Adding a section like this for example:

--------------------------

Here are the high-level steps needed to change the secrets provider for a stack with details below:
1. Export stack and config with secrets in plaintext
1. Initialize a temporary _new_ stack with the new encryption provider configuration
1. Modify your existing stack's state file with the configuration values of the new encryption provider
1. Re-import your state file with the updated encryption provider configuration

Issue #3336, opened 2020-05-15
Update RemoteStateReference docs


stack72:
https://www.pulumi.com/docs/guides/adopting/from_terraform/#referencing-terraform-state

These docs should now note that we support C# and Go - as per pulumi-terraform 2.1

PR #3349, opened 2020-05-18
Adding docs for pulumi-terraform Go and DotNet SDKs

stack72:
Fixes: #3336

### DIFF
--- a/content/docs/guides/adopting/from_terraform.md
+++ b/content/docs/guides/adopting/from_terraform.md
@@ -48,12 +48,16 @@ $ pip3 install pulumi_terraform
 {{% /choosable %}}
 {{% choosable language go %}}
 
-> Terraform RemoteStateReference is not yet supported in Go. See <https://github.com/pulumi/pulumi-terraform/issues/518>.
+```bash
+$ go get github.com/pulumi/pulumi-terraform/sdk/v2
+```
 
 {{% /choosable %}}
 {{% choosable language csharp %}}
 
-> Terraform RemoteStateReference is not yet supported in .NET. See <https://github.com/pulumi/pulumi-terraform/issues/516>.
+```bash
+$ dotnet add package Pulumi.Terraform
+```
 
 {{% /choosable %}}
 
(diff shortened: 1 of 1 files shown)


content/docs/guides/adopting/from_terraform.md
```diff
@@ -144,18 +148,92 @@ for i in range(2):
...
+        publicSubnetIds := stateRef.Outputs.ApplyT(func(args interface{}) ([]string, error) {
+            ids := args.(map[string]interface{})["public_subnet_ids"].([]interface{})
+            subnetIds := make([]string, len(ids))
+            for i, v := range ids {
+                subnetIds[i] = v.(string)
+            }
+            return subnetIds, nil
+        }).(pulumi.StringArrayOutput)
+        for x := 0; x < 2; x++ {
+            ii := x
```
lukehoban, line 183: This `ii` is not required given how this code now reads.

content/docs/guides/adopting/from_terraform.md
```diff
@@ -144,18 +148,92 @@ for i in range(2):
...
+            for (int i = 0; i < 2; i++)
+            {
+                var server = new Instance($"instance-{i}", new InstanceArgs
+                {
+                    Ami = "ami-7172b611",
+                    InstanceType = "t2.micro",
+                    SubnetId = subnetIds.GetAt(i),
+                });
+            }
+        });
```
lukehoban, line 234: I thought we now encouraged a style in C# that uses a class inheriting from `Stack` - like https://github.com/pulumi/templates/blob/master/azure-csharp/MyStack.cs#L5.

content/docs/guides/adopting/from_terraform.md
```diff
@@ -252,18 +330,70 @@ network_state = terraform.state.RemoteStateReference('network',
...
+            Token:        pulumi.StringPtrInput(token.(pulumi.StringOutput)),
+            Workspaces: state.WorkspaceStateArgs{
+                Name: pulumi.String(workspace),
+            },
+        })
+        if err != nil {
+            return err
+        }
+
+        ctx.Export("test", state.Outputs)
```
lukehoban, line 361: I don't see the equivalent of this in the other examples?

content/docs/guides/adopting/from_terraform.md
```diff
@@ -252,18 +330,70 @@ network_state = terraform.state.RemoteStateReference('network',
...
+	pulumi.Run(func(ctx *pulumi.Context) error {
+
+        conf := config.New(ctx, "")
+        token := conf.RequireSecret("tfeToken")
+        organization := conf.Require("organization")
+        workspace := conf.Require("workspaceName")
+
+        state, err := state.NewRemoteStateReference(ctx, "remote-backend-state", &state.RemoteBackendStateArgs{
+            Organization: pulumi.String(organization),
+            Token:        pulumi.StringPtrInput(token.(pulumi.StringOutput)),
```
lukehoban, line 352: Doesn't this work?

```suggestion
            Token:        token.(pulumi.StringOutput),
```

FWIW - this *should* be able to be just `token`, but I think there is a bug in the typing of `conf.*Secret`.

lukehoban: > FWIW - this should be able to be just token, but I think there is a bug in the typing of conf.*Secret.

https://github.com/pulumi/pulumi/issues/4664

stack72: You are correct :)

content/docs/guides/adopting/from_terraform.md
stack72: gone!

content/docs/guides/adopting/from_terraform.md
stack72: updated

Issue #3353, opened 2020-05-19
Add Auth0 to the cloud-providers page


stack72:
Auth0 needs to be added to this page

https://www.pulumi.com/docs/intro/cloud-providers/

PR #3354, opened 2020-05-19
Adding Auth0 to the CloudProviders page

stack72:
Also fixes links to reference docs and Go examples

Fixes: #3353


stack72:
> Should there be a link to the Auth0 provider on the Cloud Providers home page?

and that's why we have PRs ;)

PR #3355, opened 2020-05-19
Fix examples, they were using "record" instead of "records".

Sodki:
### Proposed changes

Fix examples related to Azure DNS, they were using `resource` instead of `resources`.


leezen:
Thanks for the PR and pointing out the errors! These examples are actually auto-generated, so we'll need to fix this in the way we generate the docs. I've filed https://github.com/pulumi/tf2pulumi/issues/181 to track that.

PR #3366, opened 2020-05-20
SEO optimization, KW "devspace hosting"

97thFloor:
Main KW: devspace hosting 
Secondary KW: devspace, kubernetes manifest, kubectl cp
Language/topics from semantic analysis: devspace, dev, default, command, build, controller

@cnunciato @zchase I had a difficult time understanding the content on this page, so if something doesn't quite make sense, or you are able to add any more copy featuring the above KWs and topics, feel free to do so.


zchase:
This is an autogenerated file and these changes will be overwritten when we regenerate the file. Closing.

PR #3374, opened 2020-05-21
Migrate HubSpot forms from a partial to stencil

zchase:
This PR migrates our HubSpot forms from being loaded through a partial to being loaded through a stencil.js component. This will help set up the work for a cleaner "On Demand" video experience, as we will be tracking if the user has submitted the form or not. So users should in theory only need to submit an "On Demand" form once and then always be shown the video on subsequent visits to the page.

### DIFF
--- a/components/src/store/actions/form.ts
+++ b/components/src/store/actions/form.ts
@@ -1,0 +1,15 @@
+import { TypeKeys } from "./index";
+
+export interface AddSubmittedFormId {
+    type: TypeKeys.ADD_SUBMITTED_FORM_ID;
+    key: string;
+}
+
+// Add a HubSpot formId to submitted forms list.
+export const addSubmittedFormId = (formId: string) => (dispatch, _getState) => {
+    const action: AddSubmittedFormId = {
+        type: TypeKeys.ADD_SUBMITTED_FORM_ID,
+        key: formId,
+    };
+    dispatch(action);
+};
--- a/components/src/store/actions/index.ts
+++ b/components/src/store/actions/index.ts
@@ -1,10 +1,12 @@
 import { SetLanguage, SetK8sLanguage, SetOS, SetCloud } from "./preferences";
+import { AddSubmittedFormId } from "./form";
 
 export enum TypeKeys {
     SET_LANGUAGE = "SET_LANGUAGE",
     SET_K8S_LANGUAGE = "SET_K8S_LANGUAGE",
     SET_OS = "SET_OS",
-    SET_CLOUD = "SET_CLOUD"
+    SET_CLOUD = "SET_CLOUD",
+    ADD_SUBMITTED_FORM_ID = "ADD_SUBMITTED_FORM_ID",
 }
 
-export type ActionTypes = SetLanguage | SetK8sLanguage | SetOS | SetCloud;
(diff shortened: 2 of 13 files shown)


components/src/store/state.d.ts
```diff
@@ -9,6 +9,11 @@ export interface PreferencesState {
...
 }
 
+export interface FormState {
+    submittedFormIds: string[];
+}
+
 export interface AppState {
-    preferences: PreferencesState
+    preferences: PreferencesState;
+    form: FormState;
```
cnunciato, line 18: What about something at the top level like `history`, with form submissions contained within it?

components/src/store/actions/index.ts
```diff
@@ -1,10 +1,12 @@
 import { SetLanguage, SetK8sLanguage, SetOS, SetCloud } from "./preferences";
+import { AddSubmittedFormId } from "./form";
 
 export enum TypeKeys {
     SET_LANGUAGE = "SET_LANGUAGE",
     SET_K8S_LANGUAGE = "SET_K8S_LANGUAGE",
     SET_OS = "SET_OS",
-    SET_CLOUD = "SET_CLOUD"
+    SET_CLOUD = "SET_CLOUD",
+    ADD_SUBMITTED_FORM_ID = "ADD_SUBMITTED_FORM_ID",
```
cnunciato, line 9: This should be a totally separate action, probably one having to do with form submissions specifically.

components/src/store/actions/form.ts
```diff
@@ -0,0 +1,15 @@
+import { TypeKeys } from "./index";
+
+export interface AddSubmittedFormId {
+    type: TypeKeys.ADD_SUBMITTED_FORM_ID;
+    key: string;
+}
+
+// Add a HubSpot formId to submitted forms list.
+export const addSubmittedFormId = (formId: string) => (dispatch, _getState) => {
+    const action: AddSubmittedFormId = {
```
cnunciato, line 10: Or `SubmitHubSpotForm`, maybe? 🤔  We might want to think a bit more on this one and maybe move some logic into the action if it made sense.

components/src/store/reducers/form.ts
```diff
@@ -0,0 +1,29 @@
...
+export const form = (currentState = getInitialState(), action: ActionTypes): FormState => {
+
+    switch (action.type) {
+        case TypeKeys.ADD_SUBMITTED_FORM_ID:
+            const formId = action.key;
+
+            // Check to see if the user has submitted the form before. If not,
+            // add the formId.
+            if (currentState.submittedFormIds.indexOf(formId) === -1) {
+                currentState.submittedFormIds.push(action.key);
```
cnunciato, line 22: So it's a subtle but important [Redux thing](https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns), but the aim is to return a _copy_ of the state slice, rather than mutate it. So instead, what you'd want to do here is return a new object containing a clone of the current set of IDs and the newly submitted ID, de-duped, like

```
const submittedFormIds = currentState.submittedFormIds
    .concat(action.key)
    .reduce((ids, id) => ids.includes(id) ? ids : [...ids, id], []);

// Return a modified copy of currentState.
return Object.assign({}, { submittedFormIds });
```

(The `default` case just acts as a no-op.)

cnunciato:
@zchase Can we close this? I think it's been superseded -- is that right?

Issue #2855, opened 2020-04-07
Add tooltips to resource docs


cnunciato:
- [x] Create a new `pulumi-tooltip` component
- [x] Use the tooltip for the "required" asterisks in the resource docs
- [ ] Use the tooltip on list pages (e.g., /pkg/aws) for individual modules, if possible

joeduffy:
FWIW, I recently made a CSS style derived from the code copy button, for use on the slightly revised pricing page. Not sure if that's helpful here: 
![image](https://user-images.githubusercontent.com/3953235/78618586-db85da80-782f-11ea-9317-ceaa297c5907.png)

cnunciato:
Absolutely is! I noticed that as well and was planning have a look at it. Thanks!

cnunciato:
Little update on this: It looks like this is being complicated by the way Hugo and Autoprefixer are conspiring to do their respective minifications, so it's a bit more painful than it should be. 😕 Working on it, though.

cnunciato:
Looking more closely at the data we have for the modules (which is pretty much just the module name), there isn't much value to be added with tooltips yet, and it'd also be rather annoying if they were dropped into the lists we have today:

![tooltip-mania mov](https://user-images.githubusercontent.com/274700/83310521-26352a80-a1c1-11ea-87a6-3a4d6e6c4d6e.gif)

So I'm going to forgo checking that last box for now. We've talked about rethinking these lists pages a bit, so when we do that, we can revisit whether tooltips might add value in that context.

PR #3391, opened 2020-05-27
Add a tooltip component

cnunciato:
This change adds a tooltip component that shows a little bubble when you hover over an element. In this PR, it's being used to replace the tooltips used on the Pricing page and for copy buttons.

Usage:

```
<pulumi-tooltip>
    <i class="fas fa-question-circle"></i>
    <span slot="content">
        You hovered over the thing!
    </span>
</pulumi-tooltip>
```

tooltips mov

This is the first part of #2855. (The component needs to be added before the docs that use it are incorporated.)


sean1588:
Nice. I like this!

PR #3393, opened 2020-05-27
[tools/resourcedocsgen] Pass the version to resourcedocsgen tool

### LINKED ISSUE #3378: Skip regenerating the schema for providers during resource docs generation
Now that we are starting to check-in the schema to each provider repo, we can simply reuse that schema for docs generation, and still keep the current code to generate schema but IFF the schema is not found (as a fallback.)

This is in the `scripts/gen_resource_docs.sh` script.

### LINKED ISSUE #3386: [PANIC] Building resource docs is broken
Due to a change introduced in all providers as part of v2.3.3 of the bridge, the schema.json doesn't have the version embedded in it. The version gets passed in before the schema is used in provider generation

https://github.com/pulumi/pulumi-aws/blob/master/Makefile#L34

Notice we are passing `VERSION` are the environment variable. The resource docs are unaware of this change - my apologies for this! - and  I believe this is the cause of a panic:

```
~/code/go/src/github.com/pulumi/docs/tools/resourcedocsgen ~/code/go/src/github.com/pulumi/docs
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x17bec9b]

goroutine 1 [running]:
github.com/pulumi/pulumi/pkg/v2/codegen/docs.(*modContext).genIndex(0xc0000bdc00, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/stack72/code/go/src/github.com/pulumi/pulumi/pkg/codegen/docs/gen.go:1634 +0x95b
github.com/pulumi/pulumi/pkg/v2/codegen/docs.(*modContext).gen(0xc0000bdc00, 0xc00090b1a0, 0xc0007bf8a0, 0xc00039e7e0)
	/Users/stack72/code/go/src/github.com/pulumi/pulumi/pkg/codegen/docs/gen.go:1474 +0x892
github.com/pulumi/pulumi/pkg/v2/codegen/docs.GeneratePackage(0x19b8702, 0x15, 0xc0000d8480, 0x0, 0x0, 0x0)
	/Users/stack72/code/go/src/github.com/pulumi/pulumi/pkg/codegen/docs/gen.go:1802 +0x3f1
main.generateDocsFromSchema(0x7ffeefbff6c9, 0x56, 0xc000470000, 0x1820000, 0xc000470000)
	/Users/stack72/code/go/src/github.com/pulumi/docs/tools/resourcedocsgen/main.go:189 +0x15e
main.main()
	/Users/stack72/code/go/src/github.com/pulumi/docs/tools/resourcedocsgen/main.go:77 +0x3b1
exit status 2
```

I believe the `gen_resource_docs.sh` needs to change to pass the VERSION parameter to the `go run` command so that it pushes the version back to the schema

The example of how this is handled in the schema can be found in any of the provider generate steps e.g. https://github.com/pulumi/pulumi-aws/blob/master/provider/cmd/pulumi-resource-aws/generate.go

praneetloke:
### Proposed changes

This PR updates the `resourcedocsgen` tool to accept a version param. Since I was touching the `gen_resource_docs.sh`, I have also updated it to check if a `schema.json` file exists before generating the schema, which fixes #3378.

N/A

### Related issues (optional)

Fixes #3386
Fixes #3378

### DIFF
--- a/scripts/gen_resource_docs.sh
+++ b/scripts/gen_resource_docs.sh
@@ -54,11 +54,6 @@ generate_docs() {
         pulumi plugin install resource "${provider}" "${plugin_version}"
     fi
 
-    TFGEN=pulumi-tfgen-${provider}
-    echo "Running pulumi-tfgen-${TFGEN} to generate provider schema..."
-    make generate_schema
-    popd
-
     if [ "$provider" = "kubernetes" ]; then
         SCHEMA_FILE="../../../pulumi-kubernetes/sdk/schema/schema.json"
         OVERLAY_SCHEMA_FILE="./overlays/kubernetes/overlays.json"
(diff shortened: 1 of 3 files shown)


stack72:
Thank you so much @praneetloke!! Worked a charm and way faster when using an existing schema :)

PR #3406, opened 2020-05-27
Add JFrog webinar recording and AWS "Howdy Partner" webinar

zchase:
This PR adds a gated form for the JFrog webinar. The user will be emailed a link to the recording (an unlisted youtube video). This behavior is a placeholder until we rework the UX of on-demand videos on our website.

This PR also adds the external event with AWS on Twitch.

Fixes: https://github.com/pulumi/home/issues/889


zchase:
@cnunciato good catch. I added the flag to block it from the index. Thanks!

PR #3464, opened 2020-05-31
docs: fix policyArn in lambda.md

CanRau:


leezen:
Thanks for pointing this out and fixing it!

CanRau:
Was a pleasure, thanks for Pulumi ❤️

PR #3467, opened 2020-06-01
Allow indentation on unordered lists in markdown

zchase:
Recently a linting error was merged into master breaking all subsequent builds. This PR adds a linting ignores to bypass the linting for the nested list.

### DIFF
--- a/content/docs/intro/cloud-providers/kubernetes/_index.md
+++ b/content/docs/intro/cloud-providers/kubernetes/_index.md
@@ -56,12 +56,14 @@ The following SDKs are available to work with IaaS resources, and managed or sel
 
 The [`pulumi/kubernetes`](https://github.com/pulumi/pulumi-kubernetes) SDK is available to work with, and deploy app workloads to running Kubernetes clusters:
 
+<!-- markdownlint-disable ul-indent -->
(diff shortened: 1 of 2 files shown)


content/docs/intro/cloud-providers/kubernetes/_index.md
```diff
@@ -56,12 +56,14 @@ The following SDKs are available to work with IaaS resources, and managed or sel
 
 The [`pulumi/kubernetes`](https://github.com/pulumi/pulumi-kubernetes) SDK is available to work with, and deploy app workloads to running Kubernetes clusters:
 
+<!-- markdownlint-disable ul-indent -->
```
cnunciato, line 59: Do we need to impose this linting rule?

cnunciato: (I mean, can we just remove the rule itself, since it seems reasonable to need to be able to nest Markdown list items.)

zchase: @cnunciato I guess not. I don't really have a strong opinion about it. Happy to add it to the global ignores.

cnunciato: Ok cool. Yeah, I'd probably lean toward just allowing that -- again, since there's no real reason not to (that I can think of).

Issue #3389, opened 2020-05-26
Get Resource Docs above API docs in Google search results


lukehoban:
We've recently added resource docs which are *much* better primary reference material than the previous language-specific API docs.  Although we've prominently linked them from TOC, we still see the older (and less effective) docs getting the vast majority of inbound traffic.

We should strive to significantly improve the SEO of the resource docs, and to strongly discourage the language-specific API docs as primary reference material.

A few possible specific items:
- [x] Prominent links from Node.js/Python API docs back to resource docs (https://github.com/pulumi/docs/issues/2885)
- [x] Direct SEO work on the resource docs pages (H1s, etc.)
- [x] Potentially hide from search engines (or indicate as non-canonical?) on the Node.js/Python API docs pages?
- [x] Change `relref`s that point to package docs to refer to the resource docs instead (where it makes sense to do so)

joeduffy:
@lukehoban @leezen @cnunciato I keep getting customer docs problems because a lot of folks are landing on the old API docs. As a first step towards improving this, I would strongly recommend simply adding hyperlinks to the new provider docs at the top in a very prominent way.

The other ideas are good, of course, but I'd really try to land that as quickly as we can while we figure out the more ambitious ideas that might require added design work, fwiw.

lukehoban:
> The other ideas are good, of course, but I'd really try to land that as quickly as we can while we figure out the more ambitious ideas that might require added design work, fwiw.

Absolutely.  I've recorded the list above to be more directly in "priority order".

cnunciato:
On this presently!

PR #3491, opened 2020-06-04
Direct API docs visitors to the resource docs

### LINKED ISSUE #2885: Link back to the resource docs from API docs
Let's find a cheap/simple way to have back links to the resource docs.

Minimally, just go back to the resource namespace.

cnunciato:
This change adds a shortcode and a few additions to the Python and Node.js API doc generators to show an info blurb at the top of API-docs pages directing users to the resource docs. It looks like this, and appears only on those pages whose providers have corresponding resource docs:

image

Fixes #2885.
Part of #3389.


joeduffy:
Hmm, I didn't see this on the AWS pages:

image
image

layouts/shortcodes/api-docs-alert.html
```diff
@@ -0,0 +1,21 @@
...
+    <div class="note note-info">
+        <p>
+            <strong class="mr-1">
+                <i class="fas fa-info-circle mr-1"></i>
+                Building something with {{ .Get 0 }}?
+            </strong>
+            You probably want the <a href="{{ relref . $packagePath }}">resource documentation</a>.
+        </p>
+        <p>
+            This page documents the {{ .Get 0 }} API, but it doesn't go into much detail about how to use
```
leezen, line 15: This feels a bit strong (in that it basically says, "this is not the droid you are looking for," and doesn't feel like it explains why you'd use this reference over the resource docs (or vice versa).

My suggestion would be to remove the "Building something with .... You probably want" and change this entire block to read more like:

"Note: The [resource documentation] describes the way to interact with [resource], including its inputs, outputs, and examples. This page documents the API specification for {{ .Get 0 }}, which provides additional details at the language-level."

cnunciato:
@joeduffy You won't see the new blocks in the docs as a result of this PR -- this PR just introduces the doc-generator changes and shortcode that'll be used when the next round of API docs are generated.

@leezen I'm totally not married to these words, for sure. But my understanding is that we want to be quite clear here that in most cases, these really _aren't_ the docs you probably want to be wading into. Adding the word "specification" makes sense to make that even more clear, but I do think the second paragraph clarifies (or at least tries to clarify) the difference between these docs and the resource docs. So since this isn't meant to be a "just FYI" sort of notice, but more of an actual nudge, I'm tempted to lean more toward keeping the stronger wording than playing it down.

@joeduffy @lukehoban Do y'all have thoughts? Again, I'm totally happy to make whatever changes work best; I'm just going on my current assumption of the intent.

joeduffy:
I do think it's worth making it clear that you almost certainly want the other docs. All of the feedback I get is that someone got lost in the weeds trying to navigate our (frankly not very good) Python docs, for example, and either eventually found the resource pages or -- more often than not -- gave up and assumed we didn't have what they are looking for.

"Resource docs versus API docs" is a bit inside baseball and the average user probably doesn't understand the subtlety. As such, I'm guessing they'd ignore and keep scrolling. Something that grabs your attention seems warranted.

That said, I'm certainly sympathetic to the feedback that there _are_ cases where you want the API docs. I just think it's the vast minority so if we're going to err on one side or the other, I'd suggest using a blunt instrument personally :-)

leezen:
> @leezen I'm totally not married to these words, for sure. But my understanding is that we want to be quite clear here that in most cases, these really _aren't_ the docs you probably want to be wading into. Adding the word "specification" makes sense to make that even more clear, but I do think the second paragraph clarifies (or at least tries to clarify) the difference between these docs and the resource docs. So if it's true that this isn't meant to be a "just FYI" sort of notice, but more of an actual nudge, I'm tempted to lean more toward the more suggestive wording.

I agree that in most cases the API docs aren't what customers are looking for. However, I think having clear wording about what they're reading and what other documentation is available should be enough to let them make that determination vs. trying to make it for them?

I'm not super attached to any phrasing either, but I don't want to say anything along the lines of "you want something else" because then there's no reason for them to be here in the first place when there are very valid reasons for doing so.

On a separate note, can we also incorporate this message at the top of the documentation for godoc and .NET docs? Thinking about godoc could be a useful exercise imho since it's not hosted on the docs site, so there's some sense of separation. I think whatever phrasing we use should apply equally well for the description there around what those docs are vs. using resource docs.

cnunciato:
After chatting a bit with @leezen out of band, I've adjusted the wording a bit to use "package" rather than "API", as that's consistent with what we're using on the API Reference page. Here's what it'll look like in the Node and Python pages:

image

scripts/run_typedoc.sh
```diff
@@ -76,53 +76,7 @@ generate_docs() {
 }
 
 REPOS=(
-    "aiven,pulumi-aiven,sdk/nodejs"
```
cnunciato, line 79: Whoops! Committed this by mistake -- lemme fix up first.

Issue #3504, opened 2020-06-05
Please write an example on how to import a stack using Powershell


Rookian:
[docs/intro/concepts/stack.md](https://www.pulumi.com/docs/intro/concepts/stack/)

`pulumi stack import < stack.json` --> fails when using Powershell.

PR #3508, opened 2020-06-06
Powershell example for stack import

leezen:
Fixes https://github.com/pulumi/docs/issues/3504


content/docs/intro/concepts/stack.md
```diff
@@ -122,12 +122,30 @@ A stack can be exported to see the raw data associated with the stack.  This is
...
+
+```powershell
+> pulumi stack export > stack.json
+
+> Get-Content stack.json | pulumi stack import
+```
+
+{{% /choosable %}}
+
+{{< /chooser >}}
```
lukehoban, line 147: No need to change - but an alternative would be to just move to using `—file` on all platforms (which is possibly more “natural” anyway?)

leezen: I did debate that internally as well. For whatever reason, the idea of piping a file via `stdin` felt 'natural' in that it was the 'reverse' action of piping `stdout` to a file. I think if we're going to show `--file` then it'd be in both directions as well. I'm going to just do that, which removes the need for any choosers, which makes the page less cluttered.

PR #3575, opened 2020-06-11
Add Gartner and IDC awards

cnunciato:
Fixes https://github.com/pulumi/home/issues/928.


ericrudder:
in general, looks good ... sent you a quick note in slack before i saw you tagged me!

PR #3613, opened 2020-06-16
Add a guide for using the Pulumi plugin in Spinnaker

### LINKED ISSUE #2197: Document how to use Pulumi in Spinnaker
We get asked from time to time how to use Pulumi with Spinnaker. This work item is ultimately part of https://github.com/pulumi/docs/issues/598, however it appears sufficiently complex that a dedicated work item makes sense.

In principle, the idea here is simple: we want to execute Pulumi as part of a Spinnaker deployment pipeline, as a dedicated stage, to provision infrastructure. It's already the case that you can provision base infrastructure using Pulumi and then deploy applications to it, as discussed in [this Reddit thread about Spinnaker/Terraform](https://www.reddit.com/r/Terraform/comments/aw0dex/terraform_vs_spinnaker/eizew78/). But actually running Pulumi from Spinnaker can be useful for provisioning any of the myriad of infrastructure types that Spinnaker doesn't support out of the box.

In practice, this seems quite complex. [This video demonstrates effectively what we're trying to accomplish, just in the context of Terraform](https://www.youtube.com/watch?v=FCpsI0nYHEc). This is [documented in even more detail here](https://docs.armory.io/spinnaker/terraform_integration/). Interestingly, the stage itself has to manage cloning and interacting with the source repo; I incorrectly assumed Spinnaker would arrange to make the contents of the necessary Git repos available in advance, as a separate stage.

After installing Minnaker, [per the guide here](https://docs.armory.io/spinnaker/install/#introducing-armory-minnaker) -- by far the easiest way I found to get up and running -- the `hal armory terraform` subcommand is what drives all of this integration:

```
$ hal armory
+ kubectl -n spinnaker exec -it halyard-0 -- hal armory
ARMORY

  Armory specific commands are described here.

USAGE

  hal armory [parameters] [subcommands]

...
SUBCOMMANDS

...
  terraform
    Configure Terraform integration.
```

I've searched high and low, but can't find the source for this on GitHub. My best guess is that Armory keeps their Halyard command, and the Terraform subcommand, closed source. I've tried numerous repos -- https://github.com/armory-io/spinnaker, https://github.com/armory-io/halyard, https://github.com/spinnaker/orca, https://github.com/spinnaker/deck, etc -- and also searched Google for the phrase "Armory specific commands are described here.", but have come up empty.

It _appears_ this is a custom Spinnaker stage referred to as "terraformer" in the docs -- effectively a set of Java subclasses. That process is documented [here]( https://www.spinnaker.io/guides/developer/new-stage/). For instance, [this and surrounding classes seem to be the pipeline stages that deploy Kubernetes manifests]( https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/DeployManifestStage.java).

An alternative, lower effort approach, _might_ be to use a custom script stage, [as described here](https://www.spinnaker.io/reference/pipeline/stages/#script). I am guessing that has the disadvantage of needing more custom work each time it gets set up, plus not having as rich an interface back to the Spinnaker UI to display preview results, etc. Even that seems less than straightforward: it unfortunately requires setting up a Jenkins server that is configured to run the scripts.

This is looking to be a fair bit more work than our standard CI/CD system integration!

praneetloke:
### Proposed changes

This guide walks-through how to use the Pulumi [plugin](https://github.com/pulumi/spinnaker-preconfigured-job-plugin) in Spinnaker.

N/A

### Related issues (optional)

Resolves #2197.


danfhernandez:
Can we add the spinnaker logo to the CD page?

praneetloke:
> Can we add the spinnaker logo to the CD page?

Yep. Good catch! I have added it with the latest commit.

content/docs/guides/continuous-delivery/spinnaker.md
```diff
@@ -0,0 +1,199 @@
...
+# Add the plugin itself
+hal plugins add Pulumi.PreConfiguredJobPlugin --enabled true --version 0.0.1 --extensions pulumi.PreConfiguredJobStage
+
+# Apply the updates
+hal deploy apply
+```
+
+> Depending on your Spinnaker installation you might have to run the above commands using `sudo`.
+
+**NoteE** Before you can access the newly-added plugin in the Deck UI, you must wait for the old `orca` Pod to be replaced with a new one in your cluster. This can take anywhere from a minute to a few depending on the size of your cluster.
```
praneetloke, line 128:
```suggestion
**Note** Before you can access the newly-added plugin in the Deck UI, you must wait for the old `orca` Pod to be replaced with a new one in your cluster. This can take anywhere from a minute to a few depending on the size of your cluster.
```

PR #3614, opened 2020-06-17
Correct the environment variable names for okta

mdaniel:
From personal experience of getting this error:
```
        * missing required configuration key "okta:apiToken": API Token granting privileges to Okta API.
```
and also [as seen in the example code](https://github.com/pulumi/pulumi-okta/blob/v2.1.3/examples/examples_test.go#L13)

### DIFF
--- a/content/docs/intro/cloud-providers/okta/setup.md
+++ b/content/docs/intro/cloud-providers/okta/setup.md
@@ -14,7 +14,7 @@ Once obtained, there are two ways to communicate your configuration tokens to Pu
 
 1. Set the environment variables `OKTA_ORG_NAME`, `OKTA_BASE_URL` and `OKTA_API_TOKEN`:
 
-    ```console
+    ```bash
     $ export OKTA_ORG_NAME=XXXXXX
     $ export OKTA_BASE_URL=YYYYYY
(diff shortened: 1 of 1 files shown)


content/docs/intro/cloud-providers/okta/setup.md
```diff
@@ -12,20 +12,20 @@ The [Pulumi Okta provider]({{< relref "./" >}}) uses the Okta SDK to manage reso
 
 Once obtained, there are two ways to communicate your configuration tokens to Pulumi:
 
-1. Set the environment variables `OKTA_ORG_NAME`, `OKTA_BASE_URL` and `OKTA_API_KEY`:
+1. Set the environment variables `OKTA_ORG_NAME`, `OKTA_BASE_URL` and `OKTA_API_TOKEN`:
 
-    ```bash
+    ```console
```
leezen, line 17: Can we keep these as `bash`? I'm not sure if the styling will recognize `console`. In addition, the copy button won't properly strip out the prompts unless it's a recognized (https://github.com/pulumi/docs/blob/94c6cfcef5a80b857937b261f7b3fb3855e4aa70/assets/js/copybutton.js)

PR #3627, opened 2020-06-20
Adding AzureDevOps to the Cloud Providers page

stack72:
Fixes: #3626

Requires: #3625

### DIFF
--- a/content/docs/intro/cloud-providers/_index.md
+++ b/content/docs/intro/cloud-providers/_index.md
@@ -34,6 +34,7 @@ aliases: ["/docs/reference/clouds/"]
 
 * [Aiven]({{< relref "./aiven" >}})
 * [Auth0]({{< relref "./auth0" >}})
+* [AzureDevOps]({{< relref "./azuredevops" >}})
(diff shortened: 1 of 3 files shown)


content/docs/intro/cloud-providers/_index.md
```diff
@@ -34,6 +34,7 @@ aliases: ["/docs/reference/clouds/"]
 
 * [Aiven]({{< relref "./aiven" >}})
 * [Auth0]({{< relref "./auth0" >}})
+* [AzureDevOps]({{< relref "./azuredevops" >}})
```
lukehoban, line 37: I don't feel strongly - but does this belong in "VCS Providers" section below instead?  I think of this as very similar to GitLab and GitHub providers (though "VCS" is probably not the primary job of any of these three from a Pulumi perspective).

stack72: I don't mind moving this :) It feels more like an infra tool as it does the projects etc as well - let's put it in infra for now and then we can decide to move it if it's better

PR #3654, opened 2020-06-23
Refactor link-checking workflow, notify on failure

### LINKED ISSUE #3231: Fix scheduled builds
### Problem description

1. The scheduled runs almost always fail on the broken link checking step. Examples: [1](https://github.com/pulumi/docs/actions/runs/97672972), [2](https://github.com/pulumi/docs/actions/runs/96675626), [3](https://github.com/pulumi/docs/actions/runs/95750022)
2. Those runs fail (amongst other things) because there are paths we are trying to exclude in `check-links.sh` which are not being successfully excluded (https://github.com/pulumi/docs/blob/master/scripts/check-links.sh#L16-L17)
3. There are also 429s returned from some of the links (github in particular), so we need to do some form of rate-limiting.

Related issue: https://github.com/pulumi/docs/issues/2978

cnunciato:
This change pulls the link-checking task that we run once a day into its own GitHub Actions workflow and updates the logic for that task to use pulumi.com instead of a locally running copy of the website. It also adds a Slack notification that posts to #ops-notifications when that task fails.

Closes #3231.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -1,0 +1,30 @@
+name: Scheduled jobs
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string.
+    # Run every day at 12:00AM UTC.
+    - cron:  '0 0 * * *'
+jobs:
+  buildSite:
+    env:
+      GOPATH: ${{ github.workspace }}
+    name: Install deps and build site
+    runs-on: ubuntu-latest
+    steps:
+      # Turnstyle is used to prevent multiple push jobs from
+      # running at the same time.
+      - name: Turnstyle
+        uses: softprops/turnstyle@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+
+      - name: Check links
+        run: make ci_schedule
+        env:
+            SLACK_ACCESS_TOKEN: ${{ secrets.SLACK_ACCESS_TOKEN }}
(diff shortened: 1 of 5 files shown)


scripts/check-links.sh
```diff
@@ -13,7 +15,7 @@ check_links() {
 	#     - Our Visual Studio Marketplace link for the Azure Pipelines task extension,
 	#       although valid and publicly available, is reported as a broken link.
 	#     - A number of synthetic illustrative links come from our examples/tutorials.
-    ./node_modules/.bin/blc http://localhost:$HTTP_SERVER_PORT --recursive --follow --get \
+    npx blc http://localhost:$HTTP_SERVER_PORT --recursive --follow --get \
```
cnunciato, line 18: Just for consistency with `check_get_pulumi_links`.

scripts/check-links.sh
```diff
@@ -13,7 +14,7 @@ check_links() {
 	#     - Our Visual Studio Marketplace link for the Azure Pipelines task extension,
 	#       although valid and publicly available, is reported as a broken link.
 	#     - A number of synthetic illustrative links come from our examples/tutorials.
-    ./node_modules/.bin/blc http://localhost:$HTTP_SERVER_PORT --recursive --follow --get \
+    npx blc https://www.pulumi.com --recursive --follow --get \
```
cnunciato, line 17: Just for consistency with `check_links_local`.

.github/workflows/check-links.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  buildSite:
+    env:
+      GOPATH: ${{ github.workspace }}
+    name: Install deps and build site
+    runs-on: ubuntu-latest
+    steps:
+      # Turnstyle is used to prevent multiple push jobs from
+      # running at the same time.
+      - name: Turnstyle
```
praneetloke, line 16: You can remove this step here since the trigger for _this_ workflow is only a `cron` schedule.

cnunciato: Ah, that's right! Thanks.

PR #3671, opened 2020-06-25
Add overlay schema file for Docker

praneetloke:
### Proposed changes

* Add a json schema file for documenting Docker `ComponentResource`. Specifically, the `Image` resource.
  * I added the new file in this repo instead of `pulumi-docker` since: a.) this is not the long-term solution, b.) the K8s overlay file is also in this repo alongside the `resourcedocsgen` tool.
* Update the `gen_resource_docs.sh` script to check if there is an overlay json file for a provider rather than hard-coding it.

N/A

### Related issues (optional)

https://github.com/pulumi/pulumi-docker/issues/111


praneetloke:
> Minor thing - I see DockerBuild and CacheFrom don’t get hyperlinked to the sections lower in the doc. Is that expected?

Yeah that is an issue with the resource docs generator. I'll check if there is an issue tracking it and if not, I'll add one for linking union types. It doesn't handle that today.

> Would also be nice to add one or two examples here - do we support that via overlays? If so - might be a good follow up.

Great idea! Yes, the overlay schema is intended to follow the Pulumi schema spec for provider resources. So the examples markdown can be manually authored into the resource's `description` property. I'll update this PR actually.

praneetloke:
@lukehoban the latest preview contains a TS example. Let me know what you think?

PR #3677, opened 2020-06-25
Block external webinar pages from being indexed by Google

zchase:
Currently, our external webinar pages are being indexed by Google. This PR adds the `block_external_search_index` to the webinar archetype and sets the flag to `true` on all the existing external webinars.

cc// @isaac-pulumi

### DIFF
--- a/archetypes/webinar/index.md
+++ b/archetypes/webinar/index.md
@@ -26,8 +26,11 @@ gated: false
 type: webinars
 
 # External webinars will link to an external page instead of a webinar
-# landing/registration page.
+# landing/registration page. If the webinar is external you will need
+# set the 'block_external_search_index' flag to true so Google does not index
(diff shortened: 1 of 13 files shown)


cnunciato:
Just to ask what might sound like a dumb question, but why would we want these _not_ to be indexed?

zchase:
@cnunciato They are rendered as empty pages and the webinar page just links directly to the external link. So there isn't anything valuable on the page itself, and we don't want them to "pop" in search results. @isaac-pulumi found one via Google search.

zchase:
@cnunciato Thanks! Agreed we should take some time to figure out a better approach. I also wouldn't mind wrapping some UX working into the Webinar listing page when we do that. I think there is a lot to improve with how I initially set things up :)

PR #3690, opened 2020-06-26
Added Sashu-Shankar to the team page (toml and jpeg)

sashu-shankar:


sashu-shankar:
Should be updated now

PR #3695, opened 2020-06-29
Fix a broken link in the broken-link checking script

cnunciato:
This change fixes the link to the failed GitHub Actions run and explicitly exits 1 on failure in order to properly fail the run. (Currently, we're exiting 0 because the `echo` statement completes successfully.)


cnunciato:
This PR is failing Pulumify for reasons I can't explain. Going to try closing and reopening, as another PR I just submitted passed without issue. :/

Issue #3713, opened 2020-07-01
Add version to pulumi programming model docs


stack72:
`version` is missing from the programming model docs - it should be in this section:

https://www.pulumi.com/docs/intro/concepts/programming-model/#resources

PR #3714, opened 2020-07-01
Adding docs for ResourceOptions Version

stack72:
Requires https://github.com/pulumi/pulumi/pull/4945
Fixes: #3713

### DIFF
--- a/content/docs/intro/concepts/programming-model.md
+++ b/content/docs/intro/concepts/programming-model.md
@@ -574,6 +574,7 @@ All resource constructors also accept an `options` argument which can provide th
 * [`protect`](#protect): prevent accidental deletion of a resource by marking it protected
 * [`provider`](#provider): pass an [explicitly configured provider](#explicit-provider-configuration), instead of using the default global provider
 * [`transformations`](#transformations): dynamically transform a resource's properties on the fly
+* [`version`](#version): pass a provider plugin version that should be used when operating on a resource.
 
 ###### `additionalSecretOutputs`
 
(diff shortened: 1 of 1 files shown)


content/docs/intro/concepts/programming-model.md
```diff
@@ -573,6 +573,7 @@ All resource constructors also accept an `options` argument which can provide th
 * [`parent`](#parent): establish a parent/child relationship between resources
 * [`protect`](#protect): prevent accidental deletion of a resource by marking it protected
 * [`provider`](#provider): pass an [explicitly configured provider](#explicit-provider-configuration), instead of using the default global provider
+* [`version`](#version): pass a provider plugin version that should be used when operating on a resource.
```
justinvp, line 576: Nit: I get why this is under `provider`, but the rest of the bullets are alphabetical. Should we keep the alphabetical ordering and move this below `transformations` (both here in the bullets and the linked section below)?

content/docs/intro/concepts/programming-model.md
```diff
@@ -1263,6 +1264,51 @@ var vpc = new Aws.Ec2.Vpc("vpc", new Aws.Ec2.VpcArgs(),
 
 {{< /chooser >}}
 
+###### `version`
+
+An optional provider version to use when operating on the the resource. This version overrides the version information inferred from the current package and should rarely be used
```
justinvp, line 1269: Nit:

```suggestion
An optional provider version to use when operating on the the resource. This version overrides the version information inferred from the current package and should rarely be used.
```

content/docs/intro/concepts/programming-model.md
stack72: Great point!

PR #3725, opened 2020-07-02
Build CSS and JS bundles outside of Hugo

### LINKED ISSUE #3716: Fix pulumi/docs builds
Builds are failing again, being killed somehow by GHA, probably for memory-related reasons. Fix it. 🔧

cnunciato:
This change makes a few changes to the site-building and local-development process aimed at bringing down the time it takes to run a successful build, and to make running builds more reliable than it is today. Changes include:

* Pulling CSS and JS bundling out of the Hugo process, and doing them with TypeScript, node-sass, and postcss-cli directly. Not only does this reduce the build burden on Hugo itself, but it lets us start writing our JS code in TypeScript rather than vanilla JS.

* Removing the `content` property from the search index. This also brings down the Hugo build time by reducing the volume of indexable content. There's a reduction in usefulness here as a result, but this seems okay to me given our overall position on the future of our existing search implementation.

* Slightly tidier Makefile, supporting scripts, and a shorter Pulumi build command.
 
This brings CI build times for Hugo itself down from ~8-10 minutes to ~2, and makes local builds run in about 30 seconds. The PostCSS process we run for prod builds (both in prod proper and Pulumify) is still a beast that takes a few minutes to complete, so I'm sure there's additional room for improvement there as well. On the whole, though, I'm hopeful this'll make things a good deal more stable for a while. 🤞 

Fixes #3716.


scripts/build-search-index.js
```diff
@@ -32,7 +31,6 @@ stdin.on("end", () => {
                 this.add({
                     ref: url + "|" + page.title,
                     title: page.title,
-                    content: page.content
```
justinvp, line 35: In some ways this may actually make the search experience better (smaller download, faster results), but in other ways makes it worse (not actually indexing content) as you called out.

I am wondering if we should just burn the current search to the ground and change it so that when you land on `/docs/search/?q=whatever` we grab the query parameter and immediately redirect to Google search? (Note: landing on `/docs/search/` first _before_ redirecting, so we can still get logs of searches people are making). That feels like a cheap fix in the meantime until we get around to building out the actual search experience we want, and will improve both build times for us and search experience for users. And could be done quickly as a follow-up PR.

layouts/partials/head.html
```diff
@@ -131,19 +119,4 @@
         </script>
     {{ end }}
 
-    {{/* Tests to enforce required frontmatter for certain content types. */}}
-    {{ if and (eq .Type "blog") .IsPage }}
```
justinvp, line 135: These checks were measurably affecting the build time?

scripts/build-site.sh
```diff
@@ -0,0 +1,34 @@
...
+make copy_static_prebuilt
+
+printf "Building web components...\n\n"
+make build_components
+
+printf "Running Hugo...\n\n"
+hugo --minify "$@" --templateMetrics -e production
+
+printf "Compiling the JavaScripts...\n\n"
+yarn run tsc --outFile ${JS_BUNDLE}
```
justinvp, line 26: Does this mean we're no longer minifying? Not sure it matters one way or another.

cnunciato: Ah yes, that's right, although that wasn't intentional, although gzipped, the payload is still about the same as today, so I kind of agree it's probably not worth addressing yet.

layouts/partials/head.html
cnunciato: This was actually the first thing I looked at, and no, it didn't make any measurable difference -- I just forgot to put it back in. 😄 Thanks for the catch.

scripts/build-search-index.js
cnunciato: Yep, I'm with you. I stopped short of actually doing that as part of this PR (though I almost did!), as the goal here was more just to "stop the bleeding" and unblock the pipeline. I believe @leezen has some thoughts on what the next steps for search might be as well. I'll make sure to bring this up at the next team meeting.

PR #3729, opened 2020-07-03
Update the README.

pgavlin:
Qualify the publish directory and mark it as inline code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ brew install docfx
 
 `make lint_markdown` will run `yarn lint-markdown` which lints the markdown in the `content` directory.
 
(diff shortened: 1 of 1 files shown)


pgavlin:
The primary intent behind this PR is to kick off a new build of the docs site at a new commit hash.

The last two builds ended up concurrently writing to the docs site's storage, which broke a number of links to assets that are identified by the commit to which they belong.

We believe that this is unlikely to happen again, though more work is necessary to prevent it entirely. The most likely circumstances under which this may occur are if one GH action is waiting on a prior GH action to complete and the latter writes an order of magnitude more data than the former.

Issue #3369, opened 2020-05-21
Build and deploy the website atomically


cnunciato:
Our current website deployment process updates the contents of [a single S3 bucket](https://github.com/pulumi/docs/blob/e22cac7a4ef3852f7b5f5aa49739201c8af4186f/infrastructure/index.ts#L35) with `aws s3 sync`, which can be problematic for various reasons.

Rather than continue updating this one bucket in place, we should move to producing a new bucket on each merge to master, switching the CloudFront origin to point to that new bucket (after testing it first), and keeping some number of previously deployed buckets around for a period of time for easy rollback when necessary.

Part of https://github.com/pulumi/pulumi-service/issues/4957.

pgavlin:
> Rather than continue updating this one bucket in place, we should move to producing a new bucket on each merge to master, switching the CloudFront origin to point to that new bucket (after testing it first), and keeping some number of previously deployed buckets around for a period of time for easy rollback when necessary.

How are we planning to roll this out? If we're still publishing to these bucket asynchronously, we'll need some sort of synchronization to make sure that we always have the latest content.

cnunciato:
@pgavlin Each commit will get either its own bucket or its own top-level folder within a bucket, and then Pulumi will update the CloudFront origin to point to that bucket or bucket path. By itself that doesn't solve the sequencing problem (i.e., making sure commit 1's action completes before commit 2's starts), but I believe we can continue relying on https://github.com/softprops/turnstyle for this. I'll make sure that's the case, though.

PR #3730, opened 2020-07-04
Don't pass the delete option to s3 sync

cnunciato:
By not passing the `--delete` option to `aws s3 sync`, we'll leave deleted files in the S3 bucket, reducing the risk of 404s for files that haven't yet been cached for some users by CloudFront. [See here for additional context](https://pulumi.slack.com/archives/C85BS3LJZ/p1593882723012500?thread_ts=1593818195.480600&cid=C85BS3LJZ).

### DIFF
--- a/infrastructure/sync.sh
+++ b/infrastructure/sync.sh
@@ -48,10 +48,11 @@ done
 
 cd ..
 
-# Synchronize the remaining contents of the local folder and site bucket, deleting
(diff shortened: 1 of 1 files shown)


cnunciato:
Thanks @lukehoban, and definitely agreed. Next up for me is https://github.com/pulumi/docs/issues/3369, which will essentially replace this.

PR #3756, opened 2020-07-10
Couple fixes to the Programming Model docs

cnunciato:
Should be `implements` rather than `extends`.


cnunciato:
Thanks @praneetloke!

PR #3768, opened 2020-07-13
Convert 07/08 Azure Workshop to On Demand and clean up old webinars

zchase:
This PR converts the recent Azure workshop to the On-Demand format. The workshop recording is gated so the user will need to fill out the form, then they will receive an email from GoToWebinar with a link to the recording. This PR also cleans up some older webinars as well as temporarily hides the "Upcoming" header on the "Live" section of the webinar page. There is only one current upcoming webinar so it is a bit weird to keep the header. 

cc// @isaac-pulumi

### DIFF
--- a/layouts/partials/webinar-list-item.html
+++ b/layouts/partials/webinar-list-item.html
@@ -40,8 +40,8 @@
             </div>
         </div>
     </article>
-    <!-- {{ if eq .Params.featured true }}
+    {{ if eq .Params.featured true }}
         {{ $listHeaderText := cond (eq .Params.pre_recorded true) "LATEST" "UPCOMING"}}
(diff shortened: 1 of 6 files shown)


layouts/partials/webinar-list-item.html
```diff
@@ -40,8 +40,8 @@
             </div>
         </div>
     </article>
-    {{ if eq .Params.featured true }}
+    <!-- {{ if eq .Params.featured true }}
```
cnunciato, line 43: ... or just `None at this time.` or the like. Rather than comment the code, could we instead just write some code to determine whether to show the heading, etc.?

zchase: I ended up just removing the `featured` flag from the webinar, so none of the headings show up when there is only one webinar.

PR #3776, opened 2020-07-16
Add a tf2pulumi converter page

joeduffy:


lukehoban:
A few user experience notes:
1. We probably need to disable Go for now - every example I've tried so far has failed.
2. It would be a huge improvement to the feeling of this being compelling to at least colorize the output :-)
3. The majority of TF examples I've found depend on multiple files - frequently just a `variables.tf` or similar - but still multiple files.  We likely at least want to be permissive on `AllowMissingVariables` and `AllowMissingProviders` by default.  But may also want to support `zip` inputs as well as single file inputs.
4. It might make this a lot more immediately engaging to offer a few options for code files to convert (without expecting the user to come up with something themselves) - this would also help ensure that they could see some examples that definitely work before trying their own that might not.
5. There are enough reasons this might fail, that we probably want to dress up the failure experience a little - potentially add a note like "This conversion tool doesn't yet support converted your example code, but we'd be happy to assist you with your evaluation of Pulumi at sales@pulumi.com or in the Pulumi Community Slack".
6. It would be nice if the querystring updated whenever convert was pressed to include a URL for the input in a shareable form (or perhsaps instead if there was just a separate share button that provided the shareable URL).
7. Might make sense to include a note below on "How does this work?" that points to `tf2pulumi` and links to https://www.pulumi.com/docs/intro/vs/terraform/ along with a slight bit of additional contextualization.

joeduffy:
> We probably need to disable Go for now - every example I've tried so far has failed.


Definitely not the case that everything fails. For example take any snippet that actually made its way into our resource examples, and it works; e.g.

```
resource "aws_s3_bucket" "b" {
  bucket = "s3-website-test.hashicorp.com"
  acl    = "public-read"

  cors_rule {
    allowed_headers = ["*"]
    allowed_methods = ["PUT", "POST"]
    allowed_origins = ["https://s3-website-test.hashicorp.com"]
    expose_headers  = ["ETag"]
    max_age_seconds = 3000
  }
}
```

My preference is to keep Go and just soften the error messages the way we discussed (call sales). We want to eagerly advertise that we support Go. And furthermore, I would _hope_ this just gets (much) better with time (soon).

> It would be a huge improvement to the feeling of this being compelling to at least colorize the output :-)
> The majority of TF examples I've found depend on multiple files - frequently just a variables.tf or similar - but still multiple files. We likely at least want to be permissive on AllowMissingVariables and AllowMissingProviders by default. But may also want to support zip inputs as well as single file inputs.

Agree. This is in the TODO comment at the top of the HTML file. My plan is to try to address that entire set over the weekend before calling this ready-to-go on Monday. The "UPLOAD" tab is meant to support the zipfile experience but it'll require some changes to let you download the resulting zip while also visualizing multi-file results.

> It might make this a lot more immediately engaging to offer a few options for code files to convert (without expecting the user to come up with something themselves) - this would also help ensure that they could see some examples that definitely work before trying their own that might not.

Agree, we discussed this idea and I love it! I'll be adding some pre-defined tabs or something like that.

> There are enough reasons this might fail, that we probably want to dress up the failure experience a little - potentially add a note like "This conversion tool doesn't yet support converted your example code, but we'd be happy to assist you with your evaluation of Pulumi at sales@pulumi.com or in the Pulumi Community Slack".

Agree, we discussed this one also, and I agree. Haven't had the time to add this but will do so over the weekend.

> It would be nice if the querystring updated whenever convert was pressed to include a URL for the input in a shareable form (or perhsaps instead if there was just a separate share button that provided the shareable URL).
> Might make sense to include a note below on "How does this work?" that points to tf2pulumi and links to https://www.pulumi.com/docs/intro/vs/terraform/ along with a slight bit of additional contextualization.

Great ideas, will incorporate :+1:

joeduffy:
@cnunciato I've done all sorts of questionable things in putting this functionality together. I'd love your 👀 on it if you have a moment. jQuery aesthetics aside, I suspect the part that needs to change is how this interacts with the chooser component (it depends on the layout of its HTML, meaning it is quite likely rather brittle). Any suggestions on better approaches here?

EvanBoyle:
It would be great to generate all languages by default and perhaps preload the results for the sample. Then this also functions as a "what is pulumi" for those with a TF background. 

Might require some sort of input capping to ensure consistent performance though.

content/migrate/tf2pulumi.md
```diff
@@ -0,0 +1,401 @@
...
+        } else {
+            if (!tfCode) {
+                // Populate a simple default example.
+                let comment = "#"; // to suppress Markdown lint errors.
+                tfCode = `${comment} This Terraform sample provisions an AWS EC2 instance running Ubuntu.
+${comment} Choose a language and click CONVERT below -- or try replacing it with your own!
+
+provider "aws" {
+  region = "us-west-2"
+}
```
lukehoban, line 328: Unfortunately this leads to a warning.  We can probably just leave it out (it's part of external configuration).

assets/sass/_chooser.scss
```diff
@@ -116,6 +116,7 @@ $tab-font-size: 14px !default;
 .mdl-tabs__tab.language-tab,
 .mdl-tabs__tab.os-tab,
 .mdl-tabs__tab.cloud-tab,
+.mdl-tabs__tab.input-kind-tab,
```
cnunciato, line 119: Did you find that you needed these? I'd have guessed since you're using choosers, you wouldn't need to bother with the older chooser stuff.

content/migrate/tf2pulumi.md
```diff
@@ -0,0 +1,420 @@
...
+        // If you hit enter in the URL bar, submit.
+        $("#terraform-url").keypress(function (e) {
+            if (e.which == 13) {
+                convertCode();
+                return false;
+            }
+        });
+
+        // Hook up event handlers for the language choosers.
+        $("pulumi-chooser[type='language'] > ul > li > a").each(function (i, e) {
```
cnunciato, line 402: Technically okay, but I'd probably suggest just reading this from localStorage, which we update on every select change:

```
JSON.parse(localStorage.getItem("pulumi_state")).preferences.language
```

or similar should work.

assets/sass/_chooser.scss
joeduffy: I didn't. I wasn't sure how choosers actually worked under the hood until getting through it all, so I started by copying what we (recently) did with k8s-language. I had a suspicion this wasn't required ... I'll get rid of it.

content/migrate/tf2pulumi.md
joeduffy: I'll give that a try. I still need the selector though, here, since I'm wiring up a new event handler, I think.

joeduffy:
Thanks for all the great feedback, @cnunciato!

>  I think I'd prefer we did this as a standalone app

I hear you and abstractly I agree, however, I wouldn't want to burden you with the time commitment to do this and I unfortunately have exhausted my budget to eek this out the door. If you're okay with it, I'd like to keep it as a single page on the docs website, which allows us to easily reuse the CSS, JS, styling, etc, which might be challenging with an iframe. I think it's worth considering that we are likely to do many more "marketing tools" in the future like this which aren't strictly engineering deliverables, and yet we want them to live alongside the current documentation website, etc.

> There seem to be some line-break issues in FF and Safari. 

Good catch! There was a bogus line-wrap CSS directive. I've removed that and it seems to be well again.

> Currently, the left side seems to want to select-all, which is strange

Agree, I've removed all of the weird auto-selection and clearing of fields. I did some of this because it wasn't clear I'd be able to detect the selected input-kind until I got further along. All cleaned up now -- hopefully this works better now.

> it might be nice to provide some sort of a drop-down to choose from a list of common ones

Agree! I wanted to do this but wasn't sure if I'd have time. I'll try to get this done.

> FWIW, I did dig some up, but when I tried it, it failed

Odd! That same example works for me. What error did you get?

> it might be nice if it handled common keypresses like tabs and such, to prevent jarring behavior like focus blurs

Good suggestion. I did this for tabs in the textarea. If there are others that come to mind, let me know.

> It might be good to express that this only works with AWS

It works for any cloud :smile: What made you think it's AWS-only? Did, for instance, Azure examples fail?

> It's not clear how to actually "send" something I've typed.

Yeah I hadn't implemented any of that yet. I did a fairly major rearrangement of the UI to have everything side-by-side based on @justinvp's feedback, and to try to make it work more like the TypeScript playground.

I've fixed this for now and curious what you think. It's a bit "clever" in that it's based on auto-submissions but it seems to work well in practice. The major rough edge is the errors are pretty jarring if you mistype something, but I assume most people will be copy-and-pasting existing code, not open-coding brand new HCL in the textarea, so hopefully this is fine in practice.

> I'd probably just add a tooltip to that icon and lose the Download Results button

I actually thought the button was a clearer CTA because it's very visible and "action blue," versus the other download icon which is pretty subtle and easy to miss. I renamed it from "Download Results" to "Download Code" -- does that help?

PR #3799, opened 2020-07-21
Deploy the website atomically

### LINKED ISSUE #3369: Build and deploy the website atomically

cnunciato:
This PR changes the way we build and deploy the website from using a single S3 bucket for all website content to using a either a bucket per pull request (for PRs) or per individual commit (for pushes -- e.g., direct commits to master).

There's a lot of Bash here, but I've tried to comment informatively, and this flowchart should hopefully help to explain how this works as well, as of now. On the left is what happens on GitHub `pull_request` events, and on the right, `push` events.

Deployment

A few things to note:

- In order to keep the PR churn low, I've left out removing the parts of the Pulumi program we no longer need. I'll be doing that in multiple follow-ups.
- The tests we're running now include both basic HTTP tests as well as browser tests that will validate our CSS and JS with [Cypress](https://cypress.io). 
- This PR removes Pulumify and replaces it with straight-up S3 sync. The indirection was adding complexity that we no longer need, and there's also a benefit to running the same code to produce both previews and production builds. Only downside is the reintroduction of the latency of pushing content from GH builders to S3, which is a bit of a bummer, but comparing the clock time of Pulumify builds to those produced with this change, the difference only seems to be on the order of a few minutes. Our plans to remove remove/relocate our language-specific SDK docs should help out here, though,  by reducing the ultimate size of the S3 payload almost by half.
- While PRs are slowish, merges should take only a few minutes, since all we're doing is updating the CloudFront origin for an already-built-and-tested bucket.
- We aren't doing any cleanup of old buckets yet. An additional scheduled job would probably be best for this.
- Right now, PRs are building drafts and futures, so on merge, since we're just carrying those buckets across as-is, draft and future-dated content will (at least as of now) be released publicly as well. So a bit of a tweak should probably be made to handle this, too. In practice, we don't use these features much anyway (in fact they sometimes cause confusion), but when we start wanting to schedule future builds, we'll definitely need to suppress drafts and post-dated content for production releases.
- This also adds the concept of "dev stacks" (via pulumi-dev.io) for those who want to hack on website infra. 😉 

Fixes #3369.


.github/workflows/hugo-build.yml
```diff
@@ -6,10 +6,6 @@ on:
   pull_request:
     branches:
       - master
-  schedule:
```
cnunciato, line 9: This has been replaced by [scheduled-daily](https://github.com/pulumi/docs/blob/master/.github/workflows/sheduled-daily.yml).

scripts/translate-redirects.js
```diff
@@ -0,0 +1,82 @@
+// Copyright 2016-2020, Pulumi Corporation.  All rights reserved.
```
cnunciato, line 1: There's not really any new code here -- I've just copied this coe from the Pulumi program in order to be able to run it as an individual script. (As mentioned in the PR description, I'll remove all dead code in a follow-up PR.)

infrastructure/index.ts
```diff
@@ -503,8 +562,11 @@ if (!pulumi.runtime.isDryRun()) {
 const aRecord = createAliasRecord(config.targetDomain, cdn);
 const aliasRecord = createAliasRecord(config.websiteDomain, cdn);
 
-export const archiveBucketUri = archiveBucket.bucket.apply(b => `s3://${b}`);
-export const websiteBucketUri = websiteBucket.bucket.apply(b => `s3://${b}`);
+export const archiveBucketUri = pulumi.interpolate`s3://${archiveBucket.bucket}`;
```
leezen, line 565: I assume the archiveBucket and websiteBucket (and all related code) are part of what you meant by later cleanup?

infrastructure/index.ts
```diff
@@ -26,8 +27,42 @@ const config = {
     certificateArn: stackConfig.require("certificateArn"),
     // redirectDomain is the domain to use for any redirects.
     redirectDomain: stackConfig.get("redirectDomain") || undefined,
+    // originBucketNameOverride is an optional value that can be used to manually pin the
+    // website to a specific S3 bucket. Values are of the form "bucket-name-123", rather
+    // than "s3://bucket-name-123".
+    originBucketNameOverride: stackConfig.get("originBucketNameOverride") || undefined,
```
leezen, line 33: I like that we have this option. It'd be great to have a script/documentation on how to rollback to a previous bucket. It's not clear to me how to easily lookup history of the buckets used for this stack, so having a convenience function that provides that seems useful for safe rollback.

scripts/bucketize.sh
```diff
@@ -0,0 +1,114 @@
...
+
+# The docroot of the built website.
+build_dir="public"
+
+# The text file we'll write as an output result.
+metadata_file="$(origin_bucket_metadata_filepath)"
+
+# Verify we have at least 1000 index.html files in total across the site.
+if [ ! "$(find $build_dir -type f | grep index.html | wc -l)" -ge 1000 ]; then
+    echo "Page-count check failed. See the archive extraction log above for details."
```
leezen, line 28: This error no longer applies outside the context of the archive job, right? The check is still useful, but I think we probably need to update this message.

scripts/common.sh
```diff
@@ -17,6 +17,50 @@ post_to_slack() {
...
+        echo "pr-$(cat "$GITHUB_EVENT_PATH" | jq -r ".number")"
+    else
+        echo "push-${GITHUB_SHA:=$(git_sha_short)}"
+    fi
+}
+
+# Get the AWS SSM Parameter Store key for the specified commit SHA. Used for mapping a
+# commit to a previously created bucket.
+ssm_parameter_key_for_commit() {
+    echo "pulumi-docs-bucket-for-commit-$1"
```
leezen, line 61: I think ideally, we prefix this key to avoid polluting the SSM root keyspace. Maybe something like `/docs/commits/pulumi-docs-bucket-for-commit-$1`? This will complicate some of the later logic when parsing this value, but seems worthwhile to me.

scripts/bucketize.sh
cnunciato: Ah yes, thanks -- the message doesn't make sense in the revised context. I'll update.

scripts/common.sh
cnunciato: Thanks, I appreciate that. I haven't used SSM before, so I wasn't sure whether there was a more idiomatic way to use these keys. I'll make this change.

leezen: For reference: https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-paramstore-su-organize.html

infrastructure/index.ts
cnunciato: Hadn't completely worked out the UX for this either, yeah -- having a convenience function to look up buckets and make it easier to choose one, along with some additional docs, makes sense to me. Will add.

infrastructure/index.ts
cnunciato: Yep, exactly.

infrastructure/index.ts
cnunciato: Added `scripts/list-recent-buckets.sh`, with comments, and which when run outputs the following:

```
$ ./scripts/list-recent-buckets.sh
Found 3 recent pulumi-docs-origin-* buckets:

Fetching metadata for pulumi-docs-origin-pr-3799...

Bucket URL:  http://pulumi-docs-origin-pr-3799.s3-website.us-west-2.amazonaws.com
Bucket Name: pulumi-docs-origin-pr-3799
Synced:      2020-07-29 13:20:04 UTC
Commit:      https://github.com/pulumi/docs/commit/a0135e3b79b677891ad02f94f28e75774c6d66d5

---

Fetching metadata for pulumi-docs-origin-3799...
download failed: s3://pulumi-docs-origin-3799/metadata.json to - An error occurred (404) when calling the HeadObject operation: Not Found
Missing metadata file. This bucket may not have been built and tested successfully.

---

Fetching metadata for pulumi-docs-origin-placeholder-production...
download failed: s3://pulumi-docs-origin-placeholder-production/metadata.json to - An error occurred (404) when calling the HeadObject operation: Not Found
Missing metadata file. This bucket may not have been built and tested successfully.

---

To run browser tests on one of these buckets, run:
nvm use && make ensure && CYPRESS_BASE_URL="<s3-bucket-url>" yarn run cypress run --headless

To pin the website to one of these buckets, run:
pulumi -C infrastructure config set originBucketNameOverride <bucket-name>
pulumi -C infrastrcuture preview
```

infrastructure/index.ts
```diff
@@ -12,6 +12,7 @@ import * as tar from "tar";
 import * as tmp from "tmp";
 
 const stackConfig = new pulumi.Config();
+const awsConfig = new pulumi.Config("aws");
```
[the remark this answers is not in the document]
cnunciato: Ah thanks -- good spot. I had thought about using it earlier on, but ended up now. Will remove.

Makefile
```diff
@@ -66,21 +63,19 @@ copy_static_prebuilt:
...
+lint:
+	yarn run lint-markdown
 
 .PHONY: test
 test:
 	$(MAKE) check_links_local
 
 .PHONY: check_links_local
-check_links_local::
+check_links_local:
```
praneetloke, line 78: I am no `make` expert, but isn't the second `:` required here and the few other places below where they were removed?

cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
```
praneetloke, line 1: Is this file needed?

cypress/fixtures/example.json
```diff
@@ -0,0 +1,2 @@
+{
```
praneetloke, line 1: Is this file needed?

infrastructure/Pulumi.production.yaml
```diff
@@ -1,6 +1,8 @@
 config:
   aws:region: us-west-2
   www.pulumi.com:certificateArn: arn:aws:acm:us-east-1:058607598222:certificate/a3f72a2b-e715-4639-b126-1e4efc0b634b
+  www.pulumi.com:originBucketNameOverride:
```
praneetloke, line 4: Since there is no value defined for this property, can this be removed? It looks like you are handling an `undefined` value for it anyway.

scripts/bucketize.sh
```diff
@@ -0,0 +1,114 @@
...
+# here, since in most cases, we'll be continually updating a bucket associated with a PR;
+# passing this option keeps the destination bucket clean.
+aws s3 sync "$build_dir" "$destination_bucket_uri" --acl public-read --delete --quiet
+
+echo "Sync complete."
+s3_website_url="http://${destination_bucket}.s3-website.${aws_region}.amazonaws.com"
+echo "$s3_website_url"
+
+# Set the content-type of latest-version explicitly. (Otherwise, it'll be set as binary/octet-stream.)
+aws s3 cp "$build_dir/latest-version" "${destination_bucket_uri}/latest-version" \
```
praneetloke, line 64: `latest-version` is the file you referenced earlier in the comments that would be used by Pulumi?

scripts/list-recent-buckets.sh
```diff
@@ -0,0 +1,57 @@
...
+#
+# Note that these buckets are not limited to those whose content has already been merged;
+# this includes everything, including buckets for open PRs and those that failed to build.
+# Commit links should help to identify buckets whose contents have been merged, however,
+# and in general, the presence of a metadata file implies the bucket was built and tested
+# successfully.
+
+source ./scripts/common.sh
+
+buckets="$(aws s3api list-buckets \
```
praneetloke, line 17: Ha TIL the `s3api` command.

scripts/bucketize.sh
```diff
@@ -0,0 +1,114 @@
...
+    --value "$destination_bucket" \
+    --type String \
+    --region $aws_region \
+    --overwrite
+
+# Finally, if it's a preview, post a comment to the PR that directs the user to the resulting bucket URL.
+if [ "$1" == "preview" ]; then
+    pr_comment_api_url="$(cat "$GITHUB_EVENT_PATH" | jq -r ".pull_request._links.comments.href")"
+    post_github_pr_comment \
+        "Your site preview build for for commit $(git_sha_short) is ready! :tada:\n\n${s3_website_url}." \
```
praneetloke, line 110:
```suggestion
        "Your site preview for the commit $(git_sha_short) is ready! :tada:\n\n${s3_website_url}." \
```

cypress/plugins/index.js
```diff
@@ -0,0 +1,21 @@
+/// <reference types="cypress" />
```
praneetloke, line 1: Is this file needed?

scripts/find-or-make-bucket.sh
```diff
@@ -0,0 +1,94 @@
...
+        if [ -z "$candidate_bucket" ]; then
+            echo "Push event detected, but there doesn't seem to be a bucket associated with the referenced commit ${associated_pr_commit}."
+            build_and_sync_bucket
+            exit 0
+        fi
+
+        # Validate that the bucket exists and has a metadata file, then download that file
+        # to the local path that Pulumi will be expecting. We'll use this file to run the
+        # Pulumi update if it passes all of our checks.
+        aws s3 cp s3://${candidate_bucket}/metadata.json $(origin_bucket_metadata_filepath)
```
praneetloke, line 62: What do you think about adding this file as a GH Action [artifact](https://docs.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts)? It could be useful to look at this file when looking at a specific workflow execution to see what it produced. You could download this file as part of a `push` build and avoid the SSM param thing altogether most likely.

Makefile
cnunciato: Definitely not required, no. There does seem to be _some_ use for them, but [even the GNU Make docs](https://www.gnu.org/software/make/manual/html_node/Double_002dColon.html) point out that those uses are quite specific and rare:

> Double-colon rules are somewhat obscure and not often very useful; they provide a mechanism for cases in which the method used to update a target differs depending on which prerequisite files caused the update, and such cases are rare. 

In our case, since we're just using Make targets as task wrappers (as opposed using to compile software binaries, for example), they aren't really doing anything, so I figured I'd remove them.

cypress/support/commands.js
cnunciato: It's not required, it was just generated by the Cypress setup, and is [used for extending Cypress](https://docs.cypress.io/api/cypress-api/custom-commands.html#Syntax). Figured I'd go ahead and leave it, since it's tucked away in the `cypress` folder, and when someone needs it, it'll be there.

cypress/fixtures/example.json
cnunciato: This one can go, yeah. I'll remove.

infrastructure/Pulumi.production.yaml
cnunciato: It _can_ be, yes but I chose to explicitly show it, just to be clear that it's there and available.

scripts/list-recent-buckets.sh
cnunciato: Oh, it's very useful. 😄

scripts/bucketize.sh
cnunciato: No, `latest-version` is the file used by the Pulumi CLI to determine the current released version of Pulumi.

cypress/plugins/index.js
cnunciato: This one was also Cypress-generated, so would prefer to keep for future use.

scripts/find-or-make-bucket.sh
cnunciato: Definitely like the idea of producing it as a built artifact, yeah, for reference. I'll look into adding that. 

I'd still need the ability to look up the bucket associated with a particular commit later, so having it stored solely as an artifact would make that a good deal more cumbersome, and since I want to run the whole job to succeed or fail in one step, rather than break it out into separate steps, there wouldn't be much benefit to producing and then consuming the artifact as part of the workflow itself. So as a replacement for SSM for PRs and pushes, it's not quite the right fit. Less frequently, there'll also be the occasionaly need to produce these files and use them for deployments outside the context of an Actions workflow (e.g., while developing a dev stack, deploying to prod in an emergency, etc.).

But yeah, I do like having it for reference -- it'll most likely make deleting the bucket associated with a closed-unmerged PR (which I'm working on now) way easier. 🤞 So thanks for the tip!

cnunciato: Added the metadata file and Cypress recordings as artifacts. Thanks again!

cnunciato:
> We aren't doing any cleanup of old buckets yet. An additional scheduled job would probably be best for this.

I did add a script to clean up buckets for closed, unmerged PRs, but it's not yet actually-deleting -- just logging out the deletes it _would_ make. Wanted to let this run a bit before turning it on, just for some additional safety. 😄

PR #3835, opened 2020-07-29
Add doc for configuring SAML SSO with self-hosted installations

praneetloke:
### Proposed changes

Add doc about generating and configuring the SAML key pair for enabling SSO in a self-hosted install.

N/A

### Related issues (optional)

N/A

### DIFF
--- a/content/docs/guides/self-hosted/saml-sso.md
+++ b/content/docs/guides/self-hosted/saml-sso.md
@@ -1,0 +1,57 @@
+---
+title: Enabling SAML SSO
+menu:
+    userguides:
+        parent: self_hosted
+        identifier: self_hosted_sso
+        weight: 3
+meta_desc: Learn how to make the self-hosted Pulumi ready for SAML SSO with any IdP. Self-hosting is available as part of the Enterprise Edition.
+---
+
+The self-hosted option allows you to control various aspects of the Pulumi Service including how users will logon to the [Pulumi Console]({{< relref "console" >}}).
+
+## Creating The Keys
+
+Before you can use SAML SSO to logon to the Console, however, you will need to ensure that the [API service]({{< relref "api" >}}) has a pair of keys that will be used to sign
+and validate requests/responses regardless of the IdP you choose to use.
+
+The credentials are simply a public/private key pair that will be supplied as environment variables to the API service.
+In the following snippets, we show you how you can generate a key pair by using `openssl`.
+The snippet shows the command for a self-hosted API service that is accessible via `api.company.com`.
+Be sure to adjust the value accordingly.
+
+> OpenSSL's official [wiki](https://wiki.openssl.org/index.php/Binaries) site contains links to pre-built binaries.
+
+```
+# Generate a new 2048-bit RSA key with a validity of 365 days.
+openssl \
+  req -x509 -newkey rsa:2048 \
+  -days 365 -nodes -subj "/CN=api.company.com" \
+  -keyout cert.key \
+  -out cert.cert
+```
+
+If you also want to additionally specify an SAN (Subject Alternative Name) for your public cert, you can do so by passing the `-addext` flag as shown below.
+
+> For this to work, though, you'll need to install _at least_ version 1.1. Once installed ensure that the 1.1 version is on your path when you run the command.
+> Otherwise `-addext` will not be recognized as a valid flag.
+
+```
+openssl \
+  req -x509 -newkey rsa:2048 \
+  -days 365 -nodes -subj "/CN=api.company.com" \
+  -keyout cert.key \
+  -addext "subjectAltName=DNS:anotherdomain.company.com" \
+  -out cert.cert
+```
+
+## Configure The API Service
+
+Once the key pair has been generated, set the value of the following environment variables for the API service.
+
+`SAML_CERTIFICATE_PUBLIC_KEY` should be set to the value of the `*.cert` file, i.e. the public key file.
+`SAML_CERTIFICATE_PRIVATE_KEY` should be set to the value of the `*.key` file, i.e. the private key file.
+
+For these values to take effect, you will need to restart the API Service.
+
+> Restart the service only if it is safe to do so or during a planned maintenance window.
--- a/content/docs/guides/self-hosted/saml-sso.md
+++ b/content/docs/guides/self-hosted/saml-sso.md
@@ -49,7 +49,7 @@ openssl \
 
 Once the key pair has been generated, set the value of the following environment variables for the API service:
 
-`SAML_CERTIFICATE_PUBLIC_KEY` should be set to the value of the `*cert.cert` file, i.e. the public key file.
+`SAML_CERTIFICATE_PUBLIC_KEY` should be set to the value of the `cert.cert` file, i.e. the public key file.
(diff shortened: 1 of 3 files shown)


praneetloke:
Ping.

content/docs/guides/self-hosted/saml-sso.md
```diff
@@ -0,0 +1,57 @@
...
+        identifier: self_hosted_sso
+        weight: 3
+meta_desc: Learn how to make the self-hosted Pulumi ready for SAML SSO with any IdP. Self-hosting is available as part of the Enterprise Edition.
+---
+
+The self-hosted option allows you to control various aspects of the Pulumi Service including how users will logon to the [Pulumi Console]({{< relref "console" >}}).
+
+## Creating The Keys
+
+Before you can use SAML SSO to logon to the Console, however, you will need to ensure that the [API service]({{< relref "api" >}}) has a pair of keys that will be used to sign
```
leezen, line 15:
```suggestion
Before you can use SAML SSO to logon to the Console, you will need to ensure that the [API service]({{< relref "api" >}}) has a pair of keys that will be used to sign
```

content/docs/guides/self-hosted/saml-sso.md
```diff
@@ -0,0 +1,57 @@
...
+and validate requests/responses regardless of the IdP you choose to use.
```
leezen, line 16:
```suggestion
and validate requests/responses, regardless of the IdP you choose to use.
```

content/docs/guides/self-hosted/saml-sso.md
```diff
@@ -0,0 +1,57 @@
...
+
+The credentials are simply a public/private key pair that will be supplied as environment variables to the API service.
```
leezen, line 18:
```suggestion
The credentials are simply a public/private key pair that are supplied as environment variables to the API service.
```

content/docs/guides/self-hosted/saml-sso.md
```diff
@@ -0,0 +1,57 @@
...
+openssl \
+  req -x509 -newkey rsa:2048 \
+  -days 365 -nodes -subj "/CN=api.company.com" \
+  -keyout cert.key \
+  -out cert.cert
+```
+
+If you also want to additionally specify an SAN (Subject Alternative Name) for your public cert, you can do so by passing the `-addext` flag as shown below.
+
+> For this to work, though, you'll need to install _at least_ version 1.1. Once installed ensure that the 1.1 version is on your path when you run the command.
```
leezen, line 36:
```suggestion
> For this to work, though, you'll need to install _at least_ `openssl` version 1.1. Once installed, ensure that the 1.1 version is on your path when you run the command.
```

content/docs/guides/self-hosted/saml-sso.md
```diff
@@ -0,0 +1,57 @@
...
+  req -x509 -newkey rsa:2048 \
+  -days 365 -nodes -subj "/CN=api.company.com" \
+  -keyout cert.key \
+  -addext "subjectAltName=DNS:anotherdomain.company.com" \
+  -out cert.cert
+```
+
+## Configure The API Service
+
+Once the key pair has been generated, set the value of the following environment variables for the API service.
```
leezen, line 50:
```suggestion
Once the key pair has been generated, set the value of the following environment variables for the API service:
```

content/docs/guides/self-hosted/saml-sso.md
```diff
@@ -0,0 +1,57 @@
...
+
+`SAML_CERTIFICATE_PUBLIC_KEY` should be set to the value of the `*.cert` file, i.e. the public key file.
+`SAML_CERTIFICATE_PRIVATE_KEY` should be set to the value of the `*.key` file, i.e. the private key file.
```
leezen, line 53: Since we used `cert.key` and `cert.cert` above, it probably makes sense to re-use those names here.

content/docs/guides/self-hosted/saml-sso.md
```diff
@@ -0,0 +1,57 @@
...
+
+For these values to take effect, you will need to restart the API Service.
+
+> Restart the service only if it is safe to do so or during a planned maintenance window.
```
leezen, line 57: I would remove this line. "if it is safe to do so" is pretty ambiguous and now as the reader I'm wondering, "how do I check if it's safe to do so"? In general, I would expect that you've either setup the service in some kind of HA way (e.g. behind a LB with multiple containers running) or you know you only have a single instance and that restarting the service necessarily means downtime.

content/docs/guides/self-hosted/saml-sso.md
praneetloke: I wanted to emphasize that the version of openssl is 1.1, which this blockquote is already in the context of running an `openssl` command. So I'll leave it as-is.

content/docs/guides/self-hosted/saml-sso.md
praneetloke: Fair enough. I'll drop the "safe to do so" part.

praneetloke, line 57:
```suggestion
> Restart the service only during a planned maintenance window.
```

content/docs/guides/self-hosted/saml-sso.md
```diff
@@ -0,0 +1,57 @@
...
+  -keyout cert.key \
+  -addext "subjectAltName=DNS:anotherdomain.company.com" \
+  -out cert.cert
+```
+
+## Configure The API Service
+
+Once the key pair has been generated, set the value of the following environment variables for the API service.
+
+`SAML_CERTIFICATE_PUBLIC_KEY` should be set to the value of the `*.cert` file, i.e. the public key file.
```
praneetloke, line 52:
```suggestion
`SAML_CERTIFICATE_PUBLIC_KEY` should be set to the value of the `*cert.cert` file, i.e. the public key file.
```

content/docs/guides/self-hosted/saml-sso.md
praneetloke, line 53:
```suggestion
`SAML_CERTIFICATE_PRIVATE_KEY` should be set to the value of the `cert.key` file, i.e. the private key file.
```

content/docs/guides/self-hosted/saml-sso.md
```diff
@@ -0,0 +1,57 @@
...
+  -keyout cert.key \
+  -addext "subjectAltName=DNS:anotherdomain.company.com" \
+  -out cert.cert
+```
+
+## Configure The API Service
+
+Once the key pair has been generated, set the value of the following environment variables for the API service:
+
+`SAML_CERTIFICATE_PUBLIC_KEY` should be set to the value of the `*cert.cert` file, i.e. the public key file.
```
leezen, line 52: Drop the `*`?

praneetloke: 🤦

praneetloke, line 52:
```suggestion
`SAML_CERTIFICATE_PUBLIC_KEY` should be set to the value of the `cert.cert` file, i.e. the public key file.
```

PR #3845, opened 2020-07-31
Clean up infra code

cnunciato:
With the changes in #3799, we no longer need some of our website infrastructure. This change clears it out.


infrastructure/index.ts
```diff
@@ -130,67 +74,6 @@ const originBucketPolicy = new aws.s3.BucketPolicy("origin-bucket-policy", {
...
-
-// logsBucket stores the request logs for incoming requests.
-const logsBucket = new aws.s3.Bucket(
-    "requestLogs",
-    {
-        bucket: `${config.targetDomain}-logs`,
-        acl: "private",
-    },
-    {
-        protect: true,
```
cnunciato, line 191: This resource was manually removed with a `pulumi state delete` in order to allow this change to proceed. The bucket itself is still around, and I've added a README to its contents explaining what it was used for. 

Additional context here: https://pulumi.slack.com/archives/C6Y560ADV/p1596152486079600

PR #3848, opened 2020-07-31
Updating audit logs docs

sean1588:
This is an update to the current audit log doc to include the new features we added over the last couple weeks.

We added some new Audit Log Events as well as support for CEF format.

### DIFF
--- a/content/docs/intro/console/collaboration/auditing.md
+++ b/content/docs/intro/console/collaboration/auditing.md
@@ -74,9 +74,7 @@ curl \
 
 ### Supported Audit Log Formats
 
-We currently support two formats for exporting audit logs through the API. The two formats we currently support
-are CSV and CEF. These formats can be specified by appending the `format` query parameter as follows, `format=csv`
-or `format=cef`.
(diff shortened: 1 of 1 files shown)


content/docs/intro/console/collaboration/auditing.md
```diff
@@ -71,3 +71,121 @@ curl \
 ```
 
 > _Note: Substitute `${org}`, `${user}`,  and `${time}` for your actual values - e.g. `org`, `username`, and `1583460637`._
+
+### Supported Audit Log Formats
+
+We currently support two formats for exporting audit logs through the API. The two formats we currently support
+are CSV and CEF. These formats can be specified by appending the `format` query parameter as follows, `format=csv`
+or `format=cef`.
```
ericrudder, line 79:
```suggestion
Pulumi supports multiple formats for exporting audit logs. These formats can be specified by appending the `format` query parameter, for example, `format=csv`or `format=cef`.
```

sean1588: Thanks, that sounds a million times better!

sean1588:
> This looks good, consider defining acronyms (csv, cef, siem) at first use.

Good idea. Will add what the acronyms stand for.

PR #3852, opened 2020-07-31
Update self-hosted docs to include TLS environment variables

sean1588:
Updating self-hosted docs to include the environment variables needed to configure TLS connections.

### DIFF
--- a/content/docs/guides/self-hosted/api.md
+++ b/content/docs/guides/self-hosted/api.md
@@ -61,3 +61,25 @@ The API service is a Go-based application. This is a single binary application t
 | SAML_CERTIFICATE_PRIVATE_KEY | Private key used by Pulumi to validate the SAML assertions sent by the IdP. Learn how to set this value [here]({{< relref "saml-sso" >}}). |
 | GITHUB_OAUTH_ENDPOINT | Used for GitHub API calls. |
 | GITLAB_OAUTH_ENDPOINT | Used for GitLab API calls. |
+
+### TLS Environment Variables
+
+#### API Service
+
+The service is configurable to serve the API using TLS. The following environment variables are _all_ required in order to enable TLS.
+
+| Variable Name       | Description                                                                                               |
+|---------------------|-----------------------------------------------------------------------------------------------------------|
+| API_TLS_CERTIFICATE | The TLS certificate. The certificate must be supplied in X.509 format and must be PEM encoded.            |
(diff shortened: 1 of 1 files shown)


content/docs/guides/self-hosted/api.md
```diff
@@ -61,3 +61,25 @@ The API service is a Go-based application. This is a single binary application t
...
+| Variable Name       | Description                                                                                               |
+|---------------------|-----------------------------------------------------------------------------------------------------------|
+| API_TLS_CERTIFICATE | The TLS certificate. The certificate must be supplied in X.509 format and must be PEM encoded.            |
+| API_TLS_PRIVATE_KEY | The private key associated with the TLS certificate. The private key must be PEM encoded.                 |
+| API_MIN_TLS_VERSION | The minimum version of TLS to use when serving the API (must be in \<major>.\<minor> format, e.g. `1.2`). |
+
+#### Database Connections
+
+The service is configurable to enable connections to the backend SQL database over TLS. The following environment variables are _all_
+required to connect to the database using TLS.
```
praneetloke, line 80: When the preview is ready, could you please check if this sentence renders correctly, i.e. it is a continuation of the sentence and not a new one?

content/docs/guides/self-hosted/api.md
```diff
@@ -61,3 +61,25 @@ The API service is a Go-based application. This is a single binary application t
 | SAML_CERTIFICATE_PRIVATE_KEY | Private key used by Pulumi to validate the SAML assertions sent by the IdP. Learn how to set this value [here]({{< relref "saml-sso" >}}). |
 | GITHUB_OAUTH_ENDPOINT | Used for GitHub API calls. |
 | GITLAB_OAUTH_ENDPOINT | Used for GitLab API calls. |
+
+### TLS Environment Variables
+
```
praneetloke, line 66: I think it would be good to also call out if a self-signed certificate can be used for each of these touchpoints and a sample command showing how to generate one would be good too.

content/docs/guides/self-hosted/api.md
sean1588: It rendered correctly, but I just went ahead and moved them to the same line so we could get rid of the confusion.

content/docs/guides/self-hosted/api.md
sean1588: ok, I added a note with an example of using openssl to generate self-signed certs in case they would like to use it.

PR #3863, opened 2020-08-03
Don't reuse PR builds

### LINKED ISSUE #3862: The Segment script is not loading in production
Currently, our Segment script is not loading in production which is causing an outage w/ our analytics stack. The below photo shows out Google Analytics traffic from July 30th until Today (August 3rd). There is a large dip in on July 30th indicating we made a change that stopped loading the Segment script.

broken-analytics

cnunciato:
As part of #3799, in the interest of shortening the time it takes to get a change into production, I added a bit of logic to promote the already-built-and-tested buckets we prepare for PRs directly to pulumi.com. I called this out in the PR diagram and description ---

> Right now, PRs are building drafts and futures, so on merge, since we're just carrying those buckets across as-is, draft and future-dated content will (at least as of now) be released publicly as well. 

... but what I didn't realize was that PR builds also exclude our analytics scripts, and it turns out, we need those! This change removes the find-or-rebuild script and just builds and tests a new site on every push. 

Also renames the bucketize script to call out that it also runs tests.

Fixes #3862.


cnunciato:
> If that's the case, then what's the advantage of having the bucket in the first place? Is it just to make the "CloudFront update" be atomic? (So that whenever the CloudFront distribution changes, there is no opportunity to read the wrong version of a file or something?)

These changes were originally motivated by wanting to build whole versions of the site and deploy them atomically just by switching the CloudFront origin pointer, rather than have a single bucket be continually updated on an ongoing basis, as before (which came with various problems). For PRs, the buckets are for publishing site previews, which we used to do with a separate Pulumify action, but are now doing with a single script that just knows how to handle both PR and push builds. So we do still need to create those buckets, yeah. (I have a change in progress that focuses on cleaning them up more judiciously as well.) 

Hopefully that answers your question. 😄 If not, lemme know!

PR #3868, opened 2020-08-04
Add documentation for TLS config for console.

sean1588:


content/docs/guides/self-hosted/console.md
```diff
@@ -79,3 +79,20 @@ The following are the core environment variables that are required at a minimum.
...
+| CONSOLE_TLS_PRIVATE_KEY | The private key associated with the TLS certificate. The private key must be PEM encoded.                     |
+| CONSOLE_MIN_TLS_VERSION | The minimum version of TLS to use when serving the console (must be in \<major>.\<minor> format, e.g. `1.2`). |
+
+> _Note: Self-signed certificates may be used to configure TLS in the event the need for a trusted entity is not necessary. A self-signed cert and private key may be generated using OpenSSL. The following command uses OpenSSL to generate a self-signed certificate. This example will output two files, the certificate (cert.pem) and the private key (key.pem) used to sign it._
+
+>
+```
+openssl \
+  req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem \
+  -days { days_unitl_expiration } -nodes -subj "/CN={ common_name }"
```
cnunciato, line 97:
```suggestion
  -days { days_until_expiration } -nodes -subj "/CN={ common_name }"
```

content/docs/guides/self-hosted/console.md
```diff
@@ -79,3 +79,20 @@ The following are the core environment variables that are required at a minimum.
...
+
+### TLS Environment Variables
+
+| Variable Name       | Description                                                                                                       |
+|---------------------|-------------------------------------------------------------------------------------------------------------------|
+| CONSOLE_TLS_CERTIFICATE | The TLS certificate. The certificate must be supplied in X.509 format and must be PEM encoded.                |
+| CONSOLE_TLS_PRIVATE_KEY | The private key associated with the TLS certificate. The private key must be PEM encoded.                     |
+| CONSOLE_MIN_TLS_VERSION | The minimum version of TLS to use when serving the console (must be in \<major>.\<minor> format, e.g. `1.2`). |
+
+> _Note: Self-signed certificates may be used to configure TLS in the event the need for a trusted entity is not necessary. A self-signed cert and private key may be generated using OpenSSL. The following command uses OpenSSL to generate a self-signed certificate. This example will output two files, the certificate (cert.pem) and the private key (key.pem) used to sign it._
```
cnunciato, line 91: No need to italicize:

```suggestion
> Note: Self-signed certificates may be used to configure TLS in the event the need for a trusted entity is not necessary. A self-signed cert and private key may be generated using OpenSSL. The following command uses OpenSSL to generate a self-signed certificate. This example will output two files, the certificate (cert.pem) and the private key (key.pem) used to sign it.
```

content/docs/guides/self-hosted/console.md
```diff
@@ -79,3 +79,20 @@ The following are the core environment variables that are required at a minimum.
...
 | BITBUCKET_OAUTH_ID| Atlassian Bitbucket OAuth consumer client ID. Used for Bitbucket OAuth signins. |
 | BITBUCKET_OAUTH_SECRET | |
+
+### TLS Environment Variables
+
+| Variable Name       | Description                                                                                                       |
+|---------------------|-------------------------------------------------------------------------------------------------------------------|
+| CONSOLE_TLS_CERTIFICATE | The TLS certificate. The certificate must be supplied in X.509 format and must be PEM encoded.                |
+| CONSOLE_TLS_PRIVATE_KEY | The private key associated with the TLS certificate. The private key must be PEM encoded.                     |
+| CONSOLE_MIN_TLS_VERSION | The minimum version of TLS to use when serving the console (must be in \<major>.\<minor> format, e.g. `1.2`). |
```
cnunciato, line 89: Is there a default value for this one? If so, we should mention what the value is, and that it's optional.

sean1588: No, there is not a default value. If the value is not set, a minimum version will just not be enforced. To your point though, I think it would be good to call out that it is an optional setting.

PR #3880, opened 2020-08-05
Identify deletable buckets

cnunciato:
This change updates the `list-buckets` script to identify buckets that can be safely deleted as well.

buckets mov

### DIFF
--- a/scripts/list-recent-buckets.sh
+++ b/scripts/list-recent-buckets.sh
@@ -2,31 +2,46 @@
 
 set -o errexit -o pipefail
 
-# This script lists the 20 most recent S3 buckets populated from docs-site build jobs. It
+# This script lists the 50 most recent S3 buckets populated from docs-site build jobs. It
 # works by querying S3 for buckets with names matching our origin-bucket naming prefix,
 # then fetches the metadata files we generate with each build, outputting their results.
 #
-# Note that these buckets are not limited to those whose content has already been merged;
-# this includes everything, including buckets for open PRs and those that failed to build.
-# Commit links should help to identify buckets whose contents have been merged, however,
-# and in general, the presence of a metadata file implies the bucket was built and tested
-# successfully.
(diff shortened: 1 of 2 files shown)


scripts/list-recent-buckets.sh
```diff
@@ -2,31 +2,46 @@
...
 
 if [ "$bucket_count" == "0" ]; then
-    echo "No recent $(origin_bucket_prefix)-* buckets found."
+    echo "No recent buckets matching the prefix $(origin_bucket_prefix)-${bucket_prefix} were found."
     exit
 fi
 
-echo "Found ${bucket_count} recent $(origin_bucket_prefix)-* buckets:"
+# Query for the bucket currently serving pulumi.com.
+currently_deployed_bucket="$(curl -s https://www.pulumi.com/metadata.json | jq -r '.bucket' || echo '')"
```
sean1588, line 30: My only concern here would be if the json structure changes and we return empty string it may skip the comparison below to warn that this is the current bucket serving pulumi.com, but I am guessing that `pulumi.com/metadata.json` is also being sourced the same way from s3 as as s3://....metadata.json,  where we are geting the bucket name from on line 53, so I imagine this should be fine.

cnunciato: Yes, indeed -- if this structure were somehow changed, then this script would start yielding weird results, and indeed, a few other things would probably start breaking, too. 😅 

In this case, I think the jq result would actually be `null` in both of these cases (say, if the property name were suddenly no longer `bucket`), so the comparison would match for _every_ bucket:

image

It's a good thing to bear in mind, though. It'd probably good to file an issue to harden things a bit more around this structure, if only to define it in one and only one place. I'll do that. Thanks for the head's up.

sean1588: Ahh ok. yea that makes sense. thanks for the explanation.

PR #3889, opened 2020-08-06
Update TLS info in self-hosted docs.

sean1588:
This is just adding some updates to the self-hosted docs concerning TLS configuration. I apologize, I meant to add in this information about the port changes on Tuesday before merging, but it must have slipped my mind in the rush of trying to get everything out.

### DIFF
--- a/content/docs/guides/self-hosted/api.md
+++ b/content/docs/guides/self-hosted/api.md
@@ -38,7 +38,7 @@ The Pulumi API is one of the components required for self-hosting Pulumi in your
 
 ## What's In The Container?
 
-The API service is a Go-based application. This is a single binary application that has all of the dependencies it needs in order to run.
+The API service is a Go-based application. This is a single binary application that has all of the dependencies it needs in order to run. This container will serve using port `8080` over HTTP by default or port `8443` over HTTPS when [TLS](#tls-environment-variables) is configured.
 
 ## Primary Environment Variables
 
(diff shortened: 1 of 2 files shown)


content/docs/guides/self-hosted/api.md
```diff
@@ -38,7 +38,7 @@ The Pulumi API is one of the components required for self-hosting Pulumi in your
 
 ## What's In The Container?
 
-The API service is a Go-based application. This is a single binary application that has all of the dependencies it needs in order to run.
+The API service is a Go-based application. This is a single binary application that has all of the dependencies it needs in order to run. This container will serve using port `8080` over HTTP by default or port `8443` over HTTPS when [TLS](#tls-environment-variables) is configured.
```
praneetloke, line 41: How about we create a new sub-heading so that it is easy to notice this important detail? I am thinking something like `### Ports`? If you have a better name, definitely please feel free to use that.

content/docs/guides/self-hosted/console.md
```diff
@@ -36,7 +36,7 @@ You can run this container on the same host that your API container is running o
 
 ## What's In The Container?
 
-The Console container is a Node 10-based image. A single web server handles delivering UI static assets to your web browser, as well as handling authentication callbacks.
+The Console container is a Node 10-based image. A single web server handles delivering UI static assets to your web browser, as well as handling authentication callbacks. The console will be served using port `3000` over HTTP by default. If [TLS](#tls-environment-variables) is configured the console will be served using port `3443` over HTTPS.
```
praneetloke, line 39: Same comment here as above for the API container.

content/docs/guides/self-hosted/api.md
sean1588: Agreed, I think a separate heading would give it more visibility.

PR #3903, opened 2020-08-08
Handle job re-runs, exit early when forked PRs are closed

cnunciato:
Couple of fixes to the way we handle Actions jobs:

* For those whose buckets already exist, we attempt to reuse them (which is safe, as buckets are named by the current commit hash)

* When a PR based on a fork is closed, we exit early, as the cleanup work we do isn't necessary in that case (and currently fails due to missing secrets)

👇 Below, you'll find several PR comments showing multiple re-runs of identical commits.

### DIFF
--- a/scripts/ci-pull-request-closed.sh
+++ b/scripts/ci-pull-request-closed.sh
@@ -2,6 +2,12 @@
 
 set -o errexit -o pipefail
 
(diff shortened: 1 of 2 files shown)


cnunciato:
@stack72 @jkisk Friendly pings, if either of y'all have a few to take a look at this. Thanks!

PR #3912, opened 2020-08-10
Add Sentry for client-side error reporting

### LINKED ISSUE #3371: Better detection for client-side errors on the website
We've released a few builds in recent months resulting in JavaScript errors that went undetected and had to be reported by users. We should investigate options for detecting these ourselves. Something like [Sentry](https://sentry.io) might work. 

Part of pulumi/pulumi-service#4957.

cnunciato:
This change adds Sentry to help us get notified of client-side issues with pulumi.com. Errors are posted to Slack and collected into the pulumi/docs project in Sentry that the team can use to track issues and resolutions.

Fixes #3371.


layouts/partials/head.html
```diff
@@ -97,6 +97,14 @@
     <!-- Our CSS bundle. -->
     <link rel="stylesheet" href="{{ getenv "REL_CSS_BUNDLE" }}">
 
+    <script src="https://browser.sentry-cdn.com/5.20.1/bundle.min.js" integrity="sha384-O8HdAJg1h8RARFowXd2J/r5fIWuinSBtjhwQoPesfVILeXzGpJxvyY/77OaPPXUo" crossorigin="anonymous"></script>
```
zchase, line 100: Any chance we can load these scripts with `defer` or `async`?

cnunciato: I believe in this case I'd want to _not_ defer, so as to be sure to catch errors as early as possible (which is why I placed it in the HEAD, above pretty much everything else). This is ~17K gzipped and forever-cached, though, so hopefully it won't register with the Googlebots.

cnunciato: I'll look into this once more, though, just to see if there's a practical balance we can achieve.

cnunciato: Yeah, given that we aren't deferring the component loader, and this is so cache-friendly already, I'd feel better if we didn't optimize for this just yet. We can always revisit if necessary, though.

cnunciato: Ah, on second thought, we have a lot of includes that are not deferred, looks like. Let's keep this in mind and circle back to it as a more holistic PR.

Issue #3904, opened 2020-08-08
Docs recommend that ec2 instance creation uses a deprecated parameter. 


twk:
[docs/tutorials/aws/ec2-webserver.md](https://www.pulumi.com/docs/tutorials/aws/ec2-webserver/)

```
const server = new aws.ec2.Instance("webserver-www", {
    instanceType: size,
    securityGroups: [ group.name ], // reference the security group resource above
    ami: ami.id,
});
```

In your docs, you say that usage of `securityGroups` is [discouraged](https://www.pulumi.com/docs/reference/pkg/aws/ec2/instance/#securitygroups_nodejs).

lukehoban:
Thanks - yes - we should update this tutorial and the examples it is based on in pulumi/examples.

stack72:
@lukehoban *part* of this tutorial was already updated... it's the manually created tutorial. I just opened https://github.com/pulumi/docs/pull/3917 that fixes this

PR #3917, opened 2020-08-11
Update ec2-webserver tutorial to remove deprecated parameter

stack72:
Fixes: #3904

Updates the manually created tutorial

### DIFF
--- a/content/docs/tutorials/aws/ec2-webserver.md
+++ b/content/docs/tutorials/aws/ec2-webserver.md
@@ -97,7 +97,7 @@ let group = new aws.ec2.SecurityGroup("webserver-secgrp", {
 
 let server = new aws.ec2.Instance("webserver-www", {
     instanceType: size,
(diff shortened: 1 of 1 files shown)


lukehoban:
Just to make sure - do these work in this form?  Wasn't sure if perhaps you need to specify the VPC explicitly in this form?

stack72:
> Just to make sure - do these work in this form? Wasn't sure if perhaps you need to specify the VPC explicitly in this form?

oh I see what you mean - like adding the vpc to the actual security group - I will rework the entire tutorial to adhere to this

stack72:
> Just to make sure - do these work in this form? Wasn't sure if perhaps you need to specify the VPC explicitly in this form?

I take that back - yes, these do work in this form - we have examples running in pulumi/examples of this format - https://github.com/pulumi/examples/blob/master/aws-js-webserver/index.js

https://github.com/pulumi/examples/runs/970459231?check_suite_focus=true#step:22:198

PR #3919, opened 2020-08-11
fix: typo

gunar:


justinvp:
Also fixed it in the source README: https://github.com/pulumi/pulumi-awsx/pull/562

PR #3922, opened 2020-08-12
Preload component.js files

zchase:
We seemed to have regressed a bit in our Google Page Speed score and the biggest item is the loading of `component.js` scripts. In reality, this change doesn't really make the site faster and you can see that in the screenshots below, but it does make Google Page Speed Score increase which plays a role in Google Search Rankings. Our ranking for the term `terraform` has been fluctuating a bit recently between 7/8, which is the difference between page one and page two currently. I checked terraform.io's page speed and they are tied with us. This change should get us a few points above them and in turn, help our content rank a bit better.

This PR adds two `preload` tags to `head.html`. More info on `preload` here: https://web.dev/uses-rel-preload/

I ran the Page Speed tests a bunch of times and with the `preload` tags the score was around ~6 points higher each run.

Preview Bucket Page Speed from PR w/ out `preload`
developers google com_speed_pagespeed_insights__url=http%3A%2F%2Fpulumi-docs-origin-pr-3912-367f6211 s3-website us-west-2 amazonaws com%2F

Page Speed for this PR:
developers google com_speed_pagespeed_insights__url=http%3A%2F%2Fpulumi-docs-origin-pr-3922-a4a073f1 s3-website us-west-2 amazonaws com%2F

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -120,8 +120,8 @@
     {{ end }}
 
     <!-- Preload the component JavaScript to help our Google Page Speed score. -->
-    <link rel="preload" href="/js/components/components.esm.js" as="script">
-    <link rel="preload" href="/js/components.js" as="script">
+    <link rel="preload" href="/js/components/components.esm.js" as="script" crossorigin>
+    <link rel="preload" href="/js/components.js" as="script" crossorigin>
 
     {{/* Tests to enforce required frontmatter for certain content types. */}}
     {{ if and (eq .Type "blog") .IsPage }}
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -119,6 +119,10 @@
         </script>
     {{ end }}
 
+    <!-- Preload the component JavaScript to help our Google Page Speed score. -->
+    <link rel="prefetch" href="/js/components/components.esm.js" as="script">
+    <link rel="prefetch" href="/js/components.js" as="script">
+
     {{/* Tests to enforce required frontmatter for certain content types. */}}
     {{ if and (eq .Type "blog") .IsPage }}
         {{ if not .Params.authors }}
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -119,6 +119,10 @@
         </script>
     {{ end }}
 
+    <!-- Preload the component JavaScript to help our Google Page Speed score. -->
+    <link rel="preload" href="/js/components/components.esm.js" as="script" crossorigin>
+    <link rel="preload" href="/js/components.js" as="script" crossorigin>
+
     {{/* Tests to enforce required frontmatter for certain content types. */}}
     {{ if and (eq .Type "blog") .IsPage }}
         {{ if not .Params.authors }}
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -119,10 +119,6 @@
         </script>
     {{ end }}
 
-    <!-- Preload the component JavaScript to help our Google Page Speed score. -->
-    <link rel="preload" href="/js/components/components.esm.js" as="script" crossorigin>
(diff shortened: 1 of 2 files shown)


layouts/partials/head.html
```diff
@@ -119,6 +119,10 @@
         </script>
     {{ end }}
 
+    <!-- Preload the component JavaScript to help our Google Page Speed score. -->
+    <link rel="preload" href="/js/components/components.esm.js" as="script">
```
cnunciato, line 123: Can you share some data on this? The smaller bundle (which is the one that should be loaded nearly all of the time) is like < 10K over the wire.

zchase: Page Speed Score: https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fpulumi.com%2F

I'm following the instructions they output. I'm playing around with this to see what we get back by preloading. My plan was if I end up going forward with this to add more details to the description.

pagespeedpreload

layouts/partials/head.html
```diff
@@ -119,6 +119,10 @@
         </script>
     {{ end }}
 
+    <!-- Preload the component JavaScript to help our Google Page Speed score. -->
+    <link rel="preload" href="/js/components/components.esm.js" as="script" crossorigin>
```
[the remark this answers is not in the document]
zchase: I'm fairly certain without the crossorigin the prefetch doesn't work because we need to fetch the script "anonymously" but I am going to commit this and confirm that.

layouts/partials/head.html
```diff
@@ -119,6 +119,10 @@
...
+    <link rel="preload" href="/js/components.js" as="script" crossorigin>
```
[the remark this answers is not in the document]
zchase: 👍

zchase:
@cnunciato I ended up having to keep the `crossorigin` attribute to avoid throwing a warning about fetching the file with the wrong user credentials. Setting the attribute without a value uses the default `anonymous` value so we fetch the script anonymously in head of the HTML file. This clears up the Google Page Speed issue.

I'm a bit skeptical to do this with jQuery for a few reasons:

- Google isn't complaining about it
- My reading made me uneasy about support for integrity checking. It seems like integrity checking is only recently support for `preloads` and I'd rather error on the side of caution here.

cnunciato:
Okay, I see how this works now. This is a very confusing API, but I think I understand, now. 😄  

What it means when you say `crossorigin` (or `crossorigin=""`, or `crossorigin="anonymous"`, which is probably best, since it's most explicit) is that "credentials" (meaning cookies, auth headers, TLS info and the like) should not be sent to origins _other than the current one_. They're still sent when the request origin and current origin line up, though.

So I think this is a case of Chrome (and it sounds like Googlebot) being a bit pushy, because there isn't actually any difference between the requests being made with the `crossorigin` attribute set and without; in both cases, cookies are sent with the request for this file, because the origins are the same.

In any case, I feel better now that I understand what the deal is here. 😄 I'll update my suggestion to use `crossorigin="anonymous"`, since it's same thing, but again, just a bit more explicit.

Some additional reading that helped me make sense of this, alongside some experimenting:

https://html.spec.whatwg.org/multipage/urls-and-fetching.html#cors-settings-attribute
https://fetch.spec.whatwg.org/#concept-request-credentials-mode
https://fetch.spec.whatwg.org/#credentials

layouts/partials/head.html
```diff
@@ -119,6 +119,9 @@
         </script>
     {{ end }}
 
+    <!-- Preload the component JavaScript to help our Google Page Speed score. -->
+    <link rel="preload" href="/js/components/components.esm.js" as="script" crossorigin>
```
cnunciato, line 123:
```suggestion
    <link rel="preload" href="/js/components/components.esm.js" as="script" crossorigin="anonymous">
```

PR #3946, opened 2020-08-13
Add alias to Kubernetes to handle /cloudnative links

zchase:
In our press release, we linked to `/cloudnative` instead of `/kubernetes`. This PR adds an alias to `/kubernetes` for `/cloudnative` so we can avoid 404s.


cnunciato:
This looks fine, except that I don't see any link to `/cloudnative` in the actual source of the press release. Do we still need it?

cnunciato:
Ah, I see the note about the copied version as well now.

PR #3957, opened 2020-08-14
Update k8s docs

lblackstone:
### Proposed changes
- Fix Python's yaml.ConfigFile examples

Related to https://github.com/pulumi/pulumi-kubernetes/pull/1248


content/docs/reference/pkg/kubernetes/apps/v1/daemonset.md
```diff
@@ -7290,8 +7290,8 @@ All [input](#inputs) properties are implicitly available as output properties. A
 
     <dt class="property-optional"
             title="Optional">
-        <span id="target_ww_ns_python">
-<a href="#target_ww_ns_python" style="color: inherit; text-decoration: inherit;">target_<wbr>ww_<wbr>ns</a>
+        <span id="target_wwns_python">
+<a href="#target_wwns_python" style="color: inherit; text-decoration: inherit;">target_<wbr>wwns</a>
```
justinvp, line 7294: This is a consequence of https://github.com/pulumi/pulumi/pull/4918.

We need to update the docgen to use the new `PyNameLegacy` added in https://github.com/pulumi/pulumi/pull/4918, since the names emitted here in the docs should be the names that will be used in the input/output `dict` objects (from the mapping table).

But this is really just a point-in-time problem for the next week or two, as the providers will soon have `UsesIOClasses` added to their schema for Python, and with https://github.com/pulumi/pulumi/pull/5137, the Python docs will get type annotations and should use the "correct" name (using `PyName`) for these nested types.

I've opened to https://github.com/pulumi/pulumi/issues/5175 to track fixing this short-term bug.

@lblackstone, would you prefer to wait until the short-term fix is available (and then regenerate the docs), or just manually revert the name changes in this PR so it can be merged?

lblackstone: Let's just wait until the fix is available. This isn't an urgent change.

justinvp (approved):

PR #3962, opened 2020-08-14
Publish sourcemaps to Sentry

cnunciato:
This change adds a step to the build process to publish our sourcemaps to Sentry, for better stack traces in error reports. For example:

image
 
Also contains a tiny fix to suppress the error message emitted by Pulumi when an optional config value isn't defined.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,6 @@ cypress/screenshots
 # Ignore the files we generate during the build and deployment process.
 origin-bucket-metadata.json
 redirects.txt
+
+# Ignore the temporary directory created for Sentry releases.
+sentry-release
--- a/scripts/publish-sentry-release.sh
+++ b/scripts/publish-sentry-release.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+set -o errexit -o pipefail
+
+# This script packages our JS source files and sourcemaps as a Sentry release.
+# https://docs.sentry.io/product/releases/
+
+source ./scripts/common.sh
+
+build_dir="public"
+source_dir="assets"
+
+# Make a folder to contain the source files and maps we upload to Sentry.
+release_dir="sentry-release"
+rm -rf "${release_dir}"
+mkdir -p "${release_dir}" "${release_dir}/assets"
+
+# Copy the files into that folder.
+cp -R "assets/js" "${release_dir}/assets/"
+find "${source_dir}" -name "*.js" -o -name "*.ts" -exec cp {} "${release_dir}/" \;
+find "${build_dir}" -name "*.js.map" -exec cp {} "${release_dir}/" \;
+
+# Create a new Sentry release and finalize it.
+export SENTRY_ORG="pulumi"
+export SENTRY_PROJECT="docs"
+yarn run sentry-cli releases new "$(build_identifier)"
+yarn run sentry-cli releases files "$(build_identifier)" upload-sourcemaps "${release_dir}"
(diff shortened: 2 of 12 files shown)


.gitignore
```diff
@@ -43,3 +43,6 @@ cypress/screenshots
 # Ignore the files we generate during the build and deployment process.
 origin-bucket-metadata.json
 redirects.txt
+
+# Ignore the temporary directory created for Sentry releases.
+sentry-release
```
chrsmith, line 48: The name is sufficiently clear, but would `bin/sentry-release` make more sense? (Especially since there might be other build outputs we'd want to add to `.gitignore` for things like this?

scripts/publish-sentry-release.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -o errexit -o pipefail
+
+# This script packages our JS source files and sourcemaps as a Sentry release.
+# https://docs.sentry.io/product/releases/
+
+source ./scripts/common.sh
+
+build_dir="public"
```
chrsmith, line 10: NIT: As a good practice, you probably want to declare these with `readonly` like, `readonly build_dir="public"`.

scripts/publish-sentry-release.sh
```diff
@@ -0,0 +1,28 @@
...
+source_dir="assets"
+
+# Make a folder to contain the source files and maps we upload to Sentry.
+release_dir="sentry-release"
+rm -rf "${release_dir}"
+mkdir -p "${release_dir}" "${release_dir}/assets"
```
chrsmith, line 16: My understanding is that the magic of `mkdir -p` is that it will create all subfolders if needed. So this can just be `mkdir -p "${release_dir}/assets"`.

scripts/publish-sentry-release.sh
```diff
@@ -0,0 +1,28 @@
...
+# https://docs.sentry.io/product/releases/
+
+source ./scripts/common.sh
+
+build_dir="public"
+source_dir="assets"
+
+# Make a folder to contain the source files and maps we upload to Sentry.
+release_dir="sentry-release"
+rm -rf "${release_dir}"
```
chrsmith, line 15: It doesn't quite seem right to create a folder immediately after blowing it away. Also, is there any reason why the release should be in the current directory? Could we just write the Sentry stuff to tmp?
```
# Just put the sentry release into a temp folder,
# so no need to add to .gitignore or delete on publish.
readonly release_dir=$(mktemp -d -t "sentry-release")
...
```

scripts/sync-and-test-bucket.sh
```diff
@@ -37,7 +37,7 @@ destination_bucket_uri="s3://${destination_bucket}"
 # Translate Hugo redirects into a file we'll use for making 301 redirects later. Note that
 # we do this before syncing the site, as having the redirects file available for reference
 # can be helpful for debugging purposes.
-node scripts/translate-redirects.js "$build_dir" "$(pulumi -C infrastructure config get redirectDomain || echo '')"
+node scripts/translate-redirects.js "$build_dir" "$(pulumi -C infrastructure config get redirectDomain 2>/dev/null || echo '')"
```
chrsmith, line 40: To keep things a little simpler, could we put `"$(pulumi -C infrastructure config get redirectDomain 2>/dev/null || echo '')"` on a separate line? That would make things a bit more readable, IMHO.

PR #4020, opened 2020-08-18
Add back tf2pulumi page

joeduffy:
This page was renamed to kube2pulumi and shipped (huzzah!), so this
change adds it back. It also changes the URL to adopt the endpoint we
stood up in production, rather than my personal Pulumi account (oops).


content/migrate/tf2pulumi.md
```diff
@@ -0,0 +1,633 @@
...
+}
+
+function getCannedExample(id) {
+    if (id === undefined) {
+        // Look up the currently selected example ID.
+        id = $("#terraform-canned-example").children("option:selected").attr("id");
+    }
+
+    let comment = "#"; // to suppress Markdown lint errors.
+    switch (id) {
```
EvanBoyle, line 391: Sounds like this switch was causing some sentry alerts earlier. We should pick up the fix from here https://github.com/pulumi/docs/pull/4012

joeduffy: Great, thanks for catching this -- fixed.

PR #4030, opened 2020-08-18
Replace Hire with Bamboo

cnunciato:


cnunciato:
Proactively merging this as it looks good to me and fixes a problem -- but @freddy-hernandez do let me know if you'd like anything else done for this.

Issue #4029, opened 2020-08-18
On pulumi.com/tf2pulumi the code input disappears when changing languages


zchase:
The language chooser is causing the code input the disappear on /tf2pulumi and /kube2pulumi. This is likely due to some weird issue with the input chooser listening for a change in the language chooser. 

![image (1)](https://user-images.githubusercontent.com/15146337/90567371-e7c76000-e15e-11ea-82b5-7dd628b153ea.png)

zchase:
Closed because https://github.com/pulumi/docs/pull/4266 fixed the issue.

Issue #4016, opened 2020-08-18
Refactor kube2pulumi


cnunciato:
Now that this is out, we should at least pull the script that runs in this Markdown file into the JS bundle somehow and treat it as a proper component of the website if we're going to support it going forward.

PR #4031, opened 2020-08-19
Properly set default values and allow overrides on converter pages

zchase:
Fixes: https://github.com/pulumi/docs/issues/4029

There was weird behavior going on the *2pulumi pages that caused in some cases the input text area to disappear or somethings the output would be gone. This PR changes the way we initialize the code converter and adds an optional parameter (though it is JavaScript, so I suppose everything is optional 😄 ). 

This change defaults the input to always be the "code" option (which was I understood is the expected behavior from what I was reading in the code). 

The main issue I saw is the way we were are handling state applications with some of the logic being in the stencil component and other parts of the logic being outside of the stencil component in pure JS. The issue is that calls to update state from the in the pure JS are not exactly safe. For example, when creating the component we call `setCurrentInputKind` in the pure JS, then call `convertCode` which in turn calls `getCurrentInputKind`. There is no guarantee that `setCurrentInputKind` has completed by the time we call `getCurrentInputKind` and that is what I am guessing was leading to unexpected behavior. The bug was difficult to reproduce so that makes sense it was some sort of race condition we were seeing with state updates.

We should consider moving pages like this to be its own microsite written in React or something similar. We could apply some vanity URL like `kube2.pulumi.com` as well. The way this page is written today will make it tough to debug issues and ultimately make changes to how the app functions.

### DIFF
--- a/content/migrate/tf2pulumi.md
+++ b/content/migrate/tf2pulumi.md
@@ -233,7 +233,7 @@ function displayCouldNotConvert(language) {
 }
 
 // Now set up our event handler for conversion.
-function convertCode(language) {
+function convertCode(language, inputKind) {
     // If we got called without an explicit language, look it up.
     language = language || getCurrentLanguage();
 
@@ -261,7 +261,7 @@ function convertCode(language) {
     }
 
     // Read the input kind and verify that we've got what we need.
-    let tfIk = getCurrentInputKind();
+    let tfIk = inputKind || getCurrentInputKind();
     switch (tfIk) {
     case "url":
         if (tfUrl === "") {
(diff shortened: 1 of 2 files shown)


cnunciato:
> We should consider moving pages like this to be its own microsite written in React or something similar. We could apply some vanity URL like kube2.pulumi.com as well. The way this page is written today will make it tough to debug issues and ultimately make changes to how the app functions.

Indeed, see #4016 and [this comment](https://github.com/pulumi/docs/pull/3776#pullrequestreview-452790600), and I'll be chatting with @EvanBoyle today as well about taking this over. I do think it's probably possible to get to a middle ground where the UI lives here and consumes some infra from another stack, but in general, agree it needs revisiting.

zchase:
@cnunciato I have no strong opinion between where it lives, just that the logic gets cleaned up. Happy to help contribute here and we should find some time to discuss how we are going to architect `stencil` components. Just get on the same page :)

PR #4032, opened 2020-08-19
Eliminate "real languages" from our website

joeduffy:
We know that this terminology is unfriendly and can be seen as
gatekeeping. We've moved away from it over time, but there are many
traces of it in our documentation -- most notably, even on our
homepage. Instead of "real languages", let's use "familiar languages",
and in places where deeper explanation makes sense in context, we
can say "familiar, general-purpose languages."

The current 2nd-level blurb on the homepage isn't very good right now
anyhow, so this is a good, timely change. As we make progress on various
marketing efforts, we will likely make more extensive changes, but
for now this at least fixes this problem. We recently A/B tested
updates without much of an impact, so let's move ahead with this change
and then if we want to test alternatives to "familiar", we can do that.
For instance, we considered "popular", etc.


content/blog/program-kubernetes-with-11-cloud-native-pulumi-pearls/index.md
```diff
@@ -29,7 +29,7 @@ around a little:
 <!-- just links to this same post -->
 **[Config as Real Code](#config-as-real-code)**
```
zchase, line 30:
```suggestion
**[Config as Code](#config-as-real-code)**
```

PR #4036, opened 2020-08-19
Tag new buckets with Vanta tags

leezen:
I've set myself as the owner as Vanta expects an actual user

### DIFF
--- a/scripts/sync-and-test-bucket.sh
+++ b/scripts/sync-and-test-bucket.sh
@@ -49,8 +49,10 @@ aws_region="$(pulumi -C infrastructure config get 'aws:region')"
 # fail, causing this script to exit nonzero. In either case, it's okay to continue.
 aws s3 mb $destination_bucket_uri --region $aws_region || true
 
-# Tag the bucket with ownership information
-aws s3api put-bucket-tagging --bucket $destination_bucket_uri --tagging file://bucket-tagging.json
+# Tag the bucket with ownership information for production buckets.
+if [ "$(pulumi stack --show-name)" == "production" ]; then
+    aws s3api put-bucket-tagging --bucket $destination_bucket_uri --tagging file://$(pwd)/scripts/bucket-tagging.json
+fi
 
 # Make the bucket an S3 website.
 aws s3 website $destination_bucket_uri --index-document index.html --error-document 404.html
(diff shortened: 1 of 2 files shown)


scripts/sync-and-test-bucket.sh
```diff
@@ -49,6 +49,9 @@ aws_region="$(pulumi -C infrastructure config get 'aws:region')"
 # fail, causing this script to exit nonzero. In either case, it's okay to continue.
 aws s3 mb $destination_bucket_uri --region $aws_region || true
 
+# Tag the bucket with ownership information
```
cnunciato, line 52: Super-ridiculous complete-sentence nit, sorry -- just something I picked up from others on the team and have come to believe in.  😄
 
```suggestion
# Tag the bucket with ownership information.
```

scripts/sync-and-test-bucket.sh
```diff
@@ -49,6 +49,9 @@ aws_region="$(pulumi -C infrastructure config get 'aws:region')"
...
+aws s3api put-bucket-tagging --bucket $destination_bucket_uri --tagging file://bucket-tagging.json
```
cnunciato, line 53: I think this should be relative to the project root:

```suggestion
aws s3api put-bucket-tagging --bucket $destination_bucket_uri --tagging "file://$(pwd)/scripts/bucket-tagging.json"
```

leezen:
Thanks for the suggestions. Yes, you're right that this should only apply to the production stack. These tags would be harmless for a dev stack, but also pretty unnecessary and if someone outside of the org wanted to try it, would be pretty nonsensical, so let me throw in a check for that.

scripts/sync-and-test-bucket.sh
leezen: Good idea on just having an absolute path

cnunciato:
🤔 Shouldn't `pulumi-docs-origin-pr-4036-c6dc43ad` have been tagged? It doesn't look like it was.

I think maybe [this line](https://github.com/pulumi/docs/pull/4036/files#diff-eb8fc496386a9f1c6146a288a305c02aR53) didn't work -- should it maybe be "pulumi/production"?

scripts/sync-and-test-bucket.sh
```diff
@@ -49,6 +49,11 @@ aws_region="$(pulumi -C infrastructure config get 'aws:region')"
 # fail, causing this script to exit nonzero. In either case, it's okay to continue.
 aws s3 mb $destination_bucket_uri --region $aws_region || true
 
+# Tag the bucket with ownership information for production buckets.
+if [ "$(pulumi stack --show-name)" == "production" ]; then
```
cnunciato, line 53: Looking at the log, I think this should be:

```suggestion
if [ "$(pulumi -C infrastructure stack --show-name)" == "production" ]; then
```

leezen:
Ah... I missed that last comment. You are totally right.

leezen:
Fixing in https://github.com/pulumi/docs/pull/4045

PR #4062, opened 2020-08-24
Pulumi.HCloud only has `RunAsync` method

francoishill:
### Proposed changes

The example C# code does not compile because nuget `Pulumi.HCloud` only contains a `Deployment.RunAsync` method instead of a `Deployment.Run` method.


stack72 (commented):
Thanks so much! I will ensure I get the rest of the providers fixed up as well

stack72 (approved):

PR #4072, opened 2020-08-25
Update StackReference go example

pennywisdom:
### Proposed changes

The example for go is different to the other examples and so makes the following section of the documentation a little confusing.


cnunciato:
Thanks for the submission, @pennywisdom!

@EvanBoyle, what do you think?

stack72 (approved):

PR #4080, opened 2020-08-26
Update scripts and GHA workflow file to allow for private builds.

praneetloke:
I have already added the two new env vars that the scripts as GHA secrets in this repo's settings.


.github/workflows/build-and-deploy.yml
```diff
@@ -53,6 +53,7 @@ jobs:
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.CI_AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.CI_AWS_SECRET_ACCESS_KEY }}
+          AWS_SESSION_TOKEN: ${{ secrets.CI_AWS_SESSION_TOKEN }}
```
praneetloke, line 56: This would be used only if using SSO creds. For all other cases, it should be ignored (we'll find out if the PR build fails, I guess!)

infrastructure/index.ts
```diff
@@ -336,3 +336,4 @@ async function createAliasRecord(
 export const originBucketWebsiteDomain = originBucket.websiteDomain;
 export const originBucketWebsiteEndpoint = originBucket.websiteEndpoint;
 export const cloudFrontDomain = cdn.domainName;
+export const cloudFrontDistributionUrn = cdn.urn;
```
praneetloke, line 339: I added this stack output because we are currently hard-coding the distribution URN [here](https://github.com/pulumi/docs/blob/master/scripts/run-pulumi.sh#L26). Once the stack output is available, we could replace the hard-coded value with this.

scripts/sync-and-test-bucket.sh
```diff
@@ -40,7 +40,8 @@ destination_bucket_uri="s3://${destination_bucket}"
 node scripts/translate-redirects.js "$build_dir" "$(pulumi -C infrastructure config get redirectDomain 2>/dev/null || echo '')"
 
 # Read the region from the stack's config -- we use it below.
-aws_region="$(pulumi -C infrastructure config get 'aws:region')"
+aws_region="$(pulumi -C infrastructure config get 'aws:region')" || aws_region="us-west-2"
```
praneetloke, line 43: This is so that PR previews can still be deployed without an actual stack being created, which is where we currently get the region from.

praneetloke:
@cnunciato Ping.

cnunciato:
Didn't get to it today, sorry. I'll take a look soon as I can.

scripts/sync-and-test-bucket.sh
cnunciato: Might suggest a small tweak to define the variable once:

```suggestion
aws_region="$(pulumi -C infrastructure config get 'aws:region' 2>/dev/null || echo 'us-west-2')"
```

.. but more importantly, why would someone want to do this in practice?

.github/workflows/build-and-deploy.yml
cnunciato:
```suggestion
```

> This would be used only if using SSO creds

Including this without a comment suggests it's required, but it isn't. Since we don't currently have a need for this, I'd suggest leaving it out.

infrastructure/index.ts
cnunciato: The CDN is protected, so should never change, but even if it did, this approach should still hold. Makes sense to me. 👍

scripts/ci-pull-request.sh
```diff
@@ -15,5 +15,16 @@ source ./scripts/ci-login.sh
...
+  echo "Publishing Sentry release..."
+  ./scripts/publish-sentry-release.sh
+else
+  echo "Sentry auth token not found. Will skip publishing release to Sentry."
+fi
+
+if [ -n "${PULUMI_STACK_NAME:-}" ]; then
+  ./scripts/run-pulumi.sh preview
+else
+  echo "PULUMI_STACK_NAME is not set. Will skip running a preview."
```
cnunciato, line 29: Again, I'm not sure we ever want to do this, since a merge would mean running an update, and the absence of a Pulumi preview result is an easy thing to miss. (This PR, for example, makes an infra change whose preview was skipped because PULUMI_STACK_NAME was not set.)

.github/workflows/build-and-deploy.yml
```diff
@@ -70,6 +71,7 @@ jobs:
           path: origin-bucket-metadata.json
 
       - name: Archive Sentry release
+        if: ${{ github.repository == 'pulumi/docs' }}
```
cnunciato, line 74:
```suggestion
```

I think we can just leave this line out, since there's no real reason _not_ to create build artifacts for these files but continue to do so for the others.

scripts/ci-login.sh
```diff
@@ -2,9 +2,15 @@
...
-eval $(assume-role "arn:aws:iam::058607598222:role/ContinuousIntegrationRole")
+if [ -n "${CI_ASSUME_ROLE_ARN:-}" ]; then
+  echo "Assuming the production CI role..."
+  eval "$(assume-role "${CI_ASSUME_ROLE_ARN}")"
+fi
 
-echo "Selecting the pulumi/production stack"
-pulumi login
-pulumi -C infrastructure stack select pulumi/production
+if [ -n "${PULUMI_STACK_NAME:-}" ]; then
```
cnunciato, line 10: Where is this being set?

scripts/ci-push.sh
```diff
@@ -6,6 +6,17 @@ source ./scripts/ci-login.sh
...
-./scripts/make-s3-redirects.sh
+
+if [ -n "${SENTRY_AUTH_TOKEN:-}" ]; then
+  echo "Publishing Sentry release..."
+  ./scripts/publish-sentry-release.sh
+else
+  echo "Sentry auth token not found. Will skip publishing release to Sentry."
+fi
+
+if [ -n "${PULUMI_STACK_NAME:-}" ]; then
```
cnunciato, line 17: We don't seem to be setting this anywhere. (I see it's defined as a secret, but not exported -- although really, shouldn't it be defined as a simple environment variable that's captured in code?)

scripts/ci-pull-request.sh
```diff
@@ -15,5 +15,16 @@ source ./scripts/ci-login.sh
 
 ./scripts/build-site.sh preview
 ./scripts/sync-and-test-bucket.sh preview
-./scripts/publish-sentry-release.sh
-./scripts/run-pulumi.sh preview
+
+if [ -n "${SENTRY_AUTH_TOKEN:-}" ]; then
```
cnunciato, line 19: One thing to note about the intent here -- I deliberately publish to Sentry in order to capture script errors that show up in PR previews (and also post to Slack). So to the extent the aim here is to support "real" previews that could actually be merged, I'd think we'd want to come up with some way to preserve this behavior.

praneetloke:
> But I do wonder whether the thing to do here isn't instead just to create a single private mirror of the docs repo that anyone can submit PRs for and that properly runs against pulumi/production, just as today.

Yes, we could do that. The changes in this PR doesn't prescribe one way or another. The PR simply externalizes some values that were hard-coded in the scripts. Whether we use a single private mirror (my preference) or not is something we can decide afterwards. But either way, these changes just makes our scripts general for use with any stack.

scripts/ci-pull-request.sh
praneetloke: Sure that's fine. If we go the route of having a single private mirror, we could set the token and have them be published still.

scripts/ci-push.sh
praneetloke: Yeah it's set as a GitHub Secret. But I can move it to the workflow file itself since it is not really a secret. But doing so would mean that if I wanted to fork this repo, I'd have to modify the workflow file, rather than set the GHA Secret.

.github/workflows/build-and-deploy.yml
praneetloke: This will be needed if we are not generating a Sentry artifact, then there is no artifact to publish. This step failed for me when I didn't have a Sentry token.

scripts/ci-pull-request.sh
praneetloke: > This PR, for example, makes an infra change whose preview was skipped because PULUMI_STACK_NAME was not set.

Yep. I just realized that I never mapped the `PULUMI_STACK_NAME` to the build step. I just pushed an update.

.github/workflows/build-and-deploy.yml
praneetloke: To add to what I said above, if we choose to go down the path of having a private mirror, whenever we decide that, we can update this to include that repo. Since we will want to be deliberate about using the same Sentry token elsewhere.

scripts/sync-and-test-bucket.sh
praneetloke: > but more importantly, why would someone want to do this in practice?

I could fork this repo and not create a stack and still get my PR preview builds working.

.github/workflows/build-and-deploy.yml
praneetloke: Or I'll add a comment that it is optional. :)

praneetloke:
Ahh a rebase with `master` is giving me problems. Specifically, the resource docs folder `gcp/serviceAccount` was deleted and now there is a `gcp/serviceaccount`, which is throwing off git. I'll close and open a new PR.

PR #4119, opened 2020-08-31
Fix the role policy invalid reference

runlevel5:
Line 33: `assumeRolePolicy: instance_assume_role_policy.then(instance_assume_role_policy => instance_assume_role_policy.json),`  was giving me invalid referencing error. This PR updates the variable naming to reflect the correct reference


stack72 (approved):
Thank you!

PR #4138, opened 2020-08-31
Add a banner component

cnunciato:
Fixes https://github.com/pulumi/marketing/issues/13.

### DIFF
--- a/assets/config/tailwind.config.js
+++ b/assets/config/tailwind.config.js
@@ -71,6 +71,18 @@ const green = {
     900: "#155148",
 }
 
+const fuschia= {
+    100: "#fcdbff",
+    200: "#f5b9fd",
+    300: "#eb9efa",
+    400: "#d975f5",
+    500: "#c75ceb",
+    600: "#ab3ad4",
+    700: "#9135b6",
+    800: "#6c2f89",
+    900: brand.purple,
+};
+
 module.exports = {
     theme: {
         extend: {
@@ -99,6 +111,7 @@ module.exports = {
             blue,
             orange,
             green,
+            fuschia,
             transparent,
         },
         maxHeight: {
--- a/components/src/components/banner/banner.spec.ts
+++ b/components/src/components/banner/banner.spec.ts
@@ -1,0 +1,7 @@
+import { Banner } from './banner';
+
+describe("pulumi-banner", () => {
+    it("builds", () => {
+        expect(new Banner()).toBeTruthy();
+    });
+});
--- a/components/src/store/reducers/banners.ts
+++ b/components/src/store/reducers/banners.ts
@@ -1,0 +1,21 @@
+import { BannersAction, TypeKeys } from "../actions/index";
+import { BannersState } from "../../store/state";
+
+// Define the default set of banners.
+const getInitialState = (): BannersState => {
+    return {
+        dismissed: [],
+    };
+};
+
+// The banners reducer. Note that it does not mutate state, but rather always returns a fresh copy.
+// https://redux.js.org/basics/reducers/
+export const banners = (currentState = getInitialState(), action: BannersAction): BannersState => {
+
+    switch (action.type) {
+        case TypeKeys.DISMISS_BANNER:
+            return { ...currentState, dismissed: [ ...currentState.dismissed, action.name ] };
+        default:
+            return currentState;
+    }
+};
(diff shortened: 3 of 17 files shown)


assets/config/tailwind.config.js
```diff
@@ -71,6 +71,18 @@ const green = {
     900: "#155148",
 }
 
+const fuschia= {
```
cnunciato, line 74: I ended up not actually using these in this PR, but they're part of the palette, so I chose to leave 'em in, for future use.

components/src/components/banner/banner.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { Banner } from './banner';
+
+describe("pulumi-banner", () => {
+    it("builds", () => {
+        expect(new Banner()).toBeTruthy();
+    });
+
+    it("is not visible", () => {
+
```
cnunciato, line 9: I'll fill these out.

components/src/store/reducers/banners.ts
```diff
@@ -0,0 +1,22 @@
...
+    };
+};
+
+// The banners reducer. Note that it does not mutate state, but rather always returns a fresh copy.
+// https://redux.js.org/basics/reducers/
+export const banners = (currentState = getInitialState(), action: BannersAction): BannersState => {
+
+    switch (action.type) {
+        case TypeKeys.DISMISS_BANNER:
+            return { ...currentState, dismissed: [ action.name ] }; // TODO: Make sure we append to this.
```
cnunciato, line 18: Oops. Looks like I still need to do this. 😇

components/src/store/reducers/banners.ts
```diff
@@ -0,0 +1,21 @@
+import { BannersAction, TypeKeys } from "../actions/index";
```
zchase, line 1: It looks like with this approach we could end up with an infinite amount of names stored if I am understanding correctly. I think we will want to avoid something like this

layouts/partials/banner.html
```diff
@@ -0,0 +1,11 @@
+{{/* When a banner is dismissed, its `name` property is recorded in localStorage to
+prevent it from being shown again, so be sure to name new banners uniquely to prevent them
+from being suppressed. */}}
+<pulumi-banner name="cloud-engineering-summit" dismissible>
+    <div class="content">
+        <span class="summary">Learn from the experts at the Cloud Engineering Summit, October 7&ndash;8.</span>
+        <span class="cta">
+            <a href="https://cloudengineeringsummit.com/" target="_blank" rel="noopener noreferrer">Register for free</a>
```
zchase, line 8: We are going to want to add UTM parameters here since we `cloudengineeringsummit.com` is an external domain of ours. Something like `https://cloudengineeringsummit.com/?utm_campaign=summit&utm_source=pulumi.com&utm_medium=top-banner`

It is important to note that we will only do this for external links if we are linking internally (pulumi.com/) then we will not use UTM parameters to avoid overwriting attribution data.

cnunciato: Cool. Do you want to add them here, then, so they'll be correct when we merge?

cnunciato: cc @isaac-pulumi

components/src/store/reducers/banners.ts
cnunciato: Correct, but this is based on the practical and hopefully reasonable assumption that we aren't going to be pummeling users with constantly running, constantly changing banners for all eternity. 😄 If it's really a concern, though, we can just date them and clean them up periodically. I can add the code to do that now or come back to it later. To me it'd be fine to do that later.

cnunciato:
> My recommendation would be to use a cookie to track banner dismissal. For example, we could have a cookie, ex. `pb` (pulumi banner), that's value is set to the name/id of the banner when the banner is dismissed. This way when the component mounts we can check if the cookie exists and if it does then we won't show the banner. We can eliminate the need to plumb this stuff through the State and localStorage.

This is effectively exactly how this works now; the only difference is the storage mechanism. One reason cookies are not a good fit here is that this sort of data does not need to be sent to the server on every request. Local storage is actually quite suitable for this sort of thing, and using the store as an abstraction over localStorage is definitely by design.

layouts/partials/banner.html
zchase, line 8:
```suggestion
            <a href="https://cloudengineeringsummit.com/?utm_campaign=summit&utm_source=pulumi.com&utm_medium=top-banner" target="_blank" rel="noopener noreferrer">Register for free</a>
```

PR #4172, opened 2020-09-04
Combine webinar and event pages

zchase:
Fixes: https://github.com/pulumi/docs/issues/3961

This PR combines event and webinar pages into a new `/resources` page. I've moved the code that generates the list items to be stencil.js components and reorganized the content sections. Unfortunately, this PR will break the CMS which for the next few weeks will mean I'll have to add/edit/remove things from resources. This will unblock the reminder of the Cloud Engineering Summit as once this is merged in, I'll add the recordings with transcripts as well as a new resource type for "Podcasts" which will use to list the Summit content as a Podcast on Spotify, Apple Podcasts, etc.


assets/sass/styles.scss
```diff
@@ -236,4 +236,10 @@ main {
     @apply text-yellow-700 py-2 px-4 rounded border border-yellow-600;
 }
 
+// To avoid some jumpiness of elements on the page we set the min-height
+// of the event list conatiner
+.pulumi-event-list-container {
+    min-height: 600px;
```
cnunciato, line 242: You can actually get rid of the container and just apply this styling directly on the `pulumi-event-list` element.

components/src/components/event-list/event-list.scss
```diff
@@ -0,0 +1,41 @@
+:host {
```
cnunciato, line 1: Pretty much all of this styling needs to go into assets/sass/components. We want to use the Tailwind styles and colors for these as much as possible, but keep the Tailwind references out of the components themselves, and that approach (while admittedly a bit odd, given the state of the world today) lets us accomplish both.

layouts/resources/list.html
```diff
@@ -0,0 +1,51 @@
...
+    <!-- Conver the event data to JSON -->
+    {{ $eventListEvents := slice }}
+    {{ range $listed_events }}
+        {{ if ne .Params.title "Upcoming Events" }}
+            {{ $eventListEvents = $eventListEvents | append (.Params | jsonify)}}
+        {{ end }}
+    {{ end }}
+
+    <div class="pulumi-event-list-container mt-5">
+        <pulumi-event-list
```
cnunciato, line 42: The thing about this approach that we need to be super careful about is that because these items are all rendered with JS now, we're at risk of adversely affecting the crawlability of this page. Today, for example, /webinars renders a list of links in the source of the page; we do a bit of fiddling with the ordering of things, but the source markup is baked into what we deliver to the requestor. But here, we no longer have that. I suspect we'll want to make sure the "view source" content we end up delivering somehow exposes `<a href="/link-to-thing">thing name</a>` in a bot-friendly way, unless we're absolutely sure the bots will execute our JS. That might be the case, but it's something I'd definitely want to doublecheck before proceeding with this as written.

components/src/components/event-list/event-list.tsx
```diff
@@ -0,0 +1,193 @@
...
+                <i class={`fas fa-${args.icon} text-4xl`}></i>
+                <p class="hidden sm:block m-0 mt-3">{args.text}</p>
+            </li>
+        );
+    }
+
+    private renderEventFilter() {
+        // The event filter data.
+        const items: EventFilterItem[] = [
+            { text: "All", icon: "asterisk", key: "all" },
```
cnunciato, line 144: It'd be nice to be able to link directly to these filtered views.

components/src/components/event-list/event-list.tsx
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        const pulumiEvent: PulumiEvent = {
+            archetype: "webinar",
+            title: webinar.title,
+            description: webinar.meta_desc,
+            startDate: new Date(webinar.main.sortable_date),
+            registrationUrl: webinar.url_slug,
+            external: webinar.external,
+            type: types,
```
cnunciato, line 75: So `type` is an Array? Why not call it `types` also?

components/src/components/event-list/event-list.tsx
```diff
@@ -0,0 +1,193 @@
...
+                    );
+                })}
+            </ul>
+        );
+    }
+
+    render() {
+        // Filter the events based on the currently selected filter.
+        const filteredEvents = this.eventFilterChoice === "all" ? this.pulumiEvents : this.pulumiEvents.filter((event) => {
+            return event.type.indexOf(this.eventFilterChoice) > -1;
```
cnunciato, line 179: I think if `type` were renamed to the plural (since it looks like it could be multiple types?), this could be simplified a bit as `event.types.includes(this.eventFilterChoice)`. Perhaps more, but it took me a bit to figure out what was happening here based on the naming.

components/src/components/event-list/event-list.tsx
```diff
@@ -0,0 +1,193 @@
...
+            imageSrc: webinar.preview_image,
+        };
+
+        result.push(pulumiEvent);
+    }
+
+    return result;
+}
+
+@Component({
```
cnunciato, line 85: I suspect this component is doing way more than it should. When you think about what this component is for, it's basically this:

* Expose a list of things to filter by
* Handle clicks and filter accordingly

All it needs to be able to do that is show those filters and handle their clicks, and to do _that,_ all it needs to do is set a CSS class on itself that says what the selected thing is, and then CSS can take over by hiding (`display: none`, or Tailwind `hidden`) all of the list items that shouldn't be visible. I _think_ that's about it. The rest is Hugo-specific implementation stuff that doesn't need to go into either of these components really. 

So I think you could make this much simpler by just having the Hugo template use each webinar/event's `.Params` metadata to figure out what to pass into each `pulumi-event-list-item` to ensure each item gets a class that corresponds with the kind of item it is, and scaling the list component back to deal only with filtering. I can imagine the API looking something like this, then:

```
<pulumi-event-list selected="all">
    <pulumi-event-list-item title="Something" description="Something" type="webinar" date="2020-10-31" href="/resources/blah"></pulumi-event-list-item>
    <pulumi-event-list-item title="Something else" description="Something else" type="event" date="2020-10-31" href="/resources/blah"></pulumi-event-list-item>
</pulumi-event-list>
```

cnunciato (commented):
Some initial feedback to keep you unblocked. Looking good though!

layouts/resources/list.html
cnunciato: I realized this morning that in this light, it's also worth asking: since the relevant thing here is just to be able to filter (which really means hide, and which can easily be done with CSS), why do the items themselves need to be components at all? If you had just the filtering controls as a component, and that component were able to set a prop or a class on itself, wouldn't that be enough? I think it would. 🤔 

```
<pulumi-event-filter selected="all"></filter>
<ol class="pulumi-event-items">
    <li class="webinar"></li>
    <li class="event"></li>
    ...
</ol>
```

```
pulumi-event-filter {

    // Adjacent sibling selects the list itself; all items are hidden/invisible by default.
    + .pulumi-event-items {
        li {
            @apply hidden;
        }
    }

    // Items matching what's selected are visible.
    &[selected="webinar"] + .pulumi-event-items li.webinar,
    &[selected="event"] + .pulumi-event-items li.event,
    ... {
        display: block;
    }
}
```

zchase:
New PR here: https://github.com/pulumi/docs/pull/4530. Closing this.

PR #4189, opened 2020-09-10
Update docs with TLS for migrations.

sean1588:
This is an update to the self-hosted docs with instructions on how to enable TLS for database migrations. I just added it under the Database Connections section as part of the Service doc for now. I felt this was the most logical place to put this rather than create a whole new page just for the migrations.

### DIFF
--- a/content/docs/guides/self-hosted/api.md
+++ b/content/docs/guides/self-hosted/api.md
@@ -100,7 +100,7 @@ The service is configurable to enable connections to the backend SQL database ov
 
 ##### Migrations
 
-The database migrations are configurable to enable connections to the database over TLS. If you are intending to connect to the database over TLS to run the migrations the following environment variable must be set in order to enable this. If this variable is set connections to the database will be establised over TLS, otherwise connections will default to not using TLS.
+The database migrations container is configurable to enable connections to the database over TLS. To use TLS, the following environment variable must be set. The default is to not use TLS.
 
 | Variable Name            | Description                                                                                                   |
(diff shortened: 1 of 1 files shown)


content/docs/guides/self-hosted/api.md
```diff
@@ -89,9 +89,19 @@ openssl \
...
 
 | Variable Name            | Description                                                                                                   |
 |--------------------------|---------------------------------------------------------------------------------------------------------------|
 | DATABASE_CA_CERTIFICATE  | The CA certificate used to establish TLS connections with the database. This certificate must be PEM encoded. This must be set to the value of the certificate itself and _not_ a filepath to the location of the certificate file. |
 | DATABASE_MIN_TLS_VERSION | The minimum TLS version to use for database connections (must be in \<major>.\<minor> format, e.g. `1.2`).    |
+
+##### Migrations
+
+The database migrations are configurable to enable connections to the database over TLS. If you are intending to connect to the database over TLS to run the migrations the following environment variable must be set in order to enable this. If this variable is set connections to the database will be establised over TLS, otherwise connections will default to not using TLS.
```
praneetloke, line 103:
```suggestion
The database migrations container is configurable to enable connections to the database over TLS. To use TLS, the following environment variable must be set. The default is to not use TLS.
```

sean1588: I like it. I think its a lot more concise and reads easier. Thank you.

praneetloke (approved):

PR #4192, opened 2020-09-10
Clean up website buckets daily

cnunciato:
This change adds a script to be run once daily that deletes all website buckets that can be safely deleted.

I've been running this script manually for several weeks now, and it's behaved as expected, so I feel comfortable having it run automatically now.

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -119,6 +119,7 @@ set_bucket_for_commit() {
 # https://docs.github.com/en/rest/reference/repos#list-pull-requests-associated-with-a-commit
 get_pr_for_commit() {
     curl -s \
+         -f \
          -H "Accept: application/vnd.github.groot-preview+json" \
          "https://api.github.com/repos/pulumi/docs/commits/$1/pulls" || echo ""
 }
@@ -128,7 +129,7 @@ get_pr_for_commit() {
 # suffix to filter by (e.g., "pr" or "push").
 get_recent_buckets() {
     aws s3api list-buckets \
-        --query "reverse(sort_by(Buckets,&CreationDate))[:50].{id:Name,date:CreationDate}|[?starts_with(id,'$(origin_bucket_prefix)-${1}')]" \
+        --query "reverse(sort_by(Buckets,&CreationDate))[:100].{id:Name,date:CreationDate}|[?starts_with(id,'$(origin_bucket_prefix)-${1}')]" \
         --output json | jq -r '.[].id'
 }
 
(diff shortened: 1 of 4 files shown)


scripts/common.sh
```diff
@@ -119,6 +119,7 @@ set_bucket_for_commit() {
 # https://docs.github.com/en/rest/reference/repos#list-pull-requests-associated-with-a-commit
 get_pr_for_commit() {
     curl -s \
+         -f \
```
cnunciato, line 122: Ensures curl exits nonzero for HTTP statuses outside the OK range.

scripts/list-recent-buckets.sh
```diff
@@ -71,7 +71,7 @@ deletables=()
 for bucket in $buckets; do
     maybe_echo
     maybe_echo "Fetching metadata for ${bucket}..."
-    metadata="$(aws s3 cp "s3://${bucket}/metadata.json" - || echo '')"
+    metadata="$(aws s3 cp "s3://${bucket}/metadata.json" 2>/dev/null - || echo '')"
```
cnunciato, line 74: Suppresses error messages from output (to keep the output from this one clean).

leezen (commented):
The changes in this PR make sense, but need to be reviewed in the context of the existing functionality of `list-recent-buckets.sh` so most of my questions are around that file.

I wonder if it would make sense to add an extra guard on https://github.com/pulumi/docs/blob/5dd5ecb6487adf28fee65aa9071a452bc5833045/scripts/list-recent-buckets.sh#L103 to check that the bucket was also synced at _least_ some time ago from the current time (e.g. two weeks or something)?

cnunciato:
> to check that the bucket was also synced at least some time ago

I thought about having a time-based check first, but then figured it'd be less consistent -- e.g., you can imagine over long holidays or quiet weeks not having many changes, and having things drift beyond that threshold, causing you to cleaning up more than you might want to, or merging many commits within a short time, causing things pile up. Neither of those seemed great, so I went with "some number behind current", with the assumption that what's current is good, and what's older than current only matters insofar as it might be something you might want to roll back to. Given that all releases are tested, and allowing a bit of squishy margin for things the tests might not catch, "more than 10 successful builds old" seemed like a safe threshold to start with.

How might an additional time-based threshold (presumably on top of what's in place now) provide additional safety? Totally open to it of course; I'm just not seeing the kind of scenario that it'd address.

leezen:
> If you mean something like, "more than 10 behind current and more than two weeks old", that could be fine, but again I'd be slightly concerned about things piling up: a busy week or two of releases could generate many dozens of buckets, and our current limit is 500. (Today, prior to some cleanup, we were at ~250. The max, I think, is 1000.)

Correct. I was thinking both as a way to ensure that a dozen small fixes wouldn't wipe out what's available for rollback. Though, thinking about it more, we can always trigger a build off a previous commit to rollback. So 10 is probably "good enough."

leezen (approved):

PR #4196, opened 2020-09-11
Blog post: Learn Pulumi: Inputs and Outputs

spara:


cnunciato (commented):
A bit of initial feedback for ya. 😄

content/blog/learn-pulumi-inputs-outputs/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+class LoadBalancer {
+  public readonly address: Output<string>;
+  public readonly port: Output<string>;
+
+  constructor(name: string; argos: LoadBalancerArgs, …);
+}
+```
+
+The address can't be known in advance and will only be available once the load balancer is created. However, the input port argument is the same as the output port, making it available during previews.
```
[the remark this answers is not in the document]
spara: The address is available once the load balancer is created. However, the input port argument is the same as the output port, which is known during previews.

content/blog/learn-pulumi-inputs-outputs/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+![Load balancer example](i_o-example.png)
+
+The diagram shows the dependency. The load balancer knows that it depends on the virtual machine because the target points at the virtual machines IP address. Be aware that when you're running `pulumi up`, you may encounter outputs. In this case, we're creating
+the virtual machine and load balancer, and if we click details, we'll see the target is listed as an output of string. That's because, during the preview, its value isn't known. In this case, the port is known as 80 because it’s a concrete value, but it’s possible that the port could come from another resource's output, or it was computed. In that case, it might be shown as an output of type number.
+
+Let's look at some code. The first thing to notice is that the code is wrapped in an `async ()` function. The reason for this is that we create a Virtual Private Cloud (VPC) which defines a virtual network for our application.  Resources, such as the virtual machines' public subnets used to communicate to the load balancer, aren’t known in advance. We create a security group to control egress from the VPC.
```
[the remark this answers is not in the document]
spara: I know, but there isn't a fully worked example that's simple. The post is based on Joe's video and he uses pseudo code to clarify but I want a working example that someone can copy. I know it's a lot but I think a working example is more important.

content/blog/learn-pulumi-inputs-outputs/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+For example, we have a load balancer. It takes a target in a port as inputs, and its outputs include an IP address and a port.
+
+```typescript
+interface LoadBalancerArgs {
+  target: Input<string>;
+  Port: Input<number>;
+}
+
+class LoadBalancer {
```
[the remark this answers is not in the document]
spara: I've reframed this as a pseudo code class for a load balancer and reworked the explanation.

content/blog/learn-pulumi-inputs-outputs/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+The diagram shows how this works before we examine the code. Assume that we create a virtual machine that has an IP address as its output.  Next, we'll create a load balancer that has a target and a port as inputs. We'll pass the virtual machine's IP address as the target because we want the load balancer to point at it, and then we'll set 80 as the input port. The resulting load balancer's outputs will then be the generated hostname and the port 80.
+
+![Load balancer example](i_o-example.png)
+
+The diagram shows the dependency. The load balancer knows that it depends on the virtual machine because the target points at the virtual machines IP address. Be aware that when you're running `pulumi up`, you may encounter outputs. In this case, we're creating
+the virtual machine and load balancer, and if we click details, we'll see the target is listed as an output of string. That's because, during the preview, its value isn't known. In this case, the port is known as 80 because it’s a concrete value, but it’s possible that the port could come from another resource's output, or it was computed. In that case, it might be shown as an output of type number.
```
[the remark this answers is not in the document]
spara: Done

content/blog/learn-pulumi-inputs-outputs/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+```typescript
+    // Creates an ALB associated with the default VPC for this region and listen on port 80.
+    const alb = new awsx.lb.ApplicationLoadBalancer("web-traffic", { securityGroups: [ sg ] });
+    const listener = alb.createListener("web-listener", { port: 80 });
+    const publicIps: pulumi.Output<string>[] = [];
+    const subnets = await vpc.publicSubnets;
+```
+
+It's important to remember you can't access the values of outputs directly. They might not be known, so for example, printing the value to the console returns an error instead. The most common way of interacting with an output is to pass it to another resource as an input.
```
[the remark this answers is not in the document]
spara: The most common way of interacting with an output is to pass it to another resource as an input. But, it's important to remember you can't access the values of outputs directly. They might not be known, so for example, printing the value to the console returns an error instead.

content/blog/learn-pulumi-inputs-outputs/index.md
```diff
@@ -0,0 +1,160 @@
...
+
+The diagram shows the dependency. The load balancer knows that it depends on the virtual machine because the target points at the virtual machine's IP address. Be aware that when you're running `pulumi up`, you may encounter outputs. In this case, we're creating
+the virtual machine and load balancer, and if we examine details in the preview, we'll see the target is listed as an output of string. That's because, during the preview, its value isn't known. In this case, the port is known as 80 because it’s a concrete value, but it’s possible that the port could come from another resource's output, or it was computed. In that case, it might be shown as an output of type number.
+
+Let's look at some code. This worked example uses AWS resources and presents a slightly more complex situation. However, the way inputs and outputs are used are the same as the basic example.
```
cnunciato, line 50:
```suggestion
Let's look at some code. This example uses AWS resources and presents a slightly more complex situation. However, the way inputs and outputs are used are the same as the basic example.
```

Also, which basic example?

content/blog/learn-pulumi-inputs-outputs/index.md
```diff
@@ -0,0 +1,160 @@
...
+
+The first thing to notice is that the code is wrapped in an `async ()` function. The reason for this is that we create a Virtual Private Cloud (VPC) which defines a virtual network for our application.  In place of an IP address as a target, the load balancer uses the VPC's public subnets to communicate to the load balancer, aren’t known in advance. We also create a security group to control egress from the VPC.
```
cnunciato, line 52: > In place of an IP address as a target, the load balancer uses the VPC's public subnets to communicate to the load balancer, aren’t known in advance.

This sentence doesn't seem right.

content/blog/learn-pulumi-inputs-outputs/index.md
```diff
@@ -0,0 +1,160 @@
...
+    const vpc = awsx.ec2.Vpc.getDefault();
+
+    // Create a security group to open ingress to our load balancer on port 80, and egress out of the VPC.
+    const sg = new awsx.ec2.SecurityGroup("web-sg", {
+        vpc,
+        egress: [{ protocol: "-1", fromPort: 0, toPort: 0, cidrBlocks: [ "0.0.0.0/0" ] }],
+    });
+```
+
+In this part of the code, the application load balancer is created, and port 80 is assigned to the load balancer's listener. Note that we create `publicIps` as an Output typed as a string array, which will hold the virtual machines' public IP addresses. We also call `await` to return the public subnets as an output, because we need that information before creating the virtual machines.
```
cnunciato, line 73: If we have a load balancer, why do we need the VMs to have public IPs? Why do we even need more than one VM? 

I still think this example is significantly more complicated than it needs to be to illustrate these concepts; the inputs and outputs are getting lost in all the infrastructure swirling around here. (Sorry to be a pain about it, but I do believe it.)

content/blog/learn-pulumi-inputs-outputs/index.md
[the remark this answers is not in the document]
cnunciato: > so for example, printing the value to the console returns an error instead

Where is the error? Where are we printing to the console?

content/blog/learn-pulumi-inputs-outputs/index.md
```diff
@@ -0,0 +1,160 @@
...
+        publicIps.push(vm.publicIp);
+        alb.attachTarget("target-" + i, vm);
+    }
+
+    // Export the resulting URL so that it's easy to access.
+    return { endpoint: listener.endpoint.hostname, publicIPs: publicIps  };
+};
+```
+
+As the diagram earlier depicted, another thing we can easily do with outputs is export them. When we export an output, Pulumi captures that value and makes it easily available on the stack. This includes at the command line, in the pulumi console, and even if you want one stack to depend on another stack's output. You can see it in action with the following script:
```
cnunciato, line 117:
```suggestion
As the diagram earlier depicted, another thing we can easily do with outputs is export them. When we export an output, Pulumi captures that value and makes it easily available on the stack. This includes at the command line, in the Pulumi Console console, and even if you want one stack to depend on another stack's output. You can see it in action with the following script:
```

I don't see that the diagram depicts exported outputs. Maybe the diagram "suggests"? And maybe we could point out _where_ in the code the export is? (Because it's fairly hard to find, actually.)

content/blog/learn-pulumi-inputs-outputs/index.md
```diff
@@ -0,0 +1,160 @@
...
+
+```bash
+for i in {1..25}; do curl http://$(pulumi stack output endpoint); done
+```
+
+Let's look at other ways of interacting with the concrete value of an output. In this example, we want to parse out the subdomain of the load balancer's hostname. Usually, you would expect a string, and we can call functions like substring and indexOf directly, but in this case, we can't because it's an Output. This is where we can use the apply function. Pulumi invokes a callback when it knows the concrete value and apply takes the callback and returns another output that we can export.
```
cnunciato, line 123:
```suggestion
Let's look at other ways of interacting with the concrete value of an output. In this example, we want to parse out the subdomain of the load balancer's hostname. Usually, you would expect a string, and we can call functions like `substring()` and `indexOf()` directly, but in this case, we can't because it's an Output. This is where we can use the `apply` function. Pulumi invokes a callback when it knows the concrete value, and `apply` takes the callback and returns another output that we can export.
```

We should link to this (to apply) in the docs.

content/blog/learn-pulumi-inputs-outputs/index.md
```diff
@@ -0,0 +1,160 @@
...
+export const upperHostname = alb.hostname.toUpperCase();
+```
+
+In another case, we may want to use concrete values from multiple outputs. We could do this by nesting multiple applys, but Pulumi provides `all` as a convenience function that can take multiple outputs and return a single output.
+
+```typescript
+export const url = pulumi.all([alb.hostname, alb.port]).apply( ( [hostname, port]) =>  `http://${hostname}:${port}` );
+```
+
+String concatenation is so commonplace that Pulumi offers a helper function for it. The `interpolate` function allows you to reference outputs without calling apply or all on any of
```
cnunciato, line 145:
```suggestion
String concatenation is so commonplace that Pulumi offers a helper function for it. The `interpolate` function allows you to reference outputs without calling apply or all on any of them.
```

content/blog/learn-pulumi-inputs-outputs/index.md
```diff
@@ -0,0 +1,160 @@
...
+them.
```
cnunciato, line 146:
```suggestion
```

content/blog/learn-pulumi-inputs-outputs/index.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+This article is a brief introduction to Pulumi inputs and outputs. Learn more about the Pulumi [programming model]({{< relref "/docs/intro/concepts/programming-model" >}}) by visiting Pulumi Docs.
+
+You can also watch the *Pulumi Help: Resource Inputs and Outputs (Node.js* on [PulumiTV](https://www.youtube.com/c/PulumiTV/).
+
+{{< youtube "lybOxul2otM" >}}
+
+{{% notes type="info" %}}
+Learn Pulumi is an irregular series about the Pulumi programming model.
```
cnunciato, line 159: I'd probably just drop this note, but if you want to keep it, I'd suggest using a word other than "irregular" -- it reads here as "weird/odd". 😄 Maybe "new" series, if that's the intent?

content/blog/learn-pulumi-inputs-outputs/index.md
```diff
@@ -0,0 +1,160 @@
...
+for i in {1..25}; do curl http://$(pulumi stack output endpoint); done
+```
+
+Let's look at other ways of interacting with the concrete value of an output. In this example, we want to parse out the subdomain of the load balancer's hostname. Usually, you would expect a string, and we can call functions like substring and indexOf directly, but in this case, we can't because it's an Output. This is where we can use the apply function. Pulumi invokes a callback when it knows the concrete value and apply takes the callback and returns another output that we can export.
+
+```typescript
+export const subdomain = alb.hostname.apply(hostname => hostname.substring(0, indexOf(".")));
+```
+
+You can use apply to print to the console or allocate resources inside of an apply, but it can be dangerous because the resource may not be created during a preview and apply will never execute.
```
cnunciato, line 129:
```suggestion
You can use `apply()` to print to the console, or allocate resources inside of an apply, but it can be dangerous because the resource may not be created during a preview and apply will never execute.
```

Mixed feelings about even mentioning that "you can allocate resources inside of an apply". Assuming this is even still possible (I seem to remember a warning and suggestion we'd eventually throw), it's definitely not recommended, so if we say anything, it should probably just be to advise against it.

content/blog/learn-pulumi-inputs-outputs/index.md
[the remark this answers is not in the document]
cnunciato: Why a pseudo-class and not just actual code? I'm not understanding how it's better to use code that's abstract versus code that's more like what's in, say, most of our examples (since that'll be much more common and actually map to the code you'll actually write).

cnunciato (commented):
Still some open questions & suggestions for me.

spara:
Did a rewrite, closing this PR to open one for the rewrite

PR #4204, opened 2020-09-14
Do cleanup as a separate Actions job

cnunciato:
Given how frequently the link-checker fails (often for legitimate reasons), this change breaks the bucket-cleanup scripts into their own scheduled workflow.


sean1588 (approved):
Looks good to me.

:shipit:

PR #4223, opened 2020-09-15
fix: /serverless REST API example

CanRau:


cnunciato:
Nice catch, thanks for the fix, @CanRau!

cnunciato (approved):

PR #4233, opened 2020-09-16
Move *2pulumi JS into its own file and load JS after jQuery is loaded

zchase:
This PR moves the shared JS code between the *2pulumi markdown files into its own JS file. We were still seeing the bug where the code input would disappear in production and I was still unable to reproduce locally. After investigation, I realized we were loading the JS for the *2pulumi pages before we loaded the jQuery script, so I assume the odd behavior we were seeing was due in part to a race condition of the jQuery script loading. I also added a function to wait for elements to exist, because there were cases where the stencil component wouldn't be loaded in time to register some of the click handlers we try and attach outside of the stencil code.

This will hopefully solve that bug and let us promote these types of pages.


assets/js/2pulumi.js
```diff
@@ -0,0 +1,676 @@
...
+        if (element) {
+            return callback();
+        } else {
+            return waitForElementToExists(selector, callback);
+        }
+    }, 500);
+}
+
+$(function() {
+    waitForElementToExists("pulumi-chooser[type='language'] > ul > li > a", function() {
```
cnunciato, line 565:
```suggestion
    waitForElementToExist("pulumi-chooser[type='language'] > ul > li > a", function() {
```

assets/js/2pulumi.js
```diff
@@ -0,0 +1,676 @@
...
+            setCurrentInputKind("url");
+            convertCode(getCurrentLanguage() || "typescript", "url");
+        } else {
+            if (tfCode) {
+                $("#terraform-code").val(tfCode);
+            } else {
+                loadCannedExample();
+            }
+            setCurrentInputKind("code");
+            console.log(getCurrentLanguage());
```
cnunciato, line 590:
```suggestion
```

cnunciato (approved):
Couple teeny nits, but LGTM.

PR #4235, opened 2020-09-16
Expose Pulumi access token to the cleanup workflow

cnunciato:
The `ci-login` script (which is sourced by `ci-bucket-cleanup`, which is now scheduled to run daily), needs this variable on the environment as well.


cnunciato:
Going to merge this in order to get it into use today, but will address whatever feedback is submitted afterward. Thanks!

PR #4258, opened 2020-09-18
Expire banner dismissals, run component tests on PRs

cnunciato:
This change adds a bit of logic to expire the dismissal of a banner after a period of four days. (Meaning after four days, the banner will reappear.) By virtue of this chance, all users will see the current banner again as well.

The change also adds a script that runs the Stencil tests on pull requests.

### DIFF
--- a/components/src/store/index.ts
+++ b/components/src/store/index.ts
@@ -26,11 +26,11 @@ export const configureStore = () => {
         console.error("Failed to read pulumi_state from localStorage:", e);
     }
 
-    const persistedState: Partial<AppState> = local ? JSON.parse(local) : {};
+    const persistedState: any = local ? JSON.parse(local) : {};
 
     const store = createStore(
         rootReducer,
-        persistedState,
+        normalizeState(persistedState),
         composeWithDevTools(applyMiddleware(thunk))
     );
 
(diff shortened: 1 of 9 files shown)


components/src/store/index.ts
```diff
@@ -26,11 +26,11 @@ export const configureStore = () => {
         console.error("Failed to read pulumi_state from localStorage:", e);
     }
 
-    const persistedState: Partial<AppState> = local ? JSON.parse(local) : {};
+    const persistedState: any = local ? JSON.parse(local) : {};
```
zchase, line 29: I'm not a huge fan of using `any` but knowing you, I'm sure is a good reason to have to use it here.

components/src/store/index.ts
```diff
@@ -49,3 +49,40 @@ export const configureStore = () => {
 
     return store;
 }
+
+// normalizeState transforms slices of serialized state into a shape that conforms to
+// our current expectations.
+export function normalizeState(persistedState: any): Partial<AppState> {
```
zchase, line 55: Nice! Maybe this will help out with the odd state issue we are hitting on *2pulumi pages

zchase (approved):
LGTM 🚀

components/src/store/index.ts
cnunciato: Yeah, that's because this is literally "whatever's in localStorage", so we can't rely on that being of any particular shape at the point of deserialization, since you can throw just about anything in there. 😄 Good spot, though.

components/src/store/index.ts
cnunciato: I was thinking the same! We'll see. I'm somewhat surprised to see it not being set explicitly today, and TBH I haven't had time yet to investigate. But this should indeed ensure there's always at least something in that slot.

PR #4285, opened 2020-09-21
Add arm2pulumi coming soon page

zchase:
This PR adds a `/arm2pulumi` page with a coming soon. A press release made its way into the world with a link to this page before we shipped the functional version. This is a temporary page to resolve any 404s people might hit.


cnunciato (approved):
LGTM. Can you add a note To the PR description explaining the need for this change, for historical reference?

PR #4293, opened 2020-09-23
inputs outputs blog post

spara:
This is a rewrite of the first version which was based on the video text. It has been rewritten to follow the example more closely instead of the video,


cnunciato:
Should we close this, or maybe move to pulumi-hugo?

spara:
not published

PR #4298, opened 2020-09-24
Run the broken-link-checker programmatically

cnunciato:
Running this programmatically gives us quite a bit more flexibility than we get with the command-line interface. The main goal here is to make sure we only notify ourselves when something really is broken, and with actionable info.

### DIFF
--- a/scripts/check-links.js
+++ b/scripts/check-links.js
@@ -1,0 +1,309 @@
+const { SiteChecker, HtmlUrlChecker } = require("broken-link-checker");
+const { WebClient, LogLevel } = require('@slack/web-api');
+const httpServer = require("http-server");
+const path = require("path");
+const fs = require("fs");
+
+/**
+ *  This script uses the programmatic API of https://github.com/stevenvachon/broken-link-checker
+    to check the links (including images, iframes, and client-side redirects) for either an individual page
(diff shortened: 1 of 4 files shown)


scripts/check-links.js
```diff
@@ -0,0 +1,309 @@
...
+        throw new Error("Finished, but failed with errors. See the log for details.");
+    }
+}
+
+/**
+    We exclude some links:
+    - Our generated API docs have lots of broken links.
+    - Our available versions page includes links to private repos.
+    - GitHub Edit Links may be broken, because the page might not yet exist!
+    - Our LinkedIn page, for some reason, returns an HTTP error (despite being valid).
```
zchase, line 216: LinkedIn probably does this to prevent scraping, or at least that would be my guess as to why it fails.

zchase (approved):
LGTM 🚢

scripts/check-links.js
cnunciato: Yeah, that'd make sense. This rule goes back a ways, but if you [Google around for HTTP 999](https://www.google.com/search?q="HTTP"+"999"+"linkedin") (the status code returned for every request we make to a linkedin.com URL), it's clear LinkedIn would prefer not to be crawled. 🤷

PR #4320, opened 2020-09-29
Adding console references to getting started.

sean1588:
Added some console screenshots and a small explanation to the deploy stack step of the getting started. Also I used the same screenshot for all the providers, e.g. aws, gcp, azure. I figure it is more of an example of what the console looks like to get an idea of what you will see if you go in there. If this is a big deal, then I can find a way to construct separate images with the output respective to each cloud. 
http://pulumi-docs-origin-pr-4320-344948bd.s3-website.us-west-2.amazonaws.com/docs/get-started/aws/deploy-stack/


zchase (approved):
LGTM --- Thanks for doing this! Excited to see what kind of lift we can get out of calling out the console in the guide! 🍾

content/docs/get-started/aws/deploy-stack.md
```diff
@@ -97,6 +97,10 @@ $ pulumi stack output BucketName
 
 {{% /choosable %}}
 
-Running that command will print out the name of your bucket. Now that your bucket has been provisioned, let's modify the bucket to host a static website.
+Running that command will print out the name of your bucket. If you are using a Pulumi managed backend, you can follow the Permalink displayed at the end of the CLI output. This will open the update in the Pulumi [Console](https://app.pulumi.com), where you can view the output and explore detailed information about your stack such as its activity, resources and configuration.
```
lukehoban, line 100:
```suggestion
Running that command will print out the name of your bucket. If you are using the Pulumi Service backend, you can follow the Permalink displayed at the end of the CLI output. This will open the update in the Pulumi [Console](https://app.pulumi.com), where you can view the output and explore detailed information about your stack such as its activity, resources and configuration.
```

Also:
1. Should we put this in a "note" section instead of just in the paragraph?  I feel like that might both make it stand out more and also be more clear that this is sort of "optional".
2. I think the phrasing here is not true relative to the new way the `pulumi` CLI outputs this?  We now say "View Live" instead of "Permalink" and it's at the beginning instead of end.
3. Do we want to link out on "Pulumi Service backend" to the documentation about the backend?  (Not sure we have something "great" to link to - which is another suggestion to improve!)

sean1588: > I think the phrasing here is not true relative to the new way the pulumi CLI outputs this? We now say "View Live" instead of "Permalink" and it's at the beginning instead of end.

yes very true.

> Do we want to link out on "Pulumi Service backend" to the documentation about the backend? (Not sure we have something "great" to link to - which is another suggestion to improve!)

I think that is a good idea. This is the only [page](https://www.pulumi.com/docs/intro/console/) I can think of to link, but I don't think it does a great job of showcasing the console in an organized fashion. I think we may need a console page that is more marketing oriented that has that captures more visuals, i.e. screenshots,  whereas this one seems to be more documentation oriented - unless there is one that exists already, but I am just not aware of it.

Issue #4325, opened 2020-09-30
Updated Pulumi State and Backends versus self-hosted s3 backend.


tusharshahrs:
### Update benefits of Pulumi Hosted Backend vs self-managed backend

   Evan had a great answer to this in slack:

- I saw a product documentation issue on the following slack post: <URL>
     https://pulumi-community.slack.com/archives/C019YSXN04B/p1601212645020300?thread_ts=1601061499.018300&cid=C019YSXN04B

- I saw an issue on the following blog post: <URL>

- I had trouble finding the information that I needed: <URL>
   The difference between S3 backend and Pulumi Backend

- Other: <DETAILS>

### Suggestions for a fix
Update this page: https://www.pulumi.com/docs/intro/concepts/state/

There's a few key differences between S3 and the Pulumi hosted backend:
1. S3 has no concurrency control. This is a big deal for your use case as you run the risk of automated deployments colliding and potentially corrupting state/making simultaneous edits to your infra. If you want to use the S3 backend, you'll have to implement locking around your deployments to guard against this. The pulumi backend has locking and concurrency control built in.
2. The pulumi backend is designed for software teams, and has features like stack access contorl, RBAC, built-in concurrency control, etc.
3. The pulumi backend has some interesting state storage capabilities that aren't in the other backends. You can refresh config from the last deployment, meaning you don't have to worry about storing it yourself. We have created a process that queries that pulumi backend for all ephemeral stacks matching time based criteria. This just kinda works and doesn't require us to maintain any state ourselves.  Not just state as in the state of resources, but it also makes information available about who owns a stack, who last modified it, what time that happened, the full audit log of events, etc. It will work either way, but the S3 backend will require more manual effort on your part.

tusharshahrs:
There are some other changes that are in the works that will change this.  We will close this for now.

PR #4332, opened 2020-10-02
Update Pulumi Managed Backend and Self-managed backend

tusharshahrs:
### Proposed changes

Updated the benefits of Pulumi backend vs self hosted backend to make it easier for customers/sales to make a decision.
Added additional plus points to Pulumi managed back end
Added risks that customers need to address for self-managed backend

### Related issues (optional)

 Related to: https://github.com/pulumi/docs/issues/4325
 Based on following community slack [discussion](https://pulumi-community.slack.com/archives/C019YSXN04B/p1601212645020300?thread_ts=1601061499.018300&cid=C019YSXN04B)

### DIFF
--- a/content/docs/intro/concepts/state.md
+++ b/content/docs/intro/concepts/state.md
@@ -99,6 +99,18 @@ At a glance, the Pulumi Service backend provides the following benefits:
 - Full deployment history for auditing and rollback purposes
 - Concurrent state locking to prevent corrupting your infrastructure state in a team environment
 - Advanced policy and RBAC (Role Based Access Control)
+- Stack access control, including information on who owns a stack, who last modified stack, and what time the stack was modified
+- Refresh config from the last deployment 
+- Secrets handled by a key managed by the service and option to use passpharse-based key
+
+
+### Self-managed backend risks:
+
+At a glance, the self-hosted backend requires that the following issues are addressed:
+
+- Implement your own state management including backup, sharing, and team access synchronization
+- Automated deployments colliding and potentially corrupting state/making simultaneous edits to your infrastructure because there is no concurrency control. 
+- Must use passpharse-based secrets provider
 
 #### app.pulumi.com architecture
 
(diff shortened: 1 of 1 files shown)


content/docs/intro/concepts/state.md
```diff
@@ -99,6 +99,17 @@ At a glance, the Pulumi Service backend provides the following benefits:
...
+- Refresh config from the last deployment
+- Secrets handled by a key managed by the service or the option to use passpharse-based key
+
+### Self-managed backend risks
+
+At a glance, the self-hosted backend requires that the following issues are addressed:
+
+- Implement your own state management including backup, sharing, and team access synchronization
+- Automated deployments colliding and potentially corrupting state/making simultaneous edits to your infrastructure because there is no concurrency control
+- Secrets handled by a key that must use the passphrase-based secrets provider
```
stack72, line 112: This isn't true - we can use any of the non-SaaS based secret providers right?

content/docs/intro/concepts/state.md
```diff
@@ -99,6 +99,17 @@ At a glance, the Pulumi Service backend provides the following benefits:
 - Full deployment history for auditing and rollback purposes
 - Concurrent state locking to prevent corrupting your infrastructure state in a team environment
 - Advanced policy and RBAC (Role Based Access Control)
+- Stack access control, including information on who owns a stack, who last modified stack, and what time the stack was modified
+- Refresh config from the last deployment
+- Secrets handled by a key managed by the service or the option to use passpharse-based key
+
+### Self-managed backend risks
```
stack72, line 106: I am not a fan of using the word "risks" here - they are not risks they just requirement that we need to ensure people understand the differences

lukehoban (commented):
I had written up a much longer set of notes - but looks like they got lost.  

High level - I think unfortunately a lot of what's added here isn't correct, and we'll want to make sure that we rewrite to get the details right.  I'd also like to make an effort to reframe in the positive instead of negative - highlighting with helpful clarity what users would need to do to use the self-managed backends effectively (even if that does mean doing a fair bit more heavy lifting).  @stack72 Think you could help @tusharshahrs on the details and framing here?

clstokes:
I think there are some other changes planned to this documentation, so I propose we close this PR and wait for those. @tusharshahrs, if there are some smaller changes you want to get in here, let's chat about those.

PR #4352, opened 2020-10-07
Suppress position locations, apply styles for consistency

cnunciato:
This change just applies some styling to suppress the locations currently shown by the BambooHR widget, since they seem to be giving applicants the wrong impression, and adds a note to make it clear we're currently all-remote. Also applies a bit of styling to make things look more like the rest of the site. (Copy courtesy of Liz.)

image


cnunciato:
Argh -- PostCSS! 🤛  Adding an exception...

freddy-hernandez (approved):

PR #4356, opened 2020-10-08
Update summit CTAs on home page

zchase:
This PR changes the CTA links and text on the home page now that the Summit content is live.


justinvp (approved):

cnunciato (approved):
This is good, but if you wanted to leave things in a slightly better state here, you could actually delete all of the conditional logic in the alert partial, since the home page is the only place we use it.

PR #4384, opened 2020-10-15
Document MacPorts install option

lhaeger:
### Proposed changes

Document MacPorts install option

### DIFF
--- a/content/docs/get-started/install/_index.md
+++ b/content/docs/get-started/install/_index.md
@@ -43,6 +43,22 @@ Subsequent updates can be installed in the usual way:
 $ brew upgrade pulumi
 ```
 
+### MacPorts
+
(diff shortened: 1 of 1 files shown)


cnunciato:
@davidwrede Would you like to pull this in? If so, we'll need to move it to https://github.com/pulumi/pulumi-hugo.

davidwrede:
@cnunciato Assuming this is an officially supported install option, then yeah, we should move it to pulumi-hugo. I'll try and do that before the end of the week.

leezen:
@davidwrede It's a community-supported installation mechanism. It's been up to date on our versions since this PR was submitted, so I think we should bring this in.

davidwrede:
Thanks for the contribution, @lhaeger!

This change is now tracked over in https://github.com/pulumi/pulumi-hugo/pull/33.

PR #4394, opened 2020-10-17
Revert back to the usual CTAs

joeduffy:
This goes back to the mascot and standard get started CTAs in the
hero, keeping the summit in the alert as we are still working on
adding a cloud engineering page that it will link to.


cnunciato (approved):
Thanks!

joeduffy:
I'm holding off on merging for a couple days, I'm running an A/B test to better understand a few differences between the standard CTA and the Cloud Engineering Summit flavors.

PR #4395, opened 2020-10-17
Change banner button label

cnunciato:
"Save your spot" feels less right, now that the event's passed. This just changes the button label to "Watch now".


layouts/partials/banner.html
```diff
@@ -1,7 +1,7 @@
 {{ $name := "cloud-engineering-summit" }}
-{{ $summary := "The Cloud Engineering Summit replay is available! Watch all your favorite talks, on demand." }}
+{{ $summary := "The Cloud Engineering Summit replay is available! See all your favorite talks, on demand." }}
```
cnunciato, line 2: Just to avoid using the word "watch" twice.

zchase (approved):
LGTM

PR #4406, opened 2020-10-20
gcloud doesn't support multiple roles

angelo-moreira:
### Proposed changes

Run the gcloud command twice to assign multiple roles, the gcloud command doesn't support adding multiple roles at once.

### Related issues (optional)

https://github.com/cloudfoundry/bosh-google-cpi-release/issues/142


metral (approved):
Great catch and thank you for the corrections! 🎉 

LGTM

PR #4409, opened 2020-10-20
Set a temporary timeout for Turnstyle

cnunciato:
This change sets a timeout that instructs [Turnstyle](https://github.com/softprops/turnstyle) to stop waiting after 30 minutes and proceed with a build. The timeout is intentionally short for now, as we've got several builds queued up that need to go out; I'll follow this up with another that sets it closer to the default GHA limit of 6 hours.


zchase (approved):
LGTM

PR #4410, opened 2020-10-21
Replace Turnstyle with a custom script

cnunciato:
This change replaces Turnstyle with a simple script that does essentially the same thing, but logs a bit more and gives us additional control and debuggability. The inline comments explain how it works, but the gist is that it waits for any currently running workflow of the same type on the same branch to complete before exiting (allowing the workflow to continue). Applies only to the default branch, as before.

### DIFF
--- a/scripts/await-in-progress.js
+++ b/scripts/await-in-progress.js
@@ -1,0 +1,43 @@
+const { Octokit } = require("@octokit/rest");
+
+// Wait for any in-progress runs of the same workflow on this branch to complete before
+// proceeding. In other words, if the current workflow is an instance of the "foo"
+// workflow, and there's another "foo" workflow running for a different commit on the same
+// branch as this one, wait for that workflow to complete before exiting (in order to
+// prevent the current workflow from continuing).
+// Inspired by https://github.com/softprops/turnstyle.
+async function waitForInProgressRuns() {
+    const githubToken = process.env.GITHUB_TOKEN;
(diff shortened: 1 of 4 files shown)


scripts/await-in-progress.js
```diff
@@ -0,0 +1,43 @@
...
+
+    if (recent.length > 0) {
+        const [ mostRecent ] = recent;
+        console.log(`⏱  Found ${recent.length} other ${workflowName} job(s) running on branch ${branch}.`);
+        console.log(`   Waiting for ${mostRecent.html_url} to complete before continuing...`);
+        await Promise.resolve(setTimeout(waitForInProgressRuns, 60000)); // One minute.
+    }
+}
+
+waitForInProgressRuns();
```
zchase, line 43: What do you think of adding a `.then(...).catch(...)` here? That way we can log when the waiting has completed and also catch any errors that could occur in the script.

zchase (approved):
LGTM 🎉  Just one question on error handling

scripts/await-in-progress.js
cnunciato: Ah totally, we could to that -- but in this case, if the script fails, I want to fail the whole build (as that would mean something's wrong -- workflow not defined, various things we expect to "just be there"), so I'm deliberately not catching errors so they cause Node to exit non-zero and downstream jobs to be skipped.

PR #4412, opened 2020-10-21
Run the await script after dependencies are installed

cnunciato:
This change runs the await script after dependencies are installed (rather than before, which makes it go :boom:, but still before logging in) and logs to the console on each execution (even those that pass, so we can always see what values were used to make that determination).


zchase (approved):
🚢

PR #4413, opened 2020-10-22
Exit nonzero explicitly on error

cnunciato:
Errors in promises actually don't cause the process to exit nonzero after all! TIL. This change catches and exits 1 when something goes wrong (and corrects a thing that was).


zchase (approved):
LGTM. For my own curiosity, any reason why you went with `process.on` instead of just adding a `.catch()` to the method invocation?

cnunciato:
@zchase I actually did do that first, but then ended up going this route after reading that message and thinking there might be a more Node-idiomatic way to do this. Ultimately this felt slightly better in that (a) it wasn't forcibly exiting (but rather throwing and allowing Node to handle that Error as it would any other) and (b) it'd handle _any_ `unhandledRejection` program-wide, rather than just this one. (So if someone came along later and added another async function, the script would still exit `1` on error, rather than swallowing the rejection and exiting `0`). Make sense?

zchase:
@cnunciato Yup makes complete sense! And thanks for the explanation. I generally have been relying on `catch()` but `process.on` makes total sense to use. I'll probably start taking this same approach myself 🎉

PR #4415, opened 2020-10-22
Policy as Code for All Clouds

spara:
Per Cameron's request - PaC examples for AWS/Azure/GCP/Kubernetes


cnunciato:
Ah, looks like this was closed and a commit was made at around the same time, which resulted in one cleaned-up bucket and one newly created one. (Anytime a commit is made, we make a new bucket.) Looks like I have an edge case to cover!

spara:
I deleted the branch and pushed it again, I wasn't sure what was going on

PR #4417, opened 2020-10-22
Add command to only build marketing part of docs site

zchase:
As we are all aware the `docs` site is massive and it can be a pain to run locally. I know for myself some days the `make serve` command will crash every time I save, making the dev loop really frustrating. This PR adds a `make serve_marketing` command that will only serve the marketing part of the site (everything not in `/docs`) which greatly decreases the dev loop time. To enable this I had to create a `marketing-dev` environment and in Hugo that logs `relref` errors as WARNINGs instead of ERRORs that cause the process to exit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,12 @@ serve:
 	$(MAKE) lint
 	./scripts/serve.sh
 
+.PHONY: serve_marketing
+serve_marketing:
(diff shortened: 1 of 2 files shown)


cnunciato (approved):
Makes sense to me! Might be some way to make this more generally useful and maybe the output a little less noisy, but for now, if it makes life easier for you, let's roll with it! 👍

PR #4423, opened 2020-10-23
Fix PR bucket cleanups

cnunciato:
The daily job to clean up closed PR buckets (for those that happen to slip by the GitHub action that runs when a PR is closed, as happens occasionally) has been failing to detect deletable buckets when their commits aren't associated with any PR that GitHub knows about. This makes sense given that our default merge strategy for this repo is to squash and merge, which creates a _new_ commit, but what I didn't realize is that when that happens, GitHub "forgets" all of the commits that _were_ associated with the PR before it was squash-merged. 

So this change just plucks the PR number out of the bucket name, asks GitHub whether it's still open, and if not, adds it to the list of deletables. Also adds the GITHUB_TOKEN environment variable to handle API rate limits.

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -114,17 +114,7 @@ set_bucket_for_commit() {
         --overwrite
 }
 
-# Get the GitHub pull_request object associated with a particular commit.
-# Note that this GitHub API is still in preview:
-# https://docs.github.com/en/rest/reference/repos#list-pull-requests-associated-with-a-commit
-get_pr_for_commit() {
-    curl -s \
-         -f \
-         -H "Accept: application/vnd.github.groot-preview+json" \
-         "https://api.github.com/repos/pulumi/docs/commits/$1/pulls" || echo ""
-}
-
-# List the 50 most recent bucket in the current account, sorted descendingly by
+# List the 100 most recent bucket in the current account, sorted descendingly by
 # CreationDate, matching the prefix we use to name website buckets. Supports an optional
 # suffix to filter by (e.g., "pr" or "push").
 get_recent_buckets() {
--- a/scripts/list-recent-buckets.sh
+++ b/scripts/list-recent-buckets.sh
@@ -108,19 +108,25 @@ for bucket in $buckets; do
                 deletables+=($bucket_name)
             fi
         elif [ "$1" == "pr" ]; then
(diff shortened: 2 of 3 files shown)


scripts/common.sh
```diff
@@ -114,17 +114,7 @@ set_bucket_for_commit() {
...
-# https://docs.github.com/en/rest/reference/repos#list-pull-requests-associated-with-a-commit
-get_pr_for_commit() {
-    curl -s \
-         -f \
-         -H "Accept: application/vnd.github.groot-preview+json" \
-         "https://api.github.com/repos/pulumi/docs/commits/$1/pulls" || echo ""
-}
-
-# List the 50 most recent bucket in the current account, sorted descendingly by
+# List the 100 most recent bucket in the current account, sorted descendingly by
```
cnunciato, line 117: This is actually 100 -- I just noticed the comment was wrong.

scripts/list-recent-buckets.sh
```diff
@@ -108,19 +108,25 @@ for bucket in $buckets; do
...
 
-                if [ "$pr_state" == "closed" ]; then
-                    maybe_echo
-                    maybe_echo "❌ This bucket's PR has been closed (https://github.com/pulumi/docs/pull/${pr_number}), so it can safely be deleted."
-                    maybe_echo "   aws s3 rb s3://${bucket_name} --force"
+            pr_state="$(echo $pr_metadata | jq -r '.state')"
 
-                    deletables+=($bucket_name)
-                fi
+            if [ "$pr_state" != "open" ]; then
```
sean1588, line 124: Nit:  Just wondering if there was a reason you switched it from `"$pr_state" == "closed"` to `"$pr_state" != "open"`?
Are there other states other than closed we want to mark to be deleted? I don't have all the context here but just don't want us to unintentionally mark something to be deleted...

sean1588 (approved):
Just had one question I was curious about in the comments, other than that looks fine to me. :ship:

scripts/list-recent-buckets.sh
cnunciato: Good question. I believe the only valid states are open or closed, so it probably doesn’t make much difference, but as I was debugging this issue, I thought maybe there might be others, so I figured since the only buckets I’d want _not_ to delete are open ones, that felt slightly clearer to me. But there’s probably no practical difference either way, from what I can tell.

PR #4429, opened 2020-10-26
Removing Node 13 from the tfgen docs build script

stack72:


mikhailshilkov (approved):
LGTM if that fixes the issue

PR #4459, opened 2020-10-26
Document the PULUMI_CONSOLE_DOMAIN env var

### LINKED ISSUE #2995: Document PULUMI_CONSOLE_DOMAIN env var
Added to `pulumi` in https://github.com/pulumi/pulumi/pull/4410.

chrsmith:
### Proposed changes

Add `PULUMI_CONSOLE_DOMAIN` to the documentation for supported environment variables.

This has long-since shipped in the CLI, added in https://github.com/pulumi/pulumi/pull/4410. 😓 

### Related issues (optional)

Fixes #2995


chrsmith:
Semi-related: with the "suggest change" widget, this page doesn't look all that great. Has there been any discussion on if we want the Search/Feedback widget on the right to be collapsible or anything?

image

justinvp (approved):

cnunciato (approved):
> Has there been any discussion on if we want the Search/Feedback widget on the right to be collapsible or anything?

Not specifically, but this page could also be using the space a bit better than it is. If you want, I can push a commit either onto your branch or as another PR that adjusts the layout to improve that a bit, maybe something like:

image

chrsmith:
@cnunciato the idea of changing the layout to just have two columns, and put an example at the bottom of the description looks awesome. I'm inclined to not spend any more time on this, but should the opportunity arise to adjust that page, your suggestion is 🔥 .

PR #4476, opened 2020-10-26
Fix some broken links

cnunciato:
* Unlink Menta Network (per @isaac-pulumi)
* Update a broken link (looks like it was [moved in a recent update](https://github.com/kubernetes-sigs/aws-load-balancer-controller/commit/f0ba23fb23533502764308c0ce84ad9449140b2d))
* Add an exclusion for the Coinbase blog (it sets and then expects a cookie)


zchase (approved):
LGTM

PR #4480, opened 2020-10-26
Adjust the environment variables layout

cnunciato:
This just adjusts the layout of the [Environment variables](https://pulumi.com/docs/reference/cli/environment-variables/) page to make it a little easier to scan and use.

image

image

### DIFF
--- a/assets/sass/_lists.scss
+++ b/assets/sass/_lists.scss
@@ -13,6 +13,16 @@ dl {
             dd {
                 @apply w-2/3;
             }
+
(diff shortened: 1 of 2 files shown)


susanev (approved):
whoa this is much better!!

[nit] could we get rid of the gray background on the right column?
remove-gray-on-right-column

cnunciato:
☝️ Spoke with Susan out of band and we discussed maybe changing this global style at some point later.

PR #4488, opened 2020-10-28
removing the x's from the product pricing table

susanev:
### Proposed changes

Noticed the pricing plan details section uses both x's and checkmarks which can be a bit confusing, its also more difficult to quickly scan to see which features are included in each plan. This removes the x's all together leaving an empty space if that particular edition does not include that feature. Open to feedback if anyone has concerns or questions.


cnunciato (approved):
Like it, looks good to me! ⭐

PR #4503, opened 2020-10-31
Remove double 'is' from public subnet description

qasimalbaqali:
### Proposed changes

Type fix: Remove the double 'is' from the 'public' subnet description


stack72 (approved):

stack72:
Thanks so much for this!

PR #4505, opened 2020-11-01
Adjust the definition of security groups

qasimalbaqali:
### Proposed changes

Currently the definition of security groups in the documentation states that they are meant to act as a firewall to VPCs, but in reality they are made to control the traffic of instances as they are applied to an instance rather than a VPC. This might introduce confusion to new comers of AWS. The definition has been updated to reflect the definition that is provided by the [AWS documentation](SOURCE)


stack72 (approved):
Thanks so much for this!! This reads way better and on point with what it actually should

PR #4509, opened 2020-11-03
Add missing imports to ec2-webserver example

aarongorka:
### Proposed changes

The Typescript EC2 webserver example fails with
`index.ts(4,13): errorTS2304: Cannot find name 'pulumi'.` as it is at
the moment.

Importing `@pulumi/pulumi` fixes it, and follows the pattern of the
other examples (e.g
https://www.pulumi.com/docs/tutorials/aws/rest-api/#step-3-review-your-stack-resources).

I've also updated the Javascript example as I suspect it suffers from
the same error.


stack72 (approved):
Thank you so much for this!!

PR #4516, opened 2020-11-04
Blog post: Deploy Minecraft to Azure using nextgen provider

spara:


praneetloke (approved):
I just have one suggestion. Perhaps have the VM provisioning step under another heading? If you use an h3 (I think) it shows up in the side nav to the right.

PR #4519, opened 2020-11-04
Add webinars link

cnunciato:
This change adds a link to the Webinars page to the footer of the website.


isaac-pulumi (approved):
I think this is fine but we'll definitely need to update the top nav to make this truly discoverable.

PR #4530, opened 2020-11-06
Add resources page

zchase:
Continuation of https://github.com/pulumi/docs/pull/4172. I ended up creating a new PR to clean it up and hopefully make this whole thing easier to review. 

@cnunciato you left a comment on the original PR about using CSS more than JS and another comment about making the filters "linkable". So I killed two birds with one stone with this new approach of using hashes in the URL to create the functionality we are looking for. Interested in seeing what you think about this but this approach adds no new JavaScript so probably an overall better approach for a static site. 

To address why this is named `resources` it is because this will eventually contain more than just webinars, events, and videos. Eventually, we will also have ebooks, whitepapers, podcasts, and other pieces of content listed here. 

Old PR description:

Fixes: #3961

This PR combines event and webinar pages into a new /resources page. I've moved the code that generates the list items to be stencil.js components and reorganized the content sections. Unfortunately, this PR will break the CMS which for the next few weeks will mean I'll have to add/edit/remove things from resources. This will unblock the reminder of the Cloud Engineering Summit as once this is merged in, I'll add the recordings with transcripts as well as a new resource type for "Podcasts" which will use to list the Summit content as a Podcast on Spotify, Apple Podcasts, etc.


assets/sass/_event-list.scss
```diff
@@ -0,0 +1,57 @@
+.pulumi-event-list-container {
+    #all:target {
```
cnunciato, line 2: You could cut down this duplication _even more_ with a [Sass @each](https://sass-lang.com/documentation/at-rules/control/each):

```
.pulumi-event-list-container {
    @each $filter-name in "all", "featured", "upcoming", "videos", "pulumitv" {
        ##{$filter-name}:target {
            ~ .event-list-filter li[data-filter-name="#{$filter-name}"] a {
                @apply text-blue-500;
            }

            ~ ul li[data-filters~="#{$filter-name}"] {
                @apply block;
            }
        }
    }
}
```

layouts/resources/list.html
```diff
@@ -0,0 +1,192 @@
...
+        <!-- If the webinar is not marked as unlisted we will add it to the list. -->
+        {{ if ne .Params.unlisted true }}
+            {{ $resourcesWithDate = $resourcesWithDate | append (dict "date" .Params.main.start_date "data" .) }}
+        {{ end }}
+    {{ end }}
+
+    <div class="pulumi-event-list-container mt-5">
+        <div class="container mx-auto pt-10">
+            <!-- These hidden elements help select the children element so we use CSS -->
+            <!-- to the filter, well filter. -->
```
cnunciato, line 48: Like the comment, but seemed to be missing a word or two -- hopefully this gets at your meaning, but feel free to adjust/discard as you like!

```suggestion
            <!-- These hidden elements help select the children elements so we can use CSS -->
            <!-- to have the filter, well filter. -->
```

cnunciato (commented):
Love it. One little thing, though -- you'll probably want to adjust this so that _no_ anchor means "show everything", since linking to /resources now shows nothing. But otherwise, this is great. Well done.

cnunciato:
cc @susanev for design 👀 as well.

assets/sass/_event-list.scss
zchase: TIL :)

cnunciato:
One little thing is there's a gap between the icon and the label; ideally hovering over either would highlight both. You could do this by wrapping both in an `a` tag.

gap mov

susanev (commented):
looks good!! 

- im good with calling this resources for now, but if we move to the resources pricing model and we call those resources, we may want to come up with another name for this content

- noticed a responsive issue with the date
unresponsive-date

- we should probably write out the full-date (instead of just numbers) as the month-day can through off folks not from the united states

- i think we could maybe try having the entire card be interactive instead of just the heading

layouts/resources/list.html
zchase: 👍

susanev (approved):

assets/sass/_event-list.scss
```diff
@@ -0,0 +1,39 @@
...
+                @apply text-blue-500;
+            }
+
+            ~ h3[data-filter-title="#{$filter-name}"] {
+                @apply block;
+            }
+
+            // To support intially loading all the items on the page we first need
+            // to hide all the elements and then set the specific items to be visible.
+            ~ ul li {
```
cnunciato, line 24: You're in the loop here, so this rule will simply be repeated in the generated CSS. Since I think this means "hide all elements" -- i.e., there isn't any reference to `filter-name` here -- you can pull it out of the loop and position it before the loop to make it the default (unless I'm misunderstanding).

components/package.json
```diff
@@ -22,6 +22,7 @@
     "@stencil/core": "^1.8.8",
     "@stencil/redux": "^0.1.2",
     "@stencil/sass": "^1.1.1",
+    "@types/uuid": "^8.0.0",
```
cnunciato, line 25: We probably don't need this anymore. If so, we should remove it.

cnunciato (commented):
Couple little things, but otherwise looks good to go.

Nav looks off kilter -- should probably either be left- or center-justified. (My vote would be centered, since  the three columns makes the content appear centered, but either could work.)

image

On mobile, the justification skews more rightward, and the labels disappear:

image

Since the icons aren't self-explanatory, the labels could help explain what's different about these.

components/package.json
zchase: 👍

assets/sass/_event-list.scss
zchase: I'll try and make the comment clearer but this piece needs to be here because when the filter changes we need to first hide all the elements and then only show the ones we want. Because the filter triggers off hashes in the URL we need to handle the case where there is no hash in URL and in the case of no hash we show all the items, meaning that the elements are not rendered initially with a `hidden` class. Because there is no hidden class when the element is initially rendered when need to apply the hidden class when the filter changes.

TL;DR When the page initially renders all the elements are visible when the filter changes we need to hide the elements that are not part of the selected filter.

cnunciato (approved):
Pushed a commit with a few little mobile adjustments. MWR! 🚀

PR #4531, opened 2020-11-06
Add note clearly describing current limitations of `dependsOn`

tricky42:
### Proposed changes
Add note clearly describing current limitations of `dependsOn`
More details can be found in the following discussion thread on Slack: https://pulumi-community.slack.com/archives/C84L4E3N1/p1604604723483000

### Related issues (optional)
pulumi/pulumi-kubernetes#861
pulumi/pulumi-kubernetes#1315
pulumi/pulumi-kubernetes#1364
pulumi/pulumi#3282

### DIFF
--- a/content/docs/intro/concepts/programming-model.md
+++ b/content/docs/intro/concepts/programming-model.md
@@ -848,6 +848,8 @@ The `dependsOn` option provides a list of explicit resource dependency resources
 
 Pulumi automatically tracks dependencies between resources when you supply an input argument that came from another resource's [output properties](#outputs). In some cases, however, you may need to explicitly specify additional dependencies that Pulumi doesn't know about, but must respect. This might happen if a dependency is external to the infrastructure itself---such as an application dependency---or is implied due to an ordering or eventual consistency requirement. These dependencies ensure that resource creation, update, and deletion is done in the correct order.
 
+> **Note:** Currently `dependsOn` doesn't work if you depend on [Helm Charts](https://www.pulumi.com/docs/reference/pkg/kubernetes/helm/) or [ConfigFiles](https://www.pulumi.com/docs/reference/pkg/kubernetes/yaml/configfile). The issue is tracked in the following issues: [Helm](https://github.com/pulumi/pulumi-kubernetes/issues/861), [ConfigFile](https://github.com/pulumi/pulumi-kubernetes/issues/1315). 
+
 This example demonstrates making `res2` dependent on `res1`, even if there is no property-level dependency:
 
 {{< chooser language "javascript,typescript,python,go,csharp" >}}


content/docs/intro/concepts/programming-model.md
```diff
@@ -848,6 +848,8 @@ The `dependsOn` option provides a list of explicit resource dependency resources
 
 Pulumi automatically tracks dependencies between resources when you supply an input argument that came from another resource's [output properties](#outputs). In some cases, however, you may need to explicitly specify additional dependencies that Pulumi doesn't know about, but must respect. This might happen if a dependency is external to the infrastructure itself---such as an application dependency---or is implied due to an ordering or eventual consistency requirement. These dependencies ensure that resource creation, update, and deletion is done in the correct order.
 
+> **Note:** Currently `dependsOn` doesn't work if you depend on [Helm Charts](https://www.pulumi.com/docs/reference/pkg/kubernetes/helm/) or [ConfigFiles](https://www.pulumi.com/docs/reference/pkg/kubernetes/yaml/configfile). The issue is tracked in the following issues: [Helm](https://github.com/pulumi/pulumi-kubernetes/issues/861), [ConfigFile](https://github.com/pulumi/pulumi-kubernetes/issues/1315).
```
lblackstone, line 851: @lukehoban Curious to hear your thoughts on this change. The note is accurate, but I wasn't sure about the following:

1. We could expand it to clarify that the reason this doesn't work right is that resources are created inside an apply.
1. It might be better to document more locally on the API references for YAML/Helm rather than on this general-purpose page.

tricky42: Just from a user perspective I think it is important it is mentioned in the `dependsOn` section (at least that there are currently restrictions... Otherwise a user will find the `dependsOn` documentation when he or she looks for a solution to manually define dependencies and think this will work for **all** resources. Otherwise it would be necessary that the user looks up the general documentation on dependsOn and the specific resource he wants to depend on. 

In my case I somehow didn't thought it could be a bug as it seemed to me my usecase was pretty common and the documentation didn't mention any exception.

lukehoban: Yes - my suggestion would be to document this in the Helm docs, and then here phrase this in terms of what the real limitation is in the `dependsOn` implementation itself, with a note that Helm charts are an example that might be affected by that and a link to the Helm docs.  

I'll take care of the second part in the programming-model update.  @tricky42 or @lblackstone would you like to make the update in the Helm docs?

lblackstone: Opened https://github.com/pulumi/pulumi-kubernetes/issues/1773 to track the Helm docs.

lblackstone: I'm not sure if we ever updated the programming model docs. I skimmed through https://www.pulumi.com/docs/intro/concepts/resources/#dependson and didn't see anything about it.

leezen:
A lot has changed with `dependsOn` since this PR, especially in the Helm Chart area. Sorry we couldn't get this into a merged state and thank you for the suggestions.

PR #4550, opened 2020-11-09
Blog post: Automation API as a self-service cloud platform

spara:


content/blog/automation-api-as-platform/index.md
```diff
@@ -0,0 +1,275 @@
...
+meta_desc: "With Pulumi's Automation API, you can build a REST API for creating, reading, updating, and deleting (CRUD) cloud resources."
+meta_image: automation_api.png
+authors:
+    - sophia-parafina
+tags:
+    - Automation API
+    - CRUD
+---
+
+If you could create infrastructure without using a cloud provider's console, a CLI, or a templating engine, what would you build? Pulumi's Automation API lets you create declarative infrastructure defined by your best practices and expose it behind a REST, gRPC, or custom API.
```
EvanBoyle, line 13: > lets you create declarative infrastructure

Just a general callout that I think it's important that we continue to drive home the message that this is the same declarative IaC that gives you all of the predictability, robustness, safety, and desired state management. We're just exposing those declarative building blocks over a new programmatic surface area. 

You can create custom infrastructure behind a REST API just using the raw AWS REST API or language specific SDKs. But the end result is not reliable and requires a massive amount of work to handle edge cases, error handling, group infrastructure into units and give it identity, etc. 

The doc does a pretty good job of this, but just thought it was worth reiterating.

content/blog/automation-api-as-platform/index.md
```diff
@@ -0,0 +1,275 @@
...
+
+So just what is Automation API? Think of it as Pulumi's infrastructure as code engine as an SDK. Instead of writing code and using the CLI to declare infrastructure, you can directly tell the engine to build your infrastructure. Imagine building an application that creates infrastructure via a REST interface. Get ready, because that's what we're going to do.
+
+<!--more-->
+
+## Self-service Cloud Infrastructure
+
+Automation API is just another package that runs inside your favorite frameworks and works with other packages. You can create methods or functions to create infrastructure that you can call from other frameworks. One Automation API use case is self-service infrastructure. You can create declarative infrastructure and expose it behind a REST interface.
```
EvanBoyle, line 21: > Automation API is just another package that runs inside your favorite frameworks and works with other packages. You can create methods or functions to create infrastructure that you can call from other frameworks

It might be worth calling out that the end result is still a Pulumi `stack` backed by JSON state file managed by the Pulumi service. The benefit here is being able to create, and update them dynamically and on the fly.

content/blog/automation-api-as-platform/index.md
```diff
@@ -0,0 +1,275 @@
...
+
+This example demonstrates how to create infrastructure with Automation API and the [Express](https://expressjs.com/) Node.js  framework. To keep this example simple, we'll deploy the REST API locally. To deploy the code in the cloud, for example, in a virtual machine, we'll need to create an [instance profile](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html) to store AWS credentials. Deploying Automation API is beyond this article's scope, so we'll use `localhost` as the server's endpoint. We'll cover deploying Automation API in the cloud in subsequent articles.
+
+### Infrastructure as a function
+
+We use an `inline` Pulumi program that doesn't require a separate package with `index.ts` and `Pulumi.yaml` files. Inline programs are functions that can be authored in an `index.ts` file or be imported from another package. The example deploys an AWS S3 website with the context and deployment automation defined in a function.
```
EvanBoyle, line 27: > The example deploys an AWS S3 website

Conceptually, our custom unit of infrastructure is a static S3 website in this example. Pulumi Automation API allows us to take our custom infrastructure component, and turn it into a higher level RESTful resource that can be managed programmatically. 

While the unit of custom infrastructure in this is example is a static website with content, it could just as easily be full web services, or more complex deployments like kubernetes clusters with accompanying docker images and AWS resources.

Going back to the previous paragraph:

> You can create methods or functions to create infrastructure that you can call from other frameworks.

I think we may want to consider discussing "components", "templates", or "platforms". There is a "thing" defined with pulumi, and the automation API allows you CRUD that thing dynamically.

content/blog/automation-api-as-platform/index.md
```diff
@@ -0,0 +1,275 @@
...
+            projectName,
+            // generate our pulumi program on the fly from the POST body
+            program: createPulumiProgram(content),
+        });
+        await stack.setConfig("aws:region", { value: "us-west-2" });
+        // deploy the stack, tailing the logs to console
+        const upRes = await stack.up({ onOutput: console.info });
+        res.json({ id: stackName, url: upRes.outputs.websiteUrl.value });
+    } catch (e) {
+        if (e instanceof StackAlreadyExistsError) {
```
EvanBoyle, line 110: I'm not sure if it makes sense to discuss given the scope and focus of the article, but the ability to detect and handle structured errors is pretty important. Automation API gives you tools to detect and handle error scenarios which is a necessary part of "putting your infrastructure online" or embedding it in other complex software.

content/blog/automation-api-as-platform/index.md
```diff
@@ -0,0 +1,275 @@
...
+
+Once we're done, we can delete the website.
+
+```bash
+$ curl --header "Content-Type: application/json"   --request DELETE http://localhost:1337/sites/hello
+```
+
+## Pulling it all together
+
+So what can you do with this? We used a static website in this example so that results are immediately available. However, this application does basic [CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete) operations, which are the basis of many applications. For example, an application could send photos to a bucket or send data to a database application.
```
EvanBoyle, line 270: > For example, an application could send photos to a bucket or send data to a database application.

I think this would be a good point to link back to the title "Build Self-Service Cloud Infrastructure with Automation API".

Automation API enables building our own custom cloud platforms. This example has shown us how to take our customized unit of infrastructure and expose it to our broader team via a REST interface that we define and control.  This interface is familiar to developers, and feels more comfortable than using IaC directly. Today, we expose a static website, but tomorrow it could be a Kubernetes application platform with all of our networking and security best practices.

EvanBoyle (approved):
Broadly, the article has strong bones and can be published as-is. Love the detailed walkthrough, really makes it digestible and approachable. 

I do think there is an opportunity to iterate on the messaging with respect to the value prop and framing around "platforms" if we think it's worth the time. Unfortunately, many of my comments are directional and not concrete suggestions.

content/blog/automation-api-as-platform/index.md
spara: good point, added emphasis that it's the same building blocks with new programmatic surface area.

content/blog/automation-api-as-platform/index.md
spara: good point to highlight

content/blog/automation-api-as-platform/index.md
spara: added the point

content/blog/automation-api-as-platform/index.md
spara: addressed in previous comment

content/blog/automation-api-as-platform/index.md
spara: added

PR #4561, opened 2020-11-11
Cloud Engineering Summit content

zchase:
This PR adds all the content from the Cloud Engineering Summit on the website and lists them on the `/resources` page. This PR also introduces a podcast rss.xml feed so we can list our content on Spotify, Itunes, etc. In addition to an RSS feed, there is also a listing page for Podcasts that was created for mostly SEO purposes. One issue is that we cannot currently link to the Podcast listings on the site because we need the RSS feed to exist before we can wire up the links. So there will be a follow-up PR once our Podcast channels are created.


cnunciato:
Purely on the surface, I think we should store these files outside the repository and just check the links to them. There's no need to check them into Git.

zchase:
@cnunciato @justinvp I removed all the podcast stuff so this is ready for review.

archetypes/webinar/index.md
```diff
@@ -73,6 +73,10 @@ main:
     learn:
         - ""
 
+# This section contains the transcript for a video. It is optional.
+transcript: |
+    Here is where you would put the transcrip to a recorded video.
```
lukehoban, line 78:
```suggestion
    Here is where you would put the transcript for a recorded video.
```

lukehoban (approved):
LGTM

Not blocking - but would it make sense to have these all link to the cloudengineeringsummit site?  Or to a filtered view of all content from the summit on our site?  There’s nothing in these currently indicating they were part of the summit. That’s particularly of interest because some/most content from the summit wasn’t even really about Pulumi - so it may not be clear why it’s listed here under Pulumi content resources.

PR #4568, opened 2020-11-11
Update banner to promote intro to pulumi workshop

zchase:
Just as the title states. This PR updates the banner on pulumi.com to promote the introduction to pulumi workshop page.


cnunciato (approved):
Looks fine technically, but would it make sense to shorten these a little? Here's how this banner looks on desktop and mobile for me, for example:

image

image

Seems like we could maybe tighten things up a bit, but needn't block.

Issue #4547, opened 2020-11-08
Updates to pricing page


susanev:
## overview

after a careful user experience review of the pricing page and several iterations and feedback sessions, this issue describes updates we would like to make to the pricing page. assign @susanev to review associated prs

## design details

- all designs can be viewed directly in [figma](https://www.figma.com/file/j1Fq8UsKL7HE56jz0VoLbT/pricing-page-nov-2020-updates?node-id=0%3A1)
- styles were carefully selected; styles can be accessed directly from figma by going to the inspect tab on the right-panel

### top-content area
- h1 of `Pricing`
- subhead of `The Pulumi Service helps teams adopt the Open Source SDK, with easy, robust, and secure state management and workflows.` .. slightly modified on the mobile version
- open source direction of `Prefer to manage your own backend? Try open source` with the text `Try open source` directing the user to the State and Backend guide.

### teams pricing section
- h2 of Pulumi Service for Teams
- use the fontawesome `check` icon with a color of `#2FC89F`
- schedule a demo link should go to contact us
- primary button `Try for Free`
- ? unsure if purchase now is actually a thing yet, so we may need to create a separate issue for that
- on smaller screens, the 3 plans should stack, with the primary button being repeated after each section

### compare plans table updates
- use the fontawesome `check` icon with a color of `#2FC89F`
- repeat top headings
- underline links in the left column, and remove bold from first column
- some content updates, see [figma designs](https://www.figma.com/file/j1Fq8UsKL7HE56jz0VoLbT/pricing-page-nov-2020-updates?node-id=1%3A2)
- bottom of table has action buttons

### frequently asked questions section
- reconstruct this with tabs and accordions

PR #4577, opened 2020-11-13
Revamp pricing page

zchase:
Fixes: https://github.com/pulumi/docs/issues/4547

Further details are in the issue but this PR revamps the pricing as per the agreed-upon design. The design can be viewed [here](https://www.figma.com/file/j1Fq8UsKL7HE56jz0VoLbT/pricing-page-nov-2020-updates?node-id=0%3A1). I opted to keep the form at the bottom of the page so the link for "Schedule a Demo" stays on the page.

### DIFF
--- a/assets/sass/_pricing.scss
+++ b/assets/sass/_pricing.scss
@@ -25,12 +25,15 @@
 }
 
 .pricing-compare-plans-table {
-    @apply mt-8;
+    @apply mt-4;
+
+    @screen lg {
+        @apply mt-8;
+    }
 
     .compare-plans-table-header {
         &-container {
-            @apply flex px-8 py-5;
-            background-color: rgba(42, 19, 55, 0.05);
+            @apply flex px-8 py-5 bg-gray-200;
         }
 
         &-item {
@@ -112,6 +115,10 @@
 
         .faq-item-label {
             @apply flex select-none;
+
+            p:first-child {
+                @apply pr-2;
+            }
 
             p, i {
                 @apply text-black;
--- a/assets/sass/_pricing.scss
+++ b/assets/sass/_pricing.scss
@@ -1,6 +1,6 @@
 .pricing-plans {
     &-item-container {
-        @apply w-full px-4 text-center;
+        @apply px-4 text-center;
 
         @screen lg {
             @apply w-1/3;
@@ -12,11 +12,7 @@
     }
 
     &-item-description {
-        @apply text-sm text-gray-700 mt-0 mb-2;
-
-        @screen lg {
-            @apply px-12;
-        }
+        @apply text-sm text-gray-700 mt-0 mb-2 px-3;
     }
 
     &-item-price {
--- a/assets/sass/_pricing.scss
+++ b/assets/sass/_pricing.scss
@@ -1,31 +1,141 @@
-.pricing-item-heading {
-    @apply bg-gray-200 text-purple-700 text-sm font-display font-bold border-b border-gray-300 px-2 py-4;
+.pricing-plans {
+    &-item-container {
+        @apply w-full px-4 text-center;
 
-    @screen md {
-        @apply text-base p-4;
-    }
-}
-
-.pricing-item {
-    @apply flex items-center border-b border-gray-100 text-xs text-center;
-
-    @screen md {
-        @apply text-sm border-b border-gray-200;
-    }
-
-    .pricing-item-label {
-        @apply w-1/4 border-gray-200 text-left p-2 font-bold;
-
-        @screen md {
-            @apply font-bold py-4 px-8;
+        @screen lg {
+            @apply w-1/3;
         }
     }
 
-    .pricing-item-value {
-        @apply w-1/4 p-2;
+    &-item-header {
+        @apply text-purple-500 font-bold text-lg mb-5;
+    }
 
-        @screen md {
-            @apply p-4;
+    &-item-description {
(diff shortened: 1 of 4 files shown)


susanev (approved):
thanks so much for doing this work so quickly!! 🎉 
~noticed a few small things~

~- [ ] this text is slightly different on the desktop page, is that possible to do? `The Pulumi Service helps teams adopt the Open Source SDK, with easy, robust, and secure state management and workflows.` also would love if we can keep Open Source SDK all on the same line if possible~
~- [ ] change Pulumi Service for Teams to `Pulumi Service`~
~- [ ] below the service heading add `Prefer to use the service as an individual for free? Get started`~
~- [ ] change `Contact Us` back to `Custom`~
~- [ ] swap the order back to lowest cost first~

love the accordions and the contact form (thx for adding that in!)

joeduffy:
@susanev I got a bit more feedback from a couple of folks, and two bits seem to be recurring and similar to what @cnunciato pointed out. It's making me ponder. Also curious if @jaxxstorm has feedbak.

The two points of feedback were

* Due to the font size, and ordering from enterprise-to-starter left-to-right, "Contact Us" is the first thing the eye is drawn to. This can emphasize the commercial/enterprise feel more than maybe we want to. (It used to be "Custom" which, in addition to being smaller, might land slightly better since our target audience hates talking to salespeople.)

* The lack of the individual tier on the left hand side where your eye was naturally drawn to definitely makes it feel like we don't have that free option for individuals. I explained the new phrasing and consensus was that it might not be enough to offset the initial reaction of the enterprise-forward presentation.

susanev:
@joeduffy was hoping to a/b test the ordering of the team plans after we got this out, but we can def start with the team starter first, we can def switch back to custom too, that was nit from @zchase i believe but nothing we need to change now. ill add that to my list of feedback for @zchase 

and with regards to the individual plan, we could take a step backward and go back to an earlier iteration that didn't combine all into one action if we would like to that may be a bit less drastic and folks might be more comfortable with that step, let me know
[figma earlier iteration](https://www.figma.com/file/LDzlYOa7fFDn4AKK1kNNIY/pricing-page-iterations?node-id=17%3A0)

susanev:
spoke with joe, we are gonna go back towards an earlier iteration. i will update the designs later today.

susanev:
alright designs are updated
[figma designs](https://www.figma.com/file/j1Fq8UsKL7HE56jz0VoLbT/pricing-page-nov-2020-updates?node-id=0%3A1)

i think that requires these changes

- [ ] add a community card to the top
- [ ] community card button should become primary on mobile since its at the top of the page
- [ ] adjust the heading on the team starter plan, `Everything in Community...`
- [ ] remove pulumi service heading

susanev:
@zchase everything looks good, one more small thing I noticed...
- [ ] get started button on community card on desktop view should be secondary like the one on open sdk (on mobile it becomes primary cause its at the top of the page)

joeduffy:
One small thing that just struck me, @susanev what do you think of this? I added an "Open source for teams" that links to the section below, in case someone has the common reaction we know they have (only free options for individuals):

image

Maybe some wordsmithing to do here -- e.g., "Open source options for teams" -- but that's the general idea.

susanev:
love it!

assets/sass/_pricing.scss
```diff
@@ -1,31 +1,137 @@
...
 
-.pricing-item {
-    @apply flex items-center border-b border-gray-100 text-xs text-center;
+.pricing-compare-plans-table {
+    @apply mt-8;
+
+    .compare-plans-table-header {
+        &-container {
+            @apply flex px-8 py-5;
+            background-color: rgba(42, 19, 55, 0.05);
```
cnunciato, line 33: Let's make this either `bg-gray-100` or `-200`, to keep it within the palette. (Alternatively, we could extend that palette to include another hue, but I'd rather not do one-off styles like this unless they're not really representable with Tailwind.)

assets/sass/_pricing.scss
```diff
@@ -1,31 +1,137 @@
...
 
-        @screen md {
-            @apply font-bold py-4 px-8;
+                    a {
+                        @apply text-black;
+                    }
+                }
+            }
+
+
```
cnunciato, line 97:
```suggestion
```

cnunciato (commented):
Few little things we could probably tidy up, mainly on mobile, but otherwise looking great.

On desktop, the upper and lower borders don't quite line up:

image


Text overlaps in several sections on mobile:

image

The design makes fuller use of horizontal space here:

image

The design doesn't give explicit guidance for how these tabs should look on mobile, we could probably make this look a bit better with some center alignment and/or margin/padding reduction:

image

susanev:
@zchase on mobile lets drop the tabs for the faqs and add headings instead above each group

cnunciato:
> we should at some point get a responsive table for our website, but I think we don't have to do that in this pr

💯 🌟

joeduffy:
@zchase Is tomorrow the day? 🤩

zchase:
@joeduffy Yup I'm addressing the remaining feedback this morning!

susanev:
@zchase i saw one more small thing, but also good with doing it in another pr and merging this, whatever you think is best

- [ ] open source link in community box should have the same styles as the schedule a demo one

cnunciato (approved):
Save for @susanev's comments, merge when ready. Looks great. 🚀 🚀 🚀

PR #4595, opened 2020-11-14
Add docs on assets and archives

### LINKED ISSUE #2605: Discuss Assets and Archives in Programming Model docs
Unless I'm mistaken, we don't document the concept of assets and archives very clearly anywhere, other than having API docs for them. These are core enough concepts that we should cover them, probably in the Programming Model overview, if not a sister page to it.

### LINKED ISSUE #2895: Add side-menu's entry for Components resources
Hi, 

I was looking for Documentation about Components resources and had difficulties to find it out.
It was because entry was not in side-menu:

image

See: https://www.pulumi.com/docs/intro/concepts/programming-model/#pulumipulumi

Regards,

lukehoban:
Fixes #2605.
Fixes #2895.


stack72 (approved):
Thank you for this! I've been asked this before and not been able to point to a specific location - this is awesome

PR #4677, opened 2020-11-25
Update tutorials

komalali:
Ran `mktutorial` to regenerate tutorial content.


lukehoban:
This adds 17 `azure-nextgen` tutorials, but they aren't added to the TOC page (https://www.pulumi.com/docs/tutorials/azure/).  I'm not sure exactly what we want to do here (we *could* put them into the existing Azure ToC, but intermixing them there without more work to explain the difference doesn't feel right).  I think it's fine to merge this as-is for now and deal with it as part of transitioning `azure-nextgen` to be the default in Q1 - but a little odd that these tutorials will have no inbound links at all.  cc @mikhailshilkov

leezen (approved):

PR #4709, opened 2020-12-03
Update enforcing-a-policy-pack.md

psi09:
made word bold for readability..


leezen:
Thanks for the PR and suggestion, but this isn't really inline with the standard style we use on the site. Please comment if there was a reason beyond bolding the phrase to emphasize it.

Issue #4533, opened 2020-11-06
Document import command


joeduffy:
We should document how to use the new `import` command on the [Adopting Pulumi guide's "Importing Infrastructure" page](https://www.pulumi.com/docs/guides/adopting/import/).

stack72:
I will ensure this is part of the current import docs work

joeduffy:
@stack72 I am curious, was this scoped as part of delivering the import feature?

@davidwrede Good to think about overall process for documenting feature launches. I will observe we often blog and then move on without the corresponding product docs. I love blogging, but always prefer to link end users to the official docs.

stack72:
Posting here a quick update I gave to @joeduffy on this. This was not forgotten, just a couple of time sensitive items needed to happen first

https://github.com/pulumi/docs/pull/4726

That PR should update the guide

@joeduffy / @davidwrede I'd love your thoughts on that please

davidwrede:
@joeduffy Yup, I agree that it's good to have blog posts that help raise awareness of new features, but we should always strive to have corresponding docs available for users at launch. Of course, being able to do that is highly dependent on available resources and time. 

Happy to help contribute to time estimation for upcoming feature doc work as we do future quarterly and milestone planning.

PR #4726, opened 2020-12-07
Updating the import guide based on the new CLI command

stack72:
Fixes: #4533

### DIFF
--- a/content/docs/guides/adopting/import.md
+++ b/content/docs/guides/adopting/import.md
@@ -7,9 +7,11 @@ menu:
     weight: 1
 ---
 
-Most infrastructure projects require working with existing cloud resources, either by building on top of existing resources or adopting existing resources under management with a new and more robust infrastructure provisioning solution.
+Most infrastructure projects require working with existing cloud resources, either by building on top of existing
+resources or adopting existing resources under management with a new and more robust infrastructure provisioning solution.
 
-No matter how you've provisioned these resources &mdash; manually in your cloud provider's console or CLI, using an infrastructure as code tool like Terraform or AWS CloudFormation &mdash; Pulumi enables you to adopt and manage your resources.
+No matter how you've provisioned these resources &mdash; manually in your cloud provider's console or CLI, using an
+infrastructure as code tool like Terraform or AWS CloudFormation &mdash; Pulumi enables you to adopt and manage your resources.
 
 <!--more-->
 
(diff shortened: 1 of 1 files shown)


content/docs/guides/adopting/import.md
```diff
@@ -18,204 +20,399 @@ When working with existing resources, there are two primary scenarios:
...
 
-For the first situation, consult [the user guide index]({{< relref "/docs/guides/adopting#coexistence" >}}). For the second, let's now see how to adopt existing resources.
+For the first situation, consult [the user guide index]({{< relref "/docs/guides/adopting#coexistence" >}}). For the
+second, let's now see how to adopt existing resources.
 
 ## Adopting Existing Resources
 
-To adopt existing resources so that Pulumi is able to manage subsequent updates to them, Pulumi offers the [`import`]({{< relref "/docs/intro/concepts/programming-model#import" >}}) resource option. This option request that a resource defined in your Pulumi program adopts an existing resource in the cloud provider instead of creating a new one as would normally occur. In keeping with its focus on infrastructure as _code_, Pulumi lets you specify this `import` behavior inside the Pulumi code for your infrastructure deployment, instead of outside of it in a manual workflow.
+To adopt existing resources so that Pulumi is able to manage subsequent updates to them, Pulumi offers the
+[`import`]({{< relref "/docs/reference/cli/pulumi_import" >}}) cli command option. This option request that a resource
```
joeduffy, line 29: Do we want to _replace_ the resource-level import documentation? I thought we'd want to offer this _in addition_, as an alternative, as the other programmatic model still offers benefits around automating the imports, no?

@lukehoban, curious for your feedback aslo.

lukehoban: Yes - I would like to keep both. The resource level approach is still useful for many scenarios, and thus is still the definitive documentation for that capability. 

I would lead with `pulumi import` as that is a simpler starting point.

content/docs/guides/adopting/import.md
```diff
@@ -18,204 +20,399 @@ When working with existing resources, there are two primary scenarios:
...
-
-To perform the import, run `pulumi up` as usual, and you will see `=` instead of the usual `+`, indicating an import operation:
+> Your Pulumi stack must be configured correctly---e.g., in this case, the correct AWS region---otherwise the resource will not be found.
 
+After successfully importing a resource, you can run `pulumi up`, and all subsequent operations will behave as though Pulumi provisioned
+the resource from the outset. When Pulumi performs an import, the resource is added to the Pulumi [state](https://www.pulumi.com/docs/intro/concepts/state/#state),
+it is marked as a [protected](https://www.pulumi.com/docs/intro/concepts/programming-model/#protect) resource (by default),
+and it will emit the generated code that the user can add to their program before running a `pulumi up`. The resources are
+marked as protected to ensure that imported infrastructure is not accidentally deleted if the user forgets to include
+the code for the resource in their program before doing a deployment.
```
davidwrede, line 145: Some of the added text here was covered above, so I took a stab at summarizing existing information and keeping the new info in about the protected nature of these resources.

```suggestion
After successfully importing a resource and adding the generated code to your program, you can run `pulumi up` and all subsequent operations will behave as though Pulumi provisioned
the resource from the outset. The resource is added to the Pulumi [state](https://www.pulumi.com/docs/intro/concepts/state/#state),
and marked as a [protected](https://www.pulumi.com/docs/intro/concepts/programming-model/#protect) resource (by default) to ensure that imported infrastructure is not accidentally deleted if the user forgets to include the code for the resource in their program before doing a deployment.
```

content/docs/guides/adopting/import.md
```diff
@@ -18,204 +20,399 @@ When working with existing resources, there are two primary scenarios:
...
+
+### Bulk Import Operation
+
+Should a user need to import multiple resources to be managed by Pulumi, the CLI `import command` can handle this
+scenario for us. This is helpful when using the pulumi import command as part of scripting larger bulk imports of
+cloud resource:
```
davidwrede, line 151:
```suggestion
If you need to import multiple resources, the CLI `import` command can be used with a JSON file that contains references to existing cloud resources. Using a JSON file with the `import` command can be helpful as part of scripting large bulk imports of cloud resources.
```

davidwrede (commented):
Looks good, just had some suggestions that I think will help clarify a few things.

content/docs/guides/adopting/import.md
```diff
@@ -145,13 +273,11 @@ If the resource is not found, an error will occur:
...
 
-> Your Pulumi stack must be configured correctly---e.g., in this case, the correct AWS region---otherwise the resource will not be found.
-
 After successfully importing a resource, you can delete the `import` statement, rerun `pulumi up`, and all subsequent operations will behave as though Pulumi provisioned the resource from the outset. Be careful, as this applies to `destroy` operations also.
 
-### Mismatched State
+#### Mismatched State
 
-An important part of importing resources is that the resulting Pulumi program, after the import is complete, will faithfully generate the same desired state as your existing infrastructure's actual state. After the import, of course, you may edit your program to generate and apply new desired states to update your infrastructure.
+An important part of importing resources via resource options is that the resulting Pulumi program, after the import is complete, will faithfully generate the same desired state as your existing infrastructure's actual state. After the import, of course, you may edit your program to generate and apply new desired states to update your infrastructure.
```
davidwrede, line 280:
```suggestion
An important part of importing resources using the `import` resource option is that the resulting Pulumi program, after the import is complete, will faithfully generate the same desired state as your existing infrastructure's actual state. After the import, you may edit your program to generate and apply new desired states to update your infrastructure.
```

davidwrede (commented):
Just a few minor updates to clarify the text and then I think we're good!

stack72:
All yours again @davidwrede

davidwrede (approved):
Looks great!

PR #4753, opened 2020-12-08
Fix gen_resource_docs

komalali:
### Proposed changes

The script was looking in the wrong place for the k8s schema. It follows the same pattern as other providers now.

This should fix the failing job in pulumi/pulumi https://github.com/pulumi/pulumi/pull/5894/checks?check_run_id=1520275082


stack72 (approved):
Thank you so much!!!!

PR #4763, opened 2020-12-09
Fix some broken links

cnunciato:
Couple of broken-link fixes:

* Codepen's fine, it just doesn't respond OK to regular GET requests anymore, so this addds an exception for it.

* Wallaroo's www host doesn't appears to be gone, but wallaroo.ai is still there, so this changes our link to point to that URL instead, removes a bit about the post also appearing on the Wallaroo blog -- Wallaroo appears to have two blogs now, but this post doesn't appear on either of them -- and removes a not-so-valuable heading that was causing the post to render as though it were two posts in the listing: 

image


cnunciato:
@zchase friendly ping. Thanks!

zchase (approved):
LGTM 🚀

PR #4803, opened 2020-12-16
Fix chooser references

### LINKED ISSUE #4804: Error: Chooser attribute `type` is required.
Sentry Issue: [DOCS-5F](https://sentry.io/organizations/pulumi/issues/2090441573/?referrer=github_integration)

```
Error: Chooser attribute `type` is required.
  at parseOptions (/js/components/p-cumvmial.entry.js:1:3513)
  at onRendered (/js/components/p-cumvmial.entry.js:1:2499)
  at connectedCallback/</</o.N</o</r</< (/js/components/p-28333cca.js:1:10716)
  at sentryWrapped (../src/helpers.ts:85:17)
  at emit (/js/components/p-28333cca.js:1:12509)
...
(8 additional frame(s) were not displayed)
```

cnunciato:
The `chooser` component requires a `type` attribute ([doc](https://github.com/pulumi/docs/tree/master/components#pulumi-chooser)). This adds types and options for the choosers that appear on this page.

Fixes #4804.

Alternatively, the intent were to show only one chooser on the page, we could self-close the shortcode and omit all of the other `chooser` references:

```
{{< chooser language "typescript,python,csharp,go" / >}}
                                                   ^-- self-close

{{< choosable language typescript >}}
    // Some TypeScript example
{{< /choosable >}}
...

{{< choosable language typescript >}}
    /// Some other typescript example
```

We normally include choosers alongside code listings, but they're technically optional, so it's something to keep in mind.

### DIFF
--- a/content/blog/announcing-enum-support/index.md
+++ b/content/blog/announcing-enum-support/index.md
@@ -94,7 +94,7 @@ In the above code, the S3 `Bucket` resource has a property called `acl`, where w
 
 If we look at the [resource docs](https://www.pulumi.com/docs/reference/pkg/aws/s3/bucket/#acl_nodejs), we can see that the `acl` property can only be set to one of a few different values: `private`, `public-read`, `public-read-write`, `aws-exec-read`, `authenticated-read`, and `log-delivery-write`. The `acl` property is the perfect candidate for an enum type and is emitted as one, so you can use the following code instead.
 
-{{< chooser >}}
+{{< chooser language "typescript,python,csharp,go" >}}
 {{< choosable language typescript >}}
 
 ```typescript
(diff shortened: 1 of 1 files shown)


komalali (commented):
@cnunciato I wanted all the language-specific bits to change with only one click and don't want to litter the entire page multiple language options. The page behaves as expected currently. 

I had tried to have just one choose wrap the entire page but that led to weird formatting issues. Maybe the self-close is the way to go?

cnunciato:
@komalali I thought that might be what you were thinking, yeah. In that case, self-close would do it. Again, we don't _normally_ do it this way (we usually show the tabs alongside all listings; if it feels littery, we could probably address that with some UX improvements), but you could go that route if you wanted.

cnunciato:
> The page behaves as expected currently.

Yep, it's just that these JS errors triggered some alerts, so we'll need to fix it up: 

image

I have a change queued up to use a single chooser (resulting in the behavior you have now), so if you want me to push that up, lemme know! Your preference.

komalali (approved):

cnunciato:
Thanks @komalali! 🎉

PR #4807, opened 2020-12-16
Announcing SOC 2 Accreditation

spara:


ericrudder (approved):
assuming same as Gdoc; good too go ... auditors signed off today

susanev (commented):
noticed we are using both single-space and double-space after a period, do we have a standard?

spara:
It's single space, but I didn't catch all the double spaces from the
original text.

PR #4816, opened 2020-12-17
Exclude Twitter 400s in the link checker

cnunciato:
Looks like Twitter is still experiencing issues with one of their APIs (https://api.twitterstat.us), so this change just ignores all 400s from them.

### DIFF
--- a/scripts/check-links.js
+++ b/scripts/check-links.js
@@ -260,7 +260,6 @@ function getDefaultExcludedKeywords() {
         "https://blog.coinbase.com/",
         "https://www.netfilter.org/",
         "https://codepen.io",
-        "https://twitter.com/pulumicorp"
     ];
 }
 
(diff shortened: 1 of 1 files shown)


komalali (approved):

komalali:
Thanks @cnunciato!

leezen:
@cnunciato What's the mechanism for us to revisit this? Or, alternatively, does checking Twitter provide any value whatsoever? I don't think it actually returns 404s for non-existent tweets/handles?

FWIW, I just tried `curl https://twitter.com/asdfasdfadsf1581` and the interesting thing is it returns a 400 with the following message: "This browser is no longer supported." So I think the problem is that `blc` is not coming across as a supported user agent. But, per my above comment, I wonder if excluding Twitter completely is the right thing to do anyway?

cnunciato:
It looks like the service is still degraded, so that's probably why we're still seeing this behavior. My plan was just to check on it again in a few days, mostly just out of curiosity. (I noticed the same thing, by the way.)

I do think there's some value in checking, as things do occasionally disappear, but it's obviously not a big deal in the grand scheme of things. I'll follow up though. What we have now seems okay to me for the time being.

PR #4862, opened 2020-12-28
Add UTM param tracking to hubspot forms

zchase:
In part of the migration to Salesforce, we can expand our usage of UTM parameters and begin capturing them on our form submissions. By passing these values through the form we can stamp these conversion values onto the record created in SF when a form is submitted. This is immensely helpful in attribution reporting and overall funnel metrics.

### DIFF
--- a/components/src/components/hubspot-form/hubspot-form.tsx
+++ b/components/src/components/hubspot-form/hubspot-form.tsx
@@ -35,6 +35,8 @@ export class HubspotForm {
     // propagate up to the global window object. Then once it is available
     // we should create the HubSpot form.
     componentDidLoad() {
+        const hsFormTargetId = `#hubspotForm_${this.formId}`;
+
         waitForWindowPropertyToExist("hbspt").then((hbspt: any) => {
             hbspt.forms.create({
                 portalId: "4429525",
@@ -42,18 +44,19 @@ export class HubspotForm {
                 css: "",
                 cssClass: this.class,
                 goToWebinarWebinarKey: this.goToWebinarKey,
-                target: `#hubspotForm_${this.formId}`,
+                target: hsFormTargetId,
             });
             this.isLoading = false;
             this.didLoad = true;
 
             // Hidden form fields from HubSpot created forms causes gaps to appear in
-            // the form which is less then desireable. So we need to find the hidden
+            // the form which is less then desirable. So we need to find the hidden
             // form fields and then hide the parent <fieldset> for it. We will also need
             // to populate those fields with the correct values.
-            waitForElementToExist(`#hubspotForm_${this.formId} form fieldset div`).then(() => {
-                const fieldSets = document.querySelectorAll(`#hubspotForm_${this.formId} form fieldset div[style*="display:none"]`);
-                fieldSets.forEach((fieldset) => {
+
+            waitForElementToExist(`${hsFormTargetId} form fieldset div`).then(() => {
+                const fieldSets = document.querySelectorAll(`${hsFormTargetId} form fieldset div[style*="display:none"]`);
+                fieldSets.forEach((fieldset: any) => {
                     fieldset.parentElement.style.display = "none";
                 });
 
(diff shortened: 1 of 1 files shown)


jaxxstorm (approved):

components/src/components/hubspot-form/hubspot-form.tsx
```diff
@@ -46,6 +46,39 @@ export class HubspotForm {
...
+                });
+
+                // Populate the UTM parameter fields.
+                const cookies = parseCookie();
+                const utmCookie: any = parseUTMCookieString(cookies["__utmzz"]);
+                const utmCampaign = utmCookie.utmccn || "(not set)";
+                const utmSource = utmCookie.utmcsr || "(direct)";
+                const utmMedium = utmCookie.utmcmd || "(none)";
+
+                const utmCampaignInput = document.querySelector<HTMLInputElement>(`input[name="last_utm_campaign"]`);
```
cnunciato, line 67: Should these be `document.querySelector`, or could it be scoped more specifically to this component instance?

components/src/components/hubspot-form/hubspot-form.tsx
```diff
@@ -46,6 +46,39 @@ export class HubspotForm {
             });
             this.isLoading = false;
             this.didLoad = true;
+
+            // Hidden form fields from HubSpot created forms causes gaps to appear in
+            // the form which is less then desireable. So we need to find the hidden
+            // form fields and then hide the parent <fieldset> for it. We will also need
+            // to populate those fields with the correct values.
+            waitForElementToExist(`#hubspotForm_${this.formId} form fieldset div`).then(() => {
+                const fieldSets = document.querySelectorAll(`#hubspotForm_${this.formId} form fieldset div[style*="display:none"]`);
```
cnunciato, line 55: There's some duplication here that could probably be pulled into a variable.

components/src/components/hubspot-form/hubspot-form.tsx
```diff
@@ -46,6 +46,39 @@ export class HubspotForm {
             });
             this.isLoading = false;
             this.didLoad = true;
+
+            // Hidden form fields from HubSpot created forms causes gaps to appear in
+            // the form which is less then desireable. So we need to find the hidden
```
cnunciato, line 51:
```suggestion
            // the form which is less then desirable. So we need to find the hidden
```

Still, rather that do this with code, could we do it instead with some targeted CSS?

components/src/components/hubspot-form/hubspot-form.tsx
```diff
@@ -68,7 +101,7 @@ export class HubspotForm {
         }
     }
 
-    // When the form is submiited we also send an event to segment with the relevant UTM tags.
+    // When the form is submitted we also send an event to segment with the relevant UTM tags.
```
cnunciato, line 104:
```suggestion
    // When the form is submitted we also send an event to Segment with the relevant UTM tags.
```

components/src/components/hubspot-form/hubspot-form.tsx
cnunciato: Ah, perhaps not, if it targets a parent.

components/src/components/hubspot-form/hubspot-form.tsx
```diff
@@ -46,6 +46,39 @@ export class HubspotForm {
...
+                const utmMedium = utmCookie.utmcmd || "(none)";
+
+                const utmCampaignInput = document.querySelector<HTMLInputElement>(`input[name="last_utm_campaign"]`);
+                if (utmCampaignInput) {
+                    utmCampaignInput.value = utmCampaign;
+                }
+
+                const utmSourceInput = document.querySelector<HTMLInputElement>(`input[name="last_utm_source"]`);
+                if (utmSourceInput) {
+                    utmCampaignInput.value = utmSource;
```
cnunciato, line 74: Should this be `utmSourceInput.value`?

cnunciato:
@zchase Could you point me to a form we're using somewhere that exhibits the visual behavior you're accounting for with the fieldset hiding bit? I'd just like to see how that works.

components/src/components/hubspot-form/hubspot-form.tsx
zchase: Yeah I tried a CSS approach first, but the way HubSpot applies the `hidden` style makes it I think impossible to currently in just pure CSS.

components/src/components/hubspot-form/hubspot-form.tsx
zchase: 👍

components/src/components/hubspot-form/hubspot-form.tsx
zchase: 👍

components/src/components/hubspot-form/hubspot-form.tsx
zchase: 👍

components/src/components/hubspot-form/hubspot-form.tsx
zchase: 👍

PR #4864, opened 2020-12-29
Fix import for aws.s3.BucketEvent interface

qasimalbaqali:
For the typescript example that handles an s3 `BucketEvent` the import for the TS interface is being imported from `@pulumi/pulumi` package instead of `@pulumi/aws`. This import is fixed in this PR.

### Proposed changes

Changed the import statement from `import * as aws from "@pulumi/pulumi";` to `import * as aws from "@pulumi/aws";`


stack72 (approved):
thanks so much for this!

PR #4912, opened 2021-01-08
Fix form component support non fieldset HubSpot forms

zchase:
Fixes: https://github.com/pulumi/docs/issues/4905

The form component was throwing an error because the Newsletter form doesn't have a `fieldset` element on it so the code thought it never existed. Fortunately, this did not cause any user interruption outside of an error being logged to the console. This PR fixes up the selector that is used to wait for the form to load so that both types of HubSpot forms are supported.

Also while I was in here, Wendy had suggested we order our events in ascending order on the resources page so that is included here as well.

### DIFF
--- a/assets/js/resources.js
+++ b/assets/js/resources.js
@@ -1,0 +1,36 @@
+/**
+ * sortResourceItems sorts the visible items in the resource list.
+ *
+ * @param {bool} sortDescending Whether the items should be sorted in descending order.
+ */
+function sortResourceItems(sortDescending) {
+    const resourceList = $("ul.resource-list");
+    const items = resourceList.children("li");
+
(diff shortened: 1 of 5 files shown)


cnunciato (approved):
LGTM, but do y'all think showing the oldest first is the right way to go? The list is pretty long. Maybe default to Upcoming?

zchase:
@cnunciato I added some javascript to support only order in ascending order on the `upcoming` tab. I was initially hesitant to add JS to the page that was not in a stencil component, but I think if we want this sort of logic to be in a stencil component we should file a separate issue. I think for this PR (and for the sake of progress) this is fine as-is.

cnunciato (approved):
Awesome, this looks much better. 🚀

PR #4969, opened 2021-01-15
Change order of language list to match sidebar.

clstokes:
Changes order of languages to match the sidebar, https://www.pulumi.com/docs/intro/languages/:

image

### DIFF
--- a/content/docs/intro/languages/_index.md
+++ b/content/docs/intro/languages/_index.md
@@ -18,10 +18,10 @@ capable as the other and supports the entire surface area of resources available
 
 The following language runtimes are currently supported by Pulumi:
 
-* [Node.js]({{< relref "javascript" >}}) - JavaScript, TypeScript, or any other Node.js compatible language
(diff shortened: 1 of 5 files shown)


clstokes:
http://pulumi-docs-origin-pr-4969-a1656cf6.s3-website.us-west-2.amazonaws.com/docs/intro/languages/

image

davidwrede (approved):

lukehoban:
We should definitely fix this - but I suspect we should align with the order in the prose instead of the order in the TOC (which I think was accidental).  This order should match the size and completeness of current or near future user bases - and the TOC order is almost exactly backward for that unfortunately.

I'd suggest either:
* Node.js 
* Python 
* .NET Core
* Go

Or:
* Python 
* Node.js 
* .NET Core
* Go

cnunciato:
I'll push a commit to fix the menu ordering to match what Luke suggests to finish this up.

davidwrede (approved):
Approving the new changes

PR #4990, opened 2021-01-20
Increase timeout for HubSpot forms

zchase:
Fixes: https://github.com/pulumi/docs/issues/4971

I was not able to reproduce this error myself, even with an ad blocker turned on. So I looked into the recent occurrences of this error and there wasn't a clear pattern to give me a clue as to what is going on here. Due to a lack of evidence for a clear culprit, I think that it could be slow internet connections so I increased the timeout when waiting for the form to load from HubSpot. I also added a `.catch` for the error as well.

### DIFF
--- a/components/src/components/hubspot-form/hubspot-form.tsx
+++ b/components/src/components/hubspot-form/hubspot-form.tsx
@@ -53,7 +53,7 @@ export class HubspotForm {
             // the form which is less then desirable. So we need to find the hidden
             // form fields and then hide the parent <fieldset> for it. We will also need
             // to populate those fields with the correct values.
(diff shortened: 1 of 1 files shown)


susanev:
@zchase you can artificially slow down your network in chrome if you want to try and repro that specifically. Network tab, then the online drop-down, slow 3g.

cnunciato (commented):
I'd feel more comfortable if we had repros and a better understanding of what's actually going on here. I'll take a look as well and see if I can help out at all.

cnunciato:
@zchase Closing this as I ended up doing some refactoring on the component in https://github.com/pulumi/docs/pull/5114.

PR #4994, opened 2021-01-21
Update how-pulumi-works.md

ivanIndjic:


davidwrede (approved):
Thanks for fixing this!

PR #5014, opened 2021-01-25
Draft Blog: Reduce Cloud Costs with ARM

spara:


MitchellGerdisch (approved):
I think the title should be changed to something like "Reduce Cloud Costs with EC2 ARM Instance Types"  to avoid confusion with Azure "ARM"

stack72:
Personally I can't really understand why it's anything to do with reducing costs - it's about a new instance type - this has nothing to do with Pulumi, this is an AWS feature

spara:
> Personally I can't really understand why it's anything to do with reducing costs - it's about a new instance type - this has nothing to do with Pulumi, this is an AWS feature

You're right it is a new instance, but it's cheaper to run compared to a similar Intel instance

> I think the title should be changed to something like "Reduce Cloud Costs with EC2 ARM Instance Types" to avoid confusion with Azure "ARM"

Sounds good, will change

Issue #4710, opened 2020-12-03
add converters to left nav of docs


tusharshahrs:
## design details
- [x] add a section named `Converters` in the left nav of the docs above reference
- [x] link to each converter in that section by name (e.g., arm2pulumi), ordered alphabetically by name

### Problem description
 Can all the pulumi converters be located on one page?  
Here are the current links.  You have to know the name of each one to find it, you can't see them all on one page.

[arm2pulumi](https://www.pulumi.com/arm2pulumi/)
[crd2pulumi](https://github.com/pulumi/crd2pulumi)
[kube2pulumi](https://www.pulumi.com/kube2pulumi/)
[tf2pulumi](https://www.pulumi.com/tf2pulumi/)

### Suggestions for a fix

Link all the converter tools on one page.
https://www.pulumi.com/docs/converters/arm2pulumi
https://www.pulumi.com/docs/converters/kube2pulumi
https://www.pulumi.com/docs/converters/crd2pulumi
https://www.pulumi.com/docs/converters/tf2pulumi
or
https://www.pulumi.com/converters/arm2pulumi and so on...

susanev:
yaah these are only discoverable from the blog posts if I understand correctly, we may want to consider putting something in the left nav that groups these together if we don't want to do a landing page
cc/ @davidwrede would love ur thoughts

lukehoban:
We do also link them all from the bottom of https://www.pulumi.com/docs/guides/adopting/, but it's not as prominent as it should be relative to the value of the tools.

davidwrede:
Yeah, they're pretty undiscoverable. I think a "Tools" left nav category (above Reference) would be a good place to put usage documentation on the converters and the Pulumi CLI.

susanev:
ive updated the issue description with the design details of this work

PR #5019, opened 2021-01-25
Adding a docs section for convertors

stack72:
Fixes: #4710

Please note that this adds a link title to the crd2pulumi doc that
only affects the docs navigation not the blog post listing itself

### DIFF
--- a/content/blog/introducing-crd2pulumi/index.md
+++ b/content/blog/introducing-crd2pulumi/index.md
@@ -2,6 +2,11 @@
 title: "Introducing crd2pulumi: Typed CustomResources for Kubernetes"
 date: 2020-08-12
 meta_desc: Generate Kubernetes CustomResource types in TypeScript, Python, C#, and Go.
+menu:
+  converters:
+    identifier: crd2pulumi
+linktitle: crd2pulumi
+weight: 2
 meta_image: crd.png
 authors:
     - levi-blackstone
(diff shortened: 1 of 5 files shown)


davidwrede:
This should unblock us on the discoverability issues for now, but IMHO, the better solution is to embed these topics from within the docs left nav so that you don't have to jump back and forth between the converter docs and the main docs. Having them hang right off of the root of the main site doesn't feel like the right place to me.

Is there any easy way to embed these pages into the core docs?

stack72:
@davidwrede I am going by what the issue says only - if you feel this needs a redesign of what they are then we need to rethink and plan that. There are no core docs for these pieces of work other than what is in their individual repositories. You can see that crd2pulumi doesn't even have it's own page on the site :/

davidwrede:
Yup, no worries @stack72 and yeah, I saw that we'd have to construct a page to link to the crd2pulumi repo README. 

The changes in this PR should help unblock users and make the converters more discoverable, so I don't want to block for a redesign. Just noting that there is (I think) a better solution that we'll have to follow-up on.

davidwrede (approved):

stack72:
@davidwrede FWIW, I *definitely* agree with you that we need pages on how to download a use these CLI tools as well - there are a lot of options that just need documented with good examples

PR #5091, opened 2021-02-01
Fixing the hashivault link to ensure nightly build success

stack72:


stack72:
the 308 isn't the cause of the broken link build failing

PR #5133, opened 2021-02-03
Remove and redirect language-specific provider docs 

### LINKED ISSUE #4071: Remove and redirect language-specific provider API docs
This issue tracks removing our language-specific **provider** API docs and redirecting them to their associated resource docs, which now have representation for all supported languages.

- [x] Python
- [x] .NET 
- [x] Node.js

This doesn't include our own API docs, of course (e.g., @pulumi/pulumi, @pulumi/policy, etc.) -- only those SDK docs that have corresponding entries at https://www.pulumi.com/docs/reference/pkg/.

cnunciato:
This change removes the Node.js, Python, and .NET SDK provider docs and redirects requests for those docs to their respective providers and services in the resource docs, with a Lambda@Edge 301. 

A few things to note:

* This PR is large because of the number of deletes. But I've grouped the relevant changes into individual commits, so as long as you stay away from the Files Changed tab (🦄 😱) and the "Delete the docs" commit, you should be able to review things pretty easily.

* We've still got automated processes running to build these docs, so I've added .gitignore rules to prevent them from making it into the repo going forward.

* This PR removes `@pulumi/aws`, so it also removes the docs for our `onObjectCreated` helpers, etc. I'm assuming this was expected and that we've got a plan to re-incorporate those docs soon, but figured I'd call it out, just in case.

You can kick the tires on my dev stack at https://www-christian.pulumi-dev.io. 

#### Examples that redirect
* https://www-christian.pulumi-dev.io/docs/reference/pkg/dotnet/Pulumi.Aws/Pulumi.Aws.html
* https://www-christian.pulumi-dev.io/docs/reference/pkg/dotnet/Pulumi.DigitalOcean/Pulumi.DigitalOcean.html
* https://www-christian.pulumi-dev.io/docs/reference/pkg/dotnet/Pulumi/Pulumi.html
* https://www-christian.pulumi-dev.io/docs/reference/pkg/nodejs/pulumi/aws/
* https://www-christian.pulumi-dev.io/docs/reference/pkg/nodejs/pulumi/aws/ecr
* https://www-christian.pulumi-dev.io/docs/reference/pkg/python/pulumi_aws/ecr

.. and others like these. Note that I try to redirect with a language selection, so if you click a link to Python provider, you'll get Python as your selected language, etc.  

#### Examples that do not redirect
* https://www-christian.pulumi-dev.io/docs/reference/pkg/nodejs/pulumi/pulumi/
* https://www-christian.pulumi-dev.io/docs/reference/pkg/nodejs/pulumi/cloud/
* https://www-christian.pulumi-dev.io/docs/reference/pkg/nodejs/pulumi/awsx/
* https://www-christian.pulumi-dev.io/docs/reference/pkg/nodejs/pulumi/kubernetesx/
* https://www-christian.pulumi-dev.io/docs/reference/pkg/nodejs/pulumi/eks/
* https://www-christian.pulumi-dev.io/docs/reference/pkg/nodejs/pulumi/terraform/
* https://www-christian.pulumi-dev.io/docs/reference/pkg/python/pulumi/
* https://www-christian.pulumi-dev.io/docs/reference/pkg/python/pulumi_policy/
* https://www-christian.pulumi-dev.io/docs/reference/pkg/python/pulumi_terraform/

... etc.

Fixes #4071.


infrastructure/edgeRedirectLambdaAssociations.ts
```diff
@@ -0,0 +1,124 @@
...
+    return undefined;
+};
+
+function dotnetSDKRedirect(uri: string): string | undefined {
+    const pattern = new URLPattern("/docs/reference/pkg/dotnet/Pulumi.(:provider)/Pulumi.(:providerAgain)(.(:service)).html");
+    const match = pattern.match(uri);
+    const exceptions = [
+        "Pulumi",
+        "Policy",
+        "Terraform",
```
cnunciato, line 113: Actually, this isn't right, as these don't exist. I'll have to direct them to GitHub instead, I think.

infrastructure/edgeRedirectLambdaAssociations.ts
```diff
@@ -0,0 +1,124 @@
...
+        }
+    ];
+}
+
+function getRedirect(uri: string): string | undefined {
+    return nodeSDKRedirect(uri) || pythonSDKRedirect(uri) || dotnetSDKRedirect(uri) || undefined;
+}
+
+function nodeSDKRedirect(uri: string): string | undefined {
+    const pattern = new URLPattern("/docs/reference/pkg/nodejs/pulumi/:provider(/:service)(*)");
```
praneetloke, line 66: Could these `URLPattern` instances be declared as `const`s outside the function since they don't seem to be dynamic per request? We could avoid having to instantiate them for every request.

infrastructure/edgeRedirectLambdaAssociations.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    const lambdaEdge = new LambdaEdge(
+        "redirects", {
+            func: (event: CloudFrontRequestEvent, context, callback) => {
+
+                // Get contents of response.
+                const request = event.Records[0].cf.request;
+
+                const redirect = getRedirect(request.uri);
+                if (redirect) {
```
praneetloke, line 18: What do you think about flipping the conditions here so we can eliminate the need for an `else {}` block? I think it's more readable that way. That is,

```typescript
if (!redirect) {
  ....
}

// Do redirect stuff...
```

praneetloke (changes requested):

infrastructure/edgeRedirectLambdaAssociations.ts
cnunciato: Ah, I see what you mean. They could, but I think I'd prefer to keep them close to the code that refers to them, just for readability. (It just translates the string into a regex; the execution time looks like somewhere between 0 and 1 ms.)

infrastructure/edgeRedirectLambdaAssociations.ts
cnunciato: Don't have a strong pref. I'll switch it up if you'd find that easier to read, sure.

infrastructure/edgeRedirectLambdaAssociations.ts
praneetloke: > t just translates the string into a regex

I would definitely put them outside the function since this will run for every single request the website receives in order to check if the redirect needs to happen. You could put literally outside each respective function and still maintain readability I think. But then again it's a Lambda and I am not sure how AWS will reuse function contexts in this scenario. Maybe it's ok, maybe it's not 🤷‍♂️ I would err on the side of putting it outside since it doesn't affect readability all that much anyway.

infrastructure/edgeRedirectLambdaAssociations.ts
praneetloke: Yeah I find that large conditional blocks will only get larger and the little `else {}` will get pushed down even more and easy to miss.

infrastructure/cloudfrontLambdaAssociations.ts
```diff
@@ -2,19 +2,19 @@
...
 import { LambdaEdge } from "./lambdaEdge";
-import {CloudFrontResponse, CloudFrontResponseEvent} from "aws-lambda";
+import { CloudFrontResponse, CloudFrontResponseEvent } from "aws-lambda";
 
 /**
  * Returns the lambda function associations for the cloudfront distribution.
  * @param addSecurityHeaders If true, will create a Lambda@Edge that sets security headers on
  * origin responses and adds the lambda to the list of associated triggers.
  */
 export function getLambdaFunctionAssociations(addSecurityHeaders: boolean):
```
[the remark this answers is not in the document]
cnunciato: Oh, I see -- yeah that wasn't clear. Lemme see what I can do to fix this up, then.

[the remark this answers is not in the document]
cnunciato: To confirm my understanding, then, @praneetloke were you thinking a signature and overall shape that looked something like this, then?

```
export function getLambdaFunctionAssociations(addSecurityHeaders: boolean, doRedirects: boolean): Associations[] {
    const associations = [];

    if (addSecurityHeaders) {
        //...
        associations.push({ includeBody, lambdaArn, eventType });
    }

    if (doRedirects) {
        //...
        associations.push({ includeBody, lambdaArn, eventType });
    }

    return associations;
} 
```

?

infrastructure/edgeRedirectLambdaAssociations.ts
praneetloke: Ah I just noticed that you are using an `origin-request` (and not `viewer-request`) trigger so this function wouldn't run for every request that CFN receives, since CloudFront will [cache the redirect response](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/http-3xx-status-codes.html) for the requested object based on the `cache-control` header. So feel free to leave it as-is. I would recommend bumping that `max-age`, though. 1 hour seems low for pages that have moved permanently, no?

infrastructure/lambdaEdge.ts
```diff
@@ -34,64 +38,80 @@ export class LambdaEdge extends pulumi.ComponentResource {
...
-            memorySize: 128,
-            publish: true,
-            role: this.role,
-            runtime: "nodejs12.x",
-            // Note that Lambda@Edge functions have a different max timeout of 30 seconds
-            // than the regular Lambda functions.
-            timeout: 5,
-        // tslint:disable-next-line:align
-        }, { parent: this });
+        this.lambdaEdgeFunc = new aws.lambda.CallbackFunction(`${this.name}-lambdaEdge`,
```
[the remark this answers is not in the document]
cnunciato: > This will most likely cause a problem during deployment.

Yep, I saw this during development: https://pulumi.slack.com/archives/C9SGX9QA1/p1612458433035600

Yeah, that's a bit of a bummer, but the workaround seems reasonable enough. Lemme see what I can do.

infrastructure/cloudfrontLambdaAssociations.ts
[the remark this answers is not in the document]
praneetloke: @cnunciato yeah that's along the lines of what I was thinking. If the function params grow over time, we can add a separate interface type for it then. What do you think?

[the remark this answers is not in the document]
cnunciato: Seems fine for now, yeah.

infrastructure/edgeRedirectLambdaAssociations.ts
cnunciato: I hear ya. 🤔  I do find it slightly more readable the way it's written now -- but separate from that, to your point, since everything ends up serialized into the function body anyway, and the function is run on every request, then if the `URLPatterns` were defined outside of their functions, every request would instantiate all three of them, whereas as written now, if the URL were a `nodejs` URL, you'd instantiate the Node pattern but not the Python or .NET patterns, since `getRedirect` would short-circuit on the first match.

Then again, since most requests won't be for redirectible URLs anyway, most will end up instantiating all three -- so maybe the thing to do instead is just return early unless the URL matches `/docs/reference/pkg/` to begin with. Lemme try that. What do you think?

infrastructure/lambdaEdge.ts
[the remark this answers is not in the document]
cnunciato: Seems good, yeah?

image

infrastructure/edgeRedirectLambdaAssociations.ts
cnunciato: This approach (of exiting early when the request isn't for an SDK doc) seems to work. I'll go with it for now -- thanks for prompting the additional thinking on it. Take a look and let me know if you have any objections.

On the cache-control bit, I honesty just used what was in the AWS example. But thinking about it a bit more, it looks like omitting the cache-control header altogether would make the browser cache the 301 ~permanently (i.e., for as long as it can), which sounds like what we want -- although with the caveat that that we wouldn't be able to republish at those URLs (since we wouldn't be able to invalidate the browser cache at that point).  Since that possibility seems low but not totally zero, I've opted to set this at a week, just to err on the side of having that smidge of extra control. Lemme know what you think.

praneetloke: > Since that possibility seems low but not totally zero, I've opted to set this at a week, just to err on the side of having that smidge of extra control. Lemme know what you think.

👍 That's along the lines of what I was thinking too.

praneetloke (approved):
I'll approve this while the Files Changed tab still loads!

But I am trying to understand why the diff still shows that the previous `lambdaEdge` function needs to be delete-replaced. Only the IAM role and its policy are changing, which can be updated in the function without a replacement.

infrastructure/lambdaEdge.ts
```diff
@@ -2,41 +2,48 @@
...
-        this.permissionName = permissionName;
+        // You might be wondering why the resources created in this component can't always have the prefix
+        // based on the `name` of the component resource. The answer has to do with restrictions AWS places
+        // on deleting Lambda@Edge replicated functions (namely that they don't let you do it).
+        // At update-time, if Pulumi determines that a Lambda@Edge function needs to be deleted,
+        // (for example, when it's renamed), our create-before-delete functionality does the right thing,
+        // creating the replacement function and the CloudFront association, but fails on attempting to delete
+        // the previously associated function because of a bug in the AWS provider. So to safeguard against
+        // these failures, we opt to have the component accept these names explicitly.
+        // https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-edge-delete-replicas.html
```
cnunciato, line 41: Given the changes below (which seem fine), I'd probably cut this back and reword it to explain when you'd use `disable...` and when you wouldn't, and put the explanation on the interface property instead.

praneetloke:
@cnunciato and I figured out that it was renaming the provider resource that caused the replacement of the component resources. We should be good to go after the latest PR build.

PR #5141, opened 2021-02-04
Add a concrete example of using tags on a node group.

jkinkead:
Show using labels and adding the required autoscaler tag.

This provides an explicit example of how to use labels + tags; the required ASG tag in particular is fairly hard to track down.

### DIFF
--- a/content/docs/guides/crosswalk/kubernetes/worker-nodes.md
+++ b/content/docs/guides/crosswalk/kubernetes/worker-nodes.md
@@ -437,11 +437,19 @@ const ngStandard = new eks.NodeGroup(`${projectName}-ng-standard`, {
          providers: { kubernetes: cluster.provider},
 });
 
-// Create a 2xlarge node group of t3.2xlarge workers with taints for special workloads.
+// Create a 2xlarge node group of t3.2xlarge workers with labels for special workloads.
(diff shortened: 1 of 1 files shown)


lblackstone (approved):
Thanks for the PR!

content/docs/guides/crosswalk/kubernetes/worker-nodes.md
```diff
@@ -437,11 +437,19 @@ const ngStandard = new eks.NodeGroup(`${projectName}-ng-standard`, {
...
+// Create a 2xlarge node group of t3.2xlarge workers with labels for special workloads.
 const ng2xlarge = new eks.NodeGroup(`${projectName}-ng-2xlarge`, {
         cluster: cluster,
         amiId: "ami-0ca5998dc2c88e64b", // k8s v1.14.7 in us-west-2
         instanceType: "t3.2xlarge",
+        // Below are the labels your nodes will have that you can use a `nodeSelector`
+        // to target.
+        labels: { "workload": "special" },
+        // The Auto Scaling Group must also be labelled for the cluster autoscaler to
+        // work properly.
```
jkinkead, line 449: I _think_ there's a qualification here that it only breaks the scale-from-zero scenario, but I'm not sure. :)

PR #5144, opened 2021-02-04
Update packages

infin8x:
Ran `npm update` and `npm audit fix --force` - now going to see if there are breaking changes as a result.


cnunciato:
@infin8x Can we close, or would you like to keep this open?

PR #4707, opened 2020-12-02
Programming model reorganization

spara:
Reorganize the programming model to make to make it more accessible.

The material has been reorganized to present the Pulumi concepts with definitions first and provide a high level overview how the pieces work together. This is different from the current model where users are stepped through the process while introducing the concepts. The content has been divided into sections to make it easier to maintain the document. 

Update: See follow up comment on changes. The content in this PR is now ready for review.

PR #5146, opened 2021-02-05
Remove stub Automation API topic

davidwrede:
This PR removes a stub topic for some upcoming documentation, which was accidentally included in #4707.


cnunciato (approved):
👍 but you could also set this to `draft: true` if you wanted to keep it in place while iterating on it.

davidwrede:
Nah, we can delete this. I've got a WIP branch for the Automation API doc work here: https://github.com/pulumi/docs/tree/automation_api_user_guide. 

This change must have slipped when i was pulling other changes in via `git stash apply`

Issue #5161, opened 2021-02-07
Test a new mechanism for docs creation via command dispatch


stack72:
There are a number of scenarios we need to cover:

1. We need to be able to generate resource docs for the providers
2. We need to be able to generate nodejs & python docs only for awsx
3. We need to be able to generate nodejs, python and cli docs for Pulumi

It would be nice to be able to do this via a single command dispatch style workflow - then we can control what steps run via each command dispatch style job

PR #5163, opened 2021-02-08
Rethink provider docs generation github action

stack72:
Fixes: #5161

We are now in the situation where there are some providers that
require typedoc and python doc but not resource docs

This PR introduces the ability to be able to differentiate between
these style of providers as each of the providers will trigger a
specific event type

e.g. pulumi-aws will trigger tfgen-provider job
pulumi-awsx will trigger non-resource-provider job

the name `tfgen-provider` is a legacy name that was introduced mid-2020
but this will be replaced when all providers are updated. the logic
in the event.action allows that to be updated easily in the future by
adding a new event_type and phasing out the old one

### DIFF
--- a/.github/workflows/generate-provider-docs.yml
+++ b/.github/workflows/generate-provider-docs.yml
@@ -9,6 +9,9 @@ on:
   repository_dispatch:
     types:
       - tfgen-provider
+      - non-resource-provider
+      # Non-resource providers are things like awsx / cloud / kubernetesx
+      # it's essentially anything that isn't using the resource based docs
 
 jobs:
   pull-request:
@@ -31,6 +34,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - run: echo "Building ${{ env.PROVIDER_NAME }} docs @ ${{ env.PROVIDER_VERSION }}"
+      - if: github.event.action != 'non-resource-provider'
+        run: echo "Building docs for a resource based provider"
+      - if: github.event.action == 'non-resource-provider'
+        run: echo "Building docs for a non-resource based provider"
       - name: checkout docs repo
         uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


cnunciato:
@stack72 I don't _think_ this should be necessary, since change in  #5157 should cause commands [like this](https://github.com/pulumi/docs/blob/1dbed400dd1a8e9adf37d4691178111ff4b869f5/.github/workflows/tfgen-provider.yml#L82) to just exit for providers that aren't in the list of packages to be built. For example:

```
➜  docs git:(master) NOBUILD=true PKGS=aws ./scripts/run_typedoc.sh
Done

➜  docs git:(master) NOBUILD=true PKGS=eks ./scripts/run_typedoc.sh
Generating typedocs
~/go/src/github.com/pulumi/pulumi-eks/nodejs/eks ~/go/src/github.com/pulumi/docs
...
```  

We'd still be doing some unnecessary setup work (installing deps and such) for packages we end up not generating docs for, so if this is meant to avoid doing that work, 👍 . I just wanted to make sure I understood the motivation.

stack72:
> @stack72 I don't _think_ this should be necessary, since change in #5157 should cause commands [like this](https://github.com/pulumi/docs/blob/1dbed400dd1a8e9adf37d4691178111ff4b869f5/.github/workflows/tfgen-provider.yml#L82) to just exit for providers that aren't in the list of packages to be built. For example:
> 
> ```
> ➜  docs git:(master) NOBUILD=true PKGS=aws ./scripts/run_typedoc.sh
> Done
> 
> ➜  docs git:(master) NOBUILD=true PKGS=eks ./scripts/run_typedoc.sh
> Generating typedocs
> ~/go/src/github.com/pulumi/pulumi-eks/nodejs/eks ~/go/src/github.com/pulumi/docs
> ...
> ```
> 
> We'd still be doing some unnecessary setup work (installing deps and such) for packages we end up not generating docs for, so if this is meant to avoid doing that work, 👍 . I just wanted to make sure I understood the motivation.

The only steps that do anything regarding python and nodejs are the non-tfgen providers - so this is still valid :)

cnunciato:
> The only steps that do anything regarding python and nodejs are the non-tfgen providers.

Sure, this is fine, if it's the approach you _want_ to take. I just wanted to make sure you knew you didn't _have_ to (at least to the best of my knowledge). I'll go ahead and ✅ assuming I've made that clear. 😄

stack72:
gotcha! Yeah, I'd rather not need to run yarn install anywhere where I don't need to :D)

cnunciato (approved):

Issue #5156, opened 2021-02-07
Link inside the page doesn't work https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/azure/types/input/#KubernetesClusterNetworkProfile


diegoitaliait:
[docs/reference/pkg/azure/containerservice/kubernetescluster.md](https://www.pulumi.com/docs/reference/pkg/azure/containerservice/kubernetescluster/)

This links for input and out doesn't work

https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/azure/types/input/#KubernetesClusterNetworkProfile

stack72:
Hi @diegoitaliait 

We actually removed the python docs this week and we regenerated the resource docs to ensure those links were updated correctly

Please can you tell me if you can still find the link that 404s? 

Paul

stack72:
My apologies @diegoitaliait, @susanev was able to point out the links that are broken

We will get that taken care of

Paul

cnunciato:
Ah, I see -- we're matching on `types` as the service name. This should be happening only for those resources that have `Supporting Types`. I'll look into adding a redirect rule to catch it, but it'll ultimately need to be fixed by these lines in pulumi/pulumi (there are several, in `resource.tmpl` and `function.tmpl`: 

https://github.com/pulumi/pulumi/blob/95c09d88c87b7921b05e218ef42196bbcf99d13b/pkg/codegen/docs/templates/resource.tmpl#L168

cnunciato:
Closing this issue as these links have been suppressed in the website, and for those still out there in the wild, we're redirecting properly:

```
$ curl -I https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/azure/types/input/\#KubernetesClusterNetworkProfile
HTTP/2 301
...
location: /docs/reference/pkg/azure/?language=nodejs
cache-control: max-age=604800
```

Thanks again for the head's up @diegoitaliait! And apologies for the bumpy experience.

PR #5167, opened 2021-02-08
Add redirect exceptions for types and config

cnunciato:
Our redirect rule is currently matching on `types` and `config`, which were subdirectories of some providers and aren't covered in the resource docs. This just adds a rule that prevents users still navigating to these URLs from experiencing 404s, by redirecting to the top-level provider. (We'll hide these sections client-side and remove their respective sections from doc-gen templates in separate PRs.)

Part of #5156.

### DIFF
--- a/infrastructure/cloudfrontLambdaAssociations.ts
+++ b/infrastructure/cloudfrontLambdaAssociations.ts
@@ -159,7 +159,7 @@ function nodeSDKRedirect(uri: string): string | undefined {
     ];
 
     if (match && match.provider && !exceptions.includes(match.provider)) {
-        if (match.service) {
+        if (match.service && !match.service.match(/types|config/)) {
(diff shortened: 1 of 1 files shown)


stack72:
@cnunciato unfortunately we still link to nodejs docs 

http://pulumi-docs-origin-pr-5167-1f03e3f9.s3-website.us-west-2.amazonaws.com/docs/reference/pkg/azure/containerservice/kubernetescluster/#kubernetesclusteraddonprofile

look at `See the input and output API doc for this type.` here

cnunciato:
@stack72 Yeah, sorry -- we don't run these redirects in the preview environment, so you wouldn't see them in effect here. (You're hitting the S3 bucket directly at that URL, not going through CloudFront.)

stack72:
got it! Let's 🚢 and then we can see how it goes right?

stack72 (approved):

PR #5179, opened 2021-02-09
Override default title string used by Swiftype

davidwrede:
Swiftype will pull in the HTML <title> tag by default. We append `| Pulumi` to the end of every title for external search SEO; however, that's not needed in a site search experience. 

Unfortunately, the current POC uses Swiftype to crawl the prod site, so we'll need to get this PR merged in before we can do local testing of the new Swiftype title change.

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -74,8 +74,11 @@
     {{ if .IsHome }}
         <title>{{ .Site.Title }}</title>
     {{ else if .Params.title_tag }}
+        <!-- Using the title frontmatter value to override the default title string that Swiftype pulls in. -->
+        <meta class="swiftype" name="title" data-type="string" content="{{ .Params.title_tag }}" />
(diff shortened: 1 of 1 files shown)


layouts/partials/head.html
```diff
@@ -1,7 +1,7 @@
 <head>
     <meta charset="UTF-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
```
cnunciato, line 4: Little thing -- you might want to set your editor to trim trailing whitespace and add empty lines. If you're using VSCode, the settings (in settings.json) should be:

image

cnunciato (approved):
LGTM! 🚀

PR #5201, opened 2021-02-11
Latest AzureNextGen uses Enumerations instead of strings on DefaultCo…

OleAlbers:
### Proposed changes

Latest AzureNextGen.DocumentDB changed `DatabaseAccountOfferType` and `DefaultConsistencyLevel` from strings to Enumerations on `DatabaseAccountArgs`. The C#-Documentation should do the same.


leezen:
@mikhailshilkov @viveklak Do we need to adjust the codegen to pickup on the enums in these examples?

mikhailshilkov:
@leezen Yes. It sounds like we haven't adjusted programgen when we introduced enums.

viveklak:
See https://github.com/pulumi/pulumi/issues/6263.

PR #5207, opened 2021-02-11
Specify WRITE has access to destroy

cnunciato:


praneetloke (approved):
Thank you, Christian!

PR #5208, opened 2021-02-11
Only have Swiftype crawl the main content on docs and blogs

davidwrede:
Currently, Swiftype is crawling every section on the page (e.g. submenus, footers, etc.). This change restricts the crawled area to the body of a given topic or blog post (hopefully).


cnunciato (approved):
LGTM. If this gives us the results we're looking for, we'll probably want to add it to a few more templates, but for now, seems good.

PR #5225, opened 2021-02-12
A few more link fixes

cnunciato:
These cropped up while the link checker was out.


data/awards.toml
```diff
@@ -65,12 +65,6 @@ featured = true
 img = "/logos/press/awards/ema-top-3.png"
 url = "https://www.ematop3.com/serverless.html"
 
-[[awards]]
```
cnunciato, line 68: This seems to have disappeared, so removing the entry and image.

stack72 (approved):

PR #5262, opened 2021-02-16
Denote blog entries in Swiftype search result sets

davidwrede:
Our blog posts contain a lot of good information that hasn't been included in the core docs, yet. Because of this, our Swiftype engine currently crawls blog posts and includes them in relevant queries; however, when looking through the result set, you don't know when you're clicking on a blog post or a page in the documentation. Clicking on a blog post means you will travel outside of the core docs to the blogs section of the Pulumi corp site.

This change is a simple fix to append "Blog" as a prefix to the topic title.


cnunciato (approved):
✨

Issue #5298, opened 2021-02-17
Automerge tfgen docs PRs on successful build


stack72:
We should add the work to automerge the docs PRs on a successful build - this will allow the docs PRs to be merged as expected as part of the deployment workflow

We use `pascalgn/automerge-action` for this in the provider repos and we can leverage the same deployment automation here

PR #5314, opened 2021-02-18
Add an automerge workflow for provider and cli docs

stack72:
Fixes: #5298

This new automerge functionality will mean that provider docs PRs
+ CLI docs PRs will get a new tag added `automation/merge` that if
present AND the checks for the PR pass, then the PR will be automerged

This workflow will work for *any* PR that is tagged with
`automation/merge` - the nice point here is that a non-pulumiam
cannot add labels to PRs

### DIFF
--- a/.github/workflows/automerge-workflow.yml
+++ b/.github/workflows/automerge-workflow.yml
@@ -1,0 +1,37 @@
+jobs:
+  automerge:
+    name: Automerge Workflow
+    runs-on: ubuntu-latest
+    steps:
+      - env:
+          GITHUB_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN }}
+          MERGE_COMMIT_MESSAGE: pull-request-title
+          MERGE_FORKS: "false"
+          MERGE_LABELS: automation/merge
+          MERGE_METHOD: squash
+          MERGE_REMOVE_LABELS: automation/merge
+          MERGE_RETRIES: "30"
+          MERGE_RETRY_SLEEP: "60000"
+          UPDATE_LABELS: automation/update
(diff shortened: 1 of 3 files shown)


.github/workflows/automerge-workflow.yml
```diff
@@ -0,0 +1,37 @@
+jobs:
+  automerge:
+    name: Automerge Workflow
+    runs-on: ubuntu-latest
+    steps:
+      - env:
+          GITHUB_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN }}
+          MERGE_COMMIT_MESSAGE: pull-request-title
+          MERGE_FORKS: "false"
+          MERGE_LABELS: automation/merge
```
cnunciato, line 10: Just to check my understanding: It's the presence of this label on _any_ PR that will result in an auto-merge on successful completion of the check suite. Correct? And this label only gets applied (automatically) with `build-tfgen-provider-docs` and `build-pulumi-cli-docs`?

stack72: Correct - this means that the use case can be extended where required, but the only place that adds it by default is the provider and cli docs PRs

stack72: We can add any tags in here that you want, but it still allows people in our org to get around the restrictions - this is a process / cultural problem if it happens right?

cnunciato: Ok, thanks for confirming. cc @leezen just as a head's up. I'm okay to try it as long as no objections from others.

cnunciato (approved):

davidwrede (approved):

PR #5341, opened 2021-02-23
Update gke.md

matt-slalom:
Without `--show-secrets` the `pulumi stack output` command produces a kubeconfig file that is not usable.


lblackstone (approved):

lblackstone:
Thanks for the PR!

PR #5383, opened 2021-03-03
Remove API docs of Azure NextGen

mikhailshilkov:
Redirects to be arranged outside Hugo


cnunciato:
> Redirects to be arranged outside Hugo

See https://github.com/pulumi/docs/pull/5398, which should be merged before this one.

cnunciato (approved):
👍 , but let's wait until #5398 is released before merging this.

PR #5428, opened 2021-03-11
fix links to stable and incubator helm chart repos

spara:
link to stable helm chart repo is outdated


content/docs/reference/pkg/kubernetes/helm/v2/chart.md
```diff
@@ -113,7 +113,7 @@ class HelmStack : Stack
             Version = "1.24.4",
             FetchOptions = new ChartFetchArgs
             {
-                Repo = "https://kubernetes-charts.storage.googleapis.com/"
```
cnunciato, line 116: Ordinarily we wouldn't make changes to these files, since they're generated -- but since the file that's used to generate them is included in this PR as well, and the generation process relies on this repo, I can't think of a reason not to update the generated files here, too. 😄 Mainly just calling it out for your info.

cnunciato (approved):
LGTM!

content/docs/guides/crosswalk/kubernetes/cluster-services.md
```diff
@@ -3,7 +3,7 @@ title: Deploy Kubernetes Cluster Services
 meta_desc: This page will provide an overview on how to deploy Kubernetes Cluster Services.
 no_on_this_page: true
 menu:
-  userguides:
+  userguides
```
cnunciato, line 6:
```suggestion
  userguides:
```

cnunciato: ☝️ Guessing that's what the check suite is complaining about.

content/docs/reference/pkg/kubernetes/helm/v2/chart.md
spara: good to know

PR #5431, opened 2021-03-11
Fix examples for Random provider intro.

clstokes:
Method and args name was wrong per https://www.pulumi.com/docs/reference/pkg/random/randompet/#create.


stack72 (approved):
Thank you!!!

lukehoban:
This needs to be fixed in all other languages as well - right?

clstokes:
> This needs to be fixed in all other languages as well - right?

Yea, you're right. I'll fix them too...

clstokes:
Fixed the rest of them.

lukehoban (approved):

Issue #5438, opened 2021-03-15
Cookie banner obscured by chat icon on mobile site


stack72:
Reported as per https://twitter.com/SGudbrandsson/status/1371375812784627712

cnunciato:
@zchase I believe these are both managed externally. Can you take this? I'm happy to give you alternate treatments for the cookie banner if you need.

zchase:
Reopening as the CSS change didn't fix the issue. To quickly fix this for users we are going to put the cookie banner on the top of the page instead of the bottom, while I work for a proper fix for the overlap issue.

PR #5492, opened 2021-03-17
Add z-index override for chat widget

zchase:
Fixes: https://github.com/pulumi/docs/issues/5438

The HubSpot chat widget overlaps with the cookie banner so this change overrides the z-index of the chat widget so it will end up beneath the cookie banner.


cnunciato (approved):
Seems a reasonable thing to try. 👍

PR #5494, opened 2021-03-17
Pull content and layouts into a separate module

cnunciato:
This change pulls all hand-crafted content and layout files into a [Hugo module](https://gohugo.io/hugo-modules/) managed at https://github.com/pulumi/pulumi-hugo. (All generated docs will continue to live here, as will the overall website- publishing and doc-gen processes.)

This means that if you're currently using this repository to make changes to the pages of the website, whether by writing Markdown docs or blog posts or working with front-end assets or layouts, you'll no longer do that here. Instead, you'll use [pulumi/pulumi-hugo](https://github.com/pulumi/pulumi-hugo) (see its [README](https://github.com/pulumi/pulumi-hugo) for details), and when your work is merged into the main branch of that repo, it'll be pulled into this one by way of a scheduled GitHub Actions workflow that updates this repo's Hugo module reference and creates an automated pull request (which, when merged, will result in a deployment as usual).

I've split this PR into two commits, one of which contains the files that have moved to the new repo (all deletes, with Git histories retained on the receiving end) and one that contains changes I've made to this repo to make everything work:

* Change config files to YAML (easier to read, use than TOML, IMO)
* Add a scheduled workflow to pull in updates from pulumi-hugo once an hour
* Update the README to point users to the new repo
* Make the appropriate adjustments to config files and deployment scripts
* Remove a couple of previously/accidentally committed files

Once this merges, all hand-crafted docs work (guides, blogs, etc.) will immediately need to be switched to pulumi/pulumi-hugo, and most open PRs will no longer be mergeable. I'm happy to handle informing users with open PRs that they'll need to submit them on pulumi-hugo instead.


cnunciato:
We need to get this merged for some work that's ahead of it, so if we could get a closer look today, I'd appreciate it. Thanks friends!

susanev:
a few questions...

- we will be moving issues?
- is there a plan to post something about this in community slack

cnunciato:
@susanev I'm going to look into moving issues, and yes, I'll announce in Community Slack as well once this is done.

.github/workflows/build-and-deploy.yml
```diff
@@ -6,10 +6,13 @@ on:
   pull_request:
     branches:
       - master
+  schedule:
+    # Every hour on the hour.
+    - cron:  '*/60 * * * *'
```
leezen, line 11: Why do we need to do this? Won't this unnecessarily generate cache churn for us when we re-build the site and flip origins every hour?

.github/workflows/update-theme.yml
```diff
@@ -0,0 +1,43 @@
+name: Update pulumi-hugo
+on:
+  schedule:
+    - cron:  '*/60 * * * *'
```
leezen, line 4: Instead of going in this direction, could we add a dispatch command to `pulumi/pulumi-hugo` that generates a PR in this repo whenever there's a commit to `master` there?

README.md
```diff
@@ -1,110 +1,50 @@
...
+We build the Pulumi website statically with Hugo, manage our dependencies with Node.js and Yarn, and write our documentation in Markdown. Below is a list of the tools you'll need if you'd like to work on the project:
 
-##### Install Go
+* [Go](https://golang.org/)
+* [Hugo](https://gohugo.io)
+* [Node.js](https://nodejs.org/en/)
+* [Yarn](https://classic.yarnpkg.com/en/)
 
-Download the Linux package from https://golang.org/dl/. Follow [installation and setup steps](https://tecadmin.net/install-go-on-ubuntu/) on "Installing Go on Ubuntu".
+Out TypeScript documentation is generated directly from source using [TypeDoc](http://typedoc.org/), and we check in the resulting files at `./content/reference/pkg/nodejs`.
```
leezen, line 23: 'Our'

leezen (approved):
Looked at the changes to use `pulumi/pulumi-hugo` (so I did not look at the other commit) and generally seems reasonable to me. I did have a couple questions/comments around the automation aspects between the two repos.

.github/workflows/build-and-deploy.yml
cnunciato: Only if there are changes to be made. If the job runs and finds that pulumi/hugo@master hasn't changed, it'll exit silently without generating a PR.

.github/workflows/update-theme.yml
cnunciato: Yes, and this was definitely my first inclination. But since these PRs would consist only of changes to go.mod and go.sum, I suspect we'd end up with lots of merge-conflict annoyance (say, when multiple PRs queue up and one of them gets merged) and the risk of deploying changes out of order. We could probably automate around that somehow, but ultimately it seems like what we want is for changes that merge at pulumi-hugo and pass integration tests to go out in sequence, and I _think_ that accomplishes that. But I don't disagree at all with that approach in principle.

leezen: The multi-module and conflict thinking seems reasonable enough to me to just go with this for now.

PR #5510, opened 2021-03-22
fix/update fastly golang intro example

Flamacue:
### Proposed changes

I was trying to use the fastly sdk with pulumi, and in using `v2.14.0` of the `fastly` sdk, the existing example in the documentation produces a compilation error:

```
    ./main.go:11:4: cannot use []fastly.Servicev1Backend{...} (type []fastly.Servicev1Backend) as type fastly.Servicev1BackendArrayInput in field value:
    	[]fastly.Servicev1Backend does not implement fastly.Servicev1BackendArrayInput (missing ElementType method)
    ./main.go:13:6: cannot use pulumi.String("127.0.0.1") (type pulumi.String) as type string in field value
    ./main.go:14:6: cannot use pulumi.String("localhost") (type pulumi.String) as type string in field value
    ./main.go:15:6: cannot use pulumi.Int(80) (type pulumi.Int) as type *int in field value
    ./main.go:18:4: cannot use []fastly.Servicev1Domain{...} (type []fastly.Servicev1Domain) as type fastly.Servicev1DomainArrayInput in field value:
    	[]fastly.Servicev1Domain does not implement fastly.Servicev1DomainArrayInput (missing ElementType method)
    ./main.go:20:6: cannot use pulumi.String("demo") (type pulumi.String) as type *string in field value
    ./main.go:21:6: cannot use pulumi.String("demo.pulumi.com") (type pulumi.String) as type string in field value
```

This updates the documentation to showcase a working example

### DIFF
--- a/content/docs/intro/cloud-providers/fastly/_index.md
+++ b/content/docs/intro/cloud-providers/fastly/_index.md
@@ -94,15 +94,15 @@ import (
 func main() {
 	pulumi.Run(func(ctx *pulumi.Context) error {
 		service, err := fastly.NewServicev1(ctx, "test", &fastly.Servicev1Args{
-			Backends: []fastly.Servicev1Backend{
-				{
+			Backends: fastly.Servicev1BackendArray{
+				fastly.Servicev1BackendArgs{
 					Address: pulumi.String("127.0.0.1"),
 					Name:    pulumi.String("localhost"),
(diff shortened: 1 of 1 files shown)


stack72 (approved):
thank you so much for this @Flamacue 

I am sorry this wasn't correct for you

Paul

PR #5566, opened 2021-03-25
Add a bucket where we can store larger files

zchase:
This PR adds an S3 Bucket so we can store files to serve on the website and not have to commit these larger files to source control.

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -197,6 +197,13 @@ const distributionArgs: aws.cloudfront.DistributionArgs = {
     orderedCacheBehaviors: [
         {
             ...baseCacheBehavior,
+            targetOriginId: largeFileBucket.arn,
+            pathPattern: "/large-files/*",
+            defaultTtl: oneHour,
+            maxTtl: oneHour,
+        },
+        {
+            ...baseCacheBehavior,
             pathPattern: "/css/styles.*.css",
             defaultTtl: oneYear,
             maxTtl: oneYear,
@@ -234,13 +241,6 @@ const distributionArgs: aws.cloudfront.DistributionArgs = {
         {
             ...baseCacheBehavior,
             pathPattern: "/images/home/*",
-            defaultTtl: oneHour,
-            maxTtl: oneHour,
-        },
-        {
-            ...baseCacheBehavior,
-            targetOriginId: largeFileBucket.arn,
-            pathPattern: "/large-files/*",
             defaultTtl: oneHour,
             maxTtl: oneHour,
         },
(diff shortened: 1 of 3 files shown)


layouts/partials/head.html
```diff
@@ -3,6 +3,11 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
 
+    <!-- If the site is not running in production let's block search engines from indexing. -->
+    {{ if ne hugo.Environment "production" }}
+        <meta name="robots" content="noindex">
+    {{ end }}
```
cnunciato, line 9: We actually do this already in robots.txt (driven mainly by all the PR previews we do):

https://github.com/pulumi/docs/blob/98ae287732c3a17de86455228a9bcde641b02858/layouts/robots.txt#L1-L4

```suggestion
```

infrastructure/index.ts
```diff
@@ -59,6 +59,14 @@ const originBucket = pulumi.output(aws.s3.getBucket({
     bucket: originBucketName,
 }));
 
+// Create a bucket to store large files.
+const largeFileBucket = new aws.s3.Bucket("large-file-bucket", {
```
cnunciato, line 63: Technically fine, but I'd suggest a more generic name, since we needn't use this solely for "large files", but really any file we want to be able to serve and manage by hand, outside of Git. Maybe something like "ad-hoc" or "files" or "uploads" or something. `uploads` might be good, yeah, since it's generic and connotes a separate activity and pretty much is what we're doing as well. What do you think?

cnunciato (approved):
Preview looks good, just a suggestion for a different name. Also, cc @leezen for additional visibility.

layouts/partials/head.html
cnunciato: ...although now I see that neither of these approaches seems to be working with dev stacks (though previews are good) because we seem to be running `production` builds for them. I'll get an issue filed for fixing this in the build (and add a bit more documentation for creating and tearing down dev stacks).

zchase: Ah, thanks for pointing this out.

infrastructure/index.ts
zchase: `uploads` works great for me. One thing to keep in mind is the assets in the Bucket need to live under a `uploads` folder in the bucket. You might already know that, but that tripped me up for like 10 minutes when setting this up.

cnunciato: Yep, makes sense and I think that's fine.

PR #5567, opened 2021-03-25
update/fix the Type field documentation for Servicev1Condition in golang

Flamacue:
### Proposed changes

I've been using the `golang` `fastly` provider to manage services and the documentation for `Servicev1Condition` looks like it may have been pasted from somewhere else. This updates the `golang` documentation to specify the proper accepted values.

The other languages should probably be updated too, but since I've only used the `golang` implementation, I'm only updating the documentation with values for this implementation.

### DIFF
--- a/content/docs/reference/pkg/fastly/servicev1.md
+++ b/content/docs/reference/pkg/fastly/servicev1.md
@@ -6603,7 +6603,7 @@ see [Fastly's Documentation on Conditionals][fastly-conditionals].
         <span class="property-indicator"></span>
         <span class="property-type">string</span>
     </dt>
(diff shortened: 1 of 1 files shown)


Flamacue:
I'm unsure if this documentation is being generated from https://github.com/pulumi/pulumi-fastly, but it looks like the comment here should be reflected as well https://github.com/pulumi/pulumi-fastly/blob/296c168ae7e2c4d2f9535e471c5178d3ff472286/sdk/go/fastly/pulumiTypes.go#L6146

If the documentation is being generated from there, then I'd probably consider this more of a bug report since I'd rather leave it to the experts to properly fix this :)

Flamacue:
closing in favor of this issue: https://github.com/pulumi/docs/issues/5574

PR #5594, opened 2021-03-29
Re-org event description to include new speakers

zchase:
Now that we have confirmed some more speakers for PulumiUP, I went back re-organized the event description to sort out the speakers into their respective sections. I also added a "register" CTA into the Hero and below the general event description.

I also updated the form class because links were not being rendered properly.


cnunciato (approved):
👍

PR #5595, opened 2021-03-29
adjust pulumiup button styles

susanev:
### Proposed changes

adjusting the pulumiup button styles in the header area a bit, so they stand out and don't get lost

#### before
buttons-before

#### after
button-glow

### DIFF
--- a/assets/sass/_pulumi-up.scss
+++ b/assets/sass/_pulumi-up.scss
@@ -7,6 +7,7 @@ $green700: theme("colors.green.700");
 $green900: theme("colors.green.900");
 
 $orange400: theme("colors.orange.400");
+$orange600: theme("colors.orange.600");
 $orange700: theme("colors.orange.700");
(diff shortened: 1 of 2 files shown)


assets/sass/_pulumi-up.scss
```diff
@@ -64,6 +64,12 @@ $orange900: theme("colors.orange.900");
     }
 }
 
+.orange-neon {
+    border-radius: 0.25rem;
+    border-color: #e17d47;
```
zchase, line 69: You can assign a variable that pulls in the value of the config, like some of the other variables at the top of the file. Example:
```
$orange600: theme('colors.orange.600');
// other stuff
.orange-neon {
    // ... styles
    border-color: $orange600;
}
```

zchase (approved):
LGTM --- One nit on using variables instead of the hex codes but otherwise looks great! 🚢

PR #5598, opened 2021-03-30
Make sure exported property matches Python naming conventions

davidwrede:
This fix also ensures the `curl` statement on https://www.pulumi.com/docs/get-started/azure/deploy-changes/ executes successfully when choosing Python as the language for the tutorial.

### DIFF
--- a/content/docs/get-started/azure/modify-program.md
+++ b/content/docs/get-started/azure/modify-program.md
@@ -210,7 +210,7 @@ Finally, at the end of `__main__.py`, export the resulting storage container's e
 
 ```python
 # Web endpoint to the website
-pulumi.export("staticEndpoint", account.primary_endpoints.web)
+pulumi.export("static_endpoint", account.primary_endpoints.web)
 ```
 
 {{% /choosable %}}


lukehoban (commented):
Is this the only issue here?
1. The preview and update output uses `staticEndpoint` seemingly irrespective of language choice
2. There is no language chooser on this page - which is a little strange?

davidwrede:
1. Good point. As we default to the output for a TypeScript project for all CLI outputs, it feels like adding a little note early on in the tutorial should suffice (rather than creating different outputs).
2.  Yeah, no language chooser on this page; however, if set on a previous page, the language choice will carry over to this page. As a best practice, it makes sense to have the language chooser tabs on each code block where different options exist.

lukehoban:
> As we default to the output for a TypeScript project for all CLI outputs

Note that an alternative fix to all of these issues would be to use the same stack export name for all languages - stack exports are really meant to be language neutral strings anyway - theres no great reason for them to differ between languages.

cnunciato:
@davidwrede Now that https://github.com/pulumi/docs/pull/5494 has merged, would you mind resubmitting this on https://github.com/pulumi/pulumi-hugo?

davidwrede:
Moved this fix to https://github.com/pulumi/pulumi-hugo/pull/17.

PR #5690, opened 2021-04-09
Initial docs for gcp native provider

viveklak:
Fixes https://github.com/pulumi/pulumi-gcp-native/issues/21


mikhailshilkov:
Why is there no preview?

content/docs/reference/pkg/gcp-native/_index.md
```diff
@@ -0,0 +1,136 @@
+---
+title: "gcp-native"
```
mikhailshilkov, line 2: I think we need to add a user-friendly alias in pulumi/pulumi

viveklak:
> Why is there no preview?

This is against the feature-3.0 branch which doesn't have the automation for previews. @stack72 let me know if you would like to chat about setting up preview for this branch or if we should find a way to get this against master instead.

content/docs/reference/pkg/gcp-native/_index.md
viveklak: > I think we need to add a user-friendly alias in pulumi/pulumi

👍🏾 . I discovered a codegen bug which currently makes gcp native python sdk unusable. Will roll that in as part of that fix.

cnunciato:
@mikhailshilkov If you want to see how these lay out when rendered, you can run `make ensure build serve-static` locally until you're ready to PR against master. 

I _think_ we could add `feature-3.0` to the list of `pull_request` branches in the build-and-deploy workflow as well, but since I haven't looked at the code from this perspective and tested, I'm not 100% comfortable just doing it -- though I suspect it would probably be okay.

content/docs/reference/pkg/gcp-native/_index.md
viveklak: This was merged in as part of https://github.com/pulumi/pulumi/pull/6748

PR #5692, opened 2021-04-09
Add a target and docs for running a static webserver locally

cnunciato:
This came up today and seemed worth making a little easier.


viveklak (approved):
Thanks for adding this!

PR #5695, opened 2021-04-09
Add details to module-update PRs

cnunciato:
One problem we have today is that PRs resulting from pulumi-hugo updates (e.g., [here](https://github.com/pulumi/docs/pull/5689)) don't contain enough information about the changes that would be incorporated if the PR were merged. This change adds a step to the `update-theme` workflow that attempts to convert the `go.mod` diff (via the GitHub API) into a bulleted list of GitHub links to the relevant pulumi-hugo commits, like this:

* https://github.com/pulumi/pulumi-hugo/commit/6952590c6e5f385c9b001a1dcae8611f96f85b5e
* https://github.com/pulumi/pulumi-hugo/commit/514ef32c841f9e98e6b5069c1e0b69aab7d0f858
* https://github.com/pulumi/pulumi-hugo/commit/31fdf2d2fa3392b28d6a45092c6412c261b3a2a4

### DIFF
--- a/.github/workflows/update-theme.yml
+++ b/.github/workflows/update-theme.yml
@@ -12,6 +12,11 @@ jobs:
         with:
           go-version: 1.16.x
 
+      - name: Install Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: '14.x'
+
       - name: Install Hugo
         uses: peaceiris/actions-hugo@v2
         with:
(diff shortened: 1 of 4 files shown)


cnunciato:
@stack72 This is untested and my first attempt at working with GHA outputs. Does it make sense and do you see any issues with it?

stack72:
Let's ship this - the workflow looks correct to me! :)

stack72 (approved):

PR #5698, opened 2021-04-10
WIP: Regenerate tutorial docs and tutorial index shortcodes

davidwrede:
Adds back in generated tutorial pages and index shortcodes after the changes made in #5494. 

Something to resolve is how we handle the tutorial index shortcodes that are generated by mktutorial in this repo with how a duplicate set of those shortcodes are referenced (and used) in pulumi-hugo. Currently, `scripts/mktutorial.sh` in the docs repo expects there to be a `layouts/shortcodes` subdirectory at the root of the repo (as it was before #5494); however, pulumi-hugo has these index shortcodes under `themes/default/layouts/shortcodes`, which then get used when building the site. 

Do we copy over the index files in a PR after every update to pulumi-hugo, or handle the index in a different way?

@cnunciato would love your thoughts and ideas on this one.


cnunciato:
@davidwrede Taking a look at this today to see if we can remove the need for those shortcodes somehow.

davidwrede:
Superceded by #5887

PR #5753, opened 2021-04-13
Adding build of docfx to the Pulumi CLI docs build

stack72:


.github/workflows/pulumi-cli.yml
```diff
@@ -89,6 +91,9 @@ jobs:
       - name: generate python docs
         run: ./scripts/generate_python_docs.sh "pulumi"
         working-directory: docs
+      - name: generate docfx
+        run: ./scripts/run_docfx.sh
+        working-directory: docks
```
cnunciato, line 96:
```suggestion
        working-directory: docs
```

cnunciato (approved):

.github/workflows/pulumi-cli.yml
stack72: good catch!!

PR #5779, opened 2021-04-16
google native resource docs

viveklak:


mikhailshilkov:
The input types look odd... Any idea why?

image

viveklak:
Yes - seems nodejs and python are misbehaving. C# and Go seem reasonable. Looking into it.

viveklak:
Cut https://github.com/pulumi/pulumi/issues/6806 to track this issue. This seems to be a bug in codegen and unrelated to this PR. Will escalate that bug separately.

stack72 (approved):

PR #5783, opened 2021-04-19
Add Lighthouse Check

cobraz:
**Edit**: I realize the amount of URLs are to much, so I'll have to find a way to either batch them up or something.

You can see a successful run of similar functionality here:
https://github.com/bjerkio/website/runs/2379101877?check_suite_focus=true


cnunciato:
Thanks for submitting this! I do agree testing on every single URL is unnecessary; testing with a few key URLs would probably suffice. If there were a way, say, to pass in an explicit list (like the home page, certain docs pages, etc.), that'd probably be good enough. cc @zchase who might have some suggestions for that initial set of pages might be. 

I'll take a closer look at the `lighthouse-ci-action` as well jsut to get a better idea what it does and how we'd best use it.

cobraz:
> cc @zchase who might have some suggestions for that initial set of pages might be.

That would be useful. Btw, I opened https://github.com/cobraz/parse-sitemap/issues/14 which could help. I imagine that with some iterations, we could get a concatenated report of a bunch of pages. Essentially get a health report. 🤷‍♂️

cnunciato:
@cobraz How about we start with just these URLs? 

```
https://www.pulumi.com/
https://www.pulumi.com/product/
https://www.pulumi.com/pricing/
https://www.pulumi.com/docs/get-started/
https://www.pulumi.com/docs/intro/concepts/resources/
https://www.pulumi.com/docs/reference/pkg/aws/s3/bucket/
```

I'm sure others will chime in as to additional pages we might want to add, but it'd be good just to have something to start with to get an idea. What do you think?

cobraz:
Should we add this for every push? Or just scheduled?

PR #5845, opened 2021-04-20
Incorporate pulumi-hugo-internal

cnunciato:
This change brings in the pulumi-hugo-internal module. 

Closes https://github.com/pulumi/pulumi-hugo-internal/issues/22.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -29,6 +29,10 @@ jobs:
           hugo-version: '0.82.0'
           extended: true
 
+      - name: Configure Git
+        run: |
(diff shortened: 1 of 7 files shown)


stack72 (approved):

stack72:
Will these 2 repos be amalgamated or is this a temporary thing?

If we need to push docs fixes do we now need to push them to the internal repo?

cnunciato:
Once this change merges and its content is deployed, I'll issue a follow-up PR on pulumi-hugo to bring it up to date with the new content from pulumi-hugo-internal, then remove the docs site's dependency on pulumi-hugo-internal. So we'll generally work in pulumi-hugo going forward.

Issue #5878, opened 2021-04-26
Remove the need for shortcodes in tutorials docgen


cnunciato:
With #5494, we moved all presentation (Hugo layouts, shortcodes, etc.) out of the docs repo into pulumi-hugo, with the aim of keeping all generated documentation (including tutorials) in the docs repo, and keeping hand-authored documentation in pulumi-hugo. Hugo handles this well, but the process that generates tutorials is also responsible for generating Hugo shortcodes that act as indexes, which leads to problems. We should remove the need to generate these shortcodes and just use Hugo data build these indexes instead.

PR #5887, opened 2021-04-27
Regenerate tutorials, add cloud and language params

cnunciato:
This change regenerates all tutorials and makes a couple of changes to the generation process:

* Rather than generate cloud-specific shortcodes containing HTML tables with links to each tutorial, we inject `cloud` and `language` params into each tutorial's frontmatter, then let `pulumi-hugo` use that frontmatter to build the lists with a single shortcode (which I'll submit in a pulumi-hugo PR meant to be merged after this one) 
 
* Adds a daily job to regenerate tutorials automatically and submit (or update) a pull request

As long as the preview looks correct, we should be able to merge this change now, then follow up with a pulumi-hugo PR that handles the rendering/indexing bit.

Part of #5878


davidwrede (approved):
Spot-checked each of the lists for each cloud provider. Looks good! 

I love that we're using Hugo to simplify the generation of these lists.

davidwrede (approved):

PR #5923, opened 2021-05-04
feat: Add versions to static

### LINKED ISSUE #5872: JSON with versions
I've added semver specs to the Pulumi CLI GHA (pulumi/action-install-pulumi-cli#20). In this implementation, I'm downloading all releases directly from the Github API. Since Pulumi has more than 100 releases, we have to paginate in order to get all the versions.

I feel it might be better to maintain a simple JSON that is stored on pulumi.com (e.g. pulumi.com/versions.json). Similar to [latest-version](https://github.com/pulumi/docs/blob/master/static/latest-version). 

I'd happily help with a pull request on it!

cobraz:
Adds `versions.json` to `static/` which can be consumed in Pulumi GHA.

* Do we want any other metadata? Release date? If we don't know, but might be – we should probably add an object (e.g. `[ { "tag": "v1.0.0" }, ... ]`). That way we don't make backwards-incompatible changes _as quickly_.

fixes #5872


cnunciato:
@cobraz @stack72 and I chatted about this last week, and I think what we'd like to do is have this file end up in the `data` folder instead, rather than `static`. The `data` folder is a special folder in Hugo (https://gohugo.io/templates/data-templates/) that's exposed as in-memory data at build-time, which would allow us to render these versions in multiple ways, including to power the Available Versions page. (This is what I meant with [my comment](https://github.com/pulumi/docs/issues/5872#issuecomment-827098729) in #5872.)

@stack72 can elaborate, but if we could expand this dataset in such a way that the array held not only the version numbers themselves but also the metadata associated with each one, we'd have all we need to render a resource at https://pulumi.com/versions, the page at /docs/get-started/install/versions, the file that currently lives at https://pulumi.com/latest-version, etc. The structure of the file could be something like this:

```
[
    {
        "version": "v3.1.0",
        "releaseDate": "2021-05-06",
        "downloads": {
            "linux-x64": {
                "url": "https://get.pulumi.com/releases/sdk/pulumi-v3.2.1-linux-x64.tar.gz",
                "checksum": "879420835243873760a7a4b64e4fff5ec2ac85540d07dd05e143ac2218f02e93"
            },
            "linux-arm64": {
                "url": "https://get.pulumi.com/releases/sdk/pulumi-v3.2.1-linux-arm64.tar.gz",
                "checksum": "af69944aafa899545b9b5949b44c414e82649de1fdc21f56c454bac6bf9a7357"
            },
            "darwin-x64": {
                "url": "https://get.pulumi.com/releases/sdk/pulumi-v3.2.1-darwin-x64.tar.gz",
                "checksum": "d9ef0cecc454585bfd57a17ca04b8b56b6ea496e1b53cae1d3fab60c9de9a70b"
            },
            "darwin-arm64": {
                "url": "https://get.pulumi.com/releases/sdk/pulumi-v3.2.1-darwin-arm64.tar.gz",
                "checksum": "eb94d3918ad5a688eb17ec8fd0edf149ef703329c4470bdb1b785920c6ee46e3"
            }
        }
    },
    // more...
]
```

Decoupling the data from its representations would also make it easier to do things like add optional pagination, and so on, using built-in Hugo features. Thoughts?

stack72 (approved):
This is great - we are going to ship this and immediately start to use it in the GitHub Action

PR #5981, opened 2021-05-12
Only run the automerge workflow when a token is present

cnunciato:
The automerge workflow fails when GITHUB_TOKEN is missing, so this change just skips the workflow when that's the case.

### DIFF
--- a/.github/workflows/automerge-workflow.yml
+++ b/.github/workflows/automerge-workflow.yml
@@ -4,7 +4,8 @@ jobs:
     name: Automerge Workflow
     runs-on: ubuntu-18.04
     steps:
-      - env:
+      - if: env.GITHUB_TOKEN != null
+        env:
           GITHUB_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/automerge-workflow.yml
```diff
@@ -4,7 +4,8 @@ jobs:
     name: Automerge Workflow
     runs-on: ubuntu-18.04
     steps:
-      - env:
+      - if: env.GITHUB_TOKEN != null
```
stack72, line 7: We check for this and then for Pulumi Bot token below - will that cause a second set of issues if not available?

cnunciato: I tested this with https://github.com/pulumi/docs/pull/5981/commits/68eed8278eea7f32e6be00a650b7a747679c51ac and the workflow was skipped, so I'm assuming the workflow will similarly be skipped for runs where GITHUB_TOKEN is absent. If I'm wrong we should be able to revert pretty quickly, though.

stack72 (approved):

PR #6029, opened 2021-05-24
Add missing dependency on oktokit/rest

stack72:


cobraz (approved):
Looks good!

cnunciato:
Ah, I see what happened. Thanks for the quick fix @stack72!

PR #6032, opened 2021-05-24
Exclude local Jaeger endpoint in troubleshooting guide from BLC

leezen:
Ignore the local Jaeger endpoint that's in the troubleshooting guide from the broken link checking


cnunciato (approved):
Thanks!

PR #6035, opened 2021-05-24
Generate a JSON list of versions

cnunciato:
Builds on https://github.com/pulumi/docs/pull/5923 to make this list available at pulumi.com/versions/index.json.


cnunciato:
On second thought, this probably belongs in pulumi-hugo, so will likely close after moving there.

PR #6036, opened 2021-05-24
Add the date to the versions list

cnunciato:
Builds on https://github.com/pulumi/docs/pull/5923 to add release date to the dataset, which will allow us to use the same data to build the Available Versions page that lives at https://www.pulumi.com/docs/get-started/install/versions.

### DIFF
--- a/data/versions.json
+++ b/data/versions.json
@@ -1,6 +1,7 @@
 [
   {
     "version": "v3.3.1",
+    "date": "2021-05-25T08:35:31Z",
     "downloads": {
       "linux-x64": "https://get.pulumi.com/releases/sdk/pulumi-v3.3.1-linux-x64.tar.gz",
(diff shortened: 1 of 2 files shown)


cobraz (commented):
I totally forgot to add this! Looks good :)

We should make sure we add `date` to [the runtype in Pulumi GHA](https://github.com/pulumi/actions/blob/dc3e7a2411e3a04bc47ed6c5005601c2ba8753c7/src/libs/libs/get-version.ts#L6-L17) before merging this.

cnunciato:
@cobraz Please take a look. I've updated to include today's release as well.

cnunciato:
@stack72 Been chatting with @cobraz offline and this should be mergeable as-is, without the change in pulumi/actions. We're working on getting the tests updated over there now and will add documentation for running the tests and updating the mocks as well.

stack72 (approved):

PR #6046, opened 2021-05-25
Fix date issue / Add tagger.date

cobraz:

### DIFF
--- a/scripts/get-versions.js
+++ b/scripts/get-versions.js
@@ -17,6 +17,11 @@ const query = `
         nodes {
           name
           target {
+            ... on Tag {
+              tagger {
+                date
+              }
+            }
             ... on Commit {
               committedDate
             }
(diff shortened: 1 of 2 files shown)


scripts/get-versions.js
```diff
@@ -68,7 +73,7 @@ const main = async () => {
     .filter((tag) => !tag.name.match(/sdk|pkg/))
     .map((tag) => ({
       version: tag.name,
-      date: tag.target.committedDate,
+      date: tag.target?.committedDate || tag.target?.tagger?.date,
```
cnunciato, line 76: So this essentially says, "use `target.committedDate` if possible, otherwise fall back to `tag.tagger.date`"? But if that's not there, this may still be `undefined`, right?

cnunciato (approved):

scripts/get-versions.js
cobraz: Yes, but from what I understand that shouldn't be possible. Haven't found solid documentation on it yet.

PR #6051, opened 2021-05-26
Set NODE_ENV to production for builds

cnunciato:
This should result in smaller CSS bundles.


cnunciato:
Actually this shouldn't be necessary -- we can do it in pulumi-hugo instead.

PR #6105, opened 2021-06-07
fix of documentation - not working python code in AWS pipeline example

ZeoX06:
### Fix of Python code example

Documentation for AWS pipeline creation is not working. Feel free to add the changes to make it working again. 

Thank you.


leezen:
Thanks for the suggestion. These examples are automatically generated, so the fix here won't apply. We'll need to look at the underlying code generation for why it's doing this and will open an issue for this.

Issue #1396, opened 2019-07-18
Python docs styling improvements has regressed


justinvp:
I'd made improvements to the styling in #1247 before the new website went live, but it appears to have regressed in the past week or so. We use very specific CSS selectors to cleanup the styling, so we likely just introduced a new parent/child or sibling that causes the styles to not get picked up.

PR #6158, opened 2021-06-20
Add a test for Python SDK styling

joeduffy:
Our Python SDK styling [has regressed more than once now](
https://github.com/pulumi/docs/issues/1396), because of the subtle way
that the CSS is targeting very specific layouts in the generated docs.
This change adds a test to ensure that it doesn't break in the future.

Dependent on https://github.com/pulumi/pulumi-hugo/pull/352.

### DIFF
--- a/cypress/integration/site_spec.js
+++ b/cypress/integration/site_spec.js
@@ -47,4 +47,18 @@ describe("www.pulumi.com", () => {
             });
         });
     });
+
+    describe("python SDK docs", () => {
+        beforeEach(() => {
+            cy.visit("/docs/reference/pkg/python/pulumi");
+        });
+
+        // Regression test for https://github.com/pulumi/docs/issues/1396, which has happened multiple times.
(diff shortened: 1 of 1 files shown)


cypress/integration/site_spec.js
```diff
@@ -47,4 +47,18 @@ describe("www.pulumi.com", () => {
...
     });
+
+    describe("python SDK docs", () => {
+        beforeEach(() => {
+            cy.visit("/docs/reference/pkg/python/pulumi");
+        });
+
+        // Regression test for https://github.com/pulumi/docs/issues/1396, which has happened multiple times.
+        // The CSS is applied by targeting specific node structures due to the way our Python docs are generated.
+        describe("is styled correctly", () => {
```
susanev, line 58:
```suggestion
        it("is styled correctly", () => {
```

susanev (approved):

PR #6209, opened 2021-07-07
Delete provider resource docs before generating new ones

cnunciato:
### Proposed changes

This change adds a step to `resourcedocsgen` to remove existing provider docs before generating new ones (which should be safe, since we always generate all of them). 

Related to:
* https://github.com/pulumi/pulumi/pull/7450
* https://github.com/pulumi/docs/pull/6209, 

We should merge this PR before https://github.com/pulumi/pulumi/pull/7450, as that PR changes the way we name these files. (Otherwise, we'd end up with ~2x the number of files per provider.)

### DIFF
--- a/tools/resourcedocsgen/main.go
+++ b/tools/resourcedocsgen/main.go
@@ -76,8 +76,10 @@ func main() {
 	}
 
 	// Delete existing docs before generating new ones.
-	fmt.Printf("Deleting %s", outDir)
-	os.RemoveAll(outDir)
+	if err := os.RemoveAll(outDir); err != nil {
+		glog.Infof("error deleting provider directory %v: %v", outDir, err)
+		os.Exit(1)
+	}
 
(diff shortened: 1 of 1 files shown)


tools/resourcedocsgen/main.go
```diff
@@ -74,6 +75,10 @@ func main() {
 		mergeOverlaySchemaSpec(mainSpec, overlaySpec)
 	}
 
+	// Delete existing docs before generating new ones.
+	fmt.Printf("Deleting %s", outDir)
+	os.RemoveAll(outDir)
```
stack72, line 80: Should we check that this doesn't return an error before continuing?

cnunciato: Ah, I didn't even notice that, thanks, yeah. Will add.

stack72 (approved):
💪🏻💪🏻

PR #6243, opened 2021-07-19
Skip a couple troublesome domains from broken link checks

leezen:
There are two domains that consistently fail the broken link check. They're fairly well-known domains that I don't expect to stop working anytime soon, so let's skip them for now.


cnunciato (approved):
Again, thanks for tending to these. 🙏

PR #6283, opened 2021-07-29
Redirect Node.js EKS docs to the EKS resource docs

### LINKED ISSUE #6282: EKS SEO directs users to language specific pages
### Problem description

Before we had multi language EKS, it was nodejs only. When I search for `pulumi eks` in Google, it takes me to this page: https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/eks/

This only shows the nodejs docs, whereas the users should be landing here: https://www.pulumi.com/docs/reference/pkg/eks/cluster/

cnunciato:
When these redirects were written, we didn't yet have resource docs for EKS. Now we do, so we can redirect Googlers to the new location at https://www.pulumi.com/docs/reference/pkg/eks.

We'll also want to remove these from Swiftype results; will look into what needs to be done for that and file follow-ups as needed.

Fixes #6282.

### DIFF
--- a/infrastructure/cloudfrontLambdaAssociations.ts
+++ b/infrastructure/cloudfrontLambdaAssociations.ts
@@ -160,7 +160,6 @@ function nodeSDKRedirect(uri: string): string | undefined {
         "cloud",
         "policy",
         "awsx",
-        "eks",
         "kubernetesx",
         "terraform",
     ];


praneetloke (commented):
Awesome!!

praneetloke (approved):

PR #6286, opened 2021-07-31
Stop building draft and future-dated pages in PRs

cnunciato:
Now that all of our hand-written content happens in pulumi-hugo, and we build drafts and future-dated pages for PR previews there, we no longer need to do the same thing here (and indeed, it's somewhat confusing that we do). This change just removes those two `hugo` options.


mattstratton (approved):
Thank you!

PR #6342, opened 2021-08-12
fix: add missing pkg to GetAvailabilityZonesArgs

vranystepan:
### Proposed changes

Add missing package name (`aws`) to GetAvailabilityZonesArgs from Go library.


leezen:
Thanks for the suggestion. The example code is automatically generated, so I'll file an issue for this instead so we can look at this from a codegen perspective.

PR #6352, opened 2021-08-15
fix(login): update api url

k-jun:
### Proposed changes

update api url. as shown below, can't resolve the domain.

```bash
$ host app.pulumi.acmecorp.com
Host app.pulumi.acmecorp.com not found: 3(NXDOMAIN)
```


emiliza:
This example is auto-generated from: https://github.com/pulumi/pulumi/blob/master/pkg/cmd/pulumi/login.go#L53-L55, so we'd need to pull the fix there.

However, this is correct. It's supposed to be an example if you wanted to run a self-hosted service on your own domain, eg: acmecorp.com, then this is how you'd login.

PR #6353, opened 2021-08-16
Update valid values for CloudFront distribution function event type

NartyNarts:
### Proposed changes

Added two missing valid values for CloudFront distribution function event types.


leezen:
Thanks for the suggested fix. Unfortunately, this looks like a bug in the doc generation code where we're overwriting the lambda function association with the function association property's documentation instead. I'll open up an issue to track this.

PR #6381, opened 2021-08-21
Update resourcedocsgen tool to also generate data files containing package metadata

praneetloke:
### Proposed changes

The new data files generated by the tool will be used by for docs.

* Switches the `resourcedocsgen` tool to use `spf13/cobra` to support different modes of operation for the tool
* Adds a new script to invoke the `resourcedocsgen` tool with the new `metadata` command. I am not checking in the script that calls this script for now as I have some things to work through. Specifically, I am looking at consolidate all the scripts related to resource docs generation but I want to do that in a separate PR to keep changes as minimum as possible in this PR
* Update the existing resource docs script to invoke the tool using the new format. Nothing changes for consumers of the script itself and is transparent to our existing CI processes.

N/A

### Related issues (optional)

https://github.com/pulumi/pulumi-service/issues/7091


tools/resourcedocsgen/cmd/docs.go
```diff
@@ -0,0 +1,201 @@
+package cmd
```
praneetloke, line 1: This file is basically everything that was in `main.go`. It's now specific to generating the resource docs from the unmarshalled main `PackageSpec`.

tools/resourcedocsgen/cmd/metadata.go
```diff
@@ -0,0 +1,61 @@
+package cmd
```
praneetloke, line 1: This is really the net-new code (aside from all the code organization as `cobra` commands, of course) that deals with generating the package metadata files.

tools/resourcedocsgen/pkg/metadata.go
```diff
@@ -0,0 +1,44 @@
+package pkg
```
praneetloke, line 1: I defined this in `pkg` thinking that we might have a use for it with https://github.com/pulumi/pulumi/issues/7813 but then maybe not and we'll probably discard/relocate this and use what's in `pulumi/pulumi` if this info becomes available somewhere else.

tools/resourcedocsgen/pkg/metadata.go
```diff
@@ -0,0 +1,44 @@
...
+type PackageCategory string
+
+const (
+	// PackageStatusGA indicates that a package is generally available for use.
+	PackageStatusGA PackageStatus = "ga"
+	// PackageStatusPublicPreview indicates that a package is available as
+	// pre-release and can undergo changes before it goes GA.
+	PackageStatusPublicPreview PackageStatus = "public_preview"
+
+	PackageCategoryCloud    PackageCategory = "Cloud"
```
praneetloke, line 17: This is not exhaustive at the moment and in fact all packages get categorized as Cloud instead of their respective categories as seen on our API Reference and Cloud Providers pages today.

scripts/gen_all_resource_docs.sh
```diff
@@ -3,10 +3,15 @@
...
 
-# Pass any value to the script as the first to indicate that you
+# Pass true to the script as the first argument to indicate that you
 # also this script to commit each provider's docs changes automatically.
 GIT_COMMIT=${1:-}
 
+# Pass an output dir for package metadata as the second argument to
+# additionally generate the package metadata for each package. Package
+# metadata will not be generated if an output dir is not passed.
+PKG_METADATA_OUT_DIR=${2:-}
```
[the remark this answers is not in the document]
praneetloke: We would always want to generate the metadata along with the resource docs. But for the moment, we won't generate metadata as part of our workflow. I'll generate them on-demand while I work on our docs stuff.

scripts/gen_all_resource_docs.sh
```diff
@@ -74,16 +79,20 @@ if [ "$branch" = "master" ]; then
...
 for REPO in "${REPOS[@]}" ; do \
-    ./scripts/gen_resource_docs.sh "$REPO" true
+    ./scripts/gen_resource_docs.sh "${REPO}" true
+
+    if [ -n "${PKG_METADATA_OUT_DIR:-}" ]; then
+      ./scripts/gen_package_metadata.sh "${PKG_METADATA_OUT_DIR}" "${REPO}"
+    fi
 
-    if [ -n "${GIT_COMMIT:-}" ]; then
+    if [ "${GIT_COMMIT:-}" == "true" ]; then
```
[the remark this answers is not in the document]
praneetloke: Since the left side is quoted `"true"` we are doing a string comparison.

scripts/gen_package_metadata.sh
```diff
@@ -0,0 +1,108 @@
...
+
+set -o nounset -o errexit -o pipefail
+
+# The first argument is for the output dir (required).
+METADATA_OUT_DIR=${1:-}
+# The second argument is the override for the package for which this script will generate the metadata.
+# Must not be passed without the "pulumi-" prefix.
+REPO_OVERRIDE=${2:-}
+# Pass a 3rd argument to override the package version used by this script.
+VERSION=${3:-}
```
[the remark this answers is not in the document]
praneetloke: Ah yes! I'll add that.

tools/resourcedocsgen/cmd/metadata.go
```diff
@@ -0,0 +1,127 @@
...
+				UpdatedOn:     time.Now().Unix(),
+				Publisher:     "Pulumi",
+				Title:         mainSpec.Name,
+				Description:   mainSpec.Description,
+				Category:      category,
+				PackageStatus: status,
+				Featured:      featured,
+				Native:        mainSpec.Attribution == "",
+				Version:       version,
+				Logo:          fmt.Sprintf("/logos/pkg/%s.svg", mainSpec.Name),
```
cnunciato, line 105: Let's tweak this line to exclude the specifics about these file paths and extensions and just expect that the templates in the registry repo will use the package name to assemble the image references.

go.sum
```diff
@@ -1,312 +1 @@
-github.com/pulumi/pulumi-hugo/themes/default v0.0.0-20210420180605-5a9c775a3d1c h1:ODVskkF6w7jbwIg67q4aOsc9N5524AoyA0QhJnSnK+0=
```
praneetloke, line 1: So I am about to merge this PR because I don't want to keep having to rebase this PR every time there is a new docs site build and there is a conflict with this `go.sum`. I do have a question about this file's contents in `master`, though. Why does it have every single past version of the `pulumi-hugo` module in it? It should only have the current version, correct? I ran `go mod download` on my machine and that knocked out all the previous versions and only the current version referenced by `go.mod` is in here now. cc @cnunciato

praneetloke: I went ahead and reverted this for now since this seems to be something we should fix in our cron workflow that updates this file. I didn't want to introduce a large diff in the next automated PR for this file.

PR #6451, opened 2021-09-14
typo fix for aws wafv2 reference

ertenmehmet:
my colleague has bumped into this issue while trying to create a wafv2 stack in aws. I thought this might be annoying, since waf is already a concept more complicated than many others


leezen:
Thank you for the suggested fix. This part of the documentation is auto-generated and I've filed a corresponding issue to look at the underlying code generation that causes this.

PR #6486, opened 2021-09-23
Add PurgeCSS, un-vendor pulumi-hugo, etc.

cnunciato:
Bunch of things in here, but all related:

* **Stop vendoring pulumi-hugo**. We no longer need to do this, as we're now building our CSS and JS bundles in pulumi/theme, which means we can just pull them into the Hugo build as regular Hugo `assets`. This change adjusts the Hugo config and build scripts accordingly.

* **Add PurgeCSS as a build step**. We use PurgeCSS today to filter out unused CSS, but we do so at the pulumi-hugo level. This change just moves that step to run after the website build instead -- which makes more sense, anyway, since it means PurgeCSS can analyze the actual content of the rendered website, and not just an approximation of that content through Hugo templates, Markdown files, etc.

* **Stop publishing Sentry releases**. We can't do this quite as cleanly here, with the JS build now happening in pulumi/theme (which leaves the JS/TS source files behind). But in all honesty, we weren't really making use of Sentry releases, anyway, so I'm comfortable just removing this rather than spending the time to plumb everything through; the Sentry errors we get are useful enough by themselves without having a historical mapping between a given error and the specific website release that caused it. But if we _did_ want to keep publishing these, we could certainly do so in a follow-up.

* **Remove markdown_seo_lint.js**. We're doing [the same thing in pulumi-hugo](https://github.com/pulumi/pulumi-hugo/blob/master/scripts/lint/lint-markdown.js), so we don't need to do it here.

* Remove the local reference in `go.mod` in order to rely on fetching private modules with GOPRIVATE and git config.


cnunciato:
Actually, looks like I missed checking in the Hugo config -- ⏳..

config/_default/config.yml
```diff
@@ -27,18 +27,6 @@ module:
...
           target: data
-        - source: assets
-          target: assets
-        - source: archetypes
-          target: archetypes
-        - source: components
-          target: assets
-        - source: package.json
-          target: assets
-        - source: tsconfig.json
```
djgrove, line 38: For my benefit--can we remove these from `registry` here also, or is there a distinction?

scripts/ensure.sh
```diff
@@ -13,6 +13,3 @@ fi
 # Install the Node dependencies for the website and the infrastructure.
 yarn install
 yarn --cwd infrastructure install
-
-# Fetch Hugo modules.
-hugo mod vendor
```
djgrove, line 18: One other thing for my understanding--what is the benefit of vendoring if it's essentially optional in the build process?

djgrove (approved):

djgrove:
This is great!! What was the impact on bundle size?

config/_default/config.yml
cnunciato: Yes! I'm about to push up a commit that does just that. (Just noticed them myself.)

praneetloke (approved):
TIL: When I saw that you were removing the go mod override, I went looking at the workflow in this repo and realized that it uses the [personal access token method](https://github.com/pulumi/docs/blob/master/.github/workflows/update-theme.yml#L31-L33) to configure git.

cnunciato:
> What was the impact on bundle size?

Brings it down from ~3MB to ~30K minified & gzipped.

PR #6553, opened 2021-10-01
Update gen_resource_docs.sh to add package tree files to the static dir

praneetloke:
### Proposed changes

These files are static because we'll load them on-demand. They won't be consumed to generate any static markup at this time.

CloudFront does seem to automatically compress [`application/json` content-type](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/ServingCompressedFiles.html#compressed-content-cloudfront-file-types) (I am assuming the correct content-type will be set on the file when it is uploaded to S3 since the file extension hints at it being json) already, so I don't think we need to update our CFN settings.


scripts/gen_resource_docs.sh
```diff
@@ -14,7 +14,7 @@ INSTALL_RESOURCE_PLUGIN_VERSION=${3:-}
 
 PACKDIR="./content/docs/reference/pkg"
 ABSOLUTEPACKDIR="$(pwd)/content/docs/reference/pkg"
-PACKAGE_TREE_OUT_DIR="../../../registry/packages/data/navs"
+PACKAGE_TREE_OUT_DIR="../../static/registry/packages/navs"
```
praneetloke, line 17: This also has the following benefits:
* co-located with the API docs
* will be kept up-to-date with the API docs whenever we generate those. That is, we already have a workflow for this repo to generate docs and open a PR for updated docs. So now that'll include updated package tree files for the respective package.

kimberleyamackenzie (commented):
:shipit:

scripts/gen_resource_docs.sh
cnunciato: What would you think about making this path be just `https://pulumi.com/registry/packages/aws-native.json`, for example? In other words, remove `navs`, since these files are essentially JSON representations of a package? Maybe that's reaching a bit, but I only mention it thinking these JSON files could be used for something other than just the registry tree nav. Know what I mean?

cnunciato (approved):
YES. 🚀 One little question about naming (because isn't there always a question about naming?), but needn't block, just food for thought.

scripts/gen_resource_docs.sh
praneetloke: > In other words, remove navs, since these files are essentially JSON representations of a package? 

Not entirely true. Yes, it is a JSON of the modules, resources and functions of a package but it doesn't entirely constitute a package as a whole. I guess it depends on how you see it. To me, these are the "structural" JSON files of a package.

> but I only mention it thinking these JSON files could be used for something other than just the registry tree nav. Know what I mean?

Ah I see what you mean. How about instead of `navs`, calling it `maps`? It could be considered a package map. I think `https://pulumi.com/registry/packages/aws-native.json` gives the impression that it is the package itself. In fact, would that conflict with the main path of the registry at `pulumi.com/registry`? I was trying to confirm that by looking for static path mappings and couldn't find them. How do we map the `/static` paths to the actual S3 objects? Perhaps, I should move this one more level down under a `json` dir inside `static`. So these are static json files specific to the registry. So the final path would be `https://pulumi.com/json/registry/packages/map/aws-native.json`.

cnunciato: Honestly I think we could just merge this as-is and come back to it when we do have that more-general-purpose use case. Let's :shipit:. 😄

PR #6554, opened 2021-10-01
Update provider names

infin8x:
Did a second pass over these names to true them up with the list of packages we're tracking, ensure capitalization is correct, etc.


praneetloke:
@infin8x it would be awesome if you could also update https://github.com/pulumi/pulumi/blob/master/pkg/codegen/docs/gen.go#L74 over in `pulumi/pulumi`?

infin8x:
@praneetloke ah thanks. Done: https://github.com/pulumi/pulumi/pull/8119

praneetloke (approved):

PR #6576, opened 2021-10-05
Generate package nav files

praneetloke:
### Proposed changes

These packages were previously unaccounted for in the list of packages in `gen_all_resource_docs.sh` and were actually added to that list in the last couple of days. While I generated the metadata files for those I didn't specifically generate the nav files for them.

Note that since these packages existed before I introduced the changes in `resourcedocsgen` to generate the navs, they were not automatically generated in our CI/CD automation for provider docs releases. But they will be anytime a new version is released or a new provider is introduced.


kimberleyamackenzie (approved):
:shipit:

PR #6598, opened 2021-10-07
Fix gcp account

kevinneville:
### Proposed changes
When following the guide for Python I encountered errors. 
This is now fixed.

I also made the other languages consistent.

### DIFF
--- a/content/docs/reference/pkg/gcp/serviceAccount/account/_index.md
+++ b/content/docs/reference/pkg/gcp/serviceAccount/account/_index.md
@@ -1,8 +1,8 @@
 
 ---
 title: "Account"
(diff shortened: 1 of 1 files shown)


leezen:
This is tracked in https://github.com/pulumi/pulumi-hugo/issues/477 and unfortunately a fix here won't be the right change as ultimately what needs to change is the code generation to produces this output (any changes here will end up getting overwritten on the next provider update).

kevinneville:
@leezen Aa make sense. Thank you!

leezen:
@kevinneville Thanks for submitting the fix in the first place! Really appreciate you taking a look.

PR #6612, opened 2021-10-08
Adjust the default PurgeCSS extractor

cnunciato:
Adjusts the PurgeCSS `defaultExtractor` regex to allow for decimal points in utility classnames (a Tailwind feature introduced after this regex was written).

Fixes https://github.com/pulumi/registry/issues/119 and a few of the others we've identified. Verified with a full docs build:

image

image

image

Final bundle size still comes in at ~31K gzipped.

### DIFF
--- a/scripts/minify-css.js
+++ b/scripts/minify-css.js
@@ -39,8 +39,8 @@ postcss([
         // We need to extract the Tailwind screen size selectors (e.g. sm, md, lg)
         // so that we do not strip them out. As long as a class name appears in the HTML
         // in its entirety, PurgeCSS will not remove it.
-        // Ex. https://tailwindcss.com/docs/controlling-file-size/#writing-purgeable-html
-        defaultExtractor: content => content.match(/[\w-/:]*[\w-/:]/g) || [],
+        // Ex. https://tailwindcss.com/docs/optimizing-for-production#writing-purgeable-html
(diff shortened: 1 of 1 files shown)


scripts/minify-css.js
```diff
@@ -39,8 +39,8 @@ postcss([
         // We need to extract the Tailwind screen size selectors (e.g. sm, md, lg)
         // so that we do not strip them out. As long as a class name appears in the HTML
         // in its entirety, PurgeCSS will not remove it.
-        // Ex. https://tailwindcss.com/docs/controlling-file-size/#writing-purgeable-html
-        defaultExtractor: content => content.match(/[\w-/:]*[\w-/:]/g) || [],
+        // Ex. https://tailwindcss.com/docs/optimizing-for-production#writing-purgeable-html
+        defaultExtractor: content => content.match(/[\w-/.:]+(?<!:)/g) || [],
```
praneetloke, line 43: A comment as to what the regex is trying to match (with some examples) would be great.

praneetloke (approved):
🎉

scripts/minify-css.js
cnunciato: Ack, I merged before seeing this comment. I'll add a few words to the existing comment calling out colons and dots.

PR #6630, opened 2021-10-13
Add sitemap checker script

djgrove:
Created a sitemap checker script that will scan all URLs in the `pulumi.com` sitemap and attempt to request them on the `www-reg-staging.pulumi-dev.io` site.

Worth noting--it runs quite slowly (obviously). On my machine it probably took somewhere around 30ish minutes to run, so it might be worth running this concurrently. I'm not certain what we'd like our constraints to be there. cc: @EvanBoyle @cnunciato 

## Issue
Fixes [#56](https://github.com/pulumi/registry/issues/56)

## 4xx pages
Running this script produced these errors:
```
403 https://www-reg-staging.pulumi-dev.io/blog/azure-container-solutions/
403 https://www-reg-staging.pulumi-dev.io/blog/pulumi-kubernetes-operator-1-0/
000 https://www-reg-staging.pulumi-dev.io/docs/reference/pkg/azure-native/datafactory/getexposurecontrolfeaturevalue/
403 https://www-reg-staging.pulumi-dev.io/docs/reference/pkg/azure-native/automation/getpython2package/
000 https://www-reg-staging.pulumi-dev.io/docs/reference/pkg/gcp/securitycenter/notificationconfig/
403 https://www-reg-staging.pulumi-dev.io/docs/reference/pkg/azure-native/automation/python2package/
```

It's worth noting that both of those `000` codes mean [cUrl failed to execute](https://superuser.com/questions/501690/curl-http-code-of-000)--and could be caused by either a timeout (likely) or an inability to resolve the DNS. Both of those pages are 301 redirects that appear to work when I request them manually.

Also, it looks like all the 403s (404s technically, I think some implementation detail to Hugo is surfacing these as 403 forbidden to `cUrl`) are brand new pages, and so we just haven't updated yet.


cnunciato:
Just a thought, it'd be sweet if this had a verbose mode that gave you a sense of what the script is doing, since it takes so long to run it.

cnunciato:
... also, not for now but for a V2 kind of thing, you might think about doing this as a Node script instead, just for the additional flexibility and maintainability. I know for my part, doing things like this in Bash can quickly become a lot more trouble than it's worth! 😄

scripts/validate-sitemap-urls.sh
```diff
@@ -0,0 +1,17 @@
...
+request_and_format_sitemap() {
+    curl https://www.pulumi.com/sitemap.xml > sitemap.xml && xmllint --format sitemap.xml > temp_sitemap.xml && mv temp_sitemap.xml sitemap.xml
+}
+
+parse_urls() {
+    cat sitemap.xml | grep -e loc | sed 's|<loc>\(.*\)<\/loc>$|\1|g' | sed 's/www.pulumi.com/www-reg-staging.pulumi-dev.io/g' > site_urls.txt
+}
+
+request_urls() {
+    cat site_urls.txt | xargs -I {} curl -s -o /dev/null -w "%{http_code} %{url_effective}\n" {} | grep -v '200\|301'
```
cnunciato, line 12: We might want to make the regex a little bit tighter just to make sure we don't accidentally pick up something with 200 or 301 in the URL -- not that we have any such thing yet, but you know, just in case.

cnunciato: This might do the trick (of telling `curl` to follow redirects):

```suggestion
    cat site_urls.txt | xargs -I {} curl -L -s -o /dev/null -w "%{http_code} %{url_effective}\n" {} | grep -v '200\|301'
```

It's surfacing a handful of 403s for me now, where it wasn't before.

scripts/validate-sitemap-urls.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+request_and_format_sitemap() {
+    curl https://www.pulumi.com/sitemap.xml > sitemap.xml && xmllint --format sitemap.xml > temp_sitemap.xml && mv temp_sitemap.xml sitemap.xml
```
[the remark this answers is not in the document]
djgrove: It actually adds some newlines to the sitemap, which made it easier to parse with `grep` and `sed`

PR #6653, opened 2021-10-16
Add theme and reg to the hourly module job

cnunciato:
Adds two new modules to the hourly PR job that checks for new module content, and updates the script that adds GitHub commit URLs to those PR descriptions.


praneetloke (approved):
🚀

PR #6788, opened 2021-11-10
Generate API docs during build

praneetloke:
### Proposed changes

This PR removes the API docs and nav tree JSON files from version control.

To support the plan outlined in https://github.com/pulumi/registry/issues/237#issuecomment-963384022 `resourcedocsgen` has been updated with a new command under the existing `docs` command called `registry`. The new command allows us to generate API docs for all packages from the metadata files in the registry. In other words, we'd be generating API docs for all packages from a snapshot of information about them.

The way this works is the tool clones only the registry repo and nothing else. It then looks at every metadata file, reading it and then calculating the the path to the schema.json|yaml file in the respective package's repo on some VCS (for now, only GH is supported). With the path to the schema file calculated, it downloads that via the raw GH user content URL and unmarshals it to a schema spec which can then be passed to the docs generator which is all that the docs generator has always needed. The schema spec. No need to clone package repos or have them in a certain path anymore.

With the above, we've had to make the change to the docs generator to use `go:embed` so that we can pin the version of the `pulumi/pulumi/pkg` dep in the tool so that one could essentially run, for example:

```
go install -o "${GOPATH}/bin/resourcedocsgen
resourcedocsgen docs registry --baseDocsOutDir "themes/default/content/registry/packages" --basePackageTreeJSONOutDir "themes/default/static/registry/packages/navs" --logtostderr
```
to generate API docs for packages to a path in registry for local development, if needed.

#### Other changes in this PR

* Update the `metadata` command to also include the `repo_url` and `schema_file_path` properties in the registry metadata file
* Update the resource docs workflow that responds to `repository_dispatch` events to no longer build resource docs. It still does everything else it does today.
* Update `build-site.sh` to invoke the `resourcedocsgen` tool before building the site with `hugo`
* Use `go:embed` to embed the `overlays` directory, which means I needed to move the `overlays` dir from the root of `resourcedocsgen` into `resourcedocsgen/cmd/docs`.
* I didn't realize this before but a dependency I had added to that tool to support YAML schema files was essentially a wrapper around another dep that we were using. So I eliminated the other dep and chose to keep a single dependency throughout. The dep that the `resourcedocsgen` uses now does have a requirement to use `json` struct tags instead of `yaml` to serialize the fields. That feels a bit odd but not too bad I think. The reason for using this is so that we can convert YAML-based Pulumi schema to JSON easily and then unmarshal that to our Pulumi schema spec.

### Related issues (optional)

pulumi/registry#237

This PR needs https://github.com/pulumi/registry/pull/326 to be merged and the `docs` repo updated with the latest from `registry`.


praneetloke:
I have split the commits so it is slightly easier to review. Obviously, don't try to click on the "Remove API docs and nav tree JSON files" commit 😄

praneetloke:
Let me know if any of you would like to hop on a Zoom and review this together.

tools/resourcedocsgen/cmd/docs/registry.go
```diff
@@ -215,6 +215,7 @@ func resourceDocsFromRegistryCmd() *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&baseDocsOutDir, "baseDocsOutDir", "../../content/registry/packages", "The directory path to where the docs will be written to")
+	cmd.Flags().StringVarP(&branch, "branch", "b", "master", "The branch at which the registry repo will be checked out.")
```
cnunciato: Do we need this? I ask because a bit ago, after running `hugo mod get` to use your `praneetloke/regen-metadata` branch, I ran `make build`, and saw this in the logs:

```
I1112 11:31:14.667116   82648 registry.go:155] Cloning the registry repo @ master
I1112 11:31:17.552228   82648 registry.go:162] Cloned the repo successfully!
I1112 11:31:17.552255   82648 registry.go:165] Resetting registry repo HEAD to b91845d16a01
HEAD is now at b91845d1 Merge branch 'master' into praneetloke/regen-metadata
```

So it looks like branch isn't being used, so may not be necessary. What do you think? Do we need it?

praneetloke: Yeah the flag itself isn't used by `build-site.sh`. The second message you are seeing in from `git` itself as a result of running `git reset --hard <commit>`. We don't need the `branch` flag in that it is strictly not needed for our purpose but I can see it being helpful for development purposes where someone wants to specify the branch from which to regen docs rather than a specific commit. If you specify both flags, then it becomes "checkout registry repo at that branch _and_ also reset the branch to that commit" but again we don't use that in CLI. However, it could be useful for development purposes.

stack72 (approved):

praneetloke:
Rebased this PR with the latest from `master` now that the registry PR has been absorbed into this repo with https://github.com/pulumi/docs/pull/6840. The PR build should pass now as it should now be able to build API docs purely from the metadata files alone.

praneetloke:
The API docs generation step itself passed but the build failed with the following. It appears that there is a problem fetching a Go dependency from GH. I think I saw other PRs in this repo affected by that. I'll retry the build shortly.

```
 go: gopkg.in/yaml.v2@v2.4.0: reading gopkg.in/yaml.v2/go.mod at revision v2.4.0: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /tmp/hugo_cache/modules/filecache/modules/pkg/mod/cache/vcs/748bced43cf7672b862fbc52430e98581510f4f2c34fb30c0064b7102a68ae2c: exit status 128:
	remote: Cannot obtain refs from GitHub: cannot talk to GitHub: Get https://github.com/go-yaml/yaml.git/info/refs?service=git-upload-pack: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
hugo: collected modules in 138957 ms
	fatal: unable to access 'https://gopkg.in/yaml.v2/': The requested URL returned error: 502
Error: failed to download modules: go command failed: go: gopkg.in/yaml.v2@v2.4.0: reading gopkg.in/yaml.v2/go.mod at revision v2.4.0: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /tmp/hugo_cache/modules/filecache/modules/pkg/mod/cache/vcs/748bced43cf7672b862fbc52430e98581510f4f2c34fb30c0064b7102a68ae2c: exit status 128:
	remote: Cannot obtain refs from GitHub: cannot talk to GitHub: Get https://github.com/go-yaml/yaml.git/info/refs?service=git-upload-pack: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
	fatal: unable to access 'https://gopkg.in/yaml.v2/': The requested URL returned error: 502
```

praneetloke:
Looks like it is not just us. https://github.com/niemeyer/gopkg/issues/63#issuecomment-968616241 It looks like an issue with `gopkg.in` trying to talk to GH.

praneetloke:
Alright the preview is ready for this PR 🎉

cnunciato (approved):
🚀

PR #6827, opened 2021-11-13
Exclude GitHub edit and request links from the link checker

cnunciato:
Noticed these while debugging the link checker. These edit-this-page links slow down the link checker significantly, and we don’t really need to check them — note that we’ve pretty much always excluded the same URLs for the docs repo — so this change adds exclusions for the registry and pulumi-hugo repos as well.


praneetloke (approved):
Oh wow that's interesting! Why are these links a problem for the link checker?

cnunciato:
@praneetloke yes! As we know, most of the pages of the website (except for marketing-type pages and those with the “no edit this page” param set to true) have one or two links to their corresponding files on GitHub, and each of those GitHub URLs requires the user to be logged in. That means that for every page of the site that links to GitHub in this way, the link checker makes up to two external requests — and then for each of those requests, a second request because the link checker isn’t logged into GitHub, so gets redirected. So ultimately, we end up making like tens of thousands unnecessary requests. And while the link checker does a pretty good job of remembering the links it’s already checked, it can’t do much in these cases, because every page is unique.

PR #6996, opened 2022-01-10
Add aws-apigateway to mktutorial, regenerate all

cnunciato:
This PR adds `aws-apigateway` to the list of how-to guides that we generate with `mktutorial.sh`, and regenrates all how-to guides as well.

Fixes https://github.com/pulumi/registry/issues/530 (🤞).


tools/mktutorial/main.go
```diff
@@ -28,8 +28,6 @@ import (
 	"github.com/russross/blackfriday/v2"
 )
 
-// clouds contains an index of the clouds for which we want to publish tutorials.
-var clouds = []string{"aws", "aws-native", "azure", "classic-azure", "gcp", "kubernetes"}
```
cnunciato, line 32: Removed these lines because they weren't being used.

susanev:
is it expected that the count is 3 and only 1 shows up?
Screen Shot 2022-01-10 at 2 58 50 PM

cnunciato:
Interesting -- there are three, yeah, they're just in the other language tabs (one in Py, one in Go). I suspect this is a latent bug in that most/all of the other examples we have are only in one language, so we just haven't noticed it. 😄 Probably deserves an issue of its own as it definitely looks a bit odd.

susanev:
@cnunciato wait, isnt it the same example though just in a different language 🤷 
i think i don't understand our counts

susanev (approved):

cnunciato:
> isnt it the same example though just in a different language

Conceptually yes, but techncially no --- they are three different folders in the repo (so currently treated as three different "how-to guides"):

image

PR #7077, opened 2022-02-02
Update awsx nodejs docs to v0.32.0 of the SDK

leezen:
See https://github.com/pulumi/pulumi-awsx/pull/644#issuecomment-1028467861 for context. It appears the last few times we've run docs updates for this provider, it hasn't worked (e.g. https://github.com/pulumi/docs/runs/3750842256?check_suite_focus=true).

For whatever reason, at the time of writing, I cannot retrieve the previous GHA logs to see what went wrong. For now, I propose we accept this manual change to bring the docs up to date.

I also noticed an odd path issue when generating these docs locally (see the update made to `run_typedoc.sh`). I think this is a bug unless we expected the provider source to sit at a different location. But the fact that we push two directories deep (`./tools/tscdocgen`) should mean we need to include those two directories back out as part of the path.


content/docs/reference/pkg/nodejs/pulumi/awsx/apigateway/_index.md
```diff
@@ -1575,95 +1580,39 @@ This value will be JSON.stringify'd as part of normal processing.  It should not
...
-The name of the REST API.  Defaults to the name of the awsx.apigateway.Api if unspecified.
-
-<h4 class="pdoc-member-header" id="RestApiArgs-policy">
-<a class="pdoc-child-name" href="https://github.com/pulumi/pulumi-awsx/blob/7e3319acc1119743eb7e923349542ed966380c18/nodejs/awsx/apigateway/api.ts#L439">property <b>policy</b></a>
-</h4>
-
-<pre class="highlight"><code><span class='kd'></span>policy?: <a href='/docs/reference/pkg/nodejs/pulumi/pulumi/#Input'>pulumi.Input</a>&lt;<span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String'>string</a></span>&gt;;</code></pre>
-
-JSON formatted policy document that controls access to the API Gateway.
-
```
leezen, line 1664: It's worth noting that this all goes away because we now reference the type from `@pulumi/aws` directly in `@pulumi/awsx`:

```typescript
export type RestApiArgs = aws.apigateway.RestApiArgs;
```

stack72 (approved):

PR #7249, opened 2022-03-18
Turn off Goldmark auto-linking

cnunciato:
This change disables [Goldmark's `linkify` extension](https://pkg.go.dev/github.com/yuin/goldmark#readme-linkify-extension), which converts strings that look like URLs into links. 

Today, given the following Markdown: 

```markdown
# Some Heading

Some text with http://example.com in it.  
```

you get this:

```html
<h1>Some Heading</h1>
<p>Some text with <a href="http://example.com">http://example.com</a> in it. </p>
```

However, we don't want this, for reasons described in https://github.com/pulumi/marketing/issues/124. With this change, we'll instead get this:

```html
<h1>Some Heading</h1>
<p>Some text with http://example.com in it.</p>
```

I feel okay about this change and I don't think authors will be surprised by it, as our practice has always been to encode links explicitly, whether with Markdown expressions, `<a href>`s, Hugo `{{< relref >}}`s, or the like.

We'll want to do this in module repos as well (pulumi-hugo, registry), but it's this repo'd config that manages how we build and deploy the website. (Module configs only apply to local development and PR previews.)

Fixes https://github.com/pulumi/marketing/issues/124.

### DIFF
--- a/config/_default/config.yml
+++ b/config/_default/config.yml
@@ -86,7 +86,7 @@ markup:
     extensions:
       definitionList: true
       footnote: true
-      linkify: true
+      linkify: false
       strikethrough: true
       table: true
       taskList: true


susanev:
@cnunciato do you have a sense of how many links (that we don't have concerns about) will no longer be interactive with this change?

cnunciato:
@susanev Many thousands, but nearly all are in the property descriptions of the resource docs. Here's a somewhat messy but representative list -- it's probably not perfect, but gives you an idea what we're talking about: https://gist.github.com/cnunciato/b4956603eb343c9d1ad6b05935ea9c89

For example here: https://www.pulumi.com/registry/packages/google-native/api-docs/compute/v1/instance/index.html#sourceinstancetemplate_nodejs

image

"Have concerns" is subjective. For my part, I definitely think it's  _nice_ to link some of these things, but not something users will expect, and presents problems like these with content we don't fully control.

susanev:
thx! and by "have concerns" i was trying to refer to the links mentioned in https://github.com/pulumi/marketing/issues/124.

Brant-Soler (approved):
Elegant, simple, and clean.

PR #7316, opened 2022-04-06
Fix for heap out of memory errors on build and deploy step.

kimberleyamackenzie:
### Proposed changes
Applies a fix for the sporadic JS heap out of memory failures we've been seeing in docs (see https://github.com/pulumi/docs/runs/5819314607?check_suite_focus=true as an example) in the build and deploy step.


cnunciato (approved):
LGTM

PR #7325, opened 2022-04-07
Skip the PR build for module updates

cnunciato:
This PR makes some adjustments to the hourly job we run to pick up changes from module repos like [pulumi-hugo](https://github.com/pulumi/pulumi-hugo), [registry](https://github.com/pulumi/registry), and [theme](https://github.com/pulumi/theme).

Today, this job runs `hugo mod get` for all three of these repos to figure out whether any of them has an update to deploy, and if it finds anything, aggregates the changes into a single PR, runs a build, and when that build succeeds, merges the PR automatically to kick off a new deployment. A previous version of this workflow stopped short of the automatic merge step, the idea being that a human might want to review these aggregate PRs before merging them -- but over time, as we gained confidence in these builds and started merging with little or no review, it became clear we might as well merge them automatically. And so far that's worked out okay.

What this means, though, is that we're effectively running these PR builds (which can take upwards of 30-45 minutes) for no real reason other than to prove they'll produce a viable build -- which we do anyway as part of the normal, post-merge build-and-deploy process. So this PR cuts out this step and adjusts the hourly job to run against `master` directly, committing any changes to `go.*` back to master to trigger a deployment. In practice, this means that any change merged to any module repo can be in production in as little as 20 minutes or so after a merge.

One downside to this change is that we'll no longer have a way to preview these aggregate PR builds as we do today (e.g., by temporarily disabling the auto-merge workflow), which does come in handy from time to time. This is probably okay, but for those rare situations where we do want to let things queue up without merging, we'll have to come up with an alternative mechanism.  

I've also left the scheduled part out of this workflow for now, as I'd like to run it manually first, probably over a weekend or off hours, to make sure I don't block anyone and it works as expected.


susanev (approved):

stack72:
FWIW, I think this is risky - the PRs do sometimes find that there are issues that were not caught in the hugo / registry PRs

cnunciato:
@stack72 I do agree. I wanted to get this out there for some discussion, though, as I know some folks don't love how long it takes to get content out -- but I agree there's more safety in the process we have today. Any thoughts you have on this would be great to add here.

cnunciato:
I'm going to close this for now, as it doesn't seem to have strong support. Happy to come back to it when it feels more necessary, though.

PR #7378, opened 2022-04-20
Remove the link-check step from the build-and-deploy script

cnunciato:
This step has been timing out sporadically lately, causing backups in the deployment queue. I'm not sure why that's happening -- we may just be outgrowing broken-link-checker -- but while I investigate, I'd like to remove it in order to keep from blocking deployments that would otherwise pass.

@stack72 I'd most like to hear from you on this. Do you believe it'd be risky to do this? Given the way our automated docs-generation processes work today, how plausible is it that we'd ship a docs update adding a link to a Pulumi archive or checksum that didn't exist on get.pulumi.com?


stack72 (approved):
No concerns here - as we don't push docs PRs until the system is fully published :)

cnunciato:
cc @zchase so you know this was done.

PR #7393, opened 2022-04-22
docs(CLI): improve the pulumi stack import docs

hcharley:
- added help section
- improved synopsis with links and clarity around deployment === state

### Proposed changes

Improvements to the stack import page: https://www.pulumi.com/docs/reference/cli/pulumi_stack_import/

n/a

### Related issues (optional)

n/a

### DIFF
--- a/content/docs/reference/cli/pulumi_stack_import.md
+++ b/content/docs/reference/cli/pulumi_stack_import.md
@@ -4,20 +4,65 @@ title: "pulumi stack import"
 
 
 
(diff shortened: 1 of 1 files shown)


susanev:
yes, @hcharley thank you so m uch for your contribution but we cannot accept it because these docs are generated automatically. if you wouldn't mind opening up in issue in https://github.com/pulumi/pulumi-hugo then we can figure out the best place to add in some hand-authored docs.

PR #7428, opened 2022-04-29
Switch back to GitHub runners

cnunciato:
Puts us back on a GitHub runner while we continue working on setting up self-hosted, and removes the explicit GOGC setting, as it's no longer needed (knocks wood).


cnunciato:
Hm -- I guess I have to manually update `github.com/pulumi/pulumi/pkg/v3`:

image

PR #7480, opened 2022-05-06
Upgrade the infra program

cnunciato:
After spending a bunch of time trying (and failing) to debug why the redirect that went in with #7476 isn't working, I spun up a dev stack, and saw the same odd behavior I'm seeing in prod: changes to the redirect Lambda appear to be applied, but they don't seem to be getting deployed, logging is inconsistent, and Pulumi updates that mutate the Lambda intermittently fail with an error having to do with an "update already in progress" error (for example, [here](https://github.com/pulumi/docs/runs/6314156176?check_suite_focus=true#step:9:1578)).

Then I noticed the versions of @pulumi/pulumi and @pulumi/aws were pretty old, so I updated them, and now everything works as expected -- changes to the Lambda are applied immediately, logging looks right, and no more deployment errors. 

So this change updates these and other packages to bring everything current (and fix the redirect rule).

### DIFF
--- a/infrastructure/package.json
+++ b/infrastructure/package.json
@@ -4,14 +4,14 @@
         "lint": "tslint --project tsconfig.json"
     },
     "devDependencies": {
-        "@types/aws-lambda": "^8.10.71",
-        "@types/node": "^12.7.2",
+        "@types/aws-lambda": "^8.10.95",
+        "@types/node": "^17.0.31",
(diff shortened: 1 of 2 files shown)


zchase (approved):
Nice!! I remember hitting this when prototyping some A/B stuff. Awesome you found a fix!!

PR #7541, opened 2022-05-19
Refactor the link-checking script

cnunciato:
Our link checker's been down [for a quite a while now](https://github.com/pulumi/docs/actions/workflows/check-links.yml), and debugging has been difficult, largely because of the way we're running it. (We basically point the checker at the home page and let it crawl the whole site, which can take hours, and generally either fails silently or hangs in inconsistent ways.) So this change switches things up a bit both to make debugging easier and to reduce the amount of work we're expecting to handle within a single Node.js process. Specifically:

* Moves away from pointing the checker at the home page and letting it crawl the whole site to fetching the sitemap at https://www.pulumi.com/sitemap.xml and iterating over each page in that list individually, excluding the pages we don't want to crawl (like the API docs pages, because they're so many). This lets us see clearly from the logs which pages have issues and where in the process bad links were discovered.

* Adds a way to tack additional pages onto the list of pages to be checked, to allow us to keep a closer watch on things like popular GitHub READMEs, our Community Slack page, and others.


susanev:
ohh, thx for taking a look at this!! ❤️

cnunciato:
> ohh, thx for taking a look at this

You bet. I've actually been working on this for a while without much success. Part of the challenge has been that the library we use for link checking is no longer actively maintained, and it doesn't seem to handle sites of our size all that well, either. This change is aimed more at making what we already have work more reliably, but I suspect we'll probably want to come back to it again later with a more robust long-term solution.

zchase (approved):
Nice!!! This is great.

Outside of the build error this looks good

PR #7560, opened 2022-05-23
add note about running make build first

susanev:

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ The `Makefile` exposes a number of helpers:
 * `make ensure` resolves and installs dependencies, including the Hugo module containing website layouts and content
 * `make lint` checks all Markdown files for correctness
 * `make build` generates the website and writes it to `./public`
-* `make serve-static` runs a local HTTP server that serves the contents of `./public`
+* `make serve-static` runs a local HTTP server that serves the contents of `./public` (must run `make build` first before this will work)
 * `make test` runs a script that checks for broken links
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,7 +39,7 @@ The `Makefile` exposes a number of helpers:
 * `make ensure` resolves and installs dependencies, including the Hugo module containing website layouts and content
 * `make lint` checks all Markdown files for correctness
 * `make build` generates the website and writes it to `./public`
-* `make serve-static` runs a local HTTP server that serves the contents of `./public`
+* `make serve-static` runs a local HTTP server that serves the contents of `./public` (must run `make build` first before this will work)
```
cnunciato, line 42:
```suggestion
* `make serve-static` runs a local HTTP server that serves the contents of `./public` (assumes you've run `make build` first)
```

cnunciato (approved):
Minor suggestion for language. Also, as an alternative (or addition), you could make a small tweak to the `serve-static` target to add a note for those who hit this without having read the README first: 

```
...
.PHONY: serve-static
serve-static:
	ls -al public/index.html && yarn run http-server public || echo "Missing './public/index.html'. Have you run 'make build'?"
...
```

viveklak:
Actually the specific issue I had was that my hugo was too old but make ensure didn't make that clear. This caused make build to fail - I did in fact run make build first. Should we add a warning in https://github.com/pulumi/docs/blob/master/scripts/ensure.sh#L11 to assert the hugo version? In general I rely heavily on make ensure to be authoritative on the dependencies required.

PR #7574, opened 2022-05-26
wip fix java search

dixler:

### DIFF
--- a/static-prebuilt/docs/reference/pkg/java/search.js
+++ b/static-prebuilt/docs/reference/pkg/java/search.js
@@ -317,7 +317,7 @@ $(function() {
 
                 // Nastiest (temporary) hack ever, but it works.
                 // https://github.com/pulumi/pulumi-java/issues/595
-                url = url.replace("undefined/", "");
+                //url = url.replace("undefined/", "");
 
                 if (top !== window) {
                     parent.classFrame.location = pathtoroot + url;


t0yv0 (changes requested):
We perhaps should not be editing a file copyrighted Oracle. This folder is autogenerated by the Javadoc target - and the fix will be wiped when we regenerate. Can we find and fix the root cause? Thanks.

dixler:
> We perhaps should not be editing a file copyrighted Oracle. This folder is autogenerated by the Javadoc target - and the fix will be wiped when we regenerate. Can we find and fix the root cause? Thanks.

https://stackoverflow.com/questions/52326318/maven-javadoc-search-redirects-to-undefined-url

> This bug should be fixed in JDK 12: https://bugs.openjdk.java.net/browse/JDK-8215291

Is bumping to JDK 12 fine?

t0yv0:
Interesting. It seems like changing default JDK is a big deal, more so than this error; we should at least retain some form of testing on 11. Can we scope the change to javadoc generation only?  Generate Javadoc on 12 but the rest as-is on 11?

t0yv0:
It's not possible to patch generated code by hand, as the patch will be removed. If you could include a script that auto-patches the output as part of code generation that could be a viable workaround for now.

dixler:
@t0yv0 I've updated the script to overwrite the search.js file

t0yv0:
Thank you! This makes it easy to keep updating the docs.

t0yv0 (approved):

PR #7616, opened 2022-06-03
ignore github 403s

susanev:
website notifications is full of these
can we ignore them for now?

to clarify, there is nothing wrong with the github links, but the link checker is reporting them as 403s in the website notifications channel because of how the link checker was written, 20 or something a day

### DIFF
--- a/scripts/link-checker/check-links.js
+++ b/scripts/link-checker/check-links.js
@@ -306,6 +306,9 @@ function excludeAcceptable(links) {
         // intelligently, but we can come back to that in a follow up.
         .filter(b => !(b.reason === "HTTP_429" && b.destination.match(/github.com|npmjs.com/)))
 
+        // Ignore GitHub 403s
+        .filter(b => !(b.reason === "HTTP_403" && b.destination.match(/github.com/)))
+
         // Ignore remote disconnects.
         .filter(b => b.reason !== "ERRNO_ECONNRESET")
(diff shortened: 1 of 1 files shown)


scripts/link-checker/check-links.js
```diff
@@ -306,6 +306,9 @@ function excludeAcceptable(links) {
         // intelligently, but we can come back to that in a follow up.
         .filter(b => !(b.reason === "HTTP_429" && b.destination.match(/github.com|npmjs.com/)))
 
+        // Ignore GitHub 403s
+        .filter(b => !(b.reason === "HTTP_403" && b.destination.match(/github.com/)))
```
susanev, line 310: maybe we should ignore all 403s not just github ones 🤷🏻

cnunciato: Before we just go ignoring, do we have some idea why these are returning 403 to begin with?

susanev: i thought you wrote why in slack...

susanev: https://pulumi.slack.com/archives/C03G209S6H4/p1654010275582649

susanev: if you have another idea pls lets do that, the noise of the 403s is preventing us from seeing and fixing the 404s.

PR #7704, opened 2022-06-24
Add region to redirects script

cnunciato:
Overlooked this script in #7703.


t0yv0:
@susanev @cnunciato could we merge please? It seems to fix CI. I have a feeling this may unblock master CI that will further unblock a release-related PR that will make one of pulumi/pulumi checks.

cnunciato:
@t0yv0 Just so you're aware, this particular fix wasn't blocking website deployments, just a bit of work that we do after deployments to fix up the way we handle certain redirects.

t0yv0:
Ah, that's a pity. I'm chasing down https://github.com/pulumi/docs/pull/7700 that is failing and I misdiagnozed it to the problem here. Let me try to rebase that on master and try again, and see if there's a remaning problem.

PR #7936, opened 2022-08-25
Drop BuildJet down to 4 vCPU

cnunciato:
Experimenting with different settings to try to reduce our costs a bit.


stack72 (approved):

cnunciato:
Ok @stack72 @lukehoban, the build passed and the job and build times look about the same (~27 minutes for PR builds), so I'll merge this and keep an eye on it. Thanks for the nudge. 🍻

Issue #8043, opened 2022-09-20
Checksums missing for some versions


justinvp:
On https://www.pulumi.com/docs/get-started/install/versions/

Clicking “Checksums” for `3.40.0` and `3.40.0-alpha.1662150830` leads to AccessDenied.

https://get.pulumi.com/releases/sdk/pulumi-3.40.0-checksums.txt

<img width="752" alt="Screen Shot 2022-09-20 at 9 39 51 AM" src="https://user-images.githubusercontent.com/710598/191315619-c01e5d17-828a-41c2-a6af-3048ae2470d7.png">

justinvp:
#8044 temporarily removed the checksum links. We'll add the links back once the checksum files have been published.

PR #8044, opened 2022-09-20
Update latest-version for the 3.40.1 release

justinvp:
There's an [issue](https://github.com/pulumi/pulumi/issues/10787) with typedoc generation that prevented these from being updated automatically, so updating manually.


content/docs/get-started/install/versions.md
```diff
@@ -18,6 +18,7 @@ The current stable version of Pulumi is **{{< latest-version >}}**.
         </tr>
     </thead>
     <tbody>
+        {{< changelog-table-row version="3.40.1" date="2022-09-19" showChecksum="true" >}}
```
justinvp, line 21: Note that the checksums for `3.40.0` (and `3.40.0-alpha.1662150830`) and now `3.40.1` are currently missing:

- https://get.pulumi.com/releases/sdk/pulumi-3.40.0-checksums.txt
- https://get.pulumi.com/releases/sdk/pulumi-3.40.1-checksums.txt

Tracked in https://github.com/pulumi/docs/issues/8043

If we wanted, we could temporarily set `showChecksum="false"` to hide those until we've addressed https://github.com/pulumi/docs/issues/8043

justinvp: Went ahead and added a commit that hides the checksum links for these versions temporarily. We can add the links back after we publish the checksum files.

AaronFriel (approved):
LGTM

PR #8053, opened 2022-09-21
Fix Node.js SDK generation

justinvp:
The Pulumi Node.js SDK bumped the TypeScript version from `3.7.3` to `3.8.3` to be able to use OpenTelemetry. Unfortunately, this has broken the SDK doc generation.

This change addresses the problem by upgrading TypeDoc to a version that supports TypeScript `3.8.3` and fixing our `tscdocgen` tool to handle the new output from the upgraded version of TypeDoc.

Fixes https://github.com/pulumi/pulumi/issues/10787


content/docs/reference/pkg/nodejs/pulumi/pulumi/_index.md
```diff
@@ -2711,44 +2725,44 @@ Inserts new elements at the start of an array.
...
 <h3 class="pdoc-module-header" id="UnwrapSimple" data-link-title="UnwrapSimple">
-    <a href="https://github.com/pulumi/pulumi/blob/e78abcb5b5ce42737735e571d6b82c2dded72b2f/sdk/nodejs/output.ts#L772">
+    <a href="https://github.com/pulumi/pulumi/blob/4cb43b5a1df4de0cd24bf29f622c336df05cdce6/sdk/nodejs/output.ts#L772">
         type <strong>UnwrapSimple</strong>
     </a>
 </h3>
 
-<pre class="highlight"><code><span class='kd'>type</span> UnwrapSimple = UnwrapSimple<T>;</code></pre>
+<pre class="highlight"><code><span class='kd'>type</span> UnwrapSimple = T extends primitive ? T : T extends <a href='/docs/reference/pkg/nodejs/pulumi/pulumi/#Resource'>Resource</a> ? T : T extends <a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array'>Array</a>&lt;infer U&gt; ? UnwrappedArray&lt;U&gt; : T extends <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object'>object</a></span> ? UnwrappedObject&lt;T&gt; : <span class='kd'><a href='https://www.typescriptlang.org/docs/handbook/basic-types.html#never'>never</a></span>;</code></pre>
```
AaronFriel, line 2753: I wonder if we should omit this type? This is the sort of type signature that can give people people unfamiliar with TypeScript pause (or cause for concern).

content/docs/reference/pkg/nodejs/pulumi/pulumi/_index.md
```diff
@@ -1580,12 +1594,12 @@ invoke.
...
 
 <h3 class="pdoc-module-header" id="Lifted" data-link-title="Lifted">
-    <a href="https://github.com/pulumi/pulumi/blob/e78abcb5b5ce42737735e571d6b82c2dded72b2f/sdk/nodejs/output.ts#L943">
+    <a href="https://github.com/pulumi/pulumi/blob/4cb43b5a1df4de0cd24bf29f622c336df05cdce6/sdk/nodejs/output.ts#L943">
         type <strong>Lifted</strong>
     </a>
 </h3>
 
-<pre class="highlight"><code><span class='kd'>type</span> Lifted = Lifted<T>;</code></pre>
+<pre class="highlight"><code><span class='kd'>type</span> Lifted = T extends <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String'>string</a></span> ? LiftedObject&lt;String, NonFunctionPropertyNames&lt;String&gt;&gt; : T extends <a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array'>Array</a>&lt;infer U&gt; ? LiftedArray&lt;U&gt; : T extends <span class='kd'><a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object'>object</a></span> ? LiftedObject&lt;T, NonFunctionPropertyNames&lt;T&gt;&gt; : { ... };</code></pre>
```
AaronFriel, line 1602: Likewise here

content/docs/reference/pkg/nodejs/pulumi/pulumi/_index.md
```diff
@@ -2312,12 +2326,12 @@ A regular expression the string must match. If it does not match, a ConfigPatter
...
 
 <h3 class="pdoc-module-header" id="Unwrap" data-link-title="Unwrap">
-    <a href="https://github.com/pulumi/pulumi/blob/e78abcb5b5ce42737735e571d6b82c2dded72b2f/sdk/nodejs/output.ts#L759">
+    <a href="https://github.com/pulumi/pulumi/blob/4cb43b5a1df4de0cd24bf29f622c336df05cdce6/sdk/nodejs/output.ts#L759">
         type <strong>Unwrap</strong>
     </a>
 </h3>
 
-<pre class="highlight"><code><span class='kd'>type</span> Unwrap = Unwrap<T>;</code></pre>
+<pre class="highlight"><code><span class='kd'>type</span> Unwrap = T extends <a href='https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise'>Promise</a>&lt;infer U1&gt; ? UnwrapSimple&lt;U1&gt; : T extends <a href='/docs/reference/pkg/nodejs/pulumi/pulumi/#OutputInstance'>OutputInstance</a>&lt;infer U2&gt; ? UnwrapSimple&lt;U2&gt; : UnwrapSimple&lt;T&gt;;</code></pre>
```
AaronFriel, line 2334: & here

AaronFriel (approved):
LGTM - looks like we've expanded our support for new AST node types in the type doc and it caused us to generate more precise documentation.

Mildly concerned about types like Lifted, Unwrap, etc., leaking into our docs. Really not anything end users need to be aware of, so if we can't _hide_ them, I wonder if we could note that they're types used internally to provide a smooth experience using Pulumi TypeScript?

justinvp:
> Mildly concerned about types like Lifted, Unwrap, etc., leaking into our docs. Really not anything end users need to be aware of, so if we can't _hide_ them, I wonder if we could note that they're types used internally to provide a smooth experience using Pulumi TypeScript?

I agree that these would ideally be hidden, but prefer to consider that in a follow-up since we were already including them in the docs currently. The handling of the new types (conditional and infer) from the new version of TypeDoc was done to prevent failing on unrecognized types. I guess instead of actually providing an implementation for these, I could have tried to make it so that they rendered as they did previously (perhaps by just returning an empty string).

In any case, we should be cautious about investing more in improving our `tscdocgen` tool. We originally used this to generate Node.js docs for all our providers, and the output is very oriented towards that (i.e. there's a `Resources` header, that's meant to show all the resources in a given module; which doesn't really fit with the docs for our core SDK, which doesn't have resources per se other than the base `Resource`/`CustomResource`/`ComponentResource` classes). We've long since moved away from using this tool for our provider docs (we have another tool that generates these for the registry, and emits declarations for all languages).

We have a similar problem with Python SDK docs, which are awful.

Rather investing in our bespoke tools for Node and Python, we should consider moving to standard doc generation approach like we do for [.NET](https://www.pulumi.com/docs/reference/pkg/dotnet/Pulumi/Pulumi.html), [Go](https://pkg.go.dev/github.com/pulumi/pulumi/sdk/v3/go/pulumi), and [Java](https://www.pulumi.com/docs/reference/pkg/java/).

Issue #6426, opened 2021-09-08
Create standard python documentation site for Pulumi SDK


komalali:
Related to: https://github.com/pulumi/docs/issues/4615

Our current [python SDK](https://www.pulumi.com/docs/reference/pkg/python/pulumi/) documentation page is difficult to navigate and has a number of formatting errors (as described in #4615). This is a proposal to lift and shift our python documentation to us the standard readthedocs theme, implement navigation, improve formatting and move to a structure that python users are familiar with.

An outline of the steps required to enable this:

- [ ] Build a standard doc site with Sphinx: https://docs.readthedocs.io/en/stable/intro/getting-started-with-sphinx.html
- [ ] Modify the generated sphinx files (use the existing [pulumi.rst](https://github.com/pulumi/docs/blob/master/tools/pydocgen/pydocgen/templates/providers/pulumi.rst) file and [typescript docs](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/) as templates for what information to include)
- [ ] Deploy the documentation as part of pulumi.com.
- [ ] Update any links that might require updating

Notes:
When picking up this work, sync with @susanev on bringing in the practitioner nav

komalali:
Following up to add some more color/clarity because I know there are some concerns here.

Concerns I have heard:
1. Losing design/style consistency with other language SDKs
2. We need to host this as part of pulumi.com

### Current state

Our [Go SDK docs](https://pkg.go.dev/github.com/pulumi/pulumi/sdk/v3/go/pulumi) are on pkg.go.dev so they are neither in pulumi style or hosted on pulumi.com. Our [Dotnet SDK docs](https://www.pulumi.com/docs/reference/pkg/dotnet/Pulumi/Pulumi.html) also follow dotnet-specific styling, so they are not in pulumi style (but they are hosted on pulumi.com). Realistically, there is no shared style between the 4 SDKs as even [Nodejs](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/) and [Python](https://www.pulumi.com/docs/reference/pkg/python/pulumi/) are different from each other. 

Below is a screenshot of the current layout. It is not easy to discern that `apply` is a method on `Output`. Things like `__getitem__` should not be included in this page. I could pinpoint every single issue in this image but I think it is clear that it is not "easy" to understand.

[...]

PR #8110, opened 2022-10-14
feat: statically generate python docs using sphinx + rtd

kpitzen:
This PR is intended to be the first three tasks for this issue:
https://github.com/pulumi/docs/issues/6426
The plan is to only replace the existing python doc links once we've verified this new set of documentation accomplishes what we're setting out to do.  Once that's done, we can remove the current-state documentation as well.


kpitzen:
The new version of the docs is available here: http://pulumi-docs-origin-pr-8110-486f143b.s3-website.us-west-2.amazonaws.com/docs/reference/pkg/python/providers/pulumi/

susanev:
@kpitzen can you do a search for links to the sdk docs, and update the links to the new version?
for example this page has a link to them `docs/intro/languages/python/` that's not going to the new ones
and the link under reference > pulumi sdk in the left nav is also not going to them; nor is the link on the pulumi sdk landing page

or were you doing a soft launch of these? i think its prolly gonna be confusing to have both around.. am i missing something? maybe adding a description to the pr may help.

kpitzen:
Morning @susanev !  After chatting with @cnunciato about this, the plan was to do this in stages - the first being to simply add the new documentation alongside the old (without updating links, etc) since those links exist both here and in `pulumi-hugo`.  The followup steps would be replacing existing links with the new locations, and finally removing the old docs from this repo.  I wasn't familiar with the description template on this repo, but I'm happy to fill that in as well!

susanev:
@kpitzen cool, sounds good! would suggest adding some deets to the description so reviewers have access to this information. or link to an issue that includes them.

nimbinatus (approved):
Love it! LGTM from a quick rundown as it looks the same as the existing content, just in sphinx/RST

PR #8144, opened 2022-10-25
Update the link checker to include internal URLs

cnunciato:
This change updates the link checker to check "internal" links (i.e., those that point from one page on our site to another) as well. (Up to now, we've been relying on `relref` to validate internal links for us, but once https://github.com/pulumi/pulumi-hugo/pull/2071 merges, we'll need to start checking them ourselves.) In addition to specifying `excludeInternalLinks: false`, this change also removes a bunch of dead code; we used to run the link checker in various ways (as a spider, on a page by page basis, etc.), but as of now, we only need to point it at a URL with a sitemap and say "go".

Changes assume these PRs have been merged:
* https://github.com/pulumi/pulumi-hugo/pull/2071 
* https://github.com/pulumi/docs/pull/8141 (which actually fixes a few 404s identified by the changes in this PR)


sean1588 (approved):
:shipit:

Issue #8149, opened 2022-10-26
`build-pulumi-cli-docs` GHA job currently failing due to pipenv lockfile


Frassle:
See https://github.com/pulumi/docs/actions/runs/3329923605/jobs/5509760775

susanev:
@lukehoban who needs to triage this issue?

PR #8153, opened 2022-10-26
Update latest-version to v3.44.2

dixler:


cnunciato:
This isn't usually done manually. Is this a workaround of some sort?

cnunciato (approved):
Chatted with @dixler out of band. This should get the version updated and onto the site while folks work on a fix for https://github.com/pulumi/docs/issues/8149.

PR #8223, opened 2022-11-09
fix: uses dir() to resolve sub-modules of pulumi plugins for python docs

kpitzen:
`__all__` was recently changed for pulumi_policy, which exposed an issue for submodules of plugins during docs generation.  This resolves that issue by using `dir()` instead.


content/docs/reference/pkg/python/pulumi_policy/_index.md
```diff
@@ -10,5 +10,11 @@ notitle: true
 <section id="pulumi-policy">
 <h1>Pulumi Policy<a class="headerlink" href="#pulumi-policy" title="Permalink to this headline"></a></h1>
 <div class="toctree-wrapper compound">
+<ul>
+<li class="toctree-l1"><a class="reference internal" href="deserialize/">deserialize</a></li>
+<li class="toctree-l1"><a class="reference internal" href="policy/">policy</a></li>
+<li class="toctree-l1"><a class="reference internal" href="proxy/">proxy</a></li>
+<li class="toctree-l1"><a class="reference internal" href="version/">version</a></li>
```
justinvp, line 17: These modules are supposed to be implementation details. Here's an alternative PR that addresses the problem without exposing the modules: https://github.com/pulumi/docs/pull/8224

PR #8269, opened 2022-11-23
Regenerate the classic Node.js SDK docs for AWSx 1.0 and EKS 1.0

justinvp:
Unfortunately, I had to do some [manual tweaks](https://github.com/pulumi/pulumi-awsx/compare/master...justin/typescript-3.8.3) to AWSx to get it to generate -- hopefully we won't need to continue regenerating those very often. It's a long story, but we're currently stuck on TypeScript 3.8.3 for generating the docs (matches the version of TypeScript currently used for the core `pulumi/pulumi` Node.js SDK), which works with the version of TypeDoc we're using. Upgrading these would be involved, because our tsdocgen tool looks for very specific nodes from the compiler. In any case, the easiest way to get this working was to rebuild the AWSx Node using TypeScript 3.8.3. I just had to comment out some aspects of the implementation that was using features not supported by 3.8.3. After doing that, I was able to get the docs to generate.


.gitignore
```diff
@@ -44,16 +44,6 @@ cypress/screenshots
...
-# Ignore language-specific SDK docs for providers.
-content/docs/reference/pkg/nodejs
-!content/docs/reference/pkg/nodejs/pulumi/awsx
-!content/docs/reference/pkg/nodejs/pulumi/cloud
-!content/docs/reference/pkg/nodejs/pulumi/eks
-!content/docs/reference/pkg/nodejs/pulumi/kunernetesx
-!content/docs/reference/pkg/nodejs/pulumi/policy
-!content/docs/reference/pkg/nodejs/pulumi/pulumi
-!content/docs/reference/pkg/nodejs/pulumi/terraform
-
```
justinvp, line 56: This was added in #5133, but I found that git was ignoring certain new directories within `content/docs/reference/pkg/nodejs/pulumi/awsx` (i.e. the new `classic` subdir), despite the `!` to negate the pattern.

justinvp:
Since this change is largely just re-running the tool to regenerate docs (manual tweaking aside), I'm going to go ahead and merge it. But I'm happy to address any review feedback subsequently.

PR #8336, opened 2022-12-13
Include strict-transport-policy response headers

blampe:
### Proposed changes

Refs https://github.com/pulumi/get.pulumi.com/pull/140 https://github.com/pulumi/home/issues/2490

We've seen an issue where users running old OSes or browsers will incorrectly flag our newly-rotated production certs as invalid, and we suspect adding security headers will address the issue.


AaronFriel (approved):
LGTM.

PR #8329, opened 2022-12-10
Update p/p dependency in resourcedocsgen

cnunciato:
This is the result of running:

```
go get -u github.com/pulumi/pulumi/pkg/v3
go get -u github.com/pulumi/pulumi/sdk/v3
go mod tidy
```

PR #8431, opened 2023-01-12
update resource docs

susanev:


susanev:
@cnunciato @sean1588  im running the commands here https://github.com/pulumi/docs/pull/8329#issue-1487972227
but that diff looks very wrong, thoughts?

PR #8476, opened 2023-01-25
add GH token env var

sean1588:
fixes: https://github.com/pulumi/docs/issues/8477

Adding the GITHUB_TOKEN env variable to workflow step. We saw another rate limitting error [here](https://github.com/pulumi/docs/actions/runs/4000780762/jobs/6866326356). It looks like the [script ](https://github.com/pulumi/docs/blob/master/scripts/get-module-diff-details.js#L8-L12)is set up to make the request with the GH token, but i'm guessing it wasnt set in the environment. Maybe it is already set by default though, so let me know if this makes sense. However, I do see us explicitly setting it like this in other workflows, so I am assuming this is the issue.

### DIFF
--- a/.github/workflows/update-theme.yml
+++ b/.github/workflows/update-theme.yml
@@ -45,6 +45,8 @@ jobs:
             echo 'NEW_COMMITS<<EOF' >> $GITHUB_ENV
             node scripts/get-module-diff-details.js >> $GITHUB_ENV
             echo 'EOF' >> $GITHUB_ENV
+        env:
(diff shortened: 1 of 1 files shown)


susanev:
@sean1588 do you mind getting an issue opened for tracking, sry haven't gotten to it yet

susanev (approved):

sean1588:
sure, here is the issue: https://github.com/pulumi/docs/issues/8477

zchase (approved):
LGTM --- Thanks for picking this up! Hopefully there are not too many of these left :)

PR #8529, opened 2023-02-02
add slack alerts

sean1588:
part of issue: https://github.com/pulumi/home/issues/2566

This PR adds slack notifications on failures for workflows in this repo.  These notifications will get posted to the docs-ops channel in slack, which contains a link to the failed job and the name of the workflow that failed (listed as the actions url in the image below). 

image


sean1588:
@zchase, totally not related to this PR, but do we still need this line now that theme is no longer in use?
https://github.com/pulumi/docs/pull/8529/files#diff-2a995acf6cbf1e1bd347d552e2d13117b5bf8b2ee1b8bb28ac7aa251567be2dcR39

kimberleyamackenzie (approved):

PR #8667, opened 2023-03-07
pydocgen: Don't generate HTML permalinks

abhinav:
sphinx_rtd_theme injects HTML permalinks in all pages by default,
using a FontAwesome based link icon.

This is not necessary on our website because we use Javascript
to inject header links client-side,
and instead leads to double-anchors on the website.


abhinav:
This doesn't fix it. @dixler is exploring a different short term fix.

PR #8668, opened 2023-03-07
pydocgen: Set html_permalinks to false

abhinav:
sphinx_rtd_theme injects HTML permalinks in all pages by default,
using a FontAwesome based link icon.

This is not necessary on our website because we use Javascript
to inject header links client-side,
and instead leads to double-anchors on the website.

This PR disables the setting and regenerates Python docs.


content/docs/reference/pkg/python/pulumi/_index.md
```diff
@@ -54,14 +54,14 @@ resource.</p></li>
 </dl>
 <dl class="py property">
 <dt class="sig sig-object py" id="pulumi.Resource.urn">
-<em class="property"><span class="pre">property</span><span class="w"> </span></em><span class="sig-name descname"><span class="pre">urn</span></span><em class="property"><span class="p"><span class="pre">:</span></span><span class="w"> </span><a class="reference internal" href="#pulumi.Output" title="pulumi.output.Output"><span class="pre">Output</span></a><span class="p"><span class="pre">[</span></span><span class="pre">str</span><span class="p"><span class="pre">]</span></span></em><a class="headerlink" href="#pulumi.Resource.urn" title="Permalink to this definition"></a></dt>
```
justinvp, line 57: The downside is that it appears to remove links like these, which I think we'd want to keep so that you can link to individual members.

image

We really just want to remove sphinx's icon or the permalinks it adds to headers. I'm not sure if there are config settings that do that. We may have to go back to using CSS to hide the icon.

abhinav: Yeah, this is the only dial. If we want to keep the non header links, we'll have to go back to the CSS option.

justinvp: Let's go with the CSS fix, then. https://github.com/pulumi/pulumi-hugo/pull/2539

abhinav:
closing in favor of https://github.com/pulumi/pulumi-hugo/pull/2539

Issue #8627, opened 2023-02-27
investigate transient test failures in docs build


sean1588:
We are continuing to get transient test failures when the cypress tests run as part of the docs build. We attempted to fix this by adding some retry logic to the tests, but it doesn't seem to be helping.  We should do a bit more investigation to see what is going on.

failed job:
https://github.com/pulumi/docs/actions/runs/4286641689/jobs/7466477522#step:9:1816

cnunciato:
Actually, I'll close this as a wontfix for now, since we don't seem to have seen it for a while, and we now have automation in place to file new P1 on build failures.

PR #8678, opened 2023-03-08
retry cypress tests on failure

sean1588:
fixes: #8627 

The cypress retry logic does not seem to actually be working for some reason and I cannot seem to get it to work. I'm not sure why exactly. I have reverted this to doing the retry purely in bash by trapping the exit code from the failed tests and triggering the retry.

### DIFF
--- a/cypress/integration/site_spec.js
+++ b/cypress/integration/site_spec.js
@@ -6,7 +6,7 @@ describe("www.pulumi.com", () => {
             cy.visit("/");
         });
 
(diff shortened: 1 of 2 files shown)


susanev:
should we close out https://github.com/pulumi/docs/pull/8247?

sean1588:
> should we close out #8247?

not sure. @zchase?

susanev (approved):

PR #8737, opened 2023-03-17
Add parallel workflow to deploy s3 bucket to new account

sean1588:
Part of: https://github.com/pulumi/home/issues/2564

This PR adds a separate workflow to enable us to start deploying to the new account. This workflow will run parallel to the current production workflow, which will enable us to continuously deploy to both accounts simultaneously and validate the deployment pipeline before we hook up cloudfront and/or make any dns changes. Anything deploying to the new account should be non-consequential with regards to the current deployment workflow since it is completely separeate and will just run parallel to what we have in place now.

This will only make the s3 bucket and sync the contents. The pulumi program assumes that the bucket already exists, so `pulumi up` will not be run yet. We will do this once we verify the buckets are getting created and site is functioning successfully. We will also need to provision a new pulumi stack for the new environment during the migration process.

doc: https://docs.google.com/document/d/1ADXcC8g2TQ75okm2URZrnSuDb-bN-ilRpVRJN5C-dZo/edit

### DIFF
--- a/.github/workflows/build-and-deploy-new-acct.yml
+++ b/.github/workflows/build-and-deploy-new-acct.yml
@@ -66,9 +66,9 @@ jobs:
       - name: Slack Notification
         uses: docker://sholung/action-slack-notify:v2.3.0
         env:
-          SLACK_CHANNEL: sean-ops-test
+          SLACK_CHANNEL: docs-ops-test
           SLACK_COLOR: "#F54242"
-          SLACK_MESSAGE: "[new account] build and deploy failure in pulumi/docs repo :meow_sad:"
+          SLACK_MESSAGE: "[NEW ACCOUNT] build and deploy failure in pulumi/docs repo :meow_sad:"
           SLACK_USERNAME: docsbot
           SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_URL }}
           SLACK_ICON: https://www.pulumi.com/logos/brand/avatar-on-white.png
--- a/.github/workflows/build-and-deploy-new-acct.yml
+++ b/.github/workflows/build-and-deploy-new-acct.yml
@@ -1,0 +1,74 @@
+# temporary workflow to start deploying to the new prod account.
+name: Build and deploy new account
+on:
+  push:
+    branches:
+      - master
+jobs:
+  buildSite:
+    env:
+      GOPATH: ${{ github.workspace }}/go
+    name: Install deps and build site
+    runs-on: buildjet-4vcpu-ubuntu-2004
+    steps:
+
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '16.x'
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.19.x
+
+      - uses: peaceiris/actions-hugo@v2
+        with:
+          hugo-version: '0.96.0'
+          extended: true
+
+      - name: Configure Git
+        run: |
+          git config --global url."https://${{ secrets.PULUMI_BOT_TOKEN }}:x-oauth-basic@github.com/pulumi/pulumi-hugo-internal".insteadOf "https://github.com/pulumi/pulumi-hugo-internal"
+      - name: Install assume-role
+        run: go install github.com/remind101/assume-role@latest
+
+      - name: Install Pulumi CLI
+        uses: pulumi/setup-pulumi@v2
+
+      - name: Build and deploy
+        run: make ci_${GITHUB_EVENT_NAME}
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.CI_AWS_ACCESS_KEY_ID_NEW }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.CI_AWS_SECRET_ACCESS_KEY_NEW }}
+          PULUMI_ACCESS_TOKEN: ${{ secrets.PULUMI_ACCESS_TOKEN }}
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NODE_OPTIONS: "--max_old_space_size=4096"
+
+      - name: Archive test results
+        uses: actions/upload-artifact@v2
+        with:
+          name: browser-test-results
+          path: cypress/videos
+
+      - name: Archive bucket metadata
+        uses: actions/upload-artifact@v2
+        with:
+          name: origin-bucket-metadata
+          path: origin-bucket-metadata.json
+  notify:
+    if: (github.ref == 'refs/heads/master' || (github.event_name == 'pull_request' && github.event.pull_request.user.login == 'pulumi-bot')) && failure()
+    name: Send slack notification
+    runs-on: ubuntu-latest
+    needs: [buildSite]
+    steps:
+      - name: Slack Notification
+        uses: docker://sholung/action-slack-notify:v2.3.0
+        env:
+          SLACK_CHANNEL: sean-ops-test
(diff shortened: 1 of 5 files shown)


scripts/ci-login.sh
```diff
@@ -2,9 +2,15 @@
 
 set -o errexit -o pipefail
 
-echo "Assuming the production CI role..."
-eval $(assume-role -duration "2h0m0s" "arn:aws:iam::058607598222:role/ContinuousIntegrationRole")
+if [[ "$GITHUB_WORKFLOW" == "Build and deploy new account" ]]; then
+    echo "Assuming the production CI role in new AWS account..."
+    # change this ARN to reference CI Role in new account.
+    # eval $(assume-role -duration "2h0m0s" "arn:aws:iam::058607598222:role/ContinuousIntegrationRole")
```
sean1588, line 8: I need to create this role in the new AWS account

.github/workflows/build-and-deploy-new-acct.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Install assume-role
+        run: go install github.com/remind101/assume-role@latest
+
+      - name: Install Pulumi CLI
+        uses: pulumi/setup-pulumi@v2
+
+      - name: Build and deploy
+        run: make ci_${GITHUB_EVENT_NAME}
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.CI_AWS_ACCESS_KEY_ID_NEW }}
```
sean1588, line 45: I need to get creds made and create these secrets for the new account in this github repo. I'll need admin access for that.

.github/workflows/build-and-deploy-new-acct.yml
```diff
@@ -0,0 +1,74 @@
...
+on:
+  push:
+    branches:
+      - master
+jobs:
+  buildSite:
+    env:
+      GOPATH: ${{ github.workspace }}/go
+    name: Install deps and build site
+    runs-on: buildjet-4vcpu-ubuntu-2004
```
kmosher, line 12: I didn't realize we were using buildjet for this.

We do have native 4core github runners now, which IIRC are a tad more expensive, but mean we don't need to spread our keys onto another service (I was in the middle of trying to get buildjet's SOC2 when the github runners became available)

scripts/ci-push.sh
```diff
@@ -4,11 +4,17 @@ set -o errexit -o pipefail
...
+    ./scripts/build-site.sh
+    ./scripts/sync-and-test-bucket.sh
+    node ./scripts/await-in-progress.js
+    ./scripts/run-pulumi.sh update
+elif [[ "$GITHUB_WORKFLOW" == "Build and deploy new account" ]]; then
+    # temporarily setting site to build in preview mode, to prevent search bots from indexing
+    # until we cut over DNS to start routing to the new account.
+    ./scripts/build-site.sh preview
+    ./scripts/sync-and-test-bucket.sh
+fi
```
kmosher, line 19: same `else` to bail if it doesn't understand the workflow.

scripts/sync-and-test-bucket.sh
```diff
@@ -32,6 +32,10 @@ fi
 # For previews, name the destination bucket with the PR number, to reduce the number of
 # buckets we create and to facilitate shorter sync times.
 destination_bucket="$(origin_bucket_prefix)-$(build_identifier)"
+if [[ "$GITHUB_WORKFLOW" == "Build and deploy new account" ]]; then
```
kmosher, line 35: Same `if, else-if, else` pattern. Or make `destination bucket` something the CI builds and passes in directly.

kmosher (approved):
If this were a more long-term thing, I'd probably want to clean this up to use [environments](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) and share more of the workflow code. But for a cut-over like this, it's fine, subject to the extra guarding mentioned.

scripts/ci-push.sh
sean1588: thanks @kmosher, good call on adding an else to bail out.

.github/workflows/build-and-deploy-new-acct.yml
sean1588: i made the keys and added them

scripts/ci-login.sh
sean1588: role is now configured

.github/workflows/build-and-deploy-new-acct.yml
```diff
@@ -0,0 +1,74 @@
...
+          extended: true
+
+      - name: Configure Git
+        run: |
+          git config --global url."https://${{ secrets.PULUMI_BOT_TOKEN }}:x-oauth-basic@github.com/pulumi/pulumi-hugo-internal".insteadOf "https://github.com/pulumi/pulumi-hugo-internal"
+      - name: Install assume-role
+        run: go install github.com/remind101/assume-role@latest
+
+      - name: Install Pulumi CLI
+        uses: pulumi/setup-pulumi@v2
```
cnunciato, line 37: Doesn't really impact the content of this change necessarily, but worth noting that this is now `pulumi/actions@v4` in the workflow on `master`, so we'll want to update this here as well.

scripts/ci-push.sh
```diff
@@ -4,11 +4,20 @@ set -o errexit -o pipefail
...
+    # Wait for in-progress jobs to complete before proceeding.
+    ./scripts/build-site.sh
+    ./scripts/sync-and-test-bucket.sh
+    node ./scripts/await-in-progress.js
+    ./scripts/run-pulumi.sh update
+elif [[ "$GITHUB_WORKFLOW" == "Build and deploy new account" ]]; then
+    # temporarily setting site to build in preview mode, to prevent search bots from indexing
+    # until we cut over DNS to start routing to the new account.
+    ./scripts/build-site.sh preview
+    ./scripts/sync-and-test-bucket.sh
```
cnunciato, line 18: One thing to keep in mind here is that this new account probably has the default bucket quota, which I believe is 100 buckets. The current production account allows up to 500, I think. We clean these up nightly in another workflow that won't be run here (yet), so we'll probably hit that limit fairly pretty quickly in the new account. Manual cleanup in the console is probably fine (and safer) for now, so I mainly just wanted to call that out as something to be aware of.

scripts/sync-and-test-bucket.sh
```diff
@@ -31,7 +31,17 @@ fi
...
 # For previews, name the destination bucket with the PR number, to reduce the number of
 # buckets we create and to facilitate shorter sync times.
-destination_bucket="$(origin_bucket_prefix)-$(build_identifier)"
+destination_bucket=""
+if [[ "$GITHUB_WORKFLOW" == "Build and deploy" ]]; then
+    echo "configuring bucket in production account"
+    destination_bucket="$(origin_bucket_prefix)-$(build_identifier)"
+elif [[ "$GITHUB_WORKFLOW" == "Build and deploy new account" ]]; then
+    echo "configuring bucket in new production account"
+    destination_bucket="www-$(origin_bucket_prefix)-$(build_identifier)"
```
cnunciato, line 40: I'm pretty sure you're going to have to change this value to prevent bucket-naming collisions from happening as these workflows run in parallel (since they'll both be based on the same Git SHA): https://github.com/pulumi/docs/blob/a75eb1f1c6b87b0a36af90fb9bf7db622e024ab8/scripts/common.sh#L59

cnunciato (changes requested):
I'd like to look at the sync script a little closer as well, but I believe we'll want to change the string that generates the bucket name before we merge this. The SSM parameter-bucket relationship will be an issue as well.

scripts/ci-push.sh
sean1588: good call out

cnunciato (approved):

PR #8798, opened 2023-03-30
ignore more error types

susanev:
we will add these back after we have a handle on these errors


sean1588 (approved):
thanks so much!!! :rocket:

PR #8801, opened 2023-03-30
Strip broken icon in python docs

justinvp:
Sphinx is now emitting a private unicode character ("\F0C1") corresponding to a FontAwesome chain icon next to headers, which renders as an empty box on our site.

image (1)

Strip this character out of the generated docs in the same way we stripped the "¶" character, which sphinx previously emitted.

1. First commit updates the generator.
2. Second commit has the regenerated files.

Fixes https://github.com/pulumi/pulumi-hugo/issues/2625


justinvp:
Direct link to preview: http://pulumi-docs-origin-pr-8801-4e517ece.s3-website.us-west-2.amazonaws.com/docs/reference/pkg/python/pulumi/

abhinav (approved):

static-prebuilt/docs/reference/pkg/python/.doctrees/environment.pickle
RobbieMcKinstry, line 1: It looks like you've committed several binaries. Is that intended? These look like new files, so it looks possible they were committed on accident.

RobbieMcKinstry (approved):

static-prebuilt/docs/reference/pkg/python/.doctrees/environment.pickle
justinvp: I included everything Sphinx generated as we have previously: https://github.com/pulumi/docs/commits/master/static-prebuilt/docs/reference/pkg/python/.doctrees

But I don't think it's necessary to keep the files in `.doctrees` as it looks like they're just a cache. So we could probably remove them and add to `.gitignore`.

PR #8822, opened 2023-04-03
enable cors to access rss feeds corss origin

sean1588:
This PR enables us to access the rss feeds cross origin, since I need to access them from the console UI. I tried this out in a dev environment by standing up cloudfront and the bucket and confirming I can fetch from the browser console.  This will allow origins from any subdomain of pulumi.com or any subdomain of pulumi-dev.io

### DIFF
--- a/scripts/cors/cors.json
+++ b/scripts/cors/cors.json
@@ -3,7 +3,9 @@
     {
       "AllowedOrigins": [
         "https://*.pulumi.com",
-        "https://*.pulumi-dev.io"
+        "https://*.pulumi-dev.io",
+        "http://localhost:3000",
+        "http://localhost:4200"
       ],
       "AllowedMethods": ["GET"],
       "MaxAgeSeconds": 300
--- a/scripts/cors/cors.json
+++ b/scripts/cors/cors.json
@@ -5,7 +5,6 @@
         "https://*.pulumi.com",
         "https://*.pulumi-dev.io"
       ],
-      "AllowedHeaders": ["Authorization"],
       "AllowedMethods": ["GET"],
       "MaxAgeSeconds": 300
     }
--- a/scripts/cors/cors.json
+++ b/scripts/cors/cors.json
@@ -1,7 +1,10 @@
 {
   "CORSRules": [
     {
-      "AllowedOrigins": ["https://*.pulumi.com"],
+      "AllowedOrigins": [
+        "https://*.pulumi.com",
+        "https://*.pulumi-dev.io"
+      ],
       "AllowedHeaders": ["Authorization"],
       "AllowedMethods": ["GET"],
       "MaxAgeSeconds": 300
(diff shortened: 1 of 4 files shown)


scripts/cors/cors.json
```diff
@@ -0,0 +1,13 @@
+{
+  "CORSRules": [
+    {
+      "AllowedOrigins": [
```
cnunciato, line 4: It's probably worth a comment here explaining that this is consumed by the Service. Also, do we somehow suppress this functionality in self-hosted contexts? Since those won't be running on these domains, we should probably just make sure the UI won't break when it can't load this info.

scripts/cors/cors.json
```diff
@@ -0,0 +1,13 @@
...
+        "https://*.pulumi.com",
+        "https://*.pulumi-dev.io"
+      ],
+      "AllowedHeaders": ["Authorization"],
```
cnunciato, line 8: Is this saying that auth headers are OK to send with these requests? If so, we probably should not be doing this -- but apologies if I'm reading this wrong.

cnunciato (changes requested):
Marking with "request changes" only to be on the safe side WRT that auth-header setting.

scripts/cors/cors.json
sean1588: ahhh yeah that is a very good point! I imagine we would want to show to self hosted if possible as they are pulumi users so the events will likely be relevant to them.

scripts/cors/cors.json
sean1588: yes you are right. thanks christian, we should not be doing this.

scripts/sync-and-test-bucket.sh
```diff
@@ -99,6 +99,9 @@ aws s3 cp "$metadata_file" "${destination_bucket_uri}/metadata.json" --region "$
 # Persist an association between the current commit and the bucket we just deployed to.
 set_bucket_for_commit "$(git_sha)" "$destination_bucket" "$(aws_region)"
 
+# Set cors configuration on bucket.
+aws s3api put-bucket-cors --bucket "$destination_bucket" --cors-configuration file://scripts/cors/cors.json
```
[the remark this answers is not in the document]
sean1588: ok, i'll add the region. this worked fine for me locally as is. though could be because I have the `AWS_REGION` env var set, so it took that.

scripts/cors/cors.json
sean1588: I am going to go ahead and hide the events dashboard card from self hosted folks for now. This is something we had wanted to rollout quickly to see what the response is like first and then plan to iterate on it more in the future, so its fine for now if we do not show to them.

sean1588:
@cnunciato - I removed the authorization header from the allowed list and also added a couple extra entries to the allowed origins list to include localhost:3000 and localhost:4200, so that we can render this when running locally. also, i'm going to make the adjustment in the console code to exclude this from self-hosted users for now. Let me know what you think.

PR #8823, opened 2023-04-04
add exclusions to link checker

susanev:
we regularly get 403s on these links, but they work just fine for us
so adding them to the exclude list


sean1588 (approved):
🚀

PR #8877, opened 2023-04-12
fix script syntax error

sean1588:


scripts/sync-and-test-bucket.sh
```diff
@@ -103,7 +103,7 @@ set_bucket_for_commit "$(git_sha)" "$destination_bucket" "$(aws_region)"
 aws s3api put-bucket-cors --bucket "$destination_bucket" --cors-configuration "file://scripts/cors/cors.json" --region "$(aws_region)"
 
 # Finally, if it's a preview, post a comment to the PR that directs the user to the resulting bucket URL.
-if [ "$1" == "preview" && "$GITHUB_WORKFLOW" == "Build and deploy" ]; then
```
kmosher, line 106: other option is `[[ "$1" == "preview" && "$GITHUB_WORKFLOW" == "Build and deploy" ]]`, but this works too.

kmosher (approved):

scripts/sync-and-test-bucket.sh
sean1588: thank you!

sean1588: i changed it. this way is much more pleasing to look at imo.

PR #8881, opened 2023-04-12
ignore 403s in link checker

susanev:
theres nothing we can do about these
and i think its a limitation of this tool
that we ping things like geekwire multiple times and it causes these

open to a solution, if anyone has ideas


sean1588 (approved):
thanks susan! yeah makes sense to do this for now, we can think about an alternative solution for this.

PR #8923, opened 2023-04-20
404s are under control now, so adding 403s back

susanev:
404s are under control now, so adding 403s back


sean1588 (commented):
nice :rocket:

PR #8928, opened 2023-04-21
update code comment

sean1588:
Bringing this code comment up to date, to avoid confusion down the road since this is no longer the case anymore as of https://github.com/pulumi/docs/pull/8921.  We are no longer need `public-read` to be on the bucket, since this is being applied at the object level.


susanev (approved):

cnunciato (approved):
Thank you! 🙏

PR #9115, opened 2023-05-18
Fix CSS caching rules

cnunciato:
Somewhere along the way, our CSS caching rule fell out of sync with our filenames. This change updates those rules to align with the files we're shipping today. (See [this thread](https://pulumi.slack.com/archives/C85BS3LJZ/p1684438150667419) for context/discussion.)


sean1588 (approved):
thank you christian!

PR #9230, opened 2023-06-02
Add Algolia search

cnunciato:
This PR adds the a new Node.js program to be run as part of the `build-and-deploy` workflow to update our new Algolia search indexes. PR previews write to an index called `preview` and website deployments write to `production`.

Comments are inline, and links to all relevant Algolia documentation are sprinkled throughout. It works like this:

* Generate Registry docs and run a Hugo build, as we do today
* Run `scripts/update-search-index.sh`, passing the name of the index to update (`preview` or `production`)

Merging this PR will have no effect on the user experience today, as there's not yet any front-end component to this; all this PR does is add a step (after the website is successfully deployed) to write search data to Algolia. Once this is running and we're comfortable with it, we can PR and merge the front-end code to replace the Swiftype search experience with Algolia autocomplete across the board. Plan is to make the index environment-specific as well so we can review any changes to the Algolia code by browsing PR previews and seeing the impact those changes have on search results. 

Fixes https://github.com/pulumi/devrel-team/issues/566.


.github/workflows/check-links.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Install Hugo
         uses: peaceiris/actions-hugo@v2
         with:
-          hugo-version: '0.96.0'
+          hugo-version: '0.111.0'
```
cnunciato, line 23: Upgrading Hugo to use its `Ancestors` feature, which makes breadcrumbing _way_ easier.

susanev:
thank you!!!

lets pls hold of merging this until you and Sean are able to talk through all the build impacts and he has a chance to give feedback and you have a chance to make adjustments, thank you!

scripts/search/rank.js
```diff
@@ -0,0 +1,160 @@
...
+
+            // Top-level pages again rank highest.
+            if (page.kind === "section") {
+                return 990;
+            }
+
+            return 980;
+        }
+
+        // Docs pages are next.
```
susanev, line 75: i didnt have a chance to bring this up in the meeting but we get way more daily visitors to registry than docs. should registry come before docs?

scripts/search/settings.js
```diff
@@ -0,0 +1,84 @@
...
+    // https://www.algolia.com/doc/api-reference/api-parameters/attributesForFaceting/
+    getAttributesForFaceting() {
+        return [
+            "searchable(section)",
+        ];
+    },
+
+    // Synonyms allow us to surface content that may not directly match users' queries.
+    // https://www.algolia.com/doc/api-reference/api-methods/save-synonyms/#save-synonyms
+    getSynonyms() {
```
susanev, line 17: can we pull on the synonyms from swifttype and add them here?

cnunciato:
> lets pls hold of merging this until you and Sean are able to talk

Sean is listed as a reviewer, yeah. No intention of meeting before he reviews. 😉

scripts/search/rank.js
cnunciato: I reviewed GA as part of this, yeah, and based on all the learning and experimentation I've done to this point, I'm pretty sure the answer to this question is no, because Algolia custom rankings aren't really meant to convey popularity; they exist to let us break ties between records that otherwise rank similarly. 

The [ranking criteria docs](https://www.algolia.com/doc/guides/managing-results/relevance-overview/in-depth/ranking-criteria/) explain all this in detail (and are worth reading), but the gist is that for a given query, a bunch of criteria -- text matching against attributes, synonyms, etc. -- are applied first, before custom ranking ever comes into play. Custom ranking is only applied when multiple records match a given query closely enough that Algolia can't tell which ones are "more relevant" than others. Otherwise, they don't carry much weight (if any at all).

For us, where this matters is for queries like "input" or "stack" that can match strongly for content in multiple categories. For unqualified queries like these, users will most likely be looking for Pulumi-specific content -- rather than, say, the API docs for the `aws.medialive.Input` or `aws.cloudformation.Stack` resources in the Registry (which would match for those terms). For this reason, we give docs a higher custom rank -- the assumption being that someone looking for docs about CloudFormation stacks will use additional words, like "cloudformation", to narrow the focus of those queries on their own. Also, because of how much _more_ content there is in the Registry, if we did rank Registry pages higher than docs, we'd be increasing the likelihood that our own docs content would be "drowned out" by the sheer volume of stuff that exists in the Registry; the more packages and resources we have, the more words there are to compete with our core terminology. Where legitimate ambiguity does exist -- e.g., for queries like "aws" or "kubernetes" or "awsx", where we can't really guess at what the user is looking for -- ranking docs above Registry is an overall safer bet -- especially now that we have those fine cloud-specific landing pages to help them find what they're looking for (and perhaps query more narrowly next time).

So all up, my hunch is that ranking docs above Registry is the right way to go -- but again, still learning, so I reserve the right to change my mind. 😇

scripts/search/settings.js
cnunciato: Yep -- and I did, for the most part. It isn't quite an apples-to-apples thing, as these services are different, and many of the synonyms we added to Swiftype were account for typos, which Algolia [handles natively](https://www.algolia.com/doc/guides/managing-results/optimize-search-results/typo-tolerance/) (and I believe I verified all of them by hand to make sure), so there isn't any reason to bring those over directly. A few others aren't exactly "synonyms", either (e.g., "secret" and "passphrase", which refer to different things), so I left those out as well. But otherwise, they should look pretty similar, and I've added a few of my own as well.

scripts/ci-push.sh
```diff
@@ -6,7 +6,9 @@ source ./scripts/ci-login.sh
 
 ./scripts/build-site.sh
 ./scripts/sync-and-test-bucket.sh
+
 node ./scripts/await-in-progress.js
-./scripts/run-pulumi.sh update
 
+./scripts/run-pulumi.sh update
+./scripts/update-search-index.sh production
```
sean1588, line 13: can we add a comment here about the production argument passed in at the end and that it relates to the index/environment being updated in algolia itself?

scripts/search/settings.js
```diff
@@ -0,0 +1,84 @@
...
+            ["F#", "fsharp"],
+            ["cfn", "cloudformation"],
+            ["tf", "terraform"],
+            ["stack reference", "stackreference"],
+            ["openid connect", "oidc"],
+            ["crosswalk", "awsx"],
+            ["crossguard", "pac", "policy as code"],
+            ["aws", "aws classic"],
+            ["azure", "azure native"],
+            ["how-to guide", "how to", "example"],
```
sean1588, line 34: should we include a synonym without the dash in `how-to guide` as well?

scripts/update-search-index.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+node --max-old-space-size=8192 ./scripts/search/main.js "$1"
```
sean1588, line 3: would it make sense to just configure `--max-old-space-size` using the `NODE_OPTIONS` env variable in the workflow config file?

.github/workflows/generate-provider-docs.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       - name: Install Hugo
         uses: peaceiris/actions-hugo@v2
         with:
-          hugo-version: '0.96.0'
+          hugo-version: '0.111.0'
```
sean1588, line 64: will we  need to update the version of hugo we are running locally in order to build the site locally going forward as well? I'm guessing this new version is what is allowing us to add JSON as an output type?

scripts/update-search-index.sh
cnunciato: Ah, forgot that was there. Yep! I’ll try that.

.github/workflows/generate-provider-docs.yml
cnunciato: Only if you want to build pulumi/docs locally; pulumi-hugo and others don’t need it — although we should probably upgrade everywhere (just as a rule) so everything’s on the same version. I’m happy to to that.

This PR uses the newer version specifically for the `Ancestors` property, though, just so you’re aware. We use this property in the JSON output script, but JSON output doesn’t require it specifically.

scripts/ci-push.sh
cnunciato: Added a comment to `update-search-index` to clarify what these values are and how they're used.

scripts/search/settings.js
cnunciato: I actually meant this to be "how to guide", so will add. Thanks for spotting!

sean1588 (approved):
looks good! thanks christian! :rocket:

PR #9283, opened 2023-06-09
Adjustments to prepare for search cutover

cnunciato:
Adds a few things in preparation for cutover:

* Allow-list the environment variables needed by the UI (so the search boxes know which app and index to use)
* Rank non-beta Kubernetes APIs higher than beta ones
* Add a rule to return the Inputs & Outputs page as the top result for queries for "output" or "outputs"

Also writes settings locally as well. (These files aren't technically necessary, they just help with local debugging, and could be uploaded as release artifacts if we wanted to store them more permanently.)

### DIFF
--- a/scripts/search/settings.js
+++ b/scripts/search/settings.js
@@ -1,3 +1,5 @@
+const page = require("./page");
+
 // Algolia index settings.
 //
 // Changes to this file should be made with extreme care.
@@ -34,6 +36,7 @@ module.exports = {
             ["provider", "package"],
             ["pulumi service", "pulumi console", "pulumi cloud"],
             ["rbac", "role based access control"],
+            ["saml", "sso"],
             ["stack reference", "stackreference"],
             ["tf", "terraform"],
             ["vm", "virtual machine"],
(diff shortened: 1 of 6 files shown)


scripts/search/settings.js
```diff
@@ -104,8 +107,36 @@ module.exports = {
...
+                                page.getObjectID({ href: "/docs/pulumi-cloud/" }),
+                            ],
+                            position: 0,
+                        },
+                    ],
+                    filterPromotes: true,
+                },
+            },
+
+            // When the query is for "output" or "outputs", deliver the Inputs & Outputs page as the top result.
```
sean1588, line 119: does this need to be done for querying the word "input" too, or is it only output that turns up other results?

sean1588 (approved):

scripts/search/settings.js
cnunciato: Actually it turns out we don't need this when we change the behavior of the `title` attribute to be `unordered`, which I've done here instead. (Didn't feel great having to do this, so hopefully that'll work in general. It's [documented here](https://www.algolia.com/doc/guides/managing-results/must-do/searchable-attributes/how-to/configuring-searchable-attributes-the-right-way/#understanding-word-position), btw.)

PR #9340, opened 2023-06-15
Remove some link-checker exceptions

cnunciato:


sean1588:
thanks, yeah we can remove these now.

sean1588 (approved):

PR #9365, opened 2023-06-21
Separate docs builds deploy to test account

sean1588:
fixes: https://github.com/pulumi/home/issues/2875

This PR configure docs to start deploying to the test environment. See [design doc](https://docs.google.com/document/d/10LVEG6UeCIFkprd5IzwTvHM-TYDg9xJlIp3qIFYVTSs/edit) for more context.

There will be a corresponding [PR](https://github.com/pulumi/pulumi-hugo/pull/3098) open in the hugo repo to make the changes to deploy to the test environment there as well.

Here is an outline of the changes:

**GitHub**:
- created two github environements, `testing` and `production`  with environment specific variables and reference environment in workflow
- add separate workflow to deploy in parallel to the aws test account

**Infrastructure**:
- provision a new bucket for pulumi-hugo to deploy to -(all routes will point this new bucket except for `/registry`, `/docs/install`, `/docs/cli/command`, and `/docs/reference/pkg`) in the testing env once this is merged
- provision a new bucket to house css and js bundles (this will house the bundles for both the hugo and docs repo builds)
- add cloudfront origins and behaviors for new buckets
- add testingFlow flag to conditionally configure this for the testing environment. The new buckets (bundles bucket, hugo bucket) will only be provisioned in production, but will not be wired up to cloudfront at this point.

**Scripts**:
- update await script to await updates happening in the pulumi-hugo workflows, since they share the same pulumi program and stack.
- update scripts to reference environment specific variables from the corresponding github environment (e.g. `production`, `testing`)

testing cdn endpoint: https://d298m5mqqr8wsc.cloudfront.net/

production preview: https://app.pulumi.com/pulumi/www.pulumi.com/www-production/previews/501c2c27-03a6-4815-a9f3-be4f7c5901db


infrastructure/index.ts
```diff
@@ -220,15 +295,166 @@ const baseCacheBehavior = {
...
+            },
+        },
+    )
+    bundleBehaviors.push(
+        {
+            ...baseCacheBehavior,
+            targetOriginId: bundlesBucket.arn,
+            pathPattern: "/css/*",
+            defaultTtl: oneHour,
+            maxTtl: oneHour,
```
zchase, line 410: Should we keep js/css assets around longer in the cache? Setting it to one hour just feels like we might run into issues where HTML is going to be requesting a version that is no longer available.

zchase (approved):
Overall looks good to me. I do think we will want to up the max cache time for assets but everything else looks good.

infrastructure/index.ts
sean1588: yeah we could increase it, but the previous versions of the assets will continue to live in the bundles bucket until we come up with a way to delete the old assets. Right now i'm just pushing in the new bundle files to that bucket when they get built, so the old ones are not being deleted as of now (which may not be a bad thing) so they will technically still be available. I'm thinking if we want we could just have a job that deletes assets that are over a month old or something like that if we want.

cnunciato:
Do we need this anymore, or can we close it?

cnunciato:
Closing as no longer needed.

PR #9380, opened 2023-06-23
Add resource properties to search

cnunciato:
Adds support for finding registry resources by property name.


package.json
```diff
@@ -20,6 +20,7 @@
     "http-server": "^0.12.1",
     "js-yaml": "^3.13.1",
     "jsdom": "^22.0.0",
+    "node-fetch": "2",
```
cnunciato, line 23: Once #9398 merges, we should probably update this PR to use Axios instead.

Issue #9391, opened 2023-06-26
Add a test to validate all URLs in the search index


cnunciato:
We currently crawl the pages of the website for 404s, but we don't yet do the same for the URLs we generate for the Algolia search index. We should do that, and run that check daily as part of the check-links workflow. (The link checker itself won't quite work for this -- at least not unaltered -- as it runs based on the sitemap, and its job is to check the links on each page, whereas the job of this check would just be to validate that each URL works.)

PR #9398, opened 2023-06-27
Add a script to check search URLs

cnunciato:
This change adds a script (and runs it daily) to that the URLs we send to Algolia are valid:

```
Downloading the 'production' index...
Checking 30062 search URLs...
 ↳ Checking group 1 of 31 (1000 URLs)...
 ...
 ↳ Checking group 8 of 31 (1000 URLs)...
   ↳ https://www.pulumi.com/registry/packages/azure-native/api-docs/databoxedge/getkubernetesrole/ failed (read ECONNRESET). Retrying...
   ↳ https://www.pulumi.com/registry/packages/azure/api-docs/armmsi/ failed (read ECONNRESET). Retrying...
   ...
```

Fixes https://github.com/pulumi/docs/issues/9391.

### DIFF
--- a/.github/workflows/check-search-urls.yml
+++ b/.github/workflows/check-search-urls.yml
@@ -1,0 +1,30 @@
+name: "Scheduled jobs: Check search URLs"
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string.
+    # Run every day at 3:00PM UTC.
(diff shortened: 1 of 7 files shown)


scottslowe (approved):
Seems reasonable to me, approved

PR #9561, opened 2023-07-26
Post new issues to the Docs board instead

cnunciato:
Updates the Add Issues to Project workflow to post new issues to the Docs board instead of Developer Relations.

Part of https://github.com/pulumi/devrel-team/issues/607.

### DIFF
--- a/.github/workflows/add-to-project.yml
+++ b/.github/workflows/add-to-project.yml
@@ -8,8 +8,8 @@ jobs:
   add-to-project:
     runs-on: ubuntu-latest
     steps:
-      - name: Add to DevRel
+      - name: Add to Docs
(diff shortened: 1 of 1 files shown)


cnunciato:
Reposting the same comment I [posted here](https://github.com/pulumi/registry/pull/2918#issuecomment-1652725379):

>Hm, I didn't expect this to merge automatically. I did click auto-merge, but in doing so I expected for the auto-merge to happen only after at least one approval, which is normally the case. 
>
>Since we don't seem to have approval requirements set on this repo (which probably makes sense given how it's used), I've removed support for auto-merge to make it a little less easy for direct submitters to break the build. If necessary, we can re-enable it and/or explore other options.

PR #9578, opened 2023-07-28
Address broken links

cnunciato:
One lil' link fix to close out the last one of these.

Fixes https://github.com/pulumi/docs/issues/9439.


ericrudder (approved):
This looks sane.

I happened to look at the list and noticed that Twitter was in the list ... just a heads up that x.com now redirects to twitter and might also become primary, just in case this affects your thinking on including.

PR #9619, opened 2023-08-03
Deploy registry independently

sean1588:
This PR is part of phase 1 of the docs build separation. There is a [PR](https://github.com/pulumi/registry/pull/2963) also open in the registry repo that goes along with this. The registry PR should merge first.  This will enable registry to deploy independently and manage its own infrastructure. 

This PR adds some extra infrastructure configuration that will get deployed to the testing account whenever a pull request is merged into master.  It also sets up a workflow that will run in parallel, alongside the current deployment workflow that deploys to production today, which will deploy with the new infrastructure configuration to the test environment.

- **adds a bundles bucket and cloudfront origin to manage css/js files** - this bucket will manage the bundle files for both the registry build and the docs build. All /js and /css routes will resolve here. The new bundles produced by each build will get added to this bucket. This will not be in effect for production yet. This PR will only provision the bundles bucket in the production account, but not hook it up to cloudfront, so the current flow will continue to function as it does today.
- **adds a cloudfront origin for registry**  - routes requests under /registry to the registry CDN being managed in the registry PR.

Note: once these changes get applied to production we can remove the `resourcedocsgen` tool from this repo as well as the stage of the script which generates the registry files, since this will all now live in registry

testing env cdn endpoint: d298m5mqqr8wsc.cloudfront.net

production preview: https://app.pulumi.com/pulumi/www.pulumi.com/www-production/previews/462e222e-33b5-41da-9a98-0398dc0654b1

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -8,6 +8,9 @@ import { getLambdaFunctionAssociations } from "./cloudfrontLambdaAssociations";
 
 const stackConfig = new pulumi.Config();
 
+const stackRef = new pulumi.StackReference(`pulumi/registry/registry-testing`)
+const registryCDN = stackRef.getOutput("cloudFrontDomain");
+
 const config = {
     // websiteDomain is the domain/host to serve content at.
     websiteDomain: stackConfig.require("websiteDomain"),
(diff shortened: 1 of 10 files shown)


infrastructure/index.ts
```diff
@@ -224,15 +258,113 @@ const baseCacheBehavior = {
...
+            domainName: registryCDN,
+            customOriginConfig: {
+                originProtocolPolicy: "https-only",
+                httpPort: 80,
+                httpsPort: 443,
+                originSslProtocols: ["TLSv1.2"],
+            }
+        }
+    );
+    registryBehaviors.push(
```
sean1588, line 331: remember to forward cookies and headers

zchase (approved):

PR #9664, opened 2023-08-11
Added supported languages and example

toriancrane:
### Proposed changes

On the `pulumi convert` cli page, added a table to demonstrate the supported languages for source and destination. Also added a simple yaml to java example. This will help to make it clearer to readers that YAML, Terraform, and PCL are currently the only supported source languages, and all of Pulumi's supported languages are available as a target language. The simple example will provide an easy to use reference for how to use this command.

### Related issues (optional)

Fixes [Issue 12720](https://github.com/pulumi/pulumi/issues/12720)

### DIFF
--- a/content/docs/cli/commands/pulumi_convert.md
+++ b/content/docs/cli/commands/pulumi_convert.md
@@ -14,6 +14,13 @@ Convert Pulumi programs from a supported source program into other supported lan
 
 The source program to convert will default to the current working directory.
 
+### Supported Languages
+| - | Language |
+|--------------|-------------|
+| **Source Language** | `yaml \| terraform \| pcl` |
+| **Target Language** | `typescript \| python \| dotnet \| go \| java \| yaml \| pcl` |
+
+### Command
 
 ```
 pulumi convert [flags]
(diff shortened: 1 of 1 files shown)


content/docs/cli/commands/pulumi_convert.md
```diff
@@ -14,6 +14,13 @@ Convert Pulumi programs from a supported source program into other supported lan
 
 The source program to convert will default to the current working directory.
 
+### Supported Languages
```
scottslowe, line 17: I could be wrong, but I don't think we need to expose PCL here. PCL isn't documented (I believe that's intentional) as it is an internal representation that Pulumi uses.

scottslowe (changes requested):
Useful addition, thanks! Just one comment regarding whether PCL needs to be included/exposed here.

cnunciato:
@toriancrane These docs are generated, so any changes we make to them directly will be overwritten on the next CLI release. If we want to augment them with additional content, we'll probably need to do so with something like a shortcode or change to the layout somehow.

cnunciato:
@scottslowe Yeah, we chatted about this a bit this morning at triage. Since we don't yet have a good way to inject content into these generated pages, I thought we might be able to address the core issue (conveying which languages are supported) in the CLI help text itself somehow, e.g., here: https://github.com/pulumi/pulumi/blob/2016359133bb5c7d3b213c7f6b9db212cf35e283/pkg/cmd/pulumi/convert.go#L62-L77

.. and/or on the `from` or `language` options themselves: https://github.com/pulumi/pulumi/blob/2016359133bb5c7d3b213c7f6b9db212cf35e283/pkg/cmd/pulumi/convert.go#L79-L88

Not perfect of course, but perhaps a good interim solution.

cnunciato (commented):
Do y'all think we should close this, given it can't really merge as-is?

scottslowe:
> Do y'all think we should close this, given it can't really merge as-is?

Yeah, that's probably the right thing to do.

PR #9728, opened 2023-08-25
Substitute `/ai` with AI App backend

AaronFriel:


infrastructure/index.ts
```diff
@@ -434,6 +435,34 @@ const distributionArgs: aws.cloudfront.DistributionArgs = {
...
+            maxTtl: 0,
+        },
+        {
+            ...baseCacheBehavior,
+            // allow all methods
+            allowedMethods: ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"],
+            cachedMethods: [
+                "GET", "HEAD", "OPTIONS",
+            ],
+            targetOriginId: aiAppDomain,
```
cnunciato, line 461: Running against my dev stack now, but do we need to add an `origin` for this also? I see there's one for `previewAIAppDomain`.

cnunciato: Yeah, looks like it:

```
error updating CloudFront Distribution (E3GY8OM5MAPY0D): NoSuchOrigin: One or more of your cache behavior target origins or origin groups do not exist.
```

cnunciato (changes requested):
I think all we need is to add this to the list of `origins` and we're good: 

```
 {
    originId: aiAppDomain,
    domainName: aiAppDomain,
    customOriginConfig: {
        originProtocolPolicy: "https-only",
        httpPort: 80,
        httpsPort: 443,
        originSslProtocols: ["TLSv1.2"],
    },
},
```

cnunciato (approved):
🚀

PR #9899, opened 2023-09-20
Add a sitemap-index file

cnunciato:
Part of https://github.com/pulumi/pulumi.ai/issues/206. Should be merged after https://github.com/pulumi/pulumi.ai/pull/251, as that PR adds the `/ai` sitemap.


sean1588 (approved):
thanks @cnunciato. this will be great for when I need to reference the registry map here as well.

PR #9921, opened 2023-09-26
Fix bucket cleanup

sean1588:
fixes: https://github.com/pulumi/docs/issues/9890

The bucket clean up script is now passing, but was finding 0 buckets to delete. This is because it was missing an [environment variable](https://github.com/pulumi/docs/blob/master/scripts/common.sh#L63) that we use to generate the bucket name now, which it uses to look up the buckets to delete.

https://github.com/pulumi/docs/actions/runs/6314574721/job/17145163104#step:7:31

### DIFF
--- a/.github/workflows/bucket-cleanup.yml
+++ b/.github/workflows/bucket-cleanup.yml
@@ -44,3 +44,4 @@ jobs:
             PULUMI_STACK_NAME: ${{ vars.PULUMI_STACK_NAME }}
             SLACK_ACCESS_TOKEN: ${{ secrets.SLACK_ACCESS_TOKEN }}
             GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


cnunciato (approved):
LGTM, but do you know if it's possible to read from `jobs.jobName.environment` directly? Just to remove the layer of indirection (which basically just says that the deployment environment variable for the `testing` environment is also `testing`).

sean1588:
> LGTM, but do you know if it's possible to read from `jobs.jobName.environment` directly? Just to remove the layer of indirection (which basically just says that the deployment environment variable for the `testing` environment is also `testing`).

yeah I had considered this and looked into this a bit when I was originally doing this, but couldn't quite get it working. Though I assume it must be possible to source this from the job context somehow. But then I figured maybe it is better to just explicitly add a variable (even though it **is** a bit redundant), just for the purpose of making it more explicit to follow since you can know exactly what is in there just by looking at the repo settings (i.e. does `jobs.jobName.environment` return just the actual name of the environment or is it some identifier or something). So I ended up just resorting to that, though it does _feel_ code smelly.

PR #9928, opened 2023-09-27
Upload docs index to s3 bucket and configure cronjob

sean1588:
This PR writes a file containing the docs specific search index and uploads it to s3 under the root directory and configures a cron job to access these files and combine the index with the one in registry then push to Algolia. The job will only push to the `testing` index in algolia for now. In prod we will continue to push the indexes to algolia as we are currently until we can confirm the testing index looks right. This PR should not change anything we are currently doing, it is only writing a separate index file specifically containing the non-registry related objects and pushing it to the s3 bucket and configuring a cron job which will only push to the testing index for now.

related registry [PR](https://github.com/pulumi/registry/pull/3215) to push registry indexes.

### DIFF
--- a/.github/workflows/update-search-index.yml
+++ b/.github/workflows/update-search-index.yml
@@ -1,0 +1,42 @@
+name: "Scheduled jobs: Update Seach Index"
+on:
+  schedule:
+    - cron:  '*/60 * * * *'
+  workflow_dispatch:
+jobs:
+  merge:
+    name: Update Algolia Search Index
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: '18.x'
+
+      - name: Check out the repo
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.PULUMI_BOT_TOKEN }}
+
+      - name: Update search indexes
+        run: make update_search_indexes
+        env:
+          ALGOLIA_APP_ID: ${{ vars.ALGOLIA_APP_ID }}
+          ALGOLIA_APP_SEARCH_KEY: ${{ vars.ALGOLIA_APP_SEARCH_KEY }}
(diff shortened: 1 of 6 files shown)


scripts/update-search-index.sh
```diff
@@ -1,5 +1,18 @@
 #!/bin/bash
```
sean1588, line 1: I'll update this file name to be generate-search-index.sh once we remove the parts that are actually doing the update.

.github/workflows/update-search-index.yml
```diff
@@ -0,0 +1,44 @@
+name: "Scheduled jobs: Update Seach Index"
+on:
+  schedule:
+    - cron:  '*/60 * * * *'
+  workflow_dispatch:
+jobs:
+  update:
+    name: Update Algolia Search Index
+    runs-on: ubuntu-latest
+    environment: testing
```
sean1588, line 10: we will only publish the index to the testing index in Algolia now, to confirm things there first where we can compare to prod.

.github/workflows/update-search-index.yml
```diff
@@ -0,0 +1,44 @@
+name: "Scheduled jobs: Update Seach Index"
```
cnunciato, line 1:
```suggestion
name: "Scheduled jobs: Update Search Index"
```

cnunciato (approved):
LGTM! 🚀

.github/workflows/update-search-index.yml
```diff
@@ -0,0 +1,44 @@
...
+        with:
+          node-version: '18.x'
+
+      - name: Check out the repo
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.PULUMI_BOT_TOKEN }}
+
+      - name: Update search indexes
+        run: make update_search_indexes
```
sean1588, line 23:
```suggestion
        run: make ci_update_search_index
```

PR #9943, opened 2023-09-28
Stop importing registry and genning api docs in docs

sean1588:
This PR stops importing the registry and stops genning the API docs here. We left this in for a bit so that we could continue genning the search indexes here. Now that the index generation has been split apart, we can stop building the registry in this repository. The index split work is deploying to the testing index in algolia and checking the testing index generated by the cron, ithas exactly the same number of records as does production 🎉 .  We should see a fairly significant reduction in build times once this goes in.

This PR does the folllowing:
- stop importing the registry repo in the Update Theme cron
- stop building the registry api docs
- stop updating the index, only generate the index files (the cron job will stitch these together and make the update)
- add a reference to the registry site map from the sitemap index file. 

Screen Shot 2023-09-29 at 11 19 06 AM

Second thing here is adding the reference to the registry sitemap to the sitemap index file. The registry specific sitemap will go out when this [PR](https://github.com/pulumi/registry/pull/3249) gets merged. I am going to need to merge that one right after this one and pause the workflows so this repo does not import that change from registry, since the sitemap file from registry will overwrite this one. It looks like there isn't a way around this (at least from what I understand from the [hugo documentation](https://gohugo.io/templates/sitemap-template/) stating that this file need to be placed under `themes/default`).  This will no longer be an issue once we stop importing the registry module here, which is why I need to roll this out this way so we don't end up in a state where the file gets overwritten with the registry one.


sean1588:
I think I may need to adjust this a bit more, so going to hold off on merging.

cnunciato (approved):
LGTM. Let's just make sure this all works as expected in the test environment (I believe we already have multiple times, but just calling it out one last time) and if so, 🚀 .

PR #9946, opened 2023-09-28
Add bucket ACL to enable cloudfront logging

sean1588:
fixes: https://github.com/pulumi/home/issues/2873

AWS changed their policy with regards to cloudfront logging, now requiring you to explicitly specify an ACL to grant the `awslogsdelivery` role write access to the logs bucket. This PR adds that ACL. I ran the update against the test account and logs are now being delivered in the testing environment :tada:

see: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html#AccessLogsOverview


cnunciato (approved):
LGTM!

PR #10004, opened 2023-10-05
Adjust import cron cadence

sean1588:
Now that we've shaved off around 40 mins of the deployment time by yanking out registry, we can probably run the update hugo modules cron on a tighter cadence. This will move it from hourly to every half hour.


cnunciato (approved):
🎉

PR #10009, opened 2023-10-06
Remove resourcedocsgen tool

sean1588:
This now lives in registry and no longer needed since we stopped building api docs here. Removing from here to avoid confusion.


cnunciato (approved):
Thanks! ✨

PR #10019, opened 2023-10-07
Process S3 redirects in parallel

cnunciato:
Updates the `make-s3-redirects.sh` script to process S3 301 redirects in parallel with Node.js. 

Takes about 2 seconds to run this locally, so should shave 20-30 minutes off the tail end of the `push` job. 🎉 

Run `./scripts/on-demand-build-sync-test.sh` to see it run in the context of a build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "minify-css": "node scripts/minify-css.js"
   },
   "dependencies": {
+    "@aws-sdk/client-s3": "^3.427.0",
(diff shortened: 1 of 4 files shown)


sean1588:
LGTM. nice that it only took two seconds. It took around 1 minute using bash when I did it locally with that I was doing before. yay node!! 🚀

sean1588 (approved):

PR #10047, opened 2023-10-11
Update `static/esc/latest-version`

pgavlin:
Update `static/esc/latest-version` to `0.5.2`.


static/esc/latest-version
```diff
@@ -1 +1 @@
-0.5.1
\ No newline at end of file
+0.5.2
```
pgavlin, line 1: is the lack of a newline in the original intentional?

pgavlin: looks like it is

lblackstone (approved):

PR #10053, opened 2023-10-11
stop trying to build resourcedocsgen in scripts

tgummerer:
resourcedocsgen no longer lives in this repository, so we shouldn't try to build it from here any longer (it was removed in https://github.com/pulumi/docs/pull/10009).

### DIFF
--- a/scripts/gen_resource_docs.sh
+++ b/scripts/gen_resource_docs.sh
@@ -86,16 +86,6 @@ generate_docs() {
 
     echo "Running docs generator from schema for ${provider}..."
 
-    pushd ${TOOL_RESDOCGEN}
-
-    go mod tidy
-
-    if [ -z "${GOPATH:-}" ]; then
-        echo "GOPATH is empty. Defaulting to ${HOME}/go"
-        GOPATH="${HOME}/go"
-    fi
-
-    go build -o "${GOPATH}/bin/resourcedocsgen" .
     resourcedocsgen docs \
       --docsOutDir "${ABSOLUTEPACKDIR}/${provider}/api-docs" \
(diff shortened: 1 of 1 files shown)


tgummerer:
It is still used in `pulumi/pulumi` (here: https://github.com/pulumi/pulumi/blob/master/.github/workflows/ci-test-docs-generation.yml#L125-L126), where I noticed it because a CI build was failing.

This PR (plus https://github.com/pulumi/pulumi/pull/14157) is intended to get that to work again.

cnunciato:
@tgummerer Apologies if I'm misunderstanding, but I think this is the wrong fix. What you have in https://github.com/pulumi/pulumi/pull/14157 seems generally correct and should remove the need to call this file at all anymore.

tgummerer:
It still invokes the file here: https://github.com/pulumi/pulumi/pull/14157/files#diff-5ba8315cae25efa261aeaba65beba56f61ee7f6dd77f3dcc99c6fe3d9989bf7aR142-R143

Though maybe it would be better to just do what the script does in the action directly, or have a different script in `pulumi/pulumi` if that's really the last user.  Let me try to do that.

tgummerer:
On second thought, this script seems to be used in other places as well, e.g. https://github.com/pulumi/scripts/blob/master/ci/build-package-docs.sh#L74.  Maybe the script should move to `pulumi/scripts` though?  It seems to me that we should preserve it in some way though.

tgummerer:
@cnunciato this just came up again.  This script is used in a CI job that's not run frequently, but it is still used.  Would you mind taking another look?

cnunciato (approved):

PR #10078, opened 2023-10-12
Remove go from the build-and-deploy workflow

cnunciato:
Since we aren't building resource docs anymore here, we can shave a few seconds off the build by not installing Go.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -24,18 +24,10 @@ jobs:
         with:
           node-version: '18.x'
 
-      - uses: actions/setup-go@v2
-        with:
-          go-version: 1.19.x
-
       - uses: peaceiris/actions-hugo@v2
         with:
           hugo-version: '0.111.0'
           extended: true
-
-      - name: Configure Git
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-deploy.yml
```diff
@@ -24,19 +24,11 @@ jobs:
...
-      - uses: actions/setup-go@v2
-        with:
-          go-version: 1.19.x
-
       - uses: peaceiris/actions-hugo@v2
         with:
           hugo-version: '0.111.0'
           extended: true
 
-      - name: Configure Git
```
cnunciato, line 36: This used to do something, but not anymore, so we can remove it also.

sean1588 (approved):
Thanks @cnunciato! lgtm, assuming Hugo doesn't have any dependencies on needing go to be installed? Does it implicitly still need go to do any of the things we need to do like downloading modules specified in the go.mod file, etc?

cnunciato:
>Does it implicitly still need go to do any of the things we need to do like downloading modules specified in the go.mod file, etc?

Nope! Hugo does all of that natively.

cnunciato:
Oh! I stand corrected:

https://gohugo.io/commands/hugo_mod/

```
Various helpers to help manage the modules in your project’s dependency graph. Most operations here requires a Go version installed on your system (>= Go 1.12) and the relevant VCS client (typically Git).
```

I'll verify and close if this indeed the case.

cnunciato:
Looks like it is! I'll close. Thanks for asking!

PR #10092, opened 2023-10-17
Fix Pulumi CLI command docs

justinvp:
Manual fixes for the broken links, in the meantime until the fix for https://github.com/pulumi/pulumi/issues/14237 lands.


sean1588 (approved):
nice! thank you @justinvp for the intermediary fix 🌟

PR #10093, opened 2023-10-17
Add workflow to update esc latest-version

lblackstone:
### Proposed changes
Adapted from https://github.com/pulumi/docs/blob/master/.github/workflows/pulumi-cli.yml

I cut out a bunch of steps that I don't think are relevant for ESC, but that would be good to check.

### Related issues (optional)
Part of https://github.com/pulumi/esc/issues/96

### DIFF
--- a/.github/workflows/esc-cli.yml
+++ b/.github/workflows/esc-cli.yml
@@ -21,7 +21,7 @@ jobs:
         with:
           source_branch: "pulumi/${{ github.run_id }}-${{ github.run_number }}"
           destination_branch: "master"
-          pr_title: "Regen docs esc@${{ env.PULUMI_VERSION }}"
+          pr_title: "Regen docs esc@${{ env.ESC_VERSION }}"
           pr_body: "Automated PR"
           pr_label: "automation/esc-cli-docs,automation/merge"
           github_token: ${{ secrets.PULUMI_BOT_TOKEN }}
--- a/.github/workflows/esc-cli.yml
+++ b/.github/workflows/esc-cli.yml
@@ -19,7 +19,7 @@ jobs:
       - name: pull-request
         uses: repo-sync/pull-request@v2
         with:
-          source_branch: "pulumi/${{ github.run_id }}-${{ github.run_number }}"
+          source_branch: "esc/${{ github.run_id }}-${{ github.run_number }}"
           destination_branch: "master"
           pr_title: "Regen docs esc@${{ env.ESC_VERSION }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/esc-cli.yml
```diff
@@ -0,0 +1,95 @@
...
+        uses: actions/checkout@v2
+      - name: set the esc version
+        run: |
+          echo "ESC_VERSION=${{ github.event.client_payload.ref }}" >> $GITHUB_ENV
+      - name: pull-request
+        uses: repo-sync/pull-request@v2
+        with:
+          source_branch: "pulumi/${{ github.run_id }}-${{ github.run_number }}"
+          destination_branch: "master"
+          pr_title: "Regen docs esc@${{ env.PULUMI_VERSION }}"
```
pgavlin, line 24: should this be `ESC_VERSION`?

```suggestion
          pr_title: "Regen docs esc@${{ env.ESC_VERSION }}"
```

lblackstone: Yes, good catch

.github/workflows/esc-cli.yml
```diff
@@ -0,0 +1,95 @@
...
+        run: git status && git diff
+        working-directory: docs
+      - name: commit changes
+        run: |
+          git config --local user.email "bot@pulumi.com"
+          git config --local user.name "pulumi-bot"
+          git checkout -b pulumi/${{ github.run_id }}-${{ github.run_number }}
+          git add static/
+          git add content/
+          git add data/
```
sean1588, line 73: this only updates the latest version in the `static/esc/latest-version` file, right? Do the content and data directories need to be added here?

.github/workflows/esc-cli.yml
```diff
@@ -0,0 +1,95 @@
...
+        working-directory: docs
+    strategy:
+      matrix:
+        nodeversion:
+          - "18.x"
+  notify:
+    if: failure()
+    name: Send slack notification
+    runs-on: ubuntu-latest
+    needs: [pull-request, build-pulumi-cli-docs]
```
sean1588, line 85:
```suggestion
    needs: [pull-request, build-pulumi-esc-docs]
```

.github/workflows/esc-cli.yml
lblackstone: Yes, just the `latest-version` file for now. I'll drop those directories.

pgavlin (approved):

PR #10099, opened 2023-10-18
Fix esc-cli workflow

lblackstone:
### Proposed changes
Explicitly use bash for a shell command


lblackstone:
Never mind, I just noticed that the failing workflow was already using bash

PR #10137, opened 2023-10-25
Stop pushing CSS and JS bundles to a separate bucket

cnunciato:
Follow-up to https://github.com/pulumi/registry/pull/3397, this change stops pushing CSS and JS bundles to a separate S3 bucket and instead serves them out of the same bucket as the website build. 

This is the first of two PRs: The first sets the bucket to `forceDestroy: true`, so it can be deleted later, and the second will actually delete the bucket. Verifying first in the test environment to make sure all looks well before merging.

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -118,28 +118,10 @@ const uploadsBucketAcl = new aws.s3.BucketAclV2("uploads-bucket-acl", {
 });
 
 const bundlesBucket = new aws.s3.Bucket("bundles-bucket", {
+    forceDestroy: true,
     website: {
         indexDocument: "index.html",
     },
-});
-
-const bundlesBucketPublicAccessBlock = new aws.s3.BucketPublicAccessBlock("bundles-public-access-block", {
-    bucket: bundlesBucket.id,
-    blockPublicAcls: false,
-});
-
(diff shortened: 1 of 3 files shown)


infrastructure/index.ts
```diff
@@ -294,56 +276,6 @@ if (config.registryStack) {
     const registryStack = new pulumi.StackReference(config.registryStack);
     const registryCDN = registryStack.getOutput("cloudFrontDomain");
 
-    bundleOrigins.push(
```
sean1588, line 297: you can probably  remove this line as well that concats the bundles origins to the overall origins array in the distribution args.
https://github.com/pulumi/docs/pull/10137/files#diff-32b0c3347657743da0ef35b9a3f3d8eca1191c2d669458dd8bfd202c0eeda87dR364

and this line also that adds the behaviors:
https://github.com/pulumi/docs/pull/10137/files#diff-32b0c3347657743da0ef35b9a3f3d8eca1191c2d669458dd8bfd202c0eeda87dL446

cnunciato:
Here's the update that deployed this to the testing environment. So far, all looks well: https://app.pulumi.com/pulumi/www.pulumi.com/www-testing/updates/235

infrastructure/index.ts
cnunciato: Done.

sean1588 (approved):
lgtm! thanks @cnunciato  :rocket:

PR #10300, opened 2023-11-28
Replace infinite redirect with placeholder page

cnunciato:
The page we have now works, but is strange in that it redirects to itself. This replaces that redirect with a blank page that should (🤞) have the same end result and also provide a little explanation as to why things are the way they are.


sean1588 (approved):
yup, this looks like it should work and give us the behavior we need! :rocket:

PR #10395, opened 2023-12-14
Add bucket policy for uploads in testing

zchase:
### Proposed changes

Adds a bucket policy to the testing bucket so we can upload files from an internal app.


infrastructure/Pulumi.www-production.yaml
```diff
@@ -3,11 +3,12 @@ config:
   www.pulumi.com:addSecurityHeaders: "true"
   www.pulumi.com:certificateArn: "arn:aws:acm:us-east-1:388588623842:certificate/9db6a76b-f7ba-465b-ab96-ce1d3b8ae02c"
   www.pulumi.com:doEdgeRedirects: "true"
+  www.pulumi.com:hostedZone: www.pulumi.com
```
sean1588, line 6: are all these values in the stack config files the same as they were before except for the `marketingPortalStack`? I guess it was the ordering that was switched up.

sean1588 (approved):

PR #10574, opened 2024-01-12
Re-add bucket policy

zchase:
### Proposed changes

This re-adds the bucket policy, allowing uploads from the marketing portal.

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -94,6 +94,34 @@ const uploadsBucket = new aws.s3.Bucket("uploads-bucket", {
     }],
 });
 
+if (pulumi.getStack() === "www-testing") {
+    const marketingAppStack = new pulumi.StackReference("pulumi/marketing-db/staging");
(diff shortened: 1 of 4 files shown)


infrastructure/index.ts
```diff
@@ -94,6 +98,43 @@ const uploadsBucket = new aws.s3.Bucket("uploads-bucket", {
     }],
 });
 
+if (config.marketingPortalStack) {
+    const marketingAppStack = new pulumi.StackReference(config.marketingPortalStack);
+    const ecsRoleArn = marketingAppStack.getOutput("ecsRoleArn");
```
cnunciato, line 103:
```suggestion
    const ecsRoleArn = marketingAppStack.requireOutput("ecsRoleArn");
```

cnunciato (approved):
LGTM, based on the previous PR (the issue was a missing output from the marketing-portal stack). Approving to unblock, but suggest using `requireOutput` since that output is indeed required.

infrastructure/index.ts
zchase: 👍

PR #10729, opened 2024-01-31
Update latest-version to the previous release

justinvp:
There's a regression related to the gocloud.dev upgrade. To reduce the impact, downgrade latest-version to the previous version until we get a patch release out with the revert. See https://github.com/pulumi/pulumi/issues/15329


data/versions.json
```diff
@@ -9,8 +9,7 @@
       "darwin-arm64": "https://get.pulumi.com/releases/sdk/pulumi-v3.104.0-darwin-arm64.tar.gz",
       "windows-x64": "https://get.pulumi.com/releases/sdk/pulumi-v3.104.0-windows-x64.zip"
     },
-    "checksums": "https://get.pulumi.com/releases/sdk/pulumi-3.104.0-checksums.txt",
-    "latest": true
+    "checksums": "https://get.pulumi.com/releases/sdk/pulumi-3.104.0-checksums.txt"
```
justinvp, line 12: I debated removing this whole version from this file, but not sure of the downstream impact. So just moved `"latest"` back to the previous version.

Frassle (approved):

PR #10839, opened 2024-02-18
fix: Automatically update Python docs

### LINKED ISSUE #10594: Python SDK docs are not automatically being updated when new versions are released
Python SDK docs are not being regenerated as part of the workflow that runs for a new pulumi/pulumi release.

justinvp:
When a new version of the CLI and Node.js and Python SDKs are released, we kick off a workflow to regenerate docs for these. However, the regenerated Python docs were not being included in the PR that is opened with the changes. This fixes that, by including the updated Python docs in the commit for the PR.

The Python docs are generated under `static-prebuilt/`: https://github.com/pulumi/docs/actions/runs/7925776275/job/21639491629#step:21:117

Fixes #10594


cnunciato (approved):
Thanks @justinvp!

PR #11068, opened 2024-03-21
Simplify lambda@edge config, remove for /ai

AaronFriel:
Root causing some 504 errors seen on Pulumi AI and examining individual requests, the Docs CDN reported an `OriginCommError` and the AI CDN reports a `ClientCommError`, indicating the error originates between them.

Between the two CDNs, the Docs CDN configures two lambda@edge configurations for all cache behaviors and defaults to a 30 second read timeout on origin requests. Given the errors reported, one or both could be responsible for the 504s, though this is inferential.

This PR does two things to enable us to remove the lambda@edge configurations:
* Moves response security headers from applied via origin response lambda to a ResponseHeadersPolicy, which applies the `x-frame-options: deny` header on responses. This is applied to the base cache behavior and inherited.
* Overrides the base cache behavior for `/ai` routes to skip the redirect lambda@edge. The redirect lambda remains applied to the base cache behavior.

The timeouts for the `/ai` origin are also increased to permit longer delays in serving requests.

### DIFF
--- a/infrastructure/cloudfrontLambdaAssociations.ts
+++ b/infrastructure/cloudfrontLambdaAssociations.ts
@@ -12,12 +12,10 @@ import { LambdaEdge } from "./lambdaEdge";
 
 /**
  * Returns the Lambda function associations for the CloudFront distribution.
- * @param addSecurityHeaders If true, creates Lambda@Edge that sets security headers on
- * origin responses and adds the Lambda to the list of associated triggers.
  * @param doEdgeRedirects If true, creates a Lambda@Edge function that that conditionally
 *  redirects based on the URL of the request.
  */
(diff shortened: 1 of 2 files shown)


kmosher (approved):

cnunciato (approved):

cnunciato:
Link to associated preview: https://app.pulumi.com/pulumi/www.pulumi.com/www-production/previews/adae53d1-fca1-47e2-975f-28d52e17a29d

cnunciato:
@AaronFriel @sean1588 I'm still in favor of this change as it removes an edge Lambda we no longer seem to need. Feel free to merge if you agree. Thanks!

cnunciato:
Went ahead and merged this as it worked as expected in the test environment. Will keep an eye as it rolls out.

PR #11178, opened 2024-04-03
Support returning HTTP 410 for unpublished AI Answers pages

cnunciato:
Given how long it's taking for Google to de-list our AI Answers pages, I took a closer look and saw that the 404s being returned for these pages didn't contain the `noindex` meta tags I was expecting. (The AI app actually does return this tag with its Next.js-generated 404 page, but the page we use on the website appears to be masking it.) This means that while these pages do return 404s as expected, their source doesn't contain that extra instruction that explicitly tells Google not to index them. 

To help expedite this, this change adds a CloudFront Lambda function that converts the HTTP response statuses for `/ai/answers/` pages that 404 to HTTP 410 (a much stronger signal to de-list, and one that Next.js doesn't support directly), and by doing so allows CloudFront to deliver the Next.js app's built-in 404 page -- which actually does contain the `noindex` meta tag. I'm hoping these two extra bits of info will get the job done faster than 404s alone seem to be doing.

Also upgrades the infra program use the latest AWS Classic. 

You can see this running on my dev stack at https://www-christian.pulumi-dev.io. 

An example of an unpublished AI Answer is here: https://www-christian.pulumi-dev.io/ai/answers/k3tZ42PRdjZLfs19i6riDe/deploying-helm-charts-on-kubernetes-a-how-to-guide

```
$ curl https://www-christian.pulumi-dev.io/ai/answers/k3tZ42PRdjZLfs19i6riDe/deploying-helm-charts-on-kubernetes-a-how-to-guide
HTTP/2 410 
content-type: text/html; charset=utf-8
server: CloudFront
date: Wed, 03 Apr 2024 22:57:57 GMT
x-powered-by: Next.js

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -4,7 +4,7 @@ import * as aws from "@pulumi/aws";
 import * as pulumi from "@pulumi/pulumi";
 import * as fs from "fs";
 
-import { getLambdaFunctionAssociations } from "./cloudfrontLambdaAssociations";
+import { getAIAnswersRewriteAssociation, getEdgeRedirectAssociation } from "./cloudfrontLambdaAssociations";
 
 const stackConfig = new pulumi.Config();
 
@@ -30,6 +30,10 @@ const config = {
     // doEdgeRedirects is whether to perform redirects at the CloudFront layer.
     // Setting this true will add a Lambda@Edge function that handles that redirect logic.
     doEdgeRedirects: stackConfig.getBoolean("doEdgeRedirects") || false,
+    // doAIAnswersRewrites indicates whether to rewrite HTTP status for AI Answers pages that 404
+    // (e.g., due to unpublishing) as HTTP 410 (Gone). Setting this true will add a Lambda@Edge
+    // function that handles this logic.
+    doAIAnswersRewrites: stackConfig.getBoolean("doAIAnswersRewrites") || false,
     // makeFallbackBucket toggles whether to configure a bucket for serving the website
     // directly out of S3 --- for example, when CloudFront is unavailable. In order to
     // comply with AWS configuration rules, the bucket will be named to match the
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: infrastructure/index.ts [hunk 1]
```diff
@@ -30,6 +30,9 @@ const config = {
     // doEdgeRedirects is whether to perform redirects at the CloudFront layer.
     // Setting this true will add a Lambda@Edge function that handles that redirect logic.
     doEdgeRedirects: stackConfig.getBoolean("doEdgeRedirects") || false,
+    // doEdgeRedirects is whether to perform redirects at the CloudFront layer.
```
cnunciato, line 33: Oops, 🍝 . I'll fix.

infrastructure/index.ts
```diff
@@ -583,7 +577,7 @@ const distributionArgs: aws.cloudfront.DistributionArgs = {
             pathPattern: '/ai/*',
             originRequestPolicyId: allViewerExceptHostHeaderId,
             cachePolicyId: cachingDisabledId,
-            lambdaFunctionAssociations: [],
+            lambdaFunctionAssociations: config.doAIAnswersRewrites ? [getAIAnswersRewriteAssociation()] : [],
```
AaronFriel, line 580: My main concern is whether there will be any latency/streaming impact - if this is deployed to staging first can we verify it there?

AaronFriel (approved):

infrastructure/index.ts
cnunciato: Makes sense yeah. It's currently deployed to my dev stack here if you want to have a look: 

https://www-christian.pulumi-dev.io/ai

sean1588 (approved):
this looks good to me! also thanks for upgrading the aws version, we've needed to do this for some time!

PR #11188, opened 2024-04-05
Add SDK docs to Algolia search

cnunciato:
Adds four records to Algolia for the Node, Python, Go, and .NET SDK docs.

image


sean1588 (approved):
this looks good to me and think the hard coding is good for now!!

PR #11846, opened 2024-05-16
Remove the go.mod.txt renaming stuff

cnunciato:
We had to rename these files from `go.mod` to `go.mod.txt` to work around a quirk of Go modules that prohibits importing directories containing other `go.mod` files. Now that we're no longer importing the pulumi-hugo (via Hugo) as a go module, though, we no longer have to do this! :tada:

Also adds the two workflows we use for upgrading Go modules and running the docs examples tests.


go.mod
cnunciato, line 1: This shouldn't be here -- I'll remove.

sean1588 (approved):

shughes26 (approved):

PR #11847, opened 2024-05-16
Clean up a few more references to pulumi-hugo

cnunciato:
Fixes some links to point to the right places in the docs repo and removes a few references to `pulumi-hugo` et. al. that are no longer needed.

Also fixes up the README's About section to better reflect the current reality. 😄

### DIFF
--- a/config/_default/config.yml
+++ b/config/_default/config.yml
@@ -6,7 +6,6 @@ security:
   funcs:
     getenv:
       - ASSET_BUNDLE_ID
-      - REPO_THEME_PATH
       - PULUMI_CONVERT_URL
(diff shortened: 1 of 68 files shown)


config/_default/config.yml
```diff
@@ -6,7 +6,6 @@ security:
   funcs:
     getenv:
       - ASSET_BUNDLE_ID
-      - REPO_THEME_PATH
```
cnunciato, line 9: We no longer need this, so removing it (and a few other references to it).

sean1588 (commented):
thank you @cnunciato !!

sean1588 (approved):

PR #11849, opened 2024-05-17
Separate pull request flow

sean1588:
Separates pull request flow from the build and deploy workflow and stand up a separate cleanup job for the testing environment, since that is where previews will live.

This separation will enable us to turn add the sync job to `docs-private`. It will only run the PR workflows and we can jsut disable the build-and-deploy workflow there so there will be no deploys from that repo, only PR builds. That repo does not have access to the prod environment, so can't really mess up anything there.

### DIFF
--- a/.github/workflows/bucket-cleanup-testing.yml
+++ b/.github/workflows/bucket-cleanup-testing.yml
@@ -22,10 +22,6 @@ jobs:
         with:
           node-version: '18.x'
 
-      - uses: actions/setup-go@v2
-        with:
-          go-version: 1.19.x
-
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v2
         with:
--- a/.github/workflows/bucket-cleanup-testing.yml
+++ b/.github/workflows/bucket-cleanup-testing.yml
@@ -1,9 +1,10 @@
-name: "Scheduled jobs: Bucket cleanup"
+name: "Scheduled jobs: Bucket cleanup testing"
 on:
   schedule:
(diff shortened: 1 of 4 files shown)


.github/workflows/bucket-cleanup-testing.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '18.x'
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.19.x
```
cnunciato, line 27: We should bump our Gos to 1.21 to catch up to current.

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,83 @@
...
+
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '18.x'
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.20.x
```
cnunciato, line 26: Similarly here, 1.21.

cnunciato (approved):
LGTM aside from the suggestion to bring the Go versions current. (Although do we actually need these now that we're no longer using Hugo modules? We may not. 🤔)

sean1588:
> LGTM aside from the suggestion to bring the Go versions current. (Although do we actually need these now that we're no longer using Hugo modules? We may not. 🤔)

good point! I think you may be right. I'll test it out and see.

PR #11851, opened 2024-05-17
Add new transforms docs

justinvp:
Adds docs for the new transforms APIs. Ported from https://github.com/pulumi/pulumi-hugo/pull/4044.


content/docs/concepts/options/transformations.md
```diff
@@ -17,6 +17,12 @@ The `transformations` resource option provides a list of transformations to appl
```
EronWright, line 17: I would suggest clarifying that the behavior of `transformations` on some component resources is that the children are simply not processed, contrary to the above paragraph. Is there a tip we could provide as to how to identify whether a given component transforms their children when using `transformations`?

content/docs/concepts/options/transforms.md
```diff
@@ -0,0 +1,327 @@
...
+
+You might ask why `Float64Output` instead of at least `IntOutput`. This is because the wire protocol doesn't
+actually have support for integers (being JSON based), so on receiving the untyped properties the transform
+callback can only go on their JSON values and all numbers are 64-bit floats.
+
+```go
+func(_ context.Context, rta *pulumi.XResourceTransformArgs) *pulumi.XResourceTransformResult {
+    if rta.Type == "testprovider:index:Random" {
+        length := rta.Props["length"].(pulumi.Float64Output)
+        rta.Props["length"] = length.ApplyT(func(v float64) float64 { return v * 2 })
```
EronWright, line 308: I think it would be great to expand on how to transform the props, like explain that the values are encapsulated as Outputs, that you should expect variations related to secretness etc, and that you transform the outputs as explained elsewhere.

tgummerer (commented):
A few things I noticed while experimenting with transforms based of these docs.

content/docs/concepts/options/transforms.md
EronWright: A similar example in Python:
```py
def apply_patchforce_annotation(args: ResourceTransformArgs):
    if not args.type_ == "kubernetes:helm.sh/v4:Chart":
        if not 'metadata' in args.props:
            args.props['metadata'] = {}
        if not 'annotations' in args.props['metadata']:
            args.props['metadata']['annotations'] = {}
        args.props['metadata']['annotations']['pulumi.com/patchForce'] = 'true'

    return ResourceTransformResult(
        props=args.props,
        opts=args.opts)

ingresscontroller = kubernetes.helm.v4.Chart(
    "ingresscontroller",
    chart="nginx-ingress",
    namespace=ingress_ns.metadata.name,
    repository_opts=kubernetes.helm.v3.RepositoryOptsArgs(
        repo="https://helm.nginx.com/stable",
    ),
    version="0.14.1",
    opts=pulumi.ResourceOptions(transforms=[apply_patchforce_annotation])
)
```

Frassle (approved):

content/docs/concepts/options/transforms.md
justinvp: > I think it would be great to expand on how to transform the props, like explain that the values are encapsulated as Outputs, that you should expect variations related to secretness (should you?), and that you transform the outputs as explained elsewhere.

Agree! It's somewhat shown in some of the examples, but I'd like to expand on this as a fast follow.

> Would be awesome to show an example like this for each SDK.

Yes. Haven't had a chance to port to each language, but would love to have more examples to show.

PR #11864, opened 2024-05-19
Add frequent/copyable commands to the README

cnunciato:
Just felt like this was missing, so figured I'd add it.

### DIFF
--- a/scripts/serve.sh
+++ b/scripts/serve.sh
@@ -19,4 +19,4 @@ export REL_JS_BUNDLE="/js/bundle.min.${ASSET_BUNDLE_ID}.js"
 # URL to the Pulumi conversion service.
 export PULUMI_CONVERT_URL="${PULUMI_CONVERT_URL:-$(pulumi stack output --stack pulumi/tf2pulumi-service/production url)}"
 
-hugo server --buildDrafts --buildFuture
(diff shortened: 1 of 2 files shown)


scripts/serve.sh
```diff
@@ -19,4 +19,4 @@ export REL_JS_BUNDLE="/js/bundle.min.${ASSET_BUNDLE_ID}.js"
 # URL to the Pulumi conversion service.
 export PULUMI_CONVERT_URL="${PULUMI_CONVERT_URL:-$(pulumi stack output --stack pulumi/tf2pulumi-service/production url)}"
 
-hugo server --buildDrafts --buildFuture
\ No newline at end of file
+HUGO_BASEURL=http://localhost:1313 hugo server --buildDrafts --buildFuture
```
cnunciato, line 22: This extra bit just makes it a little more convenient when you start the local dev server. (It makes the `localhost` URL clickable in most terminals.)

Before (not clickable):

```
Web Server is available at //localhost:1313/ (bind address 127.0.0.1)
``` 

After (clickable):

```
Web Server is available at http://localhost:1313/ (bind address 127.0.0.1)
```

interurban (approved):

PR #11872, opened 2024-05-20
Add pre-commit linting, version checking, and upgrade to Hugo latest (0.126.0)

### LINKED ISSUE #11855: update `hugo` cli to `0.126.x`
With the currently latest version of `hugo` cli, I get following error messages:

```bash
make serve
./scripts/serve.sh
error: [401] SAML SSO login is required.
Hugo provides its own webserver which builds and serves the site.
While hugo server is high performance, it is a webserver with limited options.

The `hugo server` command will by default write and serve files from disk, but
you can render to memory by using the `--renderToMemory` flag. This can be
faster in some cases, but it will consume more memory.

By default hugo will also watch your files for any changes you make and
automatically rebuild the site. It will then live reload any open browser pages
and push the latest content to them. As most Hugo sites are built in a fraction
of a second, you will be able to save and see your changes nearly instantly.

Usage:
  hugo server [command] [flags]
  hugo server [command]

Aliases:
  server, serve

Available Commands:
  trust       Install the local CA in the system trust store.

Flags:
      --appendPort               append port to baseURL (default true)
  -b, --baseURL string           hostname (and path) to the root, e.g. https://spf13.com/
      --bind string              interface to which the server will bind (default "127.0.0.1")
  -D, --buildDrafts              include content marked as draft
  -E, --buildExpired             include expired content
  -F, --buildFuture              include content with publishdate in the future
      --cacheDir string          filesystem path to cache directory
      --cleanDestinationDir      remove files from destination not found in static directories
  -c, --contentDir string        filesystem path to content directory
      --disableBrowserError      do not show build errors in the browser
      --disableFastRender        enables full re-renders on changes
      --disableKinds strings     disable different kind of pages (home, RSS etc.)
      --disableLiveReload        watch without enabling live browser reload on rebuild
      --enableGitInfo            add Git revision, date, author, and CODEOWNERS info to the pages
      --forceSyncStatic          copy all files when static is changed.
      --gc                       enable to run some cleanup tasks (remove unused cache files) after the build
  -h, --help                     help for server
      --ignoreCache              ignores the cache directory
  -l, --layoutDir string         filesystem path to layout directory
      --liveReloadPort int       port for live reloading (i.e. 443 in HTTPS proxy situations) (default -1)
      --minify                   minify any supported output format (HTML, XML etc.)
      --navigateToChanged        navigate to changed content file on live browser reload
      --noBuildLock              don't create .hugo_build.lock file
      --noChmod                  don't sync permission mode of files
      --noHTTPCache              prevent HTTP caching
      --noTimes                  don't sync modification time of files
      --panicOnWarning           panic on first WARNING log
      --poll string              set this to a poll interval, e.g --poll 700ms, to use a poll based approach to watch for file system changes
  -p, --port int                 port on which the server will listen (default 1313)
      --pprof                    enable the pprof server (port 8080)
      --printI18nWarnings        print missing translations
      --printMemoryUsage         print memory usage to screen at intervals
      --printPathWarnings        print warnings on duplicate target paths etc.
      --printUnusedTemplates     print warnings on unused templates.
      --renderSegments strings   named segments to render (configured in the segments config)
      --renderStaticToDisk       serve static files from disk and dynamic files from memory
      --templateMetrics          display metrics about template executions
      --templateMetricsHints     calculate some improvement hints when combined with --templateMetrics
  -t, --theme strings            themes to use (located in /themes/THEMENAME/)
      --tlsAuto                  generate and use locally-trusted certificates.
      --tlsCertFile string       path to TLS certificate file
      --tlsKeyFile string        path to TLS key file
      --trace file               write trace to file (not useful in general)
  -w, --watch                    watch filesystem for changes and recreate as needed (default true)

Global Flags:
      --clock string               set the clock used by Hugo, e.g. --clock 2021-11-06T22:30:00.00+09:00
      --config string              config file (default is hugo.yaml|json|toml)
      --configDir string           config dir (default "config")
      --debug                      debug output
  -d, --destination string         filesystem path to write files to
  -e, --environment string         build environment
      --ignoreVendorPaths string   ignores any _vendor for module paths matching the given Glob pattern
      --logLevel string            log level (debug|info|warn|error)
      --quiet                      build in quiet mode
      --renderToMemory             render to memory (mostly useful when running the server)
  -s, --source string              filesystem path to read files relative from
      --themesDir string           filesystem path to themes directory
  -v, --verbose                    verbose output

Use "hugo server [command] --help" for more information about a command.

Error: command error: unknown flag: --renderToDisk
make: *** [serve] Error 1
```

The flag is called in my hugo -> renderStaticToDisk

```bash
Error: error building site: render: failed to render pages: render of "home" failed: "/Users/dirien/Tools/repos/docs/layouts/index.json:12:32": execute of template failed: template: index.json:12:32: executing "index.json" – File is nil; wrap it in if or with: {{ with .File }}{{ .UniqueID }}{{ end }}
make: *** [serve] Error 1
```

cnunciato:
Adds the pre-commit hook (to run the linter) that we were using in pulumi-hugo, and also adds a bit of scripting to ensure required tools and versions are installed, warning on differences. Also upgrades to the latest version of Hugo (0.126.0).

Fixes #11855.

### DIFF
--- a/layouts/index.json
+++ b/layouts/index.json
@@ -6,17 +6,19 @@
         {{ $ancestors = $ancestors | append $a.LinkTitle }}
     {{- end -}}
 
-    {{-
-        $index = $index | append (
-            dict
-                "objectID" $page.File.UniqueID
-                "title" $page.Title
-                "section" $page.Section
-                "href" $page.RelPermalink
-                "params" $page.Params
-                "kind" $page.Kind
-                "ancestors" (after 1 $ancestors)
-        )
-    -}}
+    {{- if ne $page.File nil -}}
+        {{-
+            $index = $index | append (
+                dict
+                    "objectID" $page.File.UniqueID
+                    "title" $page.Title
+                    "section" $page.Section
+                    "href" $page.RelPermalink
+                    "params" $page.Params
+                    "kind" $page.Kind
+                    "ancestors" (after 1 $ancestors)
+            )
+        -}}
+    {{- end -}}
 {{- end -}}
 {{- $index | jsonify (dict "indent" "    ") -}}
--- a/layouts/page/about.html
+++ b/layouts/page/about.html
@@ -57,7 +57,7 @@
         <h3 class="font-medium text-6xl">Our purpose</h3>
         <h4 class="font-medium text-4-xl">To democratize the cloud for every engineer.</h4>
         <p class="text-gray-700 text-xl mb-10">We help engineers ship infrastructure faster with Infrastructure as Code in general-purpose languages.</p>
-        <a href="{{ relref . " /leadership" }}" class="btn-primary">Meet our leaders</a>
+        <a href="{{ relref . "/leadership" }}" class="btn-primary">Meet our leaders</a>
     </section>
 
     <section id="what-we-believe" class="text-center text-lg py-8 relative">
@@ -98,7 +98,7 @@
             </div>
         </div>
         <div class="careers-cta py-6 w-full mt-28 bg-white">
-            <h4>Like what we stand for? <a href="{{ relref . " /careers" }}" class="font-bold text-blue-600">Join us.</a></h4>
+            <h4>Like what we stand for? <a href="{{ relref . "/careers" }}" class="font-bold text-blue-600">Join us.</a></h4>
             <p>We are looking for new teammates who want to build the future of cloud infrastructure together.</p>
         </div>
     </section>
@@ -119,7 +119,7 @@
                 <img src="/images/mascot/victory-medal-pulumipus.svg" alt="Pulumipus wearing a medal with two hands raised" />
                 <div class="header mt-6 mb-1.5 text-medium text-2xl">The Puluminaries</div>
                 <p class="text-left flex-grow">A group for stellar Pulumi community members who have made "above and beyond" contributions to the community.</p>
-                <a href="{{ relref . " /community/puluminaries" }}" class="btn-primary">Meet our Puluminaries</a>
+                <a href="{{ relref . "/community/puluminaries" }}" class="btn-primary">Meet our Puluminaries</a>
             </div>
             <div class="community-spotlight mx-auto flex flex-col justify-cetner items-center w-72 md:w-1/3 my-6 px-3">
                 <img src="/images/mascot/rocketpus.svg" alt="Pulumipus wearing a medal with two hands raised" />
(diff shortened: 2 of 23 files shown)


scripts/ensure.sh
```diff
@@ -1,14 +1,29 @@
...
+    fi
+
+    if [ ! $(echo ${version_string} | grep ${required_version}) ]; then
+        printf "It looks like you're running %s %s, but this project uses version %s.\n" ${tool_name} ${version_string} ${required_version}
+        echo $details
+        exit 1
+    fi
+}
+check_version "Node.js"  "node"  "node -v | sed 's/v\([0-9\.]*\).*$/\1/'"            "18"
+check_version "Hugo"     "hugo"  "hugo version | sed 's/hugo v\([0-9\.]*\).*$/\1/'"  "0.111"
```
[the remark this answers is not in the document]
cnunciato: Yeah, agreed. I felt the same so went and looked into understanding what it'd take to address #11855, and it's not too big a deal, looks like. So I'll add that here and turn these into warnings instead.

layouts/index.json
```diff
@@ -6,17 +6,19 @@
...
-                "objectID" $page.File.UniqueID
-                "title" $page.Title
-                "section" $page.Section
-                "href" $page.RelPermalink
-                "params" $page.Params
-                "kind" $page.Kind
-                "ancestors" (after 1 $ancestors)
-        )
-    -}}
+    {{- if ne $page.File nil -}}
```
cnunciato, line 9: Turns out the only pages that have nil `File` references are the tag and author pages, whose `.UniqueID` values already resolve to `""` (so are currently being filtered out anyway). So keeping only on those pages that contain `File` references looks like it's fine after all.

scripts/ensure.sh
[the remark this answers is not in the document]
sean1588: Oh nice, yeah if it's not too much trouble to do the upgrade that would be awesome!!

layouts/page/about.html
```diff
@@ -57,7 +57,7 @@ <h2 class="text-4xl md:text-6xl bg-white px-6 mb-0 pb-6">We are Pulumi</h2>
         <h3 class="font-medium text-6xl">Our purpose</h3>
         <h4 class="font-medium text-4-xl">To democratize the cloud for every engineer.</h4>
         <p class="text-gray-700 text-xl mb-10">We help engineers ship infrastructure faster with Infrastructure as Code in general-purpose languages.</p>
-        <a href="{{ relref . " /leadership" }}" class="btn-primary">Meet our leaders</a>
+        <a href="{{ relref . "/leadership" }}" class="btn-primary">Meet our leaders</a>
```
cnunciato, line 60: Several of these were warning on `make serve`, so I went ahead and fixed them.

scripts/ensure.sh
```diff
@@ -1,14 +1,34 @@
...
+    fi
+
+    if [ ! $(echo ${version_string} | grep ${required_version}) ]; then
+        printf "${color_yellow}warning${color_end} It looks like you're running %s %s, but this project uses version %s.\n" ${tool_name} ${version_string} ${required_version}
+        echo $details
+    fi
+}
+check_version "Node.js"  "node"  "node -v | sed 's/v\([0-9\.]*\).*$/\1/'"            "18"
+check_version "Hugo"     "hugo"  "hugo version | sed 's/hugo v\([0-9\.]*\).*$/\1/'"  "0.126.0"
+check_version "Yarn"     "yarn"  "yarn -v | sed 's/v\([0-9\.]*\).*$/\1/'"            "1.22"
```
cnunciato, line 31: It'd be awesome if we could specify these versions in one place and use them everywhere. Don't yet have any great ideas for how to do this in a way that's work both here and in our GHA workflows, though.

sean1588: for `hugo` not sure, but I wouldn't be surprised if there is a way to enforce the node and yarn versions in the `package.json` file somehow.

sean1588 (approved):
thanks @cnunciato!! looks good to me. I also verified the search-index.json file produced in the preview is identical to what is in prod now, so those updates to the index.json file work and didn't regress anything! :rocket:

PR #11875, opened 2024-05-20
Add dark buttons, make light and dark the same size (matching get.pulumi.com)

cnunciato:
All buttons are now the same size (215x46) and match the dimensions of the default button hosted at get.pulumi.com:


sean1588 (approved):
:rocket:

PR #11881, opened 2024-05-21
Update blog template

sean1588:
Updates the blog template to include properties to house the social copies to be posted on twitter and linkedin. I also updated the `disabledKinds` config to update it to `taxonomyTerm` from `taxonomy`. This term was updated to this after hugo 0.73 from my understanding and also matches the config we were running in pulumi-hugo. I also added the `category` kind as well, since pulumi-hugo had this listed. It was giving me an error before I updated this when running the `hugo new` command to generate the blog files. 

We _could_ add linting here to enforce this, but it would fail unless we add these props to all the blog files. It also will presumably just pass for new blog posts going forward as these would already be filled out when generated from the template. Maybe we can do something like only running linting against newly added files in the commit, but I think this should be good enough for now.

### DIFF
--- a/archetypes/blog-post/index.md
+++ b/archetypes/blog-post/index.md
@@ -34,6 +34,13 @@ authors:
 tags:
     - change-me
 
+
+# The social copy used to promote this post on Twitter and Linkedin. These
+# properties do not actually create the post and have no effect on the
+# generated blog page. They are here strictly for reference.
+social_twitter: "social copy to post on Twitter"
+social_linkedin: "social copy to post on Linkedin"
+
 # See the blogging docs at https://github.com/pulumi/docs/blob/master/BLOGGING.md
 # for details, and please remove these comments before submitting for review.
 ---
--- a/archetypes/blog-post/index.md
+++ b/archetypes/blog-post/index.md
@@ -38,8 +38,18 @@ tags:
 # The social copy used to promote this post on Twitter and Linkedin. These
 # properties do not actually create the post and have no effect on the
 # generated blog page. They are here strictly for reference.
-social_twitter: "social copy to post on Twitter"
-social_linkedin: "social copy to post on Linkedin"
+
+# Here are some examples of posts we have made in the past for inspiration:
+# https://www.linkedin.com/feed/update/urn:li:activity:7171191945841561601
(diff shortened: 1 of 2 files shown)


cnunciato (approved):

archetypes/blog-post/index.md
```diff
@@ -34,6 +34,13 @@ authors:
 tags:
     - change-me
 
+
+# The social copy used to promote this post on Twitter and Linkedin. These
+# properties do not actually create the post and have no effect on the
+# generated blog page. They are here strictly for reference.
+social_twitter: "social copy to post on Twitter"
+social_linkedin: "social copy to post on Linkedin"
```
cnunciato, line 42: You could also do this:

```suggestion
social:
    twitter: 
    linkedin: 
```

... And then leave these blank, as we do with the meta_desc.

cnunciato:
> We could add linting here to enforce this, but it would fail unless we add these props to all the blog files.

This is true, but we could also fix this by backfilling with placeholder data, and lint going forward. I just worry that by not linting, we'll essentially in the same position we're in today, having to nag people to do it after the fact.

archetypes/blog-post/index.md
```diff
@@ -34,6 +34,13 @@ authors:
 tags:
     - change-me
 
+
+# The social copy used to promote this post on Twitter and Linkedin. These
+# properties do not actually create the post and have no effect on the
+# generated blog page. They are here strictly for reference.
```
cnunciato, line 40: People commonly ask for "good" examples of both of these (as they're slightly different) so they have something to model their own copy by. It might be good to provide links to those examples here. Here are the ones that've been shared with me:

- https://www.linkedin.com/feed/update/urn:li:activity:7171191945841561601
- https://www.linkedin.com/feed/update/urn:li:activity:7169021002394296320
- https://www.linkedin.com/feed/update/urn:li:activity:7155606616455737345
- https://twitter.com/PulumiCorp/status/1763265391042654623
- https://twitter.com/PulumiCorp/status/1762900472489185492
- https://twitter.com/PulumiCorp/status/1755637618631405655

sean1588:
thanks @cnunciato! I made updates to the structure and comments. I'll follow on with adding linting around this and also backfill the existing posts.

Issue #11688, opened 2022-04-12
Building with Pulumi tutorials are not mobile friendly


AaronFriel:
The tutorials located here:

* https://www.pulumi.com/learn/building-with-pulumi/understanding-stacks/
* https://www.pulumi.com/learn/building-with-pulumi/stack-outputs/
* ...

Render in a widescreen layout on mobile Firefox and Chrome for the main content. The banner for Business Critical Edition and the top bar, as well as the navigation hamburger menu appear fine. 

The sidebar listing the tutorials and main content however are wider than the viewport.

![Screenshot_20220412-143742.png](https://user-images.githubusercontent.com/788800/163059231-86082cbc-6819-4072-81b1-e6e8400db445.png)

Issue #10619, opened 2024-01-18
Learn Pathways code example broken for working with secrets


interurban:
My recent Pulumi Cloud sign-up was to help get me more hands-on with Pulumi in preparation for my “DevOps: Infrastructure as Code” course. In that context, I have a couple of observations for improvement to share.
When following the “Learn Pulumi” Pathways …
There are a few minor typographical inconsistencies (like “my_first_app” vs. “my-first-app”) that cause unnecessary learning friction and deflate the positive momentum of these beginner pathways.
There is at least one line of code that is entirely nonsensical. In the “Working with Secrets” section, there is an instruction to add the following Typescript line to the end of the program:
export const mongoPassword = mongoPassword;
Assigning a variable to itself makes no sense, and a variable cannot be redeclared in the same scope. The export directive should instead be placed where the variable is first declared.

Issue #10226, opened 2023-11-05
Improvement for https://www.pulumi.com/learn/building-with-pulumi/stack-references/


eguerlain:
When creating reference to the stack of `my_first_app`, the `stackRef` variable is built with a kebab-cased project name (`my-first-app`):

```ts
const stackRef = new pulumi.StackReference(`${org}/my-first-app/${stack}`)
```

However, this project was created with a snake_cased name in [Creating a Pulumi Project](https://www.pulumi.com/learn/pulumi-fundamentals/create-a-pulumi-project/). (`mkdir my_first_app`)

This leads to the following error when running `pulumi up` for the `my-second-app-staging` stack:

```bash
     Type                             Name                            Plan       Info
 +   pulumi:pulumi:Stack              my-second-app-staging           create     1 error
     └─ pulumi:pulumi:StackReference  etigrln95/my-first-app/staging             1 error

Diagnostics:
  pulumi:pulumi:Stack (my-second-app-staging):
    error: preview failed

  pulumi:pulumi:StackReference (etigrln95/my-first-app/staging):
    error: Preview failed: unknown stack "etigrln95/my-first-app/staging"
```

Using snake_cased project name works correctly:

```ts
const stackRef = new pulumi.StackReference(`${org}/my_first_app/${stack}`)
```

I suggest using kebab-cased name when creating the `my-first-app` directory in [the first tutorial](https://www.pulumi.com/learn/pulumi-fundamentals/create-a-pulumi-project/) to keep a consistant naming.

toriancrane:
Thank you @eguerlain for submitting this issue! We will be taking a look into this. We do have an updated guide available for [Stack Outputs and References](https://www.pulumi.com/docs/using-pulumi/stack-outputs-and-references/) if you'd like to check that out in the meantime.

Issue #10225, opened 2023-11-05
Issue with https://www.pulumi.com/learn/building-with-pulumi/stack-outputs/


eguerlain:
The very last instruction to run on [the "Understanding Stack Outputs" page](https://www.pulumi.com/learn/building-with-pulumi/stack-outputs/) is `pulumi up`, in order to see the difference between the `dev` and the `staging` stacks.

> Now, if you run `pulumi up` while in the `staging stack`, we should see that the frontend port is now set to 3002:

However, at this point, the `dev` stack is "up". When running `pulumi up` on the `staging` stack, I get this plan:

```bash
     Type                         Name                  Plan       
 +   pulumi:pulumi:Stack          my_first_app-staging  create     
 +   ├─ docker:index:RemoteImage  mongoImage            create     
 +   ├─ docker:index:RemoteImage  frontendImage         create     
 +   ├─ docker:index:RemoteImage  backendImage          create     
 +   ├─ docker:index:Network      network               create     
 +   ├─ docker:index:Container    mongoContainer        create     
 +   ├─ docker:index:Container    frontendContainer     create     
 +   └─ docker:index:Container    backendContainer      create    
```

Applying it result in an error when creating the `mongoContainer` since the port `27017` is already allocated.

```bash
Diagnostics:
  docker:index:Container (mongoContainer):
    error: 1 error occurred:
    	* creating urn:pulumi:staging::my_first_app::docker:index/container:Container::mongoContainer: 1 error occurred:
    	* Unable to start container: Error response from daemon: driver failed programming external connectivity on endpoint mongo-staging (148d4ec22fc40525a0bbcd2ebd54b27d7f09d6ec678280dbf9ced6c05e1417ff): Bind for 0.0.0.0:27017 failed: port is already allocated

  pulumi:pulumi:Stack (my_first_app-staging):
    error: update failed
```

I am just starting with Pulumi, maybe I have not fully grasped all the concepts.  
However, I think that there is a missing instruction `pulumi stack select dev && pulumi destroy && pulumi stack select staging` before running `pulumi up` to remove resources of the `dev` stack before creating those of the `staging` stack.

This method is a little bit annoying since the docker images are removed to be pulled again in the `pulumi up`, but a least I'm able to "up" the `staging` stack.

toriancrane:
Thank you for submitting this issue @eguerlain ! We will take a look into this.

PR #11883, opened 2024-05-22
Move Learn Pulumi to Pulumi Tutorials

cnunciato:
This change creates a new section called Pulumi Tutorials (🎉) and bootstraps it with content from Learn Pulumi. As part of this work, I also reviewed (and fixed!) all existing content and redirected all content from `/learn` to its new home at `/tutorials`. 

A few things I've not yet done:

* Documented how to create a new tutorial, how tutorials are sorted, grouped, sorted within a group, etc.
* ~~Added a `make` helper/Hugo archetype to simplify creating a new tutorial. We have these for Learn today and they're nice, but they don't map directly to the new stuff, so they'll need some adjusting.~~
* Documented the concept of "collections" as it applies to tutorials. Everything here is a collection, and collections are typed -- so there are "provider" collections (e.g., `aws`), "proficiency" (i.e., skill-level) collections (`pulumi-101`), there'll surely be "product" and "feature" collections soon (ESC, Deployments, Components), etc. Collections are the only grouping mechanism we have, but the `type` field should make this pretty flexible (certainly flexible enough for now), and we can always extend/refactor the content model when/if we outgrow this. Tutorials can also belong to multiple collections, of course, or none, in which case they'll spill into the All Tutorials section on the Tutorials home page.

    Since there was so little content to go live with here, I created a single collection to start, and called it "Pulumi 101" (feel free to suggest an alternative -- I'm mainly looking for something succinct that conveys that these are meant for new users). It contains the first three Learn modules --- Fundamentals, Building, and Encapsulation, since those are all about core concepts (whereas Automation API and Importing are more advanced topics).
* Written up authoring guidelines for tutorials -- what makes a good one, when to use multi-page vs. single-page, a general a structure to follow for both, etc.
* Consolidated/figured out what to do with the "[cloud engineering glossary](https://www.pulumi.com/learn/glossary/)". It does feel like there's a place for a glossary of sorts in this new section, so I'll probably find a way to integrate it somehow, but for now, it's got the same treatment it does today (and is redirected to /tutorials/glossary like everything else).  
* ~~Made all the pre-reqs and intros consistent. I'll be doing that while y'all review what's here.~~

It may also feel like the navigation is a bit "sparse" right now, and that's mainly because we have so little content. As we add more tutorials, collections, and collection types, this will naturally grow and begin got look more like what we've been aiming for.

Fixes https://github.com/pulumi/devrel-team/issues/641
Fixes https://github.com/pulumi/docs/issues/12017
Fixes https://github.com/pulumi/docs/issues/10226
Fixes https://github.com/pulumi/docs/issues/10225
Fixes https://github.com/pulumi/docs/issues/12418
Fixes https://github.com/pulumi/docs/issues/10619
Fixes https://github.com/pulumi/docs/issues/10202
Fixes https://github.com/pulumi/docs/issues/10173
Fixes https://github.com/pulumi/docs/issues/10163
Fixes https://github.com/pulumi/docs/issues/10155
Fixes https://github.com/pulumi/docs/issues/11688
Fixes https://github.com/pulumi/docs/issues/12423
Fixes https://github.com/pulumi/docs/issues/12426


sean1588:
Looks good overall. Couple things I noticed here with regards to how some things are rendering:

**It looks like the left nav is cutting off some of the elements within it.**
Screen Shot 2024-08-06 at 3 58 24 PM
Screen Shot 2024-08-06 at 3 57 45 PM

**I think we could do with a decrease of padding between the boundaries of the left nav and the menu itself. It seems to be less on other pages**
Screen Shot 2024-08-06 at 3 57 54 PM

interurban (approved):
Content looks good, I'll let Sean give the the code/approval.

Makefile
```diff
@@ -116,13 +108,17 @@ test-programs:
 upgrade-programs:
 	./scripts/programs/upgrade.sh
 
-.PHONY: new-learn-module
-new-learn-module:
-	./scripts/content/new-learn-module.sh
+.PHONY: new-tutorial-module
```
sean1588, line 111: good idea!!

sean1588 (approved):
I'm going to go ahead and leave you the approval here, assuming we get the nav sorted. There are a few more things I noticed when checking out the mobile view though. Not sure if all of these are blocking though, maybe we can fast follow and address these things after.

looks like the bullet points are not wrapping the text
Screen Shot 2024-08-07 at 3 13 52 PM

these look like they need some spacing and alignment adjustment
Screen Shot 2024-08-07 at 3 14 19 PM

the menu seems to be aligned all the way to the right, not sure if that was intentional though.
Screen Shot 2024-08-07 at 3 13 09 PM

sean1588:
one more thing, I noticed the search bar is not aligned to the right anymore
Screen Shot 2024-08-07 at 3 24 39 PM

sean1588:
🚀

PR #11884, opened 2024-05-22
Only run automerge if pulumi-bot

sean1588:
We have been getting quite a few failures from the automerge workflow due to github ratelimitting. This is likely happening now that there are more PRs being submitted to docs after the migration. This adds a condition to only run when the author is pulumi-bot, since that is the main reason we have this workflow.


cnunciato (approved):
LGTM.

PR #11900, opened 2024-05-24
initial draft of breakpoint debugging blog post

thoward:
### Proposed changes

Initial draft of blog post on breakpoint debugging. 

### Related issues (optional)

Closes https://github.com/pulumi/devrel-team/issues/977


content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,317 @@
...
+tags:
+    - next-level-iac
+    - debugging
+    - vscode
+    - typescript
+    - nodejs
+    - ide
+---
+
+For many of us, life is suspended between presses of the [`F5`][f5-key] key. Our IDE is the only place where everything is organized and makes sense. And while we know that the likelihood of a code hole-in-one is very rare (that magical moment where you run your build and everything *just works* the first time), we always hope that *this* time, this will be the run where you see the output that lets you know you've gotten it right. And in between? Debugging.
```
aaronkao, line 18: Great intro

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,317 @@
...
+43 packages are looking for funding
+  run `npm fund` for details
+
+found 0 vulnerabilities
+Finished installing dependencies
+
+Your new project is ready to go! ✨
+
+To perform an initial deployment, run `pulumi up`
+```
```
aaronkao, line 82: Do we need this code block? I suspect we just tell them to create a new project uusing the static website -aws-typescript template. we couuld also just provide the commands but not all the shell outputs? @interurban what is the guidance for these kinds of things?

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,317 @@
...
+
+Then launch this project to make sure it works as expected:
+
+```shell
+$ pulumi up
+
+Previewing update (dev)
+
+View in Browser (Ctrl+O): https://app.pulumi.com/troy-pulumi-corp/good-morning/dev/previews/########-####-####-####-############
```
aaronkao, line 91: @interurban same with this do we typcially want to show all this output?

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,317 @@
...
+		}
+	]
+}
+```
+
+This creates a VS Code task named `pulumi-debug` which will run the wrapper script we just created from the root of our project.
+
+Next, create a file named `launch.json` in the same directory. This will configure the behavior of the `F5` key!
+
+```json
```
aaronkao, line 210: @cnunciato can you doublecheck to make sure this is the correct way to debug in node?

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,317 @@
...
+{{% /notes %}}
+
+Finally, we need to make one more edit to the `pulumi-debug.sh` file we created. Here will will pass the necessary options to the Node.js runtime to set it to run in debug mode. 
+
+Add the following line to the beginning of the file:
+
+```shell {hl_lines=[3]}
+#!/bin/bash
+
+export NODE_OPTIONS='--inspect-brk=127.0.0.1:9292'
```
aaronkao, line 254: i believe we need to do a python version of this since all our blog posts should have both typescript and python code

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,317 @@
...
+
+* **Limited integration with VS Code:** Unfortunately our technique of running Pulumi in wrapper script as a backgrounded pre-launch task in VS Code leaves some things to be desired. It is different enough from the use case that VS Code had in mind for these features that making things "just work" is a little difficult. For example, we can't easily report on errors in the "Problems" tab. Sometimes VS Code can't tell that the program finished and so you'll have to manually stop the debugging session after Pulumi has completed.
```
aaronkao, line 289: can you rephrase this to be more clear. i don't really understand what is being said here.

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,317 @@
...
+
+Despite these limitations, Pulumi with VS Code as your IDE provides a rich experience compared to trawlling through unstructured logs, or adding in `console.log(...)` calls to echo out ad-hoc debugging information (aka. [`printf` debugging][printf-debugging]). Also, you can enable this for any language that Pulumi supports. Python, Go, and C# all have popular debuggers and IDEs that can be used in the same manner.
```
aaronkao, line 291: are there other methods to debug Pulumi programs? Are there other IDEs or other systems that you can do this? Probably should mention all the capabilities you have here

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,317 @@
...
+Open the VS Code terminal panel and first set it to be executable, then try running the script: 
+
+```shell
+$ chmod +x pulumi-debug.sh
+$ ./pulumi-debug.sh
+```
+
+Ok now let's setup VS Code to run that in debug mode when we hit `F5`!
+
+### Step 3: Configure VS Code `launch.json` and `task.json`
```
aaronkao, line 182: I think what is missing in all of this is we show "how to set up and use a debuguger", but we don't show a sample of something that is broken and using a debugger to help us fix it. I think the thinking here is like people always code errors into their programs, it will end up executing and failing at provisioning. for something like CloudFormation its a long roll back process. for TF, I don't know what happens. But in Pulumi you can debug the program with a debugger and isolate the error.

aaronkao: Showing a simple error that will normally creep up and how to find it is probably the piece that is missing

aaronkao (changes requested):
Overall looking good. I think its missing a good example of using a debugger to fix a Pulumi program

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,317 @@
...
+
+* **Pulumi programs are only part of the story:** This technique will let you inspect the program you authored, but it doesn't give access to what's happening with the Pulumi backend, the providers, or anything else outside of the program itself. You might find yourself needing to dig deeper to resolve a problem with the stack as a whole. Luckily there are other techniques we can use for that, which we will delve into in an upcoming post.
+
+* **Inputs and Outputs are asynchronous futures:** If you've used Pulumi before you will have encountered [`Input<T>`][input-docs] and [`Output<T>`][output-docs] values, which are a core function of the tool. However, at the time that the debugger hits a breakpoint on those values, they are likely not populated with their final value. That means you won't be able to inspect them in the debugger very well. Instead, you will need to add an [`apply(...)`][apply-docs] or [`all(..)`][all-docs] block to the code, and place the breakpoint on the materialized value within that block. This follows all the same rules as accessing those variables in code in a Pulumi program.
```
cnunciato, line 287: It might be nice to provide a simple pasteable example of this here, since it's the sort of thing you're so often going to want to debug -- e.g., "What is the cloud provider returning here?"

```typescript
cdn.id.apply(id => {
   // Set a breakpoint here.
});
```

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,317 @@
...
+```
+Debugger listening on ws://127.0.0.1:9292/########-####-####-####-############
+For help, see: https://nodejs.org/en/docs/inspector
+Debugger attached.
+Waiting for the debugger to disconnect...
+``` 
+
+### Step 4: Setting a debugging breakpoint 
+
+Now we can try [setting a breakpoint][vs-code-breakpoints] in our code. Navigate to a line within the code and hit the `F9` key. You'll see a small red dot appear next to the line indicating the breakpoint. You can then run the program again using `F5`. When the execution of the code reaches the line it will pause there.
```
cnunciato, line 275: One thing to note is that the debugging happens in a different terminal session/window, so if your Pulumi program relies on environment variables (e.g., your AWS creds), you'll need some way to pass them to that session. For example, I kept hitting this error (even after ensuring my AWS creds were current in the embedded terminal) until I realized the debugger wasn't actually using that terminal:

```plain
Diagnostics:
  pulumi:providers:aws (default_6_37_1):
    error: pulumi:providers:aws resource 'default_6_37_1' has a problem: unable to validate AWS credentials.
    Details: validating provider credentials: retrieving caller identity from STS: operation error STS: GetCallerIdentity, https response error StatusCode: 403, RequestID: 8e37f328-dfa8-4234-8234-5cca32d6ea48, api error ExpiredToken: The security token included in the request is expired
```

There's an `env` property you can set in `launch.json` to help with this -- e.g., if you use a named `AWS_PROFILE` (as you probably wouldn't want to hack your actual `AWS_*` creds into that file). For me, since I use AWS SSO through ESC, adding this line to the shell script before `pulumi up` did the trick:

```bash 
eval "$(pulumi env open pulumi/dev-sandbox --format shell)"
```

content/blog/next-level-iac-breakpoint-debugging/index.md
cnunciato: > Sometimes VS Code can't tell that the program finished and so you'll have to manually stop the debugging session after Pulumi has completed.

This happens 100% of the time for me, yeah. I figured this was the expected behavior, but I guess not?

It does sound like this is worth explaining in more detail, though. Like how is it "different from the use case VS Code had in mind"? Does it have something to do with the fact that the Node process is spawned by Pulumi?

content/blog/next-level-iac-breakpoint-debugging/index.md
cnunciato: We should also add this to our own docs, as our coverage of interactive debugging seems to cover only debugging providers (rather than your own IAC).

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,317 @@
+---
+title: "Next Level IaC: Breakpoint Debugging for Pulumi Programs"
+date: 2024-05-27
+draft: false
+meta_desc: "Next Level IaC: Breakpoint Debugging for Pulumi Programs"
```
cnunciato, line 5: Just a reminder to remember to tweak this before merging.

content/blog/next-level-iac-breakpoint-debugging/index.md
cnunciato: I think it is, yep!

cnunciato (commented):
This is great! A few bits of initial feedback for ya.

content/blog/next-level-iac-breakpoint-debugging/index.md
thoward: In this case, the blog post is already so complex, and the differences between Python and Node/TS here are significant enough that I think it warrants a separate follow up post. Similarly w/ Go and C#...

content/blog/next-level-iac-breakpoint-debugging/index.md
thoward: It's more about the difference between a "launch" and "attach" foreground task. Launch will run a program, attach will wait to attach to a program... So to launch the pulumi CLI we run it as a preLaunchTask (because you can't launch and then attach in the same launch config, so it needs to run it "preLaunch"), but if you don't say it's a background task, it will then have two wait conditions... waiting to attach and waiting for prelaunchtask to complete. that means you have to run as a background task (e.g. indicating that it runs forever like a service and not wait on it). Then, once it's running, the launch config will wait to attach to a debugger. but since that's a background task that can start/stop multiple times, instead of attaching then exiting once the thing it attaches to is complete, it attaches, and then goes into a wait/retry loop to try to reconnect with another background process. if it hits that timeout it will exit on its own. otherwise you have to manually exit the debug mode. Other programs behave in a way that is consistent with this model (like `tsc-watch`), which is what VS Code's implementation is designed for.

Frankly, this is all too complicated for this post which is already fairly long. I wanted to call it out without diving into the details and distracting from the "hero narrative" of Pulumi being debuggable by talking about all the ways the VS Code's debugging implementation makes it not fit perfectly into that model.

Maybe I should just remove this bullet point entirely?

content/blog/next-level-iac-breakpoint-debugging/index.md
thoward: Yeah, you can do this with any sophisticated IDE with an extension for CDP (Chrome DevTools Protocol). That would include most IDEs that are designed to work with Node (e.g. Visual Studio, Webstorm, even neovim). I added a note about that.

content/blog/next-level-iac-breakpoint-debugging/index.md
thoward: Good point! I originally wanted to write it this way but felt the complexity and length of the post would be too much. I went ahead and added that in anyway, in the current revision.

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,317 @@
...
+pulumi up -f
+```
+
+{{% notes type="info" %}}
+Let's walk through these lines to explain what's going on here:
+
+* `pulumi login` to set up credentials for the pulumi backend
+* `pulumi stack select dev` to run this against our `dev` stack
+* `pulumi up -f` to run our program udpate. The -f flag skips the preview phase and also skips the need to interactively confirm that we want to run the update.
+{{% /notes %}}
```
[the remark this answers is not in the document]
thoward: IMO this callout is for "advanced" understanding and can be skipped if you don't care about digging into the details of what this does. Leaving it in sort of railroads the reader into reading this content even if they don't care, so putting in a callout makes it easy for them to skip past it if they want to.

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,317 @@
...
+$ ./pulumi-debug.sh
+```
+
+Ok now let's setup VS Code to run that in debug mode when we hit `F5`!
+
+### Step 3: Configure VS Code `launch.json` and `task.json`
+
+To run our Pulumi program in debug mode we will need to create a couple VS Code configuration files; `launch.json` and `task.json` in a hidden directory called `.vscode` within your project. 
+
+Create the `.vscode` directory and then create a new file called `task.json` file in that directory, with the following contents:
```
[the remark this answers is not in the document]
thoward: Oops! Good catch.

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,317 @@
...
+
+pulumi login
+pulumi stack select dev
+pulumi up -f
+```
+{{% notes type="info" %}}
+The `NODE_OPTIONS` environment variable passes options to the Node.js V8 runtime. In this case, we are passing `--inspect-brk` which tells the Node.js runtime to run in debug mode, which will cause it to pause before running the code, until a debugging client connects to `localhost` on port `9292`.
+{{% /notes %}}
+
+Ok, now we should be ready to run our Pulumi program from VS Code using the `F5` key! Let's give it a try. You should see Pulumi running inside of the "Terminal" tab. You will also see output indicating that it is waiting for a debugger to attach:
```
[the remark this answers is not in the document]
thoward: This is point where VS Code is expecting Pulumi to behave differently than it does. It's kind of complicated to get into in the blog post... but essentially, as i explained in a previous comment, the only way this works is to run Pulumi as a background task, which means that VS Code expects it to keep running and produce a series of "problems" while its running which it parses out with a customizable regex-based "problem matcher" configuration to find errors and report them in the Problems tab. but for whatever reason, the problem matcher can't really be configured to work with the type of output Pulumi creates (despite it seeming like it would be possible... i went down a lengthy rabbit hole trying to get this to "just work" and all i was able to do was get partial success and it required an incredibly complex version of `tasks.json` even to get that much working). We could likely create an output mode for the Pulumi CLI that *would* output things in the way that VS Code expects, or create an extension for VS Code that would fix the need to background it, but as it stands, there's not really a way to do it. So, you get this annoying popup box until you click and tell it not to worry about it and to remember it for that task. If I knew where VS Code stored that "remember" i would include steps to configure that also, but I haven't been able to figure where it's stored yet.

content/blog/next-level-iac-breakpoint-debugging/index.md
thoward: Since this configuration will vary so much by the user's local config (e.g. Windows vs Mac, how they have creds set-up for AWS, etc), I've chosen to just leave this part unexplained rather than to try to figure out a way to explain it in a way that is exhaustive enough to cover all the likely scenarios.

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,317 @@
...
+Create the `.vscode` directory and then create a new file called `task.json` file in that directory, with the following contents:
+
+```json
+{
+	"version": "2.0.0",
+	"tasks": [
+	    {
+			"label": "pulumi-debug",
+			"type": "shell",
+			"command": ". ./pulumi-debug.sh",
```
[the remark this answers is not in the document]
thoward: Not sure. This was part of the example I worked from. I wondered the same, but just decided to leave it as is instead of spending more time trying to figure out if it was important or not (there are already so many moving parts...).

content/blog/next-level-iac-breakpoint-debugging/index.md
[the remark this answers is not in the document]
thoward: Actually, I figured out where this is stored. 
In `$HOME/Library/Application\ Support/Code/User/settings.json`
```json
{
...
    "task.problemMatchers.neverPrompt": {
        "shell": true
    },
...
```

But again, not sure if this level of detail is worth including in the article or if folks can just click the checkbox on the prompt and be happy with it. The problem of course, is that this is a global setting for all tasks that are `shell` tasks, so this would apply to everything, not just our example here. This might be too much to tell users to configure to follow along with our example as it will hide issues with problem matchers in future `shell` tasks.

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,342 @@
...
+
+In the infrastructure world, debugging is a pretty big problem. Some of the most popular tools have nothing more than simple unstructured text output logs. While tools like Terraform call themselves "Infrastructure as Code" the "code" is a proprietary DSL, not a general purpose programming language, and so there is little-to-no IDE support. There is no `F5`. Even worse, there is no `F9` for setting a breakpoint. There's no way to inspect state mid-run, or more importantly to explore in an unstructured way, because this is where the "aha!" moments usually happen.
+
+Pulumi gives you a rich IaC debugging experience by enabling breakpoint debugging in your favorite IDE. Let's walk through an example of how to enable a full-featured debugging workflow in [VS Code][vs-code] for TypeScript.
+
+## How Breakpoint Debugging Works
```
aaronkao, line 28: should this section be put in one of these

`{{% notes type="info" %}}`

And then hidden behind a expander? @cnunciato what do you think?

content/blog/next-level-iac-breakpoint-debugging/index.md
```diff
@@ -0,0 +1,342 @@
...
+</html>
+```
+
+Once that's edited and saved, go ahead let's redeploy the website:
+
+```shell
+$ pulumi up
+```
+
+Everything runs successfully... but if we navigate to the website, we don't see our updated content?! Oh no! What went wrong? Why isn't there an error message?! Time to put on our debugging hats and figure out what is going on. First, let's set up VS Code to enable breakpoint debugging right in the IDE.
```
aaronkao, line 123: Nice way to introduce the problem example

aaronkao (approved):
Overall looks good. Lets ship it

PR #11919, opened 2024-05-29
Change PR runner

sean1588:
Changes to using the "standard" ubuntu runner for PR builds. Now that this flow has been broken off, I don't think the larger ones are necessary for this. It's likely we can also back down the runner size for the deployment job as well, since I believe this was a relic from when we used to build everything here, along with registry and needed more beefy machinery.


sean1588:
Going to go ahead and push this through. No changes to anything here except the runner size and it built fine.

PR #11942, opened 2024-06-03
clarify stack transformation application

tgummerer:
Stack transformations are only applied to resources that are registered after the transformation is registered.  Clarify that in the docs.

This hopefully clarifies the potential confusion that lead to https://github.com/pulumi/pulumi/issues/16304.


Frassle (approved):
We should make sure a similar change is applied to https://github.com/pulumi/docs/pull/11851

PR #11956, opened 2024-06-04
Stage Starburst new case study

shughes26:
### Proposed changes

Staging new case study for Starburst

### Related issues (optional)

https://github.com/pulumi/marketing/issues/1058


aaronkao (approved):
There is something funny about the logo. I think when its used in the carousel it needs to be a certain size but when its in a page it probably needs to be formatted differently. I updated the quote block and filled in everything else. Please ship after you fix the logo issue. Thanks

PR #11957, opened 2024-06-04
Add ESC TypeScript & Python SDK API reference docs

justinvp:
This PR adds API reference docs for the ESC TypeScript and Python SDKs.

Notes:
- The Python docs are based on the latest published version on PyPi. (We'll want to regen when the latest version is available).
- The Python docs are currently showing everything that's exported from the module -- we may want to filter it down to only certain set of APIs
- The TypeScript docs are based on what's currently in https://github.com/pulumi/esc-sdk/tree/main/sdk/typescript/esc
- We'll definitely want to link to these from the ESC docs.
- We may want to also include links to these at the bottom of https://www.pulumi.com/docs/languages-sdks/javascript/ and https://www.pulumi.com/docs/languages-sdks/python/
- We'll likely want to add some kind of workflow, that will generate these automatically when there is a new release for https://github.com/pulumi/esc-sdk, like [what we do](https://github.com/pulumi/docs/blob/b43a9d88bf67c14dd2b70d81be3325da87f1d7cd/.github/workflows/pulumi-cli.yml#L80-L85) for new releases of `pulumi/pulumi`.


thoward:
@justinvp this is a pretty old draft. Is this ready to merge? If not, what is required to get it to the finish line?

cleverguy25:
The team plans to take this on this quarter

PR #11877, opened 2024-05-20
20k github stars celebration post

aaronkao:
New post to celebrate hitting 20k stars.

PR #11958, opened 2024-06-04
initial draft of inputs/outputs-focused blog post

thoward:
### Proposed changes

Adds a blog post about Pulumi inputs, outputs, and the apply function.

### Related issues (optional)

Fixes https://github.com/pulumi/devrel-team/issues/936


aaronkao:
> **Social Copy:** "Many tools describe themselves as "Infrastructure as Code".... but declarative config files aren't really code, and never will be. Pulumi bridges the worlds of declarative and standard programming using one simple trick; asynchronous value semantics."

You can actually just put it into the markdown file like [line 6](https://github.com/pulumi/docs/pull/11877/files)

content/blog/next-level-iac-briding-the-declarative-gap/index.md
```diff
@@ -0,0 +1,164 @@
...
+    - typescript
+    - python
+    - architecture
+---
+
+Pulumi stands out in the world of infrastructure-as-code (IaC) for its flexibility and ease of use. This is due to its unique approach of using general purpose programming languages to describe system configuration, rather than flat configuration files or domain-specific languages (DSL). It accomplishes this through a small but brilliant implementation detail that bridges the worlds of declarative and standard programming using asynchronous semantics.
+
+<!--more-->
+
+In Pulumi, you describe your infrastructure in code – real code, not a [DSL][dsl-wiki] or [YAML][yaml-wiki] – using your preferred general purpose programming language You don't have to become a specialist in a niche proprietary declarative language, like Terraform's HCL. Rather, you can reuse your existing programming skills, writing in standard imperative, object-oriented, and even functional language styles, while still gaining all of the benefits of the declarative style that other tools emphasize.
```
aaronkao, line 22: the first paragraph and this 2nd paragraph repeat a bunch of similar topics. You make a dig at DSL twice. I think we can consolidate.

content/blog/next-level-iac-briding-the-declarative-gap/index.md
```diff
@@ -0,0 +1,164 @@
...
+
+Pulumi provides a unique mix of a [declarative model][pulumi-declarative-imperative-docs] embedded and implemented inside of a standard programming language, allowing all the flexibility of custom imperative code, while still enabling Pulumi's [deployment engine][pulumi-engine-docs] to infer opportunities for parallel asynchronous execution and to converge a partially-realized system.
+
+The secret sauce that lets Pulumi achieve this is the [`apply`][apply-docs] function, and the way it handles inputs and outputs via asynchronous programming constructs, similar to [futures][futures-wiki]. We'll go deeper on that in this article, but first let's talk a little bit about the flow of operations in Pulumi.
```
aaronkao, line 26: might want to say "future or promises" since promises might be the more well known term

content/blog/next-level-iac-briding-the-declarative-gap/index.md
```diff
@@ -0,0 +1,164 @@
...
+    base64_bytes = base64.b64encode(data_bytes)
+    base64_string = base64_bytes.decode("ascii")
+
+# Export the name of the bucket, in Base64 encoding
+pulumi.export('bucket_name', bucket.id.apply(lambda data: base64_encode))
+```
+
+{{% /choosable %}}
+
+Our example shows how to do Base64 encoding, but this technique can be used for any kind of [transformation function][data-transformation-wiki]. These can be as elaborate as your use case requires, and leverage the full power of your [programming language and its entire ecosystem][next-level-iac-package-ecosystems]. Doing custom transformations like this in other tools, when possible, requires using external coordination which breaks the declarative model and interrupts the deployment engine's ability to fully understand your infrastructure's dependency graph.
```
aaronkao, line 146: I feel like it would be good to compare and contrast Pulumi with a system that doesn't have promises. What does it look like. What is not able to be done. Like is this completely not doable in TF or do they have a workaround here that you get like sorta that functionality. I believe in CFN its completely undoable because it just executes in order of the template

content/blog/next-level-iac-briding-the-declarative-gap/index.md
```diff
@@ -0,0 +1,164 @@
...
+
+While Pulumi's asynchronous semantics may seem somewhat complex for basic use cases, they unlock abilities in more complex use cases, which is when you really need your IaC tooling to show up for you. Because real life is never as simple as demo code.
```
aaronkao, line 148: Maybe give out some examples of complex use cases where this comes in handy

content/blog/next-level-iac-briding-the-declarative-gap/index.md
aaronkao, line 146: its not so much "can be used" but it "has to be used" right?

content/blog/next-level-iac-briding-the-declarative-gap/index.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+{{% /choosable %}}
+
+When this program runs, the call to create the `bucket` object will issue a request to the *deployment engine*. The deployment engine will load the *resource provider* for AWS, and ask it to create the S3 bucket. The AWS resource provider then issues the request to the AWS API, which will create the bucket and then return the `id` of the newly created resource. The provider will return that to the deployment engine, which will then *set the realized value* on your Pulumi program's `bucket.id` property and then move on to creating anything that depends on that property.
+
+This all seems pretty straight-forward, except that Pulumi programs don't wait and imperatively do each operation, one at a time. Instead the deployment engine will run them in parallel, only waiting in cases where the output of one operation is needed to run the next operation. This can create complex structures with some operations needing to wait for inputs from other operations, and some that don't, all finishing at different times, depending on how long the cloud service takes to process that request.
+
+## Inputs and Outputs as asynchronous futures
```
aaronkao, line 80: I think Inputs and Outputs are Pulumi terms. So perhaps you should make the title more generic to the use case being expressed. So maybe its really "Variables with asynchronous futures" or something?

content/blog/next-level-iac-briding-the-declarative-gap/index.md
```diff
@@ -0,0 +1,164 @@
...
+
+Back in our Pulumi program code, that means that some of the time the `bucket` object from our earlier example has its `id` property set, and sometimes it doesn't. We wouldn't want to run any operation that needed that variable before it was set, but we also wouldn't want to wait to start operations that don't need that value.
+
+To address this, Pulumi models the input and output values as [futures][futures-wiki]. This allows you to describe a complex set of needs using a simple value-like syntax. Instead of using a `string` result directly, which would only exist some of the time, you instead work with an [`Output<string>`][output-docs], that Pulumi understands as a reference to a value that currently doesn’t exist, but *will* exist when the operation completes. Similarly, [`Input<string>`][input-docs] describes a dependency on an input string that will *at some point* be provided as the result of another operation. This allows Pulumi to execute the entirety of your graph of infrastructure tasks in parallel, only waiting when necessary.
```
aaronkao, line 86: Can you elaborate here what you mean by "but will exist when the operation completes". I think there is a lot of hidden goodness in that phrase. You can probably talk about how the deployment agent waits on that value before completing the operation...

aaronkao (dismissed):
Looking good so far.

content/blog/next-level-iac-briding-the-declarative-gap/index.md
thoward: Unfortunately during interview/research phase for this article, I was unable to determine any such use cases. I'm sure they exist, but no one was able to describe one to me and I can't personally think of any either. Nonetheless, I'm confident that they *do* exist. Maybe one of our customers could answer that better than our internal team.

content/blog/next-level-iac-briding-the-declarative-gap/index.md
thoward: Well, inputs/outputs are pretty common/general terms but the way we use them in Pulumi is specific. That's sort of what I'm getting into in this section... explaining that they are not just simple input/output values/variables, but actually futures that model both asynchronous value access as well as relationship dependency between components within Pulumi, and that they too function interact in a specific and magical way that belies the simplicity of the terminology and API of Pulumi's Inputs/Outputs.

content/blog/next-level-iac-briding-the-declarative-gap/index.md
thoward: I added a bit more explanation to cover this.

content/blog/next-level-iac-briding-the-declarative-gap/index.md
thoward: This is one approach, but not the only possible one.

thoward: Yeah, It's tempting to compare/contrast, but sticking with the idea that our audience is already familiar with TF and other tools, they should already understand the pain of doing these kinds of operations.

content/blog/next-level-iac-briding-the-declarative-gap/index.md
thoward: good call. i updated it...  the semantics are fuzzy, but IMO Pulumi Inputs/Outputs are technically futures, not promises, but these terms are so often confused/interchanged that in our TypeScript implementation, "Promises" are used under the hood: https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/output.ts#L914 ... and https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/output.ts#L778 are demonstrative. So, really the distinction between the concepts, even in our own code, is unclear.

content/blog/next-level-iac-briding-the-declarative-gap/index.md
thoward: Yep. made some edits to reduce this effect.

content/blog/next-level-iac-briding-the-declarative-gap/index.md
```diff
@@ -0,0 +1,166 @@
...
+    - typescript
+    - python
+    - architecture
+---
+
+Pulumi stands out in the world of infrastructure-as-code (IaC) for its flexibility and ease of use. This is due to its unique approach of using general purpose programming languages to describe system configuration. It accomplishes this through a small but brilliant implementation detail that bridges the worlds of declarative and standard programming using asynchronous semantics.
+
+<!--more-->
+
+In Pulumi, you describe your infrastructure in code – real code, not a [DSL][dsl-wiki] or [YAML][yaml-wiki] – using your preferred general purpose programming language You don't have to become a specialist in a niche proprietary declarative language, like Terraform's HCL. Rather, you can reuse your existing programming skills, writing in standard imperative, object-oriented, and even functional language styles, while still gaining all of the benefits of the declarative style that other tools emphasize.
```
interurban, line 22:
```suggestion
In Pulumi, you describe your infrastructure in code – real code, not a [DSL][dsl-wiki] or [YAML][yaml-wiki] – using your preferred general purpose programming language you don't have to become a specialist in a niche proprietary declarative language, like Terraform's HCL. Rather, you can reuse your existing programming skills, writing in standard imperative, object-oriented, and even functional language styles, while still gaining all of the benefits of the declarative style that other tools emphasize.
```

content/blog/next-level-iac-briding-the-declarative-gap/index.md
```diff
@@ -0,0 +1,166 @@
...
+
+- [**Pulumi Programs**][pulumi-program-docs]: The code you write to describe the infrastructure you need.
+- [**Deployment Engine**][pulumi-engine-docs]: A process that executes requests to create/modify infrastructure.
+- [**Resource Providers**][pulumi-providers-docs]: Plugins that interface with cloud services like AWS, Azure, and others.
+
+These components interact together to manage your infrastructure across multiple cloud providers using a consistent programming and execution model.
+
+{{< figure src="pulumi-architecture-simplified.png" caption="Figure: Pulumi's Program, Deployment Engine, and Resource Provider architecture">}}
+
+Let's get into an example of how information flows back forth between these components. Suppose you want to write a Pulumi program to create an AWS S3 Bucket to [host a static website][pulumi-static-website-example]. When that bucket is created, it will be given a unique indentifier that you need to use in other parts of your infrastructure. Every time you create the bucket the identifier will change, and it's generated by the AWS API, so we can't know it ahead of time. We need to handle this as a variable, not a fixed configuration value.
```
interurban, line 40:
```suggestion
Let's get into an example of how information flows back and forth between these components. Suppose you want to write a Pulumi program to create an AWS S3 Bucket to [host a static website][pulumi-static-website-example]. When that bucket is created, it will be given a unique indentifier that you need to use in other parts of your infrastructure. Every time you create the bucket the identifier will change, and it's generated by the AWS API, so we can't know it ahead of time. We need to handle this as a variable, not a fixed configuration value.
```

interurban (approved):
Looking good Troy; approving to not block. I think all that's left is the CTA at the end we talked about.

PR #11959, opened 2024-06-04
update BMW logo on carousel

shughes26:
### Proposed changes

Update the BMW logo in the customer carousel to the simplified version from case studies


aaronkao:
I think this is wrong. Its not to switch the homepage back to the old BMW logo. Its to switch the logo in the case study to be the one with the mini cooper logo that the homepage has.

shughes26:
@aaronkao got it, swapped it back and updated the case study logo instead

interurban (approved):

PR #11977, opened 2024-06-05
Add better support for deploying dev stacks

cnunciato:
Adds a script and Makefile target to make it easier to deploy to a dev stack. Assumes the target dev stack already exists, and filters out stacks named `production`, `staging`, or `testing`.

Usage:

```bash
# Prompt for all values (stack name, update approval)
make deploy-dev-stack

# Optionally set the stack name, skip build, or skip previews
STACK_NAME=pulumi/christian SKIP_BUILD=true SKIP_PREVIEW=true make deploy-dev-stack
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,3 +144,7 @@ lint:
 .PHONY: format
 format:
 	./scripts/format.sh
+
+.PHONY: deploy-dev-stack
+deploy-dev-stack:
+	./scripts/deploy-dev-stack.sh $1
--- a/scripts/deploy-dev-stack.sh
+++ b/scripts/deploy-dev-stack.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+#!/bin/bash
+
+set -o errexit -o pipefail
+
+source ./scripts/common.sh
+
+# This script runs a full site build, deploys to a unique S3 bucket, and runs Pulumi to deploy the specified stack.
+#
+# Usage:
+#
+#   ./scripts/deploy-dev-stack.sh                               # Runs a full build. Prompts for stack name and before update.
+#   ./scripts/deploy-dev-stack.sh pulumi/stackname              # Runs a full build, skipping the prompt for stack name.
+#   STACK_NAME=pulumi/stackname ./scripts/deploy-dev-stack.sh   # Same as above.
+#   SKIP_BUILD=true ./scripts/deploy-dev-stack.sh               # Skips the build step and goes straight to deployment, but still prompts before update.
+#   SKIP_PREVIEW=true ./scripts/deploy-dev-stack.sh             # Skips the preview before updating. (Is the same as passing --yes.)
+
+echo
+echo "Listing available stacks for this project..."
+echo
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -144,3 +144,7 @@ lint:
 .PHONY: format
 format:
 	./scripts/format.sh
+
+.PHONY: deploy-dev-stack
+deploy-dev-stack:
+	./scripts/deploy-dev-stack.sh
```
sean1588, line 150:
```suggestion
	./scripts/deploy-dev-stack.sh $1
```

Should we pass through the first arg, since looks like the script will take this and assign it to  STACK_NAME if not set, so folks can do something like `make deploy-dev-stack pulumi/cnunciato`

scripts/deploy-dev-stack.sh
```diff
@@ -0,0 +1,66 @@
...
+    echo
+    echo "Refusing to deploy the $stack_name stack with this script. Exiting."
+    exit 0
+fi
+
+echo
+echo "Selecting the $stack_name stack..."
+pulumi -C infrastructure stack select "$stack_name"
+
+if [[ -z "$SKIP_BUILD" || "$SKIP_BUILD" == false ]]; then
```
sean1588, line 41: nice feature 🙂

scripts/deploy-dev-stack.sh
```diff
@@ -0,0 +1,66 @@
...
+
+if [[ "$SKIP_PREVIEW" == true ]]; then
+    pulumi -C infrastructure up --skip-preview
+else
+    pulumi -C infrastructure up
+fi
+
+echo "Done!"
+echo
+echo "https://$(pulumi -C infrastructure stack output websiteDomain)"
```
sean1588, line 65: In order for this to work for "anyone",  I _think_ they will need to configure a cert for the domain and and a route53 hosted zone, but I don't quite remember. I thought we just returned the cloudfront distribution endpoint in these cases where these values are not set, but looks like it is just[ returning what was specified in the config file](https://github.com/pulumi/docs/blob/master/infrastructure/index.ts#L685). We probably should make some more intelligible logic around this in our infrastrucure code. No need to block though. Maybe it would be better to just return the `cloudFrontDomain` output instead.

Makefile
cnunciato: Yeah I'd love it if that worked, but I don't believe it's possible to pass an argument to a make target like that. (I've never been able to make it work, at least.)

scripts/deploy-dev-stack.sh
cnunciato: Actually as long as the hosted zone you've configured in `websiteDomain` is available in the AWS account you're targeting, the cert will be provisioned and everything will "just work". We can definitely return both of these URLs, though. 

Reminds me it'd also be nice to make it easier to *create* a dev stack. That isn't here, but could be added pretty easily later.

Makefile
sean1588: ahhh ok, hmm... I was so sure I have gotten this to work before in the past. I could be misremembering though.

sean1588 (approved):

PR #11982, opened 2024-06-06
Adding Clark Doan

aaronkao:
Adding Clark Doan as Puluminary


sean1588:
It looks like the image is not showing up on the page. I suspect this is due to the file being a png. It looks like all the team images use a jpg image format, so it is likely looking for a file `clark-doan.jpg` when the file is actually `clark-doan.png`.

Screen Shot 2024-06-06 at 9 54 40 AM

sean1588 (approved):

PR #11984, opened 2024-06-06
Use bot token only for posting previews

sean1588:
We are running into rate limitting issues with the PULUMI_BOT_TOKEN, so using it only to post the preview links as this is what it was needed for and using the default GITHUB_TOKEN everywhere else. The default GITHUB_TOKEN is ephemeral and a new one is generated for each run, so less likely it will run into the rate limit.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -49,7 +49,7 @@ jobs:
           PULUMI_STACK_NAME: ${{ vars.PULUMI_STACK_NAME }}
           DEPLOYMENT_ENVIRONMENT: ${{ vars.DEPLOYMENT_ENVIRONMENT }}
           SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-deploy.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           PULUMI_STACK_NAME: ${{ vars.PULUMI_STACK_NAME }}
           DEPLOYMENT_ENVIRONMENT: ${{ vars.DEPLOYMENT_ENVIRONMENT }}
           SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
-          GITHUB_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
sean1588, line 52: reverting this back to using GITHUB_TOKEN. The bot token is no longer needed since the PR workflow is now separated from this workflow and this one does not post previews as it is used only to deploy prod after merging.

cnunciato (approved):

PR #11988, opened 2024-06-06
Add helpers for creating and destroying dev stacks

cnunciato:
This'll make it easier to iterate on website infrastructure generally.

```
make new-dev-stack 
make destroy-dev-stack
```

I used `make destroy-dev-stack` to clear out the `zack`, `devon`, and `leezen` stacks as well. There are a few stack in there that I know we don't use (e.g., all three `staging`s), so we should remove those as well, but we can do that as a separate thing later.


scripts/create-dev-stack.sh
```diff
@@ -0,0 +1,61 @@
...
+    read -p "What would you like to name this stack (e.g., 'christian')? " stack_name
+fi
+
+echo
+echo "Creating pulumi/${stack_name}..."
+echo
+
+pulumi -C infrastructure stack init "pulumi/${stack_name}"
+pulumi -C infrastructure config set aws:region us-west-2
+pulumi -C infrastructure config set certificateArn "arn:aws:acm:us-east-1:372027080554:certificate/54d0431c-2cf9-4c40-bd3c-324cfb8b7d32"
```
sean1588, line 32: I'm assuming the domain on this cert is for `*.pulumi-dev.io` in order for this to work? If it is, this is awesome and will be great for this!

README.md
```diff
@@ -64,6 +64,7 @@ The `Makefile` exposes a number of useful helpers for authoring:
 * `make generate` builds the TypeScript, Python, and Pulumi CLI documentation
 * `make new-blog-post` scaffolds a new, bare-bones blog post with placeholder content
 * `make new-example-program` generates a new multi-language set of examples at `./static/programs`
+* `make new-dev-stack` creates a new dev stack (in the `pulumi` organization, which you must belong to)
```
sean1588, line 67:
```suggestion
* `make new-dev-stack` creates a new dev stack (in the `pulumi` organization, which you must belong to)
* `make destroy-dev-stack` destroys dev stack (in the `pulumi` organization, which you must belong to)
```

feel free to reword that of course

scripts/destroy-dev-stack.sh
```diff
@@ -0,0 +1,75 @@
...
+echo
+echo "Destroying ${stack_name}..."
+
+echo
+echo "Emptying the contents of potentially non-empty buckets..."
+echo
+stack_name_only=$(echo $stack_name | sed -e 's/\(pulumi\/\)//')
+
+pushd infrastructure
+    for resource_name in "website-logs-bucket" "uploads-bucket" "fallback-bucket"; do
```
sean1588, line 48: good catch! I actually think there is an option available now that we can set on the s3 bucket config to empty the bucket when we call `pulumi destroy` before it deletes the bucket. Though what we have now is good for protection though so we can't accidentally delete the buckets in prod. We could also probably get around this by setting the protected resource option though. We can follow up addressing this later.

sean1588 (approved):
looks good!

scripts/create-dev-stack.sh
cnunciato: Yep, this assumed you're using that domain.

PR #11999, opened 2024-06-07
Use pulumi/dev-stacks for dev stacks

cnunciato:
Realized we could just use ESC for this and make things even easier. :sparkles:

Also adds a quick check to make sure the user belongs to the Pulumi org, exiting early if they don't.


sean1588 (approved):
nice! thanks for doing this!

PR #12001, opened 2024-06-08
ESC vs Infisical comparison 

arunkumar611:
To-do

Add explanations of major feature differentiations at the bottom like we have here: https://www.pulumi.com/docs/concepts/vs/terraform/


content/docs/esc/vs/infisical.md
```diff
@@ -31,32 +31,32 @@ Infisical is a secrets management tool that provides a centralized platform for
...
+There are a couple of fundamental differences between Infisical and Pulumi ESC. First, ESC and Infisical differ in that Infisical can only add and manage secrets stored in Infisical. ESC adopts a truly open ecosystem approach, allowing you to pull secrets stored in most secret and password managers during runtime and use them anywhere. This reduces the time to take full inventory of your secrets across your organization , enabling you to utilize the best solution for the purpose.  Second, the composability and hierarchical nature of ESC offers a familiar and powerful experience for users, reducing duplication and allowing them to get up and running quickly. Third, ESC takes a software engineering approach to versioning with ability to add tags and import specific collections of secrets and configuration via those tags, similar to Docker. Fourth, ESC takes a secure limited privilege path to provisioning dynamic short-term credentials compared to Infisical.
```
aaronkao, line 38: My theory is that its actually better to mention what Infiniscal's limitations are rather than what ESC can do. ESC's features is highlighted well in all our ESC focused materials. But What Infiniscal (or competitor) can't do or can only do is really not mentioned anywhere on the internet so its more valuable content because its opinionated.

arunkumar611: Sure. Effectively we can flip the benefits and call them as limitations.

aaronkao, line 38:
```suggestion
There are a couple of fundamental differences between Infisical and Pulumi ESC. First, ESC and Infisical differ in that Infisical can only add and manage secrets stored in Infisical. ESC adopts a truly open ecosystem approach, allowing you to pull secrets stored in most secret and password managers during runtime and use them anywhere. This reduces the time to take full inventory of your secrets across your organization , enabling you to utilize the best solution for the purpose.  Second, Infisical lacks the composability and hierarchical nature of ESC. This slows the speed to get up and running and increases duplication of secrets. Third, ESC takes a software engineering approach to versioning with ability to add tags and import specific collections of secrets and configuration via those tags, similar to Docker. Fourth, ESC takes a more secure limited privilege path to provisioning dynamic short-term credentials compared than Infisical.
```

PR #12004, opened 2024-06-10
Add a step to generate docs JSON

cnunciato:
WIP to explore generating docs JSON for easier ingestion. Includes everything under `/docs`, `/templates`, and `/learn`, and deposits the generated JSON file at `/docs/content.json`. 

Currently renders a single object per page -- so for example, [for this page](https://www.pulumi.com/docs/get-started/index.html), you'd get:

```
{
    "path": "/docs/get-started/index.html",
    "title": "Get Started with Pulumi | Pulumi Docs",
    "h1": "Get started with Pulumi",
    "main": {
        "innerHTML": "<p>Pulumi is an <a href=/what-is/what-is-infrastructure-as-code></a>infrastructure as code platform [...]",
        "textContent": "Pulumi is an infrastructure as code platform [...]",
        "markdown": "Pulumi is an [](/what-is/what-is-infrastructure-as-code)infrastructure as code platform [...]"
    }
}
```

Which would let you do something like:

```typescript
const result = await fetch("/docs/content.json"); // ~28MB, so do this once.
const json = await result.json();

// Get the Markdown for a single page.
const markdown = json.find(doc => doc.path === "/docs/concepts/stack/index.html").markdown;
console.log(markdown); // ~28K.
```

Excludes SDK docs (Node, Python, .NET) for now, but we can add those later. 

This also runs on every build, so what's in this file (e.g., at https://www.pulumi.com/docs/content.json, or whatever we choose to call it) should always match the actual docs that are live on the website.


scripts/content/generate-docs-content.js
```diff
@@ -0,0 +1,50 @@
...
+
+        if (main && main.textContent) {
+            jsonOutput.push({
+                path: path.replace("./public", ""),
+                title: doc.querySelector("title").textContent,
+                h1: doc.querySelector("h1").textContent,
+                main: {
+                    raw: main.rawText,
+                    textContent: main.textContent,
+                    markdown: service.turndown(main.innerHTML),
```
AaronFriel, line 33: It looks like markdown links are `[](url)title` instead of `[title](url)`

cnunciato: Yeah, I noticed this too. Wonder if it's the Hugo minification that's confusing the parser. 🤔

cnunciato: Looks like that was it, yeah; should be better now. (May need to fiddle a bit more to improve rendered Markdown generally, though.)

wongstein:
My understanding of this endpoint is that it's 1 call to get all the docs?  Rather than 1 call per page (for example)

cnunciato:
@wongstein Yep, that's correct, but we could do this however you want; I just figured "one big file" would be easiest to work with.

wongstein:
@cnunciato No no one big file dump is more than enough.  It'll make our ingestion job more dumb to implement, which I like (I don't want to track changes!).

wongstein (approved):

cnunciato:
@wongstein @AaronFriel I'll go ahead and merge this so y'all can begin ingesting this content regularly. (It'll be updated with every website deployment, so multiple times per day.) I'd appreciate if we could treat the endpoint as "alpha", though, as we may want to move it, add to it, etc., but will definitely let you know if we do that. As we get closer to using it "for real", we can settle on a permanent location and structural contract for it.

PR #12016, opened 2024-06-11
Add CTAs to all the Next Level IaC posts

thoward:
This adds consistent CTAs to the previous Next Level IaC posts, similar to the one [at the end of the "bridging the declarative gap" post](https://www.pulumi.com/blog/next-level-iac-briding-the-declarative-gap/#next-steps).


cnunciato (approved):
No need to block, but it'd be sweet to have this pulled into a Hugo shortcode if you're up for it (just to avoid the duplication). What do you think?

PR #12037, opened 2024-06-13
Add missing C# file content for new project

### LINKED ISSUE #12017: Broken language selector UI in create-a-pulumi-project site
### What happened?

1. Navigate to https://www.pulumi.com/learn/pulumi-fundamentals/create-a-pulumi-project/
2. Try to click C# under "Initialize your project". C# option doesn't work, the page continues showing typescript:

image

Expected behavior:

Language changes to `csharp`

### Example

N/A

### Output of `pulumi about`

N/A

### Additional context

_No response_

### Contributing

iwahbe:
Fixes #12017

The bug was caused because not all language checkers had a `csharp` entry, and since the language checker is global it doesn't make sense to display a language that can't be selected for one checker.

Ideally, we would introduce a lint failure to make this mistake impossible (and catch other places where we made the mistake). That is out of scope for this PR.

### DIFF
--- a/content/learn/pulumi-fundamentals/create-a-pulumi-project/index.md
+++ b/content/learn/pulumi-fundamentals/create-a-pulumi-project/index.md
@@ -260,7 +260,7 @@ For YAML, your {{< langfile >}} is also your program's main entrypoint file.
 Use the command <code>cat</code>{{< langfile >}} to explore the contents of your
 project's empty program:
 
-{{< chooser language "typescript,python,go,java,yaml" / >}}
+{{< chooser language "typescript,python,go,csharp,java,yaml" / >}}
 
 {{% choosable language typescript %}}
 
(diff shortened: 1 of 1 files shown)


mikhailshilkov:
Wait, but why does a chooser with a missing option down below affects the chooser that has C# and prevents it from showing it? Isn't this a bug on its own?

arturl (approved):

arturl:
Verified that the fix works in [site prerview](http://www-testing-pulumi-docs-origin-pr-12037-7a31a0c0.s3-website.us-west-2.amazonaws.com).

cnunciato (changes requested):
The [other](https://www.pulumi.com/learn/pulumi-fundamentals/create-docker-images/) [pages](https://www.pulumi.com/learn/pulumi-fundamentals/configure-and-provision/) of the guide actually don't support C#, so the bug is that there's a C# option in the chooser when there shouldn't be.

Secondarily though yes, @mikhailshilkov is right -- the chooser is supposed to support a "local" mode (so you can choose a tab for a language that may not exist elsewhere on the page), and that seems not to be working here. I'll verify and get something filed if that's what's happening here.

We're in the process of moving this content to a new section as well (in https://github.com/pulumi/docs/pull/11883), so we can address this issue there as we move things over. Sorry for the confusion on this @iwahbe -- and thank you for calling it out @arturl!

iwahbe:
I would say that having C# for the first part of the guide is better then not having it, even if it doesn't continue for the rest of the guide.

PR #12056, opened 2024-06-14
Open preview links in a new tab

cnunciato:


cnunciato:
Closing as it looks like GitHub doesn't allow this. (They appear to be stripping the `target` attribute out of the `a` tag even when we send it.)

PR #12073, opened 2024-06-18
Update runner references to ubuntu-latest

cnunciato:
The `ubuntu-22.04-8core` runner seems no longer valid (causing jobs to get stuck in the queue). This updates all of our runners to use `ubuntu-latest`.

Fixes https://github.com/pulumi/docs/issues/12074.


cnunciato:
Admin-merging to unblock the queue.

PR #12077, opened 2024-06-18
Fix CSS issues with cookie overlay

### LINKED ISSUE #12072: Bottom 20% of the nav is not clickable
In the browser, all the links in the following picture from "Environments (ESC)" and below are not clickable links.

image

The bug occurs for me in Safari and Chrome, but not Firefox. I've verified that it's related to the position of the elements on the screen, not the content. In other words, if I scroll, the bottom 20% or so of the links are unclickable no matter what the link is.

cnunciato:
Fixes some usability problems with the cookie overlay:

* Makes it easier to see by adding a border and making the button higher-contrast
* Prevents it from covering the left nav and making items unclickable
* Makes the close button a little bigger and margins and padding consistent

Fixes #12072.


layouts/partials/head.html
```diff
@@ -232,7 +232,7 @@
     </script>
 
     <script async defer src="https://buttons.github.io/buttons.js"></script>
-    {{- if eq hugo.Environment "production" }}
+    {{- if ne hugo.Environment "production" }}
```
cnunciato, line 235: This is just so we can validate this change in the PR preview. We'll switch it back before merging.

jkodroff:
Looks fixed to me.

jkodroff (approved):
IDK if anything else might've broke since I don't really understand the code, but I can say that the issue appears fixed.

PR #12100, opened 2024-06-21
docs readme updates

interurban:
Updating the readme:
- adding toc, community and additional resources links


README.md
```diff
@@ -21,7 +39,9 @@ See also:
 * [Publishing a Pulumi blog post](./BLOGGING.md)
 * [Documentation and coding style guide](./STYLE-GUIDE.md)
 
-## Toolchain
+# Setup and Development
```
sean1588, line 42: nit: I think we typically want  these heading levels to move down sequentially. So changing this to an h2 will make it that way and the following headings are all h3s already, so those should be fine.

sean1588:
```suggestion
## Setup and Development
```

sean1588 (approved):

Issue #12109, opened 2024-06-24
Add toolchain option to project file docs


julienp:
Document the toolchain option for the python runtime.

PR #12110, opened 2024-06-24
Add new runtime options

julienp:
The python runtime has a new `toolchain` option.
The nodejs runtime has a new `packagemanager` option.

[Preview](http://www-testing-pulumi-docs-origin-pr-12110-ddc9bf77.s3-website.us-west-2.amazonaws.com/docs/concepts/projects/project-file/#runtime-options)

Fixes https://github.com/pulumi/docs/issues/12109
Ref https://github.com/pulumi/pulumi/issues/15937

I'll follow up with a PR to explain the toolchain option, and the poetry choice, in the python SDK docs.

### DIFF
--- a/content/docs/concepts/projects/project-file.md
+++ b/content/docs/concepts/projects/project-file.md
@@ -57,9 +57,11 @@ The runtime attribute has an additional property called `options` where you can 
 | - | - | - |
 | `typescript` | Only applies to the `nodejs` runtime | Boolean indicating whether to use `ts-node` or not. |
 | `nodeargs` | Only applies to the `nodejs` runtime | Arguments to pass to `node`. |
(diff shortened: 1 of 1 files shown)


lukehoban (approved):

lukehoban:
Should we also add notes on these features to https://www.pulumi.com/docs/languages-sdks/python/, and possibly update a few things there (there are notes about Pip being requried for example).

julienp:
> Should we also add notes on these features to https://www.pulumi.com/docs/languages-sdks/python/, and possibly update a few things there (there are notes about Pip being requried for example).

Yes, that's on my radar. I think we can restructure the virtual environment section a bit. I don't think we need to talk about Pulumi 2.4.0 anymore.

PR #12114, opened 2024-06-24
Add docs for self-hosted deployments

seanyeh:

### DIFF
--- a/content/docs/pulumi-cloud/self-hosted/components/api.md
+++ b/content/docs/pulumi-cloud/self-hosted/components/api.md
@@ -435,3 +435,16 @@ scrape_configs:
     static_configs:
       - targets: ["api.pulumi.com"]
 ```
+
+## Deployments
+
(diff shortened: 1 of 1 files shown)


content/docs/pulumi-cloud/self-hosted/components/api.md
```diff
@@ -435,3 +435,16 @@ scrape_configs:
     static_configs:
       - targets: ["api.pulumi.com"]
 ```
+
+## Deployments
+
+In order to enable [Deployments](/docs/pulumi-cloud/deployments), the following must be configured:
```
cleverguy25, line 441: I think we should also link to how to setup a self managed agent, as that is a key piece as well

seanyeh: I have that a couple lines below

cleverguy25 (approved):

PR #12117, opened 2024-06-24
Do not recommend using a `-native` suffix for native providers

iwahbe:
### Proposed changes

My understanding was that we were moving away from `-native` naming schemes. If so, we shouldn't recommend it to others.


mjeffryes (approved):
Good call

PR #12136, opened 2024-06-26
Dynamic providers/secrets and creds

MitchellGerdisch:
### Proposed changes

Updated the dynamic providers page as follows:
- Updated the github python example to not show setting a hard-coded credential by using Pulumi config.
- This github python change was then propagated to the typescript and javascript examples for consistency. Also modified the intro text for the github example to provide more color.
- Added another example that uses environment variables to pass in the creds.
- Also added a note about how the Command provider can be used as a lighter weight solution instead of a dynamic provider.

### Related issues (optional)

The github python example is dependent this PR being merged: https://github.com/pulumi/pulumi/pull/15864


content/docs/concepts/resources/dynamic-providers.md
```diff
@@ -12,14 +12,18 @@ aliases:
 - /docs/intro/concepts/resources/dynamic-providers/
 ---
 
-There are three types of resource providers. The first are the standard resource providers. These resource providers are built and maintained by Pulumi. There is a second kind, called a dynamic resource provider, which we will discuss here. These resource providers run only in the context of your program. They are not shareable. The third type of resource provider is shareable. You write it yourself and then you can distribute it so that others can use it.
+There are three types of resource providers. The first type are standard resource providers. These resource providers are built and maintained by Pulumi and can be found in the [Pulumi Registry](https://www.pulumi.com/registry/). The second type are custom providers written by you using [Pulumi Packages](/docs/using-pulumi/pulumi-packages). Both the standard and custom providers are able to be used across all the languages Pulumi supports. The third type is the dynamic resource provider discussed on this page.
```
interurban, line 15:
```suggestion
There are three types of resource providers. The first type are standard resource providers. These resource providers are built and maintained by Pulumi and can be found in the [Pulumi Registry](https://www.pulumi.com/registry/). The second type are custom providers written by you using [Pulumi Packages](/docs/using-pulumi/pulumi-packages). Both the standard and custom providers are able to be used across all the languages Pulumi supports. The third type is the dynamic resource provider discussed on this page.
```
```suggestion
There are three types of resource providers. The first type are standard resource providers. These resource providers are built and maintained by Pulumi and can be found in the [Pulumi Registry](https://www.pulumi.com/registry/). The second type are custom providers written by you using [Pulumi Packages](/docs/using-pulumi/pulumi-packages). Both the standard and custom providers can be used across all the languages Pulumi supports. The third type is the dynamic resource provider discussed on this page.
```

content/docs/concepts/resources/dynamic-providers.md
```diff
@@ -536,7 +540,7 @@ class MyResource(Resource):
 
 ### Example: Random
```
interurban, line 541:
```suggestion
### Example: Random number generator
```

content/docs/concepts/resources/dynamic-providers.md
```diff
@@ -536,7 +540,7 @@ class MyResource(Resource):
...
+This example generates a random number using a dynamic provider. It represents the simplest dynamic provider that brings together the various topics described above. It highlights using dynamic providers to run some code only when a resource is created, and then store the results of that in the state file so that this value is maintained across deployments of the resource. Because we want our random number to be created once, and then remain stable for subsequent updates, we cannot use a random number generator in our program; we need dynamic providers. The result is a provider similar to the one provided by the standard [Random Provider](https://www.pulumi.com/registry/packages/random/), just specific to our program and language.
```
interurban, line 543:
```suggestion
This example generates a random number using a dynamic provider. It represents the simplest dynamic provider that brings together the various topics described above. It highlights using dynamic providers to run some code only when a resource is created and then stores the results in the state file so that this value is maintained across deployments of the resource. Because we want our random number to be created once, and then remain stable for subsequent updates, we cannot use a random number generator in our program; we need dynamic providers. The result is a provider similar to the one provided by the standard [Random Provider](https://www.pulumi.com/registry/packages/random/), just specific to our program and language.
```

interurban (approved):
Thx Mitch, I flagged a few things, but I'm out next week so dropping my approval as well. One other one, you have an added docs.sln file in the merge. (its an odd artifact from visual studio, can you delete it before merging?). thx so much for working on this!

thoward:
@MitchellGerdisch looks like the PR this was depending on got merged (https://github.com/pulumi/pulumi/pull/15864). Would you say this is ready to merge now?

MitchellGerdisch:
@thoward Not quite.
I need to do some rewrites based on 
https://github.com/pulumi/pulumi/pull/17673
https://github.com/pulumi/pulumi/pull/17697

PR #12139, opened 2024-06-27
Add first take on self hosted changelog

cleverguy25:
### Proposed changes

This is a first take on a self hosted changelog.  This is maintained manually for now but will look for cases where we can make ths part of the dev workflow. It is not a normal every PR changelog as that is far too granular and not every PR pertains to self hosted installations, which creates noise.

### DIFF
--- a/content/docs/pulumi-cloud/self-hosted/changelog.md
+++ b/content/docs/pulumi-cloud/self-hosted/changelog.md
@@ -1,8 +1,8 @@
 ---
-title_tag: Changelog | Self-Hosting Pulumi
+title_tag: Changelog | 
 meta_desc: Changes related to the self-hosted version of Pulumi Cloud.
 title: Changelog
-h1: Pulumi Cloud self-hosted network requirements
+h1: Self-Hosted Changelog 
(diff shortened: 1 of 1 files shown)


content/docs/pulumi-cloud/self-hosted/changelog.md
```diff
@@ -0,0 +1,68 @@
...
+* Rename email change request email
+* Stop navbar flickering after visiting the Resources page
+* Hide GH App on integrations page for self-hosted
+* Fix required marker in delete org dialog
+
+### April
+
+* [Enable OIDC Trust Relationships](http://localhost:1313/blog/oidc-trust-relationships/)
+* Enable otel metrics
+* Expose metrics endpoint
```
lukehoban, line 44: Are there docs pages we can link to on these?

lukehoban (approved):
Looks like a good start. It does raise question of whether there are any other items here that *should* have additional documentation added for them that we can link to - though nothing stands out as especially critical. I do think this provides useful anchors for consumers to at least know what to ask or about, which may lead to fleshing out more docs on demand as we go.

PR #12158, opened 2024-06-28
Add Python Poetry documentation

julienp:
### Proposed changes

We now have native `poetry` support to manage virtual enviroments and
dependencies.

### Related issues (optional)

Ref: https://github.com/pulumi/pulumi/issues/15937

### DIFF
--- a/content/docs/languages-sdks/python/_index.md
+++ b/content/docs/languages-sdks/python/_index.md
@@ -50,7 +50,7 @@ When creating a new Python project with `pulumi new`, you are offered the choice
 
 This behavior is controlled by the `toolchain` and `virtualenv` `runtime` options in `Pulumi.yaml`.
 
-Set the `toolchain` option to `pip`, along with the `virtualenv` option to use `pip`:
+To use pip, set the `toolchain` option to `pip`, along with the `virtualenv` option: 
 
 ```yaml
 runtime:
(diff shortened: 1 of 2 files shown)


content/docs/languages-sdks/python/_index.md
```diff
@@ -149,9 +136,42 @@ $ venv/bin/pip install -r requirements.txt
...
+#### Poetry
+
+To add a new dependency when using `poetry`, run the `poetry add` command in your project directory:
+
+{{< chooser os "macos,windows,linux" >}}
+
+{{% choosable os macos %}}
+
+```bash
+$ poetry add ${PACKAGE_NAME}
```
julienp, line 148: I don't have a concrete package name to use here, are we ok with using syntax like `${PACKAGE_NAME}` to indicate that this is a place holder?

cnunciato: This is okay, but for command-line args, we typically use this notation, with angle-brackets and hyphens:

```
poetry add <package-name>
```

So I'd suggest sticking with that if possible.

cnunciato: I think this is fine, yeah, although we don't have a particular standard around this yet (and probably should). I'll poll the team and see what they think, but IMO this is good for now.

layouts/shortcodes/install-python.html
```diff
@@ -11,12 +11,13 @@
     </div>
     <div class="content">
         <p>
-            <code>pip</code> is required to install dependencies. If you installed Python from source, with an installer from
+            <code>pip</code> or <code>poetry</code> are required to install dependencies. If you installed Python from source, with an installer from
```
cnunciato, line 14:
```suggestion
            Either <code>pip</code> or <code>poetry</code> is required to install dependencies. If you installed Python from source, with an installer from
```

cnunciato (approved):
Looks great! Nothing need block, just a few minor suggestions for improvement.

PR #12214, opened 2024-07-05
fix: remove stray parenthesis

ayewo:
### Proposed changes

This stray single-character was preventing a hyperlink from being rendered correctly.

N/A

### Related issues (optional)
N/A


ericrudder (approved):
thanks for the fix

PR #12252, opened 2024-07-11
Remove unused parameter from the REST API docs

### LINKED ISSUE #12156: List teams API call references `continuationToken` which is ignored by the API
### Problem description

- I saw a product documentation issue on the following URL: https://github.com/pulumi/docs/blob/master/content/docs/pulumi-cloud/cloud-rest-api/_index.md#list-teams

### Suggestions for a fix

The continuationToken is not implemented for that endpoint on pulumi cloud, so should be removed from the docs.

cnunciato:
Fixes #12156.


cnunciato:
Just noticed #12228, so will close in favor of that.

PR #12268, opened 2024-07-12
Add info about using the pulumi-std functions in yaml

MitchellGerdisch:
### Proposed changes

Updated the YAML reference to include information about using the `pulumi-std` library to do things such as add (or subtract) numbers.

### DIFF
--- a/content/docs/languages-sdks/yaml/yaml-language-reference.md
+++ b/content/docs/languages-sdks/yaml/yaml-language-reference.md
@@ -319,7 +319,7 @@ The expression `${item}` will return a JSON value `{ "key1": "value1", "key2": 1
 
 ##### `fn::invoke`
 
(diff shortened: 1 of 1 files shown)


phillipedwards (approved):
LGTM

content/docs/languages-sdks/yaml/yaml-language-reference.md
```diff
@@ -319,7 +319,9 @@ The expression `${item}` will return a JSON value `{ "key1": "value1", "key2": 1
 
 ##### `fn::invoke`
 
-Calls a function from a package and returns either the whole object or a single field if given the `return` property. The schema is:
+Calls a function from a package and returns either the whole object or a single field if given the `return` property. This can be any of the functions in the [registry providers](https://www.pulumi.com/registry/) as well as functions found in the [pulumi-std](https://github.com/pulumi/pulumi-std/tree/master/std) package that includes math functions such as `sum` and `abs` and logic functions such as `anytrue` as well as string functions such as `split`.
```
lukehoban, line 322:
```suggestion
Calls a function from a package and returns either the whole object or a single field if given the `return` property. This can be any of the functions in the [registry providers](https://www.pulumi.com/registry/) as well as functions found in the [pulumi-std](https://github.com/pulumi/pulumi-std/blob/master/FUNCTION_LIST.md) package that includes math functions such as `sum` and `abs`, logic functions such as `anytrue`, and string functions such as `split`.
```

lukehoban (approved):

PR #12278, opened 2024-07-15
Minor edit to Hcl vs pulumi blog for grammar

mjeffryes:
Great blog post but this line tripped me up when I was reading. Not sure if my edit is exactly what you meant to say, but it's an option anyway.

### Proposed changes

Just a small wording change in the first sentence.


dirien (approved):
Mega good catch @mjeffryes! 🙌🙌

PR #12299, opened 2024-07-18
Replace node-html-parser with Cheerio

cnunciato:
The node-html-parser library seems to have trouble parsing some of our docs, so this PR switches to [Cheerio](https://cheerio.js.org/) so we can evaluate that instead.


sean1588 (approved):

sean1588:
nice, i see the missing page in the content.json is in the preview now.

Issue #11326, opened 2024-04-22
Add debugging how to guide


interurban:
We have a long standing customer / user ask to better document how to debug with pulumi; see https://github.com/pulumi/pulumi/issues/1372 for the dialog of how to do this. It's also an opportunity to experiment with micro conetnt, short animated, and re-think how docs/help is consumed.

related to https://github.com/pulumi/docs/issues/11327 - work item for a poc video

interurban:
Guide launched on our blog: https://www.pulumi.com/blog/next-level-iac-briding-the-declarative-gap/

PR #12308, opened 2024-07-18
Debugging guide

mikhailshilkov:
### Proposed changes

Adds a guide to debug Pulumi programs. Summarizes the learnings in comments to https://github.com/pulumi/pulumi/issues/1372.

I tested all languages on my local machine.

### Prior art

- We have [Breakpoint Debugging for Pulumi Programs](https://www.pulumi.com/blog/next-level-iac-breakpoint-debugging/) but it's TypeScript-only and "blog-style" instead of docs-style.
- We have [Attaching a debugger to a Pulumi program](https://www.pulumi.com/blog/next-level-iac-breakpoint-debugging/) which I will link this guide from

### Related issues (optional)

- Resolves https://github.com/pulumi/docs/issues/11326 for all languages
- https://github.com/pulumi/pulumi/issues/1372
- First step of https://github.com/pulumi/pulumi/issues/15206


content/docs/using-pulumi/debugging/_index.md
```diff
@@ -0,0 +1,285 @@
...
+
+navigate to the `Run and Debug` menu on the left pane, and then click the Start Debugging (F5) button with the `.NET Core Attach` configuration. Search for the process by your project name and attach:
+
+![Debug: Attach to .NET Process](./dotnet-code-attach.png)
+
+Now you can step through the program and inspect variables. Once completed, resume the execution and let the Pulumi CLI complete.
+
+## Debugging Go programs
+
+Prepend your Pulumi program with the code to pause and give yourself time to attach the debugger:
```
EronWright, line 191: I think you should mention that the Go program should be complied with debug flags using `go build -gcflags all=-N -l`, and then use the `runtime.binary` field in the project file to launch the built executable.

mikhailshilkov: > Please mention that the Go program should be compiled with debug flags using go build -gcflags all=-N -l, then used as the binary in Pulumi.yaml. Otherwise breakpoints may not work as expected.

I did not need to change anything about the compilation flags, it just worked. Was it just luck?

EronWright: I think it _almost_ works without flags, but can be frustrating. [Here's](https://www.jetbrains.com/help/go/attach-to-running-go-processes-with-debugger.html#step-1-build-the-application) an example of the routine guidance.

content/docs/using-pulumi/debugging/_index.md
```diff
@@ -0,0 +1,295 @@
...
+            "projectName": "App",
+            "request": "attach",
+            "hostName": "localhost",
+            "port": 8000
+        }
+    ]
+}
+```
+
+navigate to the `Run and Debug` menu on the left pane, and then click the Start Debugging (F5) button with the `Debug (Attach)` configuration.
```
sean1588, line 289:
```suggestion
Navigate to the `Run and Debug` menu on the left pane, and then click the Start Debugging (F5) button with the `Debug (Attach)` configuration.
```

sean1588 (approved):

sean1588:
I have approved the content. I have not tested all of the steps in the doc.

content/docs/using-pulumi/debugging/_index.md
```diff
@@ -0,0 +1,295 @@
...
+        {
+            "name": ".NET Core Attach",
+            "type": "coreclr",
+            "request": "attach"
+        }
+    ]
+}
+```
+
+navigate to the `Run and Debug` menu on the left pane, and then click the Start Debugging (F5) button with the `.NET Core Attach` configuration. Search for the process by your project name and attach:
```
cnunciato, line 183:
```suggestion
Navigate to the `Run and Debug` menu on the left pane, and then click the Start Debugging (F5) button with the `.NET Core Attach` configuration. Search for the process by your project name and attach:
```

content/docs/using-pulumi/debugging/_index.md
```diff
@@ -0,0 +1,295 @@
...
+            "mode": "remote",
+            "remotePath": "${workspaceFolder}",
+            "port": 2345,
+            "host": "127.0.0.1"
+        }
+    ]
+}
+```
+
+navigate to the `Run and Debug` menu on the left pane, and then click the Start Debugging (F5) button with the `Connect to server` configuration.
```
cnunciato, line 252:
```suggestion
Navigate to the `Run and Debug` menu on the left pane, and then click the Start Debugging (F5) button with the `Connect to server` configuration.
```

content/docs/using-pulumi/debugging/_index.md
```diff
@@ -0,0 +1,295 @@
...
+    usingpulumi:
+        identifier: debugging
+        weight: 9
+---
+
+Because Pulumi uses general purpose programming languages to provision cloud resources, you can take advantage of native debugging tools to troubleshoot your infrastructure definitions.
+
+The tools differ depending on the language you are using, but the process stays similar:
+
+1. Start with an existing Pulumi project, that contains `Pulumi.yaml` and your program.
```
cnunciato, line 17:
```suggestion
1. Start with an existing Pulumi project that contains `Pulumi.yaml` and your program.
```

content/docs/using-pulumi/debugging/_index.md
```diff
@@ -0,0 +1,295 @@
...
+
+```bash
+pulumi up
+Previewing update (dev)
+
+ Type                 Name          Plan   Info
+ pulumi:pulumi:Stack  play-cs-dev          'dotnet build -nologo .' completed successfully
+```
+
+Navigate to your IDE and attach the debugger to the process with the name of your .NET project. For example, in VS Code, you can define the following `launch.json` file:
```
cnunciato, line 168: Any reason not to mention setting a breakpoint here also, as we do in the other sections?

content/docs/using-pulumi/debugging/_index.md
```diff
@@ -0,0 +1,295 @@
...
+```cs
+using System.Diagnostics;
+
+return await Pulumi.Deployment.RunAsync(async () =>
+{
+    while (!Debugger.IsAttached)
+    {
+        await System.Threading.Tasks.Task.Delay(1000);
+    }
+    Debugger.Break();
```
cnunciato, line 152: FWIW, I stumbled a bit on this one as I interpreted this line as "being" the breakpoint -- e.g., that I had to move this line to where I wanted to breakpoint to happen. Might be worth mentioning this is just an initial breakpoint and that others can be set within the IDE.

content/docs/using-pulumi/debugging/_index.md
```diff
@@ -0,0 +1,295 @@
...
+Now you can step through the program and inspect variables. Once completed, resume the execution and let the Pulumi CLI complete.
+
+## Debugging .NET programs
+
+Prepend your Pulumi program with the code to wait for a debugger to attach:
+
+```cs
+using System.Diagnostics;
+
+return await Pulumi.Deployment.RunAsync(async () =>
```
cnunciato, line 146: Might be worth mentioning the `async` bit is necessary here, as without it, you'll get a very noisy compile error.

content/docs/using-pulumi/debugging/_index.md
```diff
@@ -0,0 +1,295 @@
+---
+title_tag: "Debugging Pulumi Programs"
+meta_desc: "Guide to attach a debugger to Pulumi programs and step through the code."
```
cnunciato, line 3:
```suggestion
meta_desc: "Guide to attaching a debugger to Pulumi programs and stepping through the code."
```

cnunciato (commented):
So awesome to see this. Thanks @mikhailshilkov! Some suggestions, but otherwise looks awesome.

thoward:
Looks great! Nice work. 

I would suggest removing the stub section "Debugging YAML programs".

I would also edit to highlight some other IDEs. We are primarily showing how to do this with VS Code, but there are many popular IDEs which all work a little differently.

I would mention that the preview step and the update step will *both* attach to the debugger, and break on the same line, which can be confusing because you mostly will want to debug on the update step. 

Also, it might be helpful to explain the execution context that these programs are running under (as a separate subprocess launched by `pulumi`. I'm not sure if our docs highlight our execution model anywhere. If so we could just link to that. This can be confusing in a debugging context because the thing you're attaching to is not `pulumi` CLI app. Also, the `wait for attach` can be happening on multiple processes at once (though each process must attach to a unique port). So it's a good idea to define a unique port to make sure you're attaching to the correct process.

Another point is the process of launching `pulumi`. Most IDE users want to build, launch, AND debug using the `F5` key, which complicates things a little vs the process you've outlined here. I actually think this is a better method (building/launching from command-line, attaching debugger with `F5`) but it will be important to note that this is different from the way you'd build/launch/debug a typical program from an IDE in a single step (and that this is necessary because of the execution context of `pulumi`).

cnunciato:
>I would suggest removing the stub section "Debugging YAML programs".

I actually think it's better to have this, as without it, someone will inevitably land on this page and be left wondering how to debug a YAML program.

thoward:
> > I would suggest removing the stub section "Debugging YAML programs".
> 
> I actually think it's better to have this, as without it, someone will inevitably land on this page and be left wondering how to debug a YAML program.

In that case maybe we should add a little more here. "It's not possible to debug a YAML program because YAML is a configuration file, not an executable program.".. and maybe point folks to techniques to lint YAML, debug providers, etc. (e.g. explain that while breakpoint debugging isn't possible, these are the tools and techniques you WOULD want to use if your using YAML instead of a GPL).

content/docs/using-pulumi/debugging/_index.md
```diff
@@ -0,0 +1,295 @@
...
+```bash
+Previewing update (dev)
+
+     Type                      Name           Plan       Info
+ +   pulumi:pulumi:Stack       aws-ts-dev     create     For help, see: https://nodejs.org/en/docs/inspector
+```
+
+Set a breakpoint in your program.
+
+Then, attach your debugger to the process with the `node` runtime. For example, in VS Code, you can define the following `launch.json` file:
```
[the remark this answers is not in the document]
mikhailshilkov: Great callout! I changed all line to something like 

```
For example, in [VS Code](https://code.visualstudio.com/docs/editor/debugging), you can define
the following `.vscode/launch.json` file:
```

content/docs/using-pulumi/debugging/_index.md
```diff
@@ -0,0 +1,295 @@
...
+                "<node_internals>/**/*.js"
+            ]
+        }
+    ]
+}
+```
+
+navigate to the `Run and Debug` menu on the left pane, and then click the Start Debugging (F5) button with the `Node.js: Attach` configuration.
+
+Depending on your setup, you may need to attach multiple times until the right node process is found. For example, you may be running Node.js policy packs that will hit the debugger as well.
```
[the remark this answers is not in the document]
mikhailshilkov: We launch multiple `node` processes during `pulumi up` execution, and only one of them runs the user's program. Unfortunately, `NODE_OPTIONS` applies to all of them, so all `node` processes are launched in debug mode. In my sample project, I had to click `Run and Debug` twice to get to the breakpoint in my program.

content/docs/using-pulumi/debugging/_index.md
mikhailshilkov: Because `Debugger.Break();` will break for the user. However, maybe it makes sense to unify this with other sections, remove `Debugger.Break();` and ask the user to set a breakpoint. I will do that.

content/docs/using-pulumi/debugging/_index.md
mikhailshilkov: Removed now and added a suggestion for the user to set a breakpoint, similar to all other languages.

content/docs/using-pulumi/debugging/_index.md
mikhailshilkov: Added a comment to this effect

mikhailshilkov:
@sean1588 @cnunciato @thoward Thank you for your thoughtful reviews, I really appreciate it!

I tried to address most comments, see the last commit.

> I would also edit to highlight some other IDEs. We are primarily showing how to do this with VS Code, but there are many popular IDEs as well as command-line debuggers which all work a little differently.

I agree it would be useful, but I suggest opening a follow-up issue to track this. In the ideal world, we would build some interactive page where users can pick their language and their IDE and get clean instructions specific to their use case. But I suggest landing the VS Code version first.

content/docs/using-pulumi/debugging/_index.md
```diff
@@ -0,0 +1,299 @@
...
+```plain
+Previewing update (dev)
+
+     Type                      Name           Plan       Info
+ +   pulumi:pulumi:Stack       aws-ts-dev     create     For help, see: https://nodejs.org/en/docs/inspector
+```
+
+Open your Pulumi program in your IDE and set a breakpoint at the line where you want the debugger to stop.
+
+Then, attach your debugger to the process with the `node` runtime. For example, in [VS Code](https://code.visualstudio.com/docs/editor/debugging), you can define the following `.vscode/launch.json` file:
```
danielrbradley, line 50: Should we recommend using the JavaScript Debug Terminal if using VSCode - there's zero additional config required.

mikhailshilkov: Added a mention of auto-attach and debug terminal

cnunciato (approved):

PR #12319, opened 2024-07-19
Add check for empty email form submissions

sicarul:
## Description

This PR addresses issue https://github.com/pulumi/marketing/issues/864 by adding a check to make sure events generated from Hubspot's callback do not generate fake form submission events to Segment, polluting metrics with empty form fills.

### DIFF
--- a/theme/stencil/src/components/hubspot-form/hubspot-form.tsx
+++ b/theme/stencil/src/components/hubspot-form/hubspot-form.tsx
@@ -148,7 +148,7 @@ export class HubspotForm {
                 utmSource: utmData.source,
                 utmMedium: utmData.medium,
             };
-            if(emailAddress !== ""){
+            if(emailAddress !== ""){ // Don't track empty email addresses
                 analytics.track("form-submission", submissionData);
             }
         }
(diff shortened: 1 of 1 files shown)


theme/stencil/src/components/hubspot-form/hubspot-form.tsx
```diff
@@ -148,8 +148,9 @@ export class HubspotForm {
                 utmSource: utmData.source,
                 utmMedium: utmData.medium,
             };
-
-            analytics.track("form-submission", submissionData);
+            if(emailAddress !== ""){ // Don't track empty email addresses
```
cnunciato, line 151: This works, but would suggest putting this check higher up, since the work of the function is about submitting:

```
if (analyticsAvailable && emailAddress !== "") {
    // ...
```

cnunciato (changes requested):
Thanks for this fix! Just one minor suggestion.

theme/stencil/src/components/hubspot-form/hubspot-form.tsx
sicarul: Fixed!! Thanks

sicarul:
@cnunciato are any other changes needed?

cnunciato (approved):

PR #12322, opened 2024-07-20
Spruce up case studies pages

joeduffy:
This is a proof of concept to spruce up our case studies a wee bit.

Namely:

* Adds a number of new logos to the homepage. Also changes styling a little bit. We use a continuous scroller, rather than a jumpy one. We also make logos grayscale until hover, but let you click on any of them to go to the case studies page, not just those that happen to have case studies (which is not obvious to the reader).

* Instead of just saying "Trusted by your peers", say "Trusted by 2,700+ cloud innovators." This seems much stronger.

* Use this same slogan on the case studies page itself. There is apparently a subtitle capability for heros, but I doubt we use it anywhere since the formatting was funky and text illegible. I fixed that and used it for this slogan, to tighten up the page and waste less whitespace above the fold. This makes the logos visible.

* Rename the page from "Case Studies" to "Customers" -- that's what it's about! -- and in general, refer to these as "Customer Stories" versus "Case Studies" so we can hammer home the customer-centricity.

* Put all of the logos on this page too! Now they are in color and instead of scrolling, they are laid out in a clean table.

* Add a highlighted customer story. For now I have used Unity. This is a colorful little segment below the logos that has a quote, highlights the business impact, and links to the case study.

* From there, the page remains the same ... however, I have many ideas how to improve this too. But all of the above, I think, is an independently valuable and landable subset of changes.

This is not ready to ship as-is. I am a Tailwind/SCSS amateur and I'm sure some of this isn't as beautiful as it could be. I also just jammed in some inline JavaScript instead of updating our swiper component, in the interest of just getting a prototype working. Clearly we would want to fix that up if we go forward with these changes ...


joeduffy:
Example *before*:

image

And *after*:

image

joeduffy:
Here's an example of a before/after for the individual case studies:

**Before**:

image

**After**:

image

Tons of formatting issues -- the logo can be GIGANTIC -- and I want to go back and fix up the case study landing page to be consistent -- but pushing anyway just so we can iterate.

I also haven't added all of the quotes/metrics, but the intent was to leverage Marlowe's recent work here.

joeduffy:
This was forked and merged in https://github.com/pulumi/docs/pull/12456.

PR #12327, opened 2024-07-22
Document `dependsOn` workarounds for invokes

lunaris:
Invokes, or provider functions, are exposed in many SDKs alongside resources, and allow programs to call arbitrary provider-specific APIs. Just as resources may be passed resource options, so may invokes be passed invoke options, with the usual suspects such as `parent` and `provider` making an appearance and appearing as one might expect.

One resource option that is currently not available as an invoke option, despite appearing a natural addition, is `dependsOn`, which allows Pulumi users to make explicit dependencies and ordering constraints in their programs that Pulumi cannot otherwise track. One reason that `dependsOn` does not currently exist as an invoke option is that any given invoke can be called in one of two ways:

* A "direct form", whereby the function accepts plain arguments and either blocks until it returns a value or returns a language-native asynchronous value (such as a `Promise` in NodeJS or a `Task` in Python, for instance).

* An "output form", whereby the function accepts any Pulumi `Input` values and returns a Pulumi `Output` value.

Since direct-form invocations do not use Pulumi `Input`s or `Output`s, they do not form part of the Pulumi dependency graph. Passing `dependsOn` to a direct-form invocation would therefore make no sense. The sharing of the set of invoke options between the two forms means that `dependsOn` is consequently not currently available.

Thankfully, one can work around the absence of `dependsOn` for output-form invocations using e.g. `pulumi.all` and `.apply`. This commit updates the documentation on invokes/provider functions to (hopefully) make this clear and provide examples where possible. While we are here we also bolster some of the other pieces of documentation on provider functions, such as defining the terms "direct form" and "output form" which are already referenced on other pages.

Fixes https://github.com/pulumi/pulumi/issues/14243

image
image
image

### DIFF
--- a/content/docs/concepts/resources/functions.md
+++ b/content/docs/concepts/resources/functions.md
@@ -131,34 +131,170 @@ variables:
 
 Provider functions are exposed in each language as regular functions, in two variations:
 
- 1. A function that accepts plain arguments (strings and so on) and returns a Promise, or blocks until the result is available.
- 2. A function that accepts `Input` values and returns an [Output](/docs/concepts/inputs-outputs/).
+ 1. The **direct form** accepts plain arguments (e.g. `string`, as opposed to `pulumi.Input<string>`) and returns an asynchronous value (e.g. a `Promise` in NodeJS, or a `Task` in Python), or blocks until the result is available. These functions are typically named, e.g., `getFoo()`.
+ 2. The **output form** accepts `Input` values (or plain values) and returns an [Output](/docs/concepts/inputs-outputs/). These functions are typically named, e.g., `getFooOutput()`.
 
-The documentation for a provider function will tell you the name and signature for each of the variations.
+The [Pulumi Registry](/registry) contains authoritative documentation for all provider functions.
 
 #### Invoke options
 
-Each function and method also accepts "invoke options", either as an object or as varargs depending on the host language. The options are as follows:
+Each function and method also accepts "invoke options", similar to the way Pulumi resources accept [resource options](/docs/concepts/options/). Invoke options may be specified either as an object or as varargs depending on the language you're writing your Pulumi program in. The options are as follows:
 
-| Option | Explanation                                                  |
-|--------|--------------------------------------------------------------|
-| parent | Supply a parent resource, which will be used to determine default providers |
-| provider | Supply the provider to use explicitly. |
-| version | Use exactly this version of the provider plugin. |
-| pluginDownloadURL | Download the provider plugin from this URL. The download URL is otherwise inferred from the provider package. |
-| _async_ | _This option is deprecated and will be removed in a future release_ |
+- `parent`: Supply a parent resource for this function call. Much like the [parent resource option](/docs/concepts/options/parent/), the parent will be consulted when determining the provider to use.
 
-The `parent` option has a similar purpose to the [parent option](/docs/concepts/options/parent/) used when creating a resource. The parent is consulted when determining the provider to use.
+- `pluginDownloadURL`: Pass a URL from which the provider plugin should be fetched. This may be necessary for third-party packages such as those not hosted at [https://get.pulumi.com](https://get.pulumi.com).
 
-The `provider` option gives an explicit provider to use when running the invoked function. This is useful, for example, if you want to invoke a function in each of a set of AWS regions.
+- `provider`: Pass an [explicitly configured provider](/docs/concepts/resources/providers/#explicit-provider-configuration) to use for this function call, instead of using the default provider. This is useful, for example, if you want to invoke a function in each of a set of AWS regions.
 
-The `version` option specifies an exact version for the provider plugin. This can be used when you need to pin to a specific version to avoid a backward-incompatible change.
+- `version`: Pass a provider plugin version that should be used when invoking the function.
 
-The `pluginDownloadURL` option gives a URL for fetching the provider plugin. It may be necessary to supply this for third-party packages (those not hosted at [https://get.pulumi.com](https://get.pulumi.com)).
+- `async`: _This option is deprecated and will be removed in a future release_.
+
+### Dependencies and ordering
+
+While the direct and output forms of a provider function are equivalent in terms of the results they produce when invoked, they differ in how they interact with the rest of the Pulumi program and the order in which they may be executed. Specifically:
+
+- Direct form invocations execute just like any other function call in the language. Since they do not accept Pulumi `Input`s nor return Pulumi `Output`s, they are not tracked by the Pulumi engine and do not participate in the dependency graph.
+
+- Output form invocations, on the other hand, are tracked by the Pulumi engine and participate in the dependency graph. This means, for example, that Pulumi will ensure that input resources are created or updated before an invocation and that the invocation is executed before its dependent resources are created or updated.
+
+If you require that dependent resources are created or updated before an invocation, you must use a provider function's output form. If you need to specify a dependency that can't be captured by passing an appropriate input (that is, if you wish to simulate something like the [`dependsOn` resource option](/docs/concepts/options/dependson/)), you can use Pulumi's [`all`](/docs/concepts/inputs-outputs/all/) function and `Output`'s [`apply`](/docs/concepts/inputs-outputs/apply/) method:
+
+{{< chooser language "typescript,python,go,csharp,java,yaml" >}}
+{{% choosable language typescript %}}
+
+ ```typescript
+import * as pulumi from "@pulumi/pulumi";
+
+const res1 = new MyResource("res1", {});
+const res2 = new MyResource("res2", {});
+
+// Assuming `myFunctionOutput` is an output-form invocation of the `myFunction`
+// provider function, this use of `all` and `apply` will ensure that it does not
+// happen until `res1` and `res2` have been processed. This will work for any
+// set of resources, even those with no explicit outputs, since the `.urn`
+// output is always available.
+pulumi.all([res1.urn, res2.urn]).apply(() => myFunctionOutput());
+```
+
+{{% /choosable %}}
+{{% choosable language csharp %}}
+
+```csharp
+using System.Collections.Generic;
+using System.Linq;
+using Pulumi;
+
+return await Deployment.RunAsync(() =>
+{
+    var res1 = new MyResource("res1", new MyResourceArgs());
+    var res2 = new MyResource("res2", new MyResourceArgs());
+
+    // Assuming `myFunctionOutput` is an output-form invocation of the `myFunction`
+    // provider function, this use of `Tuple` and `Apply` will ensure that it does
+    // not happen until `res1` and `res2` have been processed. This will work for
+    // any set of resources, even those with no explicit outputs, since the `.Urn`
+    // output is always available.
+    Output.Tuple(res1.Urn, res2.Urn).Apply(t => myFunctionOutput());
+});
+```
+
+{{% /choosable %}}
+{{% choosable language go %}}
+
+```go
+package main
+
+import (
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+func main() {
+	pulumi.Run(func(ctx *pulumi.Context) error {
+		res1, err := NewMyResource(ctx, "res1", nil)
+		if err != nil {
+			return err
+		}
+
+		res2, err := NewMyResource(ctx, "res2", nil)
+		if err != nil {
+			return err
+		}
+
+		// Assuming `myFunctionOutput` is an output-form invocation of the `myFunction`
+		// provider function, this use of `All` and `ApplyT` will ensure that it does not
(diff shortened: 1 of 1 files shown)


content/docs/concepts/resources/functions.md
```diff
@@ -131,34 +131,195 @@ variables:
...
+import * as pulumi from "@pulumi/pulumi";
+
+const project = new gcp.organizations.Project("project", {
+    name: "project",
+    projectId: "project",
+    billingAccount: "bills",
+    autoCreateNetwork: false,
+});
+
+// Since `project.name` is immediately available, we need to use `all` and
```
tgummerer, line 182:
```suggestion
// Since `project.name` isn't immediately available, we need to use `all` and
```

I think?  Or am I misreading something?  Same with the other comments.

content/docs/concepts/resources/functions.md
```diff
@@ -131,34 +131,195 @@ variables:
...
+<pulumi-choosable type="language" values="java">
+
+Output form invocations are not yet supported in Java
+
+</pulumi-choosable>
+</div>
+<div>
+<pulumi-choosable type="language" values="yaml">
+
+Output form invocations are not yet supported in YAML
```
tgummerer, line 313:
```suggestion
Output form invocations are not supported in YAML
```

I'm not sure we'll ever want to support them?  "Not yet" gives the impression that we do want to.

Frassle: I don't see why we couldn't support them in YAML eventually?

content/docs/concepts/resources/functions.md
Frassle: Yeh I don't think this is the best example. `name` and `urn` resolve at the same time for project. Ideally this would use a different resource where it's input isn't needed to make the call but it's creation is needed for the call to actually work.

lunaris: I think this is right. Because `project.name` _is_ immediately available (since it was provided as a literal), Pulumi won't block until `project` is ready. `project.urn`, however, _won't_ be available until `project` has actually been bonafide constructed, so the `all` fixes that. @Frassle is this complete nonsense?

tgummerer: Ah then I misunderstood  In that case `project.name` wouldn't need to be part of `pulumi.all` at all though right?  Maybe something like 

// Since `project.name` is immediately available (before the project is ready), but we also want to wait for the urn to be ready, we need to use `all` and

Or maybe

// Since `project.name` is immediately available (before the project is ready), we can use `project.urn` as a proxy for it being ready.  This is why we need to use `all` to make sure it's completely ready

content/docs/concepts/resources/functions.md
tgummerer: Fair, I guess I didn't see how they fit in well with yaml, but I also didn't think long about it.

content/docs/concepts/resources/functions.md
lunaris: Yea this example isn't great it turns out -- have replaced with something synthetic but at least correct (and this seems to be in line with how `dependsOn` is currently documented)

content/docs/concepts/resources/functions.md
```diff
@@ -131,34 +131,187 @@ variables:
...
+
+- Output form invocations, on the other hand, are tracked by the Pulumi engine and participate in the dependency graph. This means, for example, that Pulumi will ensure that input resources are created or updated before an invocation and that the invocation is executed before dependent resources are created or updated.
+
+In general, you should prefer the output form of a provider function when you need to ensure that the function is executed in the correct order with respect to other resources in your program. This applies particularly if you wish to simulate something like the [`dependsOn` resource option](/docs/concepts/options/dependson/), where you want to ensure that a resource is created before the function is invoked, without necessarily passing the resource as an argument to the function. While there is no `dependsOn` [invoke option](#invoke-options) for provider functions, you can achieve the same effect using Pulumi's [`all`](/docs/concepts/inputs-outputs/all/) function and `Output`'s [`apply` method](/docs/concepts/inputs-outputs/apply/):
```
cnunciato, line 164: This wording feels a bit squishy to me. As I understand it, it's not that you should _generally prefer_ the output form when you need to ensure that the function is executed in the correct order with respect to other resources -- it's that that's your only option. Right? (In other words, you _can't_ use the direct form to invoke a function in relation to the output of a resource because resource outputs don't produce plain values.) The next sentence is also a bit confusing in that it begins with "this applies" (i.e. presumably, "using the output form applies") when you want to ensure that a resource is created before the function is invoked _without passing the resource as an argument_ -- but since doing so is actually impossible, it can't apply either.

IMO it'd be clearer and more direct to say that when a function accepts one or more output arguments, Pulumi will resolve the outputs before invoking the function, and if you need to specify a dependency on outputs that the function doesn't depend on directly, you can use `pulumi.all` to tell Pulumi to resolve those outputs first, then invoke the function yourself, as in the following example. (Or similar.) 

Mentioning `dependsOn` probably makes sense too, given we're adding this in response to a request for that property. (Which I've actually wished for myself!) Do you know if we're planning to do this, though? If we are, we might instead want to link to that issue from this page and then position this approach as an option to be used in the meantime.

content/docs/concepts/resources/functions.md
```diff
@@ -131,34 +131,187 @@ variables:
...
+    }
+}
+```
+
+</pulumi-choosable>
+</div>
+<div>
+<pulumi-choosable type="language" values="yaml">
+
+Output form invocations are not yet supported in YAML
```
cnunciato, line 305:
```suggestion
Output form invocations are not yet supported in YAML.
```

Worth asking, though -- will they be? If so, is there an issue tracking it?

content/docs/concepts/resources/functions.md
```diff
@@ -203,18 +356,18 @@ def get_kubeconfig(self,
 <div>
 <pulumi-choosable type="language" values="java">
 
-(no example available for Java)
+No example available for Java
```
cnunciato, line 359: Are there just no examples in Java and YAML, or are provider methods not supported in those languages? If not supported, is it that they aren't supported _yet_ (in which case it'd also be good to link to an issue), or that they'll never be? (Generally best to be clear on things like this, as folks will likely wonder the same.)

content/docs/concepts/resources/functions.md
```diff
@@ -203,18 +356,18 @@ def get_kubeconfig(self,
...
-(no example available for YAML)
+No example available for YAML
 
 </pulumi-choosable>
 </div>
 
 </pulumi-examples></div>
 
-Unlike provider functions, methods always take `Input` arguments, and return an `Output`. Methods do not have invoke options.
+Unlike provider functions, methods always appear in the _output form_: they take `Input` arguments, and return an `Output`. Moreover, methods do not accept invoke options.
```
cnunciato, line 373: >Moreover, methods do not accept invoke options.

Just curious -- what does it actually mean that "methods don't accept invoke options"? If "invoke options" are themselves just function arguments, and methods do take arguments, what is this sentence supposed to convey to me as a user? If we deleted it, would anything useful be lost?

cnunciato (changes requested):
Thanks for writing this up! No major issues with the content, only some suggestions, but it'd be good to fix up the code to use shortcodes instead of raw markup.

content/docs/concepts/resources/functions.md
```diff
@@ -131,34 +131,187 @@ variables:
...
+
+- Output form invocations, on the other hand, are tracked by the Pulumi engine and participate in the dependency graph. This means, for example, that Pulumi will ensure that input resources are created or updated before an invocation and that the invocation is executed before dependent resources are created or updated.
+
+In general, you should prefer the output form of a provider function when you need to ensure that the function is executed in the correct order with respect to other resources in your program. This applies particularly if you wish to simulate something like the [`dependsOn` resource option](/docs/concepts/options/dependson/), where you want to ensure that a resource is created before the function is invoked, without necessarily passing the resource as an argument to the function. While there is no `dependsOn` [invoke option](#invoke-options) for provider functions, you can achieve the same effect using Pulumi's [`all`](/docs/concepts/inputs-outputs/all/) function and `Output`'s [`apply` method](/docs/concepts/inputs-outputs/apply/):
+
+<div><pulumi-examples>
```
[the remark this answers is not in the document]
lunaris: Ah thanks!

content/docs/concepts/resources/functions.md
lunaris: Good shout -- I've gone for some stronger wording on this and would be grateful to get your thoughts! In terms of linking an issue -- for now, this is the resolution to that issue, precisely because the sharing of `InvokeOptions` across the two forms when only one (the output form) actually supports the notion, is problematic. Open to thoughts on this though.

content/docs/concepts/resources/functions.md
lunaris: I'm not actually sure on this -- I've merely (perhaps wrongly) added some cosmetic changes to this bit but am not responsible for the original docs (nor sure on the answers to your questions).

content/docs/concepts/resources/functions.md
lunaris: So for a "normal" invoke you can do something like `aws.getAmi(..., { provider: myCustomProvider })`, for instance, where `{ provider: myCustomProvider }` are the invoke options (akin to resource options). Methods don't allow this, (presumably -- again not an expert here) because they'll inherit the options from the resource on which they are being called (so e.g. we don't want to allow you to overload the provider since the owning resource will have already defined one). So at a language level, yes, the method takes arguments, but only those which are inputs to the function, not the "second set" that let you control bits around how Pulumi performs the invoke. If that makes sense? Happy to make changes as you see fit on this.

content/docs/concepts/resources/functions.md
```diff
@@ -131,34 +131,170 @@ variables:
 
 Provider functions are exposed in each language as regular functions, in two variations:
 
- 1. A function that accepts plain arguments (strings and so on) and returns a Promise, or blocks until the result is available.
- 2. A function that accepts `Input` values and returns an [Output](/docs/concepts/inputs-outputs/).
+ 1. A function that accepts plain arguments (strings and so on) and returns an asynchronous value (e.g. a `Promise` in NodeJS, or a `Task` in Python), or blocks until the result is available. This is often referred to as the _direct form_.
+ 2. A function that accepts `Input` values and returns an [Output](/docs/concepts/inputs-outputs/). This is often referred to as the _output form_.
```
jkodroff, line 135:
```suggestion
 1. The **direct form** accepts plain arguments (e.g. `string`, as opposed to `pulumi.Input<string>`) and returns an asynchronous value (e.g. a `Promise` in NodeJS, or a `Task` in Python), or blocks until the result is available. These functions are typically named, e.g., `getFoo()`.
 2. The **output form** accepts `Input` values (or plain values) and returns an [Output](/docs/concepts/inputs-outputs/). These functions are typically named, e.g., `getFooOutput()`.
```

content/docs/concepts/resources/functions.md
```diff
@@ -131,34 +131,170 @@ variables:
...
 
 The documentation for a provider function will tell you the name and signature for each of the variations.
```
jkodroff, line 137:
```suggestion
The [Pulumi Registry](/registry) contains authoritative documentation for all provider functions.
```

content/docs/concepts/resources/functions.md
```diff
@@ -131,34 +131,170 @@ variables:
...
 
 #### Invoke options
 
 Each function and method also accepts "invoke options", either as an object or as varargs depending on the host language. The options are as follows:
```
jkodroff, line 141:
```suggestion
Each function and method also accepts "invoke options", similar to the way Pulumi resources accept [resource options](concepts/options/). Invoke options may be specified either as an object or as varargs depending on the language you're writing your Pulumi program in. The options are as follows:
```

jkodroff (commented):
Added a few suggestions to make scanning the page easier and to make some things more explicit.

content/docs/concepts/resources/functions.md
cnunciato: Yes, it does make sense, thanks! I didn't realize it was this "second set" that this referred to.

content/docs/concepts/resources/functions.md
```diff
@@ -131,34 +131,170 @@ variables:
...
+
+{{% /choosable %}}
+{{% choosable language yaml %}}
+
+Output form invocations are not yet supported in YAML
+
+{{% /choosable %}}
+{{< /chooser >}}
 
 ### Provider methods
```
cnunciato, line 295: Smallish thing, but would it be clearer to call these Resource methods instead (since they're exposed on resources)? Seems so, but will leave the call up to you.

```suggestion
### Resource methods
```

cnunciato (approved):
Couple of suggestions, but otherwise looks good! 🚀

PR #12364, opened 2024-07-24
Upgrade Cypress to latest

cnunciato:
We're running a couple of major versions behind, so this PR upgrades to latest. (Changes are all the result of running the Cypress migration wizard.)


sean1588 (approved):
nice catch!!

PR #12371, opened 2024-07-25
Updating meta image

aaronkao:
Removing the color Pulumi logo on the meta image and replacing it with an all white so it doesn't blend into the background


interurban (approved):
looks good

PR #12405, opened 2024-07-29
Send production deploy failures to docs-ops-critical

sean1588:
part of: https://github.com/pulumi/devrel-team/issues/893

As a way to help differentiate more critical alerts that need to be addressed with immediacy from those that are less urgent, this will send the production build and deploy pipeline failures (indicating a blockage of our prod pipeline) to the #docs-ops critical channel. Let me know what you think.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -75,7 +75,7 @@ jobs:
       - name: Slack Notification
         uses: docker://sholung/action-slack-notify:v2.3.0
         env:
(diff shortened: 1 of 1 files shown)


cnunciato (changes requested):
[See my comment here](https://github.com/pulumi/registry/pull/5052#pullrequestreview-2206417689). I don't think there's anything particularly "critical" about these alerts now vs. historically, so I'd suggest we keep to the notification channel we already have, and adjust our process around that as needed.

PR #12429, opened 2024-08-02
Fixed headers to be more in line with the content

pierskarsenbarg:
This feels more inline with the content


pierskarsenbarg:
Feels a bit better:

image

cnunciato (approved):
Agree -- thanks!

PR #12440, opened 2024-08-04
fix currentOrg exampleon Pulumi ESC Syntax Reference page

dbeattie71:
fix currentOrg example on Pulumi ESC Syntax Reference page.


thoward (approved):
Nice catch @dbeattie71 ! Really appreciate the contribution.

PR #12454, opened 2024-08-05
Improved Kubernetes await logic blog post

blampe:
This post includes an overview of upcoming improvements to await logic in the Kubernetes provider.

Note that some of these changes are still in review, so this should not be published until they've been released.


content/blog/improved-kubernetes-await-logic/index.md
```diff
@@ -0,0 +1,293 @@
...
+* More helpful errors are surfaced when a resource doesn't become ready in time.
+* A new `pulumi.com/deletionPropagationPolicy` annotation enables faster deletions.
+* A new `pulumi.com/waitFor` annotation allows for custom readiness criteria.
+* And a new experimental feature enables readiness criteria for all resources.
+
+<!--more-->
+
+## Better error reporting
+
+Our goal is for the Pulumi Kubernetes provider to surface as much information as possible when something fails, because we don't want you to have to manually query your cluster just to figure out what happened.
```
mjeffryes, line 76: nit: technically I think our goal is "we don't want you to have to manually query your cluster just to figure out what happened." and the means is "to surface as much information as possible when something fails". 

However, I don't really know if rewording it that way is an improvement. 

"Our goal is that Pulumi Kubernetes provider users should never need to manually query the cluster to figure out why an update failed."

mjeffryes (approved):
This looks solid to me. Clear description of the changes with examples, and why they matter to users. All anchored in the core value prop of using pulumi for K8s.

content/blog/improved-kubernetes-await-logic/index.md
```diff
@@ -0,0 +1,293 @@
...
+    issuerRef: {
+      name: issuer.metadata.name,
+      kind: issuer.kind,
+      group: "cert-manager.io",
+    },
+  },
+});
+```
+
+This uses a [v2 ConfigFile](https://www.pulumi.com/blog/kubernetes-yaml-v2/) to `kubectl apply` the cert-manager custom resource definitions, and then it creates a self-signed Issuer and Certificate using those CRDs.
```
rquitales, line 201: nit: add CRD abbreviation as well

eg: `the cert-manager CustomResourceDefinition (CRD), and then it creates ...`

content/blog/improved-kubernetes-await-logic/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+Being able to specify custom readiness is powerful, but a lot of resources follow well-known conventions like the `Ready=True` condition.
+Rather than requiring you to specify `"pulumi.com/waitFor": "condition=Ready"` on every resource that needs it, there's now an opt-in feature to automatically apply well-known readiness criteria to all resources managed by Pulumi.
+
+Running your program with the environment variable `PULUMI_K8S_AWAIT_ALL=true` will cause Pulumi to automatically wait for _all_ resources to become ready when creating or updating them -- including custom resources.
+
+If a resource appears to follow well-known conventions (like the "Ready" condition, among others) Pulumi will wait for it to become ready.
+If the resource doesn't appear to follow well-known conventions, Pulumi will maintain its existing behavior and assume the resource is ready immediately.
+
+(Generic readiness criteria are determined by the [cli-utils](https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus) SIG, and behavior is unchanged for any resources that Pulumi already waits for readiness.)
```
rquitales, line 266: `cli-utils` is a collection of Go libraries that is maintained by [SIG CLI](https://github.com/kubernetes/community/blob/master/sig-cli/README.md).

rquitales (changes requested):

rquitales (approved):

content/blog/improved-kubernetes-await-logic/index.md
```diff
@@ -0,0 +1,293 @@
...
+   (You can configure this path with `terminationMessagePath`, or use `terminationMessagePolicy: FallbackToLogsOnError` to use a Pod's logs as its termination message.)
+
+## Faster deletions with "pulumi.com/deletionPropagationPolicy"
+
+Pulumi uses [foreground](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#foreground-deletion) cascading deletion by default, which deletes a resource _and_ all of its dependents.
+This is the most robust form of deletion but it is also the slowest, and until now there was no way to safely customize this behavior.
+
+The new `pulumi.com/deletionPropagationPolicy` annotation allows you to customize delete behavior while still allowing Pulumi to await readiness when creating or updating the resource.
+
+The annotation accepts three propagation policies:
```
EronWright, line 96:
```suggestion
The annotation accepts three propagation policies, corresponding to the options supported by Kubernetes (see "[Cascading Deletion](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#cascading-deletion)").
```

content/blog/improved-kubernetes-await-logic/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+1. "background": delete the owner resource and leave dependent resources to be asynchronously garbage collected.
+   This is faster than "foreground" deletion propagation, but dependent resources can remain temporarily or even indefinitely if they are not finalized.
+2. "orphan": delete the owner resource and leave dependent resources untouched.
+   This can be useful if you want to keep resources around for migration or debugging purposes.
+3. "foreground": the default behavior of deleting the resource and all of its dependents.
+   This is slower but guarantees all dependents have been cleaned up if it succeeds.
```
EronWright, line 103: This exposition makes perfect sense from a Kubernetes perspective, but might confuse when taken in a Pulumi context. Consider making the descriptions be more Kubernetes-centric, like use the term "object".

content/blog/improved-kubernetes-await-logic/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+{{% notes type="info" %}}
+Using `skipAwait` to speed up deletion is not recommended when server-side apply is enabled because it is not respected by all resources and it can lead to race conditions during replacement.
+The current behavior is considered buggy, and future releases might change the behavior of the `skipAwait` annotation to have no effect during deletion.
+
+The `pulumi.com/deletionPropagationPolicy` annotation is the recommended way to delete something quickly and safely.
```
EronWright, line 109: I don't understand this advice because they're not analogues. You're implying that deletionPropagationPolicy would make the cleanup of the managed resource be quicker, which I suppose would be true if the object has children that are slow to cleanup.

content/blog/improved-kubernetes-await-logic/index.md
```diff
@@ -0,0 +1,293 @@
...
+{{% /notes %}}
+
+## Experimental: Custom readiness with "pulumi.com/waitFor"
+
+{{% notes type="info" %}}
+The `pulumi.com/waitFor` annotation is considered experimental and its syntax may change in the future.
```
EronWright, line 115: Next time we could adopt the Kubernetes convention of using `.alpha.` in the annotation's name, for experimental stuff.  For example: `external-dns.alpha.kubernetes.io/ttl`.

content/blog/improved-kubernetes-await-logic/index.md
```diff
@@ -0,0 +1,293 @@
...
+{{% /notes %}}
+
+There is no one-size-fits-all definition for what it means for a Kubernetes resource to be "ready," so we have also introduced a new annotation, `pulumi.com/waitFor`, which allows you to specify custom readiness criteria on a per-resource basis.
+This is functionally similar to `kubectl` [wait](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_wait/).
+
+The new `pulumi.com/waitFor` annotation can be applied to any non-[Patch](https://www.pulumi.com/blog/kubernetes-server-side-apply/) resource with metadata, and it accepts three possible formats:
```
EronWright, line 121: It strikes me as a bug that a patch resource doesn't wait for readiness. Assumedly the patch is causing a spec change, meaning a rollout may occur.  Seems no different from awaiting an update. For purposes of the blog post, could we offer a note on the topic?

content/blog/improved-kubernetes-await-logic/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+        "pulumi.com/waitFor": "jsonpath={.status.phase}=Running"
+
+   If a value isn't provided, the resource is considered ready when any value exists at the given path.
+   This resource will wait until it has a web hook configured with a CA bundle:
+
+        "pulumi.com/waitFor": "jsonpath={.webhooks[*].clientConfig.caBundle}"
+
+2. A string prefixed with "condition=" followed by the type of the condition and an optional status.
+   This matches the behavior of `kubectl wait --for=condition=...` and will wait until the resource has a matching condition.
```
EronWright, line 135:
```suggestion
   This matches the behavior of `kubectl wait --for=condition=...` and will wait until the resource has a matching status condition.
```

content/blog/improved-kubernetes-await-logic/index.md
```diff
@@ -0,0 +1,293 @@
...
+      kind: issuer.kind,
+      group: "cert-manager.io",
+    },
+  },
+});
+```
+
+This uses a [v2 ConfigFile](https://www.pulumi.com/blog/kubernetes-yaml-v2/) to `kubectl apply` the cert-manager custom resource definitions (CRDs), and then it creates a self-signed Issuer and Certificate using those CRDs.
+
+The dependencies between resources _appear_ reasonable, however this program will likely fail when it's first run:
```
EronWright, line 203:
```suggestion
The dependencies between resources _appear_ reasonable, however this program will likely fail when it is first run:
```

content/blog/improved-kubernetes-await-logic/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+{{% notes type="warning" %}}
+error: resource "kubernetes:cert-manager.io/v1:ClusterIssuer::issuer" was not successfully created by the Kubernetes API server: Internal error occurred: failed calling webhook "webhook.cert-manager.io": failed to call webhook: Post "https://cert-manager-webhook.cert-manager.svc:443/validate?timeout=30s": tls: failed to verify certificate: x509: certificate signed by unknown authority
+{{% /notes %}}
+
+We see this "certificate signed by unknown authority" error because a few things need to happen before the cert-manager web hooks are usable:
```
EronWright, line 209:
```suggestion
We see this "certificate signed by unknown authority" error because a few things need to happen before the cert-manager webhooks are operational:
```

content/blog/improved-kubernetes-await-logic/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+* At least one `cainjector` pod needs to be running.
+* The `cainjector` pod(s) need to have a leader elected.
+* The `cainjector` leader needs to inject certificate authority data into the cert-manager web hooks.
```
EronWright, line 213:
```suggestion
* The `cainjector` leader needs to inject certificate authority data into the cert-manager webhooks.
```

content/blog/improved-kubernetes-await-logic/index.md
```diff
@@ -0,0 +1,293 @@
...
+
+Previously, Pulumi had no way to know whether those pre-requisite steps had taken place.
+However, now we can use the `pulumi.com/waitFor` annotation to instruct Pulumi to wait until web hooks have CA data injected before proceeding to create Certificates.
+One way to achieve this is by using the new [transforms API](https://www.pulumi.com/blog/resource-transforms/) to annotate the web hooks appropriately:
```
EronWright, line 217:
```suggestion
One way to achieve this is by using the new [transforms API](https://www.pulumi.com/blog/resource-transforms/) to annotate the webhooks appropriately:
```

content/blog/improved-kubernetes-await-logic/index.md
```diff
@@ -0,0 +1,293 @@
...
+          };
+        }
+        return undefined;
+      },
+    ],
+  },
+);
+```
+
+Here we've given the `pulumi.com/waitFor` annotation a value of `jsonpath={.webhooks[*].clientConfig.caBundle}`, which is a [JSONPath] expression that will wait until a web hook in the `webhooks` array has a `clientConfig` with a `caBundle` populated -- indicating that the web hook is usable.
```
EronWright, line 250:
```suggestion
Here we've given the `pulumi.com/waitFor` annotation a value of `jsonpath={.webhooks[*].clientConfig.caBundle}`, which is a [JSONPath] expression that will wait until a webhook in the `webhooks` array has a `clientConfig` with a `caBundle` populated -- indicating that the webhook is fully operational.
```

EronWright (approved):

content/blog/improved-kubernetes-await-logic/index.md
blampe: Exactly! Changed this to include "...if it has children that are slow to cleanup."

content/blog/improved-kubernetes-await-logic/index.md
blampe: This is a great point!

content/blog/improved-kubernetes-await-logic/index.md
blampe: I actually can't remember why I wrote this, because I don't think this is actually true -- patch resources should be fine.

PR #12483, opened 2024-08-09
Added clarification for non-pulumi cloud secrets providers

pierskarsenbarg:
### Proposed changes

Added clarification that its safe to commit the encryption key value in the stack config to source control

### DIFF
--- a/content/docs/concepts/secrets.md
+++ b/content/docs/concepts/secrets.md
@@ -603,7 +603,7 @@ config:
 
 Decrypting this ciphertext requires the encryption key that was used to create it. For stacks managed with Pulumi Cloud, these keys are obtained automatically, but only for users with [read access](/docs/pulumi-cloud/projects-and-stacks/#stack-permissions) to the stack. For self-managed backends, the keys must be supplied by the user, either by providing the stack's current passphrase (when using the [`passphrase`](#changing-the-secrets-provider-for-a-stack) provider) or by authenticating with the stack's [encryption provider](#available-encryption-providers).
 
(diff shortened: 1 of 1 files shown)


content/docs/concepts/secrets.md
```diff
@@ -603,7 +603,7 @@ config:
...
 
-It's therefore considered safe and good practice to check these files into source control (including the `encryptionSalt`s used with the passphrase provider), as doing so allows you to version your code and configuration in tandem. If you'd prefer not to check in these files, however, you can easily rebuild them, using the most recently deployed configuration, with [`pulumi config refresh`](/docs/cli/commands/pulumi_config_refresh/).
+It's therefore considered safe and good practice to check these files into source control (including the `encryptionSalt`s used with the passphrase provider or `encryptionKey` when one of the other secrets providers), as doing so allows you to version your code and configuration in tandem. If you'd prefer not to check in these files, however, you can easily rebuild them, using the most recently deployed configuration, with [`pulumi config refresh`](/docs/cli/commands/pulumi_config_refresh/).
```
lukehoban, line 606:
```suggestion
It's therefore considered safe and good practice to check these files into source control (including the `encryptionSalt`s used with the passphrase provider or `encryptedKey` when one of the other secrets providers), as doing so allows you to version your code and configuration in tandem. If you'd prefer not to check in these files, however, you can easily rebuild them, using the most recently deployed configuration, with [`pulumi config refresh`](/docs/cli/commands/pulumi_config_refresh/).
```

lukehoban (approved):
One change needed - but looks good after making that change.

PR #12485, opened 2024-08-09
Show arm64 CLI download links if available.

### LINKED ISSUE #8046: Available Versions should list both x64 and arm64 links
Currently, the Available Versions table includes x64 links for Linux, macOS, and Windows. We don't include any links to the arm64 builds, even though they're available.

https://www.pulumi.com/docs/get-started/install/versions/

Screen Shot 2022-09-20 at 10 47 03 AM

We should update the table to link to arm64 binaries as well as x64.

smithrobs:
### Proposed changes

Show arm64 CLI download links if available (seems 2.23.0 was when we started arm64 builds)

Feel free to offer UI suggestions; I tried to keep the same table spacing but there's not tons to work with:

arm64 add

### Related issues (optional)

Fixes #8046

### DIFF
--- a/layouts/shortcodes/changelog-table-row.html
+++ b/layouts/shortcodes/changelog-table-row.html
@@ -16,13 +16,15 @@
 
 {{ $showChecksum  := .Get "showChecksum" }}
 
+<!-- 2.23.0 (2021-03-17) is first arm64 build -->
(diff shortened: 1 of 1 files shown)


thoward (approved):
Looks great! Thanks for the contribution.

PR #12490, opened 2024-08-09
Fix heading attribute

### LINKED ISSUE #12488: Headers missing from right nav
One of the H2s on this page is missing from the right nav - but the others are all there: https://www.pulumi.com/docs/concepts/secrets/

I do note that it has an odd search.keywords block at the end that I've never seen before - perhaps that is what is breaking this?

```md
## Committing configuration to source control {# search.keywords="checking version control"}
```

Image

cnunciato:
The syntax here is incorrect and is preventing this heading from appearing in both the right-hand nav and in search results. This should fix it up.

Fixes #12488.


dirien (approved):
LGTM!

Issue #12506, opened 2024-08-12
Investigate client-side navigation


cnunciato:
For various reasons -- speed of navigation, persistence of state, etc. -- it'd be interesting to consider moving to more of an SPA-like navigation model in which clicks within the domain don't require a full-page refresh. There are a bunch of unknowns with an approach like this (How would existing JS behavior be affected? Analytics? Releases?), but it's worth an experiment nonetheless, as it could unlock integrations that are fundamentally easier when the DOM isn't constantly being rebuilt between pageviews.

PR #12508, opened 2024-08-12
Enable SPA-like navigation with Turbo

cnunciato:
Adds [Hotwire Turbo](https://turbo.hotwired.dev) to enable SPA-like navigation. For #12506.

Observations:

On page load, things generally seem to work, including page-to-page navigation. However, on navigation:

- [x] Docs TOC items are no longer expandable
- [x] Search disappears
- [x] Feedback widget disappears
- [x] The on-this-page menu disappears
- [x] Headings no longer get anchor links
- [x] Resource-page filtering stops working
- [x] Sticky navigation no longer sticks
- [x] Copy buttons disappear
- [x] Expand-contract widget stops working
- [x] Logo nav stops working
- [x] Pricing-page tabs stop working
- [ ] Resource tabs don't reflect selection state
- [ ] tf2pulumi converters stop working
- [ ] CSS things look _very_ funky in some places -- huge page headings, buttons that get the wrong styles, etc.; I suspect Turbo's `link` ordering may not be deterministic

When clicking the back (or forward) button:

- [ ] The search box gets duplicated
- [x] Search stops working
- [ ] Get started button gets duplicated
- [ ] Chooser tabs get duplicated
- [ ] Pretty much all Stencil components get duplicated, in fact 😭
    image


layouts/partials/docs-top-nav.html
```diff
@@ -10,7 +10,7 @@
                     <li role="none"><a role="menuitem" href="/">Pulumi home</a></li>
                     <li role="none"><a role="menuitem" href="/pricing/">Pricing</a></li>
                     <li role="none"><a role="menuitem" href="/blog/">Blog</a></li>
-                    <li role="none"><a role="menuitem" href="/resources/">Events & workshops</a></li>
+                    <li role="none"><a role="menuitem" href="/resources/">Events & worksdhops</a></li>
```
ericrudder, line 13:
```suggestion
                    <li role="none"><a role="menuitem" href="/resources/">Events & worksdhops</a></li>
```

is this *really* a change you want??

cnunciato: Ha! It is not. Good catch. 😉

cnunciato:
After a day or so trying to get this to work, I've made some good progress, but am going to set it aside for a bit. Here's where things stand as of now.

I was able to address most of the out-of-the-box issues I saw (listed above in the PR description) by refactoring our existing page-load event listeners to use Turbo's `turbo:load` event instead, along with a few minor hacks. This mostly went fine, and led me to think this might work out after all, at least for now. However, I've now hit a wall that I'm not sure I'll be able to get beyond without a potentially big refactor.

The issue is with our web components -- chooser, choosable, the converter, the pricing calculator, etc. Most of these components (32 of 38 in all) are intentionally written to be rendered into to the surrounding DOM, mainly for styling, to take advantage of the Tailwind styles we configure globally, general scriptability, and to keep the overall size of the component bundle down. For reasons that seem to have to do with how it manages its own cache, Turbo doesn't appear to "play nicely" with components written like this; when clicking around in the usual way, things generally work as expected, but when using the back and forward buttons, things go sideways, causing components to be duplicated:

https://github.com/user-attachments/assets/825e0475-01ab-41b3-bef3-533d0987e63b

Configuring the components to use the shadow DOM instead, however (i.e., using `shadow: true` instead of `false`, which is Stencil's default), seems to "fix" the duplication issue -- although it also unfortunately breaks the look/feel and behavior of the components themselves, so won't work by itself.  

Why Turbo treats these elements differently from others is unclear, but given we aren't doing anything non-standard, here, I'm assuming it's just a bug in Turbo. We _could_, in theory, rewrite all of these components to use the shadow DOM instead (which some of our components do) -- but this would of course be a potentially large undertaking and one that'd be difficult to estimate without more exploration. All up, this feels like a showstopper to me, but I'm open to keeping at it if it's something we want to pursue.

I did also notice something else we should be aware of. The one feature of Turbo we absolutely need (without which this whole effort wouldn't make sense, beyond just making the UI feel snappier) is to persist a DOM element across pageviews. This is done using a new Turbo feature called [`data-turbo-permanent`](https://turbo.hotwired.dev/handbook/building#persisting-elements-across-page-loads). Setting this attribute on any DOM node adds the node to Turbo's cache (which is how it's persisted). To get a feel for it, I added this attribute to a new node that I've placed in the docs sidebar with a timestamp showing that it does, indeed, persist, but it's not perfect. As you'll see, clicking around within the docs section itself seems to be fine, but once you click out of this section and into something else, like Tutorials, or back to the home page, into products, etc., and then back, eventually, you'll see that it's recycled. Perhaps not a huge deal, given it gives us a lot more persistence than we have today (which is none), but seems worth calling out, considering it's the single pin on which the whole structure relies.

content/tutorials/abstraction-encapsulation/_index.md
```diff
@@ -20,6 +20,7 @@ prereqs:
     - A [Docker](https://docs.docker.com/get-docker/) service installed and running locally
 collections:
     - learn-pulumi
+    - aws
```
cnunciato, line 23: Unrelated changes here and below, will remove.

sean1588:
I have a couple general concerns here:

1. How this could affect us down the road if we wanted to introduce new things to the docs site that involve dynamic elements and if this would end up hindering our options in the future if turbo happens to not play nicely with them.
2. Will this make us more susceptible to accidentally contributing breaking changes if contributors are not aware/forget special considerations that need to be taken into account when doing certain things on the site? Yes of course I am aware we generate previews and do some testing as part of the review process, but I am concerned if there are side effects that end up being caused that may be less obvious from a review.

> Setting this attribute on any DOM node adds the node to Turbo's cache (which is how it's persisted). To get a feel for it, I added this attribute to a new node that I've placed in the docs sidebar with a timestamp showing that it does, indeed, persist, but it's not perfect. As you'll see, clicking around within the docs section itself seems to be fine, but once you click out of this section and into something else, like Tutorials, or back to the home page, into products, etc., and then back, eventually, you'll see that it's recycled.

I am guessing this is likely due to the way the DOM might be structured across different sections of the site. There is a good chance the nested structure of the sidebar might be slightly different as to where it lives in the DOM across different sections of the site and I'm assuming if the parent element(s) the sidebar lives in are of a different structure on the page it is navigating to it probably yanks it from the persisted cache which would make sense. If this is actually the case, this does seem to make things more fragile (i.e. someone innocently makes a slight alteration to DOM structure on a certain page and accidentally affects the behavior of elements that were meant to be persisted but now are being recycled or vice-versa and ends up resulting in adverse affects).

cnunciato:
>How this could affect us down the road ... hindering our options
>Will this make us more susceptible to accidentally contributing breaking changes

Definitely big concerns, and I share them -- not just for how they affect this repo, but for any web project that may want to do this sort of thing going forward (and I'm sure there'll be others). It honestly makes me extremely nervous hanging so much on a single library like this, but I wanted to give it a shot anyway to uncover exactly these kinds of issues.

If all we had to do was use `turbo:load` instead of `DOMContentLoaded`, IMO, it'd be more of a two-way door and arguably worth trying, as it'd be easy to undo. But having to make what could be deep changes to make this work starts to make the choice a lot less easy. Hence my thinking it's probably a good time to pause.

cnunciato:
See https://github.com/pulumi/docs/pull/12542 for a follow-up that uses vanilla JS instead of Turbo to get around the back/forward button issues.

cnunciato:
Closing as we won't be doing this.

PR #12514, opened 2024-08-14
Remove reference to AZURE_KEYVAULT_AUTH_VIA_CLI

devnev:
### Proposed changes

Update docs on how the azurekeyvault secret provider authenticates. Specifically, removed the reference to AZURE_KEYVAULT_AUTH_VIA_CLI, and referenced the bit of docs that lists the various methods that are attempted.

AZURE_KEYVAULT_AUTH_VIA_CLI was removed in https://github.com/google/go-cloud/pull/3202


devnev:
I'm also wondering if I should include in the docs how the default credentials were tripping me up; I had AZURE_CLIENT_ID defined, which caused the ManagedIdentityCredential to fail with an error that the azidentity packages considers fatal, which means it never even tried the CLI credentials. This is not well documented, I only found it out via https://github.com/Azure/azure-sdk-for-go/issues/23113

thoward (approved):
Thanks so much for the contribution @devnev !

I reviewed the underlying code changes in `gocloud` and `azidentity` and verified that the environment variable is gone from `go-cloud` and that the versions of these libraries that we depend on (0.37.0/1.6.0) include the new approach and no longer use this environment variable.

I agree that the new implementation is a bit confusing now, and could use more clarification. Would love another docs PR on the topic if you have time to write it up!

PR #12528, opened 2024-08-16
Docs for new ESC aws parameter store provider

cleverguy25:
### Proposed changes

Docs for new ESC aws parameter store provider


thoward (approved):
Merged with master into new directory/left nav structure.

PR #12530, opened 2024-08-16
Add Managing Configuration and Secrets Tutorial

toriancrane:
### Proposed changes

This PR adds the content for the Managing Configuration and Secrets tutorial.

- [ ] Add meta image


content/tutorials/managing-config-and-secrets/index.md
```diff
@@ -0,0 +1,194 @@
...
+# An image for the tutorial. It appears on tutorial page and in social-media previews.
+meta_image: config-secrets-meta.png
+
+# The order in which the tutorial appears in most lists. Order is ascending, so higher numbers
+# mean the tutorial will appear further down the list. Positive integers only.
+weight: 999
+
+# A brief summary of the tutorial. It appears at the top of the tutorial page. Markdown is fine.
+summary: |
+    In Pulumi, an environment is a collection of values intended to capture the configuration values needed to work with a particular environment. These can be raw values like server names, environment types, region names and so on. They can also be sensitive values such as database passwords or service tokens.
```
thoward, line 21: Suggest rephrasing not to use the word "environment". The term `environment` here is confusing because that's an ESC concept that we're not actually using in this tutorial. It also conflates against the more common concept of environment variables. Later in the doc we also specify a config value called `myEnvironment` which is yet another meaning of the word (e.g. "production environment").

content/tutorials/managing-config-and-secrets/index.md
```diff
@@ -0,0 +1,194 @@
...
+    
+    In this tutorial, we'll demonstrate how to create and utilize configuration and secret values in Pulumi.
+
+# A list of three to five things the reader will have learned by the end of the tutorial.
+youll_learn:
+    - How to create a raw configuration value
```
thoward, line 27: Suggest using the term "plain-text" instead of "raw".

content/tutorials/managing-config-and-secrets/index.md
```diff
@@ -0,0 +1,194 @@
...
+    - How to create a secret configuration value
```
thoward, line 28: Suggest phrasing this as "an encrypted secret" vs "secret configuration value".

content/tutorials/managing-config-and-secrets/index.md
```diff
@@ -0,0 +1,194 @@
...
+
+# The estimated time, in minutes, for new users to complete the topic.
+estimated_time: 10
+
+# # An optional list of collections this tutorial should be belong to. Collections are defined in data/tutorials/collections.yaml.
+# collections:
+#     - some-non-existent-collection
+---
+
+## Create a new project
```
thoward, line 46: In the "Inspecting Infrastructure" tutorial, we refer to this step as optional (e.g. either use an existing project or create a new one). That said, tagging this as optional in that tutorial, especially as the first step, felt a little awkward. I think we should be consistent across the tutorials and introduce this step as "Next, either create a new project or navigate to the code for an existing one. To create a project...".

content/tutorials/managing-config-and-secrets/index.md
```diff
@@ -0,0 +1,194 @@
...
+Your new project is ready to go!
+
+To perform an initial deployment, run `pulumi up`
+```
+
+This will create simple Pulumi program without any resources or configuration details.
+
+## Create configuration values
+
+In a Pulumi project, you can locally [store and retrieve configuration values](/docs/concepts/config/) using the `pulumi config set <key> [value]` command. To demonstrate, run the following command to create a configuration value with a key of `myEnvironment` and a value of `development`:
```
thoward, line 85: See previous comment. I think we should use a different example configuration value other than the overloaded term "environment".

content/tutorials/managing-config-and-secrets/index.md
```diff
@@ -0,0 +1,194 @@
...
+To retrieve a specific configuration value, run the `pulumi config get <key>` command as shown below:
+
+```bash
+$ pulumi config get myEnvironment
+development
+```
+
+## Create secret values
+
+Pulumi Cloud always transmits and stores entire state files securely. Additionally, Pulumi supports encrypting specific values as “secrets” for extra protection. Encryption ensures that these values never appear as plain-text in your state file. By default, the encryption method uses automatic, per-stack encryption keys provided by Pulumi Cloud, but you can also use a [provider of your own choosing](/docs/concepts/secrets/#configuring-secrets-encryption) instead.
```
thoward, line 118: Minor rephrase: "Pulumi supports encrypting specific values as “secrets” for extra protection. Pulumi Cloud transmits and stores state files over a secure connection, but once received the files are viewable in plain-text. By encrypting secrets, Pulumi ensures that these values never appear as plain-text in your state file. By default, the encryption method uses automatic, per-stack encryption keys provided by Pulumi Cloud, but you can also use a [provider of your own choosing](/docs/concepts/secrets/#configuring-secrets-encryption) instead."

content/tutorials/managing-config-and-secrets/index.md
```diff
@@ -0,0 +1,194 @@
...
+
+You can see that the configuration values were successfully imported into the program and exported as outputs, and you can see that the value of `Password`, which comes from the secret configuration value `myPassword`, is still not visible.
+
+## Clean Up
+
+{{< cleanup >}}
+
+## Next Steps
+
+In this tutorial, you created raw and secret configuration values in your project's environment. You also accessed these values via your Pulumi program code.
```
thoward, line 189: IIRC these aren't "project level" config values, but rather are stack-level?

content/tutorials/managing-config-and-secrets/meta.png
thoward, line 1: Should we delete this meta image, since we're using `config-secrets-meta.png`?

thoward (changes requested):

content/tutorials/managing-config-and-secrets/index.md
toriancrane: On this one I will keep as is. The example program code being leveraged is also being used in another tutorial and I would basically have to duplicate this code just to change `myEnvironment` to something else. I have however updated the intro paragraph to make this less confusing.

content/tutorials/managing-config-and-secrets/index.md
toriancrane: I almost wonder if we should just keep it simple rather than adding the indication that it can be run against existing code.

thoward (approved):

thoward:
LGTM!

PR #12540, opened 2024-08-16
Pointing to the last version before it was cleaned up

ringods:
### Proposed changes

Fixes a link in an older blog article which rendered a 404.


cnunciato (approved):
Thanks!

ericrudder:
I'm not 100% sure if this is really the intended (or correct fix) ... do you mind taking another look please?

PR #12542, opened 2024-08-16
Enable SPA-like navigation with vanilla JS

cnunciato:
Follow-up to #12508 (for #12506), this uses vanilla JS rather than Turbo.

Still a few minor kinks in this, but it does the same thing as Turbo -- hooks navigation events (specifically clicks), fetches the requested page, and replaces `head` and `body` content in place -- but doesn't exhibit the same back/forward button issues we've seen [with Turbo](https://github.com/pulumi/docs/pull/12508#issuecomment-2289973305).

Both of these solutions suffer from the same fundamental limitation, though, in that they only work with user-initiated link clicks that bubble from direct descendants of `document`. Any sort of navigation that happens outside of a clicked link -- setting `document.location`, for example, which is common both in our own code and in third-party libraries, such as Algolia's search-results widget, which we use for our in-site search -- will necessarily trigger full-page reloads that can't be intercepted. The experimental [Navigation API](https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API) promises to make it possible to catch and redirect any navigation event, including script-initiated, but as of now, it's [only supported in Chrome](https://caniuse.com/mdn-api_navigation).

I think this PR proves the concept of SPA-like navigation for pulumi.com as technically doable, with some additional work to address all the edge cases -- and would be my preferred approach over a third-party lib, as it's code we'd be able to maintain ourselves (and actually isn't all that complicated). However, given the fundamental browser-level technical limitations, we should consider the overall approach of SPA-like navigation more of a UX optimization that'll sometimes work and sometimes won't, rather than a concrete, reliable solution to the problem of DOM-node persistence.

cc @AaronFriel @cleverguy25 @lukehoban @sean1588 @interurban


AaronFriel:
@cnunciato can we hook Algolia and other `document.location` setting via [beforeunload](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) events?

cnunciato:
@AaronFriel Alas, no (this was actually the first thing I tried), for two reasons -- one, that `beforeunload` only blocks navigation when you call `preventDefault()`, which prompts the user to confirm them want to leave the page (it's intended to [prevent data loss](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event)), and two, the destination URL isn't published with the event, which means we wouldn't know where the user was headed in order to redirect them ourselves in-page.

cnunciato:
Closing as we won't be doing this.

PR #12556, opened 2024-08-19
Document org policy violations REST API

lblackstone:
### Related issues (optional)
Fix https://github.com/pulumi/pulumi-service/issues/21563


tyler-dunkel:
This needs to be updated to include the `urn` as that will be part of the api response.

lblackstone:
> This needs to be updated to include the `urn` as that will be part of the api response.

Updated

cnunciato (approved):

PR #12565, opened 2024-08-20
Experiment with different docs nav structure

joeduffy:
This splits out IaC, ESC, and Cloud into their own independent top level nav structures. This simplifies and focuses the left nav for each one of these. There are some gray areas, like conceptual topics in IaC (or Cloud) that do apply to one or the other, but I personally find this to be way more usable than the current mammoth combined left nav. As you go between the products, the left nav drops the "Pulumi IaC", "Pulumi ESC", and "Pulumi Cloud" headers in the nav since they're superfluous given the top level tabs.

This is not final and we've certainly not built consensus as a team on this yet -- just pushing the branch so we have something to discuss!

### DIFF
--- a/content/docs/cli/_index.md
+++ b/content/docs/cli/_index.md
@@ -6,9 +6,9 @@ h1: Pulumi CLI overview
 no_on_this_page: true
 meta_image: /images/docs/meta-images/docs-meta.png
 menu:
-  cli:
-    name: Overview
-    weight: 3
+  pulumiiac:
+    identifier: cli
+    weight: 8
 aliases:
 - /docs/reference/commands/
 - /docs/tour/basics-deploying/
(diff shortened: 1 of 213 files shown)


content/docs/cli/_index.md
```diff
@@ -6,9 +6,9 @@ h1: Pulumi CLI overview
 no_on_this_page: true
 meta_image: /images/docs/meta-images/docs-meta.png
 menu:
-  cli:
-    name: Overview
-    weight: 3
+  pulumiiac:
```
joeduffy, line 9: I don't love `pulumiiac`, but I am following the convention we use elsewhere (`pulumiesc` and `pulumicloud`). The alternative would be to go with `pulumi_iac` (and rename `pulumi_esc` and `pulumi_cloud`) which makes my stomach hurt a little less. Happy to go with what the crowd prefers here!

joeduffy:
NOTE: If we do this, we'll presumably want a different /docs landing page that is inherently more "multi product" and use some permutation of the current /docs/iac page. I haven't done that part, just wanted to get the basic idea out there.

joeduffy:
This work ultimately happened in https://github.com/pulumi/docs/pull/12693.

PR #12613, opened 2024-08-28
ESC environment syntax reference: Add a multiline value example.

MitchellGerdisch:
### Proposed changes

This change is to make it easier to discover that the YAML pipe (i.e. "|") can be used to create values that are multiline text.

- Added a basic example for creating a multiline text value.
- Added another multiline example that uses `fn::secret` for an SSH key.
- Also noticed that the stack reference example had an indent issue, so fixed that.

### DIFF
--- a/content/docs/esc/reference.md
+++ b/content/docs/esc/reference.md
@@ -54,6 +54,12 @@ values:
     # Array elements are "app.items[0]" and "app.items[1]"
     items: [ "config-a", "config-b" ]
 
+    # Path is "app.multiline"
+    # If the value needs to be a multiline value, use YAML pipe.
+    multilineValue: |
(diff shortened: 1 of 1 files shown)


komalali (approved):

content/docs/esc/reference.md
```diff
@@ -54,6 +54,12 @@ values:
     # Array elements are "app.items[0]" and "app.items[1]"
     items: [ "config-a", "config-b" ]
 
+    # Path is "app.multiline"
+    # If the value needs to be a multiline value, use YAML pipe.
```
EronWright, line 58: You might say "use a YAML block" and provide a link to https://yaml-multiline.info/

PR #12636, opened 2024-08-30
Update customtimeouts.md, d(day) duration unit is not supported, remove from docs for clarity

brajshoriegzon:
### Proposed changes

This PR removes the d(day) duration unit from the https://www.pulumi.com/docs/concepts/options/customtimeouts/ page because the day unit is not supported. Replace 1d(1 day) with 12h(12 hours) since the h(hours) duration unit is supported and that will better reflect the supported units in the documentation


thoward:
Nice catch @brajshoriegzon ! Really appreciate the contribution. I verified this by inspecting the code here: https://github.com/pulumi/pulumi/blob/afac93f70cde7c89eb8c5820b490c917f540ef2e/proto/pulumi/provider.proto#L777 ... which confirms the supported input duration strings.

thoward (approved):

PR #12689, opened 2024-09-05
New, experimental layout/styling

joeduffy:
As part of the run up to PulumiUP and overall rebranding, we are exploring slightly different layout/styling. The goal is to make our website look more professional, modern, and enterprise-grade. We will eventually be tackling a more ambitious project in Q4, but this will at least serve as a stopgap for a little while.

The overall changes include:

* Removing excessive background iconography which makes thing seem overall busy and whimsical. That includes on the homepage as well as the "dotted headers."

* Doing so reveals that we waster a ton of space, especially in the page hero. It is not uncommon to burn 100 pixels of dead space before hitting "the fold" which is unfortunate. So the headers are generally tightened up as well.

* This change also reveals that we effectively use *two* textual headers on most pages. We should generally move away from this practice in favor of a tighter presentation of the facts.

* Because there is a lot more "white", I've added some light gray borders in some places. I've also moved away from overly shadowy borders in some places, since those add to the busy look.

* Tighten up some font sizes that were getting out of hand.

* Fix our buttons so they aren't so rounded; this added to the overly whimsical look. Most other comparative SaaS companies like Snowflake, Datadog, and Cloudflare use more squared off button styling, which is what I've adopted. This is site-wide.

I am sure there's a lot more we can do, but I'm pushing this to get feedback and see if there's any hope of a quick land!


joeduffy:
Before:

image

After:

image

cnunciato:
Is it okay to close this?

PR #12693, opened 2024-09-06
Reorganize the docs into product-specific sections

cnunciato:
This PR breaks our existing docs IA and L3 navigation into three major buckets: Pulumi IaC, Pulumi ESC, and Pulumi Cloud. It adds three new items to the L2 navigation, one for each product, adjusts the docs home page to serve as an entrypoint for each, adds landing pages where they don't currently exist, and replaces the L3 nav with a new component that's more declarative and easier to use (while still being driven by the Hugo menu system). 

Also adds a script to the PR build workflow to ensure that any content that moves in a given PR (e.g., this one, which relocates just about every page of the docs) contains a proper server-side (i.e., S3) redirect to prevent 404s on release. 🚀


content/docs/iac/_index.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Pulumi IaC
```
aaronkao, line 2: should the URL spell out "infrastructure-as-code"?

cnunciato: No. The product "Pulumi IaC" (that's what'll be in the nav pretty much everywhere), so we're going with `iac` in the URL as well. We can add the full phrase to descriptions and headings where it makes sense, though.

cnunciato:
image

sean1588 (approved):

content/docs/pulumi-cloud/reference/_index.md
```diff
@@ -1,13 +1,14 @@
 ---
-title: Reference
+title: Pulumi Cloud Reference
 title_tag: "Pulumi Cloud: Reference"
-meta_desc: This is a meta description that's hopefully just long enough to appease the linter.
+meta_desc: Learn how to interact with Pulumi Cloud programmatically using the REST API for automation and integration.
 menu:
     cloud:
-        name: Reference
```
cnunciato: @interurban Why change this? I would like to keep this as **Reference** across the board (and I thought it's what we agreed on). "Pulumi Cloud" is redundant, as the whole section is about Pulumi Cloud. This will be the case everywhere.

layouts/partials/docs/special-pages/docs-home.html
```diff
@@ -3,7 +3,7 @@
 
 
 <section class="content docs-home">
-    <div id="docs-home-banner" class="header">
+    <div id="docs-home-banner" class="header bg-blue-100 bg-right">
```
cnunciato, line 6: 👍🏻

sean1588 (approved):

PR #12724, opened 2024-09-09
Updating timeline image

aaronkao:


interurban (approved):

lukehoban:
This seems to look "wider" and a little harder to see.  Is that intentional?

BEfore:
Screenshot 2024-09-10 at 12 45 45 PM

After:
Screenshot 2024-09-10 at 12 45 52 PM

The image is also a lot "smaller" than before - not sure that matters given it's an SVG - but might be related to the change in how this looks in the page?

PR #12740, opened 2024-09-10
[esc] update pulumi-cloud docs for projects

dschaller:
### Proposed changes

updates pulumi-cloud docs for ESC projects

### DIFF
--- a/content/docs/pulumi-cloud/cloud-rest-api/_index.md
+++ b/content/docs/pulumi-cloud/cloud-rest-api/_index.md
@@ -3251,16 +3251,12 @@ Kubernetes Pods should not be used directly. Instead, you may want to use a Depl
 
 ## Environments
 
-{{< notes >}}
-Pulumi ESC (Environments, Secrets, and Configuration) and its associated REST API endpoints are currently in public preview.
-{{< /notes >}}
-
 <!-- ###################################################################### -->
 
 ### List Environments available to the authenticated user.
 
 ```
-GET /api/preview/environments/{organization}
+GET /esc/environments/{organization}
 ```
 
 #### Parameters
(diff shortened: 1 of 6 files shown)


content/docs/pulumi-cloud/cloud-rest-api/_index.md
```diff
@@ -3251,16 +3251,12 @@ Kubernetes Pods should not be used directly. Instead, you may want to use a Depl
 
 ## Environments
 
-{{< notes >}}
-Pulumi ESC (Environments, Secrets, and Configuration) and its associated REST API endpoints are currently in public preview.
-{{< /notes >}}
```
komalali, line 3256: I think for this we're actually going to need to do something similar to https://www.pulumi.com/docs/pulumi-cloud/deployments/api/#migrating-from-the-preview-api and we'll also need to add the note about ESC GA and how the preview routes are being deprecated (but that should go out next Wednesday).

But the migration guide should be necessary coz I think we changed some of the payloads from the preview routes.

content/docs/pulumi-cloud/cloud-rest-api/_index.md
```diff
@@ -3303,15 +3313,21 @@ Status: 200 OK
 ### Create Environment
 
 ```
-POST /api/preview/environments/{organization}/{environment}
+POST /esc/environments/{organization}/
```
seanyeh, line 3316: np no trailing slash

dschaller: 🤦 you are right!

dschaller:
> OH! almost missed this. I think we still keep /api portion?

🤦 you are right!

PR #12800, opened 2024-09-16
changed CTA on the ESC page to docs

thtmnisamnstr:
Updated the CTA on the Pulumi ESC page to point to the ESC get started in docs.


arunkumar611 (approved):
LG

PR #12818, opened 2024-09-17
fix path for versions.md in pulumi-cli version update workflow

tgummerer:
There was some reorganization of the docs in https://github.com/pulumi/docs/pull/12693, which adjusted a bunch of paths.  This path here was also updated both in the workflow and in the repo, but not quite correctly.  Fix that now.

Fixes https://github.com/pulumi/docs/issues/12814
Fixes https://github.com/pulumi/pulumi/issues/17282

### DIFF
--- a/.github/workflows/pulumi-cli.yml
+++ b/.github/workflows/pulumi-cli.yml
@@ -99,7 +99,7 @@ jobs:
       - name: Update version lists
         run: |
           NL=$'\n'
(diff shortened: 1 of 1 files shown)


justinvp:
Looks like @cnunciato took care of it with https://github.com/pulumi/docs/pull/12823. I re-ran the workflow now for yesterday's release and it opened the PR: https://github.com/pulumi/docs/pull/12825

PR #12823, opened 2024-09-17
Fix CLI workflow path

cnunciato:
Looks like we got the path slightly wrong in yesterday's update.

Fixes https://github.com/pulumi/docs/issues/12814.


toriancrane (approved):
LGTM

PR #12829, opened 2024-09-17
Allow CLI docs PRs to be merged automatically

cnunciato:
It looks like the current CODEOWNERS setting (which requires a review from the docs team for anything under /docs) is preventing CLI docs PRs from being merged automatically. This should fix things up.


sean1588 (approved):
thanks @cnunciato !

PR #12840, opened 2024-09-18
Revamp the debugging page to point to our VS Code extension

mikhailshilkov:
### Proposed changes

Simplify the debugging guide by pointing users to our VS Code extension and `--attach-debugger` parameter instead of hand-crafted steps per language.


content/docs/iac/concepts/debugging/_index.md
```diff
@@ -20,288 +20,61 @@ Because Pulumi uses general purpose programming languages to provision cloud res
...
-Now you can step through the program and inspect variables. Once completed, resume the execution and let the Pulumi CLI complete.
-
-## Debugging YAML programs
-
-It's not possible to debug a YAML program with a step-by-step debugger because YAML is a configuration file, not an executable program.
+1. Start with an existing Pulumi project that contains `Pulumi.yaml` and your program.
+2. Open your IDE and set a breakpoint in your program.
+3. Run `pulumi up --attach-debugger` from a command line.
+4. Wait until `pulumi` is paused waiting for a debugger to attach.
+5. Attach the debugger to a running process of your language runtime.
```
tgummerer, line 75: Is it worth mentioning that the full debug configuration can be found in the event log, if the user passes `--event-log` (and enables that with PULUMI_DEBUG_COMMANDS=true)?  It's a bit more complicated though, so may not be worth it.

content/docs/iac/concepts/debugging/_index.md
```diff
@@ -20,288 +20,61 @@ Because Pulumi uses general purpose programming languages to provision cloud res
...
+Access the CLI output via the Debug Console view.
 
-Now you can step through the program and inspect variables. Once completed, resume the execution and let the Pulumi CLI complete.
+![Debug Console](./debug-console.png)
 
-## Debugging Go programs
+## Debugging with other IDEs
 
-To debug your Pulumi program, you need to override the way Pulumi compiles and runs the application:
+You can use any other IDE or editor that supports attaching to running programs. The tools differ depending on your preference and the language you are using, but the process stays similar:
```
tgummerer, line 69:
```suggestion
You can use any other IDE or editor that supports attaching to running programs using the [Debug Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol/)). The tools differ depending on your preference and the language you are using, but the process stays similar:
```

Maybe?

tgummerer (approved):
I'll let others comment on the VSCode extension parts since I don't know much about those.  Debugging using other IDEs looks good, I left a couple of minor suggestions.

content/docs/iac/concepts/debugging/_index.md
```diff
@@ -20,288 +20,61 @@ Because Pulumi uses general purpose programming languages to provision cloud res
...
 
-return await Pulumi.Deployment.RunAsync(async () => // make sure you have the async keyword here
-{
-    while (!Debugger.IsAttached)
-    {
-        // keep waiting until the debugger is attached
-        await System.Threading.Tasks.Task.Delay(1000);
-    }
+1. Select the __Run and Debug__ icon.
+2. Choose __Show all automatic debug configurations__.
```
cnunciato, line 43: We may want to mention that a debuggable file like `index.ts` should also be open. I happened to have `Pulumi.dev.yaml` open when I clicked Run and Debug, and I got an error explaining that I didn't have a YAML debugging extension.

content/docs/iac/concepts/debugging/_index.md
cnunciato, line 69:
```suggestion
You can use any other IDE or editor that supports attaching to running programs. The tools differ depending on your preference and the language you are using, but the process is similar:
```

cnunciato (approved):
This is great, thanks for adding it! Agree with Thomas's feedback and dropping an approval to keep you unblocked.

It does seem like a bit of a bummer, though, to have to lose some of these runtime-specific instructions. What do you think about keeping those (sans the VS Code-specific details of course)? Do you think it'd be valuable to users who aren't using VS Code? I kind of think it might be.

mikhailshilkov:
> It does seem like a bit of a bummer, though, to have to lose some of these runtime-specific instructions. What do you think about keeping those (sans the VS Code-specific details of course)? Do you think it'd be valuable to users who aren't using VS Code? I kind of think it might be.

@cnunciato Most of those are updated now, because we baked them into the `--attach-debugger` option. Now, users should never need to modify their programs or projects to get the debugging started. With VS Code and waiting for debugger hints out, I don't see much of language-specific content remaining. Do you?

PR #12843, opened 2024-09-18
Update README.md to add search guidance

interurban:
Adding back the Algolia search guidance from pulumi-hugo
https://github.com/pulumi/pulumi-hugo/blob/master/README.md?plain=1

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,24 +223,6 @@ title: My page
 ...
 block_external_search_index: true
 ```
-
-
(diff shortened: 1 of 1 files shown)


cnunciato (approved):
Looks good, thanks!

We can probably also remove the section on Migrating from pulumi-hugo now also, as it's been a good while since the migration. Happy to do that in a separate PR, but if you're up for it, seems like we might as well.

Issue #12559, opened 2024-08-20
Docs Tutorials - Clickable Cards and Responsiveness of the left nav


arunkumar611:
1. I struggled for a minute on why I couldn't click on the 3 tutorials and 3 topics. I thought they were all placeholders. But later figured out only the top text is clickable. We should make the entire card clickable. 
2. Responsiveness of the left nav needs an update as the black border overlaps with the cards when changing the window size. 

![Image](https://github.com/user-attachments/assets/7b793f43-4d26-41d6-81c6-beee85c31223)

PR #12847, opened 2024-09-18
ESC tutorials: add "Using ESC" collection and fixes some minor nav glitches/typos

thoward:
- adds a Tutorial collection "Using ESC" / `using-esc` and places two existing tutorials in it
- fixes Collection home pages to work with `single` type layouts vs only working with tutorials with subpages
- fixes ESC Kubernetes tutorial not showing correct left nav (due to using `_index.html` vs `index.html` filename)
- fixes mismatched descriptions and grammar typos in ESC chatapp tutorial
- fixes bad link in "More tutorials" section in ESC chatapp tutorial
- fix for overlapping left nav described in Issue #12559 (but not the clickable cards part of that issue).
- fix for breadcrumb nav showing id not display name on Tutorial Collections pages

Screenshot of bad left nav rendering because of `_index.html` vs `index.html`:

Screenshot 2024-09-18 at 12 09 27 PM


data/tutorials/collections.yaml
```diff
@@ -37,3 +37,8 @@
   type: proficiency
   name: Learn Pulumi
   description: Intro-level walkthroughs to help you learn core Pulumi infrastructure-as-code concepts.
+
+- id: using-esc
+  type: proficiency
```
cnunciato, line 42: I was sort of imagining that this collection would be of type `product` and be labeled just `esc`, such that the URL would then become https://pulumi.com/tutorials/esc (or alternatively `pulumi-esc`, if we wanted to be a bit more formal about it). This way, other `product`-type tutorials could come up alongside it (`iac`, `insights`, `cloud`) and things would generally be consistent. Any opposition to that? (That's not to say we can't have more collections later of course; just thinking it'd be good to start with product-specific buckets being the main ones initially, assuming users will most likely want to pivot on those primarily.)

```suggestion
- id: pulumi-esc
  type: product
```

content/tutorials/secure-kubernetes-app-secrets-with-esc/index.md
```diff
@@ -19,6 +19,9 @@ prereqs:
     - Install and configure [Kubectl CLI](https://kubernetes.io/docs/reference/kubectl/overview)
     - Install and configure [Minikube](https://minikube.sigs.k8s.io/docs/start)
 estimated_time: 15
+collections:
+    - using-esc
```
cnunciato, line 23:
```suggestion
    - pulumi-esc
```

content/tutorials/esc-managed-chatapp/index.md
```diff
@@ -48,8 +48,9 @@ prereqs:
 estimated_time: 15
 
 # # An optional list of collections this tutorial should be belong to. Collections are defined in data/tutorials/collections.yaml.
-# collections:
-#     - some-non-existent-collection
+collections:
+    - using-esc
```
cnunciato, line 52:
```suggestion
    - pulumi-esc
```

content/tutorials/secure-kubernetes-app-secrets-with-esc/index.md
```diff
@@ -19,6 +19,9 @@ prereqs:
...
+collections_weight: 1
```
cnunciato, line 24: These are both `1`. Did you want one to sort reliably before the other? (If so, suggest making that explicit.)

data/tutorials/collections.yaml
thoward: yeah, i mostly just cargo-culted the existing collection. I like the idea of making product-specific collections for each product and also including product-integration-style posts under a collection like "Pulumi as a Platform"  other topics as well as collections to cover "ingredients" like Crossguard/Deployments/etc.

What does changing the `type` do for the end result rendering?

content/tutorials/secure-kubernetes-app-secrets-with-esc/index.md
thoward: Good call. Wasn't worried about sorting, but probably a good idea to be more explicit about that.

data/tutorials/collections.yaml
cnunciato: >What does changing the type do for the end result rendering?

Nothing yet, but I'm thinking we might use that for grouping or coloring the tags somehow, etc., sometime down the road.

interurban (approved):
thx troy!

PR #12864, opened 2024-09-19
Fix PR closed job

sean1588:
Our PR closed jobs have not been cleaning up and of the buckets associated with their PRs and have been failing silently since it thinks there were no buckets to delete. 

I made some adjustments to the job to fix this as well as remove the dependency we had on the SSM parameter store. We used to use the store to persist each commit for a PR and then find buckets based on that. I have simplified this a bit to remove this dependency altogether as it doesn't look like this complexity is needed and instead just remove all buckets via `<origin_bucket_prefits>-pr-<pr-number>`, since we do want to delete all buckets associated with that PR anyway.

### DIFF
--- a/scripts/ci-pull-request-closed.sh
+++ b/scripts/ci-pull-request-closed.sh
@@ -22,11 +22,12 @@ if [[ "$GITHUB_EVENT_NAME" == "pull_request" && ! -z "$GITHUB_EVENT_PATH" ]]; th
         pr_comment_api_url="$(echo $event | jq -r ".pull_request._links.comments.href")"
 
         # List s3 buckets and filter all buckets associated with this PR. 
-        buckets=$(aws s3 ls | grep "$(origin_bucket_prefix)-pr-${pr_number}")
+        buckets=$(aws s3 ls | grep "$(origin_bucket_prefix)-pr-${pr_number}" | awk '{print $3}')
 
         if [ ! -z "$buckets" ]; then
             for bucket in $buckets; do
-                aws s3 rb "s3://${pr_bucket_name}" --force --region "$(aws_region)"
+                echo "removing s3://${bucket}..."
+                aws s3 rb "s3://${bucket}" --force --region "$(aws_region)"
             done
         else
             echo "No buckets found for PR: 'https://github.com/pulumi/docs/pulls/${pr_number}'"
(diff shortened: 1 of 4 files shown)


scripts/ci-pull-request-closed.sh
```diff
@@ -21,33 +21,17 @@ if [[ "$GITHUB_EVENT_NAME" == "pull_request" && ! -z "$GITHUB_EVENT_PATH" ]]; th
...
-
-            if [ ! -z "$pr_bucket_name" ]; then
-                echo "Found bucket ${pr_bucket_name} associated with commit ${commit}."
-                echo "Attempting to delete..."
-
-                # aws s3api head-bucket is a quick way to determine whether the bucket exists
-                # and we have access to it.
-                if aws s3api head-bucket --bucket "$pr_bucket_name" --region "$(aws_region)" 2>/dev/null; then
-                    aws s3 rb "s3://${pr_bucket_name}" --force --region "$(aws_region)"
-                    remove_param_for_commit "$(git_sha)" "$(aws_region)"
```
cnunciato, line 43: Does this mean we no longer need SSM params to map these? Should we include the change to remove setting the params if we're going to stop deleting (or trying to delete) them?

cnunciato: Yeah, if it's truly dead code, we should remove it.

That said, I'm still curious as to what actually stopped working here, as we went for so long time without any problems with it. Were you able to figure that out? (What you have here looks fine on the surface -- I'm just _fairly_ certain I did things like this for a reason, though admittedly it's been so long I no longer remember what it was.)

sean1588: Yeah I removed it. They were failing because the call to the ssm parameter store was failing for some odd reason. I then figured I would just remove it and simplify things since it doesn't look like this really adds too much value.

I figure this was done because the state of the world was much different at the time this was written. All the previews were created in the ci account and multiple repos were writing to it. So maybe extra care was taken to avoid the chance for collision if you were to only use PR numbers. Currently we are operating in the testing-website account and this and the private repo are the only two repos writing to it. If for some reason we add other repos down the road that are standing up buckets here, the bucket naming scheme will end up being different, so there is no chance for collision. The only potential here is PR numbers colliding with the private repo, but the PR number count in the private repo is still under 100 and here we are over 12K. So the numbers will _never_ collide because it is so far behind and we rarely use the repo by comparison. Anyways, I just do not see the value for the added complexity here.

cnunciato: >extra care was taken to avoid the chance for collision if you were to only use PR numbers

Yep, I believe that was it -- PR numbers weren't specific enough. Ok, seems reasonable enough. Thanks for jogging my memory!

cnunciato (approved):

Issue #12793, opened 2024-09-14
Audit for round buttons


joeduffy:
I'm not sure if it's intentional, but I am seeing some buttons that aren't using the new border radius:

![Image](https://github.com/user-attachments/assets/df07b728-4a21-4646-91a4-ac75bf577635)

PR #12895, opened 2024-09-21
Fix a few rounded buttons

joeduffy:
We missed a handful of round buttons in the homepage update, which this change addresses. We also omitted one of the styling changes in the original PR which was to make the CTA buttons less cartoonishly large. Wasn't sure if that was intentional, but this adds it back.

Fixes pulumi/docs#12793.


sean1588 (approved):
thanks @joeduffy! this looks good! I like the slightly smaller buttons. Sorry some of these styles got missed, they probably got lost in the mix of bringing the PRs together along with the additional changes we were making in the PR that we shipped. 

thanks for addressing it 🚀

PR #12905, opened 2024-09-23
Adds support for showing the Pulumi Copilot sidebar on the docs site

foot:
### Proposed changes

Adds support for showing the Pulumi Copilot sidebar on the docs site

#### Notes

- Adds a route to the copilot host so that the iframe is SAMEORIGIN with the docs
  - This allows it to `React.portal` into the doc and interact w/ the page a bit.
  - This requires some tweaks to SecurityPolicy to allow the Copilot path to be iframe'd too

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -60,7 +60,7 @@ const config = {
 };
 
 // const aiAppStack = new pulumi.StackReference('pulumi/pulumi-ai-app-infra/prod');
-const aiAppStack = new pulumi.StackReference('pulumi/pulumi-ai-app-infra/pr-pulumi-pulumi.ai-1254');
+const aiAppStack = new pulumi.StackReference('pulumi/pulumi-ai-app-infra/pr-pulumi-pulumi.ai-1242');
 const aiAppDomain = aiAppStack.requireOutput('aiAppDistributionDomain');
 const cloudAiAppDomain = aiAppStack.requireOutput('cloudAiAppDistributionDomain');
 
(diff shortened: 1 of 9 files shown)


infrastructure/index.ts
```diff
@@ -59,8 +59,10 @@ const config = {
     marketingPortalStack: stackConfig.get("marketingPortalStack"),
 };
 
-const aiAppStack = new pulumi.StackReference('pulumi/pulumi-ai-app-infra/prod');
+// const aiAppStack = new pulumi.StackReference('pulumi/pulumi-ai-app-infra/prod');
+const aiAppStack = new pulumi.StackReference('pulumi/pulumi-ai-app-infra/pr-pulumi-pulumi.ai-1242');
```
foot, line 63: revert this

infrastructure/index.ts
```diff
@@ -59,8 +59,10 @@ const config = {
     marketingPortalStack: stackConfig.get("marketingPortalStack"),
 };
 
-const aiAppStack = new pulumi.StackReference('pulumi/pulumi-ai-app-infra/prod');
+// const aiAppStack = new pulumi.StackReference('pulumi/pulumi-ai-app-infra/prod');
+const aiAppStack = new pulumi.StackReference('pulumi/pulumi-ai-app-infra/pr-pulumi-pulumi.ai-1297');
```
foot, line 63: Revert before merging

cnunciato (approved):
Looks great @foot! Leaving an approval to unblock, but two things before we merge:

* I'd love to get a 👍🏻 from @sean1588 as well, just so we're all on the same page with this
* We'll need to fix up those lines in `infrastructure/index.ts` once the `prod` stack is ready to go

infrastructure/index.ts
sean1588, line 63: can we make the stack reference itself a config value that is specified in the stack config, instead of hard coding it, since it appears to be environment specific?

unless we only have one environment for the ai stack that all docs stacks will reference then maybe this is ok

layouts/partials/copilot/sidebar.html
```diff
@@ -0,0 +1,5 @@
+{{ $copilotApiUrl := getenv "PULUMI_COPILOT_URL" }}
+{{ if $copilotApiUrl }}
+<aside id="ai-sidebar-target"></aside>
+<iframe id="ai-sidebar-host" src="/_pulumi/cloud-ai/sidebar?hostApp=docs&apiHost={{ $copilotApiUrl | urlquery }}"></iframe>
```
sean1588, line 4: just curious here, do we still want to show these clickable example prompts in the docs section of the site since they all seem cloud console specific to querying things about the org? I imagine this would need to be handled elsewhere though and not in this PR.

Screen Shot 2024-09-26 at 5 42 48 PM

sean1588 (approved):
These changes look good to me @foot! Nice work here!!

layouts/partials/copilot/sidebar.html
foot: Good one! Will make a follow up.

infrastructure/index.ts
foot: We'll only have a one env really. Its a bit hard to align all the CORS etc to get this going on the test previews for now.

But I still like the idea a lot. Will do a follow up PR.

PR #12952, opened 2024-09-26
Add GrowthBook and segment dev env

sicarul:
This PR will:

Add GrowthBook's snippet to perform A/B tests on the website
Add a segment development key to make sure segment also loads if it's not a production environment (Useful for testing tracking works outside of production)
Fix a data-track label on the dashboard button for consistency (Relevant to the first A/B test that will be deployed)


cypress/e2e/site.cy.js
```diff
@@ -17,9 +17,10 @@ describe("www.pulumi.com", () => {
         it("loads and applies JavaScript", () => {
             // Checking the carousel validates that the JS bundle was loaded and applied
             // (excluding Stencil components, which are bundled separately).
-            cy.get(".header-container")
+            cy.get(".header-container", {timeout: 30000})
                 .should("not.have.class", "is-pinned");
 
+            cy.wait(10000)
```
sean1588, line 23: are both the timeout and the wait necessary? Also, would prefer if these were shorter if possible. Do we need to wait as long as 30 seconds to load, I imagine 10 might be good enough? Not that this is a super big deal, but do not want to add unnecessary time to our deployment time if not necessary.

sicarul: I'll try with a shorter time as soon as i get back home, i wanted to avoid flaky tests, with an external library on CI it is more likely sometimes it'll take a bit longer depending on network

thtmnisamnstr (approved):

cypress/e2e/site.cy.js
sicarul: Fixed @sean1588 ! I reduced it to 6 seconds and it worked fine

PR #12999, opened 2024-10-01
Upgrade Hugo to latest (0.135.0)

cnunciato:


.github/workflows/build-and-deploy.yml
```diff
@@ -27,7 +27,7 @@ jobs:
 
       - uses: peaceiris/actions-hugo@v2
         with:
-          hugo-version: '0.126.0'
+          hugo-version: '0.135.0'
```
sean1588, line 30: nit: if you want to change to double quotes - only because you did it in the last PR.

sean1588 (approved):

.github/workflows/build-and-deploy.yml
cnunciato: Yeah -- I may actually see if I can apply Prettier auto-formatting on this stuff across the board. Will see if I can do that in short order here. :)

PR #13005, opened 2024-10-01
wrap text to avoid platypus overlap on docs banner

shughes26:
### Proposed changes
fixed the text wrap on the docs banners


interurban (approved):
thx!

PR #13009, opened 2024-10-02
Fixes Copilot position on Tutorials page

foot:
### Proposed changes

- Fixes Copilot position on the tutorials page
- Introduces a config variable for the copilot basePath to support local development


theme/src/scss/_copilot.scss
```diff
@@ -13,8 +13,8 @@
     // Popovers on site use z-index: 20
     z-index: 10;
 
-    // Default use up the full height of the viewport
-    top: 0;
+    // Default, make space for the navbar
+    top: calc(38px + 8px + 8px + 54px);
```
sean1588, line 17: Unless we find the breakdown of this calculation useful, should this just be written out explicitly instead as it is not clear what 38, 8, 8, 54 represents? i.e. `top: 108px`, since that is what this would resolve to.

sean1588 (dismissed):
Going to just approve this. From a code glance it all seems fine to me. If this is deployed on your dev stack, then would be great to check this out. But since this is still behind the feature flag, then I think its safe enough to go ahead and ship.

layouts/partials/copilot/sidebar.html
```diff
@@ -1,5 +1,6 @@
 {{ $copilotApiUrl := getenv "PULUMI_COPILOT_URL" }}
 {{ if $copilotApiUrl }}
+{{ $copilotBasePath := or (getenv "DEV_PULUMI_COPILOT_BASE_PATH") "/_pulumi/cloud-ai/" }}
 <aside id="ai-sidebar-target"></aside>
-<iframe id="ai-sidebar-host" src="/_pulumi/cloud-ai/sidebar?hostApp=docs&apiHost={{ $copilotApiUrl | urlquery }}"></iframe>
-{{ end }}
+<iframe id="ai-sidebar-host" src="{{ $copilotBasePath }}/sidebar?hostApp=docs&apiHost={{ $copilotApiUrl | urlquery }}"></iframe>
```
sean1588, line 5: just noticed this... I think you may end up with a double slash here if the `DEV_PULUMI_COPILOT_BASE_PATH` environment variable is not set, since in that case it would end up with `/_pulumi/cloud-ai//sidebar?hostApp=docs....`

theme/src/scss/_copilot.scss
foot: Good one.

PR #13015, opened 2024-10-02
Add new url for A/B testing Contact Us form and reduce dropdown options

sicarul:
This PR reduces the forms available to users to just 2 (General Questions + Sales)
image

And adds a new URL with the same content under /contact/b/ which we'll use to enable Clearbit auto-hide fields to test the impact of that change - This will make forms only ask for additional information like company name, job title, etc, if it can't automatically guess from the work email.

image

We'll auto-redirect to contact/b 50% of traffic using GrowthBook which is already present on this repository


sicarul:
Just realized i had an extra file changed that we didn't need, pushed the latest commit to remove it ☝️

shughes26 (approved):

PR #13020, opened 2024-10-02
add azure compliance pages

sean1588:
Add compliance pages for Azure


aaronkao:
When I click into an Azure page, the lede copy still says AWS. Can we update that all

PR #13030, opened 2024-10-03
Fix additional broken links

sicarul:
With this PR, i no longer see REF_NOT_FOUND warnings in the console, which are indicative of broken links.


sicarul:
Using link checker on the generated website to see if there's any broken links caused by the PR, will report back when done.

sicarul:
The script never finished, however i inspected partial logs and upon testing i found that i had broken the aliases 🤦 I'm force pushing an update because it was easier to fix the mess up by just removing any changes on the docs folder, the broken links are usually on other pages anyway.

cnunciato:
@sicarul FWIW, All that really needs to be done here are the changes to the `relref`s (to point to valid pages) and the error-level setting. The other (hundreds of) Markdown-link changes are nice, but unnecessry.

sicarul:
> @sicarul FWIW, All that really needs to be done here are the changes to the `relref`s (to point to valid pages) and the error-level setting. The other (hundreds of) Markdown-link changes are nice, because they skip redirection, but are technically unnecessry.

Agree, i should have made the replace all less encompassing, however are you saying we should close this PR and open a new one with that? Rolling back all of those seems like more effort than starting over.

sicarul:
> >are you saying we should close this PR and open a new one
> 
> 
> 
> No no, I just wanted to make sure you understood the extra effort of replacing all the Markdown links wasn't technically necessary (especially insofar as it might increase the potential for 404s). If you're comfortable with the changes, though, it's fine, and as I said, ultimately leaves us in a net-better place.

Yeah unfortunately i understood that too late in the process but i understand now!

sicarul:
image
Seems like we're clear now!!

cnunciato (approved):

PR #13052, opened 2024-10-07
Enable tracking on some code copy buttons on install paths

sicarul:
This PR adds tracking events when clicking on the clipboard icon besides the code blocks, only for the ones that have a data-track property attached. Attached it to install instructions on https://www.pulumi.com/docs/iac/download-install/ and the snippet used by pages such as https://www.pulumi.com/docs/iac/get-started/azure/begin/.

This is to understand how many people are following this funnel that exits the browser and continues in a CLI, since we think this is a very common path.


SaraDPH (approved):

sean1588 (approved):
It might make sense to wrap this up in a component where we can just pass it the data tracking props and the copy text and then we can reuse this stuff to track across the site  so can use that component wherever we want to do this. We can consider if that is worth doing in the future. Going to approve now though, so we can start collecting data around this.

sicarul:
Thanks @sean1588! That makes sense but might be a bit involved for me, i can review that change if it helps if we work on it. It seems the PR requires "docs" to approve it?

interurban (approved):

PR #13053, opened 2024-10-07
Add custom domain for Segment snippet

sicarul:
We have enabled analytics.pulumi.com to route our tracking data, this PR switches over completely to the new setup.


sean1588 (approved):
lgtm!

PR #13067, opened 2024-10-07
add redirect for reinvent 2021 page to current

shughes26:
### Proposed changes
redirect `/aws-reinvent-2021` to `reinvent`


aaronkao:
The file that is receiving the redirect needs to be deleted from hugo

aaronkao (approved):

PR #13072, opened 2024-10-08
Add cdnHost to cookie consent manager and add dev key for growthbook

sicarul:
This PR sets the cdn host for Segment's open source cookie consent to our custom domain to avoid the cookie consent from breaking our website

As a note, we're using an archived open source library and Segment recommends other approaches: https://segment.com/docs/privacy/consent-management/consent-faq/

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -234,11 +234,12 @@
     <script async defer src="https://buttons.github.io/buttons.js"></script>
 
     <script>
-        var segmentWriteKey;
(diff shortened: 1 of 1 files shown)


cnunciato (approved):
LGTM -- so the issue was just that `cdnHost` wasn't set?

sicarul:
> LGTM -- so the issue was just that `cdnHost` wasn't set?

Exactly! We had to set it to our custom domain since we switched out from segment's cdn

PR #13084, opened 2024-10-09
Update aws-secrets.md

jkodroff:
Add a quick comment that secret name or ARN both work, since this is unclear from the docs. Longer term, we should also add this to the detailed docs in the table below (but I didn't want to risk breaking the table - in a time crunch).


interurban (approved):
thx Josh

PR #13088, opened 2024-10-09
Adding "Reference Azure Resources Across Stacks" Tutorial Content

toriancrane:
### Proposed changes

This PR adds the content for the "Reference Azure Resources Across Stacks" tutorial. It also does the following:

- Renames the original Stack Outputs and References tutorial
- Updates the test.sh script to include configuration values for the automated testing on Azure and GCP


thoward (approved):
LGTM. Thanks, @toriancrane !

PR #13101, opened 2024-10-10
Create copilot Feature Flag with GrowthBook

sicarul:
* Migrates GrowthBook from HTML Snippet into Javascript SDK
* Creates component wrapping copilot that uses copilot-on-docs feature flag defined in GrowthBook - Will use this flag for A/B testing.


layouts/partials/head.html
```diff
@@ -3,6 +3,23 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0" />
 
+    <!-- GrowthBook / Segment keys -->
+    <script>
+        {{- if eq hugo.Environment "production" }}
+            var segmentWriteKey = "UK90Ofwacetj5VCPJ7cUgkbNcKLSHO3u";
+            var snippetName = "C8Y429BDEy/06ujkUhzfL.min.js";
+            window.growthbook_sdk_key="sdk-psPofGr6jFV2ja9O";
```
arturl, line 11: Keys in source code??

theme/stencil/src/components/copilot-sidebar/copilot-sidebar.tsx
```diff
@@ -0,0 +1,35 @@
...
+  // The copilotSrc is URL where the copilot sidebar is hosted
+  @Prop()
+  copilotSrc: string;
+
+  @State()
+  showSidebar = false;
+
+  
+  componentWillLoad() {
+    this.showSidebar = gb.isOn('copilot-on-docs');
```
arturl, line 19: Can you explain who sets this GB flag? ('copilot-on-docs')

layouts/partials/head.html
sicarul: I know it looks wonky, but it's common practice since these are frontend keys (Used by the frontend library) so they are always available to end users anyway; you can see this as a common practice with any library like Segment, Google Analytics, Mixpanel, etc.

theme/stencil/src/components/copilot-sidebar/copilot-sidebar.tsx
sicarul: This is a feature flag i set up on GrowthBook:
image

We can then link the feature flag to an experiment:
image

theme/stencil/src/util/util.ts
```diff
@@ -1,4 +1,33 @@
 import * as uuid from "uuid";
+import { GrowthBook } from "@growthbook/growthbook";
+
+// Initialize GrowthBook
+export const gb = new GrowthBook({
+    apiHost: "https://cdn.growthbook.io",
+    clientKey: (window as any).growthbook_sdk_key,
+    decryptionKey: (window as any).growthbook_decrypt_key,
+    enableDevMode: (window as any).growthbook_dev_mode,
```
foot, line 9: Neat. Could we move the actual key values into this file and skip adding in the `<head />` ?

sicarul: I don't have the hugo.Environment variable here, i tried to find a way but i don't know if i can detect the environment somehow and didn't find any examples

foot: I tried to poke into this.. it looks like we _build and commit_ the components, so I don't think there is an opportunity to do prod/dev key selection in the component code.

foot (commented):
Functionally this all looks good and works for me!

I'd just wait for @sean1588 and @cnunciato to TAL too.

arturl (approved):

PR #13103, opened 2024-10-11
Document Python output lifting's use of `__getattr__`

lunaris:
The Pulumi Python SDK lifts output properties by overriding the special `__getattr__` method. This is designed to improve the ergonomics of writing Pulumi programs in Python, but technically violates some of the contracts Python expects of this method, particularly in relation to `hasattr`. This commit warns of this potential gotcha when mentioning output lifting to give users a reference point should they bump into this issue.

Fixes pulumi/pulumi#16743

### DIFF
--- a/content/docs/iac/concepts/inputs-outputs/apply.md
+++ b/content/docs/iac/concepts/inputs-outputs/apply.md
@@ -865,6 +865,10 @@ record = aws.route53.Record('validation',
 ...
 ```
 
+{{< notes type="warning" >}}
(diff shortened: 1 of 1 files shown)


lunaris:
image

thoward:
> image

I'm guessing that this screenshot is here to illustrate that the note about Python-specific information is showing, even when Java is selected as the language? To fix that I'd suggest moving the note inside of the `choosable` for Python, after the code block, which would move it up above the paragraph instead of after it, but it would then only display it if Python was your selected language, which seems more appropriate.

thoward (approved):
LGTM, though I'd recommend moving the note into the `choosable` block for Python.

PR #13111, opened 2024-10-11
Improve cookie consent UX

sicarul:
The current UX for cookie consent is a bit broken because of the deprecated library we're using; this said, we can at least improve it substantially by:

* Hiding the banner if outside Europe (Users can still click on `Your Website Preferences` and change their preferences) - This still complies with CCPA, there is no requirement to put a cookie consent banner on CCPA. Almost all our revenue comes from the US (12M ARR out of 18M), it's also appearing over Copilot which makes it harder to discover it.
* Upgrade the snippet to the latest version 5.8.1 to incorporate the fixes they've added: https://github.com/segmentio/consent-manager/blob/master/CHANGELOG.md
* Set option shouldReload to false, to avoid the snippet trying to refresh the page


sicarul:
If this PR is approved, i'll send a similar one to /registry.

sean1588 (approved):

PR #13121, opened 2024-10-15
Fix segment consent reload - Load forked library

sicarul:
This PR loads the forked library built at https://github.com/segmentio/consent-manager which removes the page reload, and fixes a CSS bug introduced in a previous PR


sean1588 (approved):
thank you @sicarul

PR #13127, opened 2024-10-15
Add ignore rule for 404ing link in dbtcloud provider docs

thoward:
This URL is correct, but is a deeplink into the dbtcloud app and requires a login, hence failing the link checker.


interurban (approved):
thx troy

PR #13142, opened 2024-10-17
Rewrite how Pulumi works page to use BucketV2

t0yv0:
### Proposed changes

Bucket is going to be deprecated and BucketV2 is recommended. IN addition to that, bucket.acl is deprecated. Unfortunately these resources were selected to demonstrate how Pulumi works. The page itself does not need to concern itself with Bucket per se, but needs a good example. To avoid the difficulty with deprecated resources and attributes, this rewrite of how-pulumi-works uses BucketV2 and tags modification as the example, instead of using Bucket and acl.

### Related issues (optional)

Relates: https://github.com/pulumi/home/issues/3583

### DIFF
--- a/content/docs/iac/concepts/how-pulumi-works.md
+++ b/content/docs/iac/concepts/how-pulumi-works.md
@@ -63,6 +63,9 @@ Let's walk through a simple example. Suppose we have the following Pulumi progra
 {{% choosable language javascript %}}
 
 ```javascript
+"use strict";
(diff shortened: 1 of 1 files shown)


t0yv0:
Going through manual examples verification.

t0yv0:
Verified examples through `pulumi preview`. PTAL.

sean1588 (approved):
LGTM! Thanks so much @t0yv0 ❤️

Issue #13132, opened 2024-10-16
Consider updating typescript code to improve output declaration


after-ephemera:
[content/docs/iac/concepts/resources/dynamic-providers.md](https://www.pulumi.com/docs/iac/concepts/resources/dynamic-providers/)

It seems that per https://github.com/pulumi/docs/issues/11639 the `declare` keyword should be used to declare outputs on the provider so that they are properly exposed to consumers.

thoward:
Thanks for the report! I filed a PR for this https://github.com/pulumi/docs/pull/13146

thoward:
Per engineering (@justinvp), it seems like the `declare` keyword is not actually necessary. Closing as won't fix.

Zemnmez:
@justinvp can it be documented somewhere how the outputs discovery mechanism works? The docs don't cover it and the open issues seem to be wrong

Issue #11639, opened 2023-02-14
Dynamic Providers: Typescript: dynamic.Resource outputs are not available if defined on Resource class


awoimbee:
### What happened?

I tried to follow the docs to create a custom resource (to generate a password and its hash), but the outputs are always `undefined`. I figured out that this is a bug caused by the class member definitions, as removing them make it work.

### Expected Behavior

It should have worked on the first try. I closely followed https://www.pulumi.com/docs/intro/concepts/resources/dynamic-providers/#dynamic-resource-outputs.

### Steps to reproduce

Original code:
```typescript
import * as pulumi from "@pulumi/pulumi";
import { strict as assert } from "node:assert";
import * as crypto from "node:crypto";

const _hashedPasswordProvider: pulumi.dynamic.ResourceProvider = {
  async create (inputs: {length: number}) {
    const password = "dummy";
    const hashedPassword = "my-super-pass-hashed-with-unique-salt";

    return { id: crypto.randomUUID(), outs: { password, hashedPassword } };
  }
};

class HashedPasswordResource extends pulumi.dynamic.Resource {
  public readonly password!: pulumi.Output<string>;
  public readonly hashedPassword!: pulumi.Output<string>;

  constructor (name: string, props: {length: pulumi.Input<number>}, opts?: pulumi.CustomResourceOptions) {
    const innerOpts = pulumi.mergeOptions(opts, { additionalSecretOutputs: ["password"] });
    const innerProps = {password: undefined, hashedPassword: undefined, ...props};
    super(_hashedPasswordProvider, name, innerProps, innerOpts);
  }
}

function main() {
  const botUsername = "robotsatan";
  const botPassword = new HashedPasswordResource("robotsatan-password", { length: 30 });

  // kaboom
  assert.notEqual(botPassword.password, undefined);
  assert.notEqual(botPassword.hashedPassword, undefined);
}
```

Diff to make it work:

```diff
-  public readonly password!: pulumi.Output<string>;
-  public readonly hashedPassword!: pulumi.Output<string>;

-  const botPassword = new HashedPasswordResource("robotml-password", { length: 30 });
+  const botPassword = <any>new HashedPasswordResource("robotml-password", { length: 30 });
```

### Output of `pulumi about`

```
CLI
Version      3.54.0
Go Version   go1.20
Go Compiler  gc

Plugins
NAME    VERSION
nodejs  unknown

Host
OS       arch
Version  22.0.2
Arch     x86_64

[...]

PR #13146, opened 2024-10-17
Use `declare` keyword in TS example for dynamic resource outputs.

thoward:
Per: https://github.com/pulumi/docs/issues/11639 and https://github.com/pulumi/docs/issues/13132 ... this should be using `declare`.


thoward:
ping @justinvp . would love to get a technical review here.

justinvp (commented):
I'm not convinced this is necessary. None of our generated provider SDKs use `declare`. I've asked about it at https://github.com/pulumi/docs/issues/11639#issuecomment-2443446338

thoward:
Gonna close this until we figure out if its necessary or not. Can reopen if we determine it's needed.

PR #13152, opened 2024-10-18
Adding AWS Immersion Workshop

tejavar:
Adding https://github.com/pulumi/marketing/issues/1129

NOTE - need to add hubspot from and salesforce campaign ID


sean1588 (approved):
thank you @tejavar!

PR #13171, opened 2024-10-24
A/B: Terraform vs. Pulumi Page Redesign

adamgordonbell:
A/B Test Described here:
https://docs.google.com/document/d/1ThaeZjfMmQsO5vTs-G9Wixs0zMoX_uUAgwaHvDtLq18/

Old:
http://www-testing-pulumi-docs-origin-pr-13171-5a7896de.s3-website.us-west-2.amazonaws.com/docs/iac/concepts/vs/terraform/?variant-vs-terraform-docs-page=0

New:
http://www-testing-pulumi-docs-origin-pr-13171-5a7896de.s3-website.us-west-2.amazonaws.com/docs/iac/concepts/vs/terraform/?variant-vs-terraform-docs-page=1


theme/src/ts/terraform-compare.ts
```diff
@@ -0,0 +1,25 @@
+import { gb } from "../../stencil/src/util/util";
+
+function initTerraformCompare() {
+    console.log("Terraform compare module initialized");
```
sicarul, line 4: We should probably remove this console log on the final version.

I've checked the version built on http://www-testing-pulumi-docs-origin-pr-13171-7992718f.s3-website.us-west-2.amazonaws.com/docs/iac/concepts/vs/terraform/ and i don't see this event in the console 🤔

adamgordonbell: fixed!

theme/src/ts/misc.ts
```diff
@@ -90,7 +90,7 @@ function bindToggles(selector) {
     });
 }
 
-function generateOnThisPage() {
+export function generateOnThisPage() {
```
adamgordonbell, line 93: I had to export this, and then redraw the sidebar after the experiment was engaged, as menu items change.

assets/css/marketing.css
```diff
@@ -1,3 +1,6 @@
+/*!*********************************************************************************************************************************************************************************************************************!*\
+  !*** css ./node_modules/css-loader/dist/cjs.js??ruleSet[1].rules[1].use[1]!./node_modules/postcss-loader/dist/cjs.js!./node_modules/sass-loader/dist/cjs.js??ruleSet[1].rules[1].use[3]!./src/scss/_marketing.scss ***!
```
sicarul, line 2: I *think* this diff appears when you don't run make build before committing, i might be wrong.

theme/src/ts/misc.ts
```diff
@@ -102,6 +102,10 @@ function generateOnThisPage() {
 
         $("h2, h3").each(function () {
             var $el = $(this);
+            // Skip if this heading is inside a hidden element
```
adamgordonbell, line 105: No heading in sidebar if hidden A/B heading

sean1588 (approved):

theme/src/ts/misc.ts
adamgordonbell: There is an "on this page" sidebar. And it is showing elements that aren't visible on the page, because they are inside a hidden element. This fixes that.

sean1588: ahhh gotcha. they are showing up in the on this page side bar because they are hidden in the main content area of the page and we are just scraping the heading tags when we generate the on this page list. ok that makes sense!!

PR #13194, opened 2024-10-29
Add docs on Property Paths

seanyeh:
### Proposed changes

Add docs on Property Paths. we refer to them on several pages but we don't go into detail what exactly they are


content/docs/iac/concepts/miscellaneous/_index.md
IaroslavTitov, line 1: Should this be in IaC? Sounds to be like an ESC concept, but it's blurry since config uses the same thing I guess?

IaroslavTitov (approved):

content/docs/iac/concepts/miscellaneous/_index.md
seanyeh: It was initially an IAC concept so I'm leaving it there. it's really a general pulumi concept 🤔

dschaller: I think it's ok to keep it in IaC since the original code lives there and reference from esc but it would be good to get @interurban's opinion

pgavlin: Yeah it's a general Pulumi concept. We use it in at least `ignoreChanges` and `esc`. Pretty sure we use it in other parts of resource options as well.

thoward (approved):
LGTM! Thanks for filling in this gap in our docs.

content/docs/iac/concepts/miscellaneous/property-paths.md
```diff
@@ -0,0 +1,36 @@
...
+menu:
+    iac:
+        name: Property Paths
+        parent: miscellaneous-concepts
+        weight: 1
+---
+
+Property paths represent a key or series of keys that one can use to refer to elements in an object or array.
+The syntax, inspired by JavaScript, supports both dot notation and bracket notation.
+For keys that contain special characters, bracket notation is required.
```
pgavlin, line 16:
```suggestion
For keys that contain special characters (i.e. `[`, `]`, `"`, or `.`) or begin with a digit, bracket notation is required.
```

seanyeh:
@thoward i just added a commit to move one of the existing pages. mind giving a second look before I merge?

PR #13197, opened 2024-10-29
Concepts docs: Add some detail about the `stackConfigDir` attribute

EronWright:
Add some detail about the `stackConfigDir` attribute, since I went looking for such information to solve an issue.


thoward (approved):
Thanks for the update! LGTM

PR #13205, opened 2024-10-30
Fix broken link in webhooks docs

thoward:
Fixes broken link to Microsoft docs by updating URL to the correct domain.


interurban (approved):
TY!

PR #13206, opened 2024-10-30
Fix broken link to automation API example

thoward:
The code was updated to a new version and the location of the referenced example changed. Previous file was removed, leading to a dead link. This updated link points to the correct location per @EronWright (the developer responsible for the change).


EronWright (approved):
LGTM

PR #13208, opened 2024-10-30
Windows arm64 builds started a bit later - only show what we have.

smithrobs:
### Proposed changes

Windows arm64 builds started a bit later - only show what we have.

### Related issues

#12485


thoward (approved):
Looks good! Thanks for digging into that to get the date ranges right.

PR #13230, opened 2024-11-04
fix link in config docs

tgummerer:
It looks like this should be a link, but is currently missing some square brackets.  Add those to fix that.


interurban (approved):
thx !!

PR #13290, opened 2024-11-07
Add some link exemptions

thoward:
These links (two Gartner links and the customer website Dutchie) are consistently a `403` to our link checker, for some reason*. This is probably something to look into long-term, but for now, adding these exemptions because these links work fine. 

*I did a little digging and this seems related to the use of Cloudflare, and its firewall rules that detect bots and then 403 them intentionally. No real way around that.


sean1588 (approved):
thank you Troy!!

Issue #11619, opened 2023-05-04
Okta: Configuring the SCIM Connector doc needing update


kelsey-dirks:
### What happened?

Our docs say, "For HTTP HEADER, you will use a token from the [Pulumi Cloud](https://app.pulumi.com/) as the authorization bearer token. To generate a token, navigate to your org in the Pulumi Cloud, click on the Settings tab, and then click SAML SSO. Scroll down to the SCIM section and generate a new token if you have never generated one for your org, or regenerate it if you have already done so in the past."

However, the Pulumi Cloud doesn't generate a token. It instead tells the user to contact support. The doc will need to be updated to reflect our new process. 

### Expected Behavior

NA

### Steps to reproduce

The doc that needs updating: [https://www.pulumi.com/docs/guides/scim/okta/#configuring-the-scim-connector](https://www.pulumi.com/docs/guides/scim/okta/#configuring-the-scim-connector)

### Output of `pulumi about`

NA

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

yoejigu:
PR has been created to make updates to the documentation. Waiting for approval and merging

PR #13299, opened 2024-11-08
Okta SCIM page updates

yoejigu:
### Proposed changes

- Made the notice to reach out to Okta to enable SCIM option if it is missing more prominent 

- Split up Enabling group provisioning and configuring group provisioning into two sections

- Changed the link names to match the new section name

- Made the warning to ensure that all users are assigned the app in Okta before configuring Push groups more prominent 

### Related issues (optional)

https://github.com/pulumi/docs/issues/11619


theme/yarn.lock
```diff
@@ -197,6 +197,13 @@
   dependencies:
     purgecss "^4.0.3"
 
+"@growthbook/growthbook@^1.2.1":
```
thoward, line 200: In theory, your above markdown changes should not require a update to `
yarn.lock`. Any idea why this is here? Can we remove changes from the commit? (e.g. if we need to bump library versions, that should be done in a separate PR and investigated with depth separately from a content update)

thoward (approved):
Thanks so much, Yosef! This all looks great.

Can you describe how you tested this? Is it possible to record a short screen capture of you doing these steps, and showing that they work? These kinds of docs are hard to verify because we can't e.g. use a code testing framework to verify that this works as described.

theme/yarn.lock
yoejigu: Hey Troy, that change in the yarn.lock showed up after running the "make ensure" command so I remove that change if it shouldn't be included

yoejigu:
https://github.com/user-attachments/assets/93af15c6-e623-4642-8e9c-c63986a85214

Issue #13261, opened 2024-11-06
enable ci to test azure code examples


sean1588:
It looks like the tests for programs under static/programs was never configured to test azure examples. We need to get azure creds in the pipeline to be able to run these tests. I am temporarily skipping those tests for now, just to get them passing until we have azure access configured.

PR #13304, opened 2024-11-09
Enable testing for the azure examples

sean1588:
fixes: https://github.com/pulumi/docs/issues/13261

Enables testing for the azure examples. It turns out all we needed to do was set the environment variables based on our azure org level secrets. I also verified this works, see this run: https://github.com/pulumi/docs/actions/runs/11751276481/job/32741012915


thoward (approved):
LGTM. Thanks for getting this unblocked.

PR #13322, opened 2024-11-11
Update Automation API getting started code to use BucketV2

t0yv0:
This change updates the code examples in the `Get started with Automation API` page.

Instead of using aws.s3.Bucket the examples now use aws.s3.BucketV2 as is recommended for our users.

In addition the code example is simplified to achieve a minimal static website wtihout manually setting up policy.

Fixes https://github.com/pulumi/home/issues/3735


sean1588 (approved):
LGTM! thanks @t0yv0 !

PR #13356, opened 2024-11-15
Add in missing rest API docs for polcies and oath token exchange

cleverguy25:
Adds in missing REST API docs around policies and oauth token exchange

Fixes https://github.com/pulumi/pulumi-cloud-requests/issues/394
Fixes https://github.com/pulumi/pulumi-service/issues/23761


glena (approved):

thoward (approved):
LGTM. Thanks for the new docs, @cleverguy25 !

PR #13367, opened 2024-11-16
Fix small url issue in REST API docs

cleverguy25:
Fixes one small issue for list environments in REST API docs


thoward (approved):

thoward:
Nice catch! Thanks @cleverguy25 .

PR #13369, opened 2024-11-16
Add documentation for using OIDC with EKS

cleverguy25:
This documents how to use our OIDC provider with EKS, similar to our GKE docs.  This scenario is basic to show login via token exchange as this is page is part of Pulumi Cloud. Later we should build a full E2E scenario with a pod accessing an ESC environment's secrets and confic specific for ESC.

I also renamed in the left nav from Google Kubernetes Engine to Google GKE as the former wrapped and looked weird.  I also fixed some minor issues in the GKE doc.


thoward (approved):
LGTM.

interurban:
hey @cleverguy25 ; Are you ready to merge your PR?

PR #13390, opened 2024-11-19
Add Importing Azure Infrastructure Tutorial

toriancrane:
This PR does the following:

- Adds the content and example code for the Azure import tutorial
- Updates the existing "Importing Infrastructure" tutorial to be named "Importing AWS Infrastructure"
- Adds the old `/tutorials/importing-infrastructure` alias to the AWS Import page.
- Adds `resourceGroupName` config value to `test.sh`.


content/tutorials/importing-azure-infrastructure/index.md
```diff
@@ -0,0 +1,406 @@
...
+
+The `pulumi import` command also enables you to import resources in bulk for scenarios in which you need to bring multiple resources under management with Pulumi. To do so, you will need to create a JSON file that has all of the required information for each resource: a `type`, a desired `name`, and an `id`.
+
+To start, return to the Azure console and create two additional Storage Accounts. Next, return to your program folder and create a new file called `resources.json`. Inside of this file, copy and paste the following JSON object, making sure to replace the values of the `id` parameters with the actual values for your environment:
+
+```json
+{
+    "resources": [
+        {
+            "type": "aws:s3/bucket:Bucket",
```
thoward, line 242: I think this JSON file needs to be updated. It seems to be copy pasta from the AWS tutorial (wrong resource type, should be `azure-native:storage:StorageAccount`).

assets/css/bundle.css
thoward, line 1: In general, you should not need to re-generate/commit these files (`assets/*`) as part of content PR. When CSS/JS changes, and the bundles need to be updated, that should be part of the PR that changed the CSS/JS, not as part of content PR. This is not really your fault, as this should have bene done by whoever modified the CSS/JS causing the need for the re-render. I'll dig in and see if I can figure out when that happened.

Reasoning: It's mostly about separation of concerns (e.g. a PR should only do one thing), but also, I have no way of verifying if this is in sync w/ what's on the main branch or not. Could be up to date, but could also be a regression. These are basically opaque files and not really reviewable, unfortunately.

Please amend the branch to remove these commits (not the files). We can re-render these and merge in a different PR to ensure they are up to date w/ latest.


@sean1588 @shughes26  any idea how we can prevent this from happening in the future? I've seen this about 3-4 times in various PRs from different folks. Maybe in PR site build automation, we check we can see if the PR includes changes to the assets, and if so, check to make sure the commit includes a re-render of asset bundles?

thoward (changes requested):

content/tutorials/importing-azure-infrastructure/index.md
toriancrane: Good catch, thank you!

thoward (approved):

PR #13398, opened 2024-11-20
Update esc cli links

sean1588:
Updates the internal links to point directly to the new location of the esc cli docs so they do not need to go through a redirect via the old route. This appears to cause issues with our link checker inaccurately reporting broken links.


thoward (approved):
LGTM. thanks for fixing this!

PR #13423, opened 2024-11-22
Update CDN behaviors to route answers

sean1588:
Adds an origin and cache behaviors to cloudfront for /answers to route to cloudfront distribution for answers project.


thoward (approved):
Looks good! Excited to ship this next iteration.

Issue #13425, opened 2024-11-23
PR previews should run against prod stack


sean1588:
Pulumi previews that run as part of the PR job should preview against the www-production stack instead of the www-testing stack.

jkodroff:
@claude Still an issue or ok to close?

CamSoper:
@claude If left open, would #13452 have fixed this? What's the full blast radius?

CamSoper:
@claude I'm unable to reopen #13452. Please create a new branch prefixed `CamSoper/` and recreate the changes from #13452. Open a new PR in Ready state, and be sure to note it fixes this issue.

CamSoper:
Closing as **won't fix**.

Running the PR-job Pulumi preview against the `www-production` stack is not viable with the current PR workflow, and making it viable would require elevating PR CI to production-account credentials — real risk for little benefit. PR #19743 (a faithful re-creation of #13452) demonstrated the failure empirically.

### Why it fails

The PR workflow (`.github/workflows/pull-request.yml`) runs in `environment: testing` and authenticates to AWS by assuming the **testing** account role `arn:aws:iam::571684982431:role/ContinuousDelivery`. Selecting the production stack (`pulumi stack select pulumi/www-production`) changes only which **Pulumi config/state** the program uses — it does **not** change the **AWS credentials** it runs with.

The production stack points at production-account resources, e.g. `hostedZone: www.pulumi.com` (`Pulumi.www-production.yaml`) and an ACM cert in account `388588623842`. During preview the program makes live AWS data-source calls — `infrastructure/index.ts` runs `aws.route53.getZone({ name: "www.pulumi.com" })` for each domain alias. With testing-account credentials that zone doesn't exist, so the invoke throws and the program aborts:

```
error: invocation of aws:route53/getZone:getZone returned an error:
  * no matching Route 53 Hosted Zone found
make: *** [Makefile:107: ci_pull_request] Error 1
```

(Failing run: https://github.com/pulumi/docs/actions/runs/27702715468/job/81943126043?pr=19743)

This corrects the earlier "narrow and low-risk" assessment in this thread, which assumed the prod stack was "already accessible because `ci-login.sh` uses the same `PULUMI_ACCESS_TOKEN`." That conflates two different credentials. The Pulumi token authorizes selecting the stack and reading its StackReferences from Pulumi Cloud; it does not grant the **AWS** access the program needs. The program still calls AWS as the runner's IAM role — testing-only.

[...]

PR #13452, opened 2024-11-26
Run pulumi previews against the production stack

sean1588:
fixes: https://github.com/pulumi/docs/issues/13425

Runs the previews against the production stack. On PR workflows we provision the preview bucket in the testing account but infrastructure previews should always run against the production stack since that is the stack they will get deployed to when merged(not the www-testing stack).


meagancojocar:
closing as stale

PR #13465, opened 2024-12-01
[esc] add at scale doc

dschaller:
### Proposed changes

add doc on using ESC at scale, breaking down the steps based on user role, and providing examples of RBAC team access, JIT access of secrets/configuration, revision tags, providers, etc.


content/docs/esc/environments/at-scale.md
```diff
@@ -0,0 +1,285 @@
...
+		escClient := esc.NewClient(configuration)
+		authCtx := esc.NewAuthContext(accessToken)
+
+        // Always reference the `stable` version of this environment
+		_, values, err := escClient.OpenAndReadEnvironmentAtVersion(authCtx, orgName, projName, envName, "stable")
+		if err != nil {
+			return fmt.Errorf("failed to open environment: %v", err)
+		}
+
+		awsProvider, err := getAwsProviderFromEnv(ctx, values)
```
nyobe, line 155: I think this example can be simplified because the default aws provider will pick up these env vars automatically if the environment is imported by the stack config.

I think using the ESC SDK within a stack makes the example appear more complicated, and we might be better served having a different non-stack example of using the SDK instead.

dschaller: 👍 yeah I did have the thought this is a somewhat exaggerated example meant to show using JIT secrets in a practical application.

I ended up sticking with it because while the provider will use the environment variables by default you have to set those which when performing updates locally means you have to open the environment as an explicit step. With this you don't have to do any of that. Just as long as your PAT is set correctly.

nyobe: is an explicit open really required? just having the environment in the config seems to work for me 😕 
```
environment:
  - dev-sandbox
```

```
const bucket = new aws.s3.BucketV2("my-bucket");
export const bucketName = bucket.id;
```

interurban:
Hey @dschaller Are you ready for a docs team review on this or still working on it?

dschaller:
Hey @interurban let me convert this to a draft based on some convo I had async with @nyobe

meagancojocar:
closing as stale, feel free to reopen!

Issue #11655, opened 2022-09-27
Add limitations to SAML & SCIM documentation


meagancojocar:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details
Bundling a couple known issues in order of priority:

1. Add a note to SCIM page that we only support one Pulumi org per SCIM application. SAML applications can support multiple applications.
2. Update Okta docs to reflect new Okta flow (they work now just slightly differences in the experience)

### Affected area/feature
Docs > User Guides > SAML Single Sign-on (SSO)
Docs > User Guides > SCIM 2.0 Integration

GeoffMillerAZ:
I created:

https://github.com/pulumi/docs/pull/13521

Have I put the note in too many places now? Let me know and I can trim it back a bit. I made it very easy to update and very easy to scale as more guides are added.

preview: http://www-testing-pulumi-docs-origin-pr-13521-619130cd.s3-website.us-west-2.amazonaws.com/docs/pulumi-cloud/access-management/saml/

GeoffMillerAZ:
Closing due to merge on https://github.com/pulumi/docs/pull/13521

PR #13521, opened 2024-12-06
Feature/11655 saml SCIM limitations

GeoffMillerAZ:
### Proposed changes

I added the info to each SAML and SCIM guide to address ensuring users are aware of the cardinality limitations of SAML and SCIM providers to Pulumi Organizations. I used a template (shortcode) that when no args are passed to it, it speaks to the concept generically. But when an idp argument is passed to it, it customizes the message for that IdP provider to make it feel more at home in the provider-specific docs. I made one for SCIM and one for SAML. This will also make it very easy to be included in any new guides going forward.

I also update the gitignore to include some dev tools I like to use. I also added a .editorconfig that autmatically fixes the issues with .md files that the git pre-commit hook tests are testing for. This makes it much easier to stay in compliance with the tests. It you're using a supported IDE, it will just put you in compliance on save if you're settings take advantage of that feature.

I also added the tools I updated the .gitignore for to the .prettierconfig so their local files don't mess with the pre-commit hook tests.

One of the tools I added support for is devbox, which sits on top of nix shell. This makes it so anyone using devbox just opens the workspace and the shell will be automatically setup with the right versions of the right tools and config to just run make build and what not and be contributing in no time.

### Related issues (optional)

#11655


content/docs/pulumi-cloud/access-management/scim/_index.md
```diff
@@ -24,6 +24,8 @@ The [Pulumi Cloud](https://app.pulumi.com) supports System for Cross-domain Iden
 If desired, in addition to the SCIM-managed teams, one can also configure and manage Pulumi-local teams in the Pulumi Cloud. See [Teams](/docs/pulumi-cloud/access-management/teams/) for how to configure teams in the Pulumi Cloud.
     {{% /notes %}}
 
+{{< sso-scim-limits-info idp="your Identity Provider" >}}
+
```
caseyyh, line 28: In terms of visuals, it does look slightly strange to have 2 notes back to back with different styling. Would it be easy to make the SAML & SCIM notes consistent with the existing notes?

Screenshot 2024-12-09 at 1 11 48 PM

layouts/shortcodes/sso-saml-limits-info.html
```diff
@@ -0,0 +1,4 @@
+{{- $idp := .Get "idp" | default "your Identity Provider" -}}
+<div class="note info">
+  <p><strong>Note:</strong> A single SAML application in {{ $idp }} can support multiple Pulumi organizations. This allows you to manage authentication for multiple teams from one centralized configuration.</p>
```
caseyyh, line 3: I hesitate to include this here, since we don't support this for SCIM and SCIM users will first walk through the SAML step. I would probably drop this note and instead issue forewarning that we don't support multi-org SCIM yet:

> If you manage multiple Pulumi organizations and plan to enable SCIM provisioning on your SAML app integration, you must configure separate applications for each organization in {{ $idp }}. Pulumi supports only one Pulumi organization per SCIM-enabled application.

caseyyh (commented):
Thank you for picking this up!!

GeoffMillerAZ:
@caseyyh ,

is it too much to have the note both

image

and

image

and then again for each IdP?

should I remove the one on `Docs  > Pulumi Cloud > Identity & access management > SAML(SSO) > Using SAML`?

content/docs/pulumi-cloud/access-management/scim/_index.md
GeoffMillerAZ: I think I fixed this now:

image

layouts/shortcodes/sso-saml-limits-info.html
GeoffMillerAZ: OK. How does this look? For the SAML parts of the docs:

image

then for the SCIM parts:

image

or were you saying just use the following for all places just like you wrote above:
> If you manage multiple Pulumi organizations and plan to enable SCIM provisioning on your SAML app integration, you must configure separate applications for each organization in {{ $idp }}. Pulumi supports only one Pulumi organization per SCIM-enabled application.

GeoffMillerAZ: OK. I have this just being the same note each time:

image

caseyyh:
> @caseyyh ,
> 
> is it too much to have the note both
> 
> and then again for each IdP?
> 
> should I remove the one on `Docs > Pulumi Cloud > Identity & access management > SAML(SSO) > Using SAML`?

Oh good question, I think just the one note on the parent page (not on individual IDP pages) should be okay!

caseyyh (approved):

thoward (changes requested):
The content changes look great, and thank so much for doing that. That said, there's quite a lot of additional changes to other files in these changesets as well (dotfiles, re-render of assets, devbox config, etc). These will need to be removed. I'll go ahead and make those changes to the commits and push and update to the branch on this one.

In the future we should be conscious of keeping PRs focused on a single concern (e.g. if you want to modify a dotfile or the asset bundles, that's fine, but it should be done in a separate PR from content changes.

thoward (approved):

PR #13530, opened 2024-12-06
Remove docs pointing to the old pulumi-cloud project. 

mjeffryes:
As part of our plan to [Sunset pulumi-cloud project](https://docs.google.com/document/d/1dS8SHjQjxVTxSVRpcjhQM657rRg7HpttqBXh4qQZq4k/edit?tab=t.0#heading=h.vmiybe9heukw)
we need to remove the olds docs for this project. 

I've removed both the "iac/clouds/aws/cloudfx/*" docs and the API reference that was in "static-prebuilt/docs/reference/pkg/nodejs/pulumi/cloud"

I adjusted a few places that the cloudfx docs pages were linked to just be links to the pulumi-cloud repo (which now has a warning about being archived) and removed the special case for this path from [infrastructure/cloudfrontLambdaAssociations.ts](https://github.com/pulumi/docs/compare/kill_pulumi_cloud?expand=1#diff-2b81c45bf2344533fee806b23c44b98d02e167ec781d524bd789fde9a2abec08)

Also updates some links in blogs to the (now deleted) examples using pulumi-cloud to point to pinned commits just before the examples we removed. This I've taken this approach for blogs since retaining old blog posts is a nice historical record, but we'd like to avoid introducing broken links


thoward (approved):
Looks like you covered all the bases. Thanks, @mjeffryes !

PR #13547, opened 2024-12-09
Update deploy-changes.md

ldgarcia:
### Proposed changes

Minor fixes: ensure imports and variables match previous definitions for Python examples.


thoward:
Nice catch @ldgarcia ! We really appreciate the contribution. I noticed a few more spots where we had the same inconsistency in variable names in our other supported languages, so I went ahead and fixed those as well. This looks good to merge, IMO.

thoward (approved):

PR #13620, opened 2024-12-13
fix incorrect query param in deployment logs endpoint docs

nyobe:
### Proposed changes

The deployment logs endpoint actually uses a [`continuationToken` query param](https://github.com/pulumi/pulumi-service/blob/31449fbea5f22e570b72a602202d7f02ccbcafd1/cmd/service/api/deployments.go#L582) instead of the documented `nextToken` one.  Unfortunately the api response returns `nextToken`, so it's a bit confusing :(

### Related issues (optional)

Fixes https://github.com/pulumi/pulumi-service/issues/24318

### DIFF
--- a/content/docs/pulumi-cloud/deployments/api.md
+++ b/content/docs/pulumi-cloud/deployments/api.md
@@ -1180,7 +1180,7 @@ The `token` is a string that contains the `job`, `offset` and `step` of the next
 
 The following query parameters are available:
 
-* **nextToken**: A string returned by the previous response, that can be used to get the next set of logs.
+* **continuationToken**: A string returned by the previous response, that can be used to get the next set of logs.
 
 ##### Example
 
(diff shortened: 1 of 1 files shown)


content/docs/pulumi-cloud/deployments/api.md
```diff
@@ -1180,7 +1180,7 @@ The `token` is a string that contains the `job`, `offset` and `step` of the next
 
 The following query parameters are available:
 
-* **nextToken**: A string returned by the previous response, that can be used to get the next set of logs.
+* **continuationToken**: A string returned by the previous response, that can be used to get the next set of logs.
```
komalali, line 1183: Can we change this to say that it's the `nextToken` value returned by the previous response?

aureq: Yes, that would be a very helpful clarification.

komalali: LGTM!

aureq (approved):
Looks great to me. Thank you @nyobe ⭐

komalali (approved):

interurban (approved):

Issue #11777, opened 2020-06-23
Document minimum system requirements


EvanBoyle:
We don't currently document the minimum system requirements (RAM, etc) required to run Pulumi. Our Go SDK is a particularly hungry resource hog due to the number of types we generate for providers (a simple AWS program binary can be 100mb). We had a user run into an OOM error when trying to use the Go SDK. The issue turned out to be at the `go run` step, before any pulumi user code even executed. The user was using a 1GB VM to run pulumi. They saw inconsistent success with a 2GB VM, and consistent success with a 4GB VM. We should explore these limits and have a page with minimum system requirements. This may vary by language, and we should decide if we want to publish a single set of limits or break it down by runtime. 

See thread for more details: https://pulumi-community.slack.com/archives/CCWP5TJ5U/p1592928208198100

GeoffMillerAZ:
are the requirements known? 

Or is part of this issue setting up a benchmarking test and then executing across each? Then just make a table or something that presents the recommendations?

If so, does this need to be integrated into some periodic process to re-evaluate benchmarks as the product evolves?

A blanket requirement was mentioned above, I could execute that as the simplest path and then address the runtime specific concern in the future. Please let me know what would be best and I'll take care of this.

thoward:
Requirements are not known (by me). I would reach out to engineering and see if they have some numbers/stats to share. Otherwise, benchmarking/doing that research is part of the project.

thoward:
Fixed in #13662

PR #13662, opened 2024-12-18
Feature/11777 min requirements

GeoffMillerAZ:
### Proposed changes

Add minimum recommended requirements for the Pulumi cli.

I build https://github.com/pulumi/pulumi-benchmarking

it uses python and ts to build the same project. I used a json file to represent a typical structure that a pulumi project might see. That is, a sets of resources with dependencies that build a dependency tree. Because not all resources can build at once. This json file is used to make it easy to reproduce the same thing across runtimes. I currently have it implemented in ts and python. I also have it so you can do multiple of these trees on the same project. Right now I have it hard-coded at 2 of these trees. I will also make it so you can set a max number of resources that it will stop doing across the number of trees you give so you can limit the size for whatever reason. So number of buckets and number or resources are the 2 variables that give you the ability to quickly resize this test.

See the screenshot for a mermaid diagram of how each tree looks like for dependencies.

I used activity monitor just casually and an up got to 10% CPU on my m1 macbook pro. And it got to like 1.5GB of RAM. The destroy got to 20% and same RAM. It's 750MB disk space.

### Related issues (optional)

https://github.com/pulumi/docs/issues/11777


content/docs/iac/download-install/_index.md
thoward, line 1: Following up on my [comment on your last PR](https://github.com/pulumi/docs/pull/13521#pullrequestreview-2507123076), this should be the only file in this PR.

I will remove all the extra files in order to get this merged, but in the future it would be helpful to ensure PRs have a clean [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns), following the [single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).

thoward (approved):
Thanks for knocking this older issue out, @GeoffMillerAZ! Fixed a few things (see note about additional files in PR) and now it's ready for merge.

PR #13695, opened 2024-12-23
🌿 add missing top-level `values:` to various ESC reference examples

aureq:
### Proposed changes

Some ESC reference examples were missing the top-level `values:` key. This could confuse the reader into thinking `values:` is not required but where in fact it's required. As ESC doesn't complain about such syntax it's even harder to identify the problem.

This PR adds those missing top-level `values:` for the relevant ESC reference examples and makes indentation consistent.

Additionally, a few trailing spaces have been removed.


interurban (approved):
Thanks @aureq !

PR #13674, opened 2024-12-19
Python for DevOps article

adamgordonbell:

PR #13718, opened 2024-12-24
Enabling Scheduled posts in blog

adamgordonbell:
- For a production build, turn off `--buildFuture`
- We run a build each morning, to publish anything scheduled
- No one has to merge things on release day manually

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -3,9 +3,12 @@ on:
   push:
     branches:
       - master
+  schedule:
+    # Run at 6 AM Eastern (11:00 UTC)
+    - cron: '0 11 * * *'
 permissions:
-      id-token: write   # This is required for requesting the JWT
-      contents: read   # This is required for actions/checkout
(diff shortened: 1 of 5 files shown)


adamgordonbell:
If this is merged, we can test it with this PR: https://github.com/pulumi/docs/pull/13674

.github/workflows/build-and-deploy.yml
```diff
@@ -3,9 +3,12 @@ on:
   push:
     branches:
       - master
+  schedule:
+    # Run at 6 AM Eastern (11:00 UTC)
+    - cron: '0 11 * * *'
```
sean1588, line 8: hmm wonder if it would be better to just run this monday to friday. I think it should be fine as it is, but just in case of anything we will be around on the weekdays.

sean1588 (approved):
thanks @adamgordonbell ! and so sorry for the late review on this! I've owed you this one for a while. Going to leave an approval, but can you hold off on merging till monday just in case. All looks good to me though. I ran `hugo list future` and it came up empty so I think all good.

PR #13723, opened 2024-12-26
Fix typo HMS -> HSM

Asamsig:


interurban (approved):
Thanks @Asamsig for the catch and PR!

Issue #13652, opened 2024-12-17
Left nav is hidden at full screen on Macbook Air


cnunciato:
At full screen (1280 wide) on a new MacBook Air, the left nav is hidden 😢:

<img width="1392" alt="Image" src="https://github.com/user-attachments/assets/a3912a59-d089-4f0a-b203-cdf3013db310" />

<img width="1392" alt="Image" src="https://github.com/user-attachments/assets/3f9936bc-091c-4064-a264-76e3d45ba727" />

This makes it quite cumbersome to use the docs, because as you navigate around, the left nav keeps closing, and you have to reopen before you can click to another page. Also, when it's closed, it's not clear where you are in the hierarchy, and there's value in knowing this as well (e.g., when you land on a page after from a Google or Algolia search). 

My recommendation would be to show the left nav at `lg` (iPad width) and up, and to hide the right/on-this-page nav instead; it's much less important than the left.

interurban:
Interesting, thanks for flagging this @cnunciato we'll take a look.

PR #13735, opened 2024-12-30
Fix to adjust the breakpoint for left nav

interurban:
Fixes https://github.com/pulumi/docs/issues/13652 to show the left nav at lg (iPad width) and up, and to hide the right/on-this-page nav instead.


sean1588 (approved):
@interurban - leaving you an approval here, for whenever you are ready to merge

PR #13741, opened 2025-01-03
Update Azure AD -> Entra ID in SAML/SCIM docs

isabeldepapel:
### Proposed changes
Polishing up some of our SAML/SCIM docs: Updates the references to Azure Active Directory to its current name, Microsoft Entra ID. Also updates the images for setting up SAML using Entra since they were fairly outdated. I stuck with only updating the AAD references in the Cloud- or SAML-specific sections, but happy to do a full find/replace if you folks think that's best.

### Related issues (optional)
Fixes https://github.com/pulumi/pulumi-service/issues/23881


glena (approved):

interurban (approved):
awesome ! thanks so much!

PR #13744, opened 2025-01-03
adding insights to featured product list on docs homepage

interurban:
Adding a featured product card for Insights


thoward:
This section uses a three-up card layout (because at the time we only had three products to feature and the layout is wider due to the lack of sidebars). With a fourth product here, I think we should change this to a 2-up layout in two rows. As it is now, It looks a little unbalanced: 

image

thoward (changes requested):
Adjust layout to work for four products.

thoward (approved):

PR #13758, opened 2025-01-06
Update REST API docs: add get team endpoint and minor fixes

caseyyh:
### Proposed changes

Fixes https://github.com/pulumi/pulumi-service/issues/21535

1. Adds `GetTeam` REST API endpoint
2. Corrects typo in List OIDC API
3. Adds default response for List Updates API


meagancojocar:
Any chance we can combine Deployments REST API docs with our Pulumi Cloud REST API docs as part of this work?

caseyyh:
> Any chance we can combine Deployments REST API docs with our Pulumi Cloud REST API docs as part of this work?

like move the link to here?

Screenshot 2025-01-07 at 3 06 25 PM

meagancojocar:
> > Any chance we can combine Deployments REST API docs with our Pulumi Cloud REST API docs as part of this work?
> 
> like move the link to here?
> 
> Screenshot 2025-01-07 at 3 06 25 PM

No, like combine them into one API doc! We have schedules and TTL in the non-deploy docs now, its a mess. @komalali let me know if you disagree

komalali:
@meagancojocar yeah, I agree with you. We'll need redirects / aliases to make sure existing links don't break.

caseyyh:
But my opinion is that the REST API doc page is already huge, rather than lean in on a single uber-huge page, we should split it out by product maybe?

thoward (approved):

PR #13762, opened 2025-01-07
Adding Torian Crane as co-presenter to upcoming workshops

toriancrane:
Adding Torian Crane as co-presenter to upcoming workshops


dirien (approved):
LGTM!

Issue #13697, opened 2024-12-23
PRs are blocked from merging due to "3 new 404s were detected"


justinvp:
The "Install deps and build site" workflow for PRs is failing with:

```
Checking http://www-testing-pulumi-docs-origin-pr-13685-7c8079ec.s3-website.us-west-2.amazonaws.com/ for new 404s...
Checking 2014 URLs...

3 new 404s were detected:

/events/exploring-the-intersection-of-cluster-api-and-infrastructure-as-code/ : 403
/events/nginx-deploy-kubernetes-and-the-kitchen-sink/ : 403
/events/winning-the-product-agility-jackpot/ : 403

These are pages that exist on pulumi.com but do not exist at http://www-testing-pulumi-docs-origin-pr-13685-7c8079ec.s3-website.us-west-2.amazonaws.com./

Please make sure any content that's been moved has been an appropriate alias or redirect in place.
```

https://github.com/pulumi/docs/actions/runs/12438037440/job/34729145238#step:7:2779

This is blocking automatic merging of CLI release PRs

sean1588:
Hey @justinvp , we have some issues with that 404 checker that was inaccurately reporting issues. We made some updates to stop erroring on these for the time being. So we just need to update the PR branches with the latest from master and they should start passing now.  Though I'm guessing we only need the latest of those failed PRs since they are updating the same version file?

justinvp:
> Though I'm guessing we only need the latest of those failed PRs since they are updating the same version file?

Yep. Just the latest. Older ones can be closed (I've done that). Thanks!

jkodroff:
The checker was disabled, and therefore this issue is probably resolved.

PR #13775, opened 2025-01-08
Disable 404 checker

thoward:
The 404 checker has been creating false positives which interrupt the PR workflow. Disabling this temporarily until we can figure out better logic for this.


shughes26 (approved):

thoward:
Original report / issue to track fixes is here: https://github.com/pulumi/docs/issues/13697

Issue #13713, opened 2024-12-23
Add docs for the new DependsOn capability for InvokeOutputOptions


justinvp:
Document the new DependsOn capability for InvokeOutputOptions at https://www.pulumi.com/docs/iac/concepts/resources/functions/#invoke-options

PR #13778, opened 2025-01-09
Update invoke docs for dependsOn

julienp:
### Proposed changes

New option for output form invokes, released for Python, Node.js and Go. Java and Dotnet should be in the next release.

Fixes https://github.com/pulumi/docs/issues/13713


thoward (approved):
Looks good! Thanks for the content update.

PR #13781, opened 2025-01-09
fixing 404s from older resources pages which are now under events

interurban:
Clean up and fixes for 404s related to /resources moving to /events


thoward (approved):
Looks good! Thanks for catching these.

PR #13782, opened 2025-01-09
publish APAC workshop for 1/22

shughes26:
### Proposed changes
Workshop added for https://github.com/pulumi/marketing/issues/1243

Known bug for APAC workshops (we're looking to fix this next week): Hugo doesn't support date displays over a date line without some extra configuration that we need to determine the blast radius of, so 1/22 will display in locations where it's actually 1/21

### DIFF
--- a/content/events/platform-engineering-done-right/index.md
+++ b/content/events/platform-engineering-done-right/index.md
@@ -1,0 +1,101 @@
+---
+# Name of the event, <= 60 characters
+title: "Platform Engineering Done Right: Steps to Success"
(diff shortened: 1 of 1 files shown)


SaraDPH:
Is it possible to show the actual TZ? I see CST for me, but for the AEST, it shows GMT, which is not user-friendly as they can see at a glance that it matches their actual TZ


image

SaraDPH (approved):

shughes26:
> Is it possible to show the actual TZ? I see CST for me, but for the AEST, it shows GMT, which is not user-friendly as they can see at a glance that it matches their actual TZ
> 
This is being displayed by the `intl` library formatting, which selects between `PST` style vs `GMT-8` style based on an ECMA weighting algorithm which would be pretty complex to control

PR #13785, opened 2025-01-09
update homepage tweet carousel to use X logo

shughes26:
### Proposed changes
https://github.com/pulumi/marketing/issues/1227


dirien:
@shughes26 the PR is only for the carousel at the moment? Should I extend it with the other places or do you wanted to create separate PRs for it?

interurban (approved):

shughes26:
> @shughes26 the PR is only for the carousel at the moment? Should I extend it with the other places or do you wanted to create separate PRs for it?

@dirien I'll do a separate task to actually find all of them. There's no guarantee they'll all render properly by just changing the image, so it'll take some verification and adjustment on each one

PR #13786, opened 2025-01-09
Don't redirect on esc sdk docs

seanyeh:


thoward (approved):
looks good! thanks for hunting this down.

PR #13794, opened 2025-01-10
Add Adam Gordon Bell's profile and remove Kat Morgan's profile from community section

adamgordonbell:


thoward (approved):
Changing of the guard. Welcome @adamgordonbell !

PR #13798, opened 2025-01-12
fix: Correct Python example syntax in Config documentation

0xThresh:
### Proposed changes

Fixes the syntax for the Python examples in the Config documentation, which currently contains invalid semicolons.


interurban (approved):
awesome thanks!

PR #13807, opened 2025-01-14
pulumi-stacks provide example

Jeroendevr:
A little example how the stack and stacks work together

### Proposed changes

The pulumi-stacks documentation was not directly clear to me. The variables `stacks` and `stack` made it a little bit confusing. 

### Comment
I am a beginner in Pulumi so I came across this documentation. 
For a maintainer or a experiences user I would also like an example how you could use different stacks together. Now `vpcInfra` is linked to the `vpc-infra` project on the `dev` stack. But I guess its use is to set the `vpcId` based on the current `stack` and not hard to the `vpc-infra/dev`. 

So something like
```yaml
values:
  stackRefs:
    fn::open::pulumi-stacks:
       stacks:
        vpcInfra:
          stack: vpc-infra/{current-stack}
```
or is there a better way to do this?

### DIFF
--- a/content/docs/esc/integrations/infrastructure/pulumi-iac/pulumi-stacks.md
+++ b/content/docs/esc/integrations/infrastructure/pulumi-iac/pulumi-stacks.md
@@ -30,11 +30,26 @@ values:
     privateSubnetIds: ${stackRefs.vpcInfra.privateSubnetIds}
 ```
 
+Where your stack file is for e.g.
+```py
+# __main__.py
+import pulumi
+import pulumi_awsx as awsx
+
+# Fetch the default VPC for the current AWS region.
+vpc = awsx.ec2.DefaultVpc("default-vpc")
+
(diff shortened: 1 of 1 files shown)


Jeroendevr:
Perhaps I understand it a little bit better now. One should not simply create a dynamic environment (one which loads the current stack). But create two seperate environments en load each environment into the appropriate stack. This would have just a little bit of duplicated code.

shughes26:
Hi Jeroen, thanks for contributing!

Because ESC only supports YAML, the example code provided in python will not work. We would otherwise like to use your `stacks` description clarification. If you could remove the python code sample and add a period to the end of the `stacks` description, that would be amazing!

Jeroendevr:
Thanks for reviewing and accepting the contribution. Here you go.

thoward (approved):
Thanks for making those changes, @Jeroendevr. We appreciate the contribution!

PR #13810, opened 2025-01-14
Document using the PKO with ArgoCD

automagic:
### Proposed changes

Adding a subsection to the PKO docs (under Continuous Delivery) about using ArgoCD with the PKO

### Related issues (optional)

https://github.com/pulumi/pulumi-kubernetes-operator/issues/775


content/docs/iac/using-pulumi/continuous-delivery/pulumi-kubernetes-operator.md
```diff
@@ -375,6 +376,109 @@ const mystack = new k8s.apiextensions.CustomResource("my-stack", {
 });
 ```
 
+### Using Argo CD
+
+Much like Flux, we can use ArgoCD to manage the lifetime of the Stack via a GitOps paradigm.  This will give us the advantage of using the ArgoCD UI or CLI to interact with the Stack, and allowing ArgoCD to reconcile changes to the repo.
+
+First, we need to define a Pulumi Stack as a kubernetes manifest that ArgoCD can understand.  We assume here that this manifest lives in the same repository as the Pulumi program, in the subfolder `deploy/`.  However, this manifest could live in a separate repository, such as an "app-of-apps" repo. In this example, the manifest declares a service account and cluster role bindings to allow the stack to create resources in the cluster. Additionally, we expect a Secret to exist on the cluster containing a valid Pulumi access token
```
interurban, line 383:
```suggestion
First, we need to define a Pulumi stack as a Kubernetes manifest that ArgoCD can understand. We assume here that this manifest lives in the same repository as the Pulumi program, in the subfolder `deploy/`.  However, this manifest could live in a separate repository, such as an "app-of-apps" repo. In this example, the manifest declares a service account and cluster role bindings to allow the stack to create resources in the cluster. Additionally, we expect a Secret to exist on the cluster containing a valid Pulumi access token
```

interurban (approved):
Small comment, thx James

EronWright (commented):
LGTM

PR #13829, opened 2025-01-16
use the latest python version for cli docs build

tgummerer:
Since the CI image has been upgraded to ubuntu 24.04, python 3.7 is no longer available.  Upgrade to the latest version to make the docs build work again.

### DIFF
--- a/.github/workflows/pulumi-cli.yml
+++ b/.github/workflows/pulumi-cli.yml
@@ -123,7 +123,7 @@ jobs:
         goversion:
           - 1.16.x
         pythonversion:
-          - "3.7"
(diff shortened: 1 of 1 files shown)


.github/workflows/pulumi-cli.yml
```diff
@@ -123,7 +131,7 @@ jobs:
         goversion:
           - 1.16.x
         pythonversion:
-          - "3.7"
+          - "3.14"
```
julienp, line 134: 3.14 isn't out yet

```suggestion
          - "3.13"
```

julienp (approved):

PR #13840, opened 2025-01-16
Fixing some editorial gaps; ensuring Insights casing

interurban:
Addressing feedback from the team, ensuring Insights casing is correct. thx Gavin!

### DIFF
--- a/content/docs/insights/accounts.md
+++ b/content/docs/insights/accounts.md
@@ -2,7 +2,7 @@
 title: Accounts
 title_tag: Create and manage Insights accounts | Pulumi Insights
 h1: Create and manage Insights accounts
-meta_desc: This page describes how to create Pulumi Insights accounts for scanning provider account resources to use within Pulumi Cloud.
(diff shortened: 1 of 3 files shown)


content/docs/insights/accounts.md
```diff
@@ -61,7 +61,7 @@ For example:
 * Child account (region): `my-aws-account/us-east-1`
```
thoward, line 61: Is the terminology here of "child accounts" part of the terminology within the tool, or are these terms we're initiating and using to describe the relationship only in documentation? If the latter, I think maybe an alternate term like "scoped" or "topical" accounts. If there is some kind of dependent relationship between them how does it function? Do child accounts get deleted if a parent account is deleted? Do they inherit permissions or access from their "parent"? Are they subclasses in some way?

thoward (approved):

content/docs/insights/accounts.md
interurban: It's the former; i'll bring this up w/the team but for now I believe the answer is to keep the notion of parent / children in terms of account scanning.

PR #13853, opened 2025-01-17
Fix `esc --help` example in documentation

nelsonfigueroa:
### Proposed changes

The [ESC docs](https://www.pulumi.com/docs/esc/cli/) state that you can run `pulumi esc --help` to see a full list of `esc` commands, but the command should really be `esc --help`.


interurban (approved):
thanks good catch!

PR #13886, opened 2025-01-27
Update docs saying wildcards can be used

glena:
Fix https://github.com/pulumi/pulumi-service/issues/25464

Update doc to explain that wildcard notation can me used for team matching when configuring oidc policies


seanyeh (approved):

interurban (approved):
Awesome thx!

PR #13889, opened 2025-01-27
Scale to Zero with Serverless APIs in Python

adamgordonbell:


content/blog/serverless-api/index.md
```diff
@@ -0,0 +1,373 @@
...
+            Total cost: $0.28/month (!!)
+            Zero traffic = Zero cost
+---
+Every developer has that service that needs to exist but barely gets used. Maybe it's an internal reporting API that get hourly requests but just from a single client, or maybe a side project you want hosted somewhere, but that is rarely used.
+
+These services present a unique challenge: they need to be reliable when called, but actually keeping them always on and running seems like overkill.
+<!--more-->
+{{< youtube "cjLpOJn0B6g?rel=0" >}}
+
+What if you could maintain these low-traffic services with zero cost when they're idle, while ensuring they spring to life exactly when needed? This is where AWS Lambda's container support shines. While many developers think of Lambda primarily for individual functions, it's equally capable of hosting complete REST APIs that springs to life when needed.
```
aaronkao, line 52: I think a decade ago this was a pretty novel concept that you can maintain zero cost these mostly idle services. But in 2025, this is old news. I think you are going to have to have a more unique insight to serverless or come up with a different hook. Remember you are sharing it over to the serverless crowd who already knows this

aaronkao: As I read the post more, I think we need to answer who is our audience. Depending on that answer, will give us slightly different editorial direction.

content/blog/serverless-api/index.md
```diff
@@ -0,0 +1,373 @@
...
+- You get [complete AWS Lambda and API Gateway management](/registry/packages/aws/) for your low-traffic REST APIs
+- And [Infrastructure testing capabilities](/docs/iac/concepts/testing/) mean you can validate your serverless configuration before deployment
+
+You can find the full code in the [service status monitor repo](https://github.com/adamgordonbell/service-status-monitor) from my [python devops article](https://www.pulumi.com/blog/python-for-devops/). A larger version with rolling updates is coming soon.
+
+{{< notes type="info" >}}
+
+## **Alternative Approaches**
+
+There are other ways to run containers that scale to zero. [Google Cloud Run](https://www.pulumi.com/registry/packages/gcp/api-docs/cloudrun/service/) offers similar functionality on GCP, and [AWS App Runner](https://www.pulumi.com/blog/deploy-applications-with-aws-app-runner/) is another AWS service that can do this. Both have similar pricing models—very cheap for low-volume services.
```
aaronkao, line 349: There is an alternative framework to Pulumi which is using SST which is under the hood using Pulumi

aaronkao (commented):
My comments are mainly around who is the audience. Someone that doesn't know Lambda and we are teaching them through Pulumi as the implementation detail or someone that knows Lambda well but doesn't know about Pulumi? The blog post reads a lot more like the former. I think we need an unique insight or unique example that hooks people otherwise there are too many other articles out there that intro Lambda and serverless.

content/blog/serverless-api/index.md
adamgordonbell: The audience is python developers. I think that a Lambda Monolith, and one that works locally the same as in lambda and that uses containers instead of the zip file lambda format is a bit novel.

So this isn't the lambdas of 10 years ago, this is using lambda do host a service just like you would ECS or EKS, which to me is a lot more palatable of a workflow.

adamgordonbell: The point I was trying to get at was, you think of lambdas as these singular functions that respond to events, but its actually a great use case for hosting full backends with several advantages.

content/blog/serverless-api/index.md
adamgordonbell: yeah, true, SST is a alternative way to setup hosting a backend on lambda. I feel like it's existence shows that people need easy way to do this. But SST is typescript isn't it and my solution here shows doing it in python and hosting anything you can put in a container.  I mean, its not as simple as SST but hopefully valuable to python people.

aaronkao (approved):

PR #13900, opened 2025-01-28
Content updates

SaraDPH:
#### Terraform vs Pulumi page
"Terraform Cloud" is the actual category and non-branded query (meaning it doesn't include Pulumi vs) for a better term to a landing page; I am increasing the term for better SEO/SEM purposes.

Adding "Cloud" will not impact "Terraform" as a single word

####  PE Series Workshop

Updating based on the direction to stay away from "Portal" and move to "Platform"

#### Meetup blog article

Updating to add newest PUGs

### DIFF
--- a/content/blog/join-pulumi-user-group-community/index.md
+++ b/content/blog/join-pulumi-user-group-community/index.md
@@ -15,7 +15,7 @@ We have a global community made up of people from many different countries, but 
 
 Members of PUGs will have access to a global network of experts who can offer help, advice, and support on using Pulumi effectively. The PUGs will also feature regular in-person meetups, webinars, in-person and virtual workshops, and other events to help members stay up-to-date with the latest cloud infrastructure and Pulumi developments.
 
-## 20 Pulumi User Groups (PUGs) Meetups in 9 countries
+## 23 Pulumi User Groups (PUGs) Meetups in 10 countries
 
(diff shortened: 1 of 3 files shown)


thtmnisamnstr:
It's done for a reason, so I'm ok with it, but almost all of the changes from Terraform to Terraform Cloud are unnecessary and slightly less accurate (except the one comparing to Automation API). No issues with it here, because it's done for SEO purposes.

thtmnisamnstr (approved):

thoward (approved):
Agree w/ @thtmnisamnstr that the use of "Terraform Cloud" is strange and seems inappropriate. I'm curious how we arrived at this terminology. HashiCorp uses the term HCP (HashiCorp Cloud Platform). A reader will not know or care about our SEO optimization intention and will just read the term and say "What is that? I've never heard of Terraform Cloud. Do they mean HCP?". 

Anyway, not going to block the commit, but this change seems awkward (and might standout to industry-savvy readers and signal that we are a little out of touch).

SaraDPH:
Thank you, @thoward and @thtmnisamnstr! I'm open to suggestions on where to adjust the placement of 'Terraform Cloud.' It's the highest-ranking non-branded search query, but Google indicates the page isn't optimized for it, so the goal was a quick fix by increasing the count.

Issue #13931, opened 2025-01-31
Alert doc-ops for code example test cron job failures


sean1588:
Send alerts to docs-ops slack channel when code example tests fail. Only alert when the failure occurs during the cron run. This job also runs on PRs to test any new code examples. We do not need to be alerted if it is a PR test failure.

PR #13932, opened 2025-01-31
Notify docs-ops slack channel on code example failures

sean1588:
fixes: https://github.com/pulumi/docs/issues/13931

Send alerts to docs-ops slack channel when cron job that runs code example tests fail.


thoward (approved):
looks good

PR #13947, opened 2025-02-04
Draft: blog linting rules

adamgordonbell:


scripts/lint/markdownlint-runner.ts
```diff
@@ -0,0 +1,199 @@
+import * as path from 'path';
```
adamgordonbell, line 1: Adapted from previous lint-markdown.js

scripts/lint/find-markdown.ts
```diff
@@ -0,0 +1,212 @@
+import * as fs from 'fs';
```
adamgordonbell, line 1: Adapted from previous lint-markdown.js but expanded to cover if file is net new when compared to parent branch.

scripts/new-blog-post.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
adamgordonbell, line 1: Posts now default to future Wednesday, to encourage scheduling

PR #13952, opened 2025-02-04
Do not fetch master if already on master

sean1588:
The job that tests our example code is failing when triggered from the cron. Since the cron runs on master, the fetch command is failing. We should only run this if the current branch is not master.

https://github.com/pulumi/docs/actions/runs/13130938858/job/36635784348


thoward (approved):
nice catch!

PR #13956, opened 2025-02-05
adding policy configuration and scan

interurban:
Adds detail about setting up policies in insights for compliance and remediation as the last step in this user journey : https://www.pulumi.com/docs/insights/get-started/


content/docs/insights/get-started/add-policies.md
```diff
@@ -0,0 +1,124 @@
...
+
+Open your terminal and run:
+
+```bash
+pulumi policy new aws-typescript
+```
+
+This will initialize your project, creating the necessary files for Pulumi to use as a policy, including module dependencies to the providers that will let us interact with AWS resources.
+
+This template sets up an example resource policy that prevents S3 buckets from being publically readable:
```
lblackstone, line 32:
```suggestion
This template sets up an example resource policy that prevents S3 buckets from being publicly readable:
```

content/docs/insights/get-started/add-policies.md
```diff
@@ -0,0 +1,124 @@
...
+
+Select your newly published policy pack from the dropdown and choose the version you want to enforce.
+
+Here you can configure the enforcement level at either a global level for all, or a granular level for each individual policy check.
+
+We'll start with an enforcement level of **advisory** then click **Enable** to confirm your settings.
+
+![Insights Policies - New Policy Pack](/docs/insights/assets/enable-policy-pack.png)
+
+1. Now add your insights account to the policy group. Click **Add accounts** and type the name of the account you want to include for Insights policies. (e.g. insights-aws-account/us-west-2) Finally, click **Add account to policy group**
```
lblackstone, line 84: Should add a note that all accounts (and stacks) are automatically added to the `default-policy-group`.

lblackstone (approved):
Couple minor notes - LGTM overall!

content/docs/insights/get-started/add-policies.md
interurban: great; not added

content/docs/insights/get-started/add-policies.md
```diff
@@ -0,0 +1,128 @@
...
+```typescript
+import * as aws from "@pulumi/aws";
+import { PolicyPack, validateResourceOfType } from "@pulumi/policy";
+
+new PolicyPack("aws-typescript", {
+    policies: [{
+        name: "s3-no-public-read",
+        description: "Prohibits setting the publicRead or publicReadWrite permission on AWS S3 buckets.",
+        enforcementLevel: "mandatory",
+        validateResource: validateResourceOfType(aws.s3.Bucket, (bucket, args, reportViolation) => {
```
thoward, line 43: This is more of an issue with the `aws-typescript` policy template, but this only checks original-style V1 `Bucket`s. It would be a different type signature for `BucketV2`, and also given how different that API is, likely would need to be a stack policy not a resource policy. I don't know what Insights will find when doing a scan of unmanaged resources or how those will be represented (e.g. will it find a bucket and present it as a `Bucket` or a `BucketV2`? One would be handled by this policy, the other wouldn't (I think?). Anyway, this could be a stumbling block for some users.

thoward (approved):
Looks good (other than the one comment, which is kind of out of scope for this doc, but worth looking into).

content/docs/insights/get-started/add-policies.md
interurban: yea let me file an issue w/Levi around that template, perhaps we create a new one covering a scenario like ensuring a tag is present.

lblackstone: It looks like discovered resources are currently using the v1 Bucket type, but I agree that this is a potential area of confusion. Maybe we should pick a different example for getting started?

PR #13994, opened 2025-02-07
Update the self-managed backend note

cnunciato:
This change updates the note under Using a Self-Managed Backend to remove the bit about locking (which I believe is no longer an issue) and adjust the overall framing to be slightly less scary ("Pulumi will not behave as you might expect", "some commands may also behave slightly differently"), while continuing to encourage users to go with Pulumi Cloud.


sean1588:
Hey @cnunciato !! Great to see you around! 🙂 

Yeah this makes sense to me. Yeah you are correct, the locking issue was resolved a long time ago, so this is probably way out of date.

interurban (approved):
Thanks!!

PR #14016, opened 2025-02-08
Forbid full internal links

iwahbe:
This PR is best reviewed commit by commit:

- fc3670caaba756b5f27ab3d2785ae0a85b3a1ad0 adds a lint that forbids internal links (links that start with `https://www.pulumi.com/{registry,docs}/`).
- c67286bcd1efe29243224806e84f9ad6c4b02c48 fixes up lint violations with a shell script. The commit message itself describes the shell script used.


thoward (approved):
Overall, the change looks good. Thanks for adding a lint rule! I'm concerned there might be valid places where relative links don't work and the fully qualified url might be necessary.

I'd like to see every changed link tested to ensure we aren't introducing new broken links that are invisible to our automation/relative linking strategy. For that, I'd suggest a one-time manual check of each link changed as part of this PR, to make sure they are still working. I rough glance suggests this is about 50 links to check. Tedious, but not too bad.

iwahbe:
> Overall, the change looks good. Thanks for adding a lint rule! I'm concerned there might be valid places where relative links don't work and the fully qualified url might be necessary.

I'd be very surprised if this was the case, but it's theoretically possible. I believe that interpreting `/<path>` as `<baseurl>/<path>` is a property of Hugo, and should work for all markdown links.

> I'd like to see every changed link tested to ensure we aren't introducing new broken links that are invisible to our automation/relative linking strategy. 

We already have automation that monitors broken links in the [`#docs-ops`](https://pulumi.slack.com/archives/C04LH6RQZHT) channel. Can we instead monitor that? If post-merge the number of broken links increases we can investigate further or revert.

> For that, I'd suggest a one-time manual check of each link changed as part of this PR, to make sure they are still working. I rough glance suggests this is about 50 links to check. Tedious, but not too bad.

The diff for 8dbf582cffba3d7b7eeff37ee792caf0d0715c26 is `+509 -484`, and every `-` represents a change in *at least* one URL. I'd guess there are ~600 URLs changed here, but there are *at least* 484 URLs changed. I don't think a manual review is feasible here.

PR #14106, opened 2025-02-19
Secret rotation docs

komalali:
### Proposed changes

Fixes https://github.com/pulumi/pulumi-service/issues/25530


content/docs/esc/environments/rotation.md
```diff
@@ -0,0 +1,86 @@
...
+```yaml
+values:
+  rotatedCredentials:
+    fn::rotate::aws-iam:
+      inputs:
+        region: us-west-2
+        login: ${environments.credentials.production.aws.login}
+        userArn: arn:aws:iam::123456789012:user/username
+      # If there are existing credentials, it's possible to specify them in the state - but this section is optional
+    state:
```
arunkumar611, line 33: State indentation issue here.

content/docs/esc/environments/rotation.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Schedule
+
+TODO: Add schedule example
+
+## Best practices
+
+### Least privilege
+
+It is recommended to follow the principle of least privilege when defining the permissions for the role that will be used to rotate the secret. The role should have only the necessary permissions to perform the rotation action, and no more.
```
arunkumar611, line 62:
```suggestion
It is recommended to follow the principle of least privilege when defining the permissions for the role or user that will be used to rotate the secret. The role or user should have only the necessary permissions to perform the rotation action, and no more.
```

content/docs/esc/environments/rotation.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Composition
+
+It is a best practice to define a separate environment for your rotated functions, and import them into the environments that require the rotated credentials. This allows you to manage the rotation logic in a single place, and allows for the rotated environment to be versioned separately from the rest of your configuration. Specifically, since a new revision is created on every rotation, you may want to always import the latest version of the rotated environment to ensure that the latest rotated credentials are always used.
+
+Another reason to keep your rotated functions in a separate environment is that an environment containing a rotation function *cannot be rolled back*, since the rotated secrets have been deactivated. By keeping your rotation functions separate, you can ensure that the rest of your configuration can be rolled back to a previous revision if needed.
```
arunkumar611, line 70: Rollback is still enabled? 

If you are saying rollback is meaningless, then I'd word it differently

content/docs/esc/environments/rotation.md
```diff
@@ -0,0 +1,86 @@
...
+### Organization
+
+You may want to organize your rotated environment by rotation schedule. Since a single environment can only have one rotation schedule, you can group your rotated environments by the frequency at which they need to be rotated. For example, you may have a `daily` environment, a `weekly` environment, and a `monthly` environment, each with their own rotation schedule.
+
+### Handling partial failures
+
+If multiple rotation functions are defined in a single environment, it is possible that some fail while others succeed. In these cases, a partial failure will be reported.
+
+To handle partial failures, failed keys can be individually retried using the `esc rotate` command with the `--key` flag. This will allow you to retry the rotation of a specific key without affecting the rotation of the other keys in the environment.
+
```
arunkumar611, line 81: Reading the partial failure best practice, I almost want to have only 1 rotated secret per environment but the Organization best practice recommended I group them by rotating schedule.

content/docs/esc/environments/rotation.md
nyobe: specifically the permissions required by the rotator are IAM List, Create, and DeleteAccessKeys.

you can further restrict the scope of the permissions to a subset of IAM users by using [paths](https://aws.amazon.com/blogs/security/optimize-aws-administration-with-iam-paths/)

content/docs/esc/environments/rotation.md
komalali: I mean, either is an option. I can add that nuance.

content/docs/esc/environments/rotation.md
komalali: We should disable rollback completely. But currently it is possible to click rollback - but the provider will not be updated back to the secrets that were valid at the time. We should really just disable this.

content/docs/esc/environments/rotation.md
```diff
@@ -0,0 +1,86 @@
...
+        accessKeyId: AKIAIOSFODNN7EXAMPLE
+        secretAccessKey: 
+          fn::secret: wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLE
+      current:
+        accessKeyId: AKIAIOSFODNN8EXAMPLE
+        secretAccessKey: 
+          fn::secret: wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLE
+```
+
+In the above example, note the `${environments.credentials.production.aws.login}` reference. This is implicit import of the `credentials/production` environment's aws.login path.
```
nyobe, line 44: consider extending the [imports docs](https://www.pulumi.com/docs/esc/environments/imports/) to describe environment references, and maybe adding a link to that, just to call out the distinction between regular imports and references a bit more?

content/docs/esc/environments/imports.md
```diff
@@ -11,12 +11,16 @@ menu:
...
 
 To address these challenges, Pulumi ESC allows you to identify common or closely related configuration settings and define them only once, as individual environments, and then _import_ those environments into other, more specialized environments as needed. Imports also allow you to expose certain environments without having to distribute any concrete values and to delegate responsibility for particular environments to other teams in your organization. Environments can import both static and dynamic values, including secrets, from any number of other environments.
 
+## Explicit imports
+
+Explicit imports are defined in the `imports` section of an environment. The `imports` section is a list of environment paths that are resolved at runtime. The values from the imported environments are merged into the current environment, with the current values overwriting the imported environment's values where keys are redefined.
```
pgavlin, line 22: Do we document the merge process more completely? Probably worth mentioning that it's [JSON Merge Patch](https://www.rfc-editor.org/rfc/rfc7396).

content/docs/esc/environments/rotation.md
pgavlin: Probably also worth mentioning somewhere that each distinct reference is evaluated exactly once. So in this environment:

```yaml
imports:
  - default/foo
values:
  a: ${environments.default.foo.bar}
```

`default/foo` is only evaluated once.

(really, this should boil down to "each referenced revision is only evaluated once", but that's not currently the case. this is only observable when the same revision is referenced by tag and by number)

content/docs/esc/environments/rotation.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Least privilege
+
+It is recommended to follow the principle of least privilege when defining the permissions for the user or role that will be used to rotate the secret. The user or role should have only the necessary permissions to perform the rotation action, and no more.
+
+The minimum required permissions for each rotation function are documented in the [Rotated Secret](/docs/esc/integrations/rotated-secrets) provider documentation.
+
+### Separation of concerns
+
+It is recommended that the login credentials required to perform the rotation action are stored in a separate environment, and imported via an [implicit import](/docs/esc/environments/imports#implicit-imports), as shown in the example above. This ensures that the credentials are not exposed to users who do not have access to the managing environment.
```
pgavlin, line 76:
```suggestion
It is recommended that the login credentials required to perform the rotation action are stored in a separate environment, and imported via an [implicit import](/docs/esc/environments/imports#implicit-imports), as shown in the example above. This ensures that the credentials are not exposed to users who unable to open the managing environment.
```

content/docs/esc/environments/rotation.md
```diff
@@ -0,0 +1,102 @@
...
+
+You may want to organize your rotated environment by rotation schedule. Since a single environment can only have one rotation schedule, you can group your rotated environments by the frequency at which they need to be rotated. For example, you may have a `daily` environment, a `weekly` environment, and a `monthly` environment, each with their own rotation schedule.
+
+Alternatively, you may want to keep a separate environment for each rotated secret. This avoids the potential for partial failures, described in more detail below.
+
+### Handling partial failures
+
+If multiple rotation functions are defined in a single environment, it is possible that some fail while others succeed. In these cases, a partial failure will be reported.
+
+To handle partial failures, failed keys can be individually retried using the `esc rotate` command with the `--path` flag. This will allow you to retry the rotation of a specific key without affecting the rotation of other keys in the environment.
```
pgavlin, line 96: did we not ship the retry button?

content/docs/esc/environments/rotation.md
```diff
@@ -0,0 +1,102 @@
...
+            fn::secret: 
+              ciphertext: ...
+        current:
+          accessKeyId: AKIA...
+          secretAccessKey: 
+            fn::secret: 
+              ciphertext: ...
+```
+
+In the above example, note the `${environments.credentials.production.aws.login}` reference. This is implicit import of the `credentials/production` environment's aws.login path.
```
IaroslavTitov, line 46:
```suggestion
In the above example, note the `${environments.credentials.production.aws.login}` reference. This is an implicit import of the `credentials/production` environment's aws.login path.
```

content/docs/esc/environments/rotation.md
```diff
@@ -0,0 +1,102 @@
...
+TODO: Add example of manual rotation
+
+### Schedule
+
+TODO: Add schedule example
+
+## Permissions
+
+- To `rotate` an environment, a user must have `WRITE` permissions on the environment, and `OPEN` permissions on any imported environment.
+- To `open` an environment, a user must have `OPEN` permissions on the environment (and does not need any permissions for the implicitly imported environment which provides the rotation credentials).
```
IaroslavTitov, line 63: How did we resolve the issue of malicious user referencing creds he shouldn't have access to, and rotating some user he shouldn't have access to? `does not need any permissions for the implicitly imported environment which provides the rotation credentials` sounds dangerous.

content/docs/esc/environments/rotation.md
```diff
@@ -0,0 +1,102 @@
...
+- To configure a rotation schedule for an environment, the user must have `WRITE` permissions on the environment.
+
+## Best practices
+
+### Least privilege
+
+It is recommended to follow the principle of least privilege when defining the permissions for the user or role that will be used to rotate the secret. The user or role should have only the necessary permissions to perform the rotation action, and no more.
+
+The minimum required permissions for each rotation function are documented in the [Rotated Secret](/docs/esc/integrations/rotated-secrets) provider documentation.
```
IaroslavTitov, line 72: This will plug into Sean's PR, right? Link broken as of now

content/docs/esc/environments/rotation.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Separation of concerns
+
+It is recommended that the login credentials required to perform the rotation action are stored in a separate environment, and imported via an [implicit import](/docs/esc/environments/imports#implicit-imports), as shown in the example above. This ensures that the credentials are not exposed to users who do not have access to the managing environment.
+
+### Composition
+
+It is a best practice to define a separate environment for your rotated functions, and import them into the environments that require the rotated credentials. This allows you to manage the rotation logic in a single place, and allows for the rotated environment to be versioned separately from the rest of your configuration. Specifically, since a new revision is created on every rotation, you may want to always import the latest version of the rotated environment to ensure that the latest rotated credentials are always used.
```
IaroslavTitov, line 80: nit
```suggestion
It is best practice to define a separate environment for your rotated functions, and import them into the environments that require the rotated credentials. This allows you to manage the rotation logic in a single place, and allows for the rotated environment to be versioned separately from the rest of your configuration. Specifically, since a new revision is created on every rotation, you may want to always import the latest version of the rotated environment to ensure that the latest rotated credentials are always used.
```

content/docs/esc/environments/rotation.md
```diff
@@ -0,0 +1,102 @@
...
+### Handling partial failures
+
+If multiple rotation functions are defined in a single environment, it is possible that some fail while others succeed. In these cases, a partial failure will be reported.
+
+To handle partial failures, failed keys can be individually retried using the `esc rotate` command with the `--path` flag. This will allow you to retry the rotation of a specific key without affecting the rotation of other keys in the environment.
+
+TODO: Add example of retrying a failed key
+
+{{% notes type="warning" %}}
+**WARNING** Beware of double rotation in the case of partial failures. If a key is rotated twice, the first rotation will be invalidated and the second rotation will be active. This can lead to unexpected behavior if not handled correctly, for example if the rotated secret has not been updated at the consumer.
```
IaroslavTitov, line 101:
```suggestion
**WARNING** Beware of double rotation in case of partial failures. If a key is rotated twice, the first rotation will be invalidated and the second rotation will be active. This can lead to downtime, if the rotated secret has not yet been pulled in by its consumers.
```

I thought I was the only one who hates articles!

content/docs/esc/environments/rotation.md
nyobe: not yet, we have an issue to fast follow on: https://github.com/pulumi/pulumi-service/issues/25970

content/docs/esc/environments/rotation.md
nyobe: should we preemptively add the tagging permissions too, since we're thinking about having the rotator add those? 🤔

content/docs/esc/environments/imports.md
```diff
@@ -11,12 +11,16 @@ menu:
...
 
 To address these challenges, Pulumi ESC allows you to identify common or closely related configuration settings and define them only once, as individual environments, and then _import_ those environments into other, more specialized environments as needed. Imports also allow you to expose certain environments without having to distribute any concrete values and to delegate responsibility for particular environments to other teams in your organization. Environments can import both static and dynamic values, including secrets, from any number of other environments.
 
+## Explicit imports
+
+Explicit imports are defined in the `imports` section of an environment. The `imports` section is a list of environment paths that are resolved at runtime. The values from the imported environments are merged into the current environment using a [JSON Merge Patch](https://www.rfc-editor.org/rfc/rfc7396), with the current values overwriting the imported environment's values where keys are redefined.
```
nyobe, line 22:
```suggestion
Explicit imports are defined in the `imports` section of an environment. The `imports` section is a list of environments that are resolved at runtime. The values from the imported environments are merged into the current environment using a [JSON Merge Patch](https://www.rfc-editor.org/rfc/rfc7396), with the current values overwriting the imported environment's values where keys are redefined.
```

(most of the time we talk about paths, we're talking about paths to values)

content/docs/esc/environments/rotation.md
komalali: I already did!

content/docs/esc/environments/imports.md
```diff
@@ -81,3 +85,85 @@ $ echo "'$GREETING'"
 ```
 
 Notice in the example that the `environmentVariables` were exposed to the `bash` command, but not to the surrounding shell, and that the values marked as secrets with `fn::secret` were protected from exposure.
+
+## Implicit imports
+
+Implicit imports are used to reference values from other environments without having to explicitly import them. This is useful when you want to reference a value from another environment without needing to expose the entire environment.
```
nyobe, line 91:
```suggestion
Implicit imports are used to reference values from other environments without having to explicitly import them. This is useful when you want to reference a value from another environment without needing to expose the entire environment.

Implicit imports take the form of a reference to the special `environments` key like `${environments.PROJECT.ENV.VALUEPATH}`.
```

maybe explicitly mention what the syntax is here, before the examples, so people know what to look for?

content/docs/esc/integrations/rotated-secrets/aws-iam.md
```diff
@@ -0,0 +1,156 @@
...
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "iam:ListAccessKeys",
+        "iam:CreateAccessKey",
+        "iam:DeleteAccessKey",
+        "iam:GetUser",
+        "iam:TagUser"
```
komalali, line 112: @nyobe re: tagging permissions

content/docs/esc/environments/rotation.md
```diff
@@ -0,0 +1,120 @@
...
+
+The rotation schedule can be defined as a [cron expression](https://crontab.cronhub.io/).
+
+![schedule.png](../schedule.png)
+
+## Permissions
+
+- To `rotate` an environment, a user must have `WRITE` permissions on the environment, and `OPEN` permissions on any imported environment.
+- To `open` an environment, a user must have `OPEN` permissions on the environment (and does not need any permissions for the implicitly imported environment which provides the rotation credentials).
+- To configure a rotation schedule for an environment, the user must have `WRITE` permissions on the environment.
```
nyobe, line 78: do we want to mention that you need read permission on imported environments when editing a rotated environment?  hmm, or maybe we hold off on that while we figure out whether we want to change that to be open permission

nyobe (approved):

content/docs/esc/integrations/rotated-secrets/aws-iam.md
nyobe: derp

content/docs/esc/environments/rotation.md
nyobe: oops! 😳

content/docs/esc/environments/imports.md
```diff
@@ -11,12 +11,16 @@ menu:
     weight: 2
 ---
 
-Environments can also be composed from other environments.
```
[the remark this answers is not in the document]
interurban: Since this wasn't addressed in this PR, I added a new issue: https://github.com/pulumi/docs/issues/14130

PR #14109, opened 2025-02-19
Show latest version of `pulumi/actions` in docs

justinvp:
Most doc pages have already been updated, but there were a few places where we were still mentioning an older version of `pulumi/actions`. This change updates to the latest version (v6).

Fixes https://github.com/pulumi/actions/issues/1297


interurban (approved):
Thx Justin

PR #14162, opened 2025-02-24
Adding pricing and features info for Insights

interurban:
Adding insights pricing detail to https://www.pulumi.com/pricing/

https://github.com/pulumi/marketing/issues/1283


content/pricing/_index.md
```diff
@@ -288,6 +300,24 @@ comparison_table:
...
+                    - content: $0.37, after 200 free
+                    - content: $1.10 / Custom
+                    - content: Custom
+                - title: Cost per minute
+                  items:
+                    - content: _blank
+                    - content: $0.01
+                    - content: $0.01 / Custom
+                    - content: $0.01 / Custom
+                - title: Minutes included/month
```
joeduffy, line 315: It seems confusing to list this here. They don’t get free minutes here and with deployments, it is the same pool of free minutes. I’d just keep it with deployments and not list this twice.

interurban: removed

AnuiDolphin:
Insights Discovery does not support for 'Free' today.

AnuiDolphin:
Also Policy Enforcement for Enterprise SKU should say 'Available' instead of a check mark

AnuiDolphin:
Here are the notes from our discussion on the UI:
* 'Price per resource/month' should be the first row
* 'Price per resource/month' for Free sku should say '200 free' to match IaC and not confuse users that it costs $200
* 'Minutes included/month' row should be merged into 'Workflow cost per minute' to show cost and number minutes free in the same cell (like we do for resources)
* 'Primary Accounts' should be a link to a docs page or FAQ describing what a Primary account is vs child account

AnuiDolphin (approved):

PR #14197, opened 2025-02-26
update css class names for top nav ad banner

shughes26:
### Proposed changes
Attempting to avoid stale css deletion by renaming the css (ran a full text search, no other instances of this class are used and the new names don't conflict)


interurban (approved):
Thx!

sean1588 (approved):
Not sure if this will work, but also wouldn't be too surprised if it did. Don't see any harm in trying this, so let's hope for the best.

PR #14211, opened 2025-02-27
Drop old ciphers vulnerable to Lucky13

kmosher:
This is the version we've been running on app.pulumi.com without complaint for many months now without issue.

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/secure-connections-supported-viewer-protocols-ciphers.html

### Related issues (optional)

https://github.com/pulumi/pulumi-service/issues/26208


sean1588 (approved):
thanks @kmosher !! LGTM

PR #14234, opened 2025-03-03
Add an entry for PULUMI_STACK

brandonpollack23:
Adding an env var doc for visibility.

This is added in pulumi/pulumi#18717


interurban (approved):
Thank you sir!

dasormeter (approved):

PR #14242, opened 2025-03-03
Update Post for SEO

asafashirov:
Added canonical to newer auto-naming config post, link to physical resource doc, and updated meta description. Intentionally not fully optimized as to not out-perform customizable resource auto-naming post.


asafashirov (commented):
resubmitting for pass failed checks before merge

PR #14257, opened 2025-03-03
Remove 6sense script

sean1588:
Removing the 6sense script to unblock our pipeline. This script is giving our tests timeout errors due to issues fetching the script. This is possibly due to issues with our account not being on a paid version. We checked in with @calon-pulumi and confirmed it was ok to temporarily (maybe permanently) remove this.


interurban (approved):
thx Sean

shughes26 (approved):

PR #14275, opened 2025-03-05
try to avoid using custom runners

tgummerer:
This runner has been deleted.  Let's see if using a non-custom runner works for running these tests, as that would be the cheaper solution.


sean1588:
thanks @tgummerer ! I'll push this through. Though, if I recall correctly we intentionally needed to bump to a paid runner for this job specifically. We'll see if this one is good enough for the job though. Thanks.

PR #14292, opened 2025-03-06
Clean up the blog post

komalali:
### Proposed changes

Removes some redundancy and makes the value prop clear in the opening line.

### DIFF
--- a/content/blog/announcing-pulumi-esc-github-action/index.md
+++ b/content/blog/announcing-pulumi-esc-github-action/index.md
@@ -11,7 +11,7 @@ tags:
   - features
 ---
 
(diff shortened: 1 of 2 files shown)


asafashirov:
I didn't add it in the file but also recommend a comparison table showing why Pulumi ESC is better than storing secrets statically, perhaps after Key Benefits section; e.g.

### **🔹 Why Use Pulumi ESC Instead of Static Secrets?**
| Feature | Pulumi ESC GitHub Action | Traditional Secrets Storage |
|---------|----------------|------------------------|
| **Secrets Storage** | Dynamic, injected only when needed | Static, stored in GitHub Secrets or CI/CD systems |
| **Security** | No long-lived credentials | Risk of secret leaks & unauthorized access |
| **Rotation** | Automatic, hands-free | Manual & error-prone |
| **Integration** | Native GitHub Actions support | Requires additional scripts |
| **Environment Variables** | Injected dynamically | Hardcoded & static |

For whatever reason comparing stuff just works well in search.

PR #14322, opened 2025-03-07
[esc] add note on large environment open failures

dschaller:
### Proposed changes

Add a note to the ESC docs on keeping environment contents slimmed down to only the values necessary as larger environment definitions (e.g. dumping a JSON blob into one) can result in the environment failing to open.

### Related issues (optional)

https://github.com/pulumi/pulumi-service/issues/22205


komalali (approved):

nyobe (approved):

thoward (approved):

Issue #14324, opened 2025-03-08
Add support for excluding tests


mjeffryes:
The docs examples test currently fail on every run because we don't have support for skipping tests where we lack the runtime environment to test properly.

PR #14366, opened 2025-03-12
Ignore example code tests when failure is expected.

sean1588:
fixes: https://github.com/pulumi/docs/issues/14324

Ignore example code tests that are expected to fail. I added an ignore file where we can specify directories as well as regex patterns for tests to skip. I also added the kubernetes examples to skip until we have the cluster configured for the ci to use. This is to reduce noise in the channel so we can focus on tests that are actually expected to pass but are failing.


mjeffryes (approved):
Nice; looking forward to a cleaner ops channel!

PR #14423, opened 2025-03-18
Fix indentation in esc api docs

seanyeh:
### Proposed changes

Fix indentation in https://www.pulumi.com/docs/pulumi-cloud/reference/cloud-rest-api/#default-response-61


interurban (approved):
looks good

PR #14449, opened 2025-03-19
docs: Move TLV PUG to Europe

dirien:


SaraDPH:
Let's ask someone in Israel what they think... Israel is in West Asia, but maybe they consider themselves something more specific. For example, Saudi Arabia is also West Asia but prefers to be grouped as Middle Eastern or Arab Nation.

PR #14456, opened 2025-03-19
docs: sort pugs by city name

dirien:


SaraDPH:
@dirien I do not approve of this change. This makes it much harder to see at a glance where there are PUGs. This is not user-friendly, they should have to go through an entire list to figure out if there is a PUG in their country/location

PR #14475, opened 2025-03-21
Update frontmatter with keywords

sean1588:
Generate keywords and add them to frontmatter.

This PR reads in the markdown files and uses the content to generate keywords. It then writes back the frontmatter to the file with keywords list under search.keywords.

I am taking a hybrid approach here that combines tf-idf and keybert. TF-IDF is used to find the keywords in a page by comparing how often they appear in a specific page relative to all the other pages. This helps us identify the unique keywords that make each page more distinct, helping to reject noise and common words that are of less meaning. Keybert is used to extract the keywords by looking at the semantic relevance of the words. We then take the keywords that are extracted that are common to both methods, with the intended result including keywords that are both semantically relevant and distinct across the corpus of pages. There are up to 7 keywords total that are added and we prioritize the keywords that are common to both methods when producing the list.

Outputs using this method to see what this produces:

https://github.com/pulumi/docs/pull/14559

### DIFF
--- a/.github/workflows/update-search-index.yml
+++ b/.github/workflows/update-search-index.yml
@@ -7,7 +7,7 @@ jobs:
   update:
     name: Update Algolia Search Index
     runs-on: ubuntu-latest
-    environment: production
+    environment: testing
     steps:
       - name: Install Node
         uses: actions/setup-node@v1
--- a/layouts/partials/docs/search.html
+++ b/layouts/partials/docs/search.html
@@ -8,7 +8,7 @@
             data-app-id="{{ getenv "ALGOLIA_APP_ID" }}"
             data-search-key="{{ getenv "ALGOLIA_APP_SEARCH_KEY" }}"
             data-facets="Docs,Registry,Tutorials"
-            data-index="production"
+            data-index="testing"
         ></div>
     </div>
     <div class="content-width-controls">
--- a/scripts/ci-push.sh
+++ b/scripts/ci-push.sh
@@ -7,7 +7,7 @@ source ./scripts/ci-login.sh
 ./scripts/build-site.sh
 ./scripts/sync-and-test-bucket.sh update
 
-./scripts/generate-search-index.sh
+# ./scripts/generate-search-index.sh
 
 node ./scripts/await-in-progress.js
 
--- a/scripts/ci-push.sh
+++ b/scripts/ci-push.sh
@@ -7,7 +7,7 @@ source ./scripts/ci-login.sh
 ./scripts/build-site.sh
 ./scripts/sync-and-test-bucket.sh update
 
-# ./scripts/generate-search-index.sh
+./scripts/generate-search-index.sh
 
 node ./scripts/await-in-progress.js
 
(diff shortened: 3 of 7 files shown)


mjeffryes (commented):
Interesting! IT doesn't feel like we're quite there yet as the keywords its picking out seem too general in a lot of cases, but let's keep playing with it! 

(It might be interesting to run it on the blog posts too; those pages have a lot of content that probably isn't making it into the index)

sean1588:
@mjeffryes - I have made some updates to the keyword extraction, see the PR description. This seems better overall IMO. See this [PR](https://github.com/pulumi/docs/pull/14555/files) that generates the keywords across all the /docs pages. I also have it deployed to the algolia testing index and it can also be tested out at [pulumi-test.io/docs](pulumi-test.io/docs). @thoward if you can take a look also. Let me know if this approach even makes sense :shrug:.

mjeffryes:
> @mjeffryes - I have made some updates to the keyword extraction, see the PR description. This seems better overall IMO. See this [PR](https://github.com/pulumi/docs/pull/14555/files) that generates the keywords across all the /docs pages. I also have it deployed to the algolia testing index and it can also be tested out at [pulumi-test.io/docs](pulumi-test.io/docs). @thoward if you can take a look also. Let me know if this approach even makes sense 🤷.
> 
> We should probably ignore any of the cli docs or other docs that are codegenned since they will be overwritten anyway.


Hmmm... Still doesn't seem like it's making a big difference. I did tried all the queries in our spreadsheet and they still all turn up empty. I wonder if it would do better on the blogs content? Overall, the keywords it's finding are too generic to be of much value, they are either already in the title and headings or very vague. Maybe this isn't as easy a win as I had thought.

mjeffryes:
Did you also add the blog keywords in the test index? https://github.com/pulumi/docs/pull/14559/files#diff-8689a0a899150c22622759c7fb62e73306a123f13db65fb0e6052a4e5ff8fca9 shows new keywords that I would think would help retrieve this blog when searching "EKS Auto Mode"

sean1588:
@mjeffryes - blog is deployed to testing index now as well as docs. Yeah agreed, still a bit too generic. I have some ideas, I'll try tomorrow on this. It does identify that eks auto mode blog post now fwiw.

sean1588:
@mjeffryes - ok so I have genned a sample using gpt. I did try using gpt as the language model for keybert, but results were not great. So I opted to just get rid of keybert altogether and just invoke gpt directly using the open ai api instead of using through keybert. The results here I think are a bit better but honestly in practice don't think this is any better than how search is currently functioning and arguably worse. I also deployed it to the algolia testing index so it is in the test env.

see sample here: https://github.com/pulumi/docs/pull/14623/

mjeffryes:
I actually think the GPT keywords might be a slight improvement. I think it's worth setting up an A/B test for it anyway

sean1588:
yeah I think the GPT keywords are better, but in practice I think still a bit questionable. I don't think we'll need any code changes in order to setup the A/B test. We should be able to configure this stuff using growthbook and just tell it to alter the dom element that specifies the index.

PR #14494, opened 2025-03-24
Adds instructions on how to handle ESM-only packages

foot:
### Proposed changes

Adds some explicit instructions on how to handle a particular nodejs error around `require`ing ESM-only modules from commonJS.

Certain large npm packages are starting to go ESM-only, e.g. https://github.com/kubernetes-client/javascript/releases/tag/1.0.0 so it may become more common to bump into this.

NodeJS is only just starting to support loading ESM modules from CommonJS.

Notes:
- This Gist helping people handle these issues has been around since 2021 https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

### DIFF
--- a/content/docs/iac/languages-sdks/javascript/_index.md
+++ b/content/docs/iac/languages-sdks/javascript/_index.md
@@ -251,6 +251,14 @@ runtime:
     nodeargs: "--loader ts-node/esm --no-warnings"
 ```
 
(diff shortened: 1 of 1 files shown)


interurban (approved):

julienp (approved):
Nice

PR #14503, opened 2025-03-24
Touch up ESC Kubernetes guide.

### LINKED ISSUE #14338: ESC: Kubernetes "integration" is actually a tutorial on cluster access
https://www.pulumi.com/docs/esc/integrations/kubernetes/kubernetes/

This doc is a howto. It's not really an integration per se because it's just using standard ESC building blocks to create a greater solution. It is, however, VERY helpful!

Change the title to something like "Kubernetes Cluster Access" to be more descriptive. The title is just "Kubernetes", which reads awkwardly in the navigation:

Image

We also link to the "Crosswalk for Kubernetes" guide, which is badly out of date.

jkodroff:
Fixes #14338


interurban (approved):
Thx Josh

Issue #14505, opened 2025-03-24
[Epic]: fix broken example tests


mjeffryes:
There are about 20 broken docs example tests that [have been skipped](https://github.com/pulumi/docs/pull/14506) so we don't miss any new failures. We should fix these as soon as we get a chance to keep our docs examples working!

jkodroff:
Adding this to the Docs Team Backlog so that we get a fresh set of eyes on this and prioritize.

PR #14506, opened 2025-03-24
Skip failing docs examples

mjeffryes:
There are about 20 failing example programs in master; this results in an alert in the #docs-ops channel everyday and makes it hard for us to notice and fix any new failures in the 300+ programs that are currently working!

This PR disables those failing test so we can catch and fix new failures in the future. I've also filed https://github.com/pulumi/docs/issues/14505 to track fixing these so we can ratchet this up to covering all of these programs.


thoward (approved):
Nice job! We really appreciate the help on this.

PR #14560, opened 2025-03-25
drop component v1 language

kramhuber:
drops language specific to components v1


interurban (approved):
thx!

thoward (approved):
Oops! This change didn't get in during launch. Had a messy git log due to WIP at that time. Cleaned up the git log and now it's ready to merge.

PR #14578, opened 2025-03-26
Remove Deployments Versus Page

jkodroff:
Remove the page b/c its tone is not customer-centric and its claims feel aspirational b/c they lack specific examples.

Also corrects CTA at the bottom of main Deployments page so that it reflects that we now have multiple guides pages.


meagancojocar:
Screenshot 2025-03-26 at 3 14 16 PM
I'll circle back and provide a more detail review but given the page views can we reword the tone and not delete it?

jkodroff:
@meagancojocar I don't think a rewrite is feasible since we don't actually provide a comparison between products (and that's a serious piece of product marketing that's beyond my skills to produce in a reasonable length of time), nor do we have code that actually demonstrates the advanced features of Deployments (in my skills wheelhouse, and something I think we should develop over time, but not feasible to do right now).

Despite the number of views, I cannot imagine how this page is useful in convincing anyone to use Deployments.

thoward:
@jkodroff given feedback from @meagancojocar, i see a few different options here:
1. merge this PR, deleting the page
2. close this PR without merging, file a ticket to update/replace the page (identify a suitable author to assign and prioritize)
3. leave this open for now and continue discussion on the topic

IMO, option 2 makes the most sense. that said, would like to get that update to the CTA merged. do you mind filing a new PR with just that?

jkodroff:
I do not mind, and was actually planning to rewrite now that I have some content in mind.

PR #14606, opened 2025-03-27
Restructure the choosers/headings to make the right nav work correctly

thoward:
Previously right nav was getting duplicate H3 entries because they were contained within multiple choosers. 

With this fix, choosers only contain H4s and H5s, which aren't shown in the menu. 

Problem Screenshot:
image


sean1588 (approved):
Thanks @thoward ! LGTM. Yeah I remember there was an issue open a long time ago to address the duplicate nav items in the right nav that get caused by this and we were looking in to it at one point. Not sure if it is still open somewhere, but yeah you can also imagine it is a bit hairy to fully solve.

Issue #14574, opened 2025-03-26
Include Blog links in the Docs search


sean1588:
Currently when the Algolia search is done at the docs portion of the site, it excludes blog links altogether. We should also list links to blog posts in the search results but down rank them beneath the docs and registry results. We should probably add a "blog" facet (tab) as well to the results listing.

PR #14609, opened 2025-03-27
Add blog search results to docs site

sean1588:
fixes: https://github.com/pulumi/docs/issues/14574

Adds blog results to docs site and but prioritizes other site sections in the result set when not under /blog. The behavior under blog section of our site remains the same.

Screenshot 2025-03-27 at 4 38 58 PM
Screenshot 2025-03-27 at 4 39 07 PM


mjeffryes:
@jkodroff Added you as a reviewer since you had shared some opinions about it!

mjeffryes (approved):
Feels a little hacky, but I think it will be helpful for our users and we could always try to develop a more sophisticated ranking later.

PR #14621, opened 2025-03-28
Add Define an Entry Point section to components blog

meagancojocar:

### DIFF
--- a/content/blog/pulumi-components/index.md
+++ b/content/blog/pulumi-components/index.md
@@ -605,6 +605,98 @@ runtime: yaml
 
 For other languages, you'd use the appropriate runtime value. This file is the key that enables Pulumi to generate SDKs in other languages from your component code.
 
+### Define an Entry Point
+
+The entrypoint analyzes components to automatically build a schema, and interact with the Pulumi engine to mange the component lifecycle.
+
+{{< chooser language "typescript,python,go,csharp,java" >}}
+
+{{% choosable language typescript %}}
+
+Not required for TypeScript.
+
+{{% /choosable %}}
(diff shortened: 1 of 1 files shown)


content/blog/pulumi-components/index.md
```diff
@@ -605,6 +605,98 @@ runtime: yaml
...
 
+### Define an Entry Point
+
+The entrypoint analyzes components to automatically build a schema, and interact with the Pulumi engine to mange the component lifecycle.
+
+{{< chooser language "typescript,python,go,csharp,java" >}}
+
+{{% choosable language typescript %}}
+
+Not required for TypeScript.
```
mikhailshilkov, line 616: Something like

```suggestion
No separate entry point is required for TypeScript. All exported component classes will automatically be exposed to be consumed from other languages.
```

?

content/blog/pulumi-components/index.md
```diff
@@ -605,6 +605,98 @@ runtime: yaml
...
+package main
+
+import (
+    "github.com/pulumi/pulumi-go-provider/infer"
+)
+
+func main() {
+    err := infer.NewProviderBuilder().
+            WithName("go-components").
+            WithNamespace("your-org-name").
```
mikhailshilkov, line 648: Minor but could drop `namespace` from this example.

mikhailshilkov (approved):
LGTM

content/blog/pulumi-components/index.md
meagancojocar: I grabbed this from docs so leaving now but we could drop in both if we want!

PR #14676, opened 2025-04-05
Improve the IaC AWS Get Started guide

joeduffy:
This streamlines, simplifies, and improves the IaC AWS Get Started guide. Our current one has been refactored a lot over time, leading to excessive introduction of distracting concepts, style and prose inconsistencies, bugs in the code (e.g., Java is currently broken), and generally speaking, something that I expect will be less effective in helping users get up and running compared to the new guide included in this pull request.

Some examples of changes:

- Use a consistent "Get started with Pulumi and AWS" header.

- Make the left-nav actually work. Today it seems broken for some reason; it is unclear if that was intentional but my experience using the navigation is that it makes it far easier to know where you are in the broader context of the get started flow.

- Have you select the language on the first page and keep it consistent throughout the guide. This eliminates a lot of visual clutter with the dozens of language choosers that were there.

- Make it very clear what prerequisites are necessary up front.

- Improve the package manager links where possible.

- Ensure we have PowerShell equivalents of commands for end users on Windows.

- Improve the styling of the buttons with "Next: " prefix, as well as enabling pages to control the links to make them sound more exciting, such as "I'm ready to begin" for the first one, and "Congratulations!" for the final button.

- Streamline the installation instructions.

- Add a specific "Configure access" step that simplifies the presentation on how to configure AWS as much as possible. Add a step to validate your configuration works. Remove needless distractions like our self-fulfilling promotion of ESC too early (configuring OIDC is a major step function increase in complexity at this stage and most likely just leads to dropoff).

- Tighten up prose wherever possible to eliminate superfluous and overly flowery prose.

- Collapse redundancy. Today we have you update your program to add an object to your bucket on one page. Then you go to the next page to actually deploy it. Then you go to the next page to change your program yet again to host a website. Instead, we now have a single page for teaching the user how to "update" things, and it includes configuring the website, adding the object, and deploying it. This not only gets to the point much quicker but also frees up conceptual space in the get started tutorial to do more interesting things.

- Introduce a "create a component" step to turn your website into a component after deploying it. This demonstrates multiple interesting concepts, including sharing and reuse of an abstraction, as well as looping over a list of files to provision multiple bucket objects.

My intent is to A/B test this, however, we'll need to discuss the best way to go about that given the extensive changes.


content/docs/iac/get-started/aws/create-component.md
```diff
@@ -0,0 +1,1015 @@
...
+Using components here also has the benefit that, as the requirements for S3 websites changes, you can
+update the one component definition and have all uses of it benefit.
+
+### Define a new component
+
+To define a new component, create a class called `AwsS3Website` that derives from `ComponentResource`. It'll have a mostly-empty
+constructor to start with but you will add the AWS S3 resources to it in the next step. You'll also define the inputs for the
+component -- the `files` to add to the website -- and outputs -- a single property with the website `url`.
+
+To get going, open a new file {{< compfile >}} alongside {{< langfile >}} and add the following:
```
kramhuber, line 102:
```suggestion
To get going, create a new file {{< compfile >}} alongside {{< langfile >}} and add the following:
```

content/docs/iac/get-started/aws/create-component.md
```diff
@@ -0,0 +1,1015 @@
...
+        // Create an AWS resource (S3 Bucket)
+        const bucket = new aws.s3.BucketV2("my-bucket", {}, { parent: this });
+
+        // Turn the bucket into a website:
+        const website = new aws.s3.BucketWebsiteConfigurationV2("website", {
+            bucket: bucket.id,
+            indexDocument: {
+                suffix: "index.html",
+            },
+        }, { parent: this });
```
kramhuber, line 317:
```suggestion
     // Set the component as the parent
        }, { parent: this });
```
Do we want to add this code comment to the first instance of setting the resource parent so the user can connect change #2 to a specific code comment?

content/docs/iac/get-started/aws/create-component.md
```diff
@@ -0,0 +1,1015 @@
...
+{{% /choosable %}}
+
+This defines a component but it doesn't do much yet.
+
+### Refactor your code into the component
+
+Next, make four changes:
+
+1. Move all resources from {{< langfile >}} ino the component's constructor
+2. Change each to use the component [as the `parent`](/docs/iac/concepts/options/parent/)
```
kramhuber, line 289:
```suggestion
2. Change each resource to use the component [as the `parent`](/docs/iac/concepts/options/parent/)
```

content/docs/iac/get-started/aws/create-component.md
```diff
@@ -0,0 +1,1015 @@
...
+
+{{% notes type="warning" %}}
+
+Unfortunately, YAML lacks the language facilities to author components. Feel free to [skip ahead](/docs/iac/get-started/aws/destroy-stack/).
+
+{{% /notes %}}
+
+{{% /choosable %}}
+
+Ensure the original resources are deleted from the {{< langfile >}} so there aren't any duplicates.
```
kramhuber, line 762: May be worth showing what the file contents should look like

kramhuber, line 762: Does the paragraph below this line make it unnecessary?

content/docs/iac/get-started/aws/create-project.md
```diff
@@ -134,6 +227,177 @@ After the command completes, the project and stack will be ready.
...
+
+Now examine the code in {{< langfile >}}:
+
+{{% choosable language javascript %}}
+
+```javascript
+"use strict";
+const pulumi = require("@pulumi/pulumi");
+const aws = require("@pulumi/aws");
+const awsx = require("@pulumi/awsx");
```
kramhuber, line 263: Is the `awsx` import needed for JS/TS?

content/docs/iac/get-started/aws/configure.md
```diff
@@ -0,0 +1,110 @@
...
+- /docs/clouds/aws/get-started/begin/
+---
+
+## Configure access to AWS
+
+Pulumi's CLI needs access to your AWS account to manage cloud resources.
+
+If you've already <a href="https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-install.html" target="_blank">installed</a> and <a href="https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html" target="_blank">configured</a> the AWS CLI, Pulumi will respect and use your configuration settings.
+
+You must use an IAM user account that has [programmatic access](https://docs.aws.amazon.com/IAM/latest/UserGuide/security-creds-programmatic-access.html) with rights to deploy and manage the resources used throughout this tutorial.
```
kramhuber, line 31:
```suggestion
You must use an IAM user account that has [programmatic access](https://docs.aws.amazon.com/IAM/latest/UserGuide/security-creds-programmatic-access.html) with rights to deploy and manage S3 buckets.
```
As far as I could tell this is the only resource they interact with--may be helpful to just be explicit.

kramhuber (approved):
Went through the TS path 👍

content/docs/iac/get-started/aws/create-component.md
joeduffy: Yes, good point!

content/docs/iac/get-started/aws/create-project.md
joeduffy: It's not, but `pulumi new` templates put it there by default, so it's going to be there when you look at it.

content/docs/iac/get-started/aws/create-component.md
joeduffy: Great idea, will do this.

joeduffy:
> Went through the TS path 👍

Thanks; I went through all languages manually as part of this. That's how I found out about `Conafiguration`: https://github.com/pulumi/docs/blob/29a87bf9eff1a203d6208cc6b342857603b7d0fd/content/docs/iac/get-started/aws/deploy-changes.md?plain=1#L194

PR #14679, opened 2025-04-06
fix: correct incorrect link to autonaming doc

TechWatching:
### Proposed changes

Correct an incorrect link to the auto-naming section of the doc.


jkodroff (approved):
Thanks for fixing this!

PR #14692, opened 2025-04-07
Add details to package installation

jkodroff:
- Correct header indentation levels.
- Add details to what gets committed to version control (driven by a customer question).
- Add information on how to store packages in the project configuration file to avoid the need to commit generated code to version control.

### DIFF
--- a/content/docs/iac/concepts/resources/providers.md
+++ b/content/docs/iac/concepts/resources/providers.md
@@ -68,7 +68,7 @@ You can add a list of your named provider packages under the [`packages`](/docs/
 
 ```yaml
 packages:
-  random: # This is an arbitrary name
+  random:
     source: terraform-provider
     parameters:
       - hashicorp/random
--- a/content/docs/iac/concepts/resources/providers.md
+++ b/content/docs/iac/concepts/resources/providers.md
@@ -70,9 +70,13 @@ You can add a list of your named provider packages under the [`packages`](/docs/
 packages:
   random:
     source: terraform-provider
+    version: 0.10.0 # The version of terraform-provider in the Pulumi registry
     parameters:
       - hashicorp/random
+      - 3.7.1 # The version of hashicorp/random in the OpenTofu Registry
 ```
+
+The versions specified above are optional, but recommended in order to ensure consistent behavior on all systems.
 
 You can install any packages tracked in the project configuration file with the [`pulumi install`](/docs/iac/cli/commands/pulumi_install/) command:
 
--- a/content/docs/iac/concepts/resources/providers.md
+++ b/content/docs/iac/concepts/resources/providers.md
@@ -40,7 +40,7 @@ The most common method of installing a provider is to use your language's packag
 
 After installing the provider using your package manager, you reference the provider in your Pulumi program to define the desired state of the resources for that provider. When you install the SDK for a provider (e.g., via `npm install <package_name>` in Node.js), the package manager (npm in this example) automatically downloads and installs the provider executable along with the SDK if the executable is not already cached on your system.
 
-## Installing a Parameterized Provider via `pulumi package add`
+### Installing a Parameterized Provider via `pulumi package add`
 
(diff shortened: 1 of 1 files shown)


content/docs/iac/concepts/resources/providers.md
```diff
@@ -50,7 +50,42 @@ For example, to generate a local SDK for the [`hashicorp/random` provider](https
...
+Adding provider packages to your project configuration file requires Pulumi version 3.157.0 or later.
+{{% /notes %}}
+
+You can avoid the need to commit any generated SDK files to version control by adding your parameterized packages to your Pulumi project configuration file (`Pulumi.yaml`).
+
+You can add a list of your named provider packages under the [`packages`](/docs/iac/concepts/projects/project-file/#packages-options) key. For the example in the previous section, the syntax would be:
+
+```yaml
+packages:
+  random: # This is an arbitrary name
```
Frassle, line 71: Its only arbitrary for "non-local" providers. If your using "source based providers" i.e. where you've checked the provider source code in side-by-side with your project then this _must_ match the provider name.

Not sure there's a good way to put that in here but figured worth pointing out that detail.

content/docs/iac/concepts/resources/providers.md
```diff
@@ -50,7 +50,42 @@ For example, to generate a local SDK for the [`hashicorp/random` provider](https
 $ pulumi package add terraform-provider hashicorp/random
 ```
 
+{{% notes type="tip" %}}
 In order to make sure Pulumi users are aware of the Any Terraform Provider's capabilities, Pulumi has included select, popular providers that can be consumed in Pulumi via the Any Terraform Provider in the Pulumi Registry, such as [The Honeycomb provider](/registry/packages/honeycombio/).
+{{% /notes %}}
+
+The generated SDK will include a `.gitignore` so it can be safely committed to version control without including all of the SDK's dependencies. The SDK contains instructions to download the provider binary to a shared (and cached) location on your local system outside of the working directory - the binary is not committed to version control.
```
mikhailshilkov, line 57: "instructions" feels like something the user has to do. Maybe "metadata" or rephrase entirely?

content/docs/iac/concepts/resources/providers.md
mikhailshilkov: It's not arbitrary: `package add` will add it as `random` and that's what the user should use. We should not encourage arbitrary names here.

content/docs/iac/concepts/resources/providers.md
```diff
@@ -50,7 +50,42 @@ For example, to generate a local SDK for the [`hashicorp/random` provider](https
...
+
+You can avoid the need to commit any generated SDK files to version control by adding your parameterized packages to your Pulumi project configuration file (`Pulumi.yaml`).
+
+You can add a list of your named provider packages under the [`packages`](/docs/iac/concepts/projects/project-file/#packages-options) key. For the example in the previous section, the syntax would be:
+
+```yaml
+packages:
+  random: # This is an arbitrary name
+    source: terraform-provider
+    parameters:
```
mikhailshilkov, line 73: Users can also specify a `version` of the provider and we should probable encourage doing this. Version is applicable both for `terraform-provider` and `hashicorp/random` (they are different versions).

mikhailshilkov (commented):
Thank you for expanding the docs! A couple of minor notes below.

content/docs/iac/concepts/resources/providers.md
jkodroff:
```suggestion
The generated SDK will include a `.gitignore` so it can be safely committed to version control without including all of the SDK's dependencies. The SDK installation process also downloads the provider binary to a shared location on your local system outside of the working directory. The binary is cached, so it will not need to be downloaded more than once, and is not committed to version control.
```

content/docs/iac/concepts/resources/providers.md
jkodroff: @mikhailshilkov Can you verify this is now correct?

mikhailshilkov (approved):
LGTM, thank you!

kramhuber (approved):

thoward (approved):

PR #14701, opened 2025-04-08
Update BLOGGING.md

asafashirov:
Added note about adding canonical tags when announcing products and features with an associated docs page, which should take precedence in the context of external search engines.


jkodroff (approved):
This explanation is perfectly clear. Thanks!

PR #14729, opened 2025-04-10
Copilot V2 new default, rm Copilot V1

foot:
### Proposed changes

- Copilot V2 is ready for production
- V2 is new default
- Remove V1 support

Also
- Remove old greenbook flag


sean1588:
this looks good to me @foot !

Issue #14326, opened 2025-03-09
Copilot docs on REST API response format incorrect for content


joeduffy:
The Copilot REST API docs [say that `content` is a `string`](https://www.pulumi.com/docs/pulumi-cloud/copilot/api/).

In the case of code, it is not, however. For example, I am seeing an object:

```
        {
            "role": "assistant",
            "kind": "program",
            "content": {
[...]

PR #14734, opened 2025-04-10
Documentation for code gen

arturl:
This PR documents the response format for the case when the caller is asking to generate code. The `content` field now becomes an object.

Fixes https://github.com/pulumi/docs/issues/14326


sean1588 (approved):
thanks Artur! this looks good to me!

PR #14769, opened 2025-04-17
Add example for aws-secrets to show kv pairs.

jkodroff:
This was in response to a customer use case, but it's also probably more common than scalar ("plain text" in AWS parlance) secrets.


jkodroff:
@pgavlin @nyobe @komalali Whoever gets to this first, that's sufficient. Just wanted to run it by someone on the dev team.

PR #14813, opened 2025-04-23
Suggested improvements to the PulumiUP page

joeduffy:
* We are launching new products! This is exciting! But you wouldn't know it from the current page. Instead, "Lap Around Pulumi" sounds like a rehash of last year's product demos. Let's spruce this up.

* Similarly, we can tighten up the overview message, with a highlight on *excitement*.

* Finally, I don't know about you, but the first thing I like to see when considering an event is the agenda. Right now it's buried at the bottom. This change pulls it up closer to the fold.

### DIFF
--- a/content/pulumi-up/_index.md
+++ b/content/pulumi-up/_index.md
@@ -122,7 +122,7 @@ schedule:
       time: "Europe & Americas"
       time2: "Asia Pacific"
 
-    - title: A Lap Around Pulumi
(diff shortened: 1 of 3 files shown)


shughes26 (approved):
Makes sense to me! I added a commit to also make that change on the post-reg confirmation page

joeduffy:
Perfect, thanks Sarah 🙌

PR #14833, opened 2025-04-24
Update STYLE-GUIDE.md

asafashirov:
Added note to avoid emojis.


gmiranda23:
Thanks @asafashirov. That looks good to me!

PR #14862, opened 2025-04-25
Update Explicit/Default Providers

jkodroff:
* Add explanation on the difference between default/explicit providers and when to use each one
* Remove `$` from bash codeblocks where we do not show the CLI output (this is an MDlint rule).


content/docs/iac/concepts/resources/providers.md
```diff
@@ -173,9 +196,11 @@ var instance = new Instance("myInstance",
 
 {{< /chooser >}}
 
-It creates a single EC2 instance in the us-west-2 region.
+It creates a single EC2 instance in the us-west-2 region, no matter what the `AWS_REGION` environment variable may be on your system.
+
+### Explicit Provider Configuration
 
-## Explicit Provider Configuration
+Because explicit providers are Pulumi resources themselves and take Pulumi inputs as values, they enable scenarios like creating a Kubernetes cluster and deploying resources onto that cluster in the same Pulumi program: The kubeconfig of the newly created cluster (a Pulumi output) can be passed to the `kubeconfig` argument of an explicit Kubernetes provider (a Pulumi input). The explicit provider can then be assigned to resources that should be deployed to the newly provisioned cluster. This scenario is _only_ possible by using explicit providers: We cannot use the default Kubernetes provider because we don't know what its kubeconfig should be until after the cluster is created.
```
mikhailshilkov, line 203: This is quite dense. Consider breaking down, e.g.

```suggestion
Explicit providers are Pulumi resources themselves and take Pulumi inputs as configuration values. This enables powerful scenarios that aren't possible with default providers. For example, you can create a Kubernetes cluster and then immediately deploy resources to that cluster in the same Pulumi program. This works because the kubeconfig of the newly created cluster (a Pulumi output) can be passed directly to the `kubeconfig` argument of an explicit Kubernetes provider (which accepts Pulumi inputs). The explicit provider can then be assigned to resources that should be deployed to the newly provisioned cluster. This scenario is _only_ possible by using explicit providers: We cannot use the default Kubernetes provider because we don't know what its kubeconfig should be until after the cluster is created.
```

content/docs/iac/concepts/resources/providers.md
```diff
@@ -91,17 +91,40 @@ Installing package 'random'...
...
+Within a Pulumi program, there are two types of providers you can use to declare resources:
 
-For example, suppose you run this CLI command:
+- **Default providers** are not declared in your Pulumi program, and use global configuration settings. Resources created with a default provider do not need the [`provider` option](/docs/iac/concepts/options/provider/) set in the [resource's options parameter](/docs/iac/concepts/options/). This is the most common way to declare Pulumi resources.
+- **Explicit providers** are explicitly declared in your Pulumi program, and use the configuration values you specify when you declare the provider. Resources created with explicit providers must have the `provider` option set in their resource options. The most common use case for explicit providers are multi-environment deployments of cloud infrastructure in a single stack. Explicit providers are themselves Pulumi resources, and their configuration values are [Pulumi inputs](/docs/iac/concepts/inputs-outputs/).
```
mikhailshilkov, line 99: Not required, but consider adding a simple table that summarizes:

Default Providers vs. Explicit Providers
When to use each
How to configure each
Limitations of each

It would repeat the text but also drum up the most important points.

content/docs/iac/concepts/resources/providers.md
```diff
@@ -91,17 +91,40 @@ Installing package 'random'...
...
 {{% /notes %}}
 
-## Default Provider Configuration
+## Default and Explicit Providers
 
-By default, each provider uses its package’s global configuration settings, which are controlled by your stack’s configuration. You can set information such as your cloud provider credentials with environment variables and configuration files. If you store this data in standard locations, Pulumi knows how to retrieve them.
+Within a Pulumi program, there are two types of providers you can use to declare resources:
 
-For example, suppose you run this CLI command:
+- **Default providers** are not declared in your Pulumi program, and use global configuration settings. Resources created with a default provider do not need the [`provider` option](/docs/iac/concepts/options/provider/) set in the [resource's options parameter](/docs/iac/concepts/options/). This is the most common way to declare Pulumi resources.
```
mikhailshilkov, line 98: The "most common way" sentence hangs in the air a bit. Why is most common?

content/docs/iac/concepts/resources/providers.md
```diff
@@ -517,23 +542,22 @@ final var myresource = new MyResource("myResource",
 
 ## Disabling Default Providers
 
-While default providers are enabled by default, they [can be disabled](/docs/concepts/config#special-configuration-options) on a per stack basis. Disabling default
-providers is a good idea if you want to ensure that your providers must be explicitly configured and should never use the default system configuration. (The meaning of "default system configuration" depends on the provider: it may be environment variables which can differ between environments, or a configuration file in a default location, and so on.)
+While default providers are enabled by default, they [can be disabled](/docs/concepts/config#special-configuration-options) on a per stack basis. Disabling default providers is a good idea if you want to ensure that your providers must be explicitly configured and should never use the default system configuration. (The meaning of "default system configuration" depends on the provider: it may be environment variables which can differ between environments, or a configuration file in a default location, and so on.)
```
mikhailshilkov, line 545: Mention the common mistake of not passing the right resource option to set the explicit provider and not noticing the error until it's too late?

mikhailshilkov (approved):
Nice change, thank you for clarifications and expanding the doc page! I left a few optional suggestions - feel free to use or ignore.

content/docs/iac/concepts/resources/providers.md
jkodroff: Added

content/docs/iac/concepts/resources/providers.md
jkodroff: Changed to "simple", which still might hang in the air, but less so. I figure with the table that breaks things down, we provide enough context for "simple" to work here.

Issue #14881, opened 2025-04-29
Document snake_case to camelCase conversion for Python components


julienp:
When authoring a component in Python, its properties will be converted from `snake_case` to `camelCase` in the schema.

PR #14882, opened 2025-04-29
Document snake_case to camelCase conversion for Python components

julienp:
### Proposed changes

When authoring a component in Python, its properties will be converted from `snake_case` to `camelCase` in the schema.

### Related issues (optional)

Fixes https://github.com/pulumi/docs/issues/14881


julienp:
I am not sure where to best place this information. I added it to the arguments section since that's where this issue first shows up, and it has a handy example with `index_content`.

PR #14907, opened 2025-05-01
Add property search and adjusting resource query limit for team sku

DukeBWard:
### Proposed changes

* `content/pricing/_index.md` : add check for property search for team and adjusting resource query limit from 2k -> 10k
* `content/docs/insights/search.md` : adjusted property search tool tip from only enterprise and biz crit to team, enterprise and biz crit

N/A

### Related issues (optional)

Reference https://github.com/pulumi/pulumi-service/issues/27667
image
image

### DIFF
--- a/content/docs/insights/search.md
+++ b/content/docs/insights/search.md
@@ -271,7 +271,7 @@ Ranges can also be one-sided. For example, to query everything modified after Ja
 ### Property Queries
 
 {{% notes "info" %}}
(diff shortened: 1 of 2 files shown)


thoward (approved):

AnuiDolphin:
We should remove the 'Per query limits' row completely. While 'per query limits' is technically true, you can work around this by grouping your results, so from a user perspective, it is 10k per group. This is too much to describe in a pricing page now that the limits are the same across all SKUs.

PR #14923, opened 2025-05-05
Initial docs for Visual Import

AnuiDolphin:
### Proposed changes

This change adds a page to the Pulumi-cloud docs to describe how to use Visual Import


arunkumar611:
A few high-level comments
1. Should this be under Pulumi IaC? Maybe we call it "Migrate to IaC" for folks who might not know what Import is, and then point to this page
2. Having an insights account set up with existing cloud resources is a pre-req, which we haven't called out here

content/docs/pulumi-cloud/import/_index.md
```diff
@@ -0,0 +1,108 @@
...
+
+![Pulumi IaC Import](import-review-enhance.png)
+
+## **Step 5: Integrate Code**
+
+Once the code is in the format you are happy with, you can:  
+
+* **Copy Code**: Copies the code to your clipboard
+* **Download Program**: Saves the code as a file to your local system
+
```
arunkumar611, line 88: Once I get here. It really is unclear for a new user what I need to do here. 

Should I be doing the following steps
1. pulumi new for the specific language in your CLI
2. then copy the code 
3. then run pulumi update?

arunkumar611: We put running a pulumi preview in best practice, maybe we move it up here as a step before running pulumi up?

PR #14939, opened 2025-05-06
Add next step to Services from Workflow (IDP)

meagancojocar:


kramhuber:
@meagancojocar you can close, I tackled as part of https://github.com/pulumi/docs/pull/14940

PR #14945, opened 2025-05-06
Replace registration CTA with 'Watch' CTA and youtube live link

shughes26:


isaac-pulumi:
@shughes26 - can we also replace the YouTube embed with the 2025 livestream embed?

isaac-pulumi (approved):

PR #15008, opened 2025-05-12
Move the Stack Permissions page under IAM

jkodroff:
Stack permissions are more commonly set up alongside teams and groups (per a customer call I was on today), so this change moves the content on permissions from the single, large stacks page and places it in a new page under IAM.

FYI, this change was authored via Claude code.


CLAUDE.md
thoward, line 1: Big step forward here.

thoward (approved):
Nice job!

PR #15025, opened 2025-05-13
Make external links open in new tabs

asafashirov:
## Summary
- Add functionality to make external links open in new tabs while keeping internal links as is
- Reduce bounce/exit rates by keeping users on pulumi.com when following external links
- Allow users to easily navigate back to pulumi.com without losing their place
- Add security protection with rel='noopener' to prevent new tab from accessing opener window

## Implementation
- Add a new TypeScript file that runs when the page loads
- Detect all links and check if they point outside the pulumi.com domain
- Add target='_blank' attribute only to external links
- Keep internal navigation behavior unchanged

## Test plan
1. Load any page with external links
2. Click on external links to verify they open in new tabs
3. Click on internal links to verify they stay in the same tab
4. Check browser console for any errors


thoward (changes requested):
@asafashirov This PR seems to include two different changes: IndexNow and also external link behavior changes. Can you split these into separate PRs?

asafashirov:
@thoward i must of branched off a previous pr's branch instead of `main`. i couldn't remove the indexnow commit from this pr so closing this one and opened a new one: https://github.com/pulumi/docs/pull/15080

PR #15026, opened 2025-05-13
Update Josh Kodroff's headshot

jkodroff:
Handsome fella.


.gitignore
```diff
@@ -66,4 +66,6 @@ _vendor/
 .buildinfo
 
 # VScode solution files.
-*.sln
\ No newline at end of file
+*.sln
+
+.claude
```
shughes26, line 71: yeah, probably a good plan

shughes26 (approved):

.gitignore
meagancojocar: I went to review this but wasn't sure about this git ignore, @thoward was going to confirm what we want to do with claude files and settings

shughes26: All this is doing is leaving a user's personal claude settings out of commits for now, which will help local settings not get overwritten and/or conflicted between branches. Safe to leave it in for sure, also safe to leave it out

PR #15028, opened 2025-05-13
First pass of breaking out REST API docs

jkodroff:
For the first pass, we'll break out the REST API docs into individual pages without paying much attention to how they relate to individual resources. We'll also order them alphabetically.

Note that this change is to be merged into a larger feature branch once everything is good.


jkodroff:
Previews are not supported for feature branches, so I opened another PR to master.

Issue #14856, opened 2025-04-25
Merge Deployments REST API docs into the main Pulumi Cloud REST API docs


jkodroff:
We have a page for Deployments' REST API reference <https://www.pulumi.com/docs/pulumi-cloud/deployments/api/> which is separate from the rest of the Pulumi Cloud API reference <https://www.pulumi.com/docs/pulumi-cloud/reference/cloud-rest-api/>. We should merge the Deployments REST API into the Pulumi Cloud REST API for consistency.

* First, we should reformat the page in-place so that it looks like the other pages in the Pulumi Cloud REST API:
  * We don't need the Authentication section - that's now redundant.
  * Add an introductory blurb at the top of the page that explains what this part of the REST API does.
  * All the stuff under "Deployment Settings" is analogous to Schema Definitions under "Resource Search" in the main REST API, so that should go last in the page and be titled "Schema Definitions" for consistency.
  * The formatting of the various REST operations should match what we have elsewhere in the rest of the API docs.
* Create a new page called "Deployments" in the main Pulumi Cloud REST API section of the docs. Ensure that it's placed in alphabetical order.
* Any links to the Deployments REST API page should be rewritten to point to the new location.
* The original page should be removed as it's no longer needed.

PR #15030, opened 2025-05-13
First pass of breaking out REST API docs

### LINKED ISSUE #11525: Re-organize our REST API docs
Problem: Our current REST API docs are too long for a single page making it harder to navigate. 

One solution: Organize the API docs like this:

- REST API Docs
    - Overview / Get started
    - Stacks
        - List/Get/Delete Stack
        - Stack Tags
        - Stack Updates
    - Deployments
    - ESC (Environments)
    - Organizations & Tokens
    - Audit Logs
    - Resource Search
    - Other
        - RUM
        - Data export
        - Schemas

jkodroff:
Fixes #11525 

Summary of changes:

- Broke out the intro material into a "basics" page that talks about authentication, etc.
- Each h2 in the old, long, single file has been broken out into its own file.
- Each file has some header material stating what the "resource" (it's not exactly aligned to resources, but close enough) is about
- Pages have been ordered alphabetically after "basics"

Given that there's at least plans in the works to generate a proper single source of truth, we should probably not put a ton of effort into getting this perfect, but this at least makes the docs more discoverable via the left nav and gives us an obvious structure so we can do #14856 later and do other smaller improvements.


djgrove (commented):
One concern: can we get redirects working from the old page? It would be ideal (and fantastic) if deep links to the old doc still take you to the correct doc

djgrove (approved):

PR #15063, opened 2025-05-15
Add example and description of provider terraform config

guineveresaenger:
Adds info on provider.terrafromConfig() method. 

Also a few wording tweaks

### DIFF
--- a/content/blog/announcing-direct-tf-modules/index.md
+++ b/content/blog/announcing-direct-tf-modules/index.md
@@ -12,8 +12,9 @@ tags:
   - features
 ---
 
-We are excited to announce that Pulumi can now execute [Terraform modules
-directly](https://github.com/pulumi/pulumi-terraform-module). This new capability unlocks a great option for users
+We are excited to announce that Pulumi now has support for [Terraform modules](https://github.com/pulumi/pulumi-terraform-module).
+
+This new capability unlocks a great option for users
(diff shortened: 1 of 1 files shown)


content/blog/announcing-direct-tf-modules/index.md
```diff
@@ -197,12 +198,35 @@ Duration: 2m26s
...
-The above program is very simple. To take it further, check out
+The above program is a very basic use case. To take it further, check out
 [examples](https://github.com/pulumi/pulumi-terraform-module/tree/main/examples) for more realistic programs showcasing
 features such as computing subnets dynamically with Pulumi `aws.getAvailabilityZonesOutput` function or passing the
 results of the VPC module to an EKS module.
 
-// TODO show-case cross-configuring the provider with a given AWS region
+### Passing through existing configuration
+
+As an example how the new Pulumi Terraform Module can interact with existing Pulumi programs, we've made it possible
```
t0yv0, line 208: This one's interesting there's a lot going on here and it can be confusing that both Pulumi and TF providers are involved and that you need to configure both. Perhaps find a quick way to call this out? 

With the caveat of it maybe being too verbose, for example start by showing how you can configure the AWS region for the TF provider using the explicit provider. And then continue with saying if you already have a *Pulumi* AWS provider you can reuse the region and everything else from that provider.

t0yv0:
Incorporated thank you!

PR #15066, opened 2025-05-15
Merge Deployments REST API docs into the main Pulumi Cloud REST API docs

### LINKED ISSUE #14856: Merge Deployments REST API docs into the main Pulumi Cloud REST API docs

jkodroff:
Fixes #14856 

- Move Deployments REST API content from its own page to the REST API reference
- Updated all links to the API reference to point to the new location
- Added Deployments section to the main REST API index page


content/docs/pulumi-cloud/deployments/api.md
thoward, line 1: Small note here: not sure why it's showing up in the git log as a delete/add of content, vs a rename/move. Did you use `git mv` then edit the file in its new location, or was this more of a "move the file on bash, edit in the new location, git add ." sort of workflow? Only reason I ask is that I think it would be easier to review if git knew it was the same file in the change log and that the move and edit were recorded separately, so that it could show a diff of changes to the file, as well as the rename/move. This could just be a limitation of git/github but especially in cases like this, when it's a 1500 line file, in order to review properly, i need to compare the two for content changes and that means just reading two very long documents and hoping that i notice and discrepancies. not ideal.

content/docs/pulumi-cloud/reference/deployments/_index.md
```diff
@@ -0,0 +1,1043 @@
+---
+title: Deployments
+title_tag: "Pulumi Cloud REST API: Deployments"
+meta_desc: Learn about the Pulumi Cloud REST API endpoints for configuring and managing Pulumi Deployments to automate infrastructure updates.
+menu:
+    cloud:
+        parent: pulumi-cloud-reference
+        weight: 4.5
```
thoward, line 8: Per Hugo docs, this should be an integer. Not sure how this will sort. Likely it will interpret both this page and the other page that has `weight: 4` as being at the same weight, then do a secondary lexicographical sort? Anyway, for consistent ordering each page needs to be an integer, which does make it a PITA to insert a new page in a large list of ordered pages like this because you have to go update all of the other pages if there weren't any numerical gaps between them.

thoward: https://gohugo.io/content-management/menus/#weight

content/docs/pulumi-cloud/reference/deployments/_index.md
```diff
@@ -0,0 +1,1043 @@
...
+    "enable": "true"
+  }
+}
+```
+
+#### Properties
+
+| Name      | Type    | Description                                      |
+|-----------|---------|--------------------------------------------------|
+| `enable`  | boolean | **Required.** Whether to use dependency caching. |
```
thoward, line 1043: This file is about 500 lines shorter than the original, and seems to be in a different ordering/format. Given the lack of ability to diff these changes, it might take some time to review this manually to ensure there was no loss of information during this conversion. I'm not a subject matter expert on this API. Should we ask an engineer from the team to review this?

jkodroff: @komalali Can you review these changes and/or do a spot check?

@thoward Alternatively, we can agree that the risk of a potential error in the content (which can be corrected later) is worth the benefit of getting our REST API docs in one place.

content/docs/pulumi-cloud/reference/deployments/_index.md
jkodroff: @thoward Are you ok with re-doing the weights as integers (multiples of 10) in a separate PR? It does display correctly for me.

thoward (approved):
Rubber stamping w/ caveat that we probably still need a SME to review the diff.

PR #15084, opened 2025-05-16
Remove duplicate Vault OIDC setup page from generic OIDC setup in Pulumi Cloud

jkodroff:
Valut OIDC setup is only relevant to Pulumi ESC, and the Pulumi ESC section of the site already has its own setup documentation, so this commit removes the duplicated page and adds an alias to the ESC page for Vault OIDC setup.


jkodroff:
Converting to draft so we can wait until #15084 is merged since this PR ignores the Vault page.

jkodroff:
For context, the overall issue is that we have:

- A complete set of maintained set of docs that describe OIDC setup for ESC in the ESC section of the docs.
- Pages under Pulumi Cloud that describe OIDC setup for both Deployments and ESC in one doc. These are difficult to find if you're trying to set up OIDC for Deployments OR ESC, and difficult to read, and in some cases link to outdated information (a Pulumi IaC program that does not have the right audience claim for ESC since we came up with ESC projects). This set is also incomplete, e.g. does not include Infiniscal. 

I am trying to get the OIDC setup for Deployments actually under Deployments in the nav, and have the content in those docs pertain _only_ to Deployments (with clear recommendations that using ESC in a stack is better way to do it).

borisschlosser (approved):
If your goal is to remove Vault from access management docs I guess you probably also want to remove this line: https://github.com/pulumi/docs/blob/master/content/docs/pulumi-cloud/access-management/oidc/provider/_index.md?plain=1#L84

From ESC perspective this change looks good to me.

PR #15124, opened 2025-05-21
Add "Use a Terraform module" docs page

thoward:
This adds a new page "Use a Terraform Module" under the "Extending Pulumi" heading. It also updates other related pages to reference this new feature.

### DIFF
--- a/content/docs/iac/adopting-pulumi/migrating-to-pulumi/from-terraform.md
+++ b/content/docs/iac/adopting-pulumi/migrating-to-pulumi/from-terraform.md
@@ -23,6 +23,7 @@ If your infrastructure was provisioned with Terraform, there are a number of opt
 * **Coexist** with resources provisioned by Terraform by referencing a `.tfstate` file.
 * **Import** existing resources into Pulumi [in the usual way](/docs/using-pulumi/adopting-pulumi/import/) or using `pulumi convert --from terraform` along with some `pulumi import --from terraform` to adopt all resources from an existing `.tfstate` file.
 * **Convert** any Terraform HCL to Pulumi code using `pulumi convert --from terraform`.
+* **Use Terraform Modules** directly within your Pulumi programs through the [Terraform Module](docs/iac/using-pulumi/extending-pulumi/use-terraform-module/) feature.
 
 This range of techniques helps to either temporarily or permanently use Pulumi alongside Terraform, in addition to fully migrating existing infrastructure to Pulumi.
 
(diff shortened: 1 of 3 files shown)


content/docs/iac/using-pulumi/extending-pulumi/use-terraform-module.md
```diff
@@ -0,0 +1,283 @@
...
+
+***Example:** Pulumi.yaml*
+
+```yaml
+name: rds-example
+runtime:
+  name: nodejs
+  options:
+    packagemanager: npm
+packages:
```
t0yv0, line 101: TIL this shipped in latest Pulumi CLI. Lovely! It does seem to work.

content/docs/iac/using-pulumi/extending-pulumi/use-terraform-module.md
```diff
@@ -0,0 +1,283 @@
...
+    input: cidr,
+    newbits: 8,
+    netnum,
+  }).result;
+}
+```
+
+In the above code, the imported Terraform module works the same as any other Pulumi code. Outputs are returned, and resource state is stored in your Pulumi state storage, alongside all your other Pulumi-native resources. This also means that resource dependencies work as expected between Pulumi-native resources and resources created by Terraform modules.
+
+## Configuring Terraform Providers
```
t0yv0, line 207: It can be confusing that Terraform providers are not the same as Pulumi bridged providers and need to be separately configured.

One cool feature possibly worth mentioning is that if you do have a Pulumi bridged provider configured you can reuse that configuration for the Terraform provider as well.  AS of today this works:


```
import * as pulumi from "@pulumi/pulumi";
import * as vpc from "@pulumi/vpc";

const awsProvider = new aws.Provider("aws-provider", {region: "us-west-2"});

const provider = new vpc.Provider("test-provider", {
    aws: awsProvider.terraformConfig().result,
});

// Use the provider with the module
const testBucket = new vpc.Module("test-bucket", {
    azs: [],
}, { provider: provider });

```

content/docs/iac/using-pulumi/extending-pulumi/use-terraform-module.md
```diff
@@ -0,0 +1,283 @@
...
+## How It Works
+
+Pulumi's Terraform Module provider allows you to consume Terraform modules as if they were native Pulumi packages. It works by:
+
+1. Automatically installing and managing [OpenTofu](https://opentofu.org/) (an open-source Terraform-compatible implementation) to execute the module.
+2. Translating Pulumi resource declarations to Terraform configurations.
+3. Managing state through your standard Pulumi state backend.
+4. Exposing module outputs as native Pulumi outputs.
+
+## Getting Started
```
t0yv0, line 46: We have an emerging `pulumi convert` capability that lets you target the modules. It's coming a little hot though so we can possibly do a quick addition  to the docs after this PR.

content/docs/iac/adopting-pulumi/migrating-to-pulumi/from-terraform.md
```diff
@@ -498,3 +499,178 @@ Before running the deployment the import file generated will be written out to t
 ### Example Conversion
 
 For an example of a full conversion, see the [Converting Full Terraform Programs to Pulumi](/blog/converting-full-terraform-programs-to-pulumi/) blog post.
+
+## Using Terraform Modules Directly
+
+Pulumi allows you to use existing Terraform modules directly in your Pulumi programs without converting or rewriting them. This feature is particularly useful for:
```
t0yv0, line 505: I wonder if this is worth bringing up or it is just a pedantic curiosity - a plain file with `tf` declarations is technically a module with no inputs and outputs. So any existing Terraform code even if it does not use modules apriori can be integrated into Pulumi by quickly turning it into a module by adding varables.tf and outputs.tf.

t0yv0 (approved):
Looking good!

PR #15128, opened 2025-05-21
Update K8s landing page with new video

adamgordonbell:
Adding K8S video to landing page.
Not sure if there are other places where we talk about K8S provider where it would be helpful as well.

### DIFF
--- a/content/docs/iac/clouds/kubernetes/_index.md
+++ b/content/docs/iac/clouds/kubernetes/_index.md
@@ -11,7 +11,6 @@ menu:
     parent: iac-clouds
     weight: 1
 cloud_overview: true
-
 description: Streamline Kubernetes cluster configuration, management, and application workload deployments using TypeScript, Python, Go, C#, Java or YAML. Use the Pulumi Kubernetes Operator to manage both Kubernetes and cloud resources.
 get_started_guide:
   link: /docs/iac/get-started/kubernetes/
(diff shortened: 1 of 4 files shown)


adamgordonbell:
Link:

http://www-testing-pulumi-docs-origin-pr-15128-224d132d.s3-website.us-west-2.amazonaws.com/kubernetes/

thtmnisamnstr:
On `/kubernetes/`, you can cut the whole "Deploy Kubernetes Workloads with Ease" section below the video. It looks out of place and doesn't really add a ton of value IMO.

It might be worthwhile to add this video to our k8s docs somewhere too.

adamgordonbell:
> It might be worthwhile to add this video to our k8s docs somewhere too.

I didn't find anywhere, where it fits in well. If we had a pulumi provider vs traditional K8S page or something it might fit there, but I don't think we do.

>you can cut the whole "Deploy Kubernetes Workloads with Ease" section below the video

Including the code? I liked the code, and that text section was sort of filler to make the two-column code layout work. I could try to get the code to be full width and below the logs perhaps?

thtmnisamnstr:
This all looks good. I wish there were some way to make the code block component look a little better when it's full width, but I think we just have to live with it as it is.

thtmnisamnstr (approved):

PR #15137, opened 2025-05-22
Add detailed Claude.md and Cursor/rules

meagancojocar:

### DIFF
--- a/.cursor/rules.mdc
+++ b/.cursor/rules.mdc
@@ -1,0 +1,245 @@
+---
+description: 
+globs: 
+alwaysApply: true
+---
+---
+description: Guidelines for contributing to Pulumi documentation and maintaining consistency across content, Hugo templates, and editor style
+globs: "**/*.md,**/*.html,**/*.js,**/*.ts,**/*.yaml,**/*.yml,**/*.json"
+alwaysApply: true
+---
+
+# Cursor Rules
+
+## Repository Overview
+
+- **Purpose:** Pulumi's Hugo-based documentation website at pulumi.com
+- **Technology Stack:** Hugo static site generator, Node.js/Yarn, TypeScript, Stencil web components, Tailwind CSS
+- **Content Types:** Documentation, blog posts, tutorials, examples, landing pages
+
+## Documentation Standards
+
+### **Language and Terminology**
+
+- Use `pulumi` or "the Pulumi CLI" to refer to the CLI
+- Use `Pulumi Cloud` to refer to the infrastructure management platform  
+- **Avoid ableist language:** Use "wild" instead of "crazy", "select" instead of "click", "placeholder" instead of "dummy"
+- **Avoid unnecessarily gendered language:** Use "folks", "y'all", or "everyone" instead of "guys"
+- **Avoid violent language** and pop-culture references
+- Use "navigate" instead of "go to"
+- Avoid directional words - link directly to sections
+- **Never use "easy" or "simple"** - they make assumptions about reader expertise
+
+### **Content Structure**
+
+- **One `h1` per page** - every page must have exactly one
+- **Heading hierarchy:** Only increment one level at a time (h2 → h3, never h2 → h4)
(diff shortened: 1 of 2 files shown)


.cursor/rules.mdc
```diff
@@ -0,0 +1,245 @@
...
+#### **Video Embeds**
+```markdown
+{{< youtube "video-id?rel=0" >}}
+```
+
+- Always use `?rel=0` for Pulumi YouTube videos to limit suggested videos
+
+#### **Pulumi CLI Commands**
+
+- Use specific shortcodes for CLI references: `{{< pulumi-new >}}`, `{{< pulumi-apply >}}`, etc.
```
mikhailshilkov, line 112: What is pulumi-apply?

.cursor/rules.mdc
```diff
@@ -0,0 +1,245 @@
...
+- These auto-generate current CLI documentation
+
+### **Frontmatter Standards**
+
+#### **Required Fields**
+
+---
+title: "Page Title (60 char limit for SEO)"
+meta_desc: "SEO description"
+author: joe-duffy
```
mikhailshilkov, line 122:
```suggestion
author: author-id
```

as below?

.cursor/rules.mdc
```diff
@@ -0,0 +1,245 @@
...
+- `/assets/` - Built CSS/JS assets (checked in)
+
+### **Images and Media**
+
+#### **Blog Images**
+
+- Place images in same directory as blog post
+- Reference relatively: `![Alt text](image.png)`
+- **Social images:** 1200×628 PNG, opaque background
+- Use [Figma template](https://www.figma.com/file/TnD7nxjIxVvXq8w0W7awPG/Build-Your-Own-Meta-Image) for consistency
```
mikhailshilkov, line 179: Oh, that works?! Nice if so.

mikhailshilkov (commented):
Two files seem similar but different in many ways. How are we thinking about keeping them in sync vs. entirely independent?

jkodroff:
I don't think I have anything useful to add to this PR, so no need to wait on my review.

.cursor/rules.mdc
thoward: In this case, it's a short code for `Output.apply`. See: https://github.com/pulumi/docs/blob/master/layouts/shortcodes/pulumi-apply.html

What the short code does is select the language-specific casing of the apply function (e.g. `Apply` vs `apply`) and links to the SDK documentation for whatever language you've chosen, using the chooser framework under the hood.

It's used in these two docs:
- https://github.com/pulumi/docs/blob/ba1649a4acb055f666d35295d267d31b344a7b6b/content/docs/iac/concepts/inputs-outputs/all.md?plain=1#L19
- https://github.com/pulumi/docs/blob/ba1649a4acb055f666d35295d267d31b344a7b6b/content/docs/iac/concepts/inputs-outputs/apply.md?plain=1#L292

 This is a way to avoid the problem I kept running into, for example, in the "Build a Component" docs, where there are subtle variations on the links/names of functions/classes that require entire prose blocks to be within a chooser (and thus duplicated with minor variations in the Markdown). However, you need to write a custom shortcode like that for every scenario and it's not super obvious what the shortcode does. I didn't do that on the components docs, but the original author of the output/apply docs did do that for apply.

CLAUDE.md
thoward, line 1: Much of this file seems duplicative of the style guide, readme, and blogging guides. My understanding is that Claude doesn't need much more directive than "read the style guide" to get this information (e.g. it doesn't have to be duplicated within `CLAUDE.md`, just mentioned there). The concern I have is drift. If we duplicate (or ~mostly duplicate with some subtle variation) between CLAUDE.md and the style guide, which one is the source of truth? which one do we maintain? seems better to keep all this information contained in the style guide, and just ask Claude to use that. That will work for human readers and for agentic AIs equally.

meagancojocar:
we can close!

PR #15179, opened 2025-05-30
Fix missing global scope + templates source role name

glena:
### Proposed changes

Move insights account create as it is a global permission.
Fix templates  scopes are they didnt follow the right syntax


glena:
@isabeldepapel done

isabeldepapel (approved):

PR #15197, opened 2025-06-02
One more step needed when OAuth App Restrictions are active.

ringods:
### Proposed changes

When a customer has OAuth App access restrictions active in their Github organization, one more step
is required to give the Pulumi Github App access to the customer's repositories.

### Related issues (optional)

Detected via a customer support issue.


thoward (approved):

PR #15207, opened 2025-06-03
Remove references to obsolete repos

jkodroff:
- Remove refs in README.md to repos whose SDK docs have been moved to pulumi/registry
- Remove the same refs from run_typedoc.sh
- Remove the same refs from update_repos.sh
- Remove the unused workflow generate_provider_docs.yml


jkodroff:
Due diligence that the provider docs workflow is not used: https://github.com/pulumi/docs/actions/workflows/generate-provider-docs.yml

iwahbe (approved):

PR #15216, opened 2025-06-04
Improve private registry docs

danielrbradley:
Avoid directing users to use local paths when publishing because this doesn't work (see https://github.com/pulumi/pulumi/issues/19536).

Lead with basic git publish command

- Publish from git not a local path.
- Use the git protocol as, if you're using a sub-folder, it's not actually a valid web address.
- Mirror syntax as presented in the CLI docs.

Expand versioning

- Add detail on including "v" in tags but not when publishing.
- Add full example workflow for publishing a version via git and Pulumi.
- Detail the default version lookup process.

Remove local path examples

- Use git source for all examples as local sources don't work.

### DIFF
--- a/content/docs/idp/get-started/private-registry.md
+++ b/content/docs/idp/get-started/private-registry.md
@@ -36,9 +36,9 @@ pulumi package publish github.com/acme/k8s-cluster
 
 #### Component Versioning
 
-Pulumi uses git tags to infer the version of the component being published. The tag must adhere to [to the semantic versioning standard](https://semver.org/) plus a "v" prefix.
+Pulumi uses git tags for versioning components. By default, the latest version tag will be used. The tag must adhere to [to the semantic versioning standard](https://semver.org/) plus a "v" prefix (e.g. `v1.2.3`).
 
-To specify a custom version when publishing, append an `@` followed by the semver version (excluding the "v") after the git source. If no version is specified, the latest version tag will be used.
+To publish a specific version, append an `@` followed by the semver version (excluding the "v") after the git source.
 
 To create a version, push it to your "origin" git remote and publish to Pulumi's private registry, you'd run:
 
--- a/content/docs/idp/get-started/private-registry.md
+++ b/content/docs/idp/get-started/private-registry.md
@@ -22,6 +22,18 @@ Developers leverage templates and components in their preferred workflows, wheth
 
 If you're new to Pulumi components, the [Build a Component](/docs/iac/using-pulumi/extending-pulumi/build-a-component/) guide is a great resource for getting started. Once you've authored your component, push it to a GitHub or GitLab repository that Pulumi can access. Private repositories are supported, but the repository must be open to inbound requests.
 
+The `publish` CLI command is used to publish components to the private registry.
+
+```bash
+pulumi package publish <provider|schema>
+```
+
+For example, if your github organization is ACME and you are publishing the `k8s-cluster` component, you'd run:
+
+```bash
+pulumi package publish git://github.com/acme/k8s-cluster
+```
+
 #### Component Versioning
 
 Pulumi checks for a git version tag when the `publish` command is executed and stores it as the component version. The tag must adhere to [to the semantic versioning standard](https://semver.org/).
@@ -40,20 +52,6 @@ If you're part of multiple organizations and do not have a [default organization
 
 ```bash
 pulumi package publish /path/to/your/component --publisher ORG_NAME
-```
-
-#### CLI Publishing
-
-The `publish` CLI command is used to publish components to the private registry.
-
-```bash
-pulumi package publish COMPONENT_LOCATION
-```
-
-For example, if your github organization is ACME and you are publishing the `k8s-cluster` component, you'd run:
-
-```bash
-pulumi package publish https://github.com/acme/k8s-cluster
 ```
 
 #### Authenticating with Private Repositories
--- a/content/docs/idp/get-started/private-registry.md
+++ b/content/docs/idp/get-started/private-registry.md
@@ -36,7 +36,7 @@ pulumi package publish github.com/acme/k8s-cluster
 
 #### Component Versioning
 
-Pulumi uses git tags for versioning components. By default, the latest version tag will be used. The tag must adhere to [to the semantic versioning standard](https://semver.org/) plus a "v" prefix (e.g. `v1.2.3`).
+Pulumi uses git tags for versioning components. By default, the latest version tag will be used. The tag must adhere to [the semantic versioning standard](https://semver.org/) plus a "v" prefix (e.g. `v1.2.3`).
 
 To publish a specific version, append an `@` followed by the semver version (excluding the "v") after the git source.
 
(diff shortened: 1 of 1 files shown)


content/docs/idp/get-started/private-registry.md
```diff
@@ -22,38 +22,46 @@ Developers leverage templates and components in their preferred workflows, wheth
...
+```bash
+pulumi package publish <provider|schema>
+```
 
-A README is required when publishing a component. Pulumi renders markdown README files in the private registry. They're a great way to provide context for a component. By default, the Pulumi CLI looks for a README in the component's root directory. The `--readme` flag can be used to specify a custom source.
+For example, if your github organization is ACME and you are publishing the `k8s-cluster` component, you'd run:
 
 ```bash
-pulumi package publish /path/to/your/component --readme README_LOCATION
+pulumi package publish git://github.com/acme/k8s-cluster
```
flostadler:
```suggestion
pulumi package publish https://github.com/acme/k8s-cluster
```

content/docs/idp/get-started/private-registry.md
```diff
@@ -22,38 +22,46 @@ Developers leverage templates and components in their preferred workflows, wheth
...
 ```
 
-#### Specifying an Organization
+#### Component Versioning
 
-If you're part of multiple organizations and do not have a [default organization](/docs/iac/cli/commands/pulumi_org_set-default/) set, you must specify the org by using the `--publisher` flag.
+Pulumi uses git tags to infer the version of the component being published. The tag must adhere to [to the semantic versioning standard](https://semver.org/) plus a "v" prefix.
```
flostadler: It doesn't need the `v` prefix, does it?

content/docs/idp/get-started/private-registry.md
```diff
@@ -22,38 +22,46 @@ Developers leverage templates and components in their preferred workflows, wheth
...
-A README is required when publishing a component. Pulumi renders markdown README files in the private registry. They're a great way to provide context for a component. By default, the Pulumi CLI looks for a README in the component's root directory. The `--readme` flag can be used to specify a custom source.
+For example, if your github organization is ACME and you are publishing the `k8s-cluster` component, you'd run:
 
 ```bash
-pulumi package publish /path/to/your/component --readme README_LOCATION
+pulumi package publish git://github.com/acme/k8s-cluster
 ```
 
-#### Specifying an Organization
+#### Component Versioning
```
flostadler, line 37: I think we should point out that it's inferring the latest version when no version is provided

content/docs/idp/get-started/private-registry.md
```diff
@@ -22,38 +22,46 @@ Developers leverage templates and components in their preferred workflows, wheth
...
 ```
 
-For example, if your github organization is ACME and you are publishing the `k8s-cluster` component, you'd run:
+#### Specifying an Organization
+
+If you're part of multiple organizations and do not have a [default organization](/docs/iac/cli/commands/pulumi_org_set-default/) set, you must specify the org by using the `--publisher` flag.
 
 ```bash
-pulumi package publish https://github.com/acme/k8s-cluster
+pulumi package publish git://github.com/acme/k8s-cluster --publisher ORG_NAME
```
flostadler: Why the change to `git://`? Both work, but I feel like `https://` is more common

content/docs/idp/get-started/private-registry.md
danielrbradley: I think it's required for the git tag .. that's how it was done in the demo repo.

content/docs/idp/get-started/private-registry.md
danielrbradley: This was deliberate as the https path isn't always actually a valid web address (specifically when accessing sub-folders) - it's kinda our own scheme. Or perhaps we should be dropping the protocol altogether.

content/docs/idp/get-started/private-registry.md
danielrbradley: > If no version is specified, the latest version tag will be used.

content/docs/idp/get-started/private-registry.md
danielrbradley: If you're referring to a sub-folder, then the https address is a 404, but is the correct format for our plugin resolution .. which is confusing. Https makes it seem like you just copy the web address in, but that won't work.

danielrbradley: I've removed the protocol completely as it's irrelevant and just confuses things.

content/docs/idp/get-started/private-registry.md
danielrbradley: I've moved this sentence up a little to make it easier to spot.

content/docs/idp/get-started/private-registry.md
flostadler: Oh wild, it's actually failing without the `v`:
```
pulumi package publish github.com/flostadler/aws-k8s@0.0.21-beta.1 --publisher initech-co
warning: error downloading provider: couldn't find remote ref "refs/tags/v0.0.21-beta.1"
Will retry in 80ms [1/5]
warning: error downloading provider: couldn't find remote ref "refs/tags/v0.0.21-beta.1"
Will retry in 160ms [2/5]
warning: error downloading provider: couldn't find remote ref "refs/tags/v0.0.21-beta.1"
Will retry in 320ms [3/5]
warning: error downloading provider: couldn't find remote ref "refs/tags/v0.0.21-beta.1"
Will retry in 640ms [4/5]
warning: error downloading provider: couldn't find remote ref "refs/tags/v0.0.21-beta.1"
Will retry in 1.28s [5/5]
```

We inherited that behavior from `pulumi package get-schema`.

I'd actually say this is a bug 🤔

flostadler: https://github.com/pulumi/pulumi/issues/19724

content/docs/idp/get-started/private-registry.md
```diff
@@ -22,38 +22,46 @@ Developers leverage templates and components in their preferred workflows, wheth
...
 ```bash
-pulumi package publish /path/to/your/component --readme README_LOCATION
+pulumi package publish github.com/acme/k8s-cluster
 ```
 
-#### Specifying an Organization
+#### Component Versioning
 
-If you're part of multiple organizations and do not have a [default organization](/docs/iac/cli/commands/pulumi_org_set-default/) set, you must specify the org by using the `--publisher` flag.
+Pulumi uses git tags for versioning components. By default, the latest version tag will be used. The tag must adhere to [to the semantic versioning standard](https://semver.org/) plus a "v" prefix (e.g. `v1.2.3`).
```
flostadler:
```suggestion
Pulumi uses git tags for versioning components. By default, the latest version tag will be used. The tag must adhere to [the semantic versioning standard](https://semver.org/) plus a "v" prefix (e.g. `v1.2.3`).
```

content/docs/idp/get-started/private-registry.md
flostadler: I'd personally vote for fixing https://github.com/pulumi/pulumi/issues/19724 and not mentioning the v prefix in the docs right now. Potentially just give an example. Wdyt?

danielrbradley: I think we should make the docs match what's required right now, then circle back on it if core take and fix that issue

flostadler: Sounds good to me!

flostadler (approved):

PR #15276, opened 2025-06-10
Fix Typescript optionality of opts for consistency

OddEssay:
### Proposed changes

Fixes the initial TypeScript code example on line 148 to match the complete code snippet on line 381.
This comes up if following the TypeScript tutorial along under [Define a new component](https://www.pulumi.com/docs/iac/get-started/aws/create-component/#define-a-new-component), where if you start with the initial code snippet, and refactor the code manually rather than just copy-pasting the finished file example, you'll end up with a TypeScript error.


thoward (approved):

PR #15367, opened 2025-06-21
Clarify stack policy timing in CrossGuard documentation

### LINKED ISSUE #11508: Pulumi policy as code concepts - Stack policies needs clarification
[themes/default/content/docs/using-pulumi/crossguard/core-concepts.md](https://www.pulumi.com/docs/using-pulumi/crossguard/core-concepts/#types-of-policies)

Stack policies run after a `preview` do not run after everything has been provisioned. We need to clarify this since stack policies are actually _most_ useful during a preview.

jkodroff:
Add detailed explanation of when stack policies execute during preview vs update operations. Clarify that stack policies are most useful during preview operations since they can catch violations before actual resource provisioning occurs.

Fixes #11508

### DIFF
--- a/content/docs/iac/crossguard/core-concepts.md
+++ b/content/docs/iac/crossguard/core-concepts.md
@@ -91,19 +91,28 @@ There are no restrictions on which policies you combine within a pack, and you s
 
 There are two broad types of policies:
 
-1. *Resource Policies*: These validate a particular resource in a stack or account before the resource is created or updated, looking at the resource's _input_ properties.
-2. *Stack Policies*: These validate all resources in the stack after they've been created/updated, but before the Pulumi preview/update has completed, looking at each resource's _output_ properties.
+1. **Resource Policies:** These validate a particular resource in a stack or account before the resource is created or updated, looking at the resource's *input* properties.
+1. **Stack Policies:** These validate all resources in the stack after the Pulumi operation has processed all resources, looking at each resource's *output* properties. Stack policies are most useful during preview operations.
 
 This table summarizes the primary differences between the two types:
 
 |                                | Resource Policies                     | Stack Policies                                                                        |
 |--------------------------------|---------------------------------------|---------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


content/docs/iac/using-pulumi/crossguard/core-concepts.md
```diff
@@ -90,20 +90,29 @@ There are no restrictions on which policies you combine within a pack, and you s
...
 | What is the type name?         | `ResourceValidationPolicy`            | `StackValidationPolicy`                                                               |
 | Supported for Insights?        | Yes                                   | No - IaC-specific feature                                                             |
 
+{{% notes type="info" %}}
+**Stack Policy Timing**: The timing of stack policy execution differs between `pulumi preview` and `pulumi update`:
+
+- During `pulumi preview`: Stack policies run after the preview completes, using whatever output values are already known from previously provisioned resources. Since no new resources are actually created during preview, stack policies are most useful in this context for catching violations before any provisioning occurs.
+- During `pulumi update`: Stack policies run after all resources have been provisioned, but before the update operation completes. This means non-compliant resources may already be created by the time the policy violation is detected.
+
+For this reason, stack policies are most effective when used with `pulumi preview` to identify issues before actual resource provisioning.
```
lblackstone: I think there are some subtle tradeoffs here. Preview won't necessarily resolve all outputs, so the policy may be evaluating a different set of properties depending on the operation. However, as you note, preview gates all changes to the stack.

jkodroff: Can you give me an example of where an output might not be resolved in a preview if a resource was already provisioned in a `pulumi up`? (That's what I was trying to communicate there, so if that was not clear I will reword it.)

lblackstone: First one that comes to mind is the docker buildx provider, which won't actually build the image during preview. There are probably a few other resource types where the state is expected to change at runtime often. It's a minor detail, and probably better to be prescriptive with advice that is right for the vast majority of cases.

PR #15411, opened 2025-06-26
Improve functions docs

### LINKED ISSUE #15368: Improve Resource Function Docs
https://www.pulumi.com/docs/iac/concepts/resources/functions/

- The existing direct form examples should not be wrapping the result in `pulumi.output()`.
- Give a simple example of using the output form of a function that takes an input. (Need some suggestions on what a practical example might be.)
- Provide guidance on when to use the direct form versus the output form. The output form should be preferred by default because:
	- By sticking strictly with Pulumi inputs and outputs, as opposed to mixing them with promises, you reduce mental overhead - there are less types/semantics to deal with.
	- Using async in TS in a Pulumi program can be difficult or impossible (I'm not sure which - need to confirm).
	- The output version of a function allows the use of `dependsOn`, so that you can specify when your function should execute.
	- Note that if you need to use a function result to determine whether a resource should be created or not, you _must_ use the direct form of a function.
- Add a note that in bridged providers, which take a Terraform/OpenTofu provider as an underlying dependency, Data Sources in the upstream provider will be exposed as Pulumi Functions in the corresponding Pulumi provider.
- "Invoke Options", "Dependencies and Ordering", and "Resource Methods" should each be an h2

Once this is done, we should add an issue to change our translation of examples to use the Output form of functions in the Registry.

jkodroff:
Fixes #15368

- Indicate which version should be preferred (output form)
- Explain how bridged providers expose data sources as functions
- Explain when each form executes
- Add scenarios where one form or the other _must_ be used
- Explain why resource methods return outputs

### DIFF
--- a/content/docs/iac/concepts/resources/functions.md
+++ b/content/docs/iac/concepts/resources/functions.md
@@ -235,7 +235,7 @@ Bridged providers, which take a Terraform provider as an underlying dependency, 
 
 Provider functions are exposed in each language as regular functions, in two variations:
 
- 1. The **direct form** accepts plain arguments (e.g. `string`, as opposed to `pulumi.Input<string>`) and returns an asynchronous value or blocks until the result is available.These functions are typically named, e.g., `getX()`.
+ 1. The **direct form** accepts plain arguments (e.g., `string`, as opposed to `pulumi.Input<string>`) and returns an asynchronous value (e.g., a `Promise` in Node.js, a `Task` in Python, etc.) or blocks until the result is available. These functions are typically named, e.g., `getX()`.
  1. The **output form** accepts Pulumi Inputs (or plain values) as arguments and returns a Pulumi Output as a result. For more information on these types, see [Inputs and Outputs](/docs/concepts/inputs-outputs/). These functions are typically named, e.g., `getXOutput()`.
 
 The [Pulumi Registry](/registry) contains authoritative documentation for all provider functions.


content/docs/iac/concepts/resources/functions.md
```diff
@@ -76,40 +78,40 @@ instance = aws.ec2.Instance(
...
 
 import (
-	"github.com/pulumi/pulumi-aws/sdk/v6/go/aws/ec2"
-	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+ "github.com/pulumi/pulumi-aws/sdk/v6/go/aws/ec2"
+ "github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
 func main() {
-	pulumi.Run(func(ctx *pulumi.Context) error {
```
iwahbe, line 84: Why have we lost the tabs here? Tabs are correct for Go.

content/docs/iac/concepts/resources/functions.md
```diff
@@ -225,40 +227,54 @@ resources:
...
+{{% / notes %}}
+
+## Direct form and output form
+
 Provider functions are exposed in each language as regular functions, in two variations:
 
- 1. The **direct form** accepts plain arguments (e.g. `string`, as opposed to `pulumi.Input<string>`) and returns an asynchronous value (e.g. a `Promise` in NodeJS, or a `Task` in Python), or blocks until the result is available. These functions are typically named, e.g., `getFoo()`.
- 2. The **output form** accepts `Input` values (or plain values) and returns an [Output](/docs/concepts/inputs-outputs/). These functions are typically named, e.g., `getFooOutput()`.
+ 1. The **direct form** accepts plain arguments (e.g. `string`, as opposed to `pulumi.Input<string>`) and returns an asynchronous value or blocks until the result is available.
+ 1. The **output form** accepts Pulumi Inputs as arguments and returns a Pulumi as a result. For more information on these types, see [Inputs and Outputs](/docs/concepts/inputs-outputs/).
```
iwahbe:
```suggestion
 1. The **output form** accepts Pulumi Inputs as arguments and returns a Pulumi Output as a result. For more information on these types, see [Inputs and Outputs](/docs/concepts/inputs-outputs/).
```

Otherwise I'm imaging a purple platypus coming out of the screen.

content/docs/iac/concepts/resources/functions.md
```diff
@@ -225,40 +227,54 @@ resources:
...
+In addition to function arguments, provider functions also accept "invoke options", similar to the way Pulumi resources accept [resource options](/docs/concepts/options/). Invoke options may be specified either as an object or as a list of arguments depending on the language you're writing your Pulumi program in. The available options are:
 
-- `parent`: Supply a parent resource for this function call. Much like the [parent resource option](/docs/concepts/options/parent/), the parent will be consulted when determining the provider to use.
+* `dependsOn`: An array of resources that this function depends on. This option is only available in the Output form of a provider function. See [Dependencies and ordering](#dependencies-and-ordering) for a full explanation.
+* `parent`: Supply a parent resource for this function call. Much like the [parent resource option](/docs/concepts/options/parent/), the parent will be consulted when determining the provider to use.
+* `pluginDownloadURL`: Pass a URL from which the provider plugin should be fetched. This may be necessary for third-party packages such as those not hosted at [https://get.pulumi.com](https://get.pulumi.com).
```
iwahbe: `pluginDownloadURL` and `version` should not be explicitly passed by users. They are automatically pre-set by the calling SDK. 

You can see this looking at 3rd party SDKs: https://github.com/wttech/pulumi-aem/blob/f1123dcace8f7812a0239f22d2a12d3b299d2b68/sdk/go/aem/internal/pulumiUtilities.go#L178-L182.

content/docs/iac/concepts/resources/functions.md
```diff
@@ -225,40 +227,54 @@ resources:
...
+* **If you need a provider function's result to determine whether a resource should be created at all**, you must use the provider function's the direct form. The direct form of a function executes _while_ the Pulumi engine is formulating the dependency graph (that is, determining what resources need to be created, updated, or deleted), so in order to figure out whether a resource belongs in the graph at all, that decision has to always be calculated up front.
+* **If you need resources to be created or updated before the function is invoked**, you must use the provider function's output form. Dependencies in the output form of a function are tracked identically to resources: all inputs to the function must be resolved before the function executes. If you need to specify a dependency that isn't already implied by an input to the function's arguments, you can use the `dependsOn` function option to specify additional dependencies (just like you can with resources).
```
iwahbe: I'll defer to you how to best write the doc, but you can use the direct form to deal with outputs, you just need to wrap it in an apply. It's generally nicer to use the outputty form, but never mandatory.

iwahbe (approved):

content/docs/iac/concepts/resources/functions.md
thoward: I think the note about naming conventions (e.g. `getFoo` vs `getFooOutput`) was helpful and we shouldn't remove that.

content/docs/iac/concepts/resources/functions.md
```diff
@@ -225,40 +227,54 @@ resources:
...
+Bridged providers, which take a Terraform provider as an underlying dependency, expose Terraform data sources in the upstream Terraform provider as provider functions in the corresponding Pulumi provider.
+{{% / notes %}}
+
+## Direct form and output form
+
 Provider functions are exposed in each language as regular functions, in two variations:
 
- 1. The **direct form** accepts plain arguments (e.g. `string`, as opposed to `pulumi.Input<string>`) and returns an asynchronous value (e.g. a `Promise` in NodeJS, or a `Task` in Python), or blocks until the result is available. These functions are typically named, e.g., `getFoo()`.
- 2. The **output form** accepts `Input` values (or plain values) and returns an [Output](/docs/concepts/inputs-outputs/). These functions are typically named, e.g., `getFooOutput()`.
+ 1. The **direct form** accepts plain arguments (e.g. `string`, as opposed to `pulumi.Input<string>`) and returns an asynchronous value or blocks until the result is available.
```
thoward: Users often report confusion about the async nature of Outputs and that's why there was clarifying information to relate them to Promises/Tasks etc. That helps folks quickly understand what kind of object they are getting back (and prevents a lot of "why can't i access this output value" type support questions). Maybe there's a better way to rephrase this, but removing it feels like a step backward in terms of helping our users understand how to use Pulumi effectively.

content/docs/iac/concepts/resources/functions.md
jkodroff: They are, but VSCode autosave turns them into spaces, and I cannot figure out what do to about it. I might vibecode myself into a solution. I don't think it's worth blocking on a solution since the code is good, but it's something longterm we do need a solution for.

content/docs/iac/concepts/resources/functions.md
jkodroff: Updated. THANK YOU.

content/docs/iac/concepts/resources/functions.md
jkodroff: Brought it back (with the less jargon-y `X` instead of `Foo`). I was originally going to lay out the signature in every language as a chooser (since the naming convention actually varies by language), but it was difficult to make readable (too wide - looped around to a new line) didn't want to block on it.

content/docs/iac/concepts/resources/functions.md
jkodroff: Noted.

content/docs/iac/concepts/resources/functions.md
jkodroff: Added back the stuff about `Promise`, etc. Again, was gonna lay out each function signature to be more inclusive, but it was getting to be a pain to get to display nicely.

Note that the direct form literally returns a `Promise`. The Inputs and Outputs page should be where we make the analogy between a promise (which I believe we do).

PR #15439, opened 2025-06-27
Re-seed weights in IaC concepts

jkodroff:
This should be a no-op from the perspective of the user - the menu items order should remain the same, but it will make future reordering easier.

### DIFF
--- a/content/docs/iac/concepts/secrets.md
+++ b/content/docs/iac/concepts/secrets.md
@@ -17,7 +17,7 @@ aliases:
 - /secrets/
 ---
 
-All resource input and output values are recorded as stack [_state_](/docs/concepts/state) and stored in Pulumi Cloud, in a state file, or in your DIY backend of choice. These values are usually just plain-text strings, such as configuration settings, computed URLs, or resource identifiers. Sometimes, hsowever, these values contain sensitive data, such as database passwords or service tokens, that must be handled carefully and protected from exposure.
+All resource input and output values are recorded as stack [_state_](/docs/concepts/state) and stored in Pulumi Cloud, in a state file, or in your DIY backend of choice. These values are usually just plain-text strings, such as configuration settings, computed URLs, or resource identifiers. Sometimes, however, these values contain sensitive data, such as database passwords or service tokens, that must be handled carefully and protected from exposure.
 
 Pulumi Cloud transmits and stores entire state files securely, but Pulumi also supports encrypting individual values as _secrets_ for additional protection. Encryption ensures that these values never appear as plain text in your state file. By default, the encryption method uses automatic, per-stack encryption keys provided by Pulumi Cloud, but you can also use a [provider of your own choosing](#configuring-secrets-encryption) instead.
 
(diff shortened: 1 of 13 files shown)


content/docs/iac/concepts/secrets.md
```diff
@@ -220,7 +220,7 @@ By default, the Pulumi CLI uses a per-stack encryption key managed by Pulumi Clo
 For example, this command sets a configuration variable named `dbPassword` to the plain-text value `S3cr37`:
 
 ```bash
-$ pulumi config set --secret dbPassword S3cr37
```
joeduffy: Note that our copy JavaScript knows to ignore the leading `$`. Is the proposal to move away from this everywhere? I suspect we do this in _a lot_ of places (it's been our style since day one, probably my fault).

thoward: I suggest that we adopt the Google style guide recommendations here: https://developers.google.com/style/code-syntax#prompt ... basically, showing the `$` prompt is always preferred (it's technically optional for single commands, but required for multiple commands in the same listing). 

In general, I consider the Google style guide our base style, with some Pulumi-specific variations that override it.

jkodroff: Markdownlint says you only include the `$` if you show the output. I'm totally cool if we change this with a global config, though.

content/docs/iac/concepts/secrets.md
```diff
@@ -17,7 +17,7 @@ aliases:
 - /secrets/
 ---
 
-All resource input and output values are recorded as stack [_state_](/docs/concepts/state) and stored in Pulumi Cloud, in a state file, or in your DIY backend of choice. These values are usually just plain-text strings, such as configuration settings, computed URLs, or resource identifiers. Sometimes, however, these values contain sensitive data, such as database passwords or service tokens, that must be handled carefully and protected from exposure.
+All resource input and output values are recorded as stack [_state_](/docs/concepts/state) and stored in Pulumi Cloud, in a state file, or in your DIY backend of choice. These values are usually just plain-text strings, such as configuration settings, computed URLs, or resource identifiers. Sometimes, hsowever, these values contain sensitive data, such as database passwords or service tokens, that must be handled carefully and protected from exposure.
```
jkodroff:
```suggestion
All resource input and output values are recorded as stack [_state_](/docs/concepts/state) and stored in Pulumi Cloud, in a state file, or in your DIY backend of choice. These values are usually just plain-text strings, such as configuration settings, computed URLs, or resource identifiers. Sometimes, however, these values contain sensitive data, such as database passwords or service tokens, that must be handled carefully and protected from exposure.
```

jkodroff:
This is now fixed. We have to disable format on save in VSCode to prevent this from happening in the future, so I've added a project settings file to fix.

The other files were also restored, save for the intended changes.

content/docs/iac/concepts/secrets.md
jkodroff: This is now fixed. Format on save in VSCode "corrected" this.

PR #15418, opened 2025-06-26
Add a blog post for resource hooks

lunaris:
This change adds the launch blog post for resource hooks, as implemented in Pulumi 3.182.0. Resource hooks allow users to execute arbitrary code at various points in the resource lifecycle.

### Tasks
- [x] Examples for all languages
    - [x] JavaScript
    - [x] TypeScript
    - [x] Python
    - [x] Go
    - [x] .Net (C#)
    - [x] Java -- not supported
    - [x] YAML -- not supported
- [ ] Merge after #15463 since this post links to docs

PR #15463, opened 2025-07-02
Document resource hooks

lunaris:
This change documents resource hooks, as implemented in Pulumi 3.182.0. Resource hooks allow users to execute arbitrary code at various points in the resource lifecycle.

### Tasks
- [x] Use health-check example from blog post -- copy in when all languages have been done
- [ ] Merge before #15418 since blog post links to docs

### DIFF
--- a/content/docs/iac/concepts/options/hooks.md
+++ b/content/docs/iac/concepts/options/hooks.md
@@ -42,30 +42,521 @@ When a hook is executed as part of a resource operation, it receives the resourc
 {{% choosable language javascript %}}
 
 ```javascript
+import * as aws from "@pulumi/aws"
+import * as pulumi from "@pulumi/pulumi"
+
+export = async () => {
+  // Define the instance's user data script to set up a simple HTTP server that
+  // serves some static content. We'll try to fetch the health.json in a
+  // lifecycle hook later on to check whether or not the server is up and running.
+  const userData = `#!/bin/bash
+echo "Hello, World!" > index.html
+echo '{"ok": true}' > health.json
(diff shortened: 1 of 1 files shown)


content/docs/iac/concepts/options/hooks.md
```diff
@@ -0,0 +1,95 @@
...
+
+| Hook type     | Old inputs | New inputs | Old outputs | New outputs |
+|---------------|------------|------------|-------------|-------------|
+| Before create |            | ✓          |             |             |
+| After create  |            | ✓          |             | ✓           |
+| Before update | ✓          | ✓          | ✓           |             |
+| After update  | ✓          | ✓          | ✓           | ✓           |
+| Before delete | ✓          |            | ✓           |             |
+| After delete  | ✓          |            | ✓           |             |
+
```
julienp, line 37: Quick note before I forget: we should mention that hooks can return an error. For before hooks this causes the action to fail. After hooks don't cause a failure (the action already ran!), but do log a warning.

lunaris: Good shout. I've taken a stab at it, though the table extension isn't the prettiest. Thoughts?
image

julienp: It's a little squished on my screen, but still looks ok. We could remove the error behaviour column from the table, since the text explains the behaviour, but it's also fine to keep in my opinion.

Screenshot 2025-07-08 at 10 13 03

julienp (approved):
Yay!

PR #15546, opened 2025-07-21
Remove Answers sitemap from sitemap-index.xml

sicarul:
The sitemap now 404s so the index points to a dead sitemap.


lucascrespo88 (approved):
LGTM

PR #15577, opened 2025-07-23
Couple of Deployments pre-run command clarifications

smithrobs:
### Proposed changes

- Clarify pre-run commands execute separately in own shell
- Add note about persisting env vars via PULUMI_ENV file

### DIFF
--- a/content/docs/pulumi-cloud/deployments/using/settings.md
+++ b/content/docs/pulumi-cloud/deployments/using/settings.md
@@ -55,14 +55,14 @@ For more information on customer-managed agents, see the [Customer Managed Agent
 
 ## Pre-Run Commands
 
-Pre-run commands allow you to execute arbitrary shell commands before the deployment process starts. This is useful for environment setup, authentication with private package repositories, or other preparatory work.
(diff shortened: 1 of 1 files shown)


content/docs/pulumi-cloud/deployments/using/settings.md
```diff
@@ -55,14 +55,14 @@ For more information on customer-managed agents, see the [Customer Managed Agent
...
-Pre-run commands allow you to execute arbitrary shell commands before the deployment process starts. This is useful for environment setup, authentication with private package repositories, or other preparatory work.
+Pre-run commands allow you to execute arbitrary shell commands before the deployment process starts. This is useful for environment setup, authentication with private package repositories, or other preparatory work. Note that each line of your pre-run command runs in a separate shell.
 
 For example, you might use pre-run commands to:
 
 - Install additional dependencies
 - Configure authentication for private repositories
 - Generate configuration files
-- Set up environment variables
+- Set up environment variables - see [here](#pulumi_env) if you need to persist these to your Pulumi program.
```
jkodroff: Use the title of the heading instead of “here” for accessibility.

jkodroff (approved):
Pre approving but note that this will likely fail linting in the future  if we keep the link text “here”

PR #15606, opened 2025-07-29
Fix more broken links

thoward:
These run a range of different problems, addressing most recent link checker report.


smithrobs:
I'm slightly conflicted about removing links to years-old links in years-old blog posts, but they're years-old blog posts. 😸 
LGTM!

thoward:
> I'm slightly conflicted about removing links to years-old links in years-old blog posts, but they're years-old blog posts. 😸 LGTM!

I agree, but the links being removed are dead already, so linking to a thing that is 404ing is ... worse? Also, noisy in the link checker logs, hiding new issues that come up in all the noise. Wherever possible I replaced links with still-valid ones, so the hard removal of a link is a last resort.

PR #15627, opened 2025-07-31
Fix typo on Typescript installation command

hakan-gencoglu:
### Proposed changes

Fix the typo of `isntall` to `install` for custom typescript version installation.


smithrobs:
Good catch, thank you!

PR #15698, opened 2025-08-11
Update Plans - remove expiramental/preview warning

automagic:
### Proposed changes

Update update-plans.md to remove warning about preview, as the feature is GA

### DIFF
--- a/content/docs/iac/concepts/update-plans.md
+++ b/content/docs/iac/concepts/update-plans.md
@@ -20,10 +20,6 @@ aliases:
 - /docs/concepts/update-plans/
 ---
 
-{{% notes type="warning" %}}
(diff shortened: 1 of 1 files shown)


CamSoper (approved):
This LGTM @automagic, let's :shipit:

smithrobs:
now I'm confused; if GA, why is `save-plan` marked [experimental and hidden](https://github.com/pulumi/pulumi/blob/8dbb586beea262f2241a4f0f1b2a68d1dd7d050d/pkg/cmd/pulumi/operations/preview.go#L588)?

Frassle:
Note that "as the feature is GA" isn't quite true. We never GA'd this, it's just "public preview", update plans still have a lot of corner cases preventing their use in general. But the they shouldn't be hidden behind the EXPERIMENTAL env var anymore.

PR #15763, opened 2025-08-15
blog: fix a broken link in the latest IDP blog

pgavlin:
This link should point to the [private registry page](https://www.pulumi.com/docs/idp/get-started/private-registry/)


dirien (approved):
LGTM!

PR #15904, opened 2025-09-03
Add event/workshop script

kimberleyamackenzie:
Still testing :)

### DIFF
--- a/content/events/testing-this-sworkshop-script/index.md
+++ b/content/events/testing-this-sworkshop-script/index.md
@@ -1,0 +1,83 @@
+---
+# Name of the event, <= 60 characters
+title: "Testing this sworkshop script!"
+meta_desc: This is a great meta description, it should be long enough.
+meta_image:
+
+# A featured webinar will display first in the list.
+featured: false
+
+# Webinars with unlisted as true will not be shown on the webinar list
+unlisted: false
+
+# Gated webinars will have a registration form and the user will need
+# to fill out the form before viewing.
+gated: true
+
+# The layout of the landing page.
+type: webinars
+
+# External webinars will link to an external page instead of a webinar
+# landing/registration page. If the webinar is external you will need
+# set the 'block_external_search_index' flag to true so Google does not index
+# the webinar page created.
+external: false
+block_external_search_index: false
+
+# The url slug for the webinar landing page. If this is an external
+# webinar, use the external URL as the value here.
+url_slug: testing-this-sworkshop-script
+
+# Content for the left hand side section of the page.
+main:
+    # Webinar title.
+    title: "Testing this sworkshop script!"
+    event_type: workshop # workshop | event
+
+    # URL for embedding a URL for ungated webinars.
+    youtube_url: 
+
+    # Sortable date. The datetime Hugo will use to sort the webinars in date order.
+    sortable_date: 2025-09-30T19:00:00.000-05:00
+
+    # Duration of the webinar.
+    duration: 90 minutes
+
+    # "virtual" will be shown under "show virtual events only", otherwise shown as City, State (seattle, wa)
+    location: virtual
+
+    # Description of the webinar.
+    description: |
+        Some more details here - you'll love this workshop!  Woohoo!
+    learn:
+        - Learn
+        - Love Pulumi
+        - Have fun!
+
+    # The webinar presenters
+    presenters:
+        - name: Pulumipus
+          role: Sr Engineer
+          photo: /images/team/pulumipus.jpg
+
+    # case-sensitive
+    tags:
+        level: begineer # Beginner, Intermediate, Advanced
+        topics:  ["AI", "Secrurity"]
+        languages: ["TypeScript", "Python"]
+        clouds: ["AWS"]
+
+# The right hand side form section.
+form:
+    # HubSpot form id.
+    hubspot_form_id: 123
+    salesforce_campaign_id: 456
+
+event_data:
+  name: "Testing this sworkshop script!"
+  start_date: 2025-09-30T19:00:00.000-05:00
+  end_date: 2025-09-30T20:30:00.000-05:00
+  url: "https://www.pulumi.com/resources/testing-this-sworkshop-script/"
+  description: |
+    Some more details here - you'll love this workshop!  Woohoo!
+---
--- a/scripts/create-event-file.js
+++ b/scripts/create-event-file.js
@@ -1,0 +1,495 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const path = require('path');
+const readline = require('readline');
+const { execSync } = require('child_process');
+
+// Create readline interface
+const rl = readline.createInterface({
+  input: process.stdin,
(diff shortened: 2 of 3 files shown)


scripts/create-event-file.js
```diff
@@ -0,0 +1,495 @@
...
+form:
+    # HubSpot form id.
+    hubspot_form_id: ${hubspotFormId}
+    salesforce_campaign_id: ${salesforceCampaignId}
+
+event_data:
+  name: "${title}"
+  start_date: ${sortableDate}
+  end_date: ${endDate}
+  url: "https://www.pulumi.com/resources/${slug}/"
```
shughes26, line 305: this will need to be extended to be able to be pulumi or external

scripts/create-event-file.js
```diff
@@ -0,0 +1,495 @@
...
+}
+
+// Helper function to format date for sortable_date
+function formatSortableDate(dateStr, timeStr, timezone) {
+  const date = new Date(`${dateStr}T${timeStr}`);
+  const offset = timezone === 'EDT' ? '-04:00' : 
+                 timezone === 'EST' ? '-05:00' :
+                 timezone === 'PDT' ? '-07:00' :
+                 timezone === 'PST' ? '-08:00' : '-00:00';
+  return date.toISOString().replace('Z', offset);
```
shughes26, line 55: I fear that sometimes we're in australia

shughes26: and europe for that matter

kimberleyamackenzie: blame cursor for being very america-forward

scripts/create-event-file.js
kimberleyamackenzie: good call, ty!!

scripts/create-event-file.js
```diff
@@ -0,0 +1,495 @@
...
+
+# The url slug for the webinar landing page. If this is an external
+# webinar, use the external URL as the value here.
+url_slug: ${slug}
+
+# Content for the left hand side section of the page.
+main:
+    # Webinar title.
+    title: "${title}"
+    event_type: workshop # workshop | event
```
shughes26, line 262: if we want this to work for in-person (which it easily could) this should be var'd

scripts/create-event-file.js
```diff
@@ -0,0 +1,495 @@
...
+    youtube_url: 
+
+    # Sortable date. The datetime Hugo will use to sort the webinars in date order.
+    sortable_date: ${sortableDate}
+
+    # Duration of the webinar.
+    duration: ${duration}
+
+    # "virtual" will be shown under "show virtual events only", otherwise shown as City, State (seattle, wa)
+    location: virtual
```
shughes26, line 274: same as above re:in-person

content/events/testing-this-sworkshop-script/index.md
```diff
@@ -0,0 +1,83 @@
+---
+# Name of the event, <= 60 characters
+title: "Testing this sworkshop script!"
+meta_desc: This is a great meta description, it should be long enough.
+meta_image:
```
tomkingkong, line 5: Okay to be empty?

kimberleyamackenzie: yeah, i think basically any field can be empty and it will just be omitted from the rendered pages - cc @shughes26 to double check me on that

scripts/create-event-file.js
```diff
@@ -0,0 +1,495 @@
...
+    });
+  });
+}
+
+// Validation functions
+function validateTitle(title) {
+  if (title.length > 60) {
+    return `❌ Title is ${title.length} characters (max 60). Please shorten it.`;
+  }
+  return `✅ Title is ${title.length} characters (within 60 limit)`;
```
tomkingkong, line 28: Does it matter to users if they know their input was within char bounds? Perhaps a brief "✅ Title" would suffice?

scripts/create-event-file.js
```diff
@@ -0,0 +1,495 @@
...
+  const topics = await askQuestion('🏷️  Topics (comma-separated): ').split(',').map(t => t.trim()).filter(t => t);
+  const languages = await askQuestion('💻 Languages (comma-separated): ').split(',').map(t => t.trim()).filter(t => t);
+  const clouds = await askQuestion('☁️  Clouds (comma-separated): ').split(',').map(t => t.trim()).filter(t => t);
+
+  // Get form IDs
+  const hubspotFormId = await askQuestion('📋 HubSpot Form ID: ');
+  const salesforceCampaignId = await askQuestion('📊 Salesforce Campaign ID: ');
+
+  // Create folder and file
+  const slug = createSlug(title);
```
tomkingkong, line 211: Should we check if slug exists and randomize if so? Low chance, but still possible no?

scripts/create-event-file.js
```diff
@@ -0,0 +1,495 @@
...
+// Main function to create event file
+async function createEventFile() {
+  console.log('\n🎯 Creating an Event File\n');
+  console.log('📋 Linting Requirements:');
+  console.log('   • Title must be ≤ 60 characters');
+  console.log('   • Meta description must be ≤ 160 characters\n');
+
+  // Get basic info
+  let title = await askQuestion('📝 Event Title: ');
+  while (title.length > 60) {
```
tomkingkong, line 341: nitty nit, but we can make the validations into consts, like `TITLE_MAX_LENGTH`

scripts/create-event-file.js
```diff
@@ -0,0 +1,495 @@
...
+    console.log(`   git commit -m "Add event: ${title}"`);
+  }
+}
+
+// Main function
+async function main() {
+  console.log('🚀 Pulumi Event/Workshop File Creator\n');
+  
+  const fileType = await askQuestion('What would you like to create? (workshop/event): ');
+
```
tomkingkong, line 475: nitty nit nit, perhaps the name of this file should be `create-file.js` under `/workshopAndEvents`, as it handles both

content/events/testing-this-sworkshop-script/index.md
shughes26: Yep, there's a default it falls back to if it's not set (on these pages it usually isn't), and adding it in after the fact is something that doesn't usually require eng support

scripts/create-event-file.js
shughes26: The character limit is good to reiterate

scripts/create-event-file.js
shughes26: Oh yes, but not randomize as the slug matters for SEO. Check if it exists, and then maybe a message to go and edit the existing slug file directly (a common pattern to reuse old ones) or pick something else if the existing one is still upcoming

meagancojocar:
@kimberleyamackenzie can we merge this?

kimberleyamackenzie:
@meagancojocar I still need to incorporate the feedback Shoes gave - this is a nice to have to try and help folks add their own events, so it fell a little lower in my pile.  Will get it done this iteration!

PR #15919, opened 2025-09-05
add note about changes to ignoreChanges resource option

tgummerer:
### Proposed changes

Add a note about how ignoreChanges treats arrays.  See https://github.com/pulumi/pulumi/pull/20465.  This should only be merged after Pulumi 3.194.0 is released.

### DIFF
--- a/content/docs/iac/concepts/options/ignorechanges.md
+++ b/content/docs/iac/concepts/options/ignorechanges.md
@@ -109,3 +109,7 @@ In addition to passing simple property names, nested properties can also be supp
 The property names passed to `ignoreChanges` should always be the "camelCase" version of the property name, as used in the core Pulumi resource model.
 For example, a property named `nested_resource` would turn into `nestedResource`.
 {{% /notes %}}
+
+{{% notes "info" %}}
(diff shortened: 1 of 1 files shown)


content/docs/iac/concepts/options/ignorechanges.md
```diff
@@ -109,3 +109,7 @@ In addition to passing simple property names, nested properties can also be supp
 The property names passed to `ignoreChanges` should always be the "camelCase" version of the property name, as used in the core Pulumi resource model.
 For example, a property named `nested_resource` would turn into `nestedResource`.
 {{% /notes %}}
+
+{{% notes "info" %}}
+For arrays with different lengths, only changes for elements that are in both arrays are ignored. If new input array is longer, additional elements will be taken from the new array. If the array is shorter, we only take that number of elements from the original array.
```
julienp:
```suggestion
For arrays with different lengths, only changes for elements that are in both arrays are ignored. If the new input array is longer, additional elements will be taken from the new array. If the new array is shorter, we only take that number of elements from the original array.

For example `ignoreChanges` on an old array `[1, 2]` and a new array `[a, b, c]` results in `[1, 2, c]`, and an old array `[1, 2, 3]` and a new array `[a, b]` results in `[1, 2]`.
```

tgummerer:
https://github.com/pulumi/pulumi/pull/20465 has been merged and released, so I'm going ahead and merge this as well.

PR #15934, opened 2025-09-09
Update parent resource options for S3 bucket components

### LINKED ISSUE #15933: Wrong usage of parent
[content/docs/iac/using-pulumi/build-a-component.md](https://www.pulumi.com/docs/iac/using-pulumi/build-a-component/)

This doc page suggests using a pulumi resource as a parent inside a Component Resource object.
This should be considered wrong as specified by the pulumi doc about parent usage [HERE](https://www.pulumi.com/docs/iac/concepts/options/parent/)

Example:
```
class StaticPage(pulumi.ComponentResource):
    endpoint: pulumi.Output[str]
    """The URL of the static website."""

    def __init__(self,
                 name: str,
                 args: StaticPageArgs,
                 opts: Optional[ResourceOptions] = None) -> None:

        super().__init__('static-page-component:index:StaticPage', name, {}, opts)

        # Create a bucket
        bucket = s3.Bucket(
            f'{name}-bucket',
            opts=ResourceOptions(parent=self))

        # Configure the bucket website
        bucket_website = s3.BucketWebsiteConfiguration(
            f'{name}-website',
            bucket=bucket.bucket,
            index_document={"suffix": "index.html"},
            opts=ResourceOptions(parent=bucket))
```

SharpEdgeMarshall:
### Proposed changes

Fix wrong `parent` usage in examples

### Related issues (optional)

Fixes #15933

### DIFF
--- a/content/docs/iac/using-pulumi/build-a-component.md
+++ b/content/docs/iac/using-pulumi/build-a-component.md
@@ -872,7 +872,7 @@ class StaticPage(pulumi.ComponentResource):
             f'{name}-website',
             bucket=bucket.bucket,
             index_document={"suffix": "index.html"},
-            opts=ResourceOptions(parent=bucket))
+            opts=ResourceOptions(parent=self))
 
         # Create a bucket object for the index document
         s3.BucketObject(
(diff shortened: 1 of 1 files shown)


jkodroff:
@SharpEdgeMarshall Thanks for this contribution, but there's a couple of reasons why we can't accept this contribution:

1. Because this change is only in Python, it would make the Python example inconsistent with the other languages
1. I believe the parenting to `bucket` instead of `self` is intentional. I believe that the note that you based this PR on is either incorrect or at the very least lacks the correct context, since our most used ComponentResource (`awsx.ec2.Vpc`) has resources that whose parent is another CustomResource within the ComponentResource (e.g. `awsx.ec2.Vpc` has `aws.ec2.Vpc` as a child which in turn has `aws.ec2.Subnet`s as children)

I'm going to close this PR and figure out what changes need to be made to the note. Totally reasonable why you would think that the existing code sample is wrong and apologies for any time wasted.

PR #16051, opened 2025-09-22
Update agent guidelines for internal links and file management

CamSoper:
Revise agent guidelines for moving, renaming, or deleting files, including the creation of redirects.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -69,12 +69,17 @@ Do not substitute other tools or commands.
 
 ## Moving and Deleting Files
 
-- When moving, renaming, or deleting files, ensure all internal references are updated accordingly.
-- When moving files within this repository, use `git mv` to preserve file history.
-- Create redirects for any moved, renamed, or deleted files to maintain link integrity.
-- Redirects should be added to the S3 redirects files located in `/scripts/redirects/`.
-- When adding redirects, place the redirect entry in the appropriate file. If an appropriate file does not exist, create a new one. Name the new file in a way that reflects the content it manages (e.g., `neo-redirects.txt` for redirects related to Neo).
-- Each redirect entry must follow the format `source-path|destination-url`.
+- **Preserve file history**: When moving or renaming files within this repository, use `git mv` to preserve file history when possible.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -68,6 +67,17 @@ Do not substitute other tools or commands.
 
 ---
 
+## Moving and Deleting Files
+
+- When moving, renaming, or deleting files, ensure all internal references are updated accordingly.
+- When moving files within this repository, use `git mv` to preserve file history.
```
jkodroff: Note that this is not always possible, for example when breaking out a single document into multiple documents.

AGENTS.md
```diff
@@ -68,6 +67,17 @@ Do not substitute other tools or commands.
 
 ---
 
+## Moving and Deleting Files
+
+- When moving, renaming, or deleting files, ensure all internal references are updated accordingly.
```
jkodroff: This should be more specific:

- We need to add an alias in the moved document.
- I'm _guessing_ we should ideally update links to use the new canonical path, but we do not enforce that today.
- Anchors links do not support aliases. (I think. Need to test.) This comes up when splitting up long docs. 
- When can aliases be removed? (I don't know for sure, but I would guess when no other internal links refer to the old URI.)

AGENTS.md
```diff
@@ -68,6 +67,17 @@ Do not substitute other tools or commands.
...
+- When moving files within this repository, use `git mv` to preserve file history.
+- Create redirects for any moved, renamed, or deleted files to maintain link integrity.
```
jkodroff: See above. "Alias" is the right term, because that's what it's called in the front matter.

AGENTS.md
```diff
@@ -68,6 +67,17 @@ Do not substitute other tools or commands.
...
+- Redirects should be added to the S3 redirects files located in `/scripts/redirects/`.
```
jkodroff: I've never touched this, so I might need some education here. Based on a quick look, I think this only applies when we're moving things that are not in Hugo?

PR #16065, opened 2025-09-23
Improve getting started experience

asafashirov:


meagancojocar:
Asaf mind adding a PR description explaining what change we are making? Is Joe the only reviewer we want?

PR #16097, opened 2025-09-26
Update ads, banner, and nav for Agentic AI course

kimberleyamackenzie:


kimberleyamackenzie:
@SaraDPH I'm looking into adding the ad to Registry too, but that happens in a different repo and it appears there is a bug there preventing the ads from being shown - I'm diving deeper, but I may not be able to get that done for Monday.

Issue #14130, opened 2025-02-20
Duplication of importing environments documentation


interurban:
Noted by @IaroslavTitov : 
"I noticed that we have exact same blob about importing environments in "Working with Environments" page. I think we should remove it out of there, since we have this separate page, or at least also update it there"

The opening paragraphs on these two docs pages are duplicative: 
https://www.pulumi.com/docs/esc/environments/working-with-environments/#importing-other-environments
and
https://www.pulumi.com/docs/esc/environments/imports/

The working with section should explain imports at a high level and reference the detailed sub-page

PR #16107, opened 2025-09-27
Removed duplicated import environment text in working-with-environmen…

tehsis:
### Proposed changes

This PR replaces the (duplicated) detailed description for environment imports in the "working with environments" page and references to the corresponding docs page.

Fixes: #14130

### DIFF
--- a/content/docs/esc/environments/working-with-environments.md
+++ b/content/docs/esc/environments/working-with-environments.md
@@ -450,11 +450,11 @@ For additional options and details, see `esc run --help`.
 
 ## Importing other environments
 
-Environments can also be composed from other environments.
+Pulumi ESC lets you define common configuration values once as reusable environments and then import them into other environments to avoid duplication, delegate ownership, and share settings (including secrets) across applications.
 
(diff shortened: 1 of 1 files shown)


content/docs/esc/environments/working-with-environments.md
```diff
@@ -450,11 +450,11 @@ For additional options and details, see `esc run --help`.
 
 ## Importing other environments
 
-Environments can also be composed from other environments.
+Pulumi ESC lets you define common configuration values once as reusable environments and then import them into other environments to avoid duplication, delegate ownership, and share settings (including secrets) across applications.
```
fnune, line 453: I'm sure there has been discussion about this before, but I would have expected "sharing configuration" rather than "settings". Configuration seems like the more generic/appropriate term to me.

Just a nit-pick! Happy to be corrected.

fnune (approved):

content/docs/esc/environments/working-with-environments.md
tehsis: In this case is mostly to avoid repeating configuration which is used as a general term at the beginning of the sentence

PR #16161, opened 2025-10-03
Update ignoreChanges documentation

corymhall:
### Proposed changes

Updating the docs on `ignoreChanges` to make it clearer to users some of the current limitations.

See the [internal doc](https://docs.google.com/document/d/1uItyB9sI9oKmo5HPjHf98NQtkcTLQRuYigZK5J57-s4/edit?tab=t.jpnfa3q186j2) for motivation.

Link to the specific page http://www-testing-pulumi-docs-origin-pr-16161-8cf51733.s3-website.us-west-2.amazonaws.com/docs/iac/concepts/options/ignorechanges/

### DIFF
--- a/content/docs/iac/concepts/options/ignorechanges.md
+++ b/content/docs/iac/concepts/options/ignorechanges.md
@@ -27,7 +27,7 @@ After the resource is created, Pulumi relies on the last recorded state for ever
 Because Pulumi reuses the value stored in the state, an external system can safely update the live resource as long as you synchronize the stack before the next update. Run `pulumi refresh` (or `pulumi up --refresh`) to pull the latest provider values into the state. Skipping the refresh step leaves stale values in the state, and Pulumi will continue to send those stale values to the provider during subsequent updates, which can overwrite the drift you meant to preserve.
 
 {{% notes type="warning" %}}
-When you skip `pulumi refresh` (or `pulumi up --refresh`) after `ignoreChanges` has been set, Pulumi keeps using the previous state value when it performs an update. Providers that require full object replacements—such as AWS load balancer listeners where the entire target group array is sent on every update—will receive the stale values from the state and may reset the live configuration.
+When you skip `pulumi refresh` (or `pulumi up --refresh`) after `ignoreChanges` has been set, Pulumi keeps using the previous state value when it performs an update. This can lead to unintentional changes if the cloud state has been changed (either through intentional external management, or unintentional drift). Providers that require full object replacements—such as AWS load balancer listeners where the entire target group array is sent on every update—will receive the potentially stale values from the state and may reset the live configuration.
 {{% /notes %}}
 
 For instance, in this example, the resource’s prop property "new-value" will be set when Pulumi initially creates the resource, but from then on, any updates will ignore it:
--- a/content/docs/iac/concepts/options/ignorechanges.md
+++ b/content/docs/iac/concepts/options/ignorechanges.md
@@ -16,6 +16,20 @@ aliases:
 
 The `ignoreChanges` resource option specifies a list of properties that Pulumi will ignore when it updates existing resources. Pulumi ignores a property by using the old value from the state instead of the value provided by the Pulumi program when determining whether an update or replace is needed. Ignored properties will still be used from the program when there is no previous value in the state, most importantly when creating the resource.
 
+## How ignoreChanges works
+
+After the resource is created, Pulumi relies on the last recorded state for every property named in `ignoreChanges`. During a preview or update, Pulumi:
+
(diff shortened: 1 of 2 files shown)


content/docs/iac/concepts/options/ignorechanges.md
```diff
@@ -16,6 +16,20 @@ aliases:
...
+
+1. Uses the program value when there is no prior state (typically the initial create).
+1. Uses the serialized value from the stack state for any later previews or updates.
+1. Does not automatically read the live value for the property unless you run a refresh.
+
+Because Pulumi reuses the value stored in the state, an external system can safely update the live resource as long as you synchronize the stack before the next update. Run `pulumi refresh` (or `pulumi up --refresh`) to pull the latest provider values into the state. Skipping the refresh step leaves stale values in the state, and Pulumi will continue to send those stale values to the provider during subsequent updates, which can overwrite the drift you meant to preserve.
+
+{{% notes type="warning" %}}
+When you skip `pulumi refresh` (or `pulumi up --refresh`) after `ignoreChanges` has been set, Pulumi keeps using the previous state value when it performs an update. Providers that require full object replacements—such as AWS load balancer listeners where the entire target group array is sent on every update—will receive the stale values from the state and may reset the live configuration.
+{{% /notes %}}
```
iwahbe, line 31: This is true if and only if you are using `ignoreChanges` to allow non-IaC management of a field. I don't think its fair to assume that all uses of `ignoreChanges` expect external management (and thus need a refresh before update).

content/docs/iac/concepts/options/ignorechanges.md
```diff
@@ -93,19 +107,319 @@ resources:
 
 {{< /chooser >}}
 
-One reason you would use the `ignoreChanges` option is to ignore changes in properties that lead to diffs. Another reason is to change the defaults for a property without forcing all existing deployed stacks to update or replace the affected resource. This is common after you’ve imported existing infrastructure provisioned by another method into Pulumi. In these cases, there may be historical drift that you’d prefer to retain, rather than replacing and reconstructing critical parts of your infrastructure.
+One reason you would use the `ignoreChanges` option is to ignore changes in properties that lead to diffs. Another reason is to change the defaults for a property without forcing all existing deployed stacks to update or replace the affected resource. This commonly occurs after importing existing infrastructure provisioned by another method into Pulumi. In these cases, there may be historical drift that you’d prefer to retain, rather than replacing and reconstructing critical parts of your infrastructure.
+
+## Example: Preserve externally managed weights
```
iwahbe, line 116: It's kinda funny to list 2 reasons why you might want to use `ignoreChanges`, then give an example for a _third another reason:

- ignore properties that lead to diffs
- change defaults without effecting existing resources

The example case should probably be listed in the set of resources you might use `ignoreChanges`.

iwahbe (commented):
I think there is a little bit of missing continuity and context (I left comments), but overall this looks good.

content/docs/iac/concepts/options/ignorechanges.md
corymhall: I can add something here to note that, but it's not that you necessarily _expect_ external management. Your cloud state could have drifted unintentionally and the result will be the same. The only way to guarantee that you are ignoring changes is if you refresh.

content/docs/iac/concepts/options/ignorechanges.md
corymhall: I updated this to be a list instead of a single paragraph. Let me know what you think.

iwahbe: Looks great.

content/docs/iac/concepts/options/ignorechanges.md
iwahbe: tldr. Good point. LGTM

---

Consider the following program + state + cloud:

State:
```
{
  "oldInputs": {
    "x": "a"
  }
}
```

Program:
```ts
new example.Resource_X("example", {
  x: "b"
}, { ignoreChanges: [ "x" ] });
```

Cloud (as returned by `Read`):
```
{
  "x": "c"
}
```

You have added `ignoreChanges: [ "x" ]`, so we definitely don't want the result of running anything to set state or the cloud to `"b"`. 

If you want `Resource_X.x` to be cloud managed, then you want to change the state to have `{ "x": "c" }` and you should run a refresh.

If you want `Resource_X.x` to ramain `"a"` (since that is the last program input)... you can't ever run refresh. This isn't really supported <-- _at this point I figured out I was wrong about how refresh interacted with `ignoreChanges`_

iwahbe (approved):

CamSoper (approved):

PR #16169, opened 2025-10-04
Add meta images to rss feed

seanyeh:
### Proposed changes

Add meta images to blog rss feed. I initially wanted to use <media:content> but somehow my rss feed reader didn't really like it so I think adding the image to the top of the description is simpler
Without adding the meta images, my feed reader was showing a random image from the article as the main image, which isn't as nice


CamSoper (approved):
This LGTM, but I haven't tested it with an RSS reader. @seanyeh I presume you have? Did you look at Claude's review?

PR #16230, opened 2025-10-09
Reinvent 2025 page

kimberleyamackenzie:
### Proposed changes

Add the new page for Pulumi's re:Invent 2025 presence.


kimberleyamackenzie:
@SaraDPH Ok, all done - meta image added, and redirect from event tile to reinvent page done!  The tile in the docs PR build will lead to the current reinvent page on prod, but will lead to the updated one once this PR is merged.

SaraDPH:
@kimberleyamackenzie looking good! Just a few items. The Happy Hour is missing the button that leads to the RSVP page.

Can you create the button but just add "RSVP coming soon" - that way we can replace the Text and add a link later. We may even self-serve if it is created, and we can just update as needed.
image

At the very bottom, the Request a Demo button gets pushed out of the purple banner (screenshot from an Android phone)

Screenshot_20251015-081442

kimberleyamackenzie:
@SaraDPH ok, both those items are addressed ✅ !

PR #16247, opened 2025-10-13
Implement "Neo Mode" for the marketing site homepage

kimberleyamackenzie:
Add a new toggle on the Marketing site that turns on "Neo mode," which opens a Neo prompt on the homepage that is deeplinked to a Neo query in the console.  Note that suggested queries have been updated since below screenshots.

Screenshot 2025-10-14 at 2 47 23 PM
Screenshot 2025-10-14 at 2 48 23 PM
Screenshot 2025-10-14 at 4 53 04 PM

### DIFF
--- a/theme/src/ts/neo-mode.ts
+++ b/theme/src/ts/neo-mode.ts
@@ -35,6 +35,17 @@ $(function () {
                 input.value = '';
             }
         }
+
+        // Track analytics if data-track is present
+        const analytics = (window as any).analytics;
+        const analyticsAvailable = analytics && analytics.track && typeof analytics.track === "function";
+
+        if (analyticsAvailable) {
+            const trackData = {
+                neoModeActive: isNeoMode ? "true" : "false",
+            };
+            analytics.track("neo-mode-toggle", trackData);
+        }
     }
 
     function submitNeoQuery() {


theme/src/ts/neo-mode.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    function submitNeoQuery() {
+        const input = document.getElementById('neo-input') as HTMLInputElement;
+        const query = input?.value.trim();
+
+        if (query) {
+            // Encode the query for URL safety
+            const encodedQuery = encodeURIComponent(query);
+            // Redirect to Pulumi Neo app with the query
+            window.location.href = `https://app.pulumi.com/neo?prompt=${encodedQuery}`;
```
arturl: What happens to user without a paid Pulumi subscription? Or users who are not logged in?

layouts/index.html
```diff
@@ -313,4 +340,69 @@ <h2 class="mb-12 text-center">Ready to try Pulumi?</h2>
...
+                            <path d="M5 12h14M12 5l7 7-7 7" stroke="white" stroke-width="2" stroke-linecap="round"
+                                stroke-linejoin="round" />
+                        </svg>
+                    </button>
+                </div>
+            </div>
+            <div class="neo-suggestions">
+                <button class="neo-suggestion-pill"
+                    onclick="populateNeoInput('What does the migration process from Terraform to Pulumi look like?')">
+                    Ditch Terraform
```
arturl: Ouch :-)
A bit on the nose, no?

theme/src/scss/_marketing.scss
```diff
@@ -189,3 +189,12 @@ body {
         background:linear-gradient(90deg, #be5188 0%, #805ac3 100%)
     }
 }
+
+// This hides Copilot on the home page, where we have a "Neo takeover" experience.
+// The section classes are generated based on the folder the page is located in, and our home page exists outside
+// a folder, hence "section-".
+.section- {
```
arturl, line 196: So, the take-over is happening on the main page (www.pulumi.com) but not docs or blogs, where Copilot is still present, right?

arturl (approved):
Looks good but need to figure out the story for anonymous users

layouts/index.html
vctrfrnndz: Yeah I think this is a bit agressive. 

`Migrate from Terraform` seems more like it!

theme/src/ts/neo-mode.ts
vctrfrnndz: They will go to the signup/login screen and be redirected to the deep link once logged in.

layouts/index.html
kimberleyamackenzie: This was updated to "Migrate to Pulumi" with feedback from Craig

theme/src/scss/_marketing.scss
kimberleyamackenzie: @arturl exactly - just the homepage, not even any other marketing pages - perhaps if its successful, we can expand it to other spots, nothing would be technically blocking to that

foot: Nice 👌

kimberleyamackenzie:
> Small nit: sample prompts here just prefill, so you have to "click sample", "click send", (inside console) "click send" again, which felt like a few clicks to get it going. However many people might end up at a signup/signin screen or something else first too.

yeah, agree that its heavy on the clicks/actions before the primary goal! re the signin/up, we decided this work was a quick way to try the concept without the longer workstream avoiding that signin/up would be.  then for the send - this seems like the expected way that the deep links work (probably so folks can add context if they want?) - may be worth adding an optional param to send right to the task.  same with the sample prompt - followed our existing pattern on the task page, but may be worth an auto send for this case.  

i expect we'll get some feedback quickly as this goes in, from internal folks at the least, so i'll connect with folks to fast follow some of the above as we start hearing impressions

theme/src/scss/neo-mode.scss
```diff
@@ -0,0 +1,331 @@
...
+body #main-content {
+    transition: opacity 0.3s ease, transform 0.3s ease;
+    max-width: 100vw;
+    overflow-x: hidden;
+}
+
+// Global body styles for Neo mode
+body {
+    overflow-x: hidden;
+    max-width: 100vw;
```
foot, line 265: Is this scoped to neo or more global? We seem to modify `body` in both main.scss and _marketing too. Which  might be a bit confusing sometimes but is a bit of nit.

foot (approved):
Very slick 💯 .

Did some testing on saf/ff/chrome and all seems to work great.

PR #16269, opened 2025-10-15
Fix S3 logging in the pulumi website

sicarul:
The legacy S3 delivery code we're using stopped pushing data on March 2024 - We need to migrate to:
https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/AWS-logs-infrastructure-V2-S3.html

This PR adds the new S3 bucket destination and removes the legacy one so we can start writing logs again, so we can consume them from the Data Warehouse - Also adds proper permissions so the data stack can access these logs.

To use an updated aws provider i had to also update nodejs, upgraded into node v22, i haven't seen any issues in the build, i hope that's ok?


CamSoper:
Thanks for doing this! This is data I badly need! Let me know if you need any help sorting the issues Claude found.

sicarul:
> Thanks for doing this! This is data I badly need! Let me know if you need any help sorting the issues Claude found.

Yeah it seems i missed some node version things, will ping again when i see the actions working, sorry!

CamSoper (approved):
Seems good to me. Thanks for doing this, @sicarul!

PR #16278, opened 2025-10-15
Add missing alias for Cloud REST API reference

komalali:
### Proposed changes

Adds an alias for the top link when googling "pulumi rest api" which is currently 404ing.


komalali:
@claude can I get another review?

kramhuber:
Confirmed that hitting `/docs/pulumi-cloud/reference/` correctly redirects

PR #16341, opened 2025-10-21
Setup OAC to restrict S3 bucket access properly

sicarul:
This PR sets up the S3 bucket as described in https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-s3.html

With this merged non-existent paths will return 404 instead of 403, e.g.:

Current behavior:

image

After this change: 

image


infrastructure/index.ts
```diff
@@ -804,6 +810,47 @@ const cdn = new aws.cloudfront.Distribution(
...
+                    Action: ["s3:GetObject", "s3:ListBucket"],
+                    Resource: [bucketArn, `${bucketArn}/*`],
+                    Condition: {
+                        StringEquals: {
+                            "AWS:SourceArn": distributionArn,
+                        },
+                    },
+                },
+                {
+                    Sid: "DenyDirectListBucket",
```
zacdirect: Deny's can be rather dangerous and I think should be expressly configured so there's not a risk they drift on context.

The allow as you have it should be sufficient in preventing list access to non-admins.  I would leave it off unless there are specific requirements for a deny.

sicarul: If you check at the beginning of the PR  this is the existing policy, but if we think it's safe we can remove it.

zacdirect: I'd remove it unless there's clear need or requirements to keep it.  The allow is already restrictive enough imo.

CamSoper (approved):
I don't have any feedback of note. Thank you for doing this!

zacdirect (approved):

infrastructure/index.ts
zacdirect: Approved to not be a blocker since it was also already there and it's really just a housecleaning item.

sicarul: I will remove it and test again tomorrow just to be safe!

PR #16358, opened 2025-10-22
Correct Neo API docs

flostadler:
### Proposed changes

The Neo API docs were wrong. This change corrects them


flostadler:
> Were this auto-gen'd from somewhere?

No, handwritten. Just wrongly, handwritten :(

foot (approved):
Looks pretty correct to me! Definitely an improvement

PR #16392, opened 2025-10-27
Fix obsolete navigation paths in ESC webhooks documentation

### LINKED ISSUE #16390: Location to webhook settings is no longer correct
### Problem description

- I saw a product documentation issue on the following URL: https://www.pulumi.com/docs/esc/environments/webhooks/#create-an-organization-webhook-in-the-pulumi-cloud-ui

### Suggestions for a fix

The location above is one of the locations in our docs where it mentions `Settings > Webhooks` to configure Stack, Deployment or Environment webhooks.

Due to the left navigation bar restructuring, the settings now live under `Settings > Integrations`, then scroll down to `Webhooks`.

jkodroff:
Fixes #16390 

Updates navigation instructions to reflect current Pulumi Cloud UI, where organization webhooks are now found under Integrations > Webhooks rather than Settings > Webhooks. Also reorders sections to present environment webhooks before organization webhooks for better logical flow.

Additional cleanup includes removing outdated trial callout and screenshot, fixing Python code example syntax errors, correcting grammar errors, and standardizing list formatting per style guide.


ringods (approved):
For organization webhooks, a user needs to be an admin of the org to set these up. Can you add that requirement somewhere?
Regarding the location, I would write `Settings > Integrations` to indicate that one needs to expand `Settings` first to find the `Integrations` item.

Do ESC environment webhooks also require admin permissions? I'm not sure about that.

PR #16401, opened 2025-10-27
Improve apply and all documentation clarity and correctness

### LINKED ISSUE #11953: Mention that calling export in an apply doesn't work
Via conversation in Community Slack, we should probably make it clearer what works and what doesn't, by language. (This issue was related to Python, but may be common to all languages.) 

https://pulumi-community.slack.com/archives/C84L4E3N1/p1717429416648089

### LINKED ISSUE #15952: Explain when *lifting* works and when it doesn't
[content/docs/iac/concepts/inputs-outputs/apply.md](https://www.pulumi.com/docs/iac/concepts/inputs-outputs/apply/)

In the Inputs/Outputs docs there is the following:
> An easier way to access deeply nested properties is by using *lifting*. Lifting allows you to access properties and elements directly from the [Output<T>](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/#Output) itself without needing [apply](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/#OutputInstance-apply).

Then after the example the section finishes with
>This approach is easier to read and write and does not lose any important dependency information that is needed to properly create and maintain the stack. This approach doesn’t work in all cases, but when it does, it can be a great help.

I agree the approach is easier to read and write and I'd like to understand when I can or cannot use it.

### LINKED ISSUE #16320: add much more information
[content/docs/iac/concepts/inputs-outputs/apply.md](https://www.pulumi.com/docs/iac/concepts/inputs-outputs/apply/)

For me (and I suspect others) apply is the most confusing aspect of working with Pulumi, and deserves much deeper information than whaat's provided here. I would expect to find guidance on:

- Limitations in using apply when working with dictionaris (apply cannot be used to resolve a single attribute of a dictionary)
- Guidance and best practices around defining resources inside an apply block
- Guidance around using nested apply blocks (apply inside apply)

One of the very reasons for using a full programming language to define IaC is that it allows complex nesting, and that there is native support for complex objects like dictionaries etc. Since using "apply" requires very careful design with both of these, I think it deserves much better documentation than what's currently provided. I think it would do wonders for users getting started with pulumi if the documentation was more real-life oriented.

### LINKED ISSUE #16393: Improve documentation for `apply`
A number of fixes related to https://www.pulumi.com/docs/iac/concepts/inputs-outputs/apply/

- The intro should start immediately with what `apply` is and what it's typically used for (with a link to each section that demonstrates the usage):
  - Printing output values for debugging Pulumi programs
  - Accessing nested values in complex types (that is, outputs that are objects or dictionaries)
  - Transforming an output by referencing its plain value
- Refer the reader to Inputs and Outputs for what outputs are and why they are necessary - don't restate that here.
- Move the note for what `all` is for as close to the top of the page as possible.
- Ensure that we call out (warning notes type) not to create resources in `apply` or `all` if it is avoidable (and it usually is avoidable) because they will not appear in `pulumi preview` unless the output's value is known when `pulumi preview` is run, in which case the `preview` output would not match the actual changes to resources when `pulumi up` is run.
- Remove the link to the gist, as it's not hosted/owned by Pulumi
- Remove the gigantic example at the top of the page that shows what happens if you try to print a Pulumi Resource since that's already covered in the "how to print a value" example. (Also fixes #11841)
- Make lifting its own section and add a note per #15952. We should also explain the types involved in the example.
- Add a note that stack outputs cannot be constructed within an apply (#11953)

jkodroff:
Fixes #16320
Fixes #16393
Fixes #15952
Fixes #11953 

Clarifies the purpose and proper usage of apply and all methods, adds important warnings about resource creation in callbacks to prevent preview inconsistencies, and fixes syntax errors in code examples to ensure accuracy.

### DIFF
--- a/content/docs/reference/cloud-rest-api/neo/_index.md
+++ b/content/docs/reference/cloud-rest-api/neo/_index.md
@@ -13,23 +13,15 @@ aliases:
 
 The Agent Tasks API allows you to create and manage AI agent tasks in Pulumi Cloud. These endpoints enable you to create tasks, monitor their status, respond to agent requests, and retrieve task events.
 
-## Authentication
+## Agent Task Operations
 
-All requests must be authenticated using a token via the `Authorization` HTTP header:
+The API provides endpoints for the following operations:
 
-```plain
-Authorization: token {token}
-```
-
-## Base URL
-
-For Managed Pulumi Cloud (app.pulumi.com):
-
-```plain
-https://api.pulumi.com
-```
-
-For Self-Hosted Pulumi Cloud, use your configured API endpoint.
+- Creating new agent tasks
+- Listing available tasks
+- Getting task details and metadata
+- Responding to agent requests
+- Retrieving task events
 
 ---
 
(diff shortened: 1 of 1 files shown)


CamSoper (approved):
Greatly improved, far better than the original. Thank you!

PR #16462, opened 2025-11-04
esc: move syntax reference under environments

pgavlin:
These changes move the ESC Environment syntax reference from the
reference portion of the docs to the Environments portion of the docs.

As it stands, the syntax reference--which is a primary source of
information for Environment authors--is not particularly discoverable,
as it is completely disconnected from the rest of the ESC documentation.


CamSoper (approved):
Good call on this, @pgavlin. This is definitely better.

The way I had it seemed like a good idea at the time. 🤷🏻‍♂️

PR #16501, opened 2025-11-06
Add Supabase case study

asafashirov:
Supabase case study

### DIFF
--- a/content/case-studies/supabase.md
+++ b/content/case-studies/supabase.md
@@ -43,7 +43,7 @@ sections:
 
 ## About Supabase {#about-supabase}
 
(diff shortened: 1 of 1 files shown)


asafashirov:
> Preview looks good to me. Do we want to include the customer's (Paul's) headshot somewhere to humanize it a bit more?

Sure - send me the headshot and I'll get it in there so we can see how it looks. I'll also be implementing some of the suggestions from the [Claude bot review](https://github.com/pulumi/docs/pull/16501#issuecomment-3497483201) in the next iteration.

scottmparker (approved):

PR #16514, opened 2025-11-07
Adding details on path parameter in automation api set config

pierskarsenbarg:
There have been a couple of questions about the `path` parameter in automation API. Here's the most recent on the subject of disabling default providers

### Proposed changes

* Updated blog post on disabling default providers to point to docs page
* Updated docs page on resource providers to include code for automation API

### DIFF
--- a/content/blog/disable-default-providers/index.md
+++ b/content/blog/disable-default-providers/index.md
@@ -314,7 +314,7 @@ configuration file:
 pulumi:disable-default-providers: ["*"]
 ```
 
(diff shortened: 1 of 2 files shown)


pierskarsenbarg:
@claude your suggestions for lines 606 and 613 are incorrect. Here's the link to the SDK for this function: https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/classes/automation.Stack.html#setConfig

CamSoper (approved):
It appears `@claude` didn't like being called out like that, @pierskarsenbarg! 😆  

This all looks fine to me. :shipit: when ready.

PR #16580, opened 2025-11-17
Correct docs in `import` with Diff

iwahbe:
The behavior of import changed with https://github.com/pulumi/pulumi/pull/19339, and our docs should reflect the current behavior.


iwahbe:
@claude Please implement your suggestions. For (3), let's use bash.

PR #16625, opened 2025-11-20
Add concept file for stash resource

Frassle:
Add a concept page for the new builtin Stash resource.

### DIFF
--- a/content/docs/iac/concepts/stash.md
+++ b/content/docs/iac/concepts/stash.md
@@ -331,19 +331,17 @@ outputs:
 
 {{< /chooser >}}
 
-When viewing stashed secret values, their plaintext content will not be shown by default. Instead, they will be displayed as `[secret]` in the CLI. Pass `--show-secrets` to the command ran to reveal the plaintext value (`--show-secrets)
+When viewing stashed secret values, their plaintext content will not be shown by default. Instead, they will be displayed as `[secret]` in the CLI. Pass `--show-secrets` to the command run to reveal the plaintext value.
 
 ## Updating stashed values
 
-To update the value stored in a `Stash` you need to replace the resource. There are a few ways to do this.
+To update the value stored in a `Stash` you need to replace it. There are a few ways to do this.
 
 1. Using the `--target-replace` argument to `up` to tell the engine to replace it.
 1. Using `pulumi state taint` to mark the resource to be replaced on the next deployment.
 1. Using the `TriggerReplacement` resource option to trigger the resource to replace on a change of value.
 
-Without a replacement any changes to the input `input` will be reflected in the output property `input`, but
-the output property `output` will not change. It will continue to return the original value the `Stash` was
-constructed with.
+Without a replacement, any changes to the `input` property will be reflected in the `input` output property, but the `output` property will not change. It will continue to return the original value the `Stash` was constructed with.
 
 ## Deleting a stash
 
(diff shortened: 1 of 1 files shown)


content/docs/iac/concepts/stash.md
```diff
@@ -0,0 +1,701 @@
...
+          Team: platform
+```
+
+{{% /choosable %}}
+
+{{< /chooser >}}
+
+## Stashing secret values
+
+The `Stash` resource respects secret annotations. If the `input` value is marked as a secret, the `output` will also be secret, and the value will be encrypted in your stack's state.
```
pgavlin, line 246: just curious--did we validate that this holds for nested inputs?

content/docs/iac/concepts/stash.md
```diff
@@ -0,0 +1,701 @@
...
+
+outputs:
+  apiKey: ${apiKeyStash.output}
+```
+
+{{% /choosable %}}
+
+{{< /chooser >}}
+
+When viewing stashed secret values, their plaintext content will not be shown by default. Instead, they will be displayed as `[secret]` in the CLI. Pass `--show-secrets` to reveal the plaintext value.
```
pgavlin: pass `--show-secrets` where? `up`/`preview`/etc.?

content/docs/iac/concepts/stash.md
```diff
@@ -0,0 +1,701 @@
...
+
+## Updating stashed values
+
+To update the value stored in a `Stash` you need to replace the resource. There are a few ways to do this.
+
+1. Using the `--target-replace` argument to `up` to tell the engine to replace it.
+1. Using `pulumi state taint` to mark the resource to be replaced on the next deployment.
+1. Using the `TriggerReplacement` resource option to trigger the resource to replace on a change of value.
```
pgavlin, line 342: I'd reorder these so the resource option is first, personally.

also wondering if we should consider a `keeper`-style property in the future that has replace semantics

content/docs/iac/concepts/stash.md
Frassle: We roundtrip the property value as-is, so it will depend on the language marshaling layer but Stash never reduces the amount of secretness

content/docs/iac/concepts/stash.md
Frassle: up, preview, stack, export, there's a load of commands that take this option. Maybe 
"Pass `--show-secrets` to the command ran to reveal the plaintext value."?

content/docs/iac/concepts/stash.md
Frassle: > also wondering if we should consider a keeper-style property in the future that has replace semantics

Briefly what would that look like? I'd have thought that's what TriggerReplacement is.

pgavlin: Ah I think I misunderstood that bullet point. Yeah we can just use that.

Though that does raise another question--where are the docs for that option? I don't see it here: https://www.pulumi.com/docs/iac/concepts/resources/options/

Kind of a meta-point that it would be great to add links to the docs for the various suggested commands as well.

content/docs/iac/concepts/stash.md
pgavlin: yeah that sounds reasonable

content/docs/iac/concepts/stash.md
Frassle: TomH is still getting around to finishing that workstream off. I was planning on linking it once it was in place.

PR #16630, opened 2025-11-21
Document provider layering, with impl guides

joeduffy:
Today we do not really explain the layering of our various means of writing providers, which as we've seen, can lead to confusion.

This does a pass over the provider authoring docs, explaining the architecture, and why only Go has a higher level SDK.

It also adds reference material and implementation guides for those who want to code to the direct gRPC/Protobuf interfaces directly, in Python, Go, or TypeScript. (These were tested locally.)

I am not 100% convinced we should add this but for power users wanting to really understand how providers work and how to author them, it would seem helpful in demystifying some confusing aspects of our ecosystem.

### DIFF
--- a/content/docs/iac/guides/building-extending/providers/implementers/typescript.md
+++ b/content/docs/iac/guides/building-extending/providers/implementers/typescript.md
@@ -45,9 +45,13 @@ npm init -y
 Install dependencies:
 
 ```bash
-npm install @pulumi/pulumi @grpc/grpc-js google-protobuf
+npm install @pulumi/pulumi @grpc/grpc-js google-protobuf@3.21.4
 npm install -D typescript @types/node @types/google-protobuf
 ```
+
+{{% notes type="info" %}}
+Use google-protobuf version 3.21.4 to match the version bundled with the Pulumi SDK. Version 4.x has incompatible API changes.
+{{% /notes %}}
 
 Create `tsconfig.json`:
 


content/docs/iac/guides/building-extending/providers/provider-architecture.md
```diff
@@ -0,0 +1,97 @@
...
+| **Schema** | Hand-written JSON | Inferred from code |
+| **Boilerplate** | More | Less |
+| **Control** | Full | Framework-guided |
+| **Learning curve** | Protocol knowledge required | Go SDK patterns |
+
+### Choosing a language
+
+Many teams prefer to write providers in their standard language, giving them access to familiar tools and libraries. However, Go has a significant advantage: providers written in Go compile to standalone binaries with no runtime dependencies. Users of your provider can consume it from any Pulumi language without installing Go.
+
+Providers written in Python or TypeScript require their respective runtimes to be installed, which adds friction for users. That said, if your team is more productive in Python or TypeScript, the familiar ecosystem and libraries may outweigh the runtime dependency—especially for internal providers where you control the deployment environment.
```
justinvp: Not sure it's worth mentioning, but it's possible to create a Python or TypeScript provider that's wrapped in a native binary. Not a very common thing to do, but it's possible.

content/docs/iac/guides/building-extending/providers/implementers/go.md
```diff
@@ -0,0 +1,497 @@
...
+
+This guide shows how to implement a Pulumi provider in Go using the gRPC bindings directly, without the [Pulumi Go Provider SDK](/docs/iac/guides/building-extending/providers/sdks/pulumi-go-provider-sdk/). This approach gives you full control over provider behavior and schema definition.
+
+{{% notes type="info" %}}
+For most Go providers, the [Pulumi Go Provider SDK](/docs/iac/guides/building-extending/providers/sdks/pulumi-go-provider-sdk/) provides a more ergonomic experience with schema inference and less boilerplate. Use this direct approach when you need precise control over the protocol or want to understand exactly how providers work.
+{{% /notes %}}
+
+## Prerequisites
+
+You'll need Go 1.21 or later, and a basic understanding of the [provider protocol](/docs/iac/guides/building-extending/providers/implementers/protocol-reference/).
```
justinvp: Nit: Go 1.24 is currently the lowest supported version of Go, so not sure why we'd mention 1.21.

content/docs/iac/guides/building-extending/providers/implementers/go.md
```diff
@@ -0,0 +1,497 @@
...
+
+```go
+func getTypeFromURN(urn string) string {
+	// URN format: urn:pulumi:<stack>::<project>::<type>::<name>
+	parts := strings.Split(urn, "::")
+	if len(parts) >= 3 {
+		return parts[2]
+	}
+	return ""
+}
```
justinvp: As of https://github.com/pulumi/pulumi/pull/17177 and [v3.132.0](https://github.com/pulumi/pulumi/releases/tag/pkg/v3.132.0) of the CLI, we send the parsed type in the requests, so probably no need to show urn parsing fallback here. Just use `req.Type` and assume users are creating providers for v3.132.0 and later of pulumi from a year ago.

content/docs/iac/guides/building-extending/providers/implementers/go.md
```diff
@@ -0,0 +1,497 @@
...
+		return &rpc.ReadResponse{}, nil
+	}
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to read file: %v", err)
+	}
+
+	oldInputs := req.Inputs.AsMap()
+	force, _ := oldInputs["force"].(bool)
+
+	state := map[string]interface{}{
```
justinvp: Nit: Might as well use modern go! There are some other uses of `interface{}` that could be cleaned up as well.

```suggestion
	state := map[string]any{
```

content/docs/iac/guides/building-extending/providers/implementers/protocol-reference.md
```diff
@@ -0,0 +1,339 @@
...
+menu:
+    iac:
+        name: Protocol reference
+        parent: iac-guides-provider-implementers
+        weight: 10
+---
+
+This document describes the gRPC protocol that all Pulumi providers implement. Understanding this protocol is essential for [implementing providers directly](/docs/iac/guides/building-extending/providers/provider-architecture/#layer-2-generated-language-bindings) without a higher-level SDK.
+
+The complete protocol definition is in [`provider.proto`](https://github.com/pulumi/pulumi/blob/master/proto/pulumi/provider.proto).
```
justinvp: We also have this published at https://pulumi-developer-docs.readthedocs.io/latest/docs/_generated/proto/provider.html. Not sure if we want to link to that from here or not.

content/docs/iac/guides/building-extending/providers/implementers/protocol-reference.md
```diff
@@ -0,0 +1,339 @@
...
+
+### Create
+
+Provisions a new resource.
+
+```protobuf
+rpc Create(CreateRequest) returns (CreateResponse)
+```
+
+**Request fields:**
```
justinvp, line 144: We should probably show `type` and `name` fields for all relevant requests.

justinvp (commented):
I like it! Some initial comments. I haven't looked at the python or typescript pages yet.

content/docs/iac/guides/building-extending/providers/provider-architecture.md
joeduffy: What do recommend for doing that?

content/docs/iac/guides/building-extending/providers/implementers/protocol-reference.md
```diff
@@ -0,0 +1,347 @@
...
+
+Initializes the provider with configuration values like credentials and region settings.
+
+```protobuf
+rpc Configure(ConfigureRequest) returns (ConfigureResponse)
+```
+
+**Request fields:**
+
+- `variables` (map<string, string>): Environment variables
```
Frassle, line 60: This isn't environment variables, it's the same as "args" but string-ified, we don't recommend using it anymore.

PR #16644, opened 2025-11-21
Add additional instructions for reviewing blog posts on new Pulumi features

CamSoper:
Enhance the review process for blog posts that introduce new Pulumi features by adding specific instructions to check for corresponding documentation and identify gaps.


meagancojocar:
@jkodroff think you'll like this

PR #16651, opened 2025-11-24
Update python & typescript provider doc updates

julienp:
* Update the minimum versions of TS and Python to match our supported versions

* The Python entrypoint should be at the root

* Fix magic values for secrets and unknowns

* Packaging & distrubution. We don’t need a wrapper script, Pulumi will run the `__main__.py` entrypoint. Virtualenvs are not portable, so we never want to add that to the archive.

### DIFF
--- a/content/docs/iac/guides/building-extending/providers/implementers/python.md
+++ b/content/docs/iac/guides/building-extending/providers/implementers/python.md
@@ -25,11 +25,10 @@ You'll need Python 3.10 or later, the `grpcio` and `grpcio-tools` packages, and 
 
 ```
 my-provider/
+├── __main__.py               # Entry point
 ├── provider/
 │   ├── __init__.py
-│   ├── __main__.py          # Entry point
-│   └── server.py            # gRPC server implementation
-├── pulumi-resource-myfiles   # Wrapper script (executable)
+│   └── server.py             # gRPC server implementation
 ├── schema.json               # Provider schema
 ├── PulumiPlugin.yaml         # Plugin metadata
 └── requirements.txt


content/docs/iac/guides/building-extending/providers/implementers/python.md
```diff
@@ -539,25 +539,25 @@ During preview, some values are unknown:
...
 ```
 
-Make it executable with `chmod +x pulumi-resource-myfiles`. This script activates the virtual environment and runs the provider module. Package everything including the `venv` directory into a distributable archive, or install dependencies system-wide and adjust the script accordingly.
+This plugin can then be installed as a regular pulumi resource plugin:
+
+```bash
+pulumi plugin install resource myfiles --file ./myfiles.tar.gz 0.0.1
+```
 
 See [Publishing packages](/docs/iac/guides/building-extending/packages/publishing-packages/) for full details.
```
julienp: Do we have a friendlier guide to packaging and SDK generation somewhere?

Something that roughly covers:

* Upload your .tar.gz somewhere with the expected filename for os/arch, etc. Note this is kinda dumb for Python, but I think this is geared towards binary providers
* Set PluginDownloadURL in the schema with the download root
* Generate language SDKs
* Publish language SDKs


Alternatively, we could provide instructions for a local SDK here instead. These only require making the provider source available somewhere, bypassing the tedium of SDK publishing.

julienp: For now I updated this to link out to the publishing docs and to the local SDK docs.

PR #16692, opened 2025-12-01
Add journaling blog post

tgummerer:
Add blog post for journaling (https://github.com/pulumi/pulumi/issues/13502)

### Related issues (optional)

https://github.com/pulumi/pulumi/issues/13502

### DIFF
--- a/content/blog/journaling/index.md
+++ b/content/blog/journaling/index.md
@@ -1,7 +1,7 @@
 ---
 title: "Speeding up Pulumi Operations by up to 20x"
 
-date: 2025-12-08T17:57:55+02:00
+date: 2026-01-12T18:57:55+02:00
 
 draft: false
 


content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+
+social:
+    twitter: Speeding up your Pulumi deployments by 10x
+    linkedin: Want faster Pulumi deployments on big stacks? Read on how we improved the performance of Pulumi deployments with lots of resources by up to 10x, and how you can get access to it today.
+
+# See the blogging docs at https://github.com/pulumi/docs/blob/master/BLOGGING.md
+# for details, and please remove these comments before submitting for review.
+---
+
+Pulumi saves a snapshot of the current state of your cloud infrastructure at every deployment, and also at every step of the deployment. This means that Pulumi always has a current view of the state even if there are crashes during an operation. However, this comes with a performance penalty especially for large stacks. Today we're introducing an improvement that can speed up deployments up to 10x. Read on for benchmarks and some technical details of the implementation.
```
julienp:
```suggestion
Pulumi saves a snapshot of the current state of your cloud infrastructure at every deployment, and also at every step of the deployment. This means that Pulumi always has a current view of the state even if there is an issue during an operation. However, this comes with a performance penalty especially for large stacks. Today we're introducing an improvement that can speed up deployments up to 10x. Read on for benchmarks and some technical details of the implementation.
```

Maybe slightly softer "issue" instead of "crashes". This also covers things like the network dropping, or it being a problem on the cloud side.

content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+
+<!--more-->
+
+## Benchmarks
+
+Before getting into the more technical details, here's a number of benchmarks demonstrating what this new experience looks like. To run the benchmarks we picked a couple of Pulumi projects, one that can be set up massively parallel, which is the worst case scenario for the old snapshot system, and another one that looks a little more like a real world example.  Note that all of these benchmarks were conducted in Europe connecting to Pulumi Cloud, which runs in `us-west-2`, so exact numbers may vary based on your location and internet connection. This should however give a good indication of the performance improvements.
```
julienp:
```suggestion
Before getting into the more technical details, here are a number of benchmarks demonstrating what this new experience looks like. To run the benchmarks we picked a couple of Pulumi projects, one that can be set up massively parallel, which is the worst case scenario for the old snapshot system, and another one that looks a little more like a real world example.  Note that all of these benchmarks were conducted in Europe connecting to Pulumi Cloud, which runs in `us-west-2`, so exact numbers may vary based on your location and internet connection. This should however give a good indication of the performance improvements.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+
+We're benchmarking two somewhat large stacks, both of which are or were used at Pulumi. The first program sets up a website using AWS bucket objects. We're using the [example-ts-static-website](https://github.com/pulumi/examples/tree/master/aws-ts-static-website) example here, but expand it a little bit to set up a version of our docs site. This means we're setting up more than 3000 bucket objects, with 3222 resources in total.
+
+The benchmarks were measured using `time` built-in command and using the best time in a best-of-three benchmarks. The network traffic using `tcpdump`, limiting the measured traffic to only the IP addresses for Pulumi Cloud. Finally `tshark` was used to process the packet captures and count the bytes sent.
```
julienp:
```suggestion
The benchmarks were measured using `time` built-in command and using the best time in a best-of-three benchmark. The network traffic was measured using `tcpdump`, limiting the measured traffic to only the IP addresses for Pulumi Cloud. Finally `tshark` was used to process the packet captures and count the bytes sent.
```

julienp (commented):
didn't read all of it yet, couple nits

content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+
+To make sure there are never any resources that are not tracked, even if a deployment is aborted unexpectedly (for example due to network issues, power outages, or bugs), Pulumi creates a new snapshot at the beginning and at the end of each operation.
+
+At the beginning of the operation, Pulumi adds a new "pending operation" to the snapshot. Pending operations declare the intent to mutate a resource. If a pending operation is left in the snapshot (in other words the operation started, but Pulumi couldn't record the end of it), in the next operation Pulumi asks the user to check the actual state of the resource, and then either removes it from the snapshot, or imports it depending on the users input.
+
+This is because it is possible that the resource has been set up correctly, or it is possible that the resource creation failed. If Pulumi aborted midway through the operation it's impossible to know which it is.
+
+Once an operation finished, the pending operation is removed, as we now know the final state of the resource, and the final state of the resource is updated in the snapshot.
```
julienp:
```suggestion
Once an operation finishes, the pending operation is removed, as we now know the final state of the resource, and the final state of the resource is updated in the snapshot.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+
+There's also some additional metadata that is stored in the snapshot, that is only updated infrequently.
+
+Here's how the snapshot looks in code. This snapshot is serialized and sent to the backend. `Resources` holds the list of known resource state, and is updated after each operation finishes, and `PendingOperations` is the list of pending operations described above.
```
julienp:
```suggestion
Here's how the snapshot looks in code. This snapshot is serialized and sent to the backend. `Resources` holds the list of known resource states, and is updated after each operation finishes, and `PendingOperations` is the list of pending operations described above.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+
+The benchmarks were measured using `time` built-in command and using the best time in a best-of-three benchmark. The network traffic was measured using `tcpdump`, limiting the measured traffic to only the IP addresses for Pulumi Cloud. Finally `tshark` was used to process the packet captures and count the bytes sent.
+
+All the benchmarks are run with journaling off (the default experience), with journaling on (the new experience), and finally with `PULUMI_SKIP_CHECKPOINTS=true` set. The last one means we skip uploading intermediate checkpoints to the backend, which in turn means potentially losing track of changes that are in flight if Pulumi exits unexpectedly due to any reason.
+
+|                    | Time   | Bytes sent |
+|--------------------|--------|------------|
+| Without journaling | 58m26s | 16.5MB     |
```
kramhuber, line 72: Nit: if you have time, this table would be very powerful as a grouped bar chart.

Edit: Created here, though we may want to add titles https://docs.google.com/spreadsheets/d/1L_CSI7_fkrJogGIHPeeteUJzSHQ3SZQx5RK75WB03Uc/edit?usp=sharing

content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+
+social:
+    twitter: Speeding up your Pulumi deployments by 10x
+    linkedin: Want faster Pulumi deployments on big stacks? Read on how we improved the performance of Pulumi deployments with lots of resources by up to 10x, and how you can get access to it today.
+
+# See the blogging docs at https://github.com/pulumi/docs/blob/master/BLOGGING.md
+# for details, and please remove these comments before submitting for review.
+---
+
+Pulumi saves a snapshot of the current state of your cloud infrastructure at every deployment, and also at every step of the deployment. This means that Pulumi always has a current view of the state even if there is an issue during an operation. However, this comes with a performance penalty especially for large stacks. Today we're introducing an improvement that can speed up deployments up to 10x. Read on for benchmarks and some technical details of the implementation.
```
kramhuber:
```suggestion
Today we're introducing an improvement that can speed up deployments up to 10x. At every deployment, and at every step within a deployment, Pulumi saves a snapshot of your cloud infrastructure. This gives Pulumi a current view of state even if something fails mid-operation, but it comes with a performance penalty for large stacks. Here's how we fixed it.
```

kramhuber: > at every **deployment**

Is **deployment** the term we want to use? Operation? Update?

content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+|--------------------|--------|------------|
+| Without journaling | 17m52s | 18.5MB     |
+| With journaling    | 9m45s  | 5.9MB      |
+| Skip checkpoints   | 8m39s  | 2MB        |
+
+Note that this feature is still behind a feature flag, but we are ready for testers. To get enrolled in the feature flag, please reach out to us, either on the [Community Slack](https://slack.pulumi.com/), or through our [Support channels](https://support.pulumi.com/hc/en-us). Once that's done, all you need to do is to set the `PULUMI_ENABLE_JOURNALING` environment variable to `true`, and your deployments will start finishing faster.
+
+If you are interested in the more technical details read on!
+
+## Introduction into snapshotting
```
kramhuber: I would make this the first section, before benchmarks; otherwise, the user doesn't know what's being benchmarked

content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+
+# See the blogging docs at https://github.com/pulumi/docs/blob/master/BLOGGING.md
+# for details, and please remove these comments before submitting for review.
+---
+
+Pulumi saves a snapshot of the current state of your cloud infrastructure at every deployment, and also at every step of the deployment. This means that Pulumi always has a current view of the state even if there is an issue during an operation. However, this comes with a performance penalty especially for large stacks. Today we're introducing an improvement that can speed up deployments up to 10x. Read on for benchmarks and some technical details of the implementation.
+
+<!--more-->
+
+## Benchmarks
```
kramhuber, line 60: I get you're trying to "get to the good stuff", but consider moving this to after "why is it slow".

content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+- Next up was adding a feature flag for the service, so journaling could be turned on selectively for some orgs. At the same time we implemented an opt-in environment variable in the CLI (`PULUMI_ENABLE_JOURNALING`), so the feature could be selectively turned on by users, if both the feature flag is enabled and the user sets the environment variable. This way we could slowly start enabling this in our repos, e.g. first in the integration tests for `pulumi/pulumi`, then in the tests for `pulumi/examples` and `pulumi/templates`, etc.
+- Allow users to start opting in.  If you want to opt-in with your org, please reach out to us, either on the [Community Slack](https://slack.pulumi.com/), or through our [Support channels](https://support.pulumi.com/hc/en-us), and we'll opt your org into the feature flag. Then you can begin seeing the performance improvements by setting the `PULUMI_ENABLE_JOURNALING` env variable to true.
```
kramhuber, line 321: This is the action we want a user to take so let's consider how we can hoist it out of the list so it's more apparent, such as "what we've done so far" and "what we're doing now"

content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+
+To make sure there are never any resources that are not tracked, even if a deployment is aborted unexpectedly (for example due to network issues, power outages, or bugs), Pulumi creates a new snapshot at the beginning and at the end of each operation.
+
+At the beginning of the operation, Pulumi adds a new "pending operation" to the snapshot. Pending operations declare the intent to mutate a resource. If a pending operation is left in the snapshot (in other words the operation started, but Pulumi couldn't record the end of it), in the next operation Pulumi asks the user to check the actual state of the resource, and then either removes it from the snapshot, or imports it depending on the users input.
+
+This is because it is possible that the resource has been set up correctly, or it is possible that the resource creation failed. If Pulumi aborted midway through the operation it's impossible to know which it is.
```
kramhuber:
```suggestion
At the beginning of the operation, Pulumi adds a new "pending operation" to the snapshot. Pending operations declare the intent to mutate a resource. If a pending operation is left in the snapshot (in other words the operation started, but Pulumi couldn't record the end of it), in the next operation Pulumi asks the user to check the actual state of the resource, and then either removes it from the snapshot, or imports it depending on the users input. This is because it is possible that the resource has been set up correctly, or it is possible that the resource creation failed. If Pulumi aborted midway through the operation it's impossible to know which it is.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+# generated blog page. They are here strictly for reference.
+
+# Here are some examples of posts we have made in the past for inspiration:
+# https://www.linkedin.com/feed/update/urn:li:activity:7171191945841561601
+# https://www.linkedin.com/feed/update/urn:li:activity:7169021002394296320
+# https://www.linkedin.com/feed/update/urn:li:activity:7155606616455737345
+# https://twitter.com/PulumiCorp/status/1763265391042654623
+# https://twitter.com/PulumiCorp/status/1762900472489185492
+# https://twitter.com/PulumiCorp/status/1755637618631405655
+
```
kramhuber, line 31:
```suggestion

```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+social:
+    twitter: Speeding up your Pulumi deployments by 10x
+    linkedin: Want faster Pulumi deployments on big stacks? Read on how we improved the performance of Pulumi deployments with lots of resources by up to 10x, and how you can get access to it today.
+
+# See the blogging docs at https://github.com/pulumi/docs/blob/master/BLOGGING.md
+# for details, and please remove these comments before submitting for review.
```
kramhuber:
```suggestion

```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+
+date: 2025-12-08T17:57:55+02:00
+
+draft: false
+
+# Use the meta_desc property to provide a brief summary (one or two sentences)
+# of the content of the post, which is useful for targeting search results or
+# social-media previews. This field is required or the build will fail the
+# linter test. Max length is 160 characters.
+meta_desc: Learn how journaling helps speed up your deployments in large stacks up to 10x, while maintaining full data integrity guarantees throughout the update
```
kramhuber:
```suggestion
meta_desc: Pulumi deployments get up to 10x faster with journaling, a new snapshotting approach that speeds up large stacks while keeping full data integrity.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+# Here are some examples of posts we have made in the past for inspiration:
+# https://www.linkedin.com/feed/update/urn:li:activity:7171191945841561601
+# https://www.linkedin.com/feed/update/urn:li:activity:7169021002394296320
+# https://www.linkedin.com/feed/update/urn:li:activity:7155606616455737345
+# https://twitter.com/PulumiCorp/status/1763265391042654623
+# https://twitter.com/PulumiCorp/status/1762900472489185492
+# https://twitter.com/PulumiCorp/status/1755637618631405655
+
+social:
+    twitter: Speeding up your Pulumi deployments by 10x
```
kramhuber: Pulumi deployments just got up to 10x faster:

- Journaling: Send only changes, not full snapshots
- Data integrity: No compromise on reliability
- Network traffic cut by 85%+ on large stacks

Try it: [link]

content/blog/journaling/index.md
```diff
@@ -0,0 +1,313 @@
...
+    linkedin: Want faster Pulumi deployments on big stacks? Read on how we improved the performance of Pulumi deployments with lots of resources by up to 10x, and how you can get access to it today.
```
kramhuber: **Pulumi Deployments Get Up to 10x Faster**

Large Pulumi stacks just got a major performance boost. Here's what changed.

**The Problem**
Pulumi saves snapshots at every deployment step to maintain data integrity. For large stacks, this creates a bottleneck: uploading the full snapshot serially slows everything down.

**The Solution: Journaling**
Instead of sending the whole snapshot, journaling sends only individual changes. These journal entries can go in parallel, and Pulumi Cloud reconstructs the full snapshot on the backend.

**The Results**
In benchmarks on a 3,000+ resource stack:
- Time dropped from 58 minutes to 3 minutes
- Network traffic cut from 16.5MB to 2.3MB

**Why This Matters**
You get the speed without sacrificing data integrity. Unlike SKIP_CHECKPOINTS, journaling still tracks all in-flight operations. If something fails mid-deployment, Pulumi still knows exactly what happened.

**Get Started**
This feature is in opt-in testing. Reach out on Pulumi Community Slack or through Support to get your org enrolled. Then set PULUMI_ENABLE_JOURNALING=true.

Read more: [link]

content/blog/journaling/index.md
tgummerer: Yeah maybe operation is easier to understand?  Happy to switch it around everywhere.

content/blog/journaling/index.md
tgummerer: Yeah, I wanted to start off with "why do I even care", and demonstrate that, and give just enough information to understand what is being benchmarked, rather than going into the technical details too early (I think they are cool and make for a well rounded blog post, but I didn't want to bury the lead).

content/blog/journaling/index.md
tgummerer: Good point, let me see if I can make this more apparent.

content/blog/journaling/index.md
tgummerer: Nice!

content/blog/journaling/index.md
tgummerer: Nice, thanks!

content/blog/journaling/index.md
tgummerer: Hmm looking at it again, I did try to highlight this in the benchmark section, so people can see it early :thinking: do you think that's not enough?  Any other ideas how to make this more visible?

content/blog/journaling/index.md
tgummerer: That's a great idea. I changed them a bit here: https://docs.google.com/spreadsheets/d/1RTJELRIYfb9xhwfNGC0gy-cyvqQUyCyycpTIP1Ucmqo/edit?gid=0#gid=0 (added pulumi colors and reduced the number of charts).

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
...
+
+date: 2025-12-08T17:57:55+02:00
+
+draft: false
+
+# Use the meta_desc property to provide a brief summary (one or two sentences)
+# of the content of the post, which is useful for targeting search results or
+# social-media previews. This field is required or the build will fail the
+# linter test. Max length is 160 characters.
+meta_desc: Pulumi deployments get up to 20x faster with journaling, a new snapshotting approach that speeds up large stacks while keeping full data integrity.
```
pgavlin: q: might suggest "operations" or "updates" rather than "deployments" throughout given potential for confusion with the Pulumi Deployments product (though that will also get faster as a side effect ;))

```suggestion
meta_desc: Pulumi operations get up to 20x faster with journaling, a new snapshotting approach that speeds up large stacks while keeping full data integrity.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
...
+    linkedin: |
+        # Pulumi Deployments Get Up to 20x Faster
+
+        Large Pulumi stacks just got a major performance boost. Here's what changed.
+
+        # The Problem
+        Pulumi saves snapshots at every deployment step to maintain data integrity. For large stacks, this creates a bottleneck: uploading the full snapshot serially slows everything down.
+
+        # The Solution: Journaling
+        Instead of sending the whole snapshot, journaling sends only individual changes. These journal entries can go in parallel, and Pulumi Cloud reconstructs the full snapshot on the backend.
```
pgavlin:
```suggestion
        Instead of sending the whole snapshot, journaling sends only individual changes. These journal entries can be written in parallel, and Pulumi Cloud can reconstruct the full snapshot on the backend.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
...
+
+        # The Results
+        In benchmarks on a 3,000+ resource stack:
```
pgavlin:
```suggestion
        In benchmarks when creating a 3,000+ resource stack:
```

content/blog/journaling/index.md
pgavlin: (happy to keep the existing wording, though, if folks disagree)

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
...
+
+---
+
+Today we're introducing an improvement that can speed up deployments up to 20x. At every deployment, and at every step within a deployment, Pulumi saves a snapshot of your cloud infrastructure. This gives Pulumi a current view of state even if something fails mid-operation, but it comes with a performance penalty for large stacks. Here's how we fixed it.
+
+<!--more-->
+
+## Benchmarks
+
+Before getting into the more technical details, here are a number of benchmarks demonstrating what this new experience looks like. To run the benchmarks we picked a couple of Pulumi projects, one that can be set up massively parallel, which is the worst case scenario for the old snapshot system, and another one that looks a little more like a real world example.  Note that all of these benchmarks were conducted in Europe connecting to Pulumi Cloud, which runs in `us-west-2`, so exact numbers may vary based on your location and internet connection. This should however give a good indication of the performance improvements.
```
pgavlin:
```suggestion
Before getting into the more technical details, here are a number of benchmarks demonstrating what this new experience looks like. To run the benchmarks we picked a couple of Pulumi projects: one that can be set up massively parallel, which is the worst case scenario for the old snapshot system, and another that looks a little more like a real world example.  Note that all of these benchmarks were conducted in Europe connecting to Pulumi Cloud, which runs in AWS's `us-west-2` region, so exact numbers may vary based on your location and internet connection. This should however give a good indication of the performance improvements.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
...
+
+We're benchmarking two somewhat large stacks, both of which are or were used at Pulumi. The first program sets up a website using AWS bucket objects. We're using the [example-ts-static-website](https://github.com/pulumi/examples/tree/master/aws-ts-static-website) example here, but expand it a little bit to set up a version of our docs site. This means we're setting up more than 3000 bucket objects, with 3222 resources in total.
```
pgavlin:
```suggestion
We're benchmarking two somewhat large stacks, both of which are or were used at Pulumi. The first program sets up a website using AWS bucket objects. We're using the [aws-ts-static-website](https://github.com/pulumi/examples/tree/master/aws-ts-static-website) example here, but expand it a little bit to set up a version of our docs site. This means we're setting up more than 3000 bucket objects, with 3222 resources in total.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
...
+
+The benchmarks were measured using `time` built-in command and using the best time in a best-of-three benchmark. The network traffic was measured using `tcpdump`, limiting the measured traffic to only the IP addresses for Pulumi Cloud. Finally `tshark` was used to process the packet captures and count the bytes sent.
+
+All the benchmarks are run with journaling off (the default experience), with journaling on (the new experience), and finally with `PULUMI_SKIP_CHECKPOINTS=true` set. The last one means we skip uploading intermediate checkpoints to the backend, which in turn means potentially losing track of changes that are in flight if Pulumi exits unexpectedly due to any reason.
```
pgavlin:
```suggestion
All the benchmarks are run with journaling off (the default experience), with journaling on (the new experience), and finally with `PULUMI_SKIP_CHECKPOINTS=true` set. The last configuration skips uploading intermediate checkpoints to the backend, which increases performance at the cost of potentially losing track of changes that are in flight if the `pulumi` CLI loses connectivity or exits unexpectedly.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
...
+
+![Time taken](time.png)
+
+![Data sent](size.png)
+
+*Note that this feature is still behind a feature flag, but we are ready for testers. To get enrolled in the feature flag, please reach out to us, either on the [Community Slack](https://slack.pulumi.com/), or through our [Support channels](https://support.pulumi.com/hc/en-us). Once that's done, all you need to do is to set the `PULUMI_ENABLE_JOURNALING` environment variable to `true`, and your deployments will start finishing faster.*
+
+If you are interested in the more technical details read on!
+
+## Introduction into snapshotting
```
pgavlin:
```suggestion
## Introduction to snapshotting
```

content/blog/journaling/index.md
pgavlin: Personally I'd keep this here--I think we risk burying the lede otherwise

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
...
+
+Pulumi keeps track of all resources in a stack in a snapshot. This snapshot is stored in the chosen backend, either in Pulumi Cloud, or in a DIY backend. Further deployments of the stack then use this snapshot to figure out which resources need to be created, updated or deleted.
```
pgavlin:
```suggestion
Pulumi keeps track of all resources in a stack in a snapshot. This snapshot is stored in the stack's configured backend, which is either the Pulumi Cloud or a DIY backend. Further deployments of the stack then use this snapshot to figure out which resources need to be created, updated or deleted.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
...
+
+To make sure there are never any resources that are not tracked, even if a deployment is aborted unexpectedly (for example due to network issues, power outages, or bugs), Pulumi creates a new snapshot at the beginning and at the end of each operation.
+
+At the beginning of the operation, Pulumi adds a new "pending operation" to the snapshot. Pending operations declare the intent to mutate a resource. If a pending operation is left in the snapshot (in other words the operation started, but Pulumi couldn't record the end of it), in the next operation Pulumi asks the user to check the actual state of the resource, and then either removes it from the snapshot, or imports it depending on the users input. This is because it is possible that the resource has been set up correctly, or it is possible that the resource creation failed. If Pulumi aborted midway through the operation it's impossible to know which it is.
```
pgavlin:
```suggestion
At the beginning of the operation, `pulumi` adds a new "pending operation" to the snapshot. Pending operations declare the intent to mutate a resource. If a pending operation is left in the snapshot (in other words the operation started, but `pulumi` couldn't record the end of it),  the next operation will ask the user to check the actual state of the resource. Depending on the user's response, `pulumi` will either remove the operation from the snapshot or import the resource . This is because it is possible that the resource has been set up correctly or that the resource creation failed. If `pulumi` aborted midway through the operation, it's impossible to know which state the resource is in.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
...
+
+Once an operation finishes, the pending operation is removed, as we now know the final state of the resource, and the final state of the resource is updated in the snapshot.
```
pgavlin:
```suggestion
Once an operation finishes, the pending operation is removed and the resource's final state is recorded in the snapshot.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
...
+
+There's also some additional metadata that is stored in the snapshot, that is only updated infrequently.
```
pgavlin:
```suggestion
There's also some additional metadata that is stored in the snapshot that is only updated infrequently.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
...
+type Snapshot struct {
+	Manifest          Manifest             // a deployment manifest of versions, checksums, and so on.
+	SecretsManager    secrets.Manager      // the secrets manager to use when serializing this snapshot.
+	Resources         []*resource.State    // all resources and their associated states.
+	PendingOperations []resource.Operation // all currently pending resource operations.
+	Metadata          SnapshotMetadata     // metadata associated with the snapshot.
+}
+```
+
+Before we dive in deeper, we also need to understand a little bit about how the Pulumi engine works internally. Whenever a Pulumi operation, e.g. `pulumi up`, `pulumi destroy`, `pulumi refresh` etc. is run, the engine internally generates a series of steps, to create, update, delete etc. resources. This series of steps is then executed. To maintain the correct relationship, the steps need to be executed in a partial order, where all steps whose dependent steps have been executed already can be executed in parallel.
```
pgavlin:
```suggestion
Before we dive in deeper, we also need to understand a little bit about how the Pulumi engine works internally. Whenever a Pulumi operation, e.g. `pulumi up`, `pulumi destroy`, `pulumi refresh` etc. is run, the engine internally generates and executes a series of steps, to create, update, delete etc. resources. To maintain correct relationships between resources, the steps need to be executed in a partial order such that no step is executed until all of its step dependencies have successfully executed. Steps may otherwise execute concurrently.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
...
+
+Our workaround for that is to serialize the snapshot uploads, uploading one snapshot at a time. This gives us the data integrity properties we want, however it can slow step execution down, especially on internet connections with lower bandwidth, and/or high latency.
+
+This impacts performance especially for large stacks, as we upload the whole snapshot every time, which can take some time if the snapshot is getting big. For the Pulumi Cloud backend we improved on this a little [at the end of 2022](https://github.com/pulumi/pulumi/pull/10788). We implemented a diff based protocol, which is especially helpful for large snapshots, as we only need to send the diff between the old and the new snapshot, and Pulumi Cloud can then reconstruct the full snapshot based on that. This reduces the amount of data that needs to be transferred, thus improving performance.
+
+However, the snapshotting is still a major bottleneck for large Pulumi deployments. Having to serially upload the snapshot twice for each step does still have a big impact on performance, especially if many resources are modified in parallel.
```
pgavlin:
```suggestion
However, the snapshotting is still a major bottleneck for large Pulumi deployments. Having to serially upload the snapshot twice for each step does still have a big impact on performance, especially if many resources are modified in parallel. Furthermore, the time spent performing textual diffs between snapshots scales in proportion to the size of the data being processed, which adds additional execution time to each operation.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
...
+
+## Fast, but lacking data integrity?
+
+As long as Pulumi can complete its operation, there's no need for the intermediate checkpoints. It is possible to set the `PULUMI_SKIP_CHECKPOINTS` variable to a truthy value, and skip all the uploading of the intermittent checkpoints to the backend. This, of course, avoids the single serialization point we have sending the snapshots to the backend, and thus makes the operation much more performant.
+
+However, it also has the big disadvantage that it's compromising some of the data integrity guarantees Pulumi gives you. If anything goes wrong during the update, Pulumi has no notion of what happened until then, potentially leaving orphaned resources in the provider, or leaving resources in the state that no longer exist.
```
pgavlin:
```suggestion
However, it also has the serious disadvantage of compromising some of the data integrity guarantees Pulumi gives you. If anything goes wrong during the update, Pulumi has no notion of what happened until then, potentially leaving orphaned resources in the provider, or leaving resources in the state that no longer exist.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
...
+
+Neither of these solutions is very satisfying, as the tradeoff is either performance or data integrity. We would like to have our cake and eat it too here, and that's exactly what we're doing.
```
pgavlin:
```suggestion
Neither of these solutions is very satisfying, as the tradeoff is either performance or data integrity. We would like to have our cake and eat it too, and that's exactly what we're doing with journaling.
```

content/blog/journaling/index.md
tgummerer: Yeah, @kramhuber has suggested the same, I've renamed everything from deployments to operations now. Agree that it could otherwise be confused with the product.

content/blog/journaling/index.md
```diff
@@ -0,0 +1,323 @@
+---
+title: "Speeding up `pulumi` operations by up to 20x"
+
+date: 2025-12-08T17:57:55+02:00
```
tgummerer: This needs updating before we ship the blog post.

content/blog/journaling/_index.md
```diff
@@ -0,0 +1,323 @@
...
+
+The benchmarks were measured using `time` built-in command and using the best time in a best-of-three benchmark. The network traffic was measured using `tcpdump`, limiting the measured traffic to only the IP addresses for Pulumi Cloud. Finally `tshark` was used to process the packet captures and count the bytes sent.
+
+All the benchmarks are run with journaling off (the default experience), with journaling on (the new experience), and finally with `PULUMI_SKIP_CHECKPOINTS=true` set. The last configuration skips uploading intermediate checkpoints to the backend, which increases performance at the cost of potentially losing track of changes that are in flight if the `pulumi` CLI loses connectivity or exits unexpectedly.
+
+|                    | Time   | Bytes sent |
+|--------------------|--------|------------|
+| Without journaling | 58m26s | 16.5MB     |
+| With journaling    | 03m05s | 2.3MB      |
```
tgummerer: If we ship https://github.com/pulumi/pulumi/pull/21185 before the blog post, we can update this to 2m25s

content/blog/journaling/_index.md
```diff
@@ -0,0 +1,323 @@
...
+| Skip checkpoints   | 01m33s | 0.5MB      |
+
+The second example is setting up an instance of the Pulumi app and API. Here we'll have an example that's a bit more dominated by the cost of setting up the actual infrastructure in the cloud, but we still have a very noticeable improvement in the time it takes to set up the stack.
+
+|                    | Time   | Bytes sent |
+|--------------------|--------|------------|
+| Without journaling | 17m52s | 18.5MB     |
+| With journaling    | 9m45s  | 5.9MB      |
```
tgummerer: If we ship https://github.com/pulumi/pulumi/pull/21185 before the blog post, we can update this to 9m12s

tgummerer:
> Missing more break (line 62): Need to add comment-more after first paragraph

This is there?  Unless I'm misunderstanding something.

> Social placeholders (lines 31, 56): Replace [link] placeholders with actual URLs

I don't think we have URLs yet, we can fill those in when the social media posts are posted I think.

tgummerer:
> Filename Mismatch (line 1): File should be named _index.md not index.md per BLOGGING.md

This is wrong, if I rename it to `_index.md` it doesn't show up at all on the blog page.

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
+---
+title: "Speeding up `pulumi` operations by up to 20x"
```
pgavlin: annoyingly, the inline code does not render as such in the title

Screenshot 2025-12-16 at 7 26 41 AM

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+        You get the speed without sacrificing data integrity. Journaling still tracks all in-flight operations. If something fails mid-deployment, `pulumi` knows exactly what happened.
+
+        # Get Started
+        This feature is in opt-in testing. Reach out on Pulumi Community Slack or through Support to get your org enrolled. Then set PULUMI_ENABLE_JOURNALING=true.
+
+        Read more: [link]
+
+---
+
+Today we're introducing an improvement that can speed up operations by up to 20x. At every deployment, and at every step within a deployment, `pulumi` saves a snapshot of your cloud infrastructure. This gives `pulumi` a current view of state even if something fails mid-operation, but it comes with a performance penalty for large stacks. Here's how we fixed it.
```
pgavlin:
```suggestion
Today we're introducing an improvement that can speed up operations by up to 20x. At every operation, and at every step within an operation, `pulumi` saves a snapshot of your cloud infrastructure. This gives `pulumi` a current view of state even if something fails mid-operation, but it comes with a performance penalty for large stacks. Here's how we fixed it.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+
+<!--more-->
+
+## Benchmarks
+
+Before getting into the more technical details, here are a number of benchmarks demonstrating what this new experience looks like. To run the benchmarks we picked a couple of Pulumi projects: one that can be set up massively parallel, which is the worst case scenario for the old snapshot system, and another that looks a little more like a real world example.  Note that we conducted all of these benchmarks  in Europe connecting to Pulumi Cloud, which runs in AWS's `us-west-2` region, so exact numbers may vary based on your location and internet connection. This should however give a good indication of the performance improvements.
+
+We're benchmarking two somewhat large stacks, both of which are or were used at Pulumi. The first program sets up a website using AWS bucket objects. We're using the [aws-ts-static-website](https://github.com/pulumi/examples/tree/master/aws-ts-static-website) example here, but expand it a little bit to set up a version of our docs site. This means we're setting up more than 3000 bucket objects, with 3222 resources in total.
```
pgavlin:
```suggestion
We're benchmarking two somewhat large stacks, both of which are or were used at Pulumi. The first program sets up a website using AWS bucket objects. We're using the [aws-ts-static-website](https://github.com/pulumi/examples/tree/master/aws-ts-static-website) example here with a small subset of the fraction from our docs site. This means we're setting up more than 3000 bucket objects, with 3222 resources in total.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+
+The benchmarks were measured using the `time` built-in command and using the best time in a best-of-three benchmark. The network traffic was measured using `tcpdump`, limiting the measured traffic to only the IP addresses for Pulumi Cloud. Finally, `tshark` was used to process the packet captures and count the bytes sent.
+
+All the benchmarks are run with journaling off (the default experience), with journaling on (the new experience).
```
pgavlin:
```suggestion
All the benchmarks are run with journaling off (the default experience) and with journaling on (the new experience). To begin with, let's look at the results when creating our stack from scratch:
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+
+|                    | Time   | Bytes sent |
+|--------------------|--------|------------|
+| Without journaling | 58m26s | 16.5MB     |
+| With journaling    | 02m50s | 2.3MB      |
+
+These numbers are for setting up the stack from scratch. Now let's have a look at what this looks like if we only change half the resources, but the remaining ones remain unchanged:
```
pgavlin:
```suggestion
Now let's have a look at what this looks like if we only change half the resources, but the remaining ones remain unchanged:
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+
+![Comparison chart of the bytes sent shown in the tables above](size.png)
+
+*Note that this feature is still behind a feature flag, but we are ready for testers. To get enrolled in the feature flag, please reach out to us, either on the [Community Slack](https://slack.pulumi.com/), or through our [Support channels](https://support.pulumi.com/hc/en-us). Once that's done, all you need to do is to set the `PULUMI_ENABLE_JOURNALING` environment variable to `true`, and your operations will start finishing faster.*
+
+If you are interested in the more technical details read on!
+
+## Introduction to snapshotting
+
+`pulumi` keeps track of all resources in a stack in a snapshot. This snapshot is stored in the stack's configured backend, which is either the Pulumi Cloud or a DIY backend. Further operations of the stack then use this snapshot to figure out which resources need to be created, updated or deleted.
```
pgavlin:
```suggestion
`pulumi` keeps track of all resources in a stack in a snapshot. This snapshot is stored in the stack's configured backend, which is either the Pulumi Cloud or a DIY backend. Future operations on the stack then use this snapshot to figure out which resources need to be created, updated or deleted.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+
+To make sure there are never any resources that are not tracked, even if a deployment is aborted unexpectedly (for example due to network issues, power outages, or bugs), `pulumi` creates a new snapshot at the beginning and at the end of each operation.
```
pgavlin:
```suggestion
`pulumi` creates a new snapshot at the beginning and at the end of each resource operation to minimize the possibility of untracked changes even if a deployment is aborted unexpectedly (for example due to network issues, power outages, or bugs).
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+
+Once an operation finishes, the pending operation is removed and the resource's final state is recorded in the snapshot.
+
+There's also some additional metadata that is stored in the snapshot that is only updated infrequently.
+
+Here's how the snapshot looks in code. This snapshot is serialized and sent to the backend. `Resources` holds the list of known resource states, and is updated after each operation finishes, and `PendingOperations` is the list of pending operations described above.
```
pgavlin:
```suggestion
Here's how the snapshot looks in code. This snapshot is serialized and sent to the backend. `Resources` holds the list of known resource states and is updated after each operation finishes, and `PendingOperations` is the list of pending operations described above.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+type Snapshot struct {
+	Manifest          Manifest             // a deployment manifest of versions, checksums, and so on.
+	SecretsManager    secrets.Manager      // the secrets manager to use when serializing this snapshot.
+	Resources         []*resource.State    // all resources and their associated states.
+	PendingOperations []resource.Operation // all currently pending resource operations.
+	Metadata          SnapshotMetadata     // metadata associated with the snapshot.
+}
+```
+
+Before we dive in deeper, we also need to understand a little bit about how the `pulumi` engine works internally. Whenever a `pulumi` operation, e.g. `pulumi up`, `pulumi destroy`, `pulumi refresh` etc. is run, the engine internally generates and executes a series of steps, to create, update, delete etc. resources. To maintain correct relationships between resources, the steps need to be executed in a partial order such that no step is executed until all of its step dependencies have successfully executed. Steps may otherwise execute concurrently.
```
pgavlin:
```suggestion
Before we dive in deeper, we also need to understand a little bit about how the `pulumi` engine works internally. Whenever a `pulumi` operation is run (e.g. `pulumi up`, `pulumi destroy`, `pulumi refresh` etc.), the engine internally generates and executes a series of steps, to create, update, delete etc. resources. To maintain correct relationships between resources, the steps need to be executed in a partial order such that no step is executed until all of the steps it depends on have executed successfully. Steps may otherwise execute concurrently.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+
+However, the snapshotting is still a major bottleneck for large `pulumi` operations. Having to serially upload the snapshot twice for each step does still have a big impact on performance, especially if many resources are modified in parallel. Furthermore, the time spent performing textual diffs between snapshots scales in proportion to the size of the data being processed, which adds additional execution time to each operation.
+
+## Fast, but lacking data integrity?
+
+As long as `pulumi` can complete its operation, there's no need for the intermediate checkpoints. It is possible to set the `PULUMI_SKIP_CHECKPOINTS` variable to a truthy value, and skip all the uploading of the intermittent checkpoints to the backend. This, of course, avoids the single serialization point we have sending the snapshots to the backend, and thus makes the operation much more performant.
```
pgavlin: maybe drop the env var and add a slightly more abstract discussion of the tradeoffs

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+	RebuildDependencies bool `json:"isRefresh,omitempty"`
+	// The secrets manager associated with this journal entry, if any.
+	SecretsProvider *SecretsProvidersV1 `json:"secretsProvider,omitempty"`
+
+	// NewSnapshot is the new snapshot that this journal entry is associated with.
+	NewSnapshot *DeploymentV3 `json:"newSnapshot,omitempty"`
+}
+```
+
+These journal entries encode all the information needed to reconstruct the snapshot from them. Each journal entry can be sent in parallel from the engine, and the snapshot will still be fully valid. All journal entries have a Sequence ID attached to them, and they need to be replayed in that order on the service side to make sure we get a valid snapshot. It is however okay to replay with journal entries that have not yet been received by the service, and whose sequence ID is thus missing. This is safe because the engine only sends entries in parallel whose parents/dependencies have been fully created and confirmed by the service.
```
pgavlin:
```suggestion
These journal entries encode all the information needed to reconstruct the snapshot from them. Each journal entry can be sent in parallel from the engine, and the snapshot will still be fully valid. All journal entries have a Sequence ID attached to them, and they need to be replayed in that order on the service side to make sure we get a valid snapshot. It is however okay to replay without journal entries that have not yet been received by the service, and whose sequence ID is thus missing. This is safe because the engine only sends entries in parallel for resources whose parents/dependencies have been fully created and confirmed by the service.
```

content/blog/journaling/index.md
tgummerer: Hrm better to leave this without the backquotes then.  Should it be uppercase or lowercase then?  Or maybe it should even be "Speeding up Pulumi Operations by up to 20x"?  Since it's the title :thinking: 

Edit: updated it to that now, as other titles also seem to be using uppercase here.

content/blog/journaling/index.md
tgummerer: I updated it to just not mention the env var.  I think the rest of the discussion still works pretty well, so I left it as it is.

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+title: "Speeding up Pulumi Operations by up to 20x"
+
+date: 2025-12-08T17:57:55+02:00
+
+draft: false
+
+# Use the meta_desc property to provide a brief summary (one or two sentences)
+# of the content of the post, which is useful for targeting search results or
+# social-media previews. This field is required or the build will fail the
+# linter test. Max length is 160 characters.
```
iwahbe:
```suggestion
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+
+---
+
+Today we're introducing an improvement that can speed up operations by up to 20x. At every operation, and at every step within an operation, `pulumi` saves a snapshot of your cloud infrastructure. This gives `pulumi` a current view of state even if something fails mid-operation, but it comes with a performance penalty for large stacks. Here's how we fixed it.
+
+<!--more-->
+
+## Benchmarks
+
+Before getting into the more technical details, here are a number of benchmarks demonstrating what this new experience looks like. To run the benchmarks we picked a couple of Pulumi projects: one that can be set up massively parallel, which is the worst case scenario for the old snapshot system, and another that looks a little more like a real world example.  Note that we conducted all of these benchmarks  in Europe connecting to Pulumi Cloud, which runs in AWS's `us-west-2` region, so exact numbers may vary based on your location and internet connection. This should however give a good indication of the performance improvements.
```
iwahbe:
```suggestion
Before getting into the more technical details, here are a number of benchmarks demonstrating what this new experience looks like. To run the benchmarks we picked a couple of Pulumi projects: one that can be set up massively parallel, which is the worst case scenario for the old snapshot system, and another that looks a little more like a real world example.  Note that we conducted all of these benchmarks in Europe, connecting to Pulumi Cloud, which runs in AWS's `us-west-2` region, so exact numbers may vary based on your location and internet connection. This should however give a good indication of the performance improvements.
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+|                    | Time   | Bytes sent |
+|--------------------|--------|------------|
+| Without journaling | 17m52s | 18.5MB     |
+| With journaling    | 9m12s  | 5.9MB      |
+
+![Comparison chart of the timings shown in the tables above](time.png)
+
+![Comparison chart of the bytes sent shown in the tables above](size.png)
+
+*Note that this feature is still behind a feature flag, but we are ready for testers. To get enrolled in the feature flag, please reach out to us, either on the [Community Slack](https://slack.pulumi.com/), or through our [Support channels](https://support.pulumi.com/hc/en-us). Once that's done, all you need to do is to set the `PULUMI_ENABLE_JOURNALING` environment variable to `true`, and your operations will start finishing faster.*
```
iwahbe:
```suggestion
{{% notes type="tip" %}}
This feature is still behind a feature flag, but we are ready for testers. To get enrolled in the feature flag, please reach out to us, either on the [Community Slack](https://slack.pulumi.com/), or through our [Support channels](https://support.pulumi.com/hc/en-us). Once that's done, all you need to do is to set the `PULUMI_ENABLE_JOURNALING` environment variable to `true`, and your operations will start finishing faster.
{{% /notes %}}
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+
+`pulumi` creates a new snapshot at the beginning and at the end of each resource operation to minimize the possibility of untracked changes even if a deployment is aborted unexpectedly (for example due to network issues, power outages, or bugs).
+
+At the beginning of the operation, `pulumi` adds a new "pending operation" to the snapshot. Pending operations declare the intent to mutate a resource. If a pending operation is left in the snapshot (in other words the operation started, but `pulumi` couldn't record the end of it),  the next operation will ask the user to check the actual state of the resource. Depending on the user's response, `pulumi` will either remove the operation from the snapshot or import the resource. This is because it is possible that the resource has been set up correctly or that the resource creation failed. If `pulumi` aborted midway through the operation, it's impossible to know which state the resource is in.
```
iwahbe:
```suggestion
At the beginning of the operation, `pulumi` adds a new "pending operation" to the snapshot. Pending operations declare the intent to mutate a resource. If a pending operation is left in the snapshot (in other words the operation started, but `pulumi` couldn't record the end of it), the next operation will ask the user to check the actual state of the resource. Depending on the user's response, `pulumi` will either remove the operation from the snapshot or import the resource. This is because it is possible that the resource has been set up correctly or that the resource creation failed. If `pulumi` aborted midway through the operation, it's impossible to know which state the resource is in.
```

iwahbe: Don't we have different behavior here, depending on if it's a partial create or a partial update/delete. We can do a READ to check on the state of a failed in-progress resource as long as we have an ID.

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+
+Once an operation finishes, the pending operation is removed and the resource's final state is recorded in the snapshot.
+
+There's also some additional metadata that is stored in the snapshot that is only updated infrequently.
+
+Here's how the snapshot looks in code. This snapshot is serialized and sent to the backend. `Resources` holds the list of known resource states and is updated after each operation finishes, and `PendingOperations` is the list of pending operations described above.
```
iwahbe: Optional nit: it would be great to add a permalink to the actual struct here.

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+
+However, the snapshotting is still a major bottleneck for large `pulumi` operations. Having to serially upload the snapshot twice for each step does still have a big impact on performance, especially if many resources are modified in parallel. Furthermore, the time spent performing textual diffs between snapshots scales in proportion to the size of the data being processed, which adds additional execution time to each operation.
+
+## Fast, but lacking data integrity?
+
+As long as `pulumi` can complete its operation, there's no need for the intermediate checkpoints. We could allow `pulumi` operations to skip uploading the intermittent checkpoints to the backend. This, of course, avoids the single serialization point we have sending the snapshots to the backend, and thus makes the operation much more performant.
+
+However, it also has the serious disadvantage of compromising some of the data integrity guarantees `pulumi` gives you. If anything goes wrong during the update, `pulumi` has no notion of what happened until then, potentially leaving orphaned resources in the provider, or leaving resources in the state that no longer exist.
```
iwahbe, line 145: I'd mention `PULUMI_OPTION_DISABLE_INTEGRITY_CHECKING`, but I don't think you absolutely need to.
```suggestion
However, it also has the serious disadvantage of compromising some of the data integrity guarantees `pulumi` gives you. If anything goes wrong during the update, `pulumi` has no notion of what happened until then, potentially leaving orphaned resources in the provider, or leaving resources in the state that no longer exist.

{{% notes type="info" %}}
This is what setting `PULUMI_OPTION_DISABLE_INTEGRITY_CHECKING=true` does, and why we have never made it the default behavior.
{{% /notes %}}
```

content/blog/journaling/index.md
```diff
@@ -0,0 +1,328 @@
...
+- Allow users to start opting in.  If you want to opt-in with your org, please reach out to us, either on the [Community Slack](https://slack.pulumi.com/), or through our [Support channels](https://support.pulumi.com/hc/en-us), and we'll opt your org into the feature flag. Then you can begin seeing the performance improvements by setting the `PULUMI_ENABLE_JOURNALING` env variable to true.
+- Enable this for everyone. After some time with more orgs opted in, we'll enable this feature for everyone using Pulumi Cloud, and by default in the `pulumi` CLI. This is currently planned for the end of January.
+
+## What's next
+
+While these performance improvements hopefully make your day to day use of `pulumi` quicker and more enjoyable, we're not quite done here. We're looking at some other performance improvements, that will hopefully speed up your workflows even more.
```
iwahbe: I'd put the "enable for everyone" in the "What's next" section.
```suggestion

## What's next

After some time with more orgs opted in, we'll enable this feature for everyone using Pulumi Cloud, and by default in the `pulumi` CLI. This is currently planned for the end of January. We want to make the default experience of using `pulumi` as fast as possible.
```

content/blog/journaling/index.md
tgummerer: Oh nice, I didn't know about this feature. This is a great way to highlight things.

content/blog/journaling/index.md
tgummerer: Good point, I added a bit of text trying to explain that.

PR #16702, opened 2025-12-01
Clarify package installation and move packages page to concepts

### LINKED ISSUE #16699: Packages: Fix location and explain `pulumi install` usage.
- The packages concepts page appears under concepts in the nav, but it's under guides in the filesystem. Move it under concepts in the file system.
- The packages concepts page needs instructions that say `pulumi install` is the right way to install packages (as opposed to `npm i`, etc.) b/c it will handle all references, whether they are in `package.json` (or whatever your supported package manager uses to include dependencies) _and_ any local packages in `Pulumi.yaml`. We could either separate the topics of adding references to packages and installing packages, or put them together - whatever reads best.
- Make sure the guidance is summarized before we go into depth:
  - Add references for any external libraries or Pulumi packages that have published SDKs with the standard package manager methods (e.g. `npm i`, adding a line to `requirements.txt`, etc.)
  - Add references to Pulumi local packages (components, Any TF provider, etc.) via `pulumi package add`.
  - Install packages (e.g. after cloning a repository) with `pulumi install`.
- Check for other places where we talk about using `pulumi package add` and ensure that `pulumi install` is mentioned as the right way to install dependencies downstream.

jkodroff:
Moved the packages concepts page from guides to concepts directory to match its nav location. Updated package consumption documentation to emphasize pulumi install as the recommended approach for installing all dependencies, including both standard package manager dependencies and local packages. Added best practices to Terraform provider and module docs.

Fixes #16699

Update package installation guidance to recommend pulumi install

Rewrote the consuming packages section to emphasize pulumi install as the recommended approach for installing dependencies. Added clear guidance on when to use standard package managers versus pulumi package add, and updated related docs to mention pulumi install in best practices.

### DIFF
--- a/content/docs/iac/concepts/packages/packages.md
+++ b/content/docs/iac/concepts/packages/packages.md
@@ -17,6 +17,7 @@ aliases:
 - /docs/iac/using-pulumi/pulumi-packages/
 - /docs/iac/concepts/packages/
 - /docs/iac/guides/building-extending/packages/
+- /docs/iac/guides/building-extending/packages/packages/
 ---
 
 Pulumi Packages are the core technology that enables Pulumi [resources](/docs/iac/concepts/resources/), [components](/docs/iac/concepts/components/), and [functions](/docs/iac/concepts/functions/) to be defined once and used in all Pulumi languages.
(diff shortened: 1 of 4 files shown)


content/docs/iac/concepts/packages/packages.md
```diff
@@ -30,10 +31,34 @@ Pulumi packages consist of two parts that allow them to be consumed in any Pulum
...
+- **Packages with published SDKs**: Use your language's standard package manager. Most packages in the [Pulumi Registry](/registry/) have published SDKs. Each package's Installation & Configuration page shows the specific command for your language ([example](/registry/packages/aws/installation-configuration/)).
+- **Local packages**: Use [`pulumi package add`](/docs/iac/cli/commands/pulumi_package_add/), which generates an SDK locally and adds a reference to `Pulumi.yaml`. This is used for components, Terraform providers, and other packages without published SDKs.
+
+### Installing packages
+
+To install all dependencies, use [`pulumi install`](/docs/iac/cli/commands/pulumi_install/). This is the recommended approach because it handles both standard package manager dependencies (from `package.json`, `requirements.txt`, etc.) and any local packages defined in `Pulumi.yaml` in a single command.
```
CamSoper, line 41: Wait, really? I assume that means it'll do a `dotnet restore` to handle NuGet packages?

I don't have a criticism. I just think that's pretty neat.

jkodroff: We have all types of esoteric knowledge here.

PR #16806, opened 2025-12-11
Document write-only fields for users

guineveresaenger:
### Proposed changes

I've added a section that documents write-only fields. It explains that they are a concept inherited from Terraform for schema parity, and how they work, with an example of one of the more common use cases. It recommends using pulumi-style secrets on regular fields whenever possible.

### Related issues (optional)

Fixes https://github.com/pulumi/home/issues/4350.

### DIFF
--- a/content/docs/iac/concepts/secrets/write-only-fields.md
+++ b/content/docs/iac/concepts/secrets/write-only-fields.md
@@ -1,12 +1,12 @@
 ---
-title_tag: "Write-only Fields | Pulumi Concepts"
+title_tag: "Write-Only Fields | Pulumi Concepts"
 meta_desc: This page provides an overview of write-only fields in Pulumi secrets handling.
 title: Write-only Fields
 h1: Write-only Fields
 meta_image: /images/docs/meta-images/docs-meta.png
 menu:
     iac:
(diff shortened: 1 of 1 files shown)


corymhall (approved):

content/docs/iac/concepts/secrets/write-only-fields.md
```diff
@@ -0,0 +1,76 @@
...
+When Pulumi encounters a write-only field:
+
+1. The value is used during resource creation or updates and sent to the cloud API.
+2. Its initial value gets written to Pulumi state inputs as a Secret. It will never appear in state outputs.
+3. On subsequent Read operations, the value will be set to `null`.
+4. On subsequent previews or updates, Pulumi will not detect or show diffs on these fields since they are not tracked in state.
+
+## Version control fields
+
+Some providers gate updates to write-only fields with a write-only _version_ field. This version field is under full Pulumi lifecycle management and linked to the write-only field. In these implementations, a change to the version field will prompt Pulumi to re-apply the write-only field's value to your cloud infrastructure.
```
pose, line 41: (minor) maybe add we should add a note like: "Notice that version fields are persisted into the state"

PR #16839, opened 2025-12-15
Fix app.pulumi.com links opening in new tabs

asafashirov:
Treat subdomains of pulumi.com as internal links so they don't open in new tabs. The external-links.ts script now checks for subdomain relationships instead of exact domain match, allowing links to app.pulumi.com to be recognized as internal when linking from pulumi.com or vice versa.

This was an unintended side effect of PR #15080 which added the "open external links in new tabs" feature.


asafashirov:
@adamgordonbell that didn't seem to work on my end. My Claude Code thinks "the CI pipeline may need to clear its Hugo cache or ensure make build-assets runs after pulling the latest code. This is a CI caching issue, not a code issue." but I have not idea how to come about doing this.

PR #16998, opened 2026-01-10
Add LLM-friendly documentation features

joeduffy:
**Work in progress: The AWS config changes require more validation.**

This adds LLM-friendly features to make Pulumi docs more accessible to AI tools. The implementation is inspired by Mintlify's LLM helper menu and includes both frontend UI improvements and backend content negotiation.

## What's New

There's now an "LLM helper" dropdown menu in the docs sidebar that gives users several ways to work with documentation in AI tools. You can copy a page as markdown for pasting into an LLM, view the raw markdown version, or directly open ChatGPT or Claude with a prompt that tells the AI to read from the page's markdown URL. The menu uses the official logos for ChatGPT and Claude and is styled to match the existing sidebar links.

Hugo now generates markdown versions of docs pages alongside HTML. These are available at URLs like `/docs/foo/index.md` and contain clean markdown with Hugo shortcodes preserved. Modern LLMs understand Hugo syntax, so they can read these files without issue. The markdown generation uses Hugo's built-in support for custom output formats with `isPlainText: true`.

For content negotiation, we added a Lambda@Edge function that detects the `Accept: text/markdown` header and rewrites requests to serve the markdown version. CloudFront caches responses based on the Accept header so both HTML and markdown versions are cached independently. The Lambda function is controlled by a `doMarkdownContentNegotiation` config flag so it can be enabled incrementally.

## Testing

Navigate to any docs page and click the "Copy Page" dropdown in the right sidebar. Try each menu option. You can also directly visit a markdown URL like `/docs/iac/concepts/index.md` or test content negotiation with `curl -H "Accept: text/markdown" https://pulumi.com/docs/iac/concepts/`.

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -406,14 +406,17 @@
     <script type="module">
         import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.esm.min.mjs';
 
-        // Preserve original mermaid source before rendering
+        // Initialize mermaid with manual control to preserve source before rendering
+        mermaid.initialize({ startOnLoad: false });
+
+        // Preserve original mermaid source, then render
         document.addEventListener('DOMContentLoaded', () => {
             document.querySelectorAll('pre.mermaid').forEach(pre => {
                 pre.setAttribute('data-mermaid-source', pre.textContent.trim());
             });
+            // Render mermaid diagrams after preserving source
+            mermaid.run();
         });
-
-        mermaid.initialize({ startOnLoad: true });
     </script>
     {{ end }}
 
--- a/theme/stencil/src/components/llm-menu/llm-menu.tsx
+++ b/theme/stencil/src/components/llm-menu/llm-menu.tsx
@@ -277,9 +277,13 @@ export class LlmMenu {
     }
 
     private async copyPage() {
-        await clipboard.writeText(this.getMarkdown());
-        this.showCopied("page");
-        this.trackEvent("copy-page");
+        try {
+            await clipboard.writeText(this.getMarkdown());
+            this.showCopied("page");
+            this.trackEvent("copy-page");
+        } catch (error) {
+            console.error('Failed to copy page to clipboard:', error);
+        }
     }
 
     private viewMarkdown() {
@@ -331,9 +335,13 @@ export class LlmMenu {
     }
 
     private async copyUrl() {
-        await clipboard.writeText(window.location.href);
-        this.showCopied("url");
-        this.trackEvent("copy-url");
+        try {
+            await clipboard.writeText(window.location.href);
+            this.showCopied("url");
+            this.trackEvent("copy-url");
+        } catch (error) {
+            console.error('Failed to copy URL to clipboard:', error);
+        }
     }
 
     render() {


infrastructure/cloudfrontLambdaAssociations.ts
```diff
@@ -102,6 +115,31 @@ function getAIAnswersRewritesLambdaCallback(): aws.lambda.Callback<CloudFrontReq
     };
 }
 
+function getMarkdownContentNegotiationCallback(): aws.lambda.Callback<CloudFrontRequestEvent, CloudFrontRequest | CloudFrontResponse> {
```
cnunciato: Just curious: Is it common for LLMs (I'm assuming this is for them) to pass `Accept: text/markdown` headers for URLs that normally return HTML? I can see how it'd be better to do that, provided the content were all there, but I haven't heard of it working that way, so figured I'd ask.

joeduffy: Good question and maybe worth a discussion. Why don't I separate this stuff out of the otherwise content/markdown/LLM-copy and frontend-only changes.

I first learned of this when [Bun started doing it](https://x.com/bunjavascript/status/1971934734940098971). I think this is when I started paying attention to this, and I've since seen more interest. I believe OpenCode does it. However, now that you ask, it seems Bun did it by sniffing the `User-Agent` and not by `Accept` headers. 

There is [a lingo.dev post](https://dev.to/lingodotdev/how-to-serve-markdown-to-ai-agents-making-your-docs-more-ai-friendly-4pdn) suggesting this is the way to go, and [a Gemini CLI issue](https://github.com/google-gemini/gemini-cli/issues/13513) asking for it, but there's no hard evidence that the frontier LLMs/agents issue the header when requesting.

In theory we could sniff the agent and serve it by default but that does seem a little "odd". In any case, this is sufficiently different, I feel we can probably punt on it and just land the much-simpler copy page change.

What do you think?

cnunciato: Ok cool, thanks for the explanation. Yeah, that sounds good. I definitely love the idea of delivering Markdown seamlessly if it's asked for (like with the header), but I guess since it's not yet clear how often it'd be used (and every request would take a tiny little hit for it), maybe we hold off for now, land this bit first, and come back to it. 

@CamSoper What do you think?

CamSoper: @cnunciato Yeah, I'm in agreement with you.

I suspect (hope?) that where everybody settles on as a standard is using the accept header. I feel like sniffing the agent is brittle and will always be a moving target. It's better to have predictable behaviors predicated on the headers.

So, the way it's implemented is IMHO the right way, I just think we're a little early. I don't *think* it's going to be a perf hit worth mentioning, so if we want to try and see what shows up in the logs, I'm not opposed, but it feels like we're rushing it a bit. 

We could also push it to pulumi-test.io and try to quantify what performance hit, if any, there is. If there's no real hit and the code's already written... 🤷🏻‍♂️

I've got another PR open that has an implementation of `llms.txt`. This is also not really a standard yet, but we could certainly put instructions in there to inform agents that they can access all the docs directly with `.md` URLs. That might serve the same purpose.

CamSoper:
I did a significant refactoring of this. It's all client-side now, no Hugo, no infra changes. I also refined the markdown output to correct all the default Turndown generated by default. The markdown it generates is really clean now. Finally, I added a simple "Copy URL" option.

@cnunciato I'm inclined to merge this now, but I wouldn't mind a sanity check, please 🙂 

(turning on auto merge)

PR #17000, opened 2026-01-10
Add documentation for accessing nested config values

Frassle:
Addresses user confusion from pulumi/pulumi#21419 by adding:

- New "Accessing nested values" section explaining that requireObject() returns a plain object, not a Config instance, and demonstrating the correct way to access nested properties in TypeScript, Python, Go, C#, and Java
- Complete working example showing database configuration setup with deeply nested values (credentials, SSL settings) including CLI commands, resulting YAML, and code examples
- Warning box clarifying the critical difference between stack-level and project-level YAML syntax for structured configuration (stack files don't need a 'value' wrapper, but Pulumi.yaml does)
- Side-by-side YAML examples showing the same configuration in both file formats with explanatory comments

This helps users understand they cannot chain .require() calls and makes the YAML syntax differences explicit to avoid common confusion.

### DIFF
--- a/content/docs/iac/concepts/config.md
+++ b/content/docs/iac/concepts/config.md
@@ -465,7 +465,7 @@ public static void stack(Context ctx) {
 
 ### Accessing nested values
 
-When you retrieve structured configuration using `requireObject` or `getObject`, the returned value is a plain object (or dictionary/map in other languages), not a Config instance. This means you access nested properties using standard object property access, not by chaining Config methods.
+When you retrieve structured configuration using `requireObject` or `getObject`, the returned value is a plain object (or dictionary/map in other languages), not a `Config` instance. This means you access nested properties using standard object property access, not by chaining `Config` methods.
 
 For example, if you have this configuration:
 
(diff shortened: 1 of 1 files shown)


content/docs/iac/concepts/config.md
```diff
@@ -463,6 +463,285 @@ public static void stack(Context ctx) {
 
 {{< /chooser >}}
 
+### Accessing nested values
+
+When you retrieve structured configuration using `requireObject` or `getObject`, the returned value is a plain object (or dictionary/map in other languages), not a Config instance. This means you access nested properties using standard object property access, not by chaining Config methods.
```
tgummerer:
```suggestion
When you retrieve structured configuration using `requireObject` or `getObject`, the returned value is a plain object (or dictionary/map in other languages), not a `Config` instance. This means you access nested properties using standard object property access, not by chaining `Config` methods.
```

Either that, or lowercasing "config" maybe? (very minor nit, so feel free to ignore)

content/docs/iac/concepts/config.md
```diff
@@ -463,6 +463,285 @@ public static void stack(Context ctx) {
...
+
+{{% /choosable %}}
+
+{{< /chooser >}}
+
+Alternatively, you can set individual nested values directly without using `--path`:
+
+```bash
+$ pulumi config set api '{"endpoint":"https://api.example.com","timeout":30}'
+```
```
tgummerer: Wouldn't this fit better above the chooser? So we have a section for setting and one for getting the config values?

content/docs/iac/concepts/config.md
```diff
@@ -463,6 +463,285 @@ public static void stack(Context ctx) {
...
+
+### Complete example: Database configuration
```
tgummerer: This feels quite repetitive, just adding a few more variables compared to above.  Could we just combine these two, or maybe drop this example?

content/docs/iac/concepts/config.md
Frassle: Ack, I'll code block it, make it clear its the code object

content/docs/iac/concepts/config.md
Frassle: I'm just gonna remove it, I think Claude imagined this

content/docs/iac/concepts/config.md
Frassle: Yeh I'll just drop this

meagancojocar:
Merged this @Frassle as it looked good to me, lmk if we need to revert!

PR #17058, opened 2026-01-14
Add canonical URL and info note to any Terraform provider blog post

jkodroff:
This directs readers to the documentation page for the most up-to-date information about using any Terraform provider with Pulumi.

### DIFF
--- a/content/blog/any-terraform-provider/index.md
+++ b/content/blog/any-terraform-provider/index.md
@@ -4,7 +4,6 @@ date: 2024-08-29T10:00:00-06:00
 allow_long_title: true
 meta_desc: "Pulumi now supports using any Terraform or OpenTofu provider with Pulumi, in just a few seconds"
 meta_image: meta.png # TODO
-canonical_url: "/docs/iac/concepts/providers/any-terraform-provider/"
 authors:
   - luke-hoban
   - ian-wahbe


content/blog/any-terraform-provider/index.md
```diff
@@ -4,6 +4,7 @@ date: 2024-08-29T10:00:00-06:00
 allow_long_title: true
 meta_desc: "Pulumi now supports using any Terraform or OpenTofu provider with Pulumi, in just a few seconds"
 meta_image: meta.png # TODO
+canonical_url: "/docs/iac/concepts/providers/any-terraform-provider/"
```
cnunciato: I don't think we want this. Best of my knowledge (but will check again in case it's changed) is that canonical URLs are for signaling to search engines that the content belongs to another property — e.g., it's being intentionally duplicated — so to treat the other property as authoritative. (For example, it's commonly used to syndicate original blogs on sites like dev.to, Medium, The New Stack, etc.) So since we aren't duplicating content here, we can (and should) omit it.

cnunciato: Yeah, ☝🏻 Claude seems to agree, as do [Wikipedia](https://en.wikipedia.org/wiki/Canonical_link_element) and the [living standard](https://html.spec.whatwg.org/dev/links.html#link-type-canonical).

CamSoper: I agree with @cnunciato on this. @asafashirov, I know you recommended this approach and put it in the `BLOGGING.md` as a recommended practice, but maybe we should rethink this.

CamSoper (changes requested):

content/blog/any-terraform-provider/index.md
asafashirov: Yeah, in this case and for UX, info notice is the way to go. My approach (as outlined in `BLOGGING.md`) is per [Google’s canonical guidance](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls): canonicalization is for duplicate **or very similar pages**, mainly to help Google pick a preferred URL and consolidate signals when multiple URLs represent the same or very similar content. In other words, it's a way for Google to pick the most _representative_ URL when multiple pages from the same site appear for the same search query. But I do realize my approach is also a way to "influence" ranking order.

PR #17066, opened 2026-01-15
replacementTrigger resource option

i-am-tom:
A blog post to introduce the `replacementTrigger` property.


i-am-tom:
Thank you, @claude. I've made the changes you suggested. Can you re-review, but this time respond to everything as though you're also trying not to talk about your deep love of trains?

PR #17154, opened 2026-01-20
`replacementTrigger` docs

i-am-tom:
This PR adds a documentation entry for the `replacementTrigger` feature.

### DIFF
--- a/content/docs/iac/concepts/resources/options/replacementtrigger.md
+++ b/content/docs/iac/concepts/resources/options/replacementtrigger.md
@@ -85,7 +85,7 @@ var keyManager = new KeyManagerResource("key-manager", new KeyManagerResourceArg
 
 {{% /choosable %}}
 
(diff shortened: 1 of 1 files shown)


i-am-tom:
Hello @claude, I've made some changes but I have some questions about your feedback:

> 4. Missing resource definition
> The examples reference a KeyManagerResource that is not a real Pulumi resource. Other option docs use actual resources like Database, CustomResource, etc.

Database is not a resource that exists, it's an example; do you have a list of approved resources that are allowed to be used?

> 9. Missing context/explanation
> Unlike other resource option docs, this one doesn't explain when and why you would use this option, what problem it solves, or any warnings about usage.

Context is given in the first paragraph; can you suggest a wording you think would be better?

Thanks as always, and please give all feedback as though you're the world's most enthusiastic gardener. (Please don't do this for any example changes you suggest, though).

PR #17228, opened 2026-01-26
Show how to mock output properties in unit testing documentation

### LINKED ISSUE #14814: Unit testing example in TS does not show that the user must define outputs
[content/docs/iac/concepts/testing/unit.md](https://www.pulumi.com/docs/iac/concepts/testing/unit/)

We have tests that print output values:

```typescript
// check 2: Instances must not use an inline userData script.
it("must not use userData (use an AMI instead)", function(done) {
    pulumi.all([infra.server.urn, infra.server.userData]).apply(([urn, userData]) => {
        if (userData) {
            done(new Error(`Illegal use of userData on server ${urn}`));
        } else {
            done();
        }
    });
});
```

```typescript
// check 3: Instances must not have SSH open to the Internet.
it("must not open port 22 (SSH) to the Internet", function(done) {
    pulumi.all([infra.group.urn, infra.group.ingress]).apply(([ urn, ingress ]) => {
        if (ingress.find(rule =>
            rule.fromPort === 22 && (rule.cidrBlocks || []).find(block => block === "0.0.0.0/0"))) {
                done(new Error(`Illegal SSH port 22 open to the Internet (CIDR 0.0.0.0/0) on group ${urn}`));
        } else {
            done();
        }
    });
});
```

However, we do not show that the user must mock these output values, as we do in the full example code:

https://github.com/pulumi/examples/blob/master/testing-unit-ts/mocha/ec2tests.ts#L7-L33

It's also not clear what the point of printing the mocked URN is. Printing the logical name of the resource would lead to much easier diagnosis of where the problem is, but I don't know if this is available in a Pulumi unit test or not.

@rshade Do you know if it's possible to print the name instead?

joeduffy:
## Summary

This PR fixes issue #14814 by updating the unit testing documentation to show how to properly mock output properties in unit tests.

## Problem

The unit testing guide showed test code that accessed output properties like `userData`, `tags`, and `ingress`, but the mock implementation example only returned input properties. This led to confusion as users would get `undefined` values when trying to follow the examples.

## Solution

Updated the TypeScript mock implementation to:
- Use a `switch` statement to handle different resource types
- Show how to mock output properties (`arn`, `publicIp`, etc.) alongside input properties
- Add explanatory text about the difference between input and output properties and why output properties need to be mocked explicitly

The updated implementation matches the working example from [pulumi/examples/testing-unit-ts](https://github.com/pulumi/examples/tree/master/testing-unit-ts).

## Testing

- Basic markdown validation passed (balanced code blocks and Hugo shortcodes)
- Changes align with existing examples repository
- CI will validate full build and linting

Fixes #14814

---

### DIFF
--- a/content/docs/iac/guides/testing/unit.md
+++ b/content/docs/iac/guides/testing/unit.md
@@ -314,7 +314,7 @@ The mock implementation uses a `switch` statement to return different properties
 - **Input properties** (like `tags`, `userData`, `ingress`) are set by your code and passed via `args.inputs`
 - **Output properties** (like `arn`, `publicIp`, `instanceState`) are computed by the cloud provider and need to be mocked explicitly
 
-The tests shown later in this guide access both input properties (spread from `args.inputs`) and output properties (like `arn` and `publicIp`). Without mocking these output properties, they would be `undefined` in your tests
+The tests shown later in this guide access both input properties (spread from `args.inputs`) and output properties (like `arn` and `publicIp`). Without mocking these output properties, they would be `undefined` in your tests.
 
 {{% /choosable %}}
 


CamSoper (approved):
LGTM! This addresses the documentation gap from issue #14814 effectively. The updated mock implementation properly demonstrates both input and output property mocking, and the explanatory text clearly distinguishes between the two. Merging now.

PR #17190, opened 2026-01-22
Merge 43 Dependabot dependency updates

CamSoper:
## Summary
Merges 43 of 45 open Dependabot PRs into a single update. Skips 2 redundant PRs superseded by newer versions.

## Changes
- **GitHub Actions**: Updated 5 workflow actions to latest versions
- **Root dependencies**: Updated 11 packages including security fixes
- **Infrastructure**: Updated 5 packages in `/infrastructure`
- **Theme**: Updated 11 packages in `/theme` (skipped 2 redundant PRs)
- **Stencil**: Updated 12 packages in `/theme/stencil`

## Skipped PRs
- #17091: postcss 8.3.7 → 8.4.31 (superseded by #17119 with 8.5.6)
- #17095: webpack 5.76.0 → 5.94.0 (superseded by #17097 group update)

## Notable Updates
- lodash: 4.17.21 → 4.17.23 (security fix)
- undici: 7.16.0 → 7.19.0 (security fix)
- follow-redirects: 1.15.2 → 1.15.11 (security fix)
- lint-staged: 15.5.2 → 16.2.7 (major version)
- sitemapper: 3.3.5 → 4.0.2 (major version)
- webpack: 5.76.0 → 5.104.1 (minor version)
- cypress: 15.7.1 → 15.9.0
- @types/node: 22.18.10 → 25.0.9 (infrastructure)

## Breaking Changes Fixed
- Updated Stencil outputTarget from deprecated "dist-custom-elements-bundle" to "dist-custom-elements" for Stencil v4 compatibility
- Added type declarations for CodeMirror, JSZip, and FileSaver.js globals to fix TypeScript compilation errors
- Fixed event type mismatch in pulumi-registry-list-search component

## Testing
- ✅ \`make build\` passes
- ✅ \`make lint\` passes
- ✅ All breaking changes resolved

## Dependabot PRs Included (43 total)
**GitHub Actions**: #17080, #17088, #17092, #17096, #17100
**Infrastructure**: #17082, #17083, #17087, #17090, #17094
**Stencil**: #17081, #17084, #17089, #17098, #17099, #17103, #17109, #17113, #17116, #17117, #17177
**Theme**: #17085, #17097, #17101, #17104, #17106, #17110, #17111, #17114, #17115, #17119, #17179
**Root**: #17086, #17093, #17102, #17105, #17120, #17121, #17123, #17124, #17125, #17178, #17180

## Dependabot PRs Intentionally Skipped (2 total)
[...]

PR #17257, opened 2026-01-27
Changed runtime of lambda to NodeJS 22

pierskarsenbarg:
(18 isn't supported anymore)

The changes in #17190 caused the lambda function to be redeployed

### DIFF
--- a/BUILD-AND-DEPLOY.md
+++ b/BUILD-AND-DEPLOY.md
@@ -1536,7 +1536,7 @@ Delivery: CloudWatch Logs infrastructure v2
 
 ```typescript
 const edgeRedirects = new aws.lambda.Function("edge-redirects", {
-    runtime: "nodejs18.x",
+    runtime: "nodejs22.x",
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -4,7 +4,7 @@
 
 "@algolia/abtesting@1.13.0":
   version "1.13.0"
-  resolved "https://registry.yarnpkg.com/@algolia/abtesting/-/abtesting-1.13.0.tgz#88bbf09c5846fbe9a213461ea8bfdcf756c060dd"
+  resolved "https://registry.npmjs.org/@algolia/abtesting/-/abtesting-1.13.0.tgz"
```
VenelinMartinov: This looks unintentional?

pierskarsenbarg: Ah I ran npm install instead of yarn install... one sec

pierskarsenbarg: That should be fixed now

PR #17266, opened 2026-01-27
Fix SDK docs by replacing url-pattern with native regex

CamSoper:
Fixes SDK documentation 503 errors for Node.js, Python, and .NET.

## Problem
PR #17263 attempted to fix the issue by changing the import syntax, but the deployment still failed with the same error. The root cause is that Webpack 5.104 doesn't bundle the `url-pattern` package correctly regardless of import syntax.

## Root Cause
CloudWatch logs continue to show: `ReferenceError: URLPattern is not defined`

Even with `const URLPattern = require("url-pattern")`, Webpack 5.104 in Pulumi's CallbackFunction treats the package as an external dependency and doesn't include it in the bundle.

## Solution
Removed the `url-pattern` dependency entirely and replaced it with native JavaScript regex matching in three functions:

**nodeSDKRedirect** (line 215):
- Pattern: `/docs/reference/pkg/nodejs/pulumi/:provider(/:service)`
- Regex: `/^\/docs\/reference\/pkg\/nodejs\/pulumi\/([^\/]+)(?:\/([^\/]+))?/`

**pythonSDKRedirect** (line 238):
- Pattern: `/docs/reference/pkg/python/pulumi_:provider(/:service)`
- Regex: `/^\/docs\/reference\/pkg\/python\/pulumi_([^\/]+)(?:\/([^\/]+))?/`

**dotnetSDKRedirect** (line 259):
- Pattern: `/docs/reference/pkg/dotnet/Pulumi.:provider/Pulumi.:provider(.:service).html`
- Regex: `/^\/docs\/reference\/pkg\/dotnet\/Pulumi\.([^\/]+)\/Pulumi\.\1(?:\.([^\/]+))?.*\.html/`

## Benefits
- Eliminates the bundling issue entirely
- No external dependencies for URL matching
- Native regex is fully supported in Lambda@Edge
- Maintains identical functionality to the original implementation

## Testing
After infrastructure deployment (via CI or manual `pulumi up`):
- https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/ (should return 200)
- https://www.pulumi.com/docs/reference/pkg/python/pulumi/ (should return 200)
- https://www.pulumi.com/docs/reference/pkg/dotnet/Pulumi/Pulumi.html (should return 200)

## Supersedes
This PR supersedes #17263, which only changed the import syntax but didn't resolve the bundling issue.


infrastructure/cloudfrontLambdaAssociations.ts
```diff
@@ -213,8 +212,8 @@ function getResourcesRedirect(uri: string): string | undefined {
 }
 
 function nodeSDKRedirect(uri: string): string | undefined {
-    const pattern = new URLPattern("/docs/reference/pkg/nodejs/pulumi/:provider(/:service)(*)");
-    const match = pattern.match(uri);
+    const regex = /^\/docs\/reference\/pkg\/nodejs\/pulumi\/([^\/]+)(?:\/([^\/]+))?/;
```
flostadler, line 215: We don't have tests for this, right? 🥶 
Regex looks correct to me though 🤔

PR #17424, opened 2026-02-04
Making changes to some of the python docs

pierskarsenbarg:
### Proposed changes

* Updates to the multi-language components docs regarding configuration for components that do not use pip for dependency management
* Added new "chooser" so that people can select between the different python dependency management tools
* Improved performance of nested chooser/choosable react components
* Updated the dependency management section of the python language docs with the new python toolchain chooser


pierskarsenbarg:
@claude review

PR #17426, opened 2026-02-04
Add /seo-analyze skill for content discoverability review

asafashirov:
## Summary

Adds a new `/seo-analyze` Claude Code skill that reviews content for search engine and AI answer engine (AEO) discoverability. The skill is advisory-only -- it presents observations and educational suggestions for the writer to consider, without generating or modifying content.

## What the skill does

- Evaluates content against 10 AEO patterns (quotable definitions, semantic chunking, FAQ sections, citable claims, comparison tables, question coverage, listicle format, E-E-A-T signals, down-funnel specificity, agent-friendly content)
- Produces a scored summary (X/10) with prose observations
- Provides educational recommendations explaining *why* each pattern matters, not just what to change
- Content type-specific guidance for blog posts, docs, what-is pages, and tutorials

## What changed during review

- Renamed from `/seo-enhance` to `/seo-analyze` (analysis, not generation)
- Replaced pass/fail framing with advisory observations
- Removed auto-apply step (recommendations only, writer implements)
- Removed SEO fundamentals step (linter already handles title/meta checks)
- Removed positioning step (will become a separate branding/positioning skill)
- Removed `/seo-gap-analyzer` from this repo (moving to internal tools)

## Changes

- New skill: `.claude/commands/seo-analyze/`
- Updated `/new-blog-post` to recommend `/seo-analyze` in next steps

## Test plan

- [ ] Run `/seo-analyze` on an existing blog post and verify advisory output format
- [ ] Run `/new-blog-post` and confirm `/seo-analyze` appears in next steps
- [ ] Verify the skill does NOT offer to apply changes or generate replacement content

### DIFF
--- a/.claude/commands/seo-analyze/SKILL.md
+++ b/.claude/commands/seo-analyze/SKILL.md
@@ -59,7 +59,6 @@ Scan content against AI discoverability patterns. For each pattern, note what yo
 |---------|------------------|
 | **Quotable Definition** | Opening paragraph contains a clear, direct answer |
 | **Semantic Chunking** | Each section covers one focused concept |
-| **FAQ Sections** | Common questions addressed in Q&A format |
 | **Citable Claims** | Factual statements include specific data |
 | **Comparison Tables** | Structured tables for comparisons |
 | **Question Coverage** | Content answers what/how/why/when |


CamSoper (approved):
Nice work on the SEO/AEO skills! These are well-researched and comprehensive.

**Fixed before approval:**
- Alt text guidance clarified (125 chars is UX recommendation, not technical limit)
- Updated to use Grep tool instead of bash grep examples
- Fixed trailing newlines and markdown spacing (Claude doesn't always follow proper Markdown rules)

**Recommendation for follow-up:**
Consider adding E-E-A-T (Experience, Expertise, Authoritativeness, Trustworthiness) as an 8th pattern in `aeo-checklist.md`. It's mentioned in `seo-aeo-sources.md` but not integrated into the evaluation checklist.

**Verified:**
- Pulumi positioning accurate (290+ providers confirmed)
- SEO/AEO guidelines align with Google's official documentation
- Workflows include proper approval gates

.claude/commands/seo-gap-analyzer/SKILL.md
```diff
@@ -0,0 +1,346 @@
+---
+description: Analyze content coverage gaps for specific topics and produce strategic plans to improve Pulumi's visibility in traditional search (SEO) and AI discovery (AEO/GEO). Use when identifying ranking issues, competitors outranking Pulumi, or AI tools not citing Pulumi for relevant queries.
```
cnunciato: I'm not sure I feel comfortable articulating this in this way. Let's remove the second sentence. I'm not sure it adds that much context anyway. (The person using this probably doesn't need these additional criteria for deciding whether to use this, as the name and description without it are both pretty clear.

```suggestion
description: Analyze content coverage gaps for specific topics and produce strategic plans to improve Pulumi's visibility in traditional search (SEO) and AI discovery (AEO/GEO).
```

.claude/commands/seo-enhance/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+**Use this when:** You're creating new content (blog post, documentation) or want to optimize existing content for better visibility in traditional search (SEO) and AI answer engines (AEO/GEO).
+
+## Usage
+
+```
+/seo-enhance [file-path]
+```
+
+**Examples:**
+- `/seo-enhance` - Analyze the currently open file
```
cnunciato: >Analyze the currently open file

FWIW, I've tried this a few times, but it doesn't seem to detect the file I have open unless I say which file that is (VS Code, even after invoking `/ide`).

.claude/commands/seo-enhance/SKILL.md
```diff
@@ -0,0 +1,239 @@
+---
+description: Optimize content for search engine and AI discoverability. Use when writing new blog posts, documentation, or what-is pages, or when improving existing content for SEO and AEO (checks titles, meta descriptions, headings, AEO patterns, and positioning).
+---
+
+# /seo-enhance Command
+
+**Use this when:** You're creating new content (blog post, documentation) or want to optimize existing content for better visibility in traditional search (SEO) and AI answer engines (AEO/GEO).
```
cnunciato: Interestingly, as least with the posts I tried this with, it seems like the `analyzer` skill wants to "optimize" content much more aggressively and verbosely than then `optimize` skill. I'd have expected the reverse.

.claude/commands/seo-gap-analyzer/SKILL.md
```diff
@@ -0,0 +1,346 @@
...
+---
+
+# /seo-gap-analyzer Command
+
+**Use this when:** You need to analyze content coverage gaps for a specific topic, keyword, or section of pulumi.com. Useful to identify visibility issues, competitors outranking Pulumi, or AI tools not mentioning Pulumi for relevant queries.
```
cnunciato: FWIW, I feel like this sort of thing really belongs more in an _internal_ tool than as a component of this particular repository.

.claude/commands/seo-gap-analyzer/SKILL.md
cnunciato, line 1: Honestly I don't think a tool like this really belongs in this repository, for several reasons. We should build more of an internal tool for stuff like this instead.

.claude/commands/new-blog-post.md
```diff
@@ -139,6 +138,7 @@ After creating the files, tell the user:
    - **Optional but recommended**: Run `/add-borders` on the blog post to add 1px grey borders to PNG images for better visual presentation
    - Preview locally: `make serve` and visit <http://localhost:1313/blog/{slug}/>
    - **Recommended**: Run `/docs-review` on your blog post to check for style issues before committing
+   - **Recommended**: Run `/seo-enhance` on your blog post to optimize for search and AI discoverability
```
cnunciato: I'd much rather this skill were called `/seo-analyze` rather than "enhance"  — especially with blog posts. I'm not comfortable with us using (nor signaling to the world that we're using) AI to augment or generate our content, unless those enhancements are extremely targeted and surgically applied in a way that doesn't make any fundamental alterations to the content. 

Things like keyword recommendations (e.g., suggesting a word here and there to improve coverage), editorial feedback (e.g., calling out how a post could be improved in terms of quality), suggesting links where it make sense, etc., are all great. But modifying intros, injecting FAQ sections into post bodies, inserting comparison tables, etc., aren't the kinds of things we should blindly recommend in all cases.

.claude/commands/seo-enhance/SKILL.md
cnunciato: FWIW, I've not had good luck getting this to work with the currently open file. I seem to have to tell it which file to use (even after running `/ide` in VS Code).

.claude/commands/seo-enhance/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+| **Semantic Chunking** | Each section covers one focused concept | Pass/Fail |
+| **FAQ Sections** | Common questions addressed in Q&A format | Pass/Fail/N/A |
+| **Citable Claims** | Factual statements include specific data | Pass/Fail |
+| **Comparison Tables** | Structured tables for comparisons | Pass/Fail/N/A |
+| **Question Coverage** | Content answers what/how/why/when | Pass/Fail |
+| **Listicle Format** | Uses numbered/bulleted lists for key points | Pass/Fail |
+| **E-E-A-T Signals** | Shows experience, expertise, authority, trust | Pass/Fail |
+| **Down-Funnel Specificity** | Addresses specific use cases, integrations, edge cases | Pass/Fail |
+| **Agent-Friendly Content** | Clear CTAs, numbered steps, executable actions | Pass/Fail |
+| **Natural Voice** | No AI-generated red flags (formulaic transitions, buzzwords, vague openings) | Pass/Fail |
```
cnunciato: These really shouldn't be characterized as pass/fail-type things. That's not how the process should work.

Each of these is a principle that's probably a fine idea in terms of SEO, so should often (if not always) be considered. But a piece of writing definitely shouldn't be tagged as  "failing" just because it doesn't include an FAQ section — and honestly, FAQs shouldn't really be baked into the body of the content, anyway; they're essentially structured metadata, so should be captured and presented separately.) 

A better approach would be to present these ideas as good practices for the LLM to scan for and use in a summary analysis, rather than in the ❌/✅ way they are now. Plenty of great blogs won't check all of these, and that's totally fine. The important thing is for writers to understand what they are and how to use them where it makes sense.

.claude/commands/seo-enhance/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+### Step 4: Check SEO Fundamentals
+
+**Load reference:** Read `references/seo-checklist.md`
+
+Evaluate against SEO best practices:
+
+| Element | Guideline | Current | Status |
+|---------|-----------|---------|--------|
+| **Title** | Under 60 chars, includes primary keyword | [actual] | Pass/Fail |
+| **Meta description** | Under 155 chars, compelling, includes keyword | [actual] | Pass/Fail |
```
cnunciato: We have a linter for these; no need to waste tokens on 'em.


```suggestion
```

.claude/commands/seo-enhance/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+**Load reference:** Read `references/pulumi-positioning.md`
+
+For content promoting Pulumi capabilities, verify:
+
+| Alignment | Check | Status |
+|-----------|-------|--------|
+| **Differentiators** | Emphasizes relevant advantages (real languages, unified platform, etc.) | Pass/Fail/N/A |
+| **Jobs-to-be-Done** | Connects to customer outcomes | Pass/Fail/N/A |
+| **Terminology** | Uses Pulumi's language, not competitor's | Pass/Fail |
+
```
cnunciato: I don't disagree with these ideas necessarily, but they aren't really SEO. We should pull these into a separate skill that's more focused on branding and positioning. (Happy to help with that BTW.)

```suggestion
```

.claude/commands/seo-enhance/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+Present findings in a structured format:
+
+---
+
+## SEO/AEO Analysis: `[filename]`
+
+### Summary
+
+| Category | Score | Priority Items |
```
cnunciato: Yeah to me, scoring is fine. I just don't love the pass/fail/blindly-generate-"fixes" thing.

.claude/commands/seo-enhance/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+3. After each major change, briefly note what was modified
+4. Run `make lint` to verify no issues introduced
+5. Summarize all changes made
+
+## Content Type-Specific Guidance
+
+### Blog Posts
+
+Additional checks for blog content:
+- **Recency signal**: Year in title or prominent placement
```
cnunciato: This should be an extreme rarity.

.claude/commands/seo-enhance/SKILL.md
cnunciato: >No AI-generated red flags

We may need to be a bit more explicit than this. 😄 (More applicably though, this probably falls outside the scope of what an SEO skill should be concerned with.)

.claude/commands/seo-enhance/SKILL.md
cnunciato: I mentioned this in another comment, but I think this should be called something like `/seo-analyze` to emphasize that it's more about feedback than automated content generation. 

One thing that feels like it's missing, though: some way to express what you're trying to analyze _for_. The basics are one thing — has good headings, applies good link practices, etc. — but there's also the goal of the piece, which is context-specific. This seems important and something we may want to capture by asking a question somehow.

.claude/commands/seo-gap-analyzer/SKILL.md
asafashirov: Makes sense. Removed it from this repo entirely. Moving it to https://github.com/pulumi/pulumi-skills-internal under a new `marketing` plugin where it’s a better fit.

.claude/commands/seo-enhance/SKILL.md
asafashirov: Good catch. The gap analyzer is now out of this repo. For the remaining skill, I’ve completely reframed it as analysis-only; so no content generation, no auto-apply. Just observations and educational suggestions the writer can consider per your guidance.

.claude/commands/new-blog-post.md
asafashirov: Done. Renamed to `/seo-analyze`. The description and behavior now reflect analysis, not enhancement. It should now only present observations and suggestions without generating replacement content or injecting sections.

.claude/commands/seo-enhance/SKILL.md
asafashirov: Updated the usage to expect an explicit file path argument. Removed the “analyze the currently open file” example. IDE context detection is still a fallback but no longer advertised as the primary mode.

.claude/commands/seo-enhance/SKILL.md
asafashirov: Replaced the pass/fail table with an advisory format; just “Pattern | What to Look For” without a status column. The output is now prose observations, not checkmarks. Added a note that good content may not check every box.

asafashirov: Fair point. Removed the Natural Voice pattern entirely. Back to 10 AEO patterns.

.claude/commands/seo-enhance/SKILL.md
asafashirov: Removed the entire SEO Fundamentals step. The skill now focuses purely on patterns the stuff the linter doesn’t cover (at least from what I was able to captured).

.claude/commands/seo-enhance/SKILL.md
asafashirov: Removed the positioning step from this skill. Happy to take you up on the offer to collaborate on a separate branding/positioning skill that's coming soon.

.claude/commands/seo-enhance/SKILL.md
asafashirov: Kept the scoring (single AEO category now, X/10). Removed pass/fail entirely. Removed the “apply changes” step so the skill now presents recommendations and stops. The writer decides what to implement and how.

.claude/commands/seo-enhance/SKILL.md
asafashirov: Updated. Changed from “Working, copy-pasteable code” (which reads as a recommendation to add code) to “If the content includes code, verify it appears functional and copy-pasteable” (just a check on existing code).

.claude/commands/seo-analyze/SKILL.md
```diff
@@ -0,0 +1,171 @@
...
+
+**Load reference:** Read `references/aeo-checklist.md`
+
+Scan content against AI discoverability patterns. For each pattern, note what you observe -- not a binary verdict. Good content may not exhibit every pattern, and that's fine.
+
+| Pattern | What to Look For |
+|---------|------------------|
+| **Quotable Definition** | Opening paragraph contains a clear, direct answer |
+| **Semantic Chunking** | Each section covers one focused concept |
+| **FAQ Sections** | Common questions addressed in Q&A format |
```
cnunciato: I'd prefer we left this out until we have a way to capture these as structured metadata — e.g., as additional frontmatter params that we render separately (and with proper schema.org tagging).


```suggestion
```

PR #17462, opened 2026-02-06
Image fix

SaraDPH:


SaraDPH:
Well @claude only the last commit made the image visible, can you add to correct the other pages with images?

Issue #17270, opened 2026-01-27
Add post-deployment health checks for critical infrastructure


CamSoper:
## Priority
**High** - Prevents production incidents from going undetected

## Problem Statement

The site lacks automated post-deployment validation and runtime monitoring for critical infrastructure components, allowing a Lambda@Edge failure to break all SDK documentation in production without immediate detection.

**What went wrong:** PR #17190 deployed successfully to production, but Lambda@Edge functions began failing at runtime with "ReferenceError: URLPattern is not defined". This broke all SDK documentation (Node.js, Python, .NET) with 503 errors. The deployment appeared successful because:
1. Pulumi deployment succeeded (infrastructure created)
2. Cypress tests passed (only check CSS/JS loading on homepage and getting-started)
3. No post-deployment health checks validated Lambda@Edge execution
4. No CloudWatch alarms detected Lambda@Edge errors

**Current Testing:**
- Cypress tests (`cypress/e2e/site.cy.js`) only validate:
  - Homepage CSS loading
  - JavaScript execution/carousel behavior
  - Getting started page OS chooser
- No tests for SDK documentation endpoints
- No tests for redirect functionality (Lambda@Edge logic)
- No validation that Lambda@Edge functions execute successfully

**Current Monitoring:**
- No CloudWatch alarms for Lambda@Edge errors
- Search URL validation disabled (`.github/workflows/check-search-urls.yml` disabled due to rate limiting issues)

### Part A: Post-Deployment Health Checks (Required)

Add a new workflow `.github/workflows/post-deployment-health-check.yml` that runs after successful deployment and validates:
- SDK documentation endpoints return 200 (not 503)
- Redirect functionality works (Lambda@Edge executes)
- Critical pages load successfully

Example workflow structure:
- Trigger on workflow_run completion
- Determine target environment (production vs testing)
- Test SDK docs for Node.js, Python, .NET, Java
- Test redirect logic
- Send Slack notification on failure

### Part B: CloudWatch Alarms for Lambda@Edge (Required)

Add CloudWatch alarms to `infrastructure/index.ts` to detect Lambda@Edge failures:

```typescript
// After Lambda@Edge function creation
const lambdaEdgeErrorAlarm = new aws.cloudwatch.MetricAlarm("lambda-edge-error-alarm", {
    name: "pulumi-docs-lambda-edge-errors",
    comparisonOperator: "GreaterThanThreshold",
    evaluationPeriods: 1,
    metricName: "Errors",
[...]

PR #17509, opened 2026-02-10
Add post-deployment health checks (#17270)

CamSoper:
Implements automated health monitoring to detect Lambda@Edge and endpoint failures after deployments.

**✅ Zero-risk change:** This PR only adds a new workflow and documentation. No existing workflows or deployment processes are modified.

## Changes
- **Health check workflow** (`.github/workflows/post-deployment-health-check.yml`): Inline bash script using curl to test 9 critical endpoints and 2 Lambda@Edge redirects. Triggers after deployments and sends Slack notifications on failure.
- **Documentation** (`BUILD-AND-DEPLOY.md`): Documents health check system and usage

## Implementation
Uses inline bash with curl (no external dependencies or repository checkout required). This keeps the workflow simple, fast, and self-contained.

## How it works
The workflow runs **after** existing deployment workflows complete successfully. It's completely independent and won't block or interfere with deployments.

## Testing
✅ All health checks pass against production
✅ Validates both production (`pulumi.com`) and testing (`pulumi-test.io`) environments


CamSoper:
We just did a little refactoring, @claude. Hit me with another review.

PR #17552, opened 2026-02-12
Update docs for enum and utility type support

julienp:
Python has supported enums for a while. TS will as of next week's release. TS will also support `Partial<T>` and `Required<T>`.


julienp:
@claude no

PR #17572, opened 2026-02-15
Add blog post: How We Load Data into Snowflake in Seconds with Pulumi

sicarul:
## Summary

  Adds a new blog post covering three production-tested patterns for loading data into Snowflake using reusable Pulumi ComponentResources.

  The post walks through a complete end-to-end example: GitHub webhooks → Lambda (HMAC validation) → Amazon Data Firehose → Snowflake via the Snowpipe Streaming API.
  Data lands in seconds with no S3 intermediate.

  ### What's covered

  - **Pulumi ESC setup**: OIDC authentication to both AWS and Snowflake with composable environments — no static secrets stored anywhere
  - **Direct Firehose-to-Snowflake streaming**: The primary pattern, using the native Snowflake destination added to Firehose. Includes TLS key pair auth, ServiceUser
  with least-privilege grants, and immediate flushing configuration
  - **S3 auto-ingest via Snowpipe**: An alternative for teams that need S3 as the system of record or can't use PrivateLink. Covers the storage integration ↔ IAM role
  circular dependency and SQS notification wiring
  - **Batch stage loading**: Orchestrator-controlled `COPY INTO` for full control over timing and deduplication
  - **Production tips**: Consolidated IAM policies, dynamic policy building with `Output.all()`, per-environment ESC configuration, and secret generation with
  `RandomPassword`
  - **Publishing components**: Git-based sharing via `pulumi package add` and the Pulumi Cloud Private Registry for cross-language SDK generation

  Based on three years of running 25+ pipelines in production. Companion template:
https://github.com/pulumi-demos/examples/tree/main/python/aws-snowflake-data-loading-real-time


sicarul:
Ignore the current date, confirming the publish date with @SaraDPH

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+---
+
+When you manage dozens of data-loading pipelines into Snowflake, copy-pasting resources becomes a maintenance problem: IAM (Identity and Access Management) policies drift, naming conventions diverge, and every new source is a chance to introduce a subtle misconfiguration. This post shows how to encapsulate those patterns into composable components and walks through the production lessons we learned running 25+ pipelines for over three years.
+
+We'll pipe GitHub webhooks into Snowflake with a few lines of Pulumi code. Lambda validates, Amazon Data Firehose streams directly into Snowflake, and data lands in seconds, all wired together by a reusable [`ComponentResource`](/docs/iac/concepts/components/). We also cover two alternative loading patterns (S3 auto-ingest and batch) available in the [companion template](https://github.com/pulumi-demos/examples/tree/main/python/aws-snowflake-data-loading-real-time). We use Pulumi ESC to handle authentication to both AWS and Snowflake.
```
cnunciato:
```suggestion
We'll pipe GitHub webhooks into Snowflake with a few lines of Pulumi code. Lambda validates, Amazon Data Firehose streams directly into Snowflake, and data lands in seconds, all wired together by a reusable [`ComponentResource`](/docs/iac/concepts/components/). We also cover two alternative loading patterns (S3 auto-ingest and batch) available in the [companion template](https://github.com/pulumi-demos/examples/tree/main/python/aws-snowflake-data-loading-real-time). We use [Pulumi ESC](/docs/esc/) to handle authentication to both AWS and Snowflake.
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+- **Log and telemetry pipelines:** application logs, CloudTrail events, or IoT telemetry flowing through Firehose into Snowflake for analytics.
+- **Multi-source data pipelines:** teams managing dozens of sources into a centralized Snowflake Data Warehouse with consistent infrastructure and naming conventions.
+
+[Josh Kodroff](/blog/author/josh-kodroff/) wrote an excellent [introduction to Snowpipe with Pulumi](https://medium.com/snowflake/lightning-fast-elt-for-python-devs-with-aws-snowpipe-and-pulumi-4eaf056dd097). This post builds on his work using the newest [Snowflake](https://www.pulumi.com/registry/packages/snowflake/) and [AWS](https://www.pulumi.com/registry/packages/aws/) provider APIs and the direct Firehose-to-Snowflake destination, which wasn't available when Josh wrote his post. Some resource names and grant patterns will also differ if you're comparing the two.
+
+## Architecture overview
+
+The following diagram shows the recommended pipeline for loading GitHub webhooks into Snowflake in seconds:
```
cnunciato:
```suggestion
The following diagram shows the recommended pipeline for loading GitHub webhooks into Snowflake:
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+
+![Architecture diagram showing the pipeline: GitHub Webhook sends events to AWS Lambda for HMAC validation, then Firehose streams directly to Snowflake's REPOSITORY_EVENTS_DIRECT landing table. S3 is used only for backup and errors.](architecture-direct.png)
+
+GitHub sends webhook events to a Lambda Function URL, which validates the HMAC (Hash-based Message Authentication Code) signature and forwards the payload to Amazon Data Firehose. Firehose streams records directly into Snowflake via the Snowpipe Streaming API. Data appears in Snowflake within seconds. S3 is used only as a backup destination for failed records.
```
cnunciato:
```suggestion
GitHub sends webhook events to a [Lambda Function URL](https://docs.aws.amazon.com/lambda/latest/dg/urls-configuration.html), which validates the webhook's HMAC (Hash-based Message Authentication Code) signature and forwards the payload to [Amazon Data Firehose](https://docs.aws.amazon.com/firehose/latest/dev/what-is-this-service.html). Firehose streams records directly into Snowflake via the [Snowpipe Streaming](https://docs.snowflake.com/en/user-guide/snowpipe-streaming/data-load-snowpipe-streaming-overview) API. Data appears in Snowflake within seconds. S3 is used only as a backup destination for failed records.
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+
+This post demonstrates the pattern with AWS (Lambda, Firehose) and Snowflake. The direct Firehose-to-Snowflake destination is an AWS-native feature that works with any Snowflake account. The S3-based patterns (auto-ingest and batch) also support [Azure Blob Storage](https://www.pulumi.com/registry/packages/azure-native/api-docs/storage/blobcontainer/) and [Google Cloud Storage](https://www.pulumi.com/registry/packages/gcp/api-docs/storage/bucket/) via Snowflake storage integrations.
```
cnunciato: Would this be a good link for this?

```suggestion
This post demonstrates the pattern with AWS (Lambda, Firehose) and Snowflake. The direct [Firehose-to-Snowflake destination](https://www.snowflake.com/en/developers/guides/getting-started-with-snowflake-and-aws-kdf/) is an AWS-native feature that works with any Snowflake account. The S3-based patterns (auto-ingest and batch) also support [Azure Blob Storage](https://www.pulumi.com/registry/packages/azure-native/api-docs/storage/blobcontainer/) and [Google Cloud Storage](https://www.pulumi.com/registry/packages/gcp/api-docs/storage/bucket/) via Snowflake storage integrations.
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+## Project setup
+
+Start a new Pulumi Python project and configure it to use [uv](https://docs.astral.sh/uv/) for dependency management:
+
+```bash
+mkdir snowpipe-data-loading && cd snowpipe-data-loading
+pulumi new python --name snowpipe-data-loading
+```
+
+Update `Pulumi.yaml` to use the `uv` toolchain:
```
cnunciato:
```suggestion
Update `Pulumi.yaml` to use [uv](https://docs.astral.sh/uv/) for dependency management:
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+
+GitHub sends webhook events to a Lambda Function URL, which validates the HMAC (Hash-based Message Authentication Code) signature and forwards the payload to Amazon Data Firehose. Firehose streams records directly into Snowflake via the Snowpipe Streaming API. Data appears in Snowflake within seconds. S3 is used only as a backup destination for failed records.
+
+This post demonstrates the pattern with AWS (Lambda, Firehose) and Snowflake. The direct Firehose-to-Snowflake destination is an AWS-native feature that works with any Snowflake account. The S3-based patterns (auto-ingest and batch) also support [Azure Blob Storage](https://www.pulumi.com/registry/packages/azure-native/api-docs/storage/blobcontainer/) and [Google Cloud Storage](https://www.pulumi.com/registry/packages/gcp/api-docs/storage/bucket/) via Snowflake storage integrations.
+
+## Project setup
+
+Start a new Pulumi Python project and configure it to use [uv](https://docs.astral.sh/uv/) for dependency management:
```
cnunciato:
```suggestion
Start a new Pulumi Python project and choose [uv](https://docs.astral.sh/uv/) for dependency management when prompted:
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+
+```bash
+mkdir snowpipe-data-loading && cd snowpipe-data-loading
+pulumi new python --name snowpipe-data-loading
```
cnunciato: Passing ``--name` isn't necessary, as the directory name will be picked up automatically and used as the project name:

```suggestion
pulumi new python
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
cnunciato: This will already have been set if the user selected `uv`:


```suggestion
Notice `Pulumi.yaml` shows `uv` as your selected toolchain:
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
kramhuber: > Lambda validates, Amazon Data Firehose streams directly...

Is there a word missing at the beginning? Feels like a sentence fragment, but I'm not sure what the intent is

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
kramhuber: Consider using a numbered list

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
kramhuber:
```suggestion
The direct Firehose-to-Snowflake destination is an AWS-native feature that works with any Snowflake account. The S3-based patterns (auto-ingest and batch) also support [Azure Blob Storage](https://www.pulumi.com/registry/packages/azure-native/api-docs/storage/blobcontainer/) and [Google Cloud Storage](https://www.pulumi.com/registry/packages/gcp/api-docs/storage/bucket/) via Snowflake storage integrations.
```

Possibly redundant

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+SELECT CONTENT:github_event::STRING AS event_type,
+       CONTENT:payload:repository:full_name::STRING AS repo,
+       METADATA:IngestionTime::TIMESTAMP AS ingested_at
+FROM LANDING_ZONE_WEBHOOKS.GITHUB.REPOSITORY_EVENTS_DIRECT
+ORDER BY ingested_at DESC;
+```
+
+You should see rows with event types like `star`, `push`, or `issues`, real GitHub events flowing through the entire pipeline. The `METADATA` column includes Firehose metadata like `IngestionTime`, which you can use to track end-to-end latency.
+
+## Alternative patterns: S3 auto-ingest and batch loading
```
kramhuber: I'd split these into separate posts. 

1. More content
2. Makes scenario 1 easy to grok. I'm not sure how many people will keep reading a post of this length

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+     'pulumi-github>=6.0.0' 'pulumi-random>=4.0.0' 'pulumi-tls>=5.0.0'
+```
+
+That's it. `uv` creates the virtual environment and lockfile automatically. Pulumi will use `uv run` under the hood to execute your program.
+
+All examples in this post are in Python, but Pulumi supports multiple languages. You can implement the same components in TypeScript, Go, .NET, Java, or YAML.
+
+## Managing credentials with Pulumi ESC
+
+This project needs credentials for two cloud providers, AWS and Snowflake. Hardcoding secrets in `Pulumi.<stack>.yaml` is insecure and doesn't scale: secrets leak into version control, rotate manually, and multiply across environments. [Pulumi ESC](/docs/esc/) (Environments, Secrets, and Configuration) solves this by providing dynamic, short-lived credentials for both providers via OIDC (OpenID Connect). No static secrets are stored anywhere.
```
cnunciato: Jumping right to "hardcoding secrets is insecure" is abrupt (no one suggested it) and could also be misread (as I read it initially) that Pulumi secrets are insecure, which they aren't (they're both secure and safe to check in). Suggest instead:

```suggestion
This project requires credentials for two cloud providers, AWS and Snowflake. To avoid having to manage credentials locally, it uses [Pulumi ESC](/docs/esc/) (Environments, Secrets, and Configuration) to obtain dynamic, short-lived credentials for both providers via OIDC (OpenID Connect). 
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+
+If you're not using Pulumi ESC, you can add these credentials directly to your stack configuration file (`Pulumi.<stack>.yaml`) using `pulumi config set`.
```
cnunciato: Wouldn't want to miss this:

```suggestion
If you're not using Pulumi ESC, you can add these credentials directly to your stack configuration file (`Pulumi.<stack>.yaml`) using [`pulumi config set --secret`](https://www.pulumi.com/docs/iac/concepts/secrets/).
```

More importantly though, if we're going with ESC, we should explain why, because it's responsible for a good portion of the length of the post, and it looks to me like normal provider config would be fine here, to keep the cognitive load down and stay focused on the topic.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+
+### How it works
+
+You define an ESC environment that opens OIDC connections to AWS and Snowflake. When you run `pulumi up`, ESC exchanges a Pulumi-issued OIDC token for short-lived credentials from each provider, then injects them into your stack config automatically. Your Pulumi program doesn't change at all; the providers read their configuration from `pulumiConfig` as usual.
+
+The following diagram shows how the ESC environments compose. The stack imports a combined dev environment (which itself imports the AWS and Snowflake base environments) plus a separate GitHub environment:
```
cnunciato:
```suggestion
The following diagram shows how this project's ESC environments compose. The stack imports a combined dev environment (which itself imports the AWS and Snowflake base environments) plus a separate GitHub environment:
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+
+AWS --> Dev
+SF --> Dev
+Dev --> Stack
+GH --> Stack
+```
+
+### One-time Snowflake setup
+
+Follow the [Snowflake OIDC login guide](/docs/esc/integrations/dynamic-login-credentials/snowflake-login/) for full details. The key step is creating a security integration that trusts Pulumi's OIDC issuer. Replace <your-pulumi-org> with your Pulumi organization name:
```
cnunciato: Brackets are causing this to be interpreted as HTML. 

Also suggest cutting out the "key step" — for one, it's AI-speak, but it's also really the _only_ step (there isn't anything to do _but_ create that integration):

```suggestion
Follow the [Snowflake OIDC login guide](/docs/esc/integrations/dynamic-login-credentials/snowflake-login/) to create nan integration that trusts Pulumi's OIDC issuer. Replace `<your-pulumi-org>` with your Pulumi organization name:
```

Perhaps importantly though: Where does this SQL get run?

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: No, this is the flow, the webhook is validated by lambda, sent to firehose and so on, i thought it was ok how it is written but i'm not native so i trust you if it's incorrect

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+        "StringEquals": {
+          "api.pulumi.com/oidc:aud": "aws:<your-pulumi-org>"
+        }
+      }
+    }
+  ]
+}
+```
+
+This policy allows any Pulumi ESC environment in your org to assume the role. You can further restrict it using subject claim conditions. The role needs whatever AWS permissions your Pulumi program requires. For this project that means Lambda, Firehose, S3, and IAM.
```
cnunciato:
````suggestion
1. **Create an IAM OIDC identity provider.** In the AWS console (or via CLI/IaC), add a new identity provider with provider URL `https://api.pulumi.com/oidc` and audience `aws:<your-pulumi-org>`. The audience string scopes trust to your specific Pulumi organization, so only OIDC tokens issued for that org are accepted.

1. **Create an IAM role with a trust policy.** The trust policy validates that the OIDC token was issued by Pulumi and carries the correct audience claim before allowing `sts:AssumeRoleWithWebIdentity`:

    ```json
    {
      "Version": "2012-10-17",
      "Statement": [
        {
          "Effect": "Allow",
          "Principal": {
            "Federated": "arn:aws:iam::<your-aws-account-id>:oidc-provider/api.pulumi.com/oidc"
          },
          "Action": "sts:AssumeRoleWithWebIdentity",
          "Condition": {
            "StringEquals": {
              "api.pulumi.com/oidc:aud": "aws:<your-pulumi-org>"
            }
          }
        }
      ]
    }
    ```

    This policy allows any Pulumi ESC environment in your org to assume the role. You can further restrict it using subject claim conditions. The role needs whatever AWS permissions your Pulumi program requires. For this project that means Lambda, Firehose, S3, and IAM.
````

File: content/blog/near-real-time-data-loading-snowflake-pulumi/index.md [hunk 14]
```diff
@@ -0,0 +1,1129 @@
...
+
+### The ESC environments
+
+Rather than putting everything in one flat environment, we split credentials by provider and compose them with [`imports`](/docs/esc/environments/imports/). Each base environment is single-concern (one cloud provider) and handles its own credential-to-config mapping. The combined environment merges them and adds cross-cutting config like `aws:region`. This way we can reuse credentials among many stacks, each one only importing what they need. It also makes switching from a development to production environment, for example, much more straightforward.
+
+**AWS-only environment.** Handles AWS OIDC login and sets environment variables for CLI operations:
+
+{{% notes type="tip" %}}
+You can run these commands or create your environments [in the Pulumi Cloud Console](https://app.pulumi.com)
```
cnunciato:
```suggestion
You can run these commands or create your environments [in the Pulumi Cloud Console](https://app.pulumi.com).
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+}
+```
+
+This policy allows any Pulumi ESC environment in your org to assume the role. You can further restrict it using subject claim conditions. The role needs whatever AWS permissions your Pulumi program requires. For this project that means Lambda, Firehose, S3, and IAM.
+
+### The ESC environments
+
+Rather than putting everything in one flat environment, we split credentials by provider and compose them with [`imports`](/docs/esc/environments/imports/). Each base environment is single-concern (one cloud provider) and handles its own credential-to-config mapping. The combined environment merges them and adds cross-cutting config like `aws:region`. This way we can reuse credentials among many stacks, each one only importing what they need. It also makes switching from a development to production environment, for example, much more straightforward.
+
+**AWS-only environment.** Handles AWS OIDC login and sets environment variables for CLI operations:
```
cnunciato:
```suggestion
#### AWS-only environment

This environment handles AWS OIDC login and sets environment variables for CLI operations:
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+          roleArn: arn:aws:iam::123456789012:role/pulumi-esc-oidc
+          sessionName: pulumi-snowpipe-demo
+  environmentVariables:
+    AWS_ACCESS_KEY_ID: ${aws.login.accessKeyId}
+    AWS_SECRET_ACCESS_KEY: ${aws.login.secretAccessKey}
+    AWS_SESSION_TOKEN: ${aws.login.sessionToken}
+    AWS_REGION: us-west-2
+```
+
+**Snowflake-only environment.** Handles Snowflake OIDC login:
```
cnunciato:
```suggestion
#### Snowflake-only environment

This environment handles Snowflake OIDC login:
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+  pulumiConfig:
+    snowflake:organizationName: ${snowflake.organizationName}
+    snowflake:accountName: ${snowflake.accountName}
+    snowflake:user: ${snowflake.login.user}
+    snowflake:token: ${snowflake.login.token}
+    snowflake:authenticator: OAUTH
+    snowflake:role: PULUMI_DEPLOYER
+```
+
+**Combined dev environment.** Imports both base environments. Since each base already maps its own credentials via `pulumiConfig`, the combined environment only needs to add `aws:region`:
```
cnunciato:
```suggestion
#### Combined dev environment

Imports both base environments. Since each base already maps its own credentials via `pulumiConfig`, the combined environment only needs to add `aws:region`:
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+```yaml
+imports:
+  - <project>/snowpipe-demo-aws
+  - <project>/snowpipe-demo-snowflake
+values:
+  pulumiConfig:
+    aws:region: us-west-2
+```
+
+The `imports` directive merges values from both base environments into the combined one. Each base environment handles its own provider config: the AWS base uses `environmentVariables` for CLI access, while the Snowflake base uses `pulumiConfig` to map OIDC credentials to Snowflake provider config keys. The combined dev environment only needs to add `aws:region`. This keeps credential configuration DRY (Don't Repeat Yourself); if the AWS role ARN or Snowflake account changes, you update one environment.
```
cnunciato: I feel like we just said this:

```suggestion
The `imports` directive merges values from both base environments into the combined one. Each base environment handles its own provider config: the AWS base uses `environmentVariables` for CLI access, while the Snowflake base uses `pulumiConfig` to map OIDC credentials to Snowflake provider config keys. This keeps credential configuration DRY (Don't Repeat Yourself); if the AWS role ARN or Snowflake account changes, you need only update one environment.
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+The webhook pipeline needs GitHub credentials to create repository webhooks via the [`pulumi-github`](https://www.pulumi.com/registry/packages/github/) provider. Pick whichever authentication method fits your security posture.
+
+Create a separate ESC environment for GitHub:
+
+```bash
+pulumi env init <your-org>/<project>/snowpipe-demo-github
+pulumi env edit <your-org>/<project>/snowpipe-demo-github
+```
+
+#### Option 1: Personal access token (recommended for individuals)
```
cnunciato: At this point this is starting to feel like a guide to setting up composable environments with Pulumi ESC. That's great, but it's not what this post is about. Someone arriving here expecting to learn about loading data into Snowflake with Pulumi may lose patience (it's a very long post, and this setup takes a while) and/or get the impression (perhaps rightly) that the post is a bit of an ad for ESC.

Ultimately, the ESC configuration should likely be skipped to trim ~600 words from the post (which, at >5K words, is already ~3x the length of a typical post). I'll come back to it to see if my impression holds, but wanted to call it out now while it was fresh in my mind as I'm reading through this.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+  snowpipe-data-loading:environment: dev
+  snowpipe-data-loading:webhook-repo: <your-test-repo-name>
+  snowflake:previewFeaturesEnabled:
+    - snowflake_table_resource
+    - snowflake_storage_integration_aws_resource
+    - snowflake_stage_external_s3_resource
+    - snowflake_pipe_resource
+```
+
+The `snowflake:previewFeaturesEnabled` list is required since `pulumi-snowflake` v2.12.0, which moved [`Table`](https://www.pulumi.com/registry/packages/snowflake/api-docs/table/), [`StorageIntegrationAws`](https://www.pulumi.com/registry/packages/snowflake/api-docs/storageintegrationaws/), [`StageExternalS3`](https://www.pulumi.com/registry/packages/snowflake/api-docs/stageexternals3/), and [`Pipe`](https://www.pulumi.com/registry/packages/snowflake/api-docs/pipe/) behind preview feature flags. The direct pattern only needs `snowflake_table_resource`; the S3-based patterns (auto-ingest and batch) need all four.
```
cnunciato: > direct pattern
> S3-based patterns

By now I've lost context as to what these refer to.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+
+That's the entire change. No credentials are stored in the Pulumi configuration yaml; the providers pick up their credentials from ESC-injected config transparently.
```
cnunciato: What entire change? We just now wrote this (i.e., nothing existed before this to be changed).

However, since this just seems to repeat the "no local creds" message, we can probably just remove it.

```suggestion
```

I will say though that at this point, it does feel like we've spent a third of the post just setting up config. What started out looking like a fairly simple thing now feels like it requires quiet a lot of up-front setup.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,1129 @@
...
+
+## Shared infrastructure
+
+The direct streaming pipeline needs an [S3 bucket](https://www.pulumi.com/registry/packages/aws/api-docs/s3/bucket/) for backup/errors, a Snowflake database, and a schema:
```
cnunciato: We seem to be missing an instruction to put this code somewhere. Does it go into `__main__.py`?

```suggestion
The direct streaming pipeline needs an [S3 bucket](https://www.pulumi.com/registry/packages/aws/api-docs/s3/bucket/) for backup/errors, a Snowflake database, and a schema. Add the following code to `__main__.py`:
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: I got a bit crazy, i'll remove those 2 and plan more posts to cover them

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,887 @@
...
+  options:
+    toolchain: uv
+description: Production-grade data loading with Pulumi, AWS, and Snowflake
+```
+
+Add the provider dependencies: the [AWS](/registry/packages/aws/), [Snowflake](/registry/packages/snowflake/), [GitHub](/registry/packages/github/), [Random](/registry/packages/random/), and [TLS](/registry/packages/tls/) providers:
+
+```bash
+uv add 'pulumi>=3.0.0,<4.0.0' 'pulumi-aws>=7.0.0' 'pulumi-snowflake>=2.0.0' \
+     'pulumi-github>=6.0.0' 'pulumi-random>=4.0.0' 'pulumi-tls>=5.0.0'
```
cnunciato: This works, and I get that it guards against future breakage, but it isn't how this is normally done. Ordinarily, and the `pulumi` dependency should already be there by virtue of using the `python` template.

```suggestion
uv add pulumi-aws pulumi-snowflake pulumi-github pulumi-random pulumi-tls
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,887 @@
...
+---
+
+When you manage dozens of data-loading pipelines into Snowflake, copy-pasting resources becomes a maintenance problem: IAM (Identity and Access Management) policies drift, naming conventions diverge, and every new source is a chance to introduce a subtle misconfiguration. This post shows how to encapsulate those patterns into composable components and walks through the production lessons we learned running 25+ pipelines for over three years.
+
+We'll pipe GitHub webhooks into Snowflake with a few lines of Pulumi code. Lambda validates the webhook signature, Amazon Data Firehose streams the payload directly into Snowflake, and data lands in seconds; all wired together by a reusable [`ComponentResource`](/docs/iac/concepts/components/). The [companion template](https://github.com/pulumi-demos/examples/tree/main/python/aws-snowflake-data-loading-real-time) also includes S3 auto-ingest and batch loading patterns, which we'll cover in upcoming posts. We use Pulumi ESC to handle authentication to both AWS and Snowflake.
```
cnunciato:
```suggestion
When you manage dozens of data-loading pipelines, copying and pasting IaC configurations between them is a recipe for mishap. IAM policies can drift, naming conventions diverge, and every new source is a new opportunity to make a mistake — not to mention compound the problem of duplication. In this post, we'll show you how you can identify and encapsulate common patterns into composable components and walk through the production lessons we've learned running 25+ pipelines for over three years.
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,887 @@
...
+- **Change Data Capture (CDC) replication:** streaming database changes via Debezium, [Fivetran](https://www.pulumi.com/registry/packages/fivetran/), or [AWS DMS](https://www.pulumi.com/registry/packages/aws/api-docs/dms/) into Snowflake.
+- **Log and telemetry pipelines:** application logs, CloudTrail events, or IoT telemetry flowing through Firehose into Snowflake for analytics.
+- **Multi-source data pipelines:** teams managing dozens of sources into a centralized Snowflake Data Warehouse with consistent infrastructure and naming conventions.
+
+[Josh Kodroff](/blog/author/josh-kodroff/) wrote an excellent [introduction to Snowpipe with Pulumi](https://medium.com/snowflake/lightning-fast-elt-for-python-devs-with-aws-snowpipe-and-pulumi-4eaf056dd097). This post builds on his work using the newest [Snowflake](https://www.pulumi.com/registry/packages/snowflake/) and [AWS](https://www.pulumi.com/registry/packages/aws/) provider APIs and the direct Firehose-to-Snowflake destination, which wasn't available when Josh wrote his post. Some resource names and grant patterns will also differ if you're comparing the two.
+
```
cnunciato: Props as always to Josh, but this doesn't seem relevant here. Let's keep the momentum going and dive right in.

```suggestion
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,887 @@
...
+- **Change Data Capture (CDC) replication:** streaming database changes via Debezium, [Fivetran](https://www.pulumi.com/registry/packages/fivetran/), or [AWS DMS](https://www.pulumi.com/registry/packages/aws/api-docs/dms/) into Snowflake.
+- **Log and telemetry pipelines:** application logs, CloudTrail events, or IoT telemetry flowing through Firehose into Snowflake for analytics.
+- **Multi-source data pipelines:** teams managing dozens of sources into a centralized Snowflake Data Warehouse with consistent infrastructure and naming conventions.
+
+[Josh Kodroff](/blog/author/josh-kodroff/) wrote an excellent [introduction to Snowpipe with Pulumi](https://medium.com/snowflake/lightning-fast-elt-for-python-devs-with-aws-snowpipe-and-pulumi-4eaf056dd097). This post builds on his work using the newest [Snowflake](https://www.pulumi.com/registry/packages/snowflake/) and [AWS](https://www.pulumi.com/registry/packages/aws/) provider APIs and the direct Firehose-to-Snowflake destination, which wasn't available when Josh wrote his post. Some resource names and grant patterns will also differ if you're comparing the two.
```
cnunciato:
```suggestion
Our own [Josh Kodroff](/blog/author/josh-kodroff/) wrote an excellent [introduction to Snowpipe with Pulumi](https://medium.com/snowflake/lightning-fast-elt-for-python-devs-with-aws-snowpipe-and-pulumi-4eaf056dd097). This post builds on his work using the newest [Snowflake](https://www.pulumi.com/registry/packages/snowflake/) and [AWS](https://www.pulumi.com/registry/packages/aws/) provider APIs and the direct Firehose-to-Snowflake destination, which wasn't available when Josh wrote his post. Some resource names and grant patterns will also differ if you're comparing the two.
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,887 @@
...
+<!--more-->
+
+## Who is this for
+
+If you're loading data into Snowflake and want reusable, composable infrastructure, this post is for you. Common use cases include:
+
+- **Webhook and event ingestion:** SaaS webhooks (GitHub, Stripe, HubSpot), application events, or any HTTP-driven data that needs to land in Snowflake quickly.
+- **Change Data Capture (CDC) replication:** streaming database changes via Debezium, [Fivetran](https://www.pulumi.com/registry/packages/fivetran/), or [AWS DMS](https://www.pulumi.com/registry/packages/aws/api-docs/dms/) into Snowflake.
+- **Log and telemetry pipelines:** application logs, CloudTrail events, or IoT telemetry flowing through Firehose into Snowflake for analytics.
+- **Multi-source data pipelines:** teams managing dozens of sources into a centralized Snowflake Data Warehouse with consistent infrastructure and naming conventions.
```
cnunciato: Let's just stay focused on what we'll cover; this other stuff isn't really relevant here.

```suggestion
## What we'll cover

If you're loading data into Snowflake and want reusable, composable infrastructure, this post is for you. Here's what we'll cover:

* Handling and validating GitHub webhooks with AWS Lambda 
* Streaming webhook payloads directly into Snowflake with Amazon Data Firehose 
* Wiring it all up with a reusable Pulumi [`ComponentResource`](/docs/iac/concepts/components/)
 
The [companion template](https://github.com/pulumi-demos/examples/tree/main/python/aws-snowflake-data-loading-real-time) also includes S3 auto-ingest and batch loading patterns, which we'll cover in upcoming posts. We also use Pulumi ESC to handle authentication to both AWS and Snowflake using OpenID Connect.
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,887 @@
...
+
+[Josh Kodroff](/blog/author/josh-kodroff/) wrote an excellent [introduction to Snowpipe with Pulumi](https://medium.com/snowflake/lightning-fast-elt-for-python-devs-with-aws-snowpipe-and-pulumi-4eaf056dd097). This post builds on his work using the newest [Snowflake](https://www.pulumi.com/registry/packages/snowflake/) and [AWS](https://www.pulumi.com/registry/packages/aws/) provider APIs and the direct Firehose-to-Snowflake destination, which wasn't available when Josh wrote his post. Some resource names and grant patterns will also differ if you're comparing the two.
+
+## Architecture overview
+
+The following diagram shows the recommended pipeline for loading GitHub webhooks into Snowflake in seconds:
```
cnunciato:
```suggestion
The following diagram shows the architecture in more detail:
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,887 @@
...
+- **Change Data Capture (CDC) replication:** streaming database changes via Debezium, [Fivetran](https://www.pulumi.com/registry/packages/fivetran/), or [AWS DMS](https://www.pulumi.com/registry/packages/aws/api-docs/dms/) into Snowflake.
+- **Log and telemetry pipelines:** application logs, CloudTrail events, or IoT telemetry flowing through Firehose into Snowflake for analytics.
+- **Multi-source data pipelines:** teams managing dozens of sources into a centralized Snowflake Data Warehouse with consistent infrastructure and naming conventions.
+
+[Josh Kodroff](/blog/author/josh-kodroff/) wrote an excellent [introduction to Snowpipe with Pulumi](https://medium.com/snowflake/lightning-fast-elt-for-python-devs-with-aws-snowpipe-and-pulumi-4eaf056dd097). This post builds on his work using the newest [Snowflake](https://www.pulumi.com/registry/packages/snowflake/) and [AWS](https://www.pulumi.com/registry/packages/aws/) provider APIs and the direct Firehose-to-Snowflake destination, which wasn't available when Josh wrote his post. Some resource names and grant patterns will also differ if you're comparing the two.
+
+## Architecture overview
+
```
cnunciato, line 40: To tighten this up and keep the momentum going, I'd suggest this:

* Remove this heading
* Bring "The following diagram..." from below (with edits) up and position it between the bullets and the Josh mention
* Bring the diagram up as well and position it right before Josh

```suggestion
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,887 @@
...
+Update `Pulumi.yaml` to use the `uv` toolchain:
+
+```yaml
+name: snowpipe-data-loading
+runtime:
+  name: python
+  options:
+    toolchain: uv
+description: Production-grade data loading with Pulumi, AWS, and Snowflake
```
cnunciato: This won't be there if they've followed this workflow unless we tell them to type it in, so we can leave it out.

```suggestion
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,887 @@
...
+```
+
+Add the provider dependencies: the [AWS](/registry/packages/aws/), [Snowflake](/registry/packages/snowflake/), [GitHub](/registry/packages/github/), [Random](/registry/packages/random/), and [TLS](/registry/packages/tls/) providers:
```
cnunciato:
```suggestion
Add the provider dependencies for [AWS](/registry/packages/aws/), [Snowflake](/registry/packages/snowflake/), [GitHub](/registry/packages/github/), and the [Random](/registry/packages/random/) and [TLS](/registry/packages/tls/) providers:
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+schema = snowflake.Schema(
+    "demo-schema",
+    name="GITHUB",
+    database=database.name,
+)
+```
+
+## Building the direct ingestion ComponentResource
+
+Amazon Data Firehose supports [Snowflake as a native destination](https://docs.aws.amazon.com/firehose/latest/dev/create-destination.html#create-destination-snowflake) via the Snowpipe Streaming API. Firehose streams records directly into Snowflake, bypassing S3 entirely. Data lands in seconds and you manage fewer resources.
```
cnunciato: I feel like I missed the memo on where S3 is getting "bypassed". Can we just remove this without losing any meaning?


```suggestion
Amazon Data Firehose supports [Snowflake as a native destination](https://docs.aws.amazon.com/firehose/latest/dev/create-destination.html#create-destination-snowflake) via the Snowpipe Streaming API. Firehose streams records directly into Snowflake.
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+
+### The Lambda handler
+
+The [Lambda function](https://www.pulumi.com/registry/packages/aws/api-docs/lambda/function/) is the entry point for GitHub webhooks. It validates the HMAC-SHA256 signature, wraps the payload in an envelope with the event type, and forwards it to Firehose:
+
+```python
```
cnunciato, line 151: Where does this go? Does the reader need to create a new file?

This makes me wonder whether the post would be better presented as a _review_ of the code (e.g., after checking it out from the examples repo), rather than an end-to-end walkthrough. This approach has several benefits: keeps the pace moving, lets you focus on the important parts and what's interesting about them, sidesteps bit rot, etc. Something to think about, especially given the length of this post.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+
+    firehose.put_record(
+        DeliveryStreamName=STREAM_NAME,
+        Record={"Data": record.encode()},
+    )
+
+    return {"statusCode": 200, "body": "OK"}
+```
+
+The envelope format `{"github_event": "<type>", "payload": {...}}\n` is important. The `github_event` field (e.g., `push`, `pull_request`, `star`) comes from the `x-github-event` header and lets downstream queries filter by event type.
```
cnunciato: What is the significance of the `\n`s here? At first it looked like a mistake, but I see it's also in the code, which suggests it's actually relevant. If it is, we should explain how.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+
+### Why direct Firehose to Snowflake?
+
+With an S3 intermediate path, you wait for Firehose to buffer records (60 seconds), then for Snowpipe to detect the new file and load it. Total latency: about two minutes. With the direct Snowflake destination, Firehose uses the Snowpipe Streaming API to insert records as soon as they arrive, in seconds.
```
cnunciato: >With an S3 intermediate path

What S3 intermediate path? Is there some explanation missing here? Something like this maybe (totally guessing here)?


```suggestion
Data like this is normally written to and loaded from Amazon S3. But with an S3 intermediate path, you must wait for Firehose to buffer records (60 seconds), then for Snowpipe to detect the new file and load it. Total latency: about two minutes. With the direct Snowflake destination, Firehose uses the Snowpipe Streaming API to insert records as soon as they arrive, in seconds.
```

Not sure if this is right, but it feels like we're trying to position against something that hasn't been stated (or if it has, I must've missed it).

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+
+The direct path also removes the need for S3 event notifications, SQS queues, external stages, and pipe resources. S3 is still used, but only as a backup destination for failed records.
+
+### The DirectSnowflakeIngestion component
+
+The component creates everything needed for the direct path: a TLS key pair for Snowflake authentication, a Snowflake service user with least-privilege grants, the landing table, a Firehose delivery stream with `destination="snowflake"`, and a Lambda function with a public URL.
+
+The `ColumnDef` dataclass is a simple schema definition used across all pipeline components. It lives in `components/snowpipe_pipeline.py`:
```
cnunciato: Should we tell the reader to create a new file here?

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+@dataclass
+class ColumnDef:
+    """Column definition for a Snowflake table."""
+
+    name: str
+    type: str
+    nullable: bool = True
+```
+
+```python
```
cnunciato, line 211: Does this code go in that file as well, or a new one?

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+    s3_backup_mode: str = "FailedDataOnly"
+    buffering_interval: int = 0
+    buffering_size: int = 1
+    retry_duration: int = 60
+    data_loading_option: str = "VARIANT_CONTENT_AND_METADATA_MAPPING"
+    content_column_name: str = "CONTENT"
+    metadata_column_name: str = "METADATA"
+```
+
+The key configuration values: `buffering_interval=0` and `buffering_size=1` tell Firehose to flush immediately (no batching delay), and `VARIANT_CONTENT_AND_METADATA_MAPPING` maps the JSON payload to a `CONTENT` column and Firehose metadata to a `METADATA` column.
```
cnunciato: The way "key" is used here sounds like AI-speak:

```suggestion
The `buffering_interval` and `buffering_size` parameters tell Firehose to flush immediately (no batching delay), and `VARIANT_CONTENT_AND_METADATA_MAPPING` maps the JSON payload to a `CONTENT` column and Firehose metadata to a `METADATA` column.
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+
+### Why direct Firehose to Snowflake?
+
+With an S3 intermediate path, you wait for Firehose to buffer records (60 seconds), then for Snowpipe to detect the new file and load it. Total latency: about two minutes. With the direct Snowflake destination, Firehose uses the Snowpipe Streaming API to insert records as soon as they arrive, in seconds.
+
+The direct path also removes the need for S3 event notifications, SQS queues, external stages, and pipe resources. S3 is still used, but only as a backup destination for failed records.
+
+### The DirectSnowflakeIngestion component
+
+The component creates everything needed for the direct path: a TLS key pair for Snowflake authentication, a Snowflake service user with least-privilege grants, the landing table, a Firehose delivery stream with `destination="snowflake"`, and a Lambda function with a public URL.
```
cnunciato: This component is ultimately what this post is really about, right? Encapsulating all this Snowflake infrastructure into a reusable component? If so, it feels like we could probably do more to set up what we're about to dive into so the reader understands and appreciates the value of the approach.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+        )
+
+        # --- Snowflake role, user, and grants ---
+        sf_role = snowflake.AccountRole(
+            f"{name}-sf-role",
+            name=args.snowflake_role_name,
+            opts=pulumi.ResourceOptions(parent=self),
+        )
+
+        user_name = f"FIREHOSE_{name.upper().replace('-', '_')}_USER"
```
cnunciato, line 288: Would it be worth calling out easily missed things, like the string interpolation here, that make solutions like these even more compelling and valuable? (Stuff like this really is unique to Pulumi.)

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+```
+
+A few things to note:
+
+- **TLS key pair for authentication.** The component generates an RSA key pair using the [`pulumi-tls`](/registry/packages/tls/) provider. The public key is assigned to the Snowflake service user; the private key (PKCS#8 format, base64-encoded) is passed to Firehose. No passwords or OAuth tokens are stored.
+- **`ServiceUser` instead of `User`.** Snowflake [service users](https://docs.snowflake.com/en/sql-reference/sql/create-user#service-type) can't log in interactively. They authenticate only via key pair, which is exactly what Firehose needs.
+- **`destination="snowflake"` on Firehose.** This tells Firehose to use the Snowpipe Streaming API rather than writing to S3. The `s3_configuration` block is still required, but only for backup/error records.
+- **Immediate flushing.** `buffering_interval=0` and `buffering_size=1` ensure records are sent to Snowflake as soon as they arrive, minimizing latency. Tune according to your needs.
+
+{{% notes type="warning" %}}
```
cnunciato, line 516: Does this need to be warn-level?

```suggestion
{{% notes %}}
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+    ),
+    events=["push", "pull_request", "issues", "star"],
+)
+
+# Exports
+pulumi.export("webhook_url", direct.function_url)
+pulumi.export("firehose_stream", direct.firehose_stream_name)
+```
+
+That's the entire pipeline. One component, one GitHub webhook, one secret. The `DirectSnowflakeIngestion` component handles the TLS key pair, Snowflake service user, landing table, Firehose stream, and Lambda function internally.
```
cnunciato: Would it be worth emphasizing the win here by mentioning something like _and now you can reuse this component across all of your pipelines_?

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+
+{{% notes type="info" %}}
+The full code for this example is available at [https://github.com/pulumi-demos/examples/tree/main/python/aws-snowflake-data-loading-real-time](https://github.com/pulumi-demos/examples/tree/main/python/aws-snowflake-data-loading-real-time).
+{{% /notes %}}
```
cnunciato: We actually built a thing just for this!

```suggestion
The full code for this example is available on GitHub:

{{< github-card repo="pulumi-demos/examples" path="python/aws-snowflake-data-loading-real-time" >}}
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+USE ROLE DATA_READER;
+
+SELECT CONTENT:github_event::STRING AS event_type,
+       CONTENT:payload:repository:full_name::STRING AS repo,
+       METADATA:IngestionTime::TIMESTAMP AS ingested_at
+FROM LANDING_ZONE_WEBHOOKS.GITHUB.REPOSITORY_EVENTS_DIRECT
+ORDER BY ingested_at DESC;
+```
+
+You should see rows with event types like `star`, `push`, or `issues`, real GitHub events flowing through the entire pipeline. The `METADATA` column includes Firehose metadata like `IngestionTime`, which you can use to track end-to-end latency.
```
cnunciato, line 653: Any chance you have a screenshot somewhere (or could make one) that shows what this looks like?

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+
+We'll walk through both patterns in detail in upcoming posts.
+
+## Production tips
+
+A few things we learned running this at scale:
+
+**Use direct Firehose to Snowflake for lowest latency.** The direct path delivers data in seconds and reduces the number of resources you manage (no stages, pipes, SQS queues, or S3 notifications).
+
+**Use separate ESC environments per deployment target.** Each environment (dev, staging, production) needs its own OIDC role ARN, so create separate ESC environments for each. Use [`imports`](/docs/esc/environments/imports/) to share common settings like provider versions and non-secret configuration:
```
cnunciato: We may want to break this out into a separate post, yeah. If we did that, this would be the place to call out where the standard config approach we took here (while pragmatic and certainly secure enough, etc.) gets complicated, and how having a more composable solution addresses that problem.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+        oidc:
+          duration: 1h
+          roleArn: arn:aws:iam::111111111111:role/pulumi-esc-production
+          sessionName: pulumi-snowpipe-production
+  pulumiConfig:
+    aws:region: us-west-2
+    snowflake:role: PULUMI_DEPLOYER
+```
+
+**Generate webhook secrets with Pulumi.** Instead of manually creating a secret and copy-pasting it into both your webhook configuration and your Lambda environment, use `random.RandomPassword` to generate it in Pulumi state. The secret is automatically wired to both the Lambda env var and the GitHub webhook config, and it rotates cleanly if you ever need to replace it.
```
cnunciato: I completely missed that we randomly generated that password. Definitely seems worth calling that out!

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,904 @@
...
+
+The `DirectSnowflakeIngestion` component in this post delivers data from GitHub webhooks into Snowflake in seconds: Lambda validates the HMAC signature, Firehose streams directly to Snowflake via the Snowpipe Streaming API, and the TLS key pair is managed entirely within Pulumi. No S3 intermediate, no SQS queues.
+
+The component accepts pluggable Lambda handlers, so swapping GitHub for Stripe webhooks or any other source is just a matter of providing different `lambda_code` and `lambda_environment` arguments. This pattern has been running in production for over three years across dozens of pipelines without significant changes to the infrastructure code. When you're ready to share components across teams, [`pulumi package add`](/docs/iac/cli/commands/pulumi_package_add/) or [`pulumi package publish`](/docs/iac/cli/commands/pulumi_package_publish/) turns them into versioned, cross-language packages. You'll find the complete example in the [GitHub repository](https://github.com/pulumi-demos/examples/tree/main/python/aws-snowflake-data-loading-real-time).
```
cnunciato: We should call out that _we've_ been running this in production! Give it its own paragraph and share how it's helped us, allowed us to roll our changes across all pipelines, let certain pipelines remain pinned if they needed to (if they did — just spitballing), etc. 

Also we already said this. I personally believe it fits better here than it does above, so feel free to 

```suggestion
The component accepts pluggable Lambda handlers, so swapping GitHub for Stripe webhooks or any other source is just a matter of providing different `lambda_code` and `lambda_environment` arguments. 

We've been using this pattern in production for over three years, across dozens of pipelines, without significant changes. _[More interestingness here]_ When you're ready to share components across teams, [`pulumi package add`](/docs/iac/cli/commands/pulumi_package_add/) or [`pulumi package publish`](/docs/iac/cli/commands/pulumi_package_publish/) turns them into versioned, cross-language packages.

You'll find the full code for this example on GitHub:

{{< github-card repo="pulumi-demos/examples" path="python/aws-snowflake-data-loading-real-time" >}}
```

cnunciato (changes requested):
This is great! Well done. ⭐ (And apologies for taking so long to get through it.) It could use another turn, though, and we should try to bring the word count down to closer to 2,000-2,500 words if we can. Hit me up and we can talk strategies. Breaking the ESC stuff out into a follow-up post might be a good way to go for several reasons, but happy to talk through it with ya.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
cnunciato: These links make sense, but I made some suggestions after this this first pass (in comments below) that reworked the intro slightly to get things moving a bit more quickly. Let me know what you think.

sicarul:
If we remove the esc configuration, i don't see what follow up post would i make to add it, it's part of the flow 🤔

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: A normal provider config would mean that we need to manage Snowflake users and their private keys directly, the length of the post would remain similar and we'd not have higlighted ESC which actually shortens the time to success. I can make the section more abridged instead of removing it?

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: It is ran on Snowflake

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: Sorry, it's leftover from when i had the 3 methods in the post.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: It's JSONL - Each line has to end with a newline
I'll make a note of it, i skipped explaining it.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: Yeah it was from a previous version, also it's the most common way to load a data warehouse from aws

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: I don't understand this comment - I thought this guide as a step-by-step guide of how to configure and implement a data pipeline from Github webhook to table in Snowflake; this is very difficult to set up with existing resources because all of the examples are very vague or use "go configure auth" language without explaining how to actually do it, leaving the reader to have to open dozens of different pages to be successful.

I understand it's long but it was my intention to be thorough, if the blog is not the place for such a guide i can close the PR until there is a better place for it.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: Why are you removing the values in your suggestion?

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: I saw the github card but seemed like a huge thing in the middle of the post so i removed it matching the pattern in https://www.pulumi.com/blog/setting-up-iam-authentication-for-postgres-on-aws/

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: But we are calling it out, i don't understand 🤔

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: I can make it in the week by bringing the infra back up

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: Why isn't it used here? https://www.pulumi.com/blog/setting-up-iam-authentication-for-postgres-on-aws/

I added it and it was huge in the post, which is why i removed it.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: We're already at a very lengthy post

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: Yes, i really think this is an important security warning

sicarul:
I've reduced the post to ~2.8k words by mainly chopping most of the ESC configuration, it'll be harder to follow with less explanation but i still feel this is better than asking the user to throw all their secrets into the yaml.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
cnunciato: We don't enforce it or anything, but it's a great way to uplevel the visibility of your repo, and IMO, this does look right (wanted to make sure my recollection was correct): 

image

Definitely encourage it, but up to you.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
cnunciato: Ah, thanks — this helps a lot. I didn't realize this. (I had no background on this post before the PR for it emerged.) 

This raises a good point, and why it can help to talk through these things first before jumping right into a post. Generally speaking (always exceptions of course), the blog _isn't_ the best place for in-depth walkthroughs like this, in part because they end up becoming canonical references, fall out of date, etc., and they're tricky to keep updated because a post is a point-in-time reference, and not really intended as a main source of instruction. Tutorials in general are usually the right place for a lengthy, deep-dive, end-to-end guide like this.

That said, this is totally fine for now. But going forward, it wouldn't hurt to reach out in the marketing channel first — we're always happy to work with you on shaping these kinds of things to minimize confusion and shorten the review/revision cycle. In this case, I might've suggested a tutorial coupled with a blog post covering the high-level story — how we use components on the data team, for instance — and then reference the tutorial as part of that post. That sort of approach usually results in a shorter, higher-level post that's more easily digestible, still carries the message, and lets the reader dive into the detailed canonical walkthrough if they want to.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,701 @@
...
+---
+
+When you manage dozens of data-loading pipelines into Snowflake, copy-pasting resources becomes a maintenance problem: IAM (Identity and Access Management) policies drift, naming conventions diverge, and every new source is a chance to introduce a subtle misconfiguration. This post shows how to encapsulate those patterns into composable components and walks through the production lessons we learned running 25+ pipelines for over three years.
+
+When you manage dozens of data-loading pipelines, copying and pasting IaC configurations between them is a recipe for mishap. IAM policies can drift, naming conventions diverge, and every new source is a new opportunity to make a mistake — not to mention compound the problem of duplication. In this post, we'll show you how you can identify and encapsulate common patterns into composable components and walk through the production lessons we've learned running 25+ pipelines for over three years.
```
cnunciato: I think we want one or the other of these, not both. Your preference!

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: Do you have an example of a tutorial+blog like you are referring to? It's the first time i've heard about this pattern.

It was a weekend project, i didn't think about reaching out to Marketing to coordinate, will take it into account if i want to write again

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: I probably misapplied one of your feedback batches by mistake, will correct

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,701 @@
...
+
+That's it. `uv` creates the virtual environment and lockfile automatically, and Pulumi uses `uv run` under the hood to execute your program.
+
+All examples in this post are in Python, but Pulumi supports multiple languages. You can implement the same components in TypeScript, Go, .NET, Java, or YAML.
+
+## Managing credentials with Pulumi ESC
+
+This project needs credentials for AWS, Snowflake, and GitHub. Rather than managing long-lived secrets locally, we can use [Pulumi ESC](/docs/esc/) to obtain dynamic, short-lived OIDC credentials for AWS and Snowflake at runtime. When you run `pulumi up`, ESC exchanges a Pulumi-issued OIDC token for temporary credentials from each provider and injects them into your stack config automatically. If you prefer not to use ESC, you can set credentials directly with [`pulumi config set --secret`](/docs/iac/concepts/secrets/).
+
+Here is a single ESC environment that handles all three providers:
```
cnunciato, line 85: So much more digestible! Thanks for doing this. 🙌🏻

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,701 @@
...
+        DeliveryStreamName=STREAM_NAME,
+        Record={"Data": record.encode()},
+    )
+
+    return {"statusCode": 200, "body": "OK"}
+```
+
+The envelope format `{"github_event": "<type>", "payload": {...}}\n` is important. The `github_event` field (e.g., `push`, `pull_request`, `star`) comes from the `x-github-event` header and lets downstream queries filter by event type. The trailing newline delimits records in the [S3 backup destination](https://docs.aws.amazon.com/firehose/latest/dev/basic-deliver.html), where Firehose concatenates them into files.
+
+Another interesting part of this snippet is that instead of manually creating a secret which would have to be copy-pasted into both your webhook configuration and your Lambda environment; we are using `random.RandomPassword` to generate it in Pulumi state.
```
cnunciato: Tiny nits:

```suggestion
Another interesting part of this snippet is that instead of manually creating a secret which would have to be copy-pasted into both your webhook configuration and your Lambda environment, we are using `random.RandomPassword` to generate and store it securely it in Pulumi state.
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,701 @@
...
+
+# Exports
+pulumi.export("webhook_url", direct.function_url)
+pulumi.export("firehose_stream", direct.firehose_stream_name)
+```
+
+That's the entire pipeline. One component, one GitHub webhook, one secret. The `DirectSnowflakeIngestion` component handles the TLS key pair, Snowflake service user, landing table, Firehose stream, and Lambda function internally, and now you can reuse this component for as many pipelines as you need.
+
+The full code for this example is available on GitHub:
+{{< github-card repo="pulumi-demos/examples" path="python/aws-snowflake-data-loading-real-time" >}}
```
cnunciato, line 619: This didn't seem to render out right. I _think_ it needs to be this:

```suggestion

{{< github-card repo="pulumi-demos/examples" path="python/aws-snowflake-data-loading-real-time" >}}
```

This fixes it locally for me, at least.

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
```diff
@@ -0,0 +1,701 @@
...
+
+The `DirectSnowflakeIngestion` component in this post delivers data from GitHub webhooks into Snowflake in seconds: Lambda validates the HMAC signature, Firehose streams directly to Snowflake via the Snowpipe Streaming API, and the TLS key pair is managed entirely within Pulumi. No S3 intermediate, no SQS queues.
+
+The component accepts pluggable Lambda handlers, so swapping GitHub for Stripe webhooks or any other source is just a matter of providing different `lambda_code` and `lambda_environment` arguments. This pattern has been running in production for over three years across dozens of pipelines without significant changes to the infrastructure code. When you're ready to share components across teams, [`pulumi package add`](/docs/iac/cli/commands/pulumi_package_add/) or [`pulumi package publish`](/docs/iac/cli/commands/pulumi_package_publish/) turns them into versioned, cross-language packages. You'll find the complete example in the [GitHub repository](https://github.com/pulumi-demos/examples/tree/main/python/aws-snowflake-data-loading-real-time).
```
cnunciato: Little things, but I think it works a bit better to explicitly say "we've" been running this ourselves (the same language we used in the intro), so would suggest that, and the mention of `package add` seems a bit more than necessary, since we just mentioned it a few lines up.

```suggestion
The component accepts pluggable Lambda handlers, so swapping GitHub for Stripe webhooks or any other source is just a matter of providing different `lambda_code` and `lambda_environment` arguments. We've been running this pattern in production for over three years across dozens of pipelines without significant changes to the infrastructure code. 

You'll find the complete example in the [GitHub repository](https://github.com/pulumi-demos/examples/tree/main/python/aws-snowflake-data-loading-real-time).
```

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
cnunciato: Yes. Here's something I wrote last year that uses the space of the blog to explain the important concepts, using a backing repo that the user checks out if they want. There's a little bit of how-to, but the post is primarily about explaining how Bazel and Buildkite work together, and less about walking the user through a step-by-step process:

https://buildkite.com/resources/blog/fully-dynamic-pipelines-with-bazel-and-buildkite/

content/blog/near-real-time-data-loading-snowflake-pulumi/index.md
sicarul: Weird, it renders fine for me 
image

PR #17610, opened 2026-02-20
Improving font size and text boundary for meta generator

sicarul:
The previous setting was too conservative


.claude/commands/blog-meta-image/references/asset-catalog.md
```diff
@@ -85,8 +85,8 @@ From `catalog.yaml`:
 - **Line-height**: 110% of font size
 - **X position**: 90px from left edge (aligned with Pulumi logo)
 - **Y position**: 80px from top edge (top-anchored, flows downward)
-- **Max width**: 520px (safe text area on left half)
-- **Font size**: variable (44-80px depending on title length)
+- **Max width**: 700px (text area on left half)
+- **Font size**: variable (60-104px depending on title length)
```
vctrfrnndz, line 90: Set the weight

```
**Font weight**: 700 (bold)
```

.claude/commands/blog-meta-image/references/asset-catalog.md
```diff
@@ -85,8 +85,8 @@ From `catalog.yaml`:
...
```
vctrfrnndz, line 89: Set letter spacing to tight: -0.025em

.claude/commands/blog-meta-image/references/asset-catalog.md
sicarul: We already use Inter Bold: https://github.com/pulumi/docs/blob/8fa30c040d1098fa938a960e22c8f4edb24e3c17/.claude/commands/blog-meta-image/assets/fonts/Inter-Bold.ttf

.claude/commands/blog-meta-image/references/asset-catalog.md
sicarul: It's set here to -0.03, we can change to -0.025, but rounding will probably land at around the same value:
https://github.com/pulumi/docs/blob/8fa30c040d1098fa938a960e22c8f4edb24e3c17/.claude/commands/blog-meta-image/scripts/compose_meta_image.py#L80

vctrfrnndz: I would adjust, at this level/size it makes a difference.

.claude/commands/blog-meta-image/references/asset-catalog.md
vctrfrnndz: Is it new? because before it didn't look like bold font weight on the generated assets, so might be something to check out. Wondering if this ttf is getting used at all.

.claude/commands/blog-meta-image/assets/fonts/Inter.ttf
cnunciato, line 1: Can we not use the font file we already have in `./static`? https://github.com/pulumi/docs/tree/master/static/fonts

sicarul: My bad, let me check.

sicarul: Fixed! Sorry about that

PR #17638, opened 2026-02-23
Add complete MyVpcComponent example to the transformations page

### LINKED ISSUE #11652: Transformation Resource Option Page Doesn't Provide a Complete Example
### What happened?

While triaging a user-reported issue today, I came across the [Transformation](https://www.pulumi.com/docs/intro/concepts/resources/options/transformations/) resource option page. I noticed this page doesn't contain a complete example, only a snippet.

Specifically, the example shows code for a component resource named `MyVpcComponent`. However, it doesn't provide the definition of `MyVpcComponent`. This opens up the possibility for confusion; users might not understand what they have to do with the component options. For example, you could implement `MyVpcComponent` incorrectly as follows:

```typescript
class MyVpcComponent extends pulumi.ComponentResource {
    constructor(name, opts) {
        super("pkg:index:MyVpcComponent", name, {}, {});
    }
}
```

### Steps to reproduce

1. Visit the [transformation](https://www.pulumi.com/docs/intro/concepts/resources/options/transformations/#transformations) resource option page.
2. Observe the code provided.
3. Note that it isn't a full example in any language and cannot be run.

### Expected Behavior

I expected all identifiers in the example program to be defined. 

### Actual Behavior

`MyVpcComponent` is undefined.

### Output of `pulumi about`

_No response_

### Additional context

_No response_

### Contributing

joeduffy:
## Summary

The `transformations` resource option page included a usage example that instantiated `MyVpcComponent` without ever defining it, leaving readers without enough context to understand how the component is structured or how transformations propagate through child resources.

This PR adds a complete `MyVpcComponent` definition in all supported languages (TypeScript, Python, Go, C#, Java, and YAML), showing how to:
- Extend `ComponentResource`
- Create child AWS VPC and Subnet resources with `parent: this`
- Call `registerOutputs()` to complete the component lifecycle

The existing transformation example follows in a second chooser block, with two correctness fixes also applied:
- **Go**: the transformation usage called `MyVpcComponent(...)` as a bare function call, which is not valid Go syntax. It is now corrected to `NewMyVpcComponent(ctx, ...)` to match the constructor pattern.
- **TypeScript**: the return object now has a trailing comma and semicolon for consistency.

Additionally, a missing trailing newline has been added to fix a markdownlint warning.

Fixes #11652

---


joeduffy:
Thanks for the thorough review! Both issues are addressed in the follow-up commit:

1. **Go error handling**: Added `if err != nil { return err }` after the `NewMyVpcComponent` call in the transformation snippet so the returned error is properly checked.

2. **Python type annotation**: Replaced `Optional[pulumi.ResourceOptions]` with `pulumi.ResourceOptions | None` which uses the built-in union syntax (Python 3.10+) and avoids the need for a `typing` import in the snippet.

PR #17677, opened 2026-02-25
Fix --import-pending-creates flag examples in troubleshooting docs and blog post

### LINKED ISSUE #15902: --import-pending-creates flag example does not work
### Problem description

The example for the --import-pending-creates flag in [Repairing State with Pulumi Refresh](https://www.pulumi.com/blog/repairing-state-with-pulumi-refresh/) fails with an error:

`pulumi refresh --import-pending-creates="aws_native:s3:Bucket::my-bucket my-bucket-6e3d099"`
`error: each URN must be followed by an ID: found an odd number of entries`

### Suggestions for a fix

This variant works but is ugly:

`pulumi refresh --import-pending-creates="aws_native:s3:Bucket::my-bucket" --import-pending-creates="my-bucket-6e3d099"`

Is there a better way intended? We should update the blog post and add an example to the cmdline docs.

joeduffy:
Fixes #15902

## Problem

Multiple users have reported that the documented examples for `pulumi refresh --import-pending-creates` do not work.

The flag is defined as a `StringArray` in cobra, which means each invocation of the flag appends exactly **one** string to the internal slice. The CLI then processes the slice as alternating URN/ID pairs. Given this behavior:

- The **blog post** (`repairing-state-with-pulumi-refresh`) combined the URN and physical ID in a single quoted string (`--import-pending-creates="URN ID"`). Cobra treats this as a slice of length 1 (odd), causing: `each URN must be followed by an ID: found an odd number of entries`

- The **troubleshooting page** (`interrupted-updates.md`) passed the ID as a second positional argument after the flag value (`--import-pending-creates "URN" "ID"`). Cobra treats the ID as an unknown subcommand, causing: `unknown command "ID" for "pulumi refresh"`

Both errors match exactly what users are seeing in the issue.

## Fix

The correct syntax passes the URN and its physical ID as **two consecutive, separate flag invocations**:

```sh
pulumi refresh \
  --import-pending-creates "urn:pulumi:dev::myproject::aws:s3/bucket:Bucket::mybucket" \
  --import-pending-creates "my-bucket-abc123" \
  --yes
```

For multiple pending creates, URN and ID invocations alternate:

```sh
pulumi refresh \
  --import-pending-creates "urn:pulumi:dev::myproject::aws:s3/bucket:Bucket::bucket1" \
  --import-pending-creates "bucket1-id" \
  --import-pending-creates "urn:pulumi:dev::myproject::aws:s3/bucket:Bucket::bucket2" \
  --import-pending-creates "bucket2-id" \
  --yes
```

## Changes

- `content/docs/support/troubleshooting/common-issues/interrupted-updates.md`: Fixed the single-resource and multi-resource examples; added guidance on how to find the full URN; added a warning note explaining both failure modes and what not to do.
- `content/blog/repairing-state-with-pulumi-refresh/index.md`: Fixed the original example (which used a short URN form and combined URN+ID in one string); updated the flag description to clarify the invocation pattern.

---

### DIFF
--- a/content/blog/repairing-state-with-pulumi-refresh/index.md
+++ b/content/blog/repairing-state-with-pulumi-refresh/index.md
@@ -81,9 +81,9 @@ You can learn more by going to the [`pulumi refresh` documentation](https://www.
 
 There are some flags you can pass to modify the new behavior:
 
-- `--import-pending-creates` tells Pulumi to import a specific resource that was created during a pending operation. Pass the resource's full URN as one invocation and its cloud provider physical ID as the next invocation. This is equivalent to selecting the `import` option when running in interactive mode, but only for the specified resources.
+- `--import-pending-creates` tells Pulumi to import resources that were created during pending operations. For each resource, pass its full URN as one flag invocation and its cloud provider physical ID as the next. This is equivalent to selecting the `import` option when running in interactive mode, but only for the specified resources.
 - `--clear-pending-creates` will remove any pending CREATEs from the state.
-- `--skip-pending-creates` will skip all pending CREATEs. This is the default behavior when running the CLI outside of interactive more.
+- `--skip-pending-creates` will skip all pending CREATEs. This is the default behavior when running the CLI outside of interactive mode.
 
 These flags can be combined:
 


joeduffy:
Thanks for the review! Both items addressed in the follow-up commit:

- Fixed the singular/plural inconsistency in the `--import-pending-creates` description — now consistently refers to "resources" (plural) throughout
- Fixed the pre-existing typo: "outside of interactive more" → "outside of interactive mode"

PR #17717, opened 2026-02-27
Fix misleading getOutputDetails example in stack references docs

### LINKED ISSUE #13855: StackReference getOutputDetails example is misleading
The [`StackReference` docs for `getOutputDetails`](https://www.pulumi.com/docs/iac/concepts/stacks/#reading-outputs-from-stack-references) seem wrong/misleading.

Namely, this part:

> On the other hand, as an example of using getOutputDetails, suppose that your referenced stack creates a VPC and exports a list of public subnets as a JSON-serialized string, and you want to add a bastion host to each subnet.

This is not what I'm seeing. If I export an array of subnet IDs from my program, and then fetch them in another stack with `getOutputDetails`, I get back _an array_, not _a string_. I was misled into assuming `getOutputDetails` would simply return those as a string, but it seems to automatically parse them.

After re-reading the docs, I see that it says "exports a list of public subnets as a JSON-serialized string"; so I guess they docs aren't technically incorrect. However, (a) most people wouldn't go out of their way to JSON serialize their stack outputs (they'd just output them) and (b) we should make it clear that if you export structured objects, their type will automatically round trip without you needing to do anything.

(Assuming that is correct -- given the docs, I am not sure!)

I am also unsure if this is true universally across all supported languages. I would hope so.

joeduffy:
## Description

This PR addresses the misleading `getOutputDetails` example in the [stack references documentation](https://www.pulumi.com/docs/iac/concepts/stacks/#reading-outputs-from-stack-references).

### What was wrong

The previous example described a scenario where the referenced stack exported a list of subnets as a **JSON-serialized string**, then used `JSON.parse` (and language equivalents like `json.Unmarshal` in Go and `JsonConvert.DeserializeObject` in C#) to decode the value back into an object. This was confusing and misleading for two reasons:

1. It is unusual to intentionally JSON-serialize a value before exporting it from a Pulumi stack. Most programs export native values (arrays, objects, strings, numbers) directly.
2. The example implied that `getOutputDetails` always returns a raw JSON string requiring manual parsing. In reality, `getOutputDetails` returns the output's native value — if you export an array, you get an array back, not a JSON string.

The bug report notes that users were misled into assuming they needed to JSON-serialize their stack outputs to use `getOutputDetails`, or that the method returned raw strings from all output types.

### What changed

- **Bullet description**: Updated the `getOutputDetails` entry in the output-reading method list to mention the `value` and `secretValue` fields of the returned `StackReferenceOutputDetails` object, which is the key distinction users need to understand about this method.
- **Example section**: Replaced the JSON-serialized string scenario with a simple, realistic example — retrieving a database hostname (a plain string) from a referenced stack. Each language example now shows direct access to `.value` (or `.Value`/`getValue()`), with a comment explaining that `.secretValue` is used for outputs marked as secret in the referenced stack.

Closes #13855

---

### DIFF
--- a/content/docs/iac/concepts/stacks.md
+++ b/content/docs/iac/concepts/stacks.md
@@ -833,12 +833,12 @@ BucketObject logFile = new BucketObject("log", new BucketObjectArgs.Builder()
 
 {{< /chooser >}}
 
-`getOutputDetails` is useful when you need direct, synchronous access to an output value.
+`getOutputDetails` is useful when you need direct access to a resolved output value.
 This is most helpful when you want to inspect whether a value is marked as a secret, or when
 you need to use the value in your program logic without calling `Output.apply()`. The method
-returns a `StackReferenceOutputDetails` object whose `value` field holds the raw value for
-non-secret outputs, and whose `secretValue` field holds the raw value for outputs the
-referenced stack has marked as secret.
+returns an `OutputDetails` object whose `value` field holds the raw value for non-secret
+outputs, and whose `secretValue` field holds the raw value for outputs the referenced stack
+has marked as secret.
 
 As an example, suppose your referenced stack exports a database hostname as a plain string:
 


joeduffy:
Both issues addressed in the follow-up commit:

1. Changed `StackReferenceOutputDetails` to `OutputDetails` in the prose paragraph to match the existing bullet description and the blog post on this feature.
2. Replaced "direct, synchronous access" with "direct access to a resolved output value" to avoid implying that `await` is unnecessary in TypeScript and C#.

@claude

PR #17727, opened 2026-02-27
Clarify that namespaced config values must be set in the stack configuration file

### LINKED ISSUE #12153: Clarify the reference to `aws:region` in the namespaced-config docs
[@thomas11 pointed out](https://pulumi.slack.com/archives/C011EMDQ8JE/p1719384336591379) that in the docs for accessing namespaced configuration, we mention how to do the lookup, but don't make it clear where the setting is that's being looked up.

This is made more confusing when the selected language is YAML because the program file contains a `config` section, but no `aws:region` setting:

image

We should clarify that these snippets all assume the existence of a stack-config file (`Pulumi.dev.yaml` for example) that looks something like this:

```yaml
config:
  aws:region: us-west-2
```

joeduffy:
## Description

This PR addresses the documentation gap described in #12153: the section of the configuration docs that shows how to read a namespaced config value (such as `aws:region`) from within a Pulumi program does not explain where that value actually comes from.

### What was wrong

The existing code examples demonstrate how to construct a namespaced `Config` object and read a value from it, but they assume without stating that `aws:region` has already been set somewhere. For most language tabs (TypeScript, Python, Go, C#, Java), this is merely an implicit assumption that experienced users might recognize. For the YAML language tab, however, the gap is especially confusing: Pulumi YAML programs define their own config values in an inline `config:` block, so readers naturally wonder why `aws:region` does not appear there — and may not realize it lives in the stack configuration file instead.

### What changed

Added a brief explanation and a concrete example immediately before the code chooser block. The new text:

- States explicitly that these examples assume `aws:region` has already been configured for the stack
- Shows how to set it using the CLI (`pulumi config set aws:region us-west-2`)
- Shows what the resulting stack configuration file (`Pulumi.dev.yaml`) looks like with that value set

The change is minimal and surgical — two sentences and a short YAML snippet — placed exactly where a reader encountering these examples for the first time needs the context.

Fixes #12153

---

### DIFF
--- a/content/docs/iac/concepts/config.md
+++ b/content/docs/iac/concepts/config.md
@@ -193,9 +193,10 @@ config:
 
 {{< /chooser >}}
 
-To access a namespaced configuration value, such as one set for a provider library like `aws`, you must pass the library's name to the constructor. The examples below assume the value has already been set in your stack's configuration file (e.g., `Pulumi.dev.yaml`) — either by running `pulumi config set aws:region us-west-2` at the command line, or by adding it to the file directly:
+To access a namespaced configuration value, such as one set for a provider library like `aws`, you must pass the library's name to the constructor. The examples below assume the value has already been set in your stack's configuration file (e.g., `Pulumi.dev.yaml`) — either by running `pulumi config set aws:region us-west-2` from the command line, or by adding it to the file directly:
 
 ```yaml
+# Pulumi.dev.yaml
 config:
   aws:region: us-west-2
 ```


joeduffy:
Thanks for the review feedback. Both suggestions addressed in the follow-up commit:

- Changed "at the command line" to "from the command line"
- Added `# Pulumi.dev.yaml` as a comment inside the YAML code block, making it immediately clear which file the snippet belongs to

CamSoper (approved):
Nice fix! The added context about needing to pre-set the value in the stack config file (with the CLI example and YAML snippet) is exactly what was missing. Clean and surgical.

PR #17766, opened 2026-03-02
[EXPERIMENTAL] Add Decap CMS PoC for blogging workflow

CamSoper:
## Summary

- Adds `static/admin/index.html` and `static/admin/config.yml` — the two static files required to run [Decap CMS](https://decapcms.org/) on top of the existing Hugo + GitHub workflow.
- Adds `make serve-cms` target that starts Hugo and the Decap local backend simultaneously via `concurrently`.
- Uses `local_backend: true` for the PoC — no OAuth app or external services needed; click **Login** at `http://localhost:1313/admin/` after running `make serve-cms`.

## How to test

1. `make ensure`
2. `make serve-cms`
3. Open `http://localhost:1313/admin/`
4. Click **Login** (no credentials required with local backend)
5. Confirm the **Blog Posts** collection loads with all frontmatter fields
6. Create a new draft post and verify `content/blog/{slug}/index.md` appears on disk

## Production path (future work)

Probably none, if we want to keep it. It can work locally.

If we wanted to make it work directly from `pulumi.com`, we'd need to do some add'l work. Switching from local backend to the real GitHub backend requires three steps:

1. **GitHub OAuth App** — Register at `github.com/organizations/pulumi/settings/applications`; set the callback URL to `<oauth-proxy>/callback`.
2. **OAuth proxy** — Deploy a Cloudflare Worker (or similar) that handles the GitHub OAuth exchange. Set `GITHUB_CLIENT_ID` and `GITHUB_CLIENT_SECRET` as secrets.
3. **Update `config.yml`** — Comment out `local_backend: true`, uncomment the `backend:` block, and set `base_url` to the proxy URL.

No CloudFront or S3 changes are needed — `/admin/` is served as static files.

### DIFF
--- a/static/admin/index.html
+++ b/static/admin/index.html
@@ -9,7 +9,7 @@
     <body>
         <script src="https://unpkg.com/decap-cms@^3.0.0/dist/decap-cms.js"></script>
         <script>
-            CMS.registerPreviewStyle('/admin/preview.css');
+            CMS.registerPreviewStyle("/admin/preview.css");
         </script>
     </body>
 </html>
--- a/static/admin/preview.css
+++ b/static/admin/preview.css
@@ -1,6 +1,5 @@
 body {
-    font-family: ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont,
-        "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif;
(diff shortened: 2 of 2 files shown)


CamSoper:
Mothballing this. @cnunciato and team have cool stuff in the works.

PR #17770, opened 2026-03-03
Clarify protect option inheritance and how to override on child resources

### LINKED ISSUE #10150: How to explicitly set protect resource option on child resources?
Submitted On: 08-17-2023
Comment(s): ""The default is to inherit this value from the parent resource""; I suppose the correct way to 'uninherit' this is to explicitly define 'protect: false' on every resource whose parent has a 'protect: true' set, but this isn't 100% clear in the wording. Additionally, are there any global config settings to change this default? If so, please document them here too!

joeduffy:
Fixes #10150

## What changed

The `protect` option docs previously had only a brief one-sentence mention of inheritance ("The default is to inherit this value from the parent resource, and `false` for resources without a parent.") with no explanation of how to override that behavior.

This PR adds two new sections to `content/docs/iac/concepts/resources/options/protect.md`:

**"Overriding inherited protection"** — explains that child resources inherit `protect` from their parent by default, and shows how to explicitly set `protect: false` on a child to allow it to be deleted independently of a protected parent. Includes full multi-language examples (TypeScript, Python, Go, C#, Java, YAML).

**"Applying protection to all resources"** — answers the issue author's question about whether there is a global config option for stack-wide protection. There isn't a dedicated flag; the answer is to use stack transforms, and this section explains that and links to the transforms reference.

## Additional fixes

- Corrected the existing YAML example in the first chooser block, which was missing the resource name key and was therefore invalid Pulumi YAML syntax.
- Fixed a double space in prose ("unprotected.  There are" → single space).
- Updated the TypeScript example to have consistent spacing (`{ protect: true }` with space before closing brace).
- Removed a spurious trailing semicolon from the Go example.

---

### DIFF
--- a/content/docs/iac/concepts/resources/options/protect.md
+++ b/content/docs/iac/concepts/resources/options/protect.md
@@ -84,7 +84,7 @@ resources:
 
 ## Overriding inherited protection
 
-Child resources inherit the `protect` option from their [parent resource](/docs/iac/concepts/options/parent). When a parent resource has `protect: true`, all of its children are also protected by default. To allow a specific child resource to be deleted independently of its protected parent, explicitly set `protect: false` on that child.
+Child resources inherit the `protect` option from their [parent resource](/docs/iac/concepts/resources/options/parent/). When a parent resource has `protect: true`, all of its children are also protected by default. To allow a specific child resource to be deleted independently of its protected parent, explicitly set `protect: false` on that child.
 
 The following example creates a protected parent resource alongside a child resource with protection explicitly disabled:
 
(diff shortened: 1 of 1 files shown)


joeduffy:
Fixed both blocking issues in the follow-up commit (5367491):

1. Corrected the link in "Overriding inherited protection" to `/docs/iac/concepts/resources/options/parent/`
2. Corrected the link in "Applying protection to all resources" to `/docs/iac/concepts/resources/options/transforms/#stack-transforms`

On suggestion #2 (stack transform code example): I intentionally kept the section as a prose pointer rather than showing a full chooser block, to avoid duplicating what is already illustrated in detail on the transforms page. The section's purpose is to answer the "is there a global flag?" question concisely and point readers in the right direction — a full code example would make the section significantly longer without adding much that isn't already one click away. Happy to revisit if the team disagrees.

Suggestion #3 (Go error discarding) is consistent with existing style in this file; no action taken.

PR #17774, opened 2026-03-03
Improve SAML SSO docs structure and cross-linking

### LINKED ISSUE #17773: Ensure consistent instructions and troubleshooting for SAML SSO setup
We need some structural changes first:

- Move all SSO troubleshooting content to a separate page, put it last in the nav, and ensure all SSO tutorial pages link to it.
- In the troubleshooting page, explain that SAML Admin is designed to solve this problem, but if they do not have a SAML Admin, they will need to contact support
- Put SAML Admin in a separate page as it's an important concept that deserves its own page

Then we have several pages that have SAML SSO setup instructions:

- https://www.pulumi.com/docs/administration/access-identity/saml/auth0/
- https://www.pulumi.com/docs/administration/access-identity/saml/entra/
- https://www.pulumi.com/docs/administration/access-identity/saml/onelogin/
- https://www.pulumi.com/docs/administration/access-identity/saml/gsuite/
- https://www.pulumi.com/docs/administration/access-identity/saml/okta/

The instructions in each page should be consistent when it comes time to paste the SAML document into Pulumi Cloud:

- Add a warning to all SSO config guide pages near the top that Pulumi Orgs should set at least one SAML admin (it's not clear whether there can be more than one) so that their orgs stay accessible (not clear at the present time if this can be done before setting up SSO, so let's assume it has to be done after)
- The current path in the UI is Access Management (under Settings) -> Other tab -> Change Requirements (under Membership Requirements) -> Select SAML SSO -> Next -> paste the doc -> Apply Changes
- The section should have the same heading and nearly identical text, except possibly for the paste instructions, which may mention where we're copying the text from
- There should be no screenshots of the Pulumi Cloud UI in these pages and any unused screenshots should be removed.

jkodroff:
Splits saml-admin and troubleshooting into separate pages, adds self-hosted <> cloud cross-links, removes prerequisites sections, and converts blockquotes to notes shortcodes.

Fixes #17773

### DIFF
--- a/content/docs/administration/access-identity/saml/okta.md
+++ b/content/docs/administration/access-identity/saml/okta.md
@@ -2,14 +2,14 @@
 title_tag: Configuring Okta | SAML SSO
 meta_desc: This page provides a walkthrough important aspects of configuring
            Okta as a SAML SSO identity provider (IdP).
+meta_image: /images/docs/meta-images/docs-meta.png
 title: Okta
 h1: "SAML: Configuring Okta"
-meta_image: /images/docs/meta-images/docs-meta.png
 menu:
   administration:
     name: Okta
     parent: administration-access-identity-saml
-    weight: 5
+    weight: 50
     identifier: pulumi-cloud-access-management-saml-okta
 aliases:
 - /docs/reference/service/saml-okta/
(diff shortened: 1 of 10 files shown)


content/docs/administration/access-identity/saml/okta.md
```diff
@@ -129,23 +121,20 @@ The JSON payload the inline hook sends to Okta should contain the following:
         {
           "op": "add",
           "path": "/authentication/sessionLifetime",
-          "value": 21600 // lifetime in seconds
+          "value": 21600
```
isabeldepapel: If we're going to take out the comment that this is lifetime in seconds, do we want to add that info to the text somewhere? Otherwise I'm not sure it's super obvious that value should be in seconds (but maybe that's just me!)

isabeldepapel (approved):
SAML info looks good to me, thanks for cleaning this up!

content/docs/administration/access-identity/saml/okta.md
jkodroff: Oof, that is NOT intentional. Will fix.

PR #17827, opened 2026-03-05
RBAC docs: role assignment to users and teams, tag-based access rules, org-wide settings

djgrove:
Updates the RBAC documentation to reflect the full permission model launch.

## What changed

The custom roles system now supports assigning roles directly to organization members and to teams, in addition to organization access tokens. Teams can hold multiple role assignments simultaneously, with members receiving the union of all team roles and their own baseline role. Custom roles can also include tag-based access rules that dynamically grant a permission set to any resource whose tags match specified conditions, without needing to name each resource individually.

## Pages updated

`_index.md` adds a section explaining how organization-wide access settings (the Access Management toggles for actions like creating stacks or teams) differ from RBAC permission scopes. These two systems coexist: org-wide settings apply unconditionally to all members, while scopes are granted per-role.

`roles.md` fixes two misleading references to the Access Management page in the context of the organization default role. The default role is set from Settings > Roles, not from the Access Management page. The Access Management page is where the "Assign custom roles to users" feature toggle lives.

`teams.md` adds the role escalation constraint to the team role assignment description: a team admin must hold the `role:create` and `role:update` scopes themselves in order to assign a role to the team, and cannot assign a role that grants more permissions than they hold.

`scopes.md` adds an upfront note distinguishing RBAC scopes from the legacy org-wide access toggles.

`permission-sets.md` adds a callout under the Organization settings entity type clarifying that RBAC org-settings scopes are separate from the org-wide Access Management toggles.

## Scope additions and removals

Several scopes were corrected or removed to match the actual API.

Removed from permission set tables: `stack_annotations:read` and `stack_annotations:update` (annotation scopes are intentionally not user-facing), `environment:read_decrypt` and `environment_version:read_decrypt` (replaced by the correct `environment:open` and `environment_version:open` names used in the API), and `insights_account:update_policy_results` from Account Write (this scope is internal and not in the public registry).

Removed entirely: the deprecated `stack:list`, `environment:list`, and `insights_account:list` scopes, which are no longer used for authorization. The `services` scope group (`services:read/write/create/admin`) was also removed as these scopes are not part of the public scope registry.

Added: `policy_results:create`, `policy_results:update`, and `policy_results:delete` to the Insights Policy section in org settings. `change_gate:create`, `change_gate:update`, and `change_gate:delete` (approval rules) were added to both the Environment scopes page and the Org settings scopes page, reflecting that change gates exist at both levels.


CamSoper (approved):
Looks great — thorough update to the RBAC docs covering the full permission model launch. Made minor style fixes: title case on H1s and normalized numbered lists per style guide.

PR #17937, opened 2026-03-12
Blog Feature and Meta Image Updates

jeffmerrick:
### Proposed changes
To make the blog images consistent but unique moving forward I'm proposing we default to custom designed feature images with an option to use template fallbacks. (Designing the custom images will be on me.)

- Adds `feature_image` in addition to the `meta_image`, feature images are displayed in the blog listing and on the single blog posts, meta images now are only used as meta images. (This sets us up to move to a grid layout for the blog listing.)
- Updates the `/blog-meta-image` skill to generate feature and meta images from either the new templates or a provided custom image.
- Removes old templates and provides some new basic generic ones. New templates can be added as we have more artwork designed. Leaving out Pulumipus illustrations for now as they may be revisited.
- Adds custom feature images for the 4 newest blog posts - if we move forward with this approach I can create more going back further. (Maybe the first two pages of the blog.)
- Hides the image completely if there is no feature image in the name of consistency, but leaves old meta images untouched.

### Future changes
- More and more specialized templates for specific types of blog posts (announcements, releases, features, etc.)
- Possibly review older high traffic/evergreen content as candidates for a custom designed image.
- Grid layout for blog listings.

---
Resolves MA-542

### DIFF
--- a/.claude/commands/blog-meta-image/scripts/compose_meta_image.py
+++ b/.claude/commands/blog-meta-image/scripts/compose_meta_image.py
@@ -297,6 +297,10 @@ def tint_image(img: Image.Image, hex_color: str, mode: str = "overlay") -> Image
         blend_s *= 0.7
         arr = np.array(img.convert("RGB"), dtype=np.float32) / 255.0
         h_px, l_px, s_px = _rgb_to_hls_np(arr)
+        # Remap lightness so dark pixels are pushed up to a visible floor.
+        # Maps L=0 → l_floor and L=1 → 1.0, preserving relative contrast.
+        l_floor = 0.45
+        l_px = l_floor + l_px * (1.0 - l_floor)
         out = _hls_to_rgb_np(
             np.full_like(h_px, blend_h),
             l_px,


theme/stencil/src/components.d.ts
jeffmerrick, line 1: This autogenerated file keeps getting changed, not sure if it needs to be included or should be added to the .gitignore

cnunciato: @jeffmerrick #17956 should fix things up for now (so I think if you rebase/merge with latest `master`, your PR should be clean), but yep — I think @CamSoper has a change in the works that adds this file to the gitignore.

jeffmerrick: @cnunciato Awesome, think that fixed it up!

jeffmerrick:
Thanks @CamSoper!

Just a note that I'm going to spend some time creating images for the first two pages or so before merging.

.claude/commands/blog-meta-image/assets/templates/feature-logo-3.png
cnunciato, line 1: I almost hesistate bring it up because it's so minor, but it's poking at my OCD, so I will 😄 To my eyes at least, the top circle looks _ever so slightly_ off center.

.claude/commands/blog-meta-image/SKILL.md
```diff
@@ -106,7 +115,27 @@ Based on technologies detected in the blog post (AWS, Kubernetes, TypeScript, Do
...
+
+---
+
+### Question 4: Logo tint style
+
+_Only ask if logos were selected in Question 3._
+
+```
+header: "Logo Tint Style"
+question: "How should the tint color be applied to the logos?"
```
cnunciato, line 128: This is likely the fault of the logos we're using, but I tried both options with a faked-up post involving AWS, TypeScript, and Docker, and neither one came out quite right:

Overlay: 

Image

Colorize: 

Image

As a follow-up (or before merge, if it wouldn't be too big a deal), we may want to run through these logos and check 'em all against our new darker backgrounds to make sure they look right. (Just wanted to call it out now to make sure it didn't get lost.)

layouts/blog/single.html
```diff
@@ -18,18 +18,26 @@ <h1 class="no-anchor">{{ .Params.h1 }}</h1>
...
+                                <span>Posted on <time>{{ .Date.Format "Monday, Jan 2, 2006" }}</time></span>
+                            {{ end }}
+                        </div>
+
+                        {{ partial "blog/full-poster.html" . }}
+
+                        <div class="text-sm text-gray-700">
+                           
+                        </div>
+
```
cnunciato: Guessing we can remove this guy:

```suggestion
```

layouts/blog/single.html
```diff
@@ -18,18 +18,26 @@ <h1 class="no-anchor">{{ .Params.h1 }}</h1>
...
-
-                        <div class="flex items-center">
-                            <span class="text-sm text-gray-700">
+                        <div class="flex flex-wrap items-center my-4 text-sm text-gray-700">
+                            <span>
                                 {{ partial "blog/authors.html" (dict "context" . "authors" .Params.authors) }}
                             </span>
                         </div>
+
+                       <div class="text-sm text-gray-700 mb-4">
```
cnunciato:
```suggestion
                        <div class="text-sm text-gray-700 mb-4">
```

.claude/commands/blog-meta-image/SKILL.md
jeffmerrick: @cnunciato Thanks for catching this, the colorize wasn't working as expected.

I've made some updates so the overlay method should work with all the logos we currently have, since the skill can also download logos, it may not work with all future logos. I've also adjusted the colorize method to look slightly better, but it really depends on the logo (it's basically a backup in case overlay doesn't work). So results may not be always great on the 2x and 3x versions since different logos might work better with different methods.

Hopefully we will mostly be designing custom images so it won't come up too much.

**Overlay:**
test-overlay

**Colorize:**
test-color

.claude/commands/blog-meta-image/assets/templates/feature-logo-3.png
jeffmerrick: Thank you for bringing it up! That would've bugged me when I noticed it later. Fixed!

.claude/commands/blog-meta-image/SKILL.md
cnunciato: Awesome. Looks great!

PR #17978, opened 2026-03-14
Remove Google Ads pages from the site

cnunciato:
These are no longer in use, so we can safely remove them.

## Changes
- Removed 31 gads landing pages under `content/gads/`
- Removed 3 gads layout templates and 1 partial (`gads-header.html`)


cnunciato:
@alexleventer No worries — if we're using them, I'll them to the visual refresh and content-review list. (Let me know if there are any we can definitely delete, though — looking over that list, I suspect there are a few we could probably can.)

PR #18004, opened 2026-03-18
Fix pricing page FAQ tabs broken by Tailwind v4

cnunciato:
## Summary
- The FAQ tabs (Pricing, Product, Billing) on the pricing page stopped working after the Tailwind v4 upgrade. Tailwind v4 places utility classes in `@layer utilities`, which always beats `@layer components` where the SCSS `:target` rules live. Inline utilities like `lg:hidden`, `border-b-4`, and `text-gray-500` were overriding the tab-switching CSS.
- Moved all tab-state styling from inline HTML utilities into SCSS using explicit CSS properties (`border-bottom-color`, `display`) to stay within the same layer as the `:target` rules.
- Added a small script to preserve scroll position when clicking FAQ tabs, preventing the page from jumping to the anchor targets.

## Test plan
- [ ] Visit `/pricing/` and scroll down to the FAQ section
- [ ] Click each tab (Pricing, Product, Billing) and verify the correct section appears
- [ ] Verify the active tab has a blue underline and the inactive tabs are gray
- [ ] Verify clicking a tab does not scroll the page
- [ ] Check on mobile that all FAQ sections are visible (no tab filtering on small screens)

### DIFF
--- a/layouts/page/pricing.html
+++ b/layouts/page/pricing.html
@@ -212,23 +212,20 @@
         <span id="faq-billing"></span>
 
         <ul id="faq-headers" class="hidden lg:flex list-none mt-10">
-            <li class="px-10 py-3 mr-3 border-b-4 border-solid border-blue-600" data-faq-label="pricing">
-                <h4><a href="#faq-pricing" class="text-black">Pricing</a></h4>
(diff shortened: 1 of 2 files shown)


CamSoper (approved):
LGTM!

PR #18005, opened 2026-03-18
docs(components): clarify parent option requirement for child resources

### LINKED ISSUE #14717: Components: Clarify that parent must be explicitly set when authoring
### Problem description

The way that resource options are handled by components should be made more explicit, perhaps in https://www.pulumi.com/docs/iac/concepts/resources/components/ . For example, components can automatically pass providers to child resources, but child resources are not automatically passed the `parent` resource option. Since it is possible for the component author to pass resource options from component to child resources in the component definition, the docs should contain guidance on whether this ever makes sense, or how to make it make sense. 

The guidance on which resource options function for the component level could also be made more top level with an explanation of why some resource options work, and some don't.

Thread here https://pulumi.slack.com/archives/C07U860415K/p1743614649568389?thread_ts=1743092770.603079&cid=C07U860415K

### LINKED ISSUE #15319: Please provide more context about why `parent` is required for ComponentResources. Or is it?
### Problem description

In the [Component Resources - Authoring a New Component Resource | Pulumi Docs](https://www.pulumi.com/docs/iac/concepts/resources/components/#authoring-a-new-component-resource) documentation, it says:

> Also inside of its constructor, allocate any child resources, passing the [`parent`](https://www.pulumi.com/docs/concepts/options/parent) option as appropriate to ensure component resource children are parented correctly.

What does "as appropriate" mean here?

Similarly, in the [S3 Folder Pulumi Component | Pulumi Registry](https://www.pulumi.com/registry/packages/aws/how-to-guides/s3-folder-component/) (linked as the `Pulumi Components tutorial` in the Component Resources page), it says:

> When creating a resource within a component, add a `parent` property as the last argument to the constructor.

Should the `parent` property always be added to any resources in a ComponentResource? What are the impacts of leaving it out (spoiler: anything depending on the `componentresource` will not wait for a non-child resource to build).

It'd be lovely to get a bit more information about the importance of the `parent` property when building Component Resources, when it should be added and what impacts it has on the bahaviour of the Component.

Thanks 😊

jkodroff:
Replaces vague "as appropriate" language with a clear explanation of why `parent` must always be set when creating resources inside a component. Adds a warning callout explaining the consequences of omitting parent: broken resource graph, missing provider inheritance, and dependency tracking issues.

Fixes #15319
Fixes #14717

### DIFF
--- a/content/docs/iac/concepts/components/_index.md
+++ b/content/docs/iac/concepts/components/_index.md
@@ -28,7 +28,7 @@ The implicit `pulumi:pulumi:Stack` resource is itself a component resource that 
 
 ## Authoring a New Component Resource
 
-To author a new component, either in a program or for a reusable library, create a subclass of [`ComponentResource`](/docs/reference/pkg/python/pulumi/#pulumi.ComponentResource). Inside of its constructor, chain to the base constructor, passing its type string, name, arguments, and options. Also inside of its constructor, allocate any child resources, passing the [`parent`](/docs/concepts/options/parent) option as appropriate to ensure component resource children are parented correctly.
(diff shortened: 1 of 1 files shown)


jkodroff:
@claude Does this also fix https://github.com/pulumi/docs/issues/14717?

PR #18051, opened 2026-03-19
Add a blog for the new bun runtime

julienp:
<s>Requires CLI 3.227.0 (about to be released) and `@pulumi/pulumi` 3.226.0 (already released)</s>
Edit: everything's been released

Need to first merge:
* Docs: https://github.com/pulumi/docs/pull/17935
* Templates: https://github.com/pulumi/templates/pull/1075

Fixes https://github.com/pulumi/docs/issues/18052


content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md
```diff
@@ -0,0 +1,204 @@
...
+    packagemanager: npm
+```
+
+After:
+
+```yaml
+runtime: bun
+```
+
+When the runtime is set to `bun`, Bun is also used as the package manager — there's no need to configure a separate `packagemanager` option.
```
pierskarsenbarg, line 102: Can we make this a note - to draw attention to it:

```suggestion
{{% notes type="info" %}}
When the runtime is set to `bun`, Bun is also used as the package manager — there's no need to configure a separate `packagemanager` option.
{{% /notes %}}
```

content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md
```diff
@@ -0,0 +1,204 @@
...
+- **[`module: "Preserve"`](https://www.typescriptlang.org/tsconfig/#module)** and **[`moduleResolution: "bundler"`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#moduleresolution-bundler)**: Let Bun handle module resolution instead of compiling to CommonJS. The `bundler` resolution strategy allows extensionless imports while still respecting `package.json` exports, matching how Bun resolves modules in practice.
+- **[`verbatimModuleSyntax: true`](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax)**: Enforces consistent use of ESM `import`/`export` syntax. TypeScript will flag any remaining CommonJS patterns like `require()` at compile time.
+
+### 3. Use ESM syntax
+
+With the Bun runtime, your code should use [ECMAScript module (ESM)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) syntax. ESM uses `import` and `export` statements instead of CommonJS's `require()` and `module.exports`:
+
+```typescript
+// CommonJS (Node.js default)
+const pulumi = require("@pulumi/pulumi");
```
pierskarsenbarg: In our docs, we don't do this: https://www.pulumi.com/registry/packages/aws/api-docs/account/region/#example-usage

content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md
```diff
@@ -0,0 +1,204 @@
...
+// ESM (used with Bun)
+import * as pulumi from "@pulumi/pulumi";
+import * as aws from "@pulumi/aws";
+```
+
+For stack outputs, use `export const` instead of assigning to `module.exports`:
+
+```typescript
+// CommonJS
+module.exports = { bucketName: bucket.id };
```
pierskarsenbarg: We don't do this in our docs: https://www.pulumi.com/docs/iac/concepts/stacks/#outputs

content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md
julienp: I updated the ESM section to talk about `export = ...` instead, a case we actually do have in our documentation.

content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md
```diff
@@ -0,0 +1,218 @@
...
+    - bun
+schema_type: auto
+
+# Optional: Social media promotional copy (for reference only, does not auto-post)
+social:
+    twitter:
+    linkedin:
+---
+
+Last year we added support for [Bun as a package manager](/blog/bun-package-manager/) for Pulumi TypeScript projects. Today we're taking the next step: Bun is now a fully supported runtime for Pulumi programs. Set `runtime: bun` in your `Pulumi.yaml` and Bun will execute your entire Pulumi program, with no Node.js required. This was one of our [most requested features](https://github.com/pulumi/pulumi/issues/13904).
```
cnunciato:
```suggestion
Last year we added support for [Bun as a package manager](/blog/bun-package-manager/) for Pulumi TypeScript projects. Today we're taking the next step: Bun is now a fully supported runtime for Pulumi programs. Set `runtime: bun` in your `Pulumi.yaml` and Bun will execute your entire Pulumi program, with no Node.js required. Since Bun's 1.0 release, this has been one of our [most requested features](https://github.com/pulumi/pulumi/issues/13904).
```

content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Why Bun?
+
+[Bun](https://bun.sh/) is a JavaScript runtime designed as an all-in-one toolkit: runtime, package manager, bundler, and test runner. For Pulumi users, the most relevant advantages are:
+
+- **Native TypeScript support**: Bun runs TypeScript directly without requiring [ts-node](https://typestrong.org/ts-node/) or a separate compile step.
+- **Fast package management**: Bun's built-in package manager can install dependencies significantly faster than npm.
+- **Node.js compatibility**: Bun [aims for 100% Node.js compatibility](https://bun.sh/docs/runtime/nodejs-apis), so the npm packages you already use with Pulumi should work out of the box.
+
+With `runtime: bun`, Pulumi uses Bun for both running your program and managing your packages, giving you a streamlined single-tool experience.
```
cnunciato, line 34: Presumably also much faster startup times, native ESM support for Pulumi programs, all the goodies that come with that, etc.?

content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md
```diff
@@ -0,0 +1,218 @@
...
+
```
cnunciato, line 35: This is all good, but one thing that feels like it's somewhat missing (and could add a bit of color) is an answer to the the question, "What took you so long?" 😄 Looking back at the issue, it seems the missing piece was mainly support for http2, but as a user of both Bun and Pulumi, I'd love to hear more about how that matters and whether we hit any other issues along the way. Not a deep dive of course, just a bit more of the technical story (as it's probably pretty interesting!).

We should also update [Piers's post](https://www.pulumi.com/blog/bun-package-manager/#what-is-this-bun-thing-youre-talking-about) to revise the bit about only supporting package management. I'm happy to take that on if you like — up to you!

content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md
```diff
@@ -0,0 +1,218 @@
...
+```typescript
+import * as aws from "@pulumi/aws";
+
+const azs = await aws.getAvailabilityZones({ state: "available" });
+
+const buckets = azs.names.map(az => new aws.s3.BucketV2(`my-bucket-${az}`));
+
+export const bucketNames = buckets.map(b => b.id);
+```
```
cnunciato: This section threw me for a loop, because most Pulumi TypeScript programs aren't written this way. When I read it, I was confused until I got to the next section, which essentially told me I could ignore the previous one. 😄 

I think what we want to say here instead is that async workarounds are no longer needed because ESM and Bun support top-level await. So I might suggest switching that up a bit (feel free to take or leave any of this by the way — just offering a suggestion):


````suggestion
With [ECMAScript module (ESM)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) syntax, one thing that gets easier is working with async code. In a CommonJS Pulumi program, if you need to await a data source or other async call before declaring resources, the program must be wrapped in an [async entrypoint function](https://github.com/docs/iac/languages-sdks/javascript/#enabling-async-support). With ESM and Bun, [top-level await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#top_level_await) just works, so you can skip the wrapper function entirely and `await` directly at the module level:

```typescript

const azs = await aws.getAvailabilityZones({ state: "available" });

const buckets = azs.names.map(az => new aws.s3.BucketV2(`my-bucket-${az}`));

export const bucketNames = buckets.map(b => b.id);
```

If your existing program does use an async entrypoint with `export =`, just replace it with the ESM-standard `export default`:

```typescript
// CommonJS (Node.js default)
export = async () => {
    const bucket = new aws.s3.BucketV2("my-bucket");
    return { bucketName: bucket.id };
};

// ESM (used with Bun)
export default async () => {
    const bucket = new aws.s3.BucketV2("my-bucket");
    return { bucketName: bucket.id };
};
```
````

File: content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md (hunk 5 above)
Comment (julienp) [reply content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md:34]: > native ESM support for Pulumi programs

Pulumi supports native ESM just fine with Node.js, but we don't default to it.

I don't think the startup time matters much, so I'd rather not talk about better performance when it's unlikely to meaningfully impact the total duration of the operation.

File: content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md (hunk 6 above)
Comment (julienp) [reply content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md:35]: Apparently bun started supporting GRPC on the server side in https://bun.sh/blog/bun-v1.1.31#support-for-node-http2-server-and-grpc in October 18, 2024. So I think the main reason we didn't support it as a runtime earlier was that we didn't get around to it.

File: content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md (hunk 5 above)
Comment (joeduffy) [reply content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md:34]: But is it faster? That's one noticeable difference with TypeScript which is that the compilation step can add 100ms+, sometimes perceptably multiple seconds. I think that's very worthy of noting (and quantifying).

Comment (julienp) [reply content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md:34]: A simple example using a `pulumi preview` against the local backend of a program with a single `@pulumi/random` resource, once with default setup (ts-node), once with Node.js's builtin type stripping, once as plain JS and finally using bun. Used `hyperfine` with 20 runs to benchmark.

Bun is 140ms faster than the default setup here.

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `ts + ts-node` | 713.7 ± 3.3 | 707.4 | 720.1 | 1.24 ± 0.01 |
| `ts + strip` | 753.8 ± 6.1 | 745.9 | 768.2 | 1.31 ± 0.01 |
| `js` | 707.0 ± 4.0 | 701.9 | 716.4 | 1.23 ± 0.01 |
| `ts + bun` | 573.7 ± 4.6 | 565.3 | 585.3 | 1.00 |

**Comment by pgavlin** [comment]
@pierskarsenbarg @joeduffy @cnunciato does this look good to you all? eager to get this out!

File: content/blog/introducing-bun-as-a-runtime-for-pulumi/index.md [hunk 8]
```diff
@@ -0,0 +1,219 @@
...
+        "noUncheckedIndexedAccess": true,
+        "noImplicitOverride": true
+    }
+}
+```
+
+Key differences from a typical Node.js `tsconfig.json`:
+
+- **[`module: "Preserve"`](https://www.typescriptlang.org/tsconfig/#module)** and **[`moduleResolution: "bundler"`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#moduleresolution-bundler)**: Let Bun handle module resolution instead of compiling to CommonJS. The `bundler` resolution strategy allows extensionless imports while still respecting `package.json` exports, matching how Bun resolves modules in practice.
+- **[`verbatimModuleSyntax: true`](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax)**: Enforces consistent use of ESM `import`/`export` syntax. TypeScript will flag any remaining CommonJS patterns like `require()` at compile time.
```
cnunciato:
```suggestion
- [`module: "Preserve"`](https://www.typescriptlang.org/tsconfig/#module) and [`moduleResolution: "bundler"`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#moduleresolution-bundler): Let Bun handle module resolution instead of compiling to CommonJS. The `bundler` resolution strategy allows extensionless imports while still respecting `package.json` exports, matching how Bun resolves modules in practice.
- [`verbatimModuleSyntax: true`](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax): Enforces consistent use of ESM `import`/`export` syntax. TypeScript will flag any remaining CommonJS patterns like `require()` at compile time.
```

cnunciato (approved):

PR #18099, opened 2026-03-20
[Supersonic Page Speed] Dedicated homepage CSS bundle

sicarul:
## Summary

Create a lean homepage-specific CSS bundle that only includes the ~15 SCSS modules the homepage actually uses (out of 70+). The full bundle carries docs, registry, pricing, API reference, and event page styles that are never used on the homepage.

### How it works
- New `homepage.scss` entry point with only homepage-relevant imports (header, footer, hero, buttons, carousel, chooser, swiper, neo-mode, etc.)
- Pages set `css_bundle: homepage` in frontmatter to use it; pages without the field get the default full `bundle.css`
- `assets.html` reads the `css_bundle` param and loads the corresponding CSS file
- Production `minify-css.js` PurgeCSS pipeline processes the homepage bundle alongside the existing ones

### Results
- Homepage CSS: 659 KiB -> 260 KiB in dev (60% smaller)
- After gzip: **30 KiB** vs production's current 61 KiB (**-51%**)
- Other pages are unaffected (they continue to use `bundle.css`)

### Extensibility
Any page can set `css_bundle` in frontmatter to use a page-specific bundle. This pattern can be extended to other high-traffic pages (pricing, product, etc.) in the future.


sicarul:
Replying to Claude's feedback...

**1. PurgeCSS scans all pages** -- Not an issue. PurgeCSS in `minify-css.js` runs on the main `bundle.css`, not the homepage bundle. The homepage CSS uses Tailwind `@source` directives already scoped to homepage templates only.

**2. `marketing/pricing` import** -- Intentional. It styles the "Ready to try Pulumi?" rainbow box at the bottom of the homepage.

**3. No validation of `css_bundle` frontmatter** -- Hugo already fails at build time with a clear error if the resource doesn't exist. No need for extra validation.

**4. `homepage.ts` generates unused JS** -- Standard webpack pattern. A JS entry point is required to produce the CSS bundle. The output JS is effectively empty.

CamSoper (changes requested):
Nice work on the dedicated homepage CSS bundle — the 51% gzip reduction is a great win!

One issue I spotted:

**`infrastructure/index.ts`** — The new CloudFront cache behavior for `/css/homepage.*.css` (lines 723-728) is missing `responseHeadersPolicyId: ImmutableCachePolicy.id`. Both the `bundle.*.css` and `marketing.*.css` behaviors include this policy to set immutable cache headers. Since the homepage bundle also uses content-hash fingerprinting in the filename, it should get the same treatment.

```typescript
{
    ...baseCacheBehavior,
    pathPattern: "/css/homepage.*.css",
    defaultTtl: oneYear,
    maxTtl: oneYear,
    responseHeadersPolicyId: ImmutableCachePolicy.id,  // Add this
},
```

CamSoper:
Looks great, @sicarul!

PR #18124, opened 2026-03-21
fix: preserve IDP homepage image aspect ratio

sicarul:
## Summary
- PR #18097 added `width` and `height` attributes to homepage images for CLS reduction
- The IDP image has a `h-96` class that constrains height to 384px, but without `w-auto` the browser stretches the width to fill the container, distorting the aspect ratio
- Adds `w-auto` so the width scales proportionally with the fixed height


CamSoper (approved):
LGTM! Clean fix — `w-auto` is the correct approach to preserve aspect ratio when `h-96` and explicit width/height attributes are both set.

PR #18126, opened 2026-03-21
[Supersonic Page Speed] Lazy-load consent manager and GitHub buttons

sicarul:
## Summary

- Lazy-load `consent-manager.js` (180KB) after page `load` event instead of `defer`
- Lazy-load GitHub `buttons.js` after page `load` event instead of `async defer`

## Why

`defer` scripts download during HTML parsing and compete for bandwidth with critical resources (CSS, fonts, main JS bundle). They also block the `DOMContentLoaded` event. By moving these to post-`load` lazy loading, the browser focuses on rendering the page first, then downloads these non-critical scripts after everything else is done.

## Impact

- **Consent manager**: 180KB moved off the critical path. Segment buffers analytics events until the consent manager loads, so there's no compliance risk. For non-EU users (`shouldRequireConsent: exports.inEU`), the banner never shows anyway. For EU users, the banner appears ~200-500ms later.
- **GitHub buttons**: Decorative star/fork count badges, not needed for page render.

## Risk

Low. Both scripts are non-critical for initial page functionality. The consent manager's event buffering ensures no tracking fires before consent is given.

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -190,7 +190,7 @@
     <!--Segment tracking-->
     <script>
         // Do not load Segment for these user agents.
-        userAgentBlacklist = [
+        userAgentBlocklist = [
             'Mozilla/5.0 (compatible; SiteAuditBot/0.97; +http://www.semrush.com/bot.html)',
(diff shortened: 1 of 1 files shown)


CamSoper (approved):
LGTM! Clean performance optimization — moving non-critical scripts off the critical path is a solid win. The lazy-load pattern is correct, and the blocklist rename is a nice inclusive language improvement.

PR #18139, opened 2026-03-21
[Supersonic Page Speed] Fix header CLS on slow connections

sicarul:
## Summary

- Make `.header-container` sticky by default so it doesn't flash/blink on mid-page refresh (previously `sticky` was only applied via JS after IntersectionObserver fired)
- Add pre-hydration CSS to prevent layout shifts before Stencil and third-party scripts load:
  - Hide the `signed-in` slot so only "Sign In" shows (prevents "Pulumi Cloud Sign In" overflow)
  - Hide the raw GitHub "Star" anchor before `buttons.github.io` replaces it with an iframe
- Add `flex items-center` to the sign-in and Neo links to prevent icon/text stacking before hydration
- Normalize icon margins across all top nav items: replace whitespace-based spacing with consistent `mr-1`

## How to test

1. Open Chrome DevTools (F12)
1. Go to the **Network** tab
1. Open the **Throttling** dropdown (defaults to "No throttling")
1. Select **Slow 4G** (or create a custom profile with lower speeds)
1. Check **Disable cache**
1. Reload the page and watch the header area -- the top nav bar should not change height or flash during load

**Before:** On slow connections, the header blinks into position on scroll, the GitHub widget shows raw "Star" text, and the sign-in area shows "Pulumi Cloud Sign In" stacked vertically before snapping inline.

**After:** Header is sticky from first paint, GitHub button is hidden until the widget loads, and sign-in renders as a single inline "Sign In" immediately.


sicarul:
Thanks @claude - both items are intentional:

1. **Neo icon margin** - The SVG icon has its own built-in spacing. Adding `mr-1` creates inconsistent gaps compared to the other nav items.
2. **Dropdown menu margins** - Intentional tightening for visual consistency. Verified it looks correct.

No changes needed.

CamSoper (approved):
Looks good! Clean CLS fix — making the header sticky by default and hiding pre-hydration elements is the right approach. The margin normalization to `mr-1` is a nice cleanup too. 👍

PR #18147, opened 2026-03-22
[Supersonic Page Speed] Replace Swiper library with lightweight carousel and enable chunk splitting

sicarul:
## Summary

- Replace the 77KB Swiper library with a custom ~8KB carousel (2KB gzipped), cutting the largest Stencil dependency
- Enable webpack chunk splitting with content-hash filenames
- Fix long tweet text overflow on homepage cards

## Carousel replacement

The `pulumi-swiper` Stencil component previously imported `SwiperCore` with the Autoplay and Navigation modules. Every usage on the site is `loop="true"` with autoplay, and only the about page uses prev/next nav buttons. The full Swiper bundle is unnecessary for this.

The new implementation uses CSS `translateX` transitions on the wrapper div. For infinite loop, it clones the first and last N slides (where N = slidesPerView) and instantly jumps to the real slide on `transitionend` at clone boundaries. Autoplay uses `setInterval`, cleared on mouseenter and restarted on mouseleave. Touch/drag swiping uses pointer events with a 50px threshold to advance or snap back. Interactive elements (links, buttons, nav controls) bypass the drag handler so clicks pass through normally. The `@Prop()` interface is unchanged, so zero layout file changes were needed.

## Chunk splitting

Webpack now outputs to `static/js/` with `contenthash` chunk filenames instead of going through Hugo's `resources.Get` + `fingerprint` pipeline. This lets Stencil lazy-load component chunks directly and enables proper long-term caching via content-addressed filenames.

## Tweet overflow fix

Desktop and mobile tweet cards on the homepage now use `line-clamp-5` to truncate long text with an ellipsis instead of letting it overflow the fixed-height card.

### DIFF
--- a/scripts/minify-css.js
+++ b/scripts/minify-css.js
@@ -14,6 +14,7 @@ const sharedSafelist = [
     /^carousel/,
     /^st-/,
     /^pulumi-/,
+    /^swiper/,
 ];
 
 // Per-bundle PurgeCSS configuration.
--- a/theme/src/scss/components/_swiper.scss
+++ b/theme/src/scss/components/_swiper.scss
@@ -88,8 +88,7 @@
 
 pulumi-swiper {
     .swiper {
-        @apply w-full h-full overflow-visible;
-        overflow: visible !important;
+        @apply w-full h-full;
     }
 
     .swiper-slide {
--- a/theme/stencil/src/components/swiper/swiper.tsx
+++ b/theme/stencil/src/components/swiper/swiper.tsx
@@ -86,6 +86,7 @@ export class Swiper {
 
         this.resizeObserver = new ResizeObserver(() => {
             this.applySlideWidths();
+            this.isTransitioning = false;
             this.jumpTo(this.currentIndex);
         });
         this.resizeObserver.observe(this.container);


sicarul:
@CamSoper sorry i thought i tested properly on mobile, will check!

sicarul:
## Fix: Carousel broken in production builds

**Root cause**: The production CSS post-processing step (`scripts/minify-css.js`) runs PurgeCSS, which strips CSS selectors not found in the scanned content files. PurgeCSS scans `public/**/*.html` and `public/js/bundle.*.js` for class usage, but the `.swiper`, `.swiper-wrapper`, and `.swiper-slide` classes are only present in Stencil component chunks (`chunk-*.js`), which aren't included in the scan. PurgeCSS strips the base styles, including the critical `.swiper-wrapper { display: flex }` that makes the carousel work.

This is why the carousel worked locally (`make serve-all` skips PurgeCSS) but was broken in the PR preview (production build runs PurgeCSS).

The existing `/^pulumi-/` safelist entry preserved `pulumi-swiper .swiper` (the scoped override) but not the bare `.swiper-wrapper` base rule.

**Fixes in 8d25668:**

1. **PurgeCSS safelist** (`scripts/minify-css.js`) - Added `/^swiper/` to preserve all swiper-prefixed class selectors through the purge step
2. **Removed `overflow: visible !important`** (`_swiper.scss`) - This was a workaround for the old Swiper library. The new carousel needs the base `overflow: hidden` on `.swiper` to clip off-screen slides
3. **ResizeObserver stuck state** (`swiper.tsx`) - Reset `isTransitioning` in the ResizeObserver callback to prevent the carousel from freezing if a resize fires mid-transition

Verified locally with `make build` + `make serve-static` that the production CSS now contains the base swiper rules and the carousel renders correctly.

Issue #18149, opened 2026-03-23
Add "Accept All" to simplify our cookie consent form


joeduffy:
I am traveling in Europe and, I don't know if I'm doing something wrong, but our cookie consent form seems more complicated than is typical. It forces me to manually approve every category of cookie, and lacks a simple Accept/Reject All button.

For example, notice we only have "Manage Cookies" which pops the detailed selector as shown:

<img width="1316" height="901" alt="Image" src="https://github.com/user-attachments/assets/5779e2c1-358b-463f-a7e7-a6442ecce991" />

In comparing to other sites, our approach seems to introduce far more friction for the average visitor. See Claude for instance, which, per European laws, still allows you to customize the specific selections, but also offers a much simpler "Accept All":

<img width="976" height="664" alt="Image" src="https://github.com/user-attachments/assets/4ecd116d-ef16-4cd9-9d13-aeada768334b" />

Most sites I look at offer a simple option like Claude does.

When I look at our consent form configuration in HubSpot, it even explicitly sets the Accept All button text, but I don't think we are using HubSpot any longer for this. It seems we have some [custom Segment frontend rendering code to control this](https://github.com/pulumi/docs/blob/master/layouts/partials/head.html#L262), and I am unsure how it even relates, if at all. Claude Code claims fixing this is as simple as:

```diff
+            var bannerActionsBlock = function(props) {
+                return React.createElement(
+                    'div',
+                    { class: "consent-banner-actions" },
+                    React.createElement(
+                        'button',
+                        { class: "accept-all-btn", type: "button", onClick: props.acceptAll },
+                        'Accept All'
+                    ),
+                    React.createElement(
+                        'button',
+                        { class: "manage-cookies-btn", type: "button", onClick: props.changePreferences },
+                        'Manage cookies'
+                    )
+                )
+            }
```

But I'm not as familiar with how we're using Segment, and why we prefer it over HubSpot, here to know if that'll fix things for sure.

PR #18172, opened 2026-03-23
feat: add Accept All / Reject All buttons to cookie consent banner

sicarul:
fixes https://github.com/pulumi/docs/issues/18149

- Add explicit Accept All and Reject All buttons alongside Manage cookies
- Defer analytics.load() to consent manager so tracking only starts after user consent
- Hide close button since users now have explicit accept/reject actions
- Update banner styling: centered text, violet brand colors, secondary style for Manage/Reject

image

### DIFF
--- a/theme/src/scss/_consent-banner.scss
+++ b/theme/src/scss/_consent-banner.scss
@@ -40,10 +40,6 @@
         @apply font-bold no-underline;
     }
 
-    button[aria-label="Close"] {
-        @apply text-violet-800 mt-4 top-0 right-0 text-lg;
-    }
-
     .consent-actions {
         @apply flex gap-2 w-full pb-4;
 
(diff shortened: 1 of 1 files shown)


CamSoper (approved):
Looks good! Verified via Playwright with EU timezone emulation and production Segment key:

- **Accept All**: Dismisses banner, sets all tracking preferences to `true`, initializes analytics
- **Reject All**: Dismisses banner, sets all preferences to `false`, analytics does NOT load
- **Manage cookies**: Opens preferences dialog with per-category toggles
- **Close button**: Hidden as intended

I pushed a small commit adding `focus-visible` styles to the consent buttons for keyboard accessibility and removing the dead `button[aria-label="Close"]` CSS rule.

Note: The banner won't render on the test deployment because the dev Segment write key has zero integrations configured — this is expected behavior from the consent manager library, not a bug.

PR #18222, opened 2026-03-26
Defer GrowthBook SDK init to break critical request chain

alexleventer:
## Summary

- Defers GrowthBook SDK initialization from eager module-level execution to a `window.load` callback
- **Before**: importing `util.ts` (used by multiple Stencil components) immediately created a `GrowthBook` instance and called `gb.init({ streaming: true })`, which fetches from `cdn.growthbook.io` — creating a critical request chain: `HTML → JS bundle → GrowthBook API` (635ms critical path latency)
- **After**: GrowthBook is initialized only after the page finishes loading, breaking the chain entirely
- GrowthBook is only used for experiment tracking (not rendering decisions), so deferring it has no visual impact
- The `gb` export was never imported by any other file — it was purely a side effect in `util.ts`

## Test plan

- [ ] Verify GrowthBook experiment tracking still fires (check network tab for `cdn.growthbook.io` request after page load)
- [ ] Verify Segment `"Experiment Viewed"` events still appear when experiments are active
- [ ] Verify no console errors on page load
- [ ] Re-run PageSpeed Insights to confirm the "Network dependency tree" critical chain is resolved

### DIFF
--- a/theme/stencil/src/util/util.ts
+++ b/theme/stencil/src/util/util.ts
@@ -1,41 +1,56 @@
 import * as uuid from "uuid";
 import { GrowthBook } from "@growthbook/growthbook";
 
-// Initialize GrowthBook
-export const gb = new GrowthBook({
-    apiHost: "https://cdn.growthbook.io",
-    clientKey: (window as any).growthbook_sdk_key,
-    decryptionKey: (window as any).growthbook_decrypt_key,
-    enableDevMode: (window as any).growthbook_dev_mode,
-});
+// Defer GrowthBook initialization until after page load to avoid a critical
+// request chain (HTML → JS → cdn.growthbook.io fetch) that blocks LCP.
+// GrowthBook is used for experiment tracking, not rendering, so deferring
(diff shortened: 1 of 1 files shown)


CamSoper (changes requested):
The `gb` export removal breaks two downstream consumers that still import it:

- `theme/src/ts/experiments/cta-activations-direct-vs-docs.ts` — `import { gb }` + calls `gb.setRenderer()`, `gb.isOn()`
- `theme/src/ts/experiments/terraform-compare.ts` — `import { gb }` + calls `gb.setRenderer()`, `gb.isOn()`

This will cause a build failure (missing export) and, even if re-exported, a runtime crash since `gb` is `null` until `window.load` fires while both experiment files call `gb.setRenderer()` at module execution time.

A couple of options:
- Export a `getGrowthBook()` accessor and have experiment files defer their setup to `window.load` as well
- Register the experiment renderers inside `initGrowthBook()` itself

Also a minor nit: trailing whitespace on line 54.

PR #18251, opened 2026-03-27
Fix case-sensitive .NET SDK docs URLs

### LINKED ISSUE #18238: Fix case-sensitive .NET SDK docs URLs
## Problem

The .NET SDK docs at `/docs/reference/pkg/dotnet/Pulumi/Pulumi.html` are case-sensitive. Requesting the URL with different casing (e.g., `/docs/reference/pkg/dotnet/pulumi/pulumi.html`) returns a 404.

This happens because:
- S3 keys are case-sensitive
- CloudFront passes URIs as-is to S3
- DocFX generates files with PascalCase names matching C# namespaces (`Pulumi/`, `Pulumi.Automation/`, `Pulumi.FSharp/`)

## Proposed fix

1. **CloudFront Function** (`infrastructure/cloudfrontFunctions.ts`) — extend the existing markdown negotiation function to lowercase incoming URIs for `/docs/reference/pkg/dotnet/*`
2. **Post-processing in `scripts/run_docfx.sh`** — after DocFX generates files, lowercase all filenames and rewrite internal `href` attributes so S3 stores lowercase keys
3. **Lambda@Edge regex** (`infrastructure/cloudfrontLambdaAssociations.ts`) — add `/i` flag to the `dotnetSDKRedirect` regex since it will now see lowercased URIs
4. **One-time rename** of the ~204 committed files in `static-prebuilt/docs/reference/pkg/dotnet/`

Old PascalCase URLs (bookmarks, search engines, external links) continue to work transparently via the CloudFront Function rewrite.

## Key details

- Fragment anchors (`#Pulumi_Alias_Urn`) must NOT be lowercased — they're client-side and must match `id` attributes
- GNU `sed` `\L` is available on the Ubuntu GitHub Actions runners
- The `xrefmap.yml` href values also need lowercasing
- After deploy, a CloudFront invalidation for `/docs/reference/pkg/dotnet/*` is needed

CamSoper:
Lowercase all DocFX-generated filenames and add a CloudFront Function to
normalize incoming URIs, so `/docs/reference/pkg/dotnet/` URLs resolve
regardless of casing. Update Lambda@Edge redirect regex to match the
now-lowercase paths.

## Changes
- Add CloudFront Function to lowercase URIs for `/docs/reference/pkg/dotnet/*`
- Add ordered cache behavior to attach the function at viewer-request
- Update `dotnetSDKRedirect` Lambda@Edge regex to match lowercase URIs
- Add post-processing to `run_docfx.sh` to lowercase DocFX output
- Rename ~177 static-prebuilt HTML files and directories to lowercase
- Rewrite internal hrefs, xrefmap.yml, and manifest.json paths

Fixes #18238

### DIFF
--- a/scripts/run_docfx.sh
+++ b/scripts/run_docfx.sh
@@ -49,7 +49,7 @@ done
 #    but NOT the fragment (after #). Fragment anchors must stay PascalCase
 #    because they must match id= attributes in the HTML.
 find "$DOTNET_OUT" -type f -name '*.html' -exec sed -i -E \
-    's/href="([^"#]+)(#[^"]*)?"/href="\L\1\E\2"/g' {} +
+    's/href="([^"#:]+)(#[^"]*)?"/href="\L\1\E\2"/g' {} +
 
 # 4. Rewrite xrefmap.yml: lowercase the path portion of href values (before #).
 sed -i -E 's/^(  href: )([^#]+)(#.*)?$/\1\L\2\E\3/' "$DOTNET_OUT/xrefmap.yml"


sicarul:
@CamSoper let's wait for yoda too but i'm checking it!

sicarul (approved):
Approved your PR is

PR #18253, opened 2026-03-27
Fix oversized SVG icons in LLM menu dropdown

CamSoper:
The `.llm-menu-icon` CSS class set `width` but not `height`, causing inline SVG icons (link, ChatGPT, Claude logos) to expand to fill their flex container while FontAwesome icons rendered correctly.

Adds `height: 1.25rem` to match the existing width constraint.

## Before / After
| Before | After |
|--------|-------|
| SVG icons fill full container height | All icons render at 20x20px |


sicarul (approved):
LGTM

PR #18264, opened 2026-03-30
Fix Event schema issues reported by GWT

alexleventer:
## Summary
- Default missing `location` to "virtual" in event schema markup so all event pages emit the required `location` field — fixes 38 pages flagged as invalid by Google Webmaster Tools
- Skip Event schema on the `/events/` list page (only apply to individual event pages) to avoid invalid Event markup on a non-event page

## Test plan
- [x] Verify event single pages include `VirtualLocation` in JSON-LD when no `location` is set in frontmatter
- [x] Verify `/events/` list page no longer emits Event schema
- [x] Verify events with explicit `location` (e.g. "virtual", "New York, NY") still render correctly

### DIFF
--- a/layouts/partials/schema/collectors/event-entity.html
+++ b/layouts/partials/schema/collectors/event-entity.html
@@ -24,28 +24,26 @@
   {{ $schema = merge $schema (dict "duration" .) }}
 {{ end }}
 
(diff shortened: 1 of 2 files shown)


sicarul (approved):
Lgtm

PR #18267, opened 2026-03-30
Add template deployment settings documentation

fnune:
Add template deployment settings and central registry URL reference.

- Create `/docs/reference/registry-urls/` as the central reference for the `registry://` URL scheme
- Update organization-templates.md to show registry URL format with link to reference
- Update deployments REST API with template sourceContext documentation linking to reference

## Test plan

- `make lint` passes
- Preview pages:
  - `/docs/reference/registry-urls/`
  - `/docs/idp/concepts/organization-templates/`
  - `/docs/reference/cloud-rest-api/deployments/`

### DIFF
--- a/content/docs/reference/registry-urls.md
+++ b/content/docs/reference/registry-urls.md
@@ -7,6 +7,7 @@ meta_image: /images/docs/meta-images/docs-meta.png
 menu:
     reference:
         name: Registry URLs
+        parent: reference-home
         weight: 5
 ---
 


content/docs/reference/registry-urls.md
borisschlosser, line 1: This produces a top level navigation entry in the left nav bar. Might be better suited under `/docs/reference/cloud-rest-api/deployments/#schema-definitions`

fnune: This isn't specific to deployments. Registry URLs are a core concept of Pulumi at this point, they're accepted e.g. by `pulumi new` and other things.

It being in its own tab is a bug though, I'll see why that's happening. It should be under Reference.

PR #18282, opened 2026-03-30
Increase static asset cache TTLs to 30 days

alexleventer:
## Summary

- Increases CloudFront edge TTLs and browser `Cache-Control` max-age for `/icons/*`, `/logos/brand/*`, and `/logos/*` from 30m–1h to **30 days** (2,592,000 seconds)
- Renames `BrandLogoCachePolicy` → `StaticAssetCachePolicy` and consolidates all three path patterns under it
- Unifies `/logos/*` onto the same policy as `/icons/*` and `/logos/brand/*` (previously used `OneHourCachePolicy`)

## Motivation

Google's performance guidance states:

> All cacheable subresource requests should have a cache lifetime of at least 30 days (2,592,000 seconds). Cacheable resources should have a very long lifetime (30 days or 1 year).

Our SVG icons and logos (`pdi-neo.svg`, `logo-on-black.svg`, customer logos, social icons, etc.) were being served with cache lifetimes of only 30 minutes to 1 hour — well below Google's recommended minimum. These are static assets that change very rarely, so short cache times cause unnecessary revalidation round-trips and hurt page performance scores.

## Changes

| Path pattern | CDN edge TTL (before → after) | Browser Cache-Control (before → after) |
|---|---|---|
| `/icons/*` | 1h → **30 days** | `max-age=1800` (30m) → `max-age=2592000` (30d) |
| `/logos/brand/*` | 30m → **30 days** | `max-age=1800` (30m) → `max-age=2592000` (30d) |
| `/logos/*` | 1h → **30 days** | `max-age=3600` (1h) → `max-age=2592000` (30d) |

Since these files are not fingerprinted (no content hash in filenames), we use 30 days rather than 1 year/immutable. If an SVG needs an urgent update, a CloudFront cache invalidation will flush it immediately.

## Test plan

- [ ] Verify `pulumi preview` shows expected CloudFront distribution and response headers policy updates
- [ ] After deploy, confirm `Cache-Control: public, max-age=2592000` header on an SVG request (e.g., `curl -I https://www.pulumi.com/icons/pdi-neo.svg`)

### DIFF
--- a/BUILD-AND-DEPLOY.md
+++ b/BUILD-AND-DEPLOY.md
@@ -1598,9 +1598,9 @@ Delivery: CloudWatch Logs infrastructure v2
 | /ai/* | S3 Main | 1 week | 410 Gone (Lambda@Edge) |
 | /uploads/* | Uploads | 1 hour | User uploads |
 | /fonts/* | S3 Main | 1 hour | Web fonts |
-| /icons/* | S3 Main | 30 days | Icons |
-| /logos/brand/* | S3 Main | 30 days | Brand logos |
-| /logos/* | S3 Main | 30 days | Logos |
+| /icons/* | S3 Main | 1 hour | Icons |
+| /logos/brand/* | S3 Main | 30 minutes | Brand logos |
+| /logos/* | S3 Main | 1 hour | Logos |
 | /fingerprinted/* | S3 Main | 1 year (immutable) | Content-hashed assets |
 | *.woff,*.woff2 | S3 Main | 1 year | Font files |
 


sicarul:
Commented on slack i think this is a bit risky if we need to replace a logo quickly, we'll need to rename it. Better if we can use fingerprints for the remaining logos like we did on https://github.com/pulumi/docs/pull/18183

alexleventer:
@claude review this PR

infrastructure/index.ts
```diff
@@ -571,16 +573,15 @@ const CopilotSecurityHeadersPolicy = new aws.cloudfront.ResponseHeadersPolicy('c
     },
 });
 
-// Fingerprinted/hashed assets get immutable browser caching (1 year).
-// This is separate from CloudFront edge TTLs (defaultTtl/maxTtl) which only
-// control CDN-level caching. Without this policy, browsers see no Cache-Control
-// header and fall back to heuristic caching with 304 revalidation round-trips.
-const BrandLogoCachePolicy = new aws.cloudfront.ResponseHeadersPolicy('brand-logo-cache-headers', {
+// Static assets (icons, logos, SVGs) get 30-day browser caching to meet
```
sicarul: @alexleventer i don't think we should do this as the default - for fingerprinted we can do a year, but i still think it's dangerous for non fingerprinted assets to do 30 days, we can get ourselves into a mess

sicarul: This is the only issue i'd fix, the rest of the PR LGTM

PR #18290, opened 2026-03-31
fix: reduce CloudFront error cache TTL from 5min to 10s

sicarul:
## Summary
- Reduces `errorCachingMinTtl` for 404 and 403 custom error responses from 5 minutes (300s) to 10 seconds
- During deploys, when CloudFront switches its origin to a new S3 bucket, transient 404s can get cached at edge locations. With a 5-minute error cache TTL, the homepage stays down until the cache expires. 10 seconds limits the blast radius of these transient errors.

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -1034,7 +1034,7 @@ const distributionArgs: aws.cloudfront.DistributionArgs = {
         {
             errorCode: 404,
             responseCode: 404,
-            errorCachingMinTtl: fiveMinutes,
+            errorCachingMinTtl: 10,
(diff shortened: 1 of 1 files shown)


CamSoper (approved):
LGTM! Clean, well-motivated change. Reducing the error cache TTL from 5 minutes to 10 seconds will significantly limit the blast radius of transient 404s during origin switches.

Minor optional suggestion: consider defining `const tenSeconds = 10;` for consistency with the other named TTL constants (`fiveMinutes`, `tenMinutes`, etc.) — but not a blocker.

PR #18343, opened 2026-04-02
Update name and title for Sabrina

isaac-pulumi:
Sabrina's promotion to Partner and name change (congrats Sabrina!).


data/team/board/sabrina-wu.toml
isaac-pulumi:
```suggestion
linkedin = "sabrina-albert"
```

PR #18353, opened 2026-04-02
Add social media review CI workflow

adamgordonbell:
Adds a separate Claude-powered review for social media copy on blog post PRs. Reuses schedule-posts.py with a new --check mode that queries S3 state to skip already-posted content, then feeds the social copy into Claude for tone/compliance review.

Also softens the general docs review criteria so missing social blocks are informational, not merge blockers.

### DIFF
--- a/BUILD-AND-DEPLOY.md
+++ b/BUILD-AND-DEPLOY.md
@@ -1404,7 +1404,7 @@ The repository uses 24 GitHub Actions workflows organized into categories. All w
 
 ### Other Workflows
 
-The repository includes 9 additional utility workflows for automation and project management:
+The repository includes 10 additional utility workflows for automation and project management:
 
 **Automation and Auto-merge:**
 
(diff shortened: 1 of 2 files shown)


.claude/commands/social-media-review.md
```diff
@@ -0,0 +1,105 @@
...
+
+# Review Social Post
+
+## Intent
+
+You are Claude, acting as a strict reviewer protecting Pulumi’s brand.
+
+Your role is to prevent low-quality or non-compliant social posts from being published, while helping authors understand how posts should be structured.
+
+These posts are published from Pulumi’s corporate social accounts, not from individual authors. Even if the blog post was written by a guest or community member, the social copy represents Pulumi’s voice.
```
adamgordonbell, line 14: I assume we will have to iterate on this prompt, but I tested it on some of our previous social media posts and it should catch a number of issues.

CamSoper (changes requested):
Nice approach — dedicated CI for social review is a solid idea, and the check mode logic is clean.

Two issues to address:

**Bug: `--check` will crash at runtime**

`CHECK_MODE` is set via `"--check" in sys.argv` at module level (line 83), but `--check` is never registered with the argparse parser. When CI runs `uv run scripts/social/schedule-posts.py --check`, `parser.parse_args()` at line 979 will exit with `error: unrecognized arguments: --check` before `main()` is ever called.

Fix: add `--check` to the argparse parser, or handle check mode before `parse_args()` runs.

**Missing BUILD-AND-DEPLOY.md update**

The new `claude-social-review.yml` workflow should be documented in the "Other Workflows" → "AI-Assisted Development" section of BUILD-AND-DEPLOY.md (alongside `claude.yml` and `claude-code-review.yml`), and the workflow count note updated.

PR #18374, opened 2026-04-03
Decouple preview-link and Lighthouse PR comments

cnunciato:
### Proposed changes

The preview-link comment (`sync-and-test-bucket.sh`) and Lighthouse comment (`run-lighthouse-pr.sh`) were sharing a single PR comment via the `<!-- lighthouse-report -->` HTML marker. This meant skipping Lighthouse on content-only PRs (#18371) left orphaned markers and made the two scripts implicitly dependent on each other.

This PR makes four changes:

- **Decouple the PR comments.** Each script now manages its own independent comment with its own marker (`<!-- preview-link -->` for previews, `<!-- lighthouse-report -->` for Lighthouse). Both follow the same find-by-marker/PATCH-or-POST pattern, so ordering and conditional skipping no longer matter.
- **Use GitHub API for changed-files detection.** Replace the fragile `git fetch`/`git diff` approach with `gh pr diff --name-only`, which queries GitHub directly and doesn't depend on clone depth or local ref state. The previous approach silently produced empty output in CI's shallow checkout, causing Lighthouse to be skipped even when performance-relevant files changed.
- **Only update comments authored by pulumi-bot.** Other bots (e.g., Claude) can quote HTML markers in review comments, causing the scripts to match and overwrite the wrong comment. Both scripts now check `.user.login == "pulumi-bot"` before updating.
- **Update BUILD-AND-DEPLOY.md** to reflect that Lighthouse results are now posted as a separate comment.

### Related issues (optional)

Follow-up to #18371.

### DIFF
--- a/scripts/run-lighthouse-pr.sh
+++ b/scripts/run-lighthouse-pr.sh
@@ -16,7 +16,7 @@ if [[ ! -f "$metadata_file" ]]; then
 fi
 
 base_url="$(jq -r '.url' "$metadata_file")"
-commit_sha="$(jq -r '.commit' "$metadata_file" | cut -c1-8)"
+commit_sha="$(jq -r '.commit' "$metadata_file" | cut -c1-7)"
 
 if [[ -z "$base_url" || "$base_url" == "null" ]]; then
     echo "No preview URL found in metadata. Skipping Lighthouse."


cnunciato:
Interesting: Because Claude's review contained the HTML marker, the script found and updated _it_ rather than generate a new one:

image

We should probably limit these to comments submitted by pulumi-bot specifically.

cnunciato:
Adding the `do-not-merge` while I do some validation on this.

PR #18375, opened 2026-04-04
Prune orphaned dep from yarn.lock, ignore unused package-lock

cnunciato:
Noticed I kept getting `yarn.lock` diffs after running `make ensure`. This seems to be why!

---

## Summary

- Removes an orphaned `dom-mutator@^0.6.0` entry from `theme/yarn.lock` that was left behind when GrowthBook was removed in #18319. This stale entry caused Yarn to prune it on every `yarn install`, producing a local diff after `make ensure`/`make build`/`make serve`.
- Removes `theme/package-lock.json` from git and adds it to `.gitignore`. The repo uses Yarn exclusively — no Makefile target, CI workflow, or script runs `npm install` in `theme/`. The root `package-lock.json` was already gitignored for this reason, but `theme/` was missed, so Dependabot has been updating a lockfile nothing consumes.

## Test plan

- [x] Run `make ensure` and verify `git diff` shows no changes to lockfiles
- [x] Run `make build` and verify the same

### DIFF
--- a/scripts/run-lighthouse-pr.sh
+++ b/scripts/run-lighthouse-pr.sh
@@ -84,7 +84,7 @@ comment_body=""
 add_line() { comment_body+="$1"$'\n'; }
 add_line "## Lighthouse Performance Report"
 add_line ""
-add_line "Commit: \`${commit_sha}\` | [Metric definitions](https://web.dev/articles/vitals)"
+add_line "Commit: ${commit_sha} | [Metric definitions](https://web.dev/articles/vitals)"
 add_line ""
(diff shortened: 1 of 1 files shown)


scripts/run-lighthouse-pr.sh
```diff
@@ -84,7 +84,7 @@ comment_body=""
 add_line() { comment_body+="$1"$'\n'; }
 add_line "## Lighthouse Performance Report"
 add_line ""
-add_line "Commit: \`${commit_sha}\` | [Metric definitions](https://web.dev/articles/vitals)"
+add_line "Commit: ${commit_sha} | [Metric definitions](https://web.dev/articles/vitals)"
```
cnunciato, line 87: Sneaking this in because I noticed (on the Lighthouse comment generated for this PR) that with the backticks in place, GitHub fails to recognize the SHA, so skips linking it. (Seems to read the backticks as an "escape".)

Before and after:

image

image

PR #18468, opened 2026-04-10
fix(infra): enable CloudFront gzip compression for /registry/* and /guides/*

CamSoper:
The `/registry/*`, `/fingerprinted/logos/pkg/*`, and `/guides/*` CloudFront cache behaviors attached the `AllViewerExceptHostHeader` managed origin request policy, which forwards `Accept-Encoding` to the origin. CloudFront silently disables its own auto-compression whenever `Accept-Encoding` is forwarded — so `compress: true` was a no-op and these paths shipped uncompressed (~5-10x wire size on every cache miss).

This drops `originRequestPolicyId` from the three affected behaviors so CloudFront falls back to its default minimal-headers forwarding and re-enables gzip. Comments added so the policy isn't reintroduced.

## Affected paths
- `/registry/*` (HTML, nav JSON, sitemaps, schemas <10 MB)
- `/fingerprinted/logos/pkg/*` (registry SVG logos)
- `/guides/*`

## Test plan
- [ ] Deploy to `www-testing`, confirm `content-encoding: gzip` on representative `/registry/*` and `/guides/*` URLs after the 30-min TTL or a manual invalidation
- [ ] Diff rendered bodies before/after to confirm nothing downstream depended on the forwarded headers
- [ ] After prod deploy, capture before/after wire sizes for `/registry/packages/aws/api-docs/s3/bucket/`, `/registry/packages/navs/aws.json`, `/registry/packages/random/schema.json`

Out of scope: oversized schemas (>10 MB CloudFront auto-compress cap) — handled separately in `pulumi/registry` via the pre-gzip-at-upload pattern from pulumi/registry#10628.

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -508,9 +508,14 @@ const oneHour = fiveMinutes * 12;
 const oneWeek = oneHour * 24 * 7;
 const oneYear = oneWeek * 52;
 
-// AllViewerExceptHostHeader passes all cookies, querystrings, and headers except the Host header.
+// UserAgentRefererHeaders forwards only User-Agent and Referer to the origin.
+// Used by the chained-CDN behaviors (/registry/*, /guides/*) so the inner CDNs
+// still see useful viewer signals for their own logs/observability, while NOT
+// forwarding Accept-Encoding. Forwarding Accept-Encoding silently disables
+// CloudFront auto-compression at the outer layer (compress: true becomes a
+// no-op), which is what was causing /registry/* responses to ship uncompressed.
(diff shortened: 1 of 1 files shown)


CamSoper:
Closing in favor of #18425 — Pablo independently arrived at the same root-cause fix (cache-policy `enableAcceptEncodingGzip`/`enableAcceptEncodingBrotli` flags) and his refactor is strictly broader (Brotli for the entire site, not just /registry/*).

Empirical confirmation that the cache-policy fix works on a live CloudFront distribution posted at #18425 (comment).

PR #18473, opened 2026-04-10
fix: improve Organization schema.org best practices

alexleventer:
## Summary

Audited the homepage Organization structured data against Google and schema.org best practices and made the following fixes:

### Schema improvements in `graph-builder.html`

- **`@type` broadened to `["Organization", "Corporation"]`** — the graph-builder was using only `"Corporation"`, which is a subtype of `Organization`. Search engines and AI systems that only recognize `Organization` would miss it. Using both ensures maximum compatibility.
- **Fixed invalid `AwardOrHonor` schema type** — `AwardOrHonor` is not a valid schema.org type. Replaced the object with a plain text string (`"SOC 2 Type II Certified"`), which is the correct value type for the `award` property.
- **Added `numberOfEmployees`** with a `QuantitativeValue` range (100–200) — this is a Google Knowledge Panel signal.
- **Added `x.com/PulumiCorp` to `sameAs`** — Twitter rebranded to X; including both URLs ensures entity disambiguation works regardless of which domain search engines resolve.

### Dead code removal

Deleted 4 standalone `base/` schema partials that were **never called** by any template (confirmed via grep). All schema output goes through `graph-builder.html` via `loader.html` → `head.html`. These files had drifted from the graph-builder versions and were a source of confusion:

- `layouts/partials/schema/base/organization.html`
- `layouts/partials/schema/base/website.html`
- `layouts/partials/schema/base/software-app.html`
- `layouts/partials/schema/base/breadcrumb.html`

## Test plan

- [ ] Verify homepage renders correctly with `make serve`
- [ ] Inspect homepage source for the `@graph` JSON-LD block and confirm:
  - Organization `@type` is `["Organization", "Corporation"]`
  - `award` is a plain string, not an object
  - `numberOfEmployees` and `x.com` sameAs are present
- [ ] Validate with [Google Rich Results Test](https://search.google.com/test/rich-results) or [Schema Markup Validator](https://validator.schema.org/)
- [ ] Confirm no other templates break from the deleted `base/` partials


CamSoper (approved):
LGTM. Nice cleanup — net −212 lines, removes the invalid `AwardOrHonor` type (schema.org only defines `Award`; the `award` property takes `Text`), and safely deletes the four unreferenced `schema/base/*` partials (verified via grep — no templates reference them).

After merge, suggest running the homepage through Google's Rich Results Test to confirm the `@graph` parses cleanly with the multi-type `["Organization", "Corporation"]` array.

PR #18475, opened 2026-04-10
fix: disable compression encoding on no-cache CloudFront policy

CamSoper:
Fixes the production deployment failure from #18425.

CloudFront rejects `EnableAcceptEncodingGzip` on cache policies with caching disabled (TTL=0). The `cacheKeyPolicy()` helper now conditionally sets encoding flags only when TTL > 0.

## Changes
- `infrastructure/index.ts` — gate Brotli/gzip encoding on `cachingEnabled` (TTL > 0) in `cacheKeyPolicy()`

cc @sicarul @alexleventer


CamSoper:
Verified: the testing build-and-deploy workflow passed against pulumi-test.io on this branch. [Run 24258642915](https://github.com/pulumi/docs/actions/runs/24258642915).

sicarul (approved):
Makes sense, ship it!

PR #18479, opened 2026-04-10
feat: enable Origin Shield on docs CloudFront origin

sicarul:
## Summary
- Enables CloudFront Origin Shield on the docs (S3 main) origin
- Uses `us-west-2` to match the S3 bucket region for lowest latency
- Adds a centralized caching layer that collapses duplicate requests to the origin
- Adds comments on registry and guides origins noting Origin Shield should be configured in their respective repos (pulumi/registry, pulumi/guides) since they have their own CloudFront distributions
- Updates BUILD-AND-DEPLOY.md to document Origin Shield configuration

## Impact
- Customers report origin load reductions of ~60% and p90 latency reductions of ~67%
- Improves cache hit ratio by adding an extra cache layer between regional edge caches and the origin
- Cost: ~$0.0035 per 10,000 requests (modest for a static docs site)

### DIFF
--- a/BUILD-AND-DEPLOY.md
+++ b/BUILD-AND-DEPLOY.md
@@ -1611,7 +1611,7 @@ Delivery: CloudWatch Logs infrastructure v2
 
 **Compression:** Enabled (gzip, brotli)
 
-**Origin Shield:** Enabled on docs (S3 main) and guides origins in `us-west-2` to improve cache hit ratio and reduce origin load.
+**Origin Shield:** Enabled on the docs (S3 main) origin in `us-west-2` to improve cache hit ratio and reduce origin load. Registry and guides origins have their own CloudFront distributions and should configure Origin Shield in their respective repos.
 
 **Price Class:** All (global distribution)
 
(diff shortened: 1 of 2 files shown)


sicarul:
@claude please review this PR

sicarul:
Updated from the initial version:

- Removed Origin Shield from the registry origin. The registry has its own CloudFront distribution in pulumi/registry, so Origin Shield should be configured there instead.
- Added Origin Shield to the guides origin (was missing in the first push).
- Added a code comment on the registry origin explaining why it's excluded.
- Updated BUILD-AND-DEPLOY.md to document the Origin Shield configuration.

sicarul:
@claude check once more but talk like Heisenberg from breaking bad

CamSoper (approved):
LGTM. Narrowly-scoped and correct:

- `originShield` is applied only to the docs S3 main origin, with `us-west-2` matching the bucket region (verified against all stack configs).
- Registry and guides origins are correctly left untouched — those distributions live in their own repos.
- `BUILD-AND-DEPLOY.md` is updated to document the new setting.

I ran `testing-build-and-deploy.yml` against this branch and verified pulumi-test.io serves `/docs/` cleanly through CloudFront (200, `x-cache: Hit from cloudfront`). Heads up: the "Update Changelog" check is red — add the `impact/no-changelog-required` label before merge.

PR #18509, opened 2026-04-13
chore: prune scripts/programs/ignore.txt

jkodroff:
## Summary

Audit branch to determine which entries in `scripts/programs/ignore.txt` are still needed.

- Emptied `scripts/programs/ignore.txt` (keeping only `tf-state-ref-terraform`, which is intentional — not a Pulumi program).
- Appended a `# audit-touch` comment to `Pulumi.yaml` in each of the 52 previously-ignored program directories so PR-mode CI (`scheduled-test.yml`, `TEST_MODE=pull_request`) picks them up via its `git diff --name-only master` filter in `scripts/programs/test.sh:18-49`.

This scopes the test run to exactly the previously-ignored set — no full run of every program.

**Do not merge as-is.** After the workflow run produces pass/fail results, a follow-up commit on this branch will:
1. Revert the `Pulumi.yaml` touches.
2. Repopulate `ignore.txt` with only the programs that still fail, each annotated with a reason and issue link.

## Test plan

- [ ] `Run Example Code Tests` workflow triggers and reports `Number of programs to test: 52`.
- [ ] Review the final pass/fail report to classify each previously-ignored program.
- [ ] Update `ignore.txt` based on results; revert touches; re-run CI to confirm green.

### DIFF
--- a/scripts/programs/ignore.txt
+++ b/scripts/programs/ignore.txt
@@ -10,53 +10,10 @@
 # aws-.* will ignore all programs that start with "aws-"
 #
 
-
-
-# Java SDK serialization bug: FargateService container/containers inputs
-# not received by AWSX provider. https://github.com/pulumi/pulumi-java/issues/2090
-awsx-load-balanced-fargate-nginx-java
-awsx-vpc-fargate-service-java
-
-# API Gateway authorizer parameter `providerArns` is mistyped.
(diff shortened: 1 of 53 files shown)


jkodroff:
This is marked `do-not-merge` because I want CI to run against all the currently ignored programs to figure out which should be tested.

CamSoper (approved):

PR #18510, opened 2026-04-13
Restructure publishing-packages docs to cover private registry workflow

djperlovsky:
The publishing-packages page was exclusively about the public Pulumi Registry, leaving users who want to publish components to their private registry without guidance. Users searching for "how to publish a package" landed here and hit a dead end.

## Changes
- Added a "Publish to your organization's private registry" section with `pulumi package publish` workflow, placed before the public registry content
- Restructured existing public registry content under its own "Publish to the public Pulumi Registry" heading
- Added cross-references to Packaging Components, Private Registry, and Publishing from GitHub Actions docs
- Clarified that boilerplate templates are for provider-based packages, with link to cross-language component approach
- Updated parent index page description to mention both registries
- Removed dismissive note box that sent private registry users to Build a Component (which doesn't cover publishing)

### DIFF
--- a/content/docs/iac/guides/building-extending/_index.md
+++ b/content/docs/iac/guides/building-extending/_index.md
@@ -48,7 +48,7 @@ Package and distribute your components and providers for use across teams and pr
 
 **[Local Packages](/docs/iac/guides/building-extending/packages/local-packages/)** - Develop and test packages locally before publishing them to registries.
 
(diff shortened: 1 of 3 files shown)


CamSoper (approved):
Clean restructure — the publisher-first framing is much better for users landing here looking to publish to a private registry. Internal links and CLI flags verified.

Pre-existing nits (not blocking, flagging in case you want to sweep them): `publishing-packages.md` has a mix of "&" and "and" in headings, and one note block in `private-registry.md` is missing a terminal period.

CamSoper (commented):
Merging this, LMK if we need to revert

PR #18513, opened 2026-04-13
Free runner disk space for example code tests

### LINKED ISSUE #14403: Address disk usage issues in GitHub code example runner
Our Github runner for the code example tests is running out of disk space. Temporarily we are going to bump to use one of the paid runners. But we should do some digging to find out what is causing this and figure out a way to manage it.

CamSoper:
Switches the example code test workflow back from the paid `pulumi-service-ubuntu-24.04-4core` runner to `ubuntu-latest` and adds `jlumbroso/free-disk-space` as the first step in the `test` job to reclaim disk space before tests run.

Matches the pattern already used in `pulumi/examples` (`.github/workflows/test-examples.yml`, `providers` job). The action is configured with `tool-cache: false` and `dotnet: false` to preserve caches that later `setup-*` steps rely on.

I ran the full test workflow against this, it seems to work fine.

Fixes #14403

### DIFF
--- a/.github/workflows/scheduled-test.yml
+++ b/.github/workflows/scheduled-test.yml
@@ -47,7 +47,7 @@ jobs:
 
     steps:
       - name: Free Disk Space (Ubuntu)
-        uses: jlumbroso/free-disk-space@54081f13873cdfa15788a463838342cd2f914a1be # v1.3.1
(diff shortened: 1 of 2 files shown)


CamSoper:
@claude Update BUILD-AND-DEPLOY as suggested.

cnunciato (approved):

PR #18544, opened 2026-04-15
Simplify Terraform state backend auth docs

nyobe:
Leads with `terraform login api.pulumi.com` (which now works out of the box) as the primary authentication path, with `TF_TOKEN_api_pulumi_com` documented as the non-interactive fallback for CI/CD.

## Changes
- Replace step 3 "Set your access token" with `terraform login` instructions
- Extend the self-hosted callout to cover both backend config and auth
- Update "Failed to get existing workspaces" troubleshooting to mention both auth paths

### DIFF
--- a/content/docs/iac/get-started/terraform/terraform-state-backend.md
+++ b/content/docs/iac/get-started/terraform/terraform-state-backend.md
@@ -82,18 +82,22 @@ Replace the placeholders:
 | `<stack>` | The Pulumi stack name | `prod` |
 
 {{% notes "info" %}}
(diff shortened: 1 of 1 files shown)


CamSoper (approved):
Clean simplification -- leading with `terraform login` and demoting `TF_TOKEN_api_pulumi_com` to the CI/CD fallback reads better, and the self-hosted callout expansion is a nice touch. Fact-checked the auth flow claims against standard Terraform behavior; all check out. Thanks, Claire!

PR #18555, opened 2026-04-16
Update pulumi-go-provider SDK doc and move under Packages

### LINKED ISSUE #18553: Update content for pulumi-go-provider
https://www.pulumi.com/docs/iac/guides/building-extending/providers/sdks/pulumi-go-provider-sdk

- Remove the FAQ, but do keep the fact that we use it internally for the EKS package.
- Revise the tone to be more neutral and more in line with the rest of the docs tone.
- Revise the wording "traditional authoring experience". The alternative is hand authoring the schema. Maybe remove this comparison altogether since the advantages are fairly clear anyway.
- Remove the migration guide - no one is migrating a handwritten provider to pulumi-go-sdk in all likelihood.
- Make sure that it's clear that the library is for authoring Pulumi *packages* and can be used to include components for distribution (either as source-based or executable-based plugin), but can also include functions (for which we need an example and might not have one - check https://github.com/pulumi/pulumi-go-provider/tree/main/examples), and custom resources (which is in the example code).

jkodroff:
Rewrites the Pulumi Go Provider SDK page per #18553:

- Neutral, less marketing-y tone.
- Clarifies that pulumi-go-provider authors Pulumi packages, which can contain custom resources, components, and functions.
- Adds a function example adapted from the upstream `str` example.
- Adds a Components section pointing to `examples/component-provider` and the source-based plugin guide.
- Removes the FAQ, migration guide, "traditional authoring" comparison table, and TBD testing stub.
- Preserves the fact that Pulumi uses pulumi-go-provider internally for `pulumi-eks`.

Also moves the page from `providers/sdks/` to `packages/` (with aliases for the old URL) and updates internal links in `docs/`.

Fixes #18553


CamSoper (changes requested):
Two factual issues:

**`content/docs/iac/guides/building-extending/packages/pulumi-go-provider-sdk.md:27`** -- `pulumi-eks` isn't built on pulumi-go-provider. Its `provider/go.mod` has no `pulumi-go-provider` require, and the provider entrypoint is `nodejs/eks/cmd/provider/index.ts` using `@pulumi/pulumi`'s component provider interface. Real first-party users: `pulumi-command`, `pulumi-std`, `pulumi-docker-build`, `pulumi-terraform`, `pulumi-aws-native`.

```suggestion
The library uses Go reflection to derive the package's [schema](/docs/iac/guides/building-extending/packages/schema/) from your Go types, so multi-language SDKs can be generated without hand-authoring a JSON schema. Pulumi uses pulumi-go-provider internally for several of its own packages, including [`pulumi-command`](/registry/packages/command/).
```

**`content/docs/iac/guides/building-extending/packages/pulumi-go-provider-sdk.md:144`** -- `resourcex` was removed from pulumi-go-provider in [#355](https://github.com/pulumi/pulumi-go-provider/pull/355) (May 2025). Current top-level packages are `infer`, `integration`, `middleware`.

```suggestion
- **integration**: in-process helpers for testing providers end-to-end without the Pulumi engine.
```

Also optional: line 62 of `build-a-provider.md` still points to `/docs/iac/using-pulumi/extending-pulumi/pulumi-provider-sdk/`. You're updating every other link in that file, so worth catching in the same PR.

CamSoper (approved):
Thanks Josh, all three items are resolved -- the `pulumi-command` swap and the `integration` replacement both verified against upstream. Approving. :shipit:

PR #18562, opened 2026-04-16
Update functions docs with language-specific descriptions

### LINKED ISSUE #15457: Expand functions docs to use language-specific terminology
In the Functions docs, we use some terminology that is not necessarily applicable in all languages, e.g. Get functions are not statically attached to a type in Go - they're just free-standing functions that are part of the namespace. For Provider Functions, we might want to ensure that the types are specifically mentioned (e.g. Promise in TS, Task in Python, etc.)

In order to enhance the docs, we should ensure that the functions pages are idiomatic and technically accurate for all languages.

I did attempt to do this for invokes (Provider Functions), but I couldn't get the signatures to display on a single line due to the layout, so I abandoned the effort.

jkodroff:
Add language-specific chooser blocks to the functions concept pages so each language shows accurate terminology (e.g. static methods in TS/C#/Java, class methods in Python, package-level functions in Go, get stanzas in YAML).

## Changes
- Add chooser blocks to functions index, get functions, and provider functions pages
- Fix broken list rendering in direct/output form section
- Fix wording and bolding in choosing section

Fixes #15457

### DIFF
--- a/content/docs/iac/concepts/functions/_index.md
+++ b/content/docs/iac/concepts/functions/_index.md
@@ -26,7 +26,33 @@ For example, you might use a provider function to get the latest virtual machine
 
 ## Get functions
 
-**[Get functions](/docs/iac/concepts/functions/get-functions/)** are static functions available on all Pulumi resource types that allow you to reference an existing resource that is not managed by Pulumi. Unlike the `pulumi import` command which brings resources under Pulumi management, get functions simply allow you to read the properties of existing resources.
+{{< chooser language "typescript,python,go,csharp,java,yaml" >}}
+
+{{% choosable language "typescript,csharp,java" %}}
+
+**[Get functions](/docs/iac/concepts/functions/get-functions/)** are static methods available on all Pulumi resource classes that allow you to reference an existing resource that is not managed by Pulumi. Unlike the `pulumi import` command which brings resources under Pulumi management, get functions simply allow you to read the properties of existing resources.
+
+{{% /choosable %}}
+
+{{% choosable language python %}}
+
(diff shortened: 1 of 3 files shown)


CamSoper (changes requested):
A couple of accuracy nits before this lands -- the chooser scaffolding and most of the per-language naming claims are spot on. Just two:

1. **Python `get` is a static method, not a class method.** In `_index.md:39` and `get-functions.md:31`, the Python entry says "class method(s)" -- but the Pulumi Python SDK implements `get` as `@staticmethod`, not `@classmethod` (a Python reader will read "class method" as `@classmethod` that receives `cls`). Since the call site is `Resource.get(...)` in all four of TS/C#/Java/Python, you can either:
   - Merge Python into the TS/C#/Java choosable block ("static methods on resource classes"), or
   - Just change "class method(s)" to "static method(s)" in the Python entry.

2. **Go provider-function naming is mixed `Lookup`/`Get`.** In `provider-functions.md`, the direct form ("typically named `LookupX()`") and output form ("typically named `LookupXOutput()`") overstate the convention. The registry has `LookupAmi`/`LookupAmiOutput` but also `GetAmiIds`/`GetAmiIdsOutput`, `GetCallerIdentity`/`GetCallerIdentityOutput`, etc. -- the Go SDK uses both prefixes to avoid collisions with resource-level `GetX` methods. Suggest "typically named `LookupX()` or `GetX()`" / "`LookupXOutput()` or `GetXOutput()`".

Pushed a small commit fixing some pre-existing Hugo shortcode bracket mismatches in `provider-functions.md` (`{{% /chooser %}}` -> `{{< /chooser >}}` and `{{% / notes %}}` -> `{{% /notes %}}`) while I had the file open. Feel free to revert if you'd rather keep them out of this PR.

CamSoper (approved):
Both items from my prior review are addressed -- Python is merged into the TS/C#/Java "static methods" block, and the Go naming now correctly shows `LookupX()`/`GetX()` and `LookupXOutput()`/`GetXOutput()`. Spot-checked the rest of the per-language return types and naming conventions against the code examples on this page and they all line up. Thanks for the quick turnaround!

PR #18572, opened 2026-04-16
Add launch blog for OpenAPI-generated REST API docs

djgrove:
### Proposed changes

Adds a launch blog post announcing that the Pulumi Cloud REST API reference at `/docs/reference/cloud-rest-api/` is now generated from the live OpenAPI spec (shipped in #18551).

### Related issues
Announces feature shipped in #18551


CamSoper (approved):
Reads well and the facts all check out against #18551's diff (path, tag categories, schema page all line up).

Two optional prose nits, take or leave:

1. The sentence "We've heard feedback in this area, and wanted to respond with a durable solution for docs accuracy that will scale with us." (line 26) reads a little marketing-y. Something like "We wanted a durable fix that keeps the docs in sync as the API grows" tightens it up.

2. The numbered list under "What's new" has each item as a heading + single sub-bullet, which reads a bit clunky. Consider collapsing each into a single bolded lead-in, e.g. `1. **Find what you need faster.** Endpoints are grouped by product area...`

Neither is blocking. Assuming #18551 lands first, this is good to ship.

tyler-dunkel:
As the blog post, we want to focus more on "why" vs "what" which we touch on here some but it would be great to add a section on something you can do with an agent and the api docs and talk about how it works better now because of being always in sync (ie add a novel workflow or 2 a user might do that uses the pulumi api docs such as setup roles) this will help expose why someone might use these api docs. (Not blocking. Just an idea)

djgrove:
> As the blog post, we want to focus more on "why" vs "what" which we touch on here some but it would be great to add a section on something you can do with an agent and the api docs and talk about how it works better now because of being always in sync (ie add a novel workflow or 2 a user might do that uses the pulumi api docs such as setup roles) this will help expose why someone might use these api docs. (Not blocking. Just an idea)

Good feedback. Added an agents/impact section in [4c816d0](https://github.com/pulumi/docs/pull/18572/commits/4c816d0b9c8aca86e696079dccc0395fb387a510)

content/blog/rest-api-docs-from-openapi/index.md
```diff
@@ -0,0 +1,72 @@
+---
+title: "Pulumi Cloud REST API Docs, Now Generated from OpenAPI"
+date: 2026-04-16
```
djgrove: @claude  Please update the date here so that it deploys Monday morning at 7AM Eastern

djgrove:
@claude will the social media front matter trigger posts to our social platforms automatically? What takes precedence if the blog post is configured for a date in the future? If the blog post is not scheduled until Monday, will we attempt to post to socials on merge?

djgrove:
@claude let's remove the specific timing on the frontmatter date and just use Monday's date only

PR #18573, opened 2026-04-16
Revise steps to view Customer Managed Keys

smithrobs:
### Proposed changes

Updated instructions for viewing Customer Managed Keys to include selecting the Organization before accessing the Customer Managed Keys tab.


CamSoper (approved):
Looks good -- the reorganized steps match the current Pulumi Cloud UI flow, and the ordered list uses the repo's `1.`-for-every-item convention.

PR #18588, opened 2026-04-17
remove false claim from the blogpost

aacotroneo:
I overlooked this statement. The first part was true

"and every scanner API call targets that partition's regional endpoints"

but the latter is not 

"Discovery traffic doesn't cross the boundary." 

as we don't deploy anything to their accounts.

I'm removing the paragraph as it doesn't really add much. The blogpost mentions the new feature, the partitions and a screenshot. Simple.


CamSoper (approved):

CamSoper (approved):
Happens to the best of us!

PR #18590, opened 2026-04-17
Remove Go direct-implementation provider guide in favor of pulumi-go-provider

jkodroff:
Deletes the "Direct Implementation → Go" guide because `pulumi-go-provider` is Pulumi's supported, published framework for building Go providers — there is no reason to document a raw-gRPC path for Go when it actively steers users away from the framework.

## Changes

- Deleted `content/docs/iac/guides/building-extending/providers/implementers/go.md`
- Added the old URL as a Hugo alias on `build-a-provider.md` so bookmarks and search traffic land on the framework guide
- Removed inbound links from `provider-architecture.md`, `protocol-reference.md`, and `implementers/_index.md`
- Clarified in `provider-architecture.md` that the higher-level SDK (Layer 3) is **Go only**, with Python/TypeScript readers routed explicitly to direct implementation
- Reworded the notes block in `build-a-provider.md` to replace the "gRPC bindings for full control" phrasing with "protocol-level control"

## Test plan

- [x] `make lint` passes
- [x] `make build` passes (2484 aliases registered)
- [x] Verified old `/implementers/go/` URL generates redirect HTML pointing to `/build-a-provider/`
- [x] No remaining orphaned references to the deleted page


CamSoper (approved):
LGTM.

PR #18601, opened 2026-04-18
Document NPW cloud filter for org templates

### LINKED ISSUE #13571: New Project Wizard: How does the filter by cloud work?
The NPW has a filter by cloud:

Image

But there's no info on how to set that in the template metadata here:

<https://www.pulumi.com/docs/pulumi-cloud/developer-portals/templates/>

nor here:

<https://www.pulumi.com/docs/pulumi-cloud/developer-portals/new-project-wizard/>

It's probably something in here: <https://www.pulumi.com/docs/iac/concepts/projects/project-file/#template-options>

but even in the reference it's not explicitly documented.

jkodroff:
Documents how `template.metadata.cloud` controls the New Project Wizard "Filter by cloud" dropdown for organization templates.

## Changes
- Added a "New Project Wizard cloud filter" subsection to the Organization Templates page covering recognized values (`aws`, `azure`, `gcp`, `kubernetes`), the single-string constraint, and behavior when the field is unset.
- Expanded the `template.metadata` row in the project file reference to point readers at the new section.

Closes #13571.

### DIFF
--- a/content/docs/iac/concepts/projects/project-file.md
+++ b/content/docs/iac/concepts/projects/project-file.md
@@ -163,7 +163,7 @@ Schemas are only valid for project property keys. For setting the value of a pro
 | `displayName` | optional | A user-friendly name for the template. This should follow `Title Case` format and be succinct. This field is only supported by Pulumi CLI >= 3.95. |
 | `description` | optional | Description of the template. |
 | `config` | required | Config to request when using this template with `pulumi new`. |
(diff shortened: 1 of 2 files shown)


CamSoper (approved):
LGTM. Spot-checked the three substantive claims (recognized cloud values, single-string constraint, unset behavior) against the NPW source and they line up.

Issue #18602, opened 2026-04-18
Audit and document custom vs component resource options


jkodroff:
## Background

The current docs describe resource options (`parent`, `providers`, `dependsOn`, `aliases`, `protect`, `ignoreChanges`, `customTimeouts`, `transforms`, `hooks`, etc.) but it's not always clear which options apply to which resource types. Some options are component-only (e.g., `providers`); some are no-ops on components (e.g., `ignoreChanges`, `customTimeouts`). The classification isn't surfaced consistently on each option's reference page, and the SDK-level enforcement varies.

## Tasks

- [ ] Audit every resource option in `content/docs/iac/concepts/resources/options/` and classify each as:
  - Custom-resource only
  - Component-resource only
  - Both (and note any behavior differences)
- [ ] Surface the classification on each option's reference page in a consistent place (e.g., a frontmatter field rendered as a callout, or a fixed line in the intro).
- [ ] Investigate how each SDK enforces (or doesn't enforce) the distinction between options that apply to custom vs component resources, and whether \`CustomResourceOptions\` can be passed where \`ComponentResourceOptions\` is expected (or vice versa):
  - **TypeScript**: \`pulumi.ResourceOptions\` vs \`pulumi.CustomResourceOptions\` vs \`pulumi.ComponentResourceOptions\` — what's the type hierarchy, and does the type system catch misuse?
  - **Python**: There's a single \`pulumi.ResourceOptions\` class — does it accept all options regardless of resource type, and what happens at runtime if you pass a component-only option to a custom resource?
  - **Go**: \`pulumi.ResourceOption\` is a variadic functional-options pattern — does the SDK enforce which options are valid on which resource type, or does it silently accept incorrect combinations?
  - **C#**: \`Pulumi.ResourceOptions\` vs \`CustomResourceOptions\` vs \`ComponentResourceOptions\` — same question as TS.
  - **Java**: similar split.
- [ ] Document the SDK-level enforcement (or lack thereof) on each option's reference page so users know whether passing the wrong option is a compile error, runtime error, or silent no-op.

## Why this matters

The \`build-a-component.md\` guide currently states \"component resources accept a \`providers\` option that custom resources don't have\" without an authoritative source for that claim, and there's a \`Resource options and component resources\` section linked from the components concept page that lists a few options as not applying to components — but those lists may be incomplete and the format is inconsistent across pages. Authors building components or consuming them are left to discover behavior empirically.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #18603, opened 2026-04-18
Refactor Build a Component guide and expand providers resource option docs

jkodroff:
Refocuses the Build a Component guide on authoring and points readers to the source-based plugin guide for packaging.

## Changes

- **Build a Component**: removed the project-scaffolding and consumer-program walkthrough sections that duplicated content in the source-based plugin guide.
- Added a new **Implementation details** parent section grouping the reference subsections (component arguments, defining a component resource, child resources, registering outputs, passing resource options).
- Wrapped **Component arguments and type requirements** in a per-language chooser (TypeScript, Python, Go, C#, Java, YAML), with each pane documenting that language's args shape, supported/unsupported types, optional/required signaling, naming, and best practices — sourced from the analyzers in pulumi/pulumi, pulumi-dotnet, pulumi-java, and pulumi-component-provider-go-boilerplate.
- Tightened the **Child resources** walkthrough across all six panes to two principles (parent + name templating) plus a warning about hard-coded names.
- Fixed Go and C# code-block line lengths to avoid horizontal scroll.
- Various copy fixes (terminology: annotation vs decorator, factory function vs constructor for Go; path corrections to canonical resource-options paths; removed dynamic-providers reference; replaced "sub-resources" with "child resources" throughout).
- Repaired stale TS line ranges in the static-program example references.

- **`providers` resource option page**: documented that the option is component-only; added the lookup-order explanation, inheritance behavior, and a worked multi-region example moved here from the guide.

Related: #18602 (audit custom vs component resource options across SDKs).

### DIFF
--- a/content/docs/iac/concepts/resources/options/providers.md
+++ b/content/docs/iac/concepts/resources/options/providers.md
@@ -15,77 +15,163 @@ aliases:
   - /docs/concepts/options/providers/
 ---
 
-The `providers` resource option sets a map of providers for the resource and its children. This map is combined with resource parent's `providers` map. If no value is provided, the providers map is identical to the parent's providers map. When determining which provider to use for a resource, the `providers` map is used if [provider](/docs/concepts/options/provider/) is not supplied.
+The `providers` resource option is supported only on [component resources](/docs/iac/concepts/components/). It sets a map of providers (keyed by package name) for the component and its child resources — the map flows from the component to every child via parent inheritance. To pass an explicit provider to a single custom resource, use the [`provider`](/docs/iac/concepts/resources/options/provider/) option instead.
 
-{{< chooser language "typescript,python,go,yaml" >}}
+When the SDK creates a resource it picks a provider by checking, in order:
+
+1. The explicit `provider` option, if set.
+1. A matching provider in the resource's `providers` map (looked up by package name).
+1. A matching provider inherited from the parent's `providers` map.
+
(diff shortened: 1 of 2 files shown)


CamSoper (approved):
Approving. The restructure is a big improvement — pulling scaffolding out to the source-based-plugin guide, unifying "sub-resources" → "child resources", and wrapping the per-language args reference in a chooser all make the page much easier to navigate. Line-range repairs on the static-program references and the path corrections to the canonical `/docs/iac/concepts/resources/options/` URLs are nice catches.

A few things surfaced that you may want to follow up on (none blocking):

1. **`providers` as "component-only"** (`providers.md:18`) — accurate for TypeScript/.NET/Java where `providers` lives on `ComponentResourceOptions`, but in Python (`sdk/python/lib/pulumi/resource.py`) and Go (`ProviderMap` returns `ResourceOption`) the field is on base `ResourceOptions` and applies to any resource plus its children. The old phrasing actually matches the Python docstring verbatim. Consider "primarily used with component resources" or a brief note on the language variance.

2. **Go `pulumi:"fieldName,skip"` tag** (build-a-component, Go args "Struct tags") — the `skip` suffix isn't recognized in pulumi-go-provider's introspect. Fields are excluded by omitting the `pulumi` tag or keeping them unexported. Worth dropping that bullet.

3. **Go enums listed as unsupported** (Go "Unsupported types") — `pulumi-go-provider`'s `infer` package does support enums via the `infer.Enum[T]` interface (T constrained to `~string | ~float64 | ~bool | ~int`) with a `Values() []EnumValue[T]` method. The narrower true claim is that `infer` doesn't auto-synthesize enums from bare `const` declarations — worth rewording.

4. **Java "constructor must take exactly one parameter assignable to `ResourceArgs`"** — the analyzer (`ComponentAnalyzer.java`) picks the first constructor and looks for *a* parameter assignable to `ResourceArgs` among its params; it doesn't require the constructor to take only one parameter. Something like "The constructor must have a parameter assignable to `ResourceArgs`" matches the actual enforcement.

Nothing urgent — happy to see this land.

PR #18612, opened 2026-04-20
Fix mobile header: restore close icon, remove empty top bar, lock scroll, prevent layout shift

dwahbe:
### Proposed changes

Four small, related fixes to the mobile site header, touching `layouts/partials/header.html`, `theme/src/scss/_header.scss`, and `theme/src/ts/misc.ts`.

- **Restore the close (X) icon when the burger menu is open.** The open-state `<label>` carried the Tailwind `hidden` class. Tailwind v4 puts `.hidden` in the `utilities` layer, which beats the `display: block` override in the `components` layer regardless of selector specificity, so the X never rendered. Removed `hidden` from the markup and replaced `@apply hidden/block` with plain `display: none/block` so the toggle override stays in the same layer as its default.

Before
Screenshot 2026-04-20 at 5 07 25 PM

After
Screenshot 2026-04-20 at 5 06 47 PM

- **Remove ~32px of empty space above the header on mobile.** The `.top-nav-bar` wraps its desktop-only links UL in a `<div class="w-full p-4">`. On mobile the UL is `hidden`, but the wrapper's 16px top+bottom padding still reserves 32px. Added `hidden lg:block` so the wrapper only takes space at the breakpoint where its content is visible.

Before
Screenshot 2026-04-20 at 5 08 24 PM

After
Screenshot 2026-04-20 at 5 08 54 PM

- **Lock body scroll while the mobile menu is open.** Previously the page behind the menu could still scroll. Added `html:has(.mobile-menu-toggle:checked) { overflow: hidden }`. Because `overflow: hidden` on `<html>` breaks `position: sticky`, the rule also pins `.header-container` with `position: fixed` while the menu is open so it stays in the viewport regardless of scroll position.

- **Prevent a 1px layout shift when opening the menu.** The header's sticky uses `top: -1px` for IntersectionObserver-based pinned-state detection. When the menu opens, the header grows to `100vh` and the IO fired and added `is-pinned`, which shifted the fixed header up by 1px. The observer now early-returns when the mobile menu toggle is checked, so `is-pinned` doesn't flip based on menu state.

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -8,7 +8,7 @@
 <nav class="top-nav-bar transition-all">
     <div class="top-nav-container lg:container lg:px-4 flex-wrap justify-end">
         {{ partial "banner" . }}
-        <div class="w-full p-4">
+        <div class="hidden lg:block w-full p-4">
             <ul class="hidden w-full lg:flex justify-end items-center order-first lg:order-last">
                 <li class="github-widget mr-8">
                     <a data-track="header-github-pulumi" class="github-button" href="https://github.com/pulumi/pulumi"
@@ -357,7 +357,7 @@
                     <label class="mobile-menu-closed" for="mobile-menu">
                         {{ partial "icon" (dict "icon" "hamburger-closed") }}
                     </label>
-                    <label class="mobile-menu-open hidden" for="mobile-menu">
+                    <label class="mobile-menu-open" for="mobile-menu">
                         {{ partial "icon" (dict "icon" "hamburger-open") }}
                     </label>
                 </span>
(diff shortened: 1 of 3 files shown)


CamSoper (approved):
Thanks @dwahbe -- the writeup made this easy to review. Approving from the docs side. Tagging @cnunciato and @jeffmerrick for a second look since this touches the homepage header.

cnunciato:
Hi @dwahbe! Thanks for this submission, we appreciate it. The timing's a little unfortunate, though, since we're about to ship a completely updated nav in #18806, and the disappearing-icon issue was fixed in #18764. So it probably wouldn't make sense to merge this.

I'll therefore close for now, but thanks again for taking the time to help make Pulumi better!

dwahbe:
The new nav looks great!

PR #18620, opened 2026-04-20
Tighten REST API tag-list descriptions on cloud-rest-api landing

djgrove:
## Summary

* The "API documentation by category" list on https://www.pulumi.com/docs/reference/cloud-rest-api/ previously rendered identical boilerplate for every entry ("API reference for X endpoints in the Pulumi Cloud REST API"). Each item now gets a curated 1-2 sentence description.
* Promote the `assets/openapi/tag-intros/*.md` sidecar files to carry a `meta_desc` field in YAML frontmatter. The content adapter (`_content.gotmpl`) reads it and uses it as both the page's SEO `<meta name="description">` and the landing-page list-item description (rendered via the `openapi-tag-list` shortcode), so there's a single source of truth per tag.
* Switched the list-item separator from em dash to colon.
* Filled in the two missing intros (`AI`, `RegistryPreview`) — also clears the existing `openapi: missing intro` build warnings.

### Before
Screenshot 2026-04-20 at 3 07 50 PM

Every entry: `Access Tokens — API reference for Access Tokens endpoints in the Pulumi Cloud REST API.`

### After

Screenshot 2026-04-20 at 3 07 33 PM

`Access Tokens: Access tokens authenticate users and automation with the Pulumi Cloud API. The Access Tokens API manages personal access tokens (PATs) for individuals and organization-level tokens for team and CI use.`

## Test plan

* [x] `hugo --buildFuture` succeeds with no openapi-related warnings
* [x] Landing page list renders all 31 entries with `: ` separator and curated descriptions (verified in generated HTML)
* [x] Tag pages emit the curated `meta_desc` as `<meta name="description">` (verified for `cloud-setup`, `registry`)
* [x] Tag page intro bodies render without YAML frontmatter leaking (no stray `<hr>` or raw `meta_desc:` text)
* [ ] Maintainer spot-check: visit `/docs/reference/cloud-rest-api/` after deploy preview is up

### DIFF
--- a/assets/openapi/tag-intros/README.md
+++ b/assets/openapi/tag-intros/README.md
@@ -11,9 +11,13 @@ Markdown fragments rendered at the top of each generated tag page under
   `CloudSetup` -> `Cloud Setup` -> `cloud-setup`), so the file for the
   `CloudSetup` tag must be `cloud-setup.md`.
 - YAML frontmatter holds a `meta_desc` field — a 1-2 sentence description
-  of the API area, used as the tag page's `<meta name="description">` and
-  as the list-item description on the cloud-rest-api landing page. Keep
-  `meta_desc` to plain text (no Markdown links) — it renders unescaped.
+  (≤160 chars, so Google search snippets aren't truncated) used as the
+  tag page's `<meta name="description">` and as the list-item description
+  on the cloud-rest-api landing page. Use plain text — `meta_desc` is
+  not processed as Markdown, so links or emphasis appear literally.
+- Avoid a literal `: ` in unquoted `meta_desc` values; YAML parses it as
+  a nested mapping and the build fails. Reword (e.g., use a comma) or
+  quote the value.
 - Body is Markdown. Headings (`##`, `###`) and lists are fine; they render
   inside the page above the auto-generated endpoint sections.
 - Keep the body editorial: purpose of the API area, when to use it, links
--- a/assets/openapi/tag-intros/README.md
+++ b/assets/openapi/tag-intros/README.md
@@ -10,11 +10,27 @@ Markdown fragments rendered at the top of each generated tag page under
   through `urlize` after inserting spaces into CamelCase names (e.g.
   `CloudSetup` -> `Cloud Setup` -> `cloud-setup`), so the file for the
   `CloudSetup` tag must be `cloud-setup.md`.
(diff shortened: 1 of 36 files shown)


djgrove:
Thanks for the review @claude. Addressed in three commits on this branch:

* **Length, repetition, and casing (#1, #5, #6):** 0fe384d1d5a93ed0be6e99762dc81b619e761216 — every `meta_desc` is now ≤160 chars (range 95–159), leads with an action verb, and no longer restates the page-title noun. Two values had to be reworded (organizations, services) because a bare `: ` in unquoted YAML breaks the parse during build.
* **Frontmatter split hardness (#2):** 5f618f73e6276c6771c65cd71fb0bbcb33aadad8 — switched `open-api-gen.html` to `delimit (after 2 $parts) "---\n"` so a future `---` HR in an intro body won't truncate the rendered page. Note: `_content.gotmpl` only reads `index $parts 1` (the frontmatter, never the body), so it isn't affected by the same bug — left it alone.
* **README accuracy (#3):** fb55907c8b0b2436ad1813d7c61f443f08283011 — replaced "renders unescaped" with the actual constraint ("not processed as Markdown"). Also added the 160-char target and the YAML bare-`: ` hazard surfaced during the tightening pass.

**Declined / no action:**

* **Stack Config Markdown link (#4):** intentional. Body keeps the inline link to Stack Updates; `meta_desc` drops it because the field is rendered as plain text.

Going to spot-check the deploy preview once it's up, then this is ready for another pass if you have one.

CamSoper (approved):
LGTM. Template logic is clean -- frontmatter split and rejoin correctly protects against body-level horizontal rules, and the fallback meta_desc preserves behavior when a tag-intro is missing or frontmatter-less. All 31 meta_desc values are under 160 chars and none contain an unquoted `: ` that would trip YAML parsing. Landing page list reads much better now.

Non-blocking nits for a future pass:

- `miscellaneous.md` and `registry-preview.md` start with noun phrases ("Utility endpoints...", "Preview endpoints...") rather than action verbs -- defensible since both tags are essentially descriptive buckets, but worth noting if we ever tighten the guideline.
- `stacks.md` is 159/160 chars, so any future tweak will blow the budget. Might be worth shaving a few now.
- Minor: `_content.gotmpl` uses `os.ReadFile` while the partial uses `readFile`. Both work; `readFile` is the more conventional Hugo idiom if you want consistency.

PR #18622, opened 2026-04-20
Show posted and updated dates on blog posts

alexleventer:
## Summary

Blog post headers now always show the original posted date, and append a `· Updated on <date>` suffix when the post's frontmatter includes an `updated:` field. Previously the header was either-or: posts with an `updated:` field hid their original publish date entirely.

Also drops the weekday name from both dates. It wasn't adding information in a post header, and it made the single-line layout busy once two dates were visible side by side.

Before:

> Posted on Wednesday, Jun 12, 2024

or (if `updated:` is set):

> Updated on Wednesday, March 12, 2025

After:

> Posted on Jun 12, 2024 · Updated on March 12, 2025

Posts without `updated:` in frontmatter are unchanged other than the weekday removal — they still read "Posted on \<date\>".

## Test plan

- [ ] Visit a blog post with an `updated:` field (e.g. `/blog/pulumi-copilot/`) and confirm both dates render on one line separated by `·`
- [ ] Visit a blog post without an `updated:` field and confirm only "Posted on \<date\>" renders
- [ ] Confirm the weekday name is no longer shown in either case
- [ ] Spot-check responsive behavior on narrow viewports (the line should wrap cleanly)

### DIFF
--- a/layouts/blog/single.html
+++ b/layouts/blog/single.html
@@ -25,7 +25,7 @@
                         </div>
 
                         <div class="text-sm text-gray-700 mb-4">
-                            <span>Posted on <time datetime="{{ .Date.Format "2006-01-02" }}">{{ .Date.Format "Jan 2, 2006" }}</time>{{ if .Params.updated }} · Updated on <time datetime="{{ (time .Params.updated).Format "2006-01-02" }}">{{ (time .Params.updated).Format "Jan 2, 2006" }}</time>{{ end }}</span>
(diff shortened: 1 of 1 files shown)


alexleventer:
Here's what it looks like: 
image

jeffmerrick (approved):
Looks good!

alexleventer:
@claude - one final review pls

PR #18642, opened 2026-04-21
Fix live reload not working reliably, speed up dev build times

jeffmerrick:
**`make serve-all`: CSS changes now reliably update in the browser**

**Bug fixes**

SCSS edits made during `make serve-all` were not appearing in the browser without a server restart. Three root causes fixed:

- **`@tailwindcss/postcss` stale cache** (postcss.config.js): Tailwind's internal cache only tracked `main.scss`'s mtime, so changes to imported partials were silently ignored. A new PostCSS plugin registers all SCSS files as dependencies so Tailwind detects any partial change and fully rebuilds.
- **Webpack module cache** (webpack.config.js): Webpack's default in-memory cache served a stale CSS module even when partials changed. Disabled in dev (`cache: false`).
- **Hugo LiveReload with stable CSS URLs** (assets.html): In dev, CSS `<link>` tags now use the stable `/css/bundle.css` URL instead of a fingerprinted hash URL, so Hugo's LiveReload can hot-swap the stylesheet without a full page reload.

**Performance improvements**

In addition, there were some long build times with local dev, some tweaks brought that down to ~1-3s:

- **Skip `js_manifest.json` write when unchanged** (webpack.config.js): CSS-only rebuilds previously always wrote a new manifest, triggering a full Hugo page reload. Now only written when JS content hashes actually change.
- **Skip Terser in dev** (webpack.config.js): Minification now only runs in production builds.
- **Run stencil only on startup** (webpack.config.js): Stencil was rebuilt on every watch recompilation; now runs once on startup only.
- **Remove `--disableFastRender`** (serve.sh): Hugo was re-rendering the entire site on every asset change, causing multi-second rebuild windows. Fast render limits rebuilds to only affected pages. This was added for a reason, so I want to confirm that we're okay with removing it.

I _think_ all of these are safe and should only affect local development. The potential downside is bugs potentially only manifesting in production builds if they're related to minification or other things being skipped in dev.

@cnunciato Please feel free to add anyone else that might want to take a look at these changes.

### DIFF
--- a/theme/webpack.config.js
+++ b/theme/webpack.config.js
@@ -27,6 +27,14 @@ module.exports = function (env, { mode }) {
         devServer: {
             writeToDisk: true,
         },
+        // Disable webpack's module cache in development. When enabled (the default),
+        // changes to SCSS partials imported from main.scss/_marketing.scss don't
+        // invalidate the extracted CSS module — webpack rebuilds the JS stub but
+        // serves stale CSS from cache until the watcher is restarted. See the
+        // output of `make serve-all` showing "cached modules ... 766 KiB (css/mini-extract)"
+        // after a partial change. Disabling the cache in dev mode forces sass-loader
+        // to re-run on every rebuild. Production builds are unaffected.
+        cache: mode === "production",
         module: {
             rules: [
                 {
@@ -75,6 +83,10 @@ module.exports = function (env, { mode }) {
                 filename: "../../assets/css/[name].css",
             }),
             new WebpackShellPluginNext({
+                // dev: true clears the script list after the first run, so
+                // stencil is not rebuilt on every watch recompilation — only on startup.
+                // Has no effect on production builds (single compilation).
+                dev: mode !== "production",
                 onBuildStart: {
                     blocking: true,
                     parallel: false,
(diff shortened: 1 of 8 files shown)


theme/webpack.config.js
```diff
@@ -27,6 +27,14 @@ module.exports = function (env, { mode }) {
...
             writeToDisk: true,
         },
+        // Disable webpack's module cache in development. When enabled (the default),
+        // changes to SCSS partials imported from main.scss/_marketing.scss don't
+        // invalidate the extracted CSS module — webpack rebuilds the JS stub but
+        // serves stale CSS from cache until the watcher is restarted. See the
+        // output of `make serve-all` showing "cached modules ... 766 KiB (css/mini-extract)"
+        // after a partial change. Disabling the cache in dev mode forces sass-loader
+        // to re-run on every rebuild. Production builds are unaffected.
+        cache: mode === "production",
```
cnunciato: For some reason, I'm not able to get consistent behavior out of this. Maybe it's my Mac (an older M2), or there's something else is going on, but yeah, something's not quite working. 

Here's what I see:

1. `make serve-all`, make a change to `main.scss`. 
    1. Nearly all the time, nothing happens.
    2. Rarely, after a minute or so, Webpack rebuilds, and I see the change with a live reload. 
  
2. Make another change, wait several minutes. Nothing happens ever.

I'll poke around. Somehow it feels like there could be multiple things contributing to the mishap, though.

cnunciato: Opus 4.7 seems to have found a fix — at least one that works for me! 

```js
// In some environments, Webpack's default fs.watch-based watcher misses SCSS
// and TS edits, causing `make serve-all` to sit idle after a change. Polling 
// works, and at 300ms, is fast enough that rebuilds still feel responsive. 
// Only active under `webpack --watch`.
watchOptions: {
    poll: 300,
    aggregateTimeout: 200,
    ignored: ["**/node_modules/**", "**/stencil/dist/**"],
},
```

I'll push a commit and you can give it a try @jeffmerrick.

CamSoper (approved):
Hey Jeff -- LGTM. Took a look at the build-config changes and they're all mode-gated correctly: `cache`, `minimize`, and the `WebpackShellPluginNext` `dev` flag only flip in dev mode, and the stable-CSS-URL branch in `assets.html` only triggers under `hugo.IsServer`.

Belt-and-suspenders check: deployed this branch to pulumi-test.io and confirmed the production build path is clean -- CSS bundles still fingerprint (`bundle.workflow-dispatch-d7ed47be.css`), JS still minifies, all assets reachable. No regressions.

Confirming the `--disableFastRender` removal is fine -- the env-var toggle (`DISABLE_FAST_RENDER=true`) preserves the escape hatch for when you actually need it.

One pre-existing nit worth knowing about (not blocking, not yours to fix here):

- `.gitignore` has the `static/js/*.js` patterns duplicated in two blocks (L43 and L55); your additions only landed in the second block, leaving the first out of sync. Either keep it or consolidate eventually.

theme/webpack.config.js
jeffmerrick: Working for me! I previously tested by editing `_templates.scss` and am also running in a container so those could have contributed to the different behavior.

cnunciato (approved):

PR #18659, opened 2026-04-22
Update all possible attributes example with runtime options

andrzejressel:
### Proposed changes

Updates `all possible attributes` example in project file page, so it also includes runtime's options


CamSoper (approved):
Thanks for the contribution, @andrzejressel! Good catch — the previous example used the simple-string form for `runtime`, which left out the `options` attribute that the rest of the page documents. The structured form with `compiler: cue export` is a nice, meaningful demonstration of the yaml runtime's `compiler` option.

PR #18688, opened 2026-04-24
docs: document the template quickstart property

### LINKED ISSUE #11460: Document the `quickstart` property for templates
I have a use case for `template:quickstart:` in my program, and was unaware it existed because it doesn't seem to be documented.

Example this quickstart value is output once I create the template
```yaml
name: ${PROJECT}
description: ${DESCRIPTION}
runtime: nodejs
template:
  quickstart: |
    Please check your program once it is generated.
    This is my multiline string
    Also have fun!
  description: "A minimal Azure Native Typescript Pulumi program"
  important: true
  config:
    azure-native:location:
      description: The Azure location to use
      default: WestUS2

```
References:
https://github.com/pulumi/pulumi/blob/master/sdk/go/common/workspace/project.json

[themes/default/content/docs/concepts/projects/project-file.md](https://www.pulumi.com/docs/concepts/projects/project-file/)

joeduffy:
Fixes #11460.

The `template.quickstart` property has been supported by `pulumi new` for some time but was missing from the project file reference. This change adds a row to the `template` options table describing what `quickstart` does and includes it in the end-of-document YAML example.

When a template author sets `quickstart` in `Pulumi.yaml`, `pulumi new` prints that text to stdout after generating the new project — after the standard "Next steps" block. It is intended for follow-up instructions specific to the template: required environment variables, dependency installation steps, links to related guides, and so on.

I deliberately did not document the `important` template field: it is marked `Deprecated: We don't use this field any more.` in the Pulumi CLI source (`sdk/go/common/apitype/templates.go`), so guiding users toward it would be a regression.

---

### DIFF
--- a/content/docs/iac/concepts/projects/project-file.md
+++ b/content/docs/iac/concepts/projects/project-file.md
@@ -162,7 +162,7 @@ Schemas are only valid for project property keys. For setting the value of a pro
 | - | - | - |
 | `displayName` | optional | A user-friendly name for the template. This should follow `Title Case` format and be succinct. This field is only supported by Pulumi CLI >= 3.95. |
 | `description` | optional | Description of the template. |
-| `quickstart` | optional | Optional text to display after the template is instantiated — for example, after `pulumi new` generates a new project. Use this field to surface follow-up instructions specific to the template, such as required environment variables, dependency installation steps, or links to getting-started guides. |
+| `quickstart` | optional | Text to display after the template is instantiated — for example, after `pulumi new` generates a new project. Use this field to surface follow-up instructions specific to the template, such as required environment variables, dependency installation steps, or links to getting-started guides. |
 | `config` | required | Config to request when using this template with `pulumi new`. |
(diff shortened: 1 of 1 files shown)


joeduffy:
Thanks for the review. I've applied the two substantive suggestions (removed the redundant "Optional" prefix, and trimmed the duplicate "run" in the example).

On the minimum CLI version question: `quickstart` was added in commit `e08aafe4` on 2022-09-03 and first shipped in Pulumi CLI v3.40.0 (released 2022-09-14). That's over three years ago, making it quite unlike `displayName` and `metadata` (which were added in v3.95 and warrant a caveat). Adding a version note for something this old would be more confusing than clarifying, so I'm leaving it without one.

jkodroff (approved):

CamSoper (approved):
LGTM. Verified each claim against the Pulumi CLI source:
- `quickstart` field exists in the template schema (`apitype/templates.go`)
- `pulumi new` prints it after the "Next steps" block (`newcmd/new.go`)
- `important` is indeed deprecated, so deliberately omitting it is the right call

Thanks!

PR #18693, opened 2026-04-24
Document reviewStackLabels for conditionally creating review stacks

zacdirect:
Documents new functionality for configuring a gating review stack creation by requiring a PR having a specific label on it.

### DIFF
--- a/content/docs/deployments/deployments/review-stacks.md
+++ b/content/docs/deployments/deployments/review-stacks.md
@@ -323,3 +323,16 @@ const prMigrationSettings = new pulumiservice.DeploymentSettings("prMigrationSet
 });
 
 ```
+
+### Gating review stacks by GitHub label
+
+By default, every pull request against a configured branch creates a review stack. To restrict this to a subset of pull requests, set `reviewStackLabels` on the GitHub Deployment Settings to a list of labels. Review stacks are only created for pull requests carrying at least one matching label (case-sensitive); labels added after the pull request is opened also trigger creation. Applies to GitHub only.
+
+```typescript
+github: {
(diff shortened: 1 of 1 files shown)


content/docs/deployments/deployments/review-stacks.md
```diff
@@ -323,3 +323,29 @@ const prMigrationSettings = new pulumiservice.DeploymentSettings("prMigrationSet
 });
 
 ```
+
+### Gating review stacks by GitHub label
+
+By default, every pull request against a configured branch creates a review stack. To restrict this to a subset of pull requests, set `reviewStackLabels` on the GitHub Deployment Settings to a list of labels. Review stacks are then created only for pull requests carrying at least one matching label. Matching is case-sensitive, and labels added after a pull request is opened also trigger creation.
```
fnune: Nit/probably a pre-coffee brain thought:

This was a tiny bit hard to parse for me. The punctuation doesn't split the different ideas in the paragraph so well IMO. Here's my (and Claude's) attempt:

```suggestion
By default, every pull request against a configured branch creates a review stack. Set `reviewStackLabels` in your GitHub deployment settings to limit review stack creation to pull requests that carry at least one of the specified labels. Matching is case-sensitive. Labels added after a pull request is opened will also trigger creation.
```

fnune (approved):

CamSoper (approved):
Looks good — verified the docs against the merged pulumi-service feature PR (#41897). Field name `reviewStackLabels`, case-sensitive matching, and the "label-added-after-open triggers creation" claim all match the implementation. Heading and code style are consistent with the rest of the file.

PR #18695, opened 2026-04-27
Persist UTM and gclid params across /gads/ signup clicks

alexleventer:
## Summary

Adds a small inline script to the `/gads/` Google Ads landing pages that forwards the landing-page querystring (UTM params, `gclid`, etc.) onto every link pointing at `app.pulumi.com`, so paid-traffic attribution survives the click into signup. The Neo page's JS-driven redirect (`gadsSubmitNeoQuery`) was updated to do the same.

Files touched:

- [`layouts/gads/baseof.html`](layouts/gads/baseof.html) — inline script that runs on every `/gads/` page and rewrites every `app.pulumi.com` link to carry the current querystring through
- [`layouts/gads/gads-neo.html`](layouts/gads/gads-neo.html) — Neo prompt redirect appends the current querystring to the destination URL

## Why this matters

Without this, the params Google Ads (and other ad platforms) append to the landing-page URL are dropped the moment the visitor clicks "Get Started" / "Try Neo for FREE" and lands on `app.pulumi.com`. That has real cost:

- **Broken attribution.** `gclid` is what Google Ads uses to tie a signup back to the specific ad click. If it doesn't reach the signup event (or the conversion ping that fires after signup), the conversion can't be attributed to the campaign, ad group, or keyword that produced it. We end up over-spending on ads that look like they don't convert and under-spending on ones that actually do.
- **Bid optimization breaks.** Smart Bidding and conversion-based optimization rely on a steady stream of attributed conversions to learn from. Drop the `gclid` and the algorithm trains on a much noisier signal.
- **UTM-based reporting goes blank.** `utm_source`, `utm_medium`, `utm_campaign`, and especially `utm_term` (the matched keyword) are how the marketing team segments funnel performance in analytics. Losing them at the page boundary means downstream funnel analysis can't distinguish paid search from organic, brand from non-brand, or one campaign from another.
- **Enhanced Conversions / offline imports.** When we later upload offline conversion data back to Google Ads, the click ID is the join key. Missing click IDs = unmatched conversions = wasted attribution.

The fix is intentionally narrow: it only touches links pointing at `app.pulumi.com` and runs entirely client-side with no dependencies.

## Test plan

- [ ] Build passes locally (`make build` ✅)
- [ ] Visit `/gads/aws/?utm_source=google&utm_medium=cpc&utm_campaign=test&utm_term=aws+iac&gclid=ABC123` and confirm the header "Get Started" link, hero CTA, and every `btn-secondary` CTA in the page have the params appended
- [ ] Visit `/gads/neo/?utm_source=google&gclid=ABC123` and confirm the "Try Neo for FREE" buttons and the prompt-submit redirect both carry the params through
- [ ] Confirm direct visits (no UTM/gclid in URL) leave links unchanged


alexleventer:
@claude please review

alexleventer:
@claude - take another look pls

CamSoper (approved):
Reconciled the PR description with the inlined implementation — the simplification commits dropped the partial, the param allowlist, and the sessionStorage hydration that the original write-up still described, and the test plan's last bullet was no longer going to pass.

Two minor things worth considering, neither blocking:

- `layouts/gads/baseof.html:7-9` has a "why" comment in inline JS, which `AGENTS.md` says to drop. The why isn't obvious from the code, so I'd lean toward keeping it, but flagging since the rule is what it is.
- The script forwards the whole querystring rather than the `utm_*` / `gclid` / `gbraid` / `wbraid` / `fbclid` / `msclkid` allowlist the original description mentioned. Fine in practice today, but worth a thought if marketing/analytics ever cares about constraining what gets passed through.

Otherwise looks good — CI green, both rewrite paths traced, the `a[href*="app.pulumi.com"]` selector covers all the existing CTAs.

Issue #18717, opened 2026-04-29
Combine Newsroom and Awards into a single filterable Newsroom page


dwahbe:
## Summary

Today, Pulumi's press, news, and awards content lives across two separate pages:
* https://www.pulumi.com/about/newsroom/ — a two-column layout splitting Press Releases and Recent News
* https://www.pulumi.com/awards/ — a standalone page for industry awards and recognitions

I'd propose consolidating these into a single Newsroom page with a filter bar at the top — All | News | Press | Awards — so visitors can browse everything in one place and narrow down to what they're looking for.

## Why
* One destination. Press, analysts, and prospective customers researching Pulumi today have to know that "Awards" is a separate URL. A single Newsroom page is the conventional pattern and easier to link to from media kits, footers, and outreach.
* Awards belong in the same story. Awards are part of the same external-validation narrative as press coverage and releases; surfacing them alongside reinforces Pulumi's momentum rather than hiding them on a separate page.
* The two-column layout doesn't scale. As the press and news lists grow, the side-by-side layout creates uneven column heights and forces visitors to scan two parallel timelines instead of one chronological feed.

## Proposed design

<img width="425" height="783" alt="Image" src="https://github.com/user-attachments/assets/d00e618d-fa00-404d-85b9-55edc2f9ba8a" />

<img width="770" height="766" alt="Image" src="https://github.com/user-attachments/assets/6310e17a-2867-4de3-aba8-404197e50205" />

<img width="1267" height="886" alt="Image" src="https://github.com/user-attachments/assets/9f4ff6e0-323e-42c3-b630-ec4eddfe3f3c" />

<img width="362" height="307" alt="Image" src="https://github.com/user-attachments/assets/6d24aa51-455e-451d-a7ba-7864dee9481b" />

PR #18721, opened 2026-04-29
Combine Newsroom and Awards into a single filterable page

dwahbe:
Addresses #18717, feel free to ignore if not the right direction.

### Proposed changes

This PR consolidates the existing Newsroom (`/about/newsroom/`) and Awards (`/awards/`) pages into a single Newsroom at `/about/newsroom/`, with filter tabs for News, Press, and Awards. The Awards URL aliases to the new page so external links and search rankings carry over.

#### A unified Newsroom page

- **Filter tabs** at the top: All, News, Press, Awards. State lives in the URL hash so individual filters are linkable and the back/forward buttons work.
- **Featured cards** for the most recent items, followed by an **Archive list** for older items.
- The Archive paginates with a "Load more" button (page size 20). Pagination state is per-filter, so switching tabs starts fresh.
- "All" interleaves press releases, news coverage, and awards in chronological order; each tab also has its own featured + archive split computed from its filtered feed.

#### Card design

- A uniform card grid with image, type chip, date, title, and (for press releases) a short summary.
- On the All tab, each card shows a colored chip indicating its type. On a single-type tab, the chips are hidden since the type is implied by the active filter.
- The bold-on-active tab uses a hidden ghost label to reserve width, and the date and chip share matching vertical padding, so toggling tabs doesn't cause horizontal or vertical layout shifts.

#### Content and routing

- `content/about/newsroom.md` becomes the unified page; `content/awards/_index.md` is removed and `/awards/` aliases to the new URL.
- `data/awards.toml` gets a `date` field on each entry (sourced from the original publication or announcement date) so awards sort chronologically alongside press and news.
- Top nav and footer collapse the previous two entries into a single "Newsroom — News, press, and awards" entry.
- The About page (`/about/`) keeps a compact "In the Newsroom" preview using the same card partial, with one item from each category.

#### Implementation notes

- New layout in `layouts/page/newsroom.html` with four small partials in `layouts/partials/newsroom/`:
  - `card.html` — the featured-card layout
  - `archive-row.html` — the compact archive row
  - `type-meta.html` — single source of truth for type → label / badge color / track-name mapping
  - `sort-key.html` — parses the human-readable dates from the data files into a sortable \`YYYY-MM-DD\` key
- Filter behavior in `theme/src/ts/newsroom.ts`. Uses the existing \`analytics.track()\` pattern (see \`llm-menu.tsx\`, \`hubspot-form.tsx\`) to fire \`newsroom-filter\` and \`newsroom-load-more\` events. Existing per-link \`data-track\` attributes carry over.
- Tab styling follows the established \`:target\`-driven pattern from \`_event-list.scss\`, with the gradient-border mixin from \`_mixins.scss\`.


cnunciato:
Hello again @dwahbe! 👋🏻 

Thanks for this submission as well! I generally like this idea, but there's a bunch here to consider, so I'll need to discuss it with the team. We may also want to provide you with some additional guidance to align more closely with the visual updates we've been making across the site lately. Will keep you posted!

dwahbe:
Sounds great! Let me know :)

CamSoper:
@dwahbe -- first off, thank you. Genuinely. You put real work into this (a full Newsroom/Awards rethink across 21 files) and it was a good idea, well executed.

I also owe you an apology: this sat since early May, and that's on us, not you. It slipped through a gap in our review process and nobody picked it back up -- not the experience any contributor should get, least of all a first-time one.

At this point the branch has drifted far enough from current main that rebasing and re-reviewing would be more churn than it's worth, so I'm going to close it. That's a reflection of the calendar, not the quality of what you sent.

If you're up for contributing again, I'd genuinely welcome it, and I'll make sure it doesn't go quiet on you next time. Thanks again for the effort and the patience.

PR #18806, opened 2026-05-05
Add new main nav

jeffmerrick:
A new main nav! It now matches a shadcn style nav that we may use in other codebases. This also required some fixes to how Inter was being loaded as it was rendering differently as a static instead of variable font.

**Considerations:**
- Some of the text descriptions in the dropdowns could maybe be trimmed to be shorter.
- Should we keep "For" in front of engineers and enterprise?
- I have only tested the logged in user functionality by spoofing a cookie, but it appears to work.

---

Moves the header navigation structure into a YAML data file (`data/header_nav.yaml`) and splits the rendering into three focused partials — `header/nav.html`, `header/nav-desktop.html`, and `header/nav-mobile.html`. The old monolithic `top-nav.html` — which used the `pulumi-user-toggle` web component to show "Pulumi Cloud" for signed-in users — is deleted. The docs nav (`docs-top-nav.html`) is left in place and continues to serve `/docs/` and `/tutorials/` pages.

Also adds Inter variable fonts (`InterVariable.woff2`, `InterVariable-Italic.woff2`) and updates `_fonts.scss` to use them.

The sign-in and get-started buttons are still auth-aware: when a user is signed in, both are hidden and replaced by a Dashboard button with a user-circle icon. Auth state is read from the `pulumi_web_user_info` cookie in `header-nav.js`, using the same logic as the existing `header-cta` Stencil component.

Resolves: #18403


vctrfrnndz:
@jeffmerrick the mobile nav starts open for me in the staging env, safari

jeffmerrick:
@vctrfrnndz PurgeCSS was purging some Tailwind classes only on production builds, [should be working now](http://www-testing-pulumi-docs-origin-pr-18806-d03d7b12.s3-website.us-west-2.amazonaws.com/).
 
@cnunciato Re: the above, I think this has been happening without anyone noticing with the new classes available in Tailwind v4 since it only manifests on a production build. It was purging a bunch of stuff including hover states and arbitrary classes. Since Tailwind doesn't generate unused classes anymore, I've just removed PurgeCSS. The file size difference looks to be around +15kb gzipped, but that's also restoring missing css.
 
@claude I've addressed the orphaned code and accessibility. The Dashboard CTA href idea seems like a good one, but I'm not sure so I'll leave that to other reviewers to decide. Please re-review.

vctrfrnndz:
@jeffmerrick This is looking awesome btw!

> PurgeCSS was purging some Tailwind classes only on production builds, [should be working now](http://www-testing-pulumi-docs-origin-pr-18806-d03d7b12.s3-website.us-west-2.amazonaws.com/).

(1) Roger! I would still check that out and make sure we have some css that hides that by default because I still see it flash for a small second and then correct.

(2) Thoughts on maybe decreasing the gray selection color and also using a subtler gray color for the top menu carets?

CleanShot 2026-05-06 at 17 20 59@2x

jeffmerrick:
@vctrfrnndz Good calls, color tweaks added and FOUC should be fixed!

Screenshot 2026-05-06 at 12 31 34 PM

CamSoper:
@jeffmerrick — quick perf check on the variable font, cc'ing @sicarul.

The preload went from ~83 KB (`inter-regular.woff2` 40 KB + `inter-semibold.woff2` 43 KB) to 352 KB (`InterVariable.woff2`). That's a ~4× jump in preloaded font bytes on first paint, directly in the LCP critical path. Have you measured the impact, or have a sense of whether it's hurting on uncached visits?

If it shows up as a real regression, a few cheap mitigations:

- Subset to Latin only (Inter ships per-script subsets, or `pyftsubset` can produce one) — typical drop is 350 KB → 120–180 KB.
- `unicode-range`-split `@font-face` declarations so the browser only fetches glyph ranges in use.
- Pair Inter with a `size-adjust`-tuned fallback to flatten CLS during the swap.

For context: the CSS-bundle delta from the PurgeCSS removal is small (+16 KB gzipped, async-loaded, mostly absorbed by beasties critical-CSS inlining), so the font preload is the bigger perf knob in this PR.

jeffmerrick:
@CamSoper Good catch, I swapped in the variable font because it renders slightly differently than where I'm using it elsewhere, but not worth the extra KB - I can just use the static fonts elsewhere also. 

Also noticed I was using the variable version of Monaspace Neon, I think we're really only using the regular weight, so I also swapped that in so we can save ~74KB there (not much, but not nothing!)

data/header_nav.yaml
cnunciato, line 1: Nice — love having this in YAML instead of pages of HTML! 🙌🏻

cnunciato: One thing, though: What do you think about keeping the ordering as we have it today, putting Registry beside Docs?

- Product
- For Engineers
- For Enterprises
- Docs
- Registry
- Blog
- Pricing
- Company

Also, what do you think about adding Contact Us to the Company menu? (Just since we're losing it from the top nav.)

data/header_nav.yaml
```diff
@@ -0,0 +1,155 @@
...
+
+plainLinks:
+  - { label: Registry, href: /registry/, track: header-registry }
+  - { label: Blog,     href: /blog/,     track: header-blog }
+  - { label: Pricing,  href: /pricing/,  track: header-pricing }
+  - { label: Docs,     href: /docs/,     track: header-docs }
+
+cta:
+  signIn:    { label: Sign in,    href: "https://app.pulumi.com/signin", track: header-signin }
+  getStarted: { label: Get started, href: "https://app.pulumi.com/signup", track: header-get-started }
```
cnunciato: I kind of like [Claude's suggestion](https://github.com/pulumi/docs/pull/18806#issuecomment-4382600977) of having the Dashboard button link to the app at the root (`/`), rather than `/signup`. That extra hop seems unnecessary for someone who's already signed in.

Any reason not to do that @alexleventer? Seems like this ought to be fine, but I don't want to break any metrics we might be measuring with this particular link. (This would only affect clicks to "Dashboard", which is only shown to signed-in users.)

layouts/partials/header/nav-desktop.html
```diff
@@ -0,0 +1,59 @@
...
+  <ul class="flex items-center gap-0">
+    {{ range $i, $dropdown := $nav.dropdowns }}
+      <li>
+        <button type="button"
+          data-nav-trigger-button
+          data-nav-index="{{ $i }}"
+          aria-expanded="false"
+          aria-haspopup="menu"
+          aria-controls="nav-desktop-popup"
+          class="group/trigger inline-flex h-9 items-center rounded-md px-4 py-2 text-sm font-regular text-gray-950 whitespace-nowrap outline-none transition-all hover:bg-gray-100/70 focus:bg-gray-100/70 focus-visible:ring-3 focus-visible:ring-violet-400/50 focus-visible:outline-1 data-[open=true]:bg-gray-100/70">
```
cnunciato: Any reason not to add `cursor-pointer` to these, and to the links? I rather like them myself, so I usually add them, but if there's a trend away from them, always flexible.

layouts/partials/header/nav-desktop.html
```diff
@@ -0,0 +1,59 @@
...
+          {{ partial "icon.html" (dict "name" "caret-down" "weight" "bold" "class" "text-gray-500 relative top-px ml-1 size-3 transition duration-300 [[data-open=true]_&]:rotate-180") }}
+        </button>
+      </li>
+    {{ end }}
+
+    {{ range $link := $nav.plainLinks }}
+      <li>
+        <a href="{{ $link.href }}"
+          {{ with $link.track }}data-track="{{ . }}"{{ end }}
+          class="inline-flex h-9 items-center justify-center rounded-md px-4 py-2 text-sm font-regular text-gray-950 whitespace-nowrap transition-all hover:bg-gray-100/70 focus:bg-gray-100/70 focus-visible:ring-3 focus-visible:ring-violet-400/50">
```
cnunciato: Also here, I think (for the pointer).

cnunciato (commented):
This looks awesome. A few questions we should work out first, but otherwise, huge improvement. 🙌🏻 ✨

layouts/partials/header/nav-desktop.html
jeffmerrick: There is apparently [some debate around this](https://github.com/tailwindlabs/tailwindcss/issues/8961) I wasn't aware of. Personally, I prefer having them as well, so I just added a global rule to add it to all `button` elements.

data/header_nav.yaml
jeffmerrick: Makes sense to me! Reordered and added contact us into the company menu (and it's tracking attribute).

cnunciato: One thing @jeffmerrick: Could we pull the Contact Us link out of the menu after all and position it to the left of the the Sign In / Dashboard CTAs? @alexleventer wants to make sure we don't bury that, as it sounds like it's commonly for sales inquiries.

data/header_nav.yaml
cnunciato: Talked to @alexleventer and we can link directly to `/` for the signed-in user (so no need to pass through `/signup` from the Dashboard button).

jeffmerrick: Done!

data/header_nav.yaml
jeffmerrick: Done! I've also made it so it's still visible next to the "Dashboard" button when the user is logged in.

cnunciato (approved):

PR #18811, opened 2026-05-06
docs(ai): document Neo browser and audio notifications

kramhuber:
### Proposed changes

Documents the per-user browser and audio notification toggles added to Neo in [pulumi/pulumi-service#39653](https://github.com/pulumi/pulumi-service/pull/39653).

Adds a new "Notifications" section to the Neo Settings page, framed explicitly as user-level (not org-level) settings. Covers:

- The two toggles (browser notifications, audio notifications), both off by default.
- When notifications fire (final assistant message or pending user approval) and that they are suppressed while the Pulumi Cloud tab has focus.
- Browser permission flow, including the toggle reverting if the user declines or has previously blocked notifications.
- The OS-level requirement that notifications be enabled for the browser, since browser permission alone is not sufficient.

Also updates the page intro to reflect that Neo configuration spans organization, repository, and user levels, and refreshes ``meta_desc`` accordingly.

### Related issues (optional)

Tracks pulumi/pulumi-service#39653.

### DIFF
--- a/content/docs/ai/settings/_index.md
+++ b/content/docs/ai/settings/_index.md
@@ -2,7 +2,7 @@
 title: Settings
 title_tag: Neo Settings
 h1: Neo Settings
-meta_desc: Configure Neo with Custom Instructions, Repository Instructions, Slash Commands, Integrations, access controls, and task modes.
+meta_desc: Configure Neo with Custom Instructions, Repository Instructions, Slash Commands, Integrations, access controls, task modes, and notification preferences.
 meta_image: /images/docs/meta-images/docs-meta.png
 menu:
     ai:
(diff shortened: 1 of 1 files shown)


CamSoper (approved):
All facts cross-checked against pulumi/pulumi-service#39653 — defaults, triggers (`is_final` AssistantMessage / UserApprovalRequest), `document.hasFocus()` suppression, the two-note chime, the toggle/permission flow, and the Neo-enabled-org gating all line up. Style is clean.

Two optional polish items, fine as a follow-up:
- L15: "at the organization, repository, and user level" reads slightly awkward across three distinct levels — your PR description uses the plural ("levels") and that scans more naturally to me.
- The source PR also covers an "unsupported browser" message and notes that the "Enable notifications…" warning persists whenever permission is denied (e.g., revoked outside the page), not only after a toggle attempt. Both are edge cases — worth a sentence each if you want full coverage.

PR #18816, opened 2026-05-06
Document scaling/concurrency and complete config reference for customer-managed runners

borisschlosser:
## Summary

- Add a **Scaling and concurrency** section to the customer-managed workflow runners page, explaining that each runner process executes one workflow job at a time, that scaling is done by adding runner instances, and that Pulumi Cloud claims each pending job exclusively (so two runners never process the same job simultaneously, with crash recovery once a claim expires). Includes three concrete scaling patterns: long-running replicas, ephemeral `single_run` jobs, and specialized pools via `enabled_workflow_types`.
- Expand the **Configuration reference**:
  - Add the previously undocumented `health_threshold` / `PULUMI_AGENT_HEALTH_THRESHOLD` setting.
  - Add a top note covering the `PULUMI_AGENT_*` env-var convention, env-over-file precedence, and Go-style duration syntax.
  - Reorganize the YAML block into logical groups (required, OIDC, polling/retry, health/observability).
  - Tighten descriptions on existing keys (e.g. `deploy_target`, `circuit_breaker_failures`, `single_run`, `env_forward_allowlist` always-forwarded `DOCKER_HOST`).

## Test plan

- [ ] Render the page locally (`make serve` or equivalent) and confirm formatting.
- [ ] Verify the new "Scaling and concurrency" section renders under "Using customer-managed workflow runners".
- [ ] Verify the YAML configuration block parses cleanly and the new `health_threshold` entry is visible.

### DIFF
--- a/content/docs/deployments/deployments/customer-managed-agents.md
+++ b/content/docs/deployments/deployments/customer-managed-agents.md
@@ -53,15 +53,18 @@ Workflow runners support multiple workflow types beyond deployments, including P
 
 ### Scaling and concurrency
 
-Each workflow runner process executes **one workflow job at a time** and has no internal worker pool to configure. To increase the number of jobs your pool can run in parallel, add more workflow runner instances to the pool — each instance contributes one additional concurrency slot.
+Each workflow runner process runs **one deployment at a time**, plus optionally **one Insights scan or policy evaluation in parallel**, and has no internal worker pool to configure. To increase the number of jobs your pool can run in parallel, add more workflow runner instances to the pool — each instance contributes one deployment slot and, if the pool also handles non-deployment workflow types, one additional slot for Insights scans or policy evaluations.
 
-Pulumi Cloud assigns each pending job to exactly one runner using an exclusive claim. When multiple runners poll the same pool simultaneously, the service hands each pending job to a single runner, so the same job is never processed by two runners at the same time. If a runner crashes or loses connectivity in the middle of a job, the claim eventually expires and another runner in the pool picks up the job.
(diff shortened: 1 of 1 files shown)


borisschlosser:
@claude review again

borisschlosser:
@claude collaps the "each runner" sentences to one as suggested by yourself

fnune (approved):

CamSoper (approved):
LGTM — the YAML reorganization and the new "Scaling and concurrency" section are real readability wins, and I spot-checked the defaults and the `health_threshold` derivation against the agent source; both match.

Three small things you might want to fold into a follow-up (none are blockers):

1. **Line 149** — `# Required unless using OIDC (see oidc_token_file below).` is a directional reference. The earlier review pass caught the same pattern with "above" on line 207 and replaced it with a section name. Easiest fix here is to drop "below" entirely: `# Required unless using OIDC.` reads fine since `oidc_token_file` is named again a few lines later.

2. **Line 208** — backtick consistency in the OIDC group header: 

  ```markdown
  ## organization_name and runner_pool_id are required, and `token` is not used.
  ```
   ...puts backticks on `token` but not on the other keys. Either backtick all or none.

3. **Line 99 (outside this PR's diff)** — the OIDC prose bullet still says `token_expiration` is "the expiration in seconds for the tokens requested by the workflow runner," but the agent reads it via `GetDuration` (`config.go:154`), and the new general note at line 141 establishes Go duration syntax across the board. Worth tightening to something like "the lifetime of tokens requested by the workflow runner (Go duration syntax, e.g. \`1h\`)" so the page tells one consistent story.

PR #18855, opened 2026-05-08
Improve header nav keyboard and screen reader accessibility

jeffmerrick:
### Proposed changes

Tightens keyboard and screen-reader behavior on the header nav: a skip-to-main-content link, an `inert`-managed Tab order for the desktop popup, mobile sheet, and mobile collapsibles, full Arrow/Home/End/Escape/Tab handling on desktop dropdowns (including correctly walking past plain-link nav items like Docs/Registry/Blog/Pricing on forward Tab), and focus return to the sheet trigger on close. Also drops `aria-haspopup="menu"` since the popup isn't a menu pattern, polishes focus-visible styles, and shows Contact us above Dashboard in the mobile sheet for signed-in users to match the desktop CTAs.

### Related issues (optional)

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -2,11 +2,12 @@
 <html lang="en-US" prefix="og: http://ogp.me/ns#">
     {{ partial "head.html" . }}
     <body class="section-{{ .Section }}">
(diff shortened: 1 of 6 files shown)


cnunciato (approved):
Looks great!

Issue #18847, opened 2026-05-08
Missing margin on https://www.pulumi.com/about/


iwahbe:
This is what it looks like on mobile (iPhone 17 Pro):

<img width="1206" height="2622" alt="Image" src="https://github.com/user-attachments/assets/0625756e-438d-4837-b6d4-4ee7b47e06dc" />

We should have the same margin on this page as the rest of our site.

jeffmerrick:
Fixed in https://github.com/pulumi/docs/pull/18857

PR #18857, opened 2026-05-08
Update about page hero and migrate images to fingerprinted assets

jeffmerrick:
### Proposed changes

- Fix page heading and navigation extending to edges of the page on mobile.
- Replace the team photo hero with new photo, switch the rendered `` to use the `fingerprinted-img.html` partial.
- Move `journey.svg` from `static/images/about-us/` to `assets/fingerprinted/images/about-us/` so it gets fingerprinted/responsive treatment.
- Update the `fingerprinted-img.html` partial to always emit `alt` (defaulting to empty) so callers can pass `alt=""` for decorative images without the attribute being dropped.

### Related issues

Refs #18847


iwahbe (approved):
It fixes the margin thing.

PR #18860, opened 2026-05-08
Fix mobile nav disabling cookie consent, refactor consent dialog styles

jeffmerrick:
### Proposed changes

The mobile nav sheet was marking the cookie consent banner and preferences dialog `inert`, making them unreachable while the sheet was open. `setBackgroundInert` in `header-nav.ts` now skips `#segment-consent-manager` and `.consent-dialog-overlay` so those higher-priority overlays stay interactive.

Also refactors `_consent-banner.scss`: simplified container layout (mobile hugs the bottom edge, desktop floats with `max-w-xl`), updated borders and hover states, restyled the secondary button as neutral gray, and added section comments.

### DIFF
--- a/theme/src/scss/_consent-banner.scss
+++ b/theme/src/scss/_consent-banner.scss
@@ -1,70 +1,78 @@
 @import "mixins";
 
-#segment-consent-manager:empty {
-    @apply hidden;
-}
-#segment-consent-manager {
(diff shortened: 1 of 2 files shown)


jeffmerrick:
To test in US, spoof location in Chrome devtools `Sensors` section.

cnunciato (approved):

PR #18868, opened 2026-05-08
Redesign site footer with data-driven nav, CTA, and statuspage badge

### LINKED ISSUE #18537: Global: Update Footer

jeffmerrick:
### Proposed changes

Replaces the legacy footer with a new dark-themed design driven by `data/footer.yml`. Adds a persistent footer CTA section, live Pulumi statuspage badge (fetched client-side with a 5-minute localStorage cache), GitHub star button, and inline HubSpot newsletter signup. Removes redundant inline `footer_cta` sections from the homepage and the IaC product page since every page now has a CTA in the footer. Also refactors `.btn`/`.btn-orange` styles to work correctly with Tailwind v4's `@utility`/`@layer` ordering, drops the unused `btn-orange-translucent`, and tweaks the global CTA labels to sentence case.

---

TODO outside of this PR: add placeholder to email signup form field in Hubspot.

Resolves: https://github.com/pulumi/docs/issues/18537


cnunciato (approved):
This looks great! I made a few minor adjustments (I hope you don't mind) and one fix to account for duplicate CTAs in some of our existing templated pages, for example:

image

I opted to keep your `footer/cta` partial, drop the existing one, and then to plumb through any page specific CTA frontmatter params into your partial. That should give each page the ability to customize those slots, falling back to the site-wide, configured primary and secondary CTAs by default.

I also made a few tweaks to the footer menu just to tighten things up a bit and to keep the link to Security, which is an important for various reasons (inquiries, reports, etc.).

Lastly, something about the `muted` violet felt just a little bit too bright to me (I'm not sure why — maybe set against the purple, or with the black text, or something), so I figured I'd try something slightly lighter for it. What do you think of this `background` violet instead? It feels a little softer on the eyes to me, but just a suggestion.

If that slight visual tweak does work for for you, then I think this is good to go! Leaving you an approval to merge when you're ready, but feel free to make any additional adjustments you need to before you do so.

🙌🏻

layouts/partials/footer/github-star.html
```diff
@@ -1,7 +1,7 @@
 {{/* buttons.github.io/buttons.js is loaded in layouts/partials/head.html. */}}
 <a class="github-button"
    href="https://github.com/pulumi/pulumi"
-   data-color-scheme="no-preference: dark; light: dark; dark: dark;"
+   data-color-scheme="no-preference: light; light: light; dark: light;"
```
cnunciato: @jeffmerrick Also, what do you think about this, just to help the widget stand out a little more? The dark version definitely works better aesthetically, but it does get a bit lost against the dark background.

image

jeffmerrick: Made some changes that should address this!

jeffmerrick:
@cnunciato Good with your changes! I like the CTA override setup and the lighter purple is good.

Screenshot 2026-05-11 at 9 13 19 AM

---

I went ahead and added a supabase-style custom component that we can render in both the header nav and the footer - it updates the star count at build time.

Couple of questions/notes:
- Should we add back case studies? I accidentally removed it. It could maybe go under 'Pulumi for' and we rename that to 'Solutions'? Or just go under 'Resources'?
- The nav now collapses at 1345px which is pretty wide but I think will be okay (a small laptop might be 1440px wide) However, we might look into what we can prune in the future to keep if from getting too crowded.

@claude Re-review

cnunciato:
>However, we might look into what we can prune in the future to keep if from getting too crowded.

@jeffmerrick Yep, agree. First on my list would be the Contact Us button; if the data don't show that getting used, we can pull that into the Company menu to win back some space. We'll keep an eye on it. 

>Should we add back case studies? I accidentally removed it. It could maybe go under 'Pulumi for' and we rename that to 'Solutions'? Or just go under 'Resources'?

Hm, yeah. It does seem valuable to have something there. How about something like "Customers" — or maybe better/slightly more fun, "Teams like yours"? 🤔

PR #18875, opened 2026-05-11
docs(ci): refresh "Adding CI/CD support" guide

borisschlosser:
## Summary

Refresh `content/docs/iac/guides/continuous-delivery/add-support-for-cicd-systems.md` so it reflects the current state of [`sdk/go/common/util/ciutil`](https://github.com/pulumi/pulumi/blob/master/sdk/go/common/util/ciutil) in `pulumi/pulumi`:

- Fix inline references to the package path. The body text said `pkg/util/ciutil`, but the package lives at `sdk/go/common/util/ciutil`. (The existing links were already pointing at the correct path; only the prose was wrong.)
- Describe the `detectors` collection as a **slice** (`[]system`) rather than a map, and update the `AppVeyor` Go example so it matches the current slice-literal syntax — the previous `AppVeyor: baseCI{…}` map-literal form wouldn't compile against today's code.
- Use the accurate Go term "embed `baseCI`" instead of "add `baseCI`".
- Expand the metadata properties list to include **Build ID**, **Build Number**, and **Branch Name** — all already exposed by the `Vars` struct.
- Expand the generic fallback env var list to include the four entries that `generic.go` reads but the doc was missing: `PULUMI_CI_BRANCH_NAME`, `PULUMI_CI_BUILD_NUMBER`, `PULUMI_COMMIT_MESSAGE`, `PULUMI_PR_NUMBER`.
- Add a note clarifying that `PULUMI_CI_BRANCH_NAME` and `PULUMI_COMMIT_MESSAGE` are **fallbacks**, not overrides: when the CLI runs inside a Git repo with a non-detached HEAD and a non-empty commit message, the Git values win and these two env vars are ignored. They take effect for detached-HEAD checkouts (typical of CI runners), Mercurial repos, or when no repo is present.

## Test plan

- [x] `make lint` passes
- [x] Page renders correctly under `make serve` at `/docs/iac/guides/continuous-delivery/add-support-for-cicd-systems/`
- [x] Existing aliases still resolve

### DIFF
--- a/content/docs/iac/guides/continuous-delivery/add-support-for-cicd-systems.md
+++ b/content/docs/iac/guides/continuous-delivery/add-support-for-cicd-systems.md
@@ -31,11 +31,14 @@ The Pulumi CLI already supports enhanced metadata for several popular CI/CD syst
 The detection of metadata in a CI environment depends on some key environment variables that a CI system injects into the build environment of their build agents. The Pulumi CLI uses those environment variables to try and determine the values for the following properties:
 
 - CI System Name
+- Build ID
+- Build Number
 - Build Type
 - Build URL
 - Commit SHA
-- PR Number
+- Branch Name
 - Commit Message
+- PR Number
 
 The metadata about your CI environment is displayed in the [Pulumi Cloud](https://app.pulumi.com/signin) stack activity log. The metadata from your CI environment and the information about your Git repository allow Pulumi to provide links to pull requests and commits.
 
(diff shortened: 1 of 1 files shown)


content/docs/iac/guides/continuous-delivery/add-support-for-cicd-systems.md
```diff
@@ -90,9 +93,23 @@ That's it! Send us a new [PR](https://github.com/pulumi/pulumi/pulls) in the [`p
...
+- `PULUMI_CI_BRANCH_NAME`
 - `PULUMI_CI_BUILD_ID`
+- `PULUMI_CI_BUILD_NUMBER`
 - `PULUMI_CI_BUILD_TYPE`
 - `PULUMI_CI_BUILD_URL`
 - `PULUMI_CI_PULL_REQUEST_SHA`
+- `PULUMI_COMMIT_MESSAGE`
+- `PULUMI_PR_NUMBER`
+
+You can also find these variables in the [`generic.go`](https://github.com/pulumi/pulumi/blob/master/sdk/go/common/util/ciutil/generic.go) file in the [`sdk/go/common/util/ciutil`](https://github.com/pulumi/pulumi/blob/master/sdk/go/common/util/ciutil) folder.
```
fnune: Nit: use permalinks pointing to a commit instead of `master`

fnune (approved):

content/docs/iac/guides/continuous-delivery/add-support-for-cicd-systems.md
borisschlosser: On reflection, reverted back to `blob/master/...` in 732815155d. Since these links point at whole directories/files rather than specific code lines, `master` keeps the doc tracking the live code; pinning to a SHA would just freeze the references and let the page fall behind reality over time.

PR #18895, opened 2026-05-12
Add sitewide announcement banner

### LINKED ISSUE #18885: Sidewide banner
An announcement banner above the nav that appears on all pages.

jeffmerrick:
Adds a sitewide banner with a few configurable options.
- Multiple announcements can be set, if there are more than one, one is chosen at random on page load.
- They can be dismissed (if set as dismissible). Their `id` is stored in localstorage so that specific announcement is not shown to the user again.
- They can be set to just show on marketing pages, docs pages or both (default). If we don't want these on the docs by default we could change that.

I have a few blog posts added as announcements to show how it works, **we'll need to clear those out before we merge.**

Resolves #18885 
Resolves https://github.com/pulumi/docs/pull/18895

---

## Summary

Adds a configurable sitewide announcement banner above the main header on every page (marketing + docs).

- **Configurable** from `data/announcements.yml` — each entry has `active`, `display` scope (`marketing` | `docs` | `both`), optional `icon` *or* `badge` (mutually exclusive), Markdown-formatted `text`, optional `cta`, and `dismissible`.
- **Random rotation** across active+in-scope+not-dismissed entries on each page load.
- **No FOUC / no layout shift (CLS = 0):** randomization happens in a tiny inline `<script>` that runs *before* the wrapper is parsed, injecting an `!important` reveal rule so the banner paints in its final state on first render. If the visitor has dismissed everything, the wrapper stays `display: none` and reserves no space.
- **Dismiss swaps** in the next eligible entry instead of just hiding; only hides when nothing's left. Dismissed ids persist in `localStorage`.
- **Marquee** on narrow viewports — text auto-scrolls (with a seamless clone, no blank gap) while the icon/badge stay on the left and the CTA + close button stay fixed on the right. Respects `prefers-reduced-motion`.
- **A11y:** wrapper is `role="region" aria-label="Site announcement"`; dismiss button is labeled; clone is `aria-hidden`.

**Cleanup:** removes the commented-out `banner.html`, the unused `home/dismissible-banner.html` + its SCSS, and the obsolete `banner-state` dismiss block in `nav.ts`.

**Drive-by fix:** header logo CLS — added `width`/`height` and `loading="eager"` to the marketing and docs nav logos so the browser reserves aspect-ratio space before the SVG loads.

## Test plan

- [ ] Visit `/` and `/docs/`; confirm banner renders above the header on both, scrolls away while the marketing header sticks
- [ ] Reload several times; verify different announcements appear (random rotation across the placeholders)
- [ ] Dismiss the visible announcement → confirm the next eligible one takes its place; dismiss again → wrapper hides; reload → stays hidden; clear `localStorage` → reappears
- [ ] Resize to narrow width → text marquees seamlessly; icon, CTA, and X stay fixed on the right
- [ ] Toggle a YAML entry to `active: false`, `display: docs`, or set `dismissible: false` and confirm behavior matches
- [ ] Set both `icon` and `badge` on an entry → build fails with a clear `errorf`
- [ ] DevTools → Performance / Web Vitals: CLS = 0 on first paint for both "banner shown" and "all-dismissed" cases


theme/src/ts/announcement-banner.ts
```diff
@@ -0,0 +1,136 @@
...
+
+        const setupMarquee = (el: HTMLElement) => {
+            if (marqueeCleanup) {
+                marqueeCleanup();
+                marqueeCleanup = null;
+            }
+            const marquee = el.querySelector<HTMLElement>("[data-announcement-marquee]");
+            const content = el.querySelector<HTMLElement>("[data-announcement-content]");
+            const text = el.querySelector<HTMLElement>(".announcement-text");
+            if (!marquee || !content || !text) return;
```
cnunciato, line 50: Tiny nit and less important because TS, but I'd have expected the linter to catch this. 🤔 Maybe something to look into later.

cnunciato (approved):
Love it. Clear out the array and merge when ready. My only suggestion would be to include an example block or two in the usage instructions just to make it a little easier for folks to copy and paste into place.

Also, one question we can totally come back to in a follow-up: Would it make sense to extend this YAML-driven model to the hero widget on the home page also (the one that currently says "Meet Neo")?

theme/src/ts/announcement-banner.ts
jeffmerrick: @cnunciato I'm not sure there is a linter set up for JS/TS? Just prettier config?

jeffmerrick:
@cnunciato Good idea, example block added!

I'd probably lean towards the home hero one being more static but I could see something like an animated badge that rotates through a few working there.

PR #18909, opened 2026-05-12
Fix mobile social carousel tap-vs-drag handling

### LINKED ISSUE #18861: Homepage quote carousel mobile padding
Carousel has extra left padding on mobile.

https://github.com/user-attachments/assets/5e0a7311-7083-4c07-a79e-f131a2d3cf60

jeffmerrick:

Fixes the social carousel and the case study carousel on the homepage so they loop cleanly on mobile instead of dragging into empty whitespace past the last slide. Both components are rendered by the shared `<pulumi-swiper>` Stencil component, so the swiper-level fix covers both.

**Root cause.** The pointer-move handler was applying the raw drag delta directly to the wrapper transform with no bounds. Dragging past the first or last cloned slide pulled the wrapper beyond its content extent, exposing whitespace, and because the loop teleport only runs on `transitionend`, the carousel had no way to recover during an in-progress drag. The previous mitigation — `allow-touch-move=\"false\"` in the markup — was inert because the Stencil component never declared a matching `@Prop`, so touch dragging was never actually disabled.

**What changed in `swiper.tsx`:**

- Added `clampTranslate` to bound the wrapper translate to the content extent during drag, including the `centeredSlides` offset so other pages using `centered-slides=\"true\"` clamp to the right bounds.
- Defer drag commit until pointer movement crosses a 5px threshold; capture the pointer only after that. This keeps taps on slide anchors from being retargeted away from links by `setPointerCapture`.
- After a committed drag, install a one-shot capture-phase `click` suppressor on the container (with a 300ms `setTimeout` fallback for touch devices that don't fire a trailing click) so the user doesn't accidentally open the card they were swiping.
- Add a `duration` parameter to `slideTo` / `advance` and snap post-drag in 300ms instead of inheriting `speed` (which is tuned to 8000ms for the slow autoplay drift).

**What changed in the template:** removed the inert `allow-touch-move=\"false\"` attribute from all five `<pulumi-swiper>` instances in the social carousel (mobile / tablet / desktop variants). The case study carousel template is unchanged — it picks up the swiper fix automatically.

### Related issues

Fixes #18861

### DIFF
--- a/layouts/partials/template-partials/template-social-carousel.html
+++ b/layouts/partials/template-partials/template-social-carousel.html
@@ -32,7 +32,7 @@
     {{/* Mobile: 1 slide at a time */}}
     <div class="template-social-carousel-rows block md:hidden">
         <div class="template-social-carousel-row">
-            <pulumi-swiper slides="1" centered-slides="false" loop="true" autoplay="true" autoplay-delay="3000" speed="8000" free-mode="true" allow-touch-move="false">
+            <pulumi-swiper slides="1" centered-slides="false" loop="true" autoplay="true" autoplay-delay="3000" speed="8000" free-mode="true">
                 {{ range $tweet := $tweets }}
                 <pulumi-swipeable>
                     <div class="w-full px-4">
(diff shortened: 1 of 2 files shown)


cnunciato (approved):
👏🏻

PR #18923, opened 2026-05-13
[blog] Add post on agent-friendly Markdown registry docs

fnune:
Announce the five public Pulumi Cloud registry documentation endpoints
(`/readme`, `/installation`, `/nav`, `/docs/{token}`, `/examples`),
their Markdown response format with size savings over JSON, the `Link:
...; rel="alternate"` advertisement on JSON responses, and the `pulumi
api` docs hint that `pulumi package add` now emits when it detects an
AI coding agent. Targets publish on 2026-05-19.

Also swap `--format=markdown` to `--output=markdown` (and `--format=json`
to `--output=json`) in `content/docs/iac/cli/cloud-api.md` and
`layouts/index.llms.txt` to match the `pulumi api` flag rename in
https://github.com/pulumi/pulumi/pull/23072. The post relies on the new
flag name; the existing docs and llms.txt copies were stale.

## Test plan

- `make lint` passes
- `make serve` renders the post at `/blog/agent-friendly-registry-docs-api/`
- URLs and `pulumi api` examples in the post were verified against the
  live `api.pulumi.com` (public packages, no auth needed) and against a
  local build of pulumi/pulumi master (v3.238.0-dev), including `Link:`
  header presence, `Cache-Control: immutable` on resolved versions,
  content negotiation behavior, and the two token-encoding forms

PR #18927, opened 2026-05-13
ci: validate pulumi/esc-action fix (pin to 197ccaa)

borisschlosser:
## Summary

Validates the upstream fix in [pulumi/esc-action@197ccaa](https://github.com/pulumi/esc-action/commit/197ccaa42ab49560ce838a3010eb8345ce086896) ("Fix oidc-auth default breaking ESC_ACTION_OIDC_AUTH env-var fallback", from [pulumi/esc-action#43](https://github.com/pulumi/esc-action/pull/43), fixes [pulumi/esc-action#42](https://github.com/pulumi/esc-action/issues/42)) by pinning every esc-action invocation in this repo to that exact SHA.

**A green CI on this PR is the validation signal**: if jobs authenticate successfully against the pinned SHA *without* any consumer-side workaround, the upstream fix is working.

## Background

The `pulumi/esc-action` `v1` tag was moved on 2026-05-13 ~02:26 UTC (when [pulumi/esc-action#41](https://github.com/pulumi/esc-action/pull/41) landed) to a commit that included an earlier regression from [pulumi/esc-action#35](https://github.com/pulumi/esc-action/pull/35) — `oidc-auth` was declared in `action.yml` with `default: 'false'`.

The action's input/env helper is:

```ts
function getInput(name: string, envVar: string): string | undefined {
    const val = core.getInput(name) || process.env[`ESC_ACTION_${envVar}`];
    ...
}
```

Once `core.getInput('oidc-auth')` started returning the non-empty string `'false'` by default, the `|| process.env[...]` fallback never fired — silently disabling OIDC for every consumer that configured the action via the `ESC_ACTION_OIDC_AUTH=true` env-var pattern. `esc` then errored out with:

```
##[error]`esc open` command failed:
Error: could not determine current cloud: PULUMI_ACCESS_TOKEN must be set for login during non-interactive CLI sessions
```

This broke every open PR (#18925, #18923, #18920, ...) and scheduled jobs.

The upstream fix changes the `oidc-auth` default in `action.yml` from `'false'` to `''`, matching the other `oidc-*` inputs, so `core.getInput` returns falsy and the env-var fallback works again.

## What this PR does

1. **Commit 1** added a consumer-side workaround (`with: oidc-auth: true` on every invocation) to unblock CI immediately.
2. **Commit 2** removes that workaround and pins every `pulumi/esc-action@v1` reference to `@197ccaa42ab49560ce838a3010eb8345ce086896` so this PR's CI exercises *only* the upstream fix.

24 workflow files updated; all 30 workflow YAML files parse OK.

## Follow-up

Once the `v1` tag is moved to include the fix, swap the pinned SHA back to `@v1` in a follow-up PR.

## Test plan

- [ ] CI on this PR passes — proves the upstream fix restores the env-var fallback
- [ ] Spot-check `pull-request.yml`, `build-and-deploy.yml`, `pulumi-cli.yml`, and `esc-cli.yml` show the pinned SHA and no `with:` blocks on esc-action steps
- [ ] After validation, follow-up PR to move pin back to `@v1`

### DIFF
--- a/.github/workflows/add-to-project.yml
+++ b/.github/workflows/add-to-project.yml
@@ -11,9 +11,7 @@ jobs:
     steps:
       - name: Fetch secrets from ESC
         id: esc-secrets
-        uses: pulumi/esc-action@v1
-        with:
-          oidc-auth: true
+        uses: pulumi/esc-action@197ccaa42ab49560ce838a3010eb8345ce086896
       - name: Add to Docs
         uses: actions/add-to-project@v1.0.2
         with:
--- a/.github/workflows/add-to-project.yml
+++ b/.github/workflows/add-to-project.yml
@@ -12,6 +12,8 @@ jobs:
       - name: Fetch secrets from ESC
         id: esc-secrets
         uses: pulumi/esc-action@v1
+        with:
+          oidc-auth: true
       - name: Add to Docs
         uses: actions/add-to-project@v1.0.2
         with:
--- a/.github/workflows/bucket-cleanup-testing.yml
+++ b/.github/workflows/bucket-cleanup-testing.yml
@@ -19,6 +19,8 @@ jobs:
       - name: Fetch secrets from ESC
         id: esc-secrets
         uses: pulumi/esc-action@v1
+        with:
+          oidc-auth: true
       - uses: actions/checkout@v6
 
       - uses: actions/setup-node@v6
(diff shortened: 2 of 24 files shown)


fnune (approved):
TY

borisschlosser:
Pushed v1 tag in esc-action repo which includes the actual fix: https://github.com/pulumi/esc-action/pull/43

PR #18931, opened 2026-05-13
docs(concepts): clarify the difference between direct-form and output-form provider functions

### LINKED ISSUE #12373: Clarify the difference between getFoo and getFooOutput
There are two forms of "provider functions" -- the "direct form", which accepts plain/prompt inputs, and the "output form", which accepts Pulumi inputs. The example and content at the top of the [Provider functions page](https://www.pulumi.com/docs/concepts/resources/functions/) mention this, but both could be clearer in that the example wraps the function in `pulumi.output` (it's unclear why) and the bullets fail to express that the two "forms" are actually two different functions (e.g., `getZone` and `getZoneOutput`), rather than overloads of the same function.

We should clarify this and provide an example of where and why you'd choose to use one over the other.

joeduffy:
## What changed

This PR makes two changes to the [Provider functions](https://www.pulumi.com/docs/iac/concepts/functions/provider-functions/) concepts page to address the gaps described in issue #12373.

**Change 1 — clarify that the two forms are distinct functions, not overloads**

A new paragraph has been added immediately before the "Direct form" subsection to state plainly that in most languages the two variations are two separately named functions rather than overloads of a single function. It includes the concrete naming examples (`aws.ec2.getAmi` / `aws.ec2.getAmiOutput`), calls out Java's inverted convention (`getAmi()` is the output form, `getAmiPlain()` is the direct form), and notes that YAML uses `fn::invoke` for both.

**Change 2 — concrete code examples showing when to choose each form**

Two new choosable code blocks have been added at the end of the "Choosing between direct form and output form" section, one per form:

- The direct-form example uses `getAmiIds` to look up AMIs and gates instance creation on whether any matching AMIs were found — illustrating that the direct form returns plain values suitable for conditional logic.
- The output-form example uses `getAmiOutput` (and its per-language equivalents) with a config secret (`requireSecret`) as a filter value — illustrating that the output form accepts Pulumi `Input` values directly, avoiding an `apply()` wrapper.

Both examples use AWS EC2 AMI lookups, consistent with the existing examples on the page, and are provided for TypeScript, Python, Go, C#, Java, and YAML.

## Verification

- Front-matter unchanged; exactly two `---` delimiters confirmed.
- All `{{< chooser >}}` and `{{% choosable %}}` tags balanced (5 chooser blocks, 30 choosable blocks).
- No trailing whitespace.
- File ends with a newline.
- `provider-functions.md` is hand-authored content, not auto-generated.

Fixes #12373

---

### DIFF
--- a/content/docs/iac/concepts/functions/provider-functions.md
+++ b/content/docs/iac/concepts/functions/provider-functions.md
@@ -236,6 +236,8 @@ Bridged providers, which take a Terraform provider as an underlying dependency, 
 
 Provider functions are exposed in each language as regular functions, in two variations:
 
+In most languages, the two variations take the form of **two distinct, separately named functions** rather than overloads of a single function. For example, the AWS function `aws.ec2.getAmi` has a corresponding output form named `aws.ec2.getAmiOutput`. Java's naming convention is inverted: `Ec2Functions.getAmi()` is the _output_ form, while `Ec2Functions.getAmiPlain()` is the direct form. In YAML, both forms are invoked using `fn::invoke`, and the runtime handles the distinction transparently.
+
 ### Direct form
 
 The **direct form** accepts plain arguments (e.g., `string`, as opposed to `pulumi.Input<string>`):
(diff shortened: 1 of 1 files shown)


joeduffy:
Thanks for the detailed review! Addressed all feedback in the follow-up commit:

- **Colon → period**: Fixed the dangling colon at the section lead-in.
- **Redundancy**: Removed "distinct" from "two distinct, separately named functions".
- **Python filter style**: Updated the direct-form Python example to use `GetAmiIdsFilterArgs` consistently with the output-form example.
- **YAML direct-form**: Dropped YAML from the direct-form chooser entirely and replaced it with an info note explaining that YAML does not distinguish between direct and output forms at the language level.
- **Java wildcard imports**: Replaced `import com.pulumi.*;` with explicit `import com.pulumi.Context; import com.pulumi.Pulumi;` in both new Java examples.
- **apply wrapper**: Changed to "no apply call required" without backticks.

@claude

CamSoper (changes requested):
Two things in the new "Using the output form" examples before this merges:

**1. Go example won't compile (line 615)**

```go
amiNameFilter := cfg.RequireSecret(ctx, "amiNameFilter")
```

The `*Config.RequireSecret` method takes only `key`:

```go
func (c *Config) RequireSecret(key string) pulumi.StringOutput
```

(`pulumi/sdk/go/pulumi/config/config.go:155`). The two-arg form is the package-level `config.RequireSecret(ctx, key)`, not the method on `cfg`. Easiest fix:

```go
amiNameFilter := cfg.RequireSecret("amiNameFilter")
```

**2. AMI field inconsistency**

The new examples reference `latestAmi.id` (or `${latestAmi.id}` in YAML). The existing examples on the same page all use `.imageId` (lines 45, 69, 107, 150, 191, 223). Both fields exist on `GetAmiResult`, so neither is wrong — but the inconsistency between the new block and the rest of the page is jarring. Worth picking one and standardizing, probably `.imageId` since that's what the rest of the page already uses.

**Small wording suggestion (non-blocking)**

The sentence on line 359 says "no apply call required" but the C# and Java output-form examples below still need `.Apply(a => a.Id)` / `.applyValue(ami -> ami.id())` to extract the AMI ID. Tightening to something like "no apply call required to pass the secret into the filter" would head off the misread.

joeduffy:
Thanks for the thorough review. All three points addressed in d4ac90d:

1. **Go RequireSecret**: Fixed to `cfg.RequireSecret("amiNameFilter")` — the method form takes only the key, not `(ctx, key)`.
2. **AMI field consistency**: Standardized all new output-form examples to `.imageId` / `.image_id` / `.ImageId()` across TypeScript, Python, Go, C#, Java, and YAML, consistent with the existing examples on the page.
3. **Wording**: Tightened to "no apply call required to pass the secret into the filter" — making clear the scope of what is avoided (the apply wrapper around the secret input), since C# and Java still use apply to extract the AMI ID from the Output result.

CamSoper (approved):

PR #18971, opened 2026-05-13
docs(yaml): reconcile YAML config schema with project file reference

### LINKED ISSUE #12417: Inconsistencies in YAML config docs
### Problem description

I noticed a few inconsistencies in our YAML docs about the project-level config.

- [Pulumi YAML reference](https://www.pulumi.com/docs/languages-sdks/yaml/yaml-language-reference/) right at the top links to [config options](https://www.pulumi.com/docs/reference/pulumi-yaml/#config-options), which says that the type property can be “either string, boolean, integer, or array.” But the “Config” section on Pulumi YAML reference says “It can be one of: String, Number, List<Number>, or List<String>.”
  - Note the capitalization, too.
- Also, [config options](https://www.pulumi.com/docs/reference/pulumi-yaml/#config-options) lists type as required whereas Pulumi YAML reference doesn’t (“Either type or default is required.“). Nevertheless the description says “Type is the (**required**) data type …”
- The “Config” section on Pulumi YAML reference doesn’t mention the variant where a config key maps directly to a value (foo: bar) instead of a structured object, which is documented on [config options](https://www.pulumi.com/docs/reference/pulumi-yaml/#config-options).

### Suggestions for a fix

Apart from fixing the issues above, I wonder if we could avoid having two sections on config on two different pages? It's YAML either way.

rshade:
## Summary

- Aligns the YAML language reference with the canonical project-level `config` schema documented in `project-file.md`, resolving the long-standing inconsistencies flagged in #12417 (mismatched type names, drifted property tables, and a stale anchor link).
- Replaces the YAML page's Title-Case type list (`String`, `Number`, `List<String>`, etc.) with the lowercase types the engine actually accepts today: `string`, `integer`, `boolean`, `array`, and `object`, plus the `items` form that replaces `List<T>`.
- Moves the legacy `configuration` key guidance into a warning callout that documents the full deprecated schema, so existing users can still recognize their programs while new programs get steered to the supported form.
- Adds a note on `project-file.md` clarifying that the `config` schema there is the canonical, engine-parsed schema used by every language (including YAML), with a cross-link to the YAML page for YAML-flavored examples.

## Test plan

- [ ] `make lint` clean (project's markdownlint-cli2 over built output)
- [ ] Visual spot-check via `make serve`: rendered `{{% notes %}}` callouts on both pages, in-page `#config` anchor from the top property table, and cross-links between the two pages
- [ ] Spot-check that the example YAML program still parses with `pulumi preview` against a scratch project

### Modified files

- `content/docs/iac/languages-sdks/yaml/yaml-language-reference.md` — Fix the top-of-page `config` link to use the in-page `#config` anchor; rewrite the Config section to mirror the project-level schema (lowercase types, `items` row, secret guidance); move the `configuration` deprecation into a warning callout with the full legacy type list; refresh the example program to use the supported schema and add a `boolean` example.
- `content/docs/iac/concepts/projects/project-file.md` — Add a note above `### config options` clarifying that this schema is parsed by the engine and applies to every language, with a link to the YAML reference.

Closes #12417

### DIFF
--- a/content/docs/iac/concepts/projects/project-file.md
+++ b/content/docs/iac/concepts/projects/project-file.md
@@ -121,6 +121,10 @@ Alternatively, a package value can be an object with the following properties:
 
 ### `config` options
 
+{{% notes %}}
+This is the project-level `config` schema parsed by the Pulumi engine. It applies to programs in every language — TypeScript, JavaScript, Python, Go, .NET, Java, and YAML — including the `config` block of a Pulumi YAML program. See the [Pulumi YAML reference](/docs/iac/languages-sdks/yaml/yaml-language-reference/#config) for a YAML-flavored example.
+{{% /notes %}}
+
 `config` is a map of config property keys to either values or structured declarations.
 
 Non-object values are allowed to be set directly. Anything more complex must be defined using the structured
@@ -138,7 +142,7 @@ Schemas are only valid for project property keys. For setting the value of a pro
 
 | Name | Required | Description |
 | - | - | - |
-| `type` | required | The type of this config property, either `string`, `boolean`, `integer`, `array`, or `object`. |
+| `type` | optional | The type of this config property, either `string`, `boolean`, `integer`, `array`, or `object`. Either `type` or `default` is required; when omitted, the type is inferred from `default`. |
 | `description` | optional | A description for this config property. |
 | `secret` | optional | True if this config property should be a secure secret value. |
 | `default` | optional | The default value for this config property, must match the given type. |
--- a/content/docs/iac/languages-sdks/yaml/yaml-language-reference.md
+++ b/content/docs/iac/languages-sdks/yaml/yaml-language-reference.md
@@ -23,7 +23,7 @@ Pulumi supports programs written in YAML or JSON.  In both cases, the programs (
 
 | Property | Type | Required | Expression | Description |
 | - | - | - | - | - |
-| `config` | [config options](/docs/reference/pulumi-yaml/#config-options) | No | No | Config specifies the [Pulumi config](/docs/concepts/config/) inputs to the deployment. |
+| `config` | [config options](#config) | No | No | Config specifies the [Pulumi config](/docs/concepts/config/) inputs to the deployment. |
 | `resources` | map[string]Resource | No | No | Resources declares the [Pulumi resources](/docs/concepts/resources/) that will be deployed and managed by the program |
 | `variables` | map[string]Expression | No | Yes | Variables specifies intermediate values of the program, the values of variables are expressions that can be re-used. |
 | `outputs` | map[string]Expression | No | Yes | Outputs specifies the [Pulumi stack outputs](/docs/concepts/stack#outputs) of the program and how they are computed from the `resources` is a value of the appropriate type for the template to use if no value is specified. |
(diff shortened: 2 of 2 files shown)


CamSoper (changes requested):
Hi @rshade — thanks for tackling this; harmonizing the type lists is genuinely useful and a real step on #12417. Before it lands, I'd like to revisit a few framing claims.

**Heads-up on provenance**: my fact-checker (a subagent I dispatched against the Pulumi sources) flagged the items below — I haven't independently verified the line citations. If anything below is wrong, push back.

#### 1. The "engine canonical schema, every language" framing is overstated

`project-file.md:125` and `yaml-language-reference.md:42`:
> "This is the project-level `config` schema parsed by the Pulumi engine. It applies to programs in every language..."

The canonical JSON schema does limit types to `string`/`boolean`/`integer`/`array`/`object` (`pulumi/pulumi` → `sdk/go/common/workspace/project.json`, `simpleConfigType` enum).

But the YAML *runtime* has its own parser (`pulumi/pulumi-yaml` → `pkg/pulumiyaml/config/config.go`, `Parse()`) that accepts a superset: `number`, `bool`, `int`, and `list<X>` — none of which are in `project.json`. So the engine isn't the single source of truth for what *runs*; the YAML runtime is more permissive for backward compatibility.

Suggest softening to: "This is the canonical schema. The YAML runtime accepts a few legacy forms for backward compatibility, but new programs should use the canonical types shown here."

#### 2. The Title-Case story conflates two separate things

`yaml-language-reference.md:55`:
> "In beta, Pulumi YAML projects used a top-level \`configuration\` key, which accepted a different schema with Title-Case types..."

Two distinct deprecations are bundled into one claim:

- **The `configuration` top-level key is deprecated.** Confirmed: `pulumi-yaml` → `pkg/pulumiyaml/run.go` emits a runtime warning when the `configuration` block is present (something close to `"Pulumi.yaml: root-level \`configuration\` field is deprecated; please use \`config\` instead."`).
- **Title-Case type names** (`String`, `Number`, `List<String>`) are *not* tied to the `configuration` key. The runtime parser calls `strings.ToLower(s)` before matching, so those forms still work under `config` today — without any deprecation warning we could find.

Suggest splitting the callout: (a) `configuration` key deprecation (with the real warning text), (b) older type-name forms still work via case-insensitive matching but new programs should use canonical lowercase forms, (c) `List<T>` and `number` aren't part of the canonical schema even though the runtime accepts them.

#### 3. "Closes #12417" overstates resolution

#12417 listed three bullets. This PR resolves bullet 1 (capitalization). Two remain:

- **Bullet 2** (`type` required-or-not): The YAML page (line 49) now says `type` Required = "No" with prose "Either `type` or `default` is required." `project-file.md` (line 145) still says `type` "required". Same disagreement Thomas flagged.
- **Bullet 3** (value form on the YAML page): The simple `key: value` form is documented on `project-file.md` (lines 135–137) but not on the YAML reference. Specifically called out in the issue.

Either fold those in or downgrade the link to "Addresses #12417 (partial)" so the issue stays open for the rest.

#### Optional, while you're in there

`project-file.md:145` claims `type` is "required", but `project.json`'s `configTypeDeclaration` has no `required` array — `type` is optional and inferable from `default`. Pre-existing, but your new cross-link now sends YAML readers to that table.

Thanks again — happy to iterate on any of this.

CamSoper (approved):
Chatted with @rshade, my agent-generated findings were wrong.

LGTM

rshade:
Hey @CamSoper — thanks for the thoughtful review and the explicit "push back if any of this is wrong" invitation. I took you up on it and verified each claim against a real `pulumi-yaml` program (random `RandomPet`, `pulumi up` end-to-end, local backend). The fact-checker subagent was right on some things and wrong on others — details below.

## What I verified

I built a small YAML program that exercises each disputed form and ran `pulumi preview`/`up` against it. The CLI's `project.json` schema validation runs **before** the YAML runtime ever loads the program, which turns out to be the key fact:

### 1. "Engine canonical schema, every language" framing — empirically accurate

Tested `type: number`, `type: bool`, `type: int`, `type: list<string>` under `config`. All four are rejected at validation time:

```
error: could not validate 'Pulumi.yaml': ... errors occurred:
  * #/config/countNumber/type: value must be one of "string", "integer", "boolean", "array", "object"
```

The YAML runtime's `Parse()` may be permissive at the parser level, but it never sees these — the CLI blocks them upstream. So from a user's perspective the canonical schema **is** the operative source of truth. Softening to "the YAML runtime is more permissive" would invite readers to try forms that don't actually work, so I'd like to keep the current framing.

### 2. Title-Case types under `config` — does **not** work in practice

Tested `type: String` under top-level `config`:

```
error: could not validate 'Pulumi.yaml':
  * #/config/envTitleCase/type: value must be one of "string", "integer", "boolean", "array", "object"
```

Tested the same `type: String` (plus `Integer`, `List<String>`) under the deprecated `configuration` block — works fine, with exactly the warning text you quoted:

```
Warning: Pulumi.yaml: root-level `configuration` field is deprecated; please use `config` instead.
```

So the `strings.ToLower(s)` call in the runtime parser is real, but unreachable from `config` because `project.json` gatekeeps the type strings. Title-Case is empirically tied to the `configuration` key. The current warning callout reflects what users actually experience, so I'd like to keep it as-is rather than splitting it.

### 3a. `type` required on `project-file.md:145` — confirmed, fixed

You're right, this disagrees with the schema (no `required` array on `configTypeDeclaration`) and with the YAML page. Verified that `default` alone (no `type`) deploys cleanly with the type inferred. Updated the row to `optional` with: *"Either `type` or `default` is required; when omitted, the type is inferred from `default`."*

### 3b. Simple `key: value` form on the YAML page — confirmed, added

Verified `greeting: hello` deploys and resolves to `"hello"`. Added a one-sentence note plus a shorthand line in the example program on the YAML page.

## Net result for #12417

With 3a and 3b folded in, the three bullets in the original issue are all addressed:
- Bullet 1 (capitalization) — original PR
- Bullet 2 (`type` required disagreement) — fixed now
- Bullet 3 (simple value form on YAML page) — fixed now

So I'd like to keep "Closes #12417" rather than downgrading to partial.

Happy to keep iterating if any of the above doesn't match what you're seeing — if you have a repro where Title-Case works under `config` without a `configuration` block, I'd love to see it because that would change the framing.

PR #18981, opened 2026-05-14
Make each Python SDK its own independent sphinx docset

### LINKED ISSUE #18934: Decouple CLI and SDK docs generation workflow
A release in the `pulumi/pulumi` repository generates the Pulumi CLI docs, and the SDK docs, and the ESC SDK docs in a monolithic workflow based on a single `repository_dispatch`. This is not terribly intuitive and because everything is thrown into a single workflow, an error can cause a lot work to potentially need to be redone.

- Create individual workflows triggered via `repository_dispatch` that generate each kind of docs:
  - Pulumi CLI
  - ESC CLI
  - Pulumi SDKs (one per language that uses the current monolithic workflow)
  - Pulumi Policy SDKs (which are in `pulumi-policy`)
  - ESC SDKs (one per language that uses the current monolithic workflow)
- Change the monolithic workflow so that it dispatches to the finer-grained workflow to ensure operational continuity.

In order to test this, we will need to run the workflows in GitHub itself. Normally, when we do this, we generate a PR that should be automatically merged. However, during development, we do not want PRs to be automatically merged - we will examine their output in the PR to verify their correctness. Once we are sure the automatically generated PRs are correct, we will then turn them back on to be automerged.

The recent PR that adds docs generation for the Pulumi .NET and Java SDKs should be the model for this work, but it's ok to include iterative improvements upon those workflows.

The generated workflows should have consistent and intuitive naming, e.g. "Pulumi CLI docs", "Pulumi SDK docs - TypeScript", "ESC SDK docs - Python", etc.

IMPORTANT: Do not change the public interface of the current `repository_dispatch` workflow. We will submit changes to repos that emit `repository_dispatch` and change them to use `workflow_dispatch` directly.

When this work is finished, we will need to be sure to create issues in all repositories that fire the `repository_dispatch`: `pulumi/pulumi` and `pulumi/esc-sdk` at least, and probably `pulumi/esc`.

If the scope of this work is too much of a change at one time to be easily understood, we can cut out a smaller slice of the work and push that PR through to completion.

jkodroff:
Fixes #18934

## Summary

Fixes the regression where PR #18980's preview site showed only `pulumi` package content without sibling-package sidebar links. The per-package builds were correctly producing isolated output, but the *layout* assumed a shared docset spanning all packages.

This PR switches to **Option 2** from our discussion: each Python SDK is now its own independent sphinx docset.

### What changed

- **Per-package sphinx source dirs.** Each package gets its own `conf.py` + `index.rst`:
  - `tools/pydocgen/source/pulumi/`
  - `tools/pydocgen/source/pulumi_policy/`
  - `tools/pydocgen/source/pulumi_esc_sdk/`
- **Version is now displayed** in the rendered docs. Each `conf.py` reads its package's installed version via `importlib.metadata.version()` and sets `version` / `release` so the RTD theme renders it in the sidebar.
- **Simpler script.** `scripts/generate_python_docs.sh` picks `./source/${PACKAGE}/` and builds directly to `static-prebuilt/docs/reference/pkg/python/${PACKAGE}/`. No temp dirs, no fabricated one-package toctree.
- **Hugo hub page.** New `content/docs/reference/pkg/python/_index.md` renders the parent `/docs/reference/pkg/python/` page as a hand-curated list of the three SDKs. Replaces the role of the sphinx-generated unified index. Hugo runs after the static-prebuilt copy in `build-site.sh`, so this Hugo page wins the path collision on next deploy.

### Trade-offs

- ✅ Truly independent regens. A broken `pulumi_policy` release on PyPI no longer blocks `pulumi` docs.
- ✅ Brings Python parity with TypeScript, which already works as independent docsets.
- ✅ Each rendered page now shows the SDK version.
- ⚠️ No automatic cross-references between packages. `pulumi_policy` referencing `pulumi.Resource` will render as a code span rather than a link. In practice these refs are rare and the trade-off is worth the decoupling. If we miss them later, intersphinx is a clean follow-up.

### Orphaned files

The shared top-level sphinx artifacts (`static-prebuilt/docs/reference/pkg/python/{_static,genindex,search,...}/`, `objects.inv`, `searchindex.js`, the old `index.html`) become orphans after this PR. They're still referenced by the *existing* `pulumi/index.html`, `pulumi_policy/index.html`, `pulumi_esc_sdk/index.html` on master (which link to `../_static/`, `../genindex/`, etc.).

As each per-package workflow runs post-merge, the new self-contained per-package output will replace those files. After all three packages have been regenerated, the top-level sphinx assets can be deleted in a follow-up. They're harmless dead weight in the meantime.

### Pre-merge (already done)

- [x] `bash -n scripts/generate_python_docs.sh` — shell syntax OK.
- [x] All three `conf.py` files parse cleanly as Python (project, version, release fields populate).
- [x] `make lint` clean.

### Post-merge

1. **Visit the new Hugo hub.** Browse to https://www.pulumi.com/docs/reference/pkg/python/ (or the preview site) and confirm the page lists the three SDKs as a Hugo-rendered hub, not the old sphinx index.

2. **Regenerate each Python SDK with the new layout.** Each workflow now produces a self-contained docset; the resulting PR's diff should rewrite many files in `static-prebuilt/docs/reference/pkg/python/<pkg>/`.

   \`\`\`bash
   gh workflow run pulumi-sdk-python-docs.yml         --repo pulumi/docs --ref master -f version=3.238.0
   gh workflow run pulumi-policy-sdk-python-docs.yml  --repo pulumi/docs --ref master -f version=1.20.0
   gh workflow run pulumi-esc-sdk-python-docs.yml     --repo pulumi/docs --ref master -f version=0.13.1
   \`\`\`

   For each PR, **strip `automation/merge` label and disable auto-merge** before reviewing (consistent with our usual test posture):

   \`\`\`bash
   gh pr edit <N> --remove-label automation/merge --repo pulumi/docs
   gh pr merge <N> --disable-auto --repo pulumi/docs
   \`\`\`

3. **Verify each preview site.** For each generated PR's preview:
   - `/docs/reference/pkg/python/<pkg>/` renders the package's docs.
   - The rendered sidebar/header includes the package version (e.g. "3.238.0" for pulumi).
   - The page does *not* try to link to siblings (those navs are intentionally gone).

4. **Verify the hub page.** After all three packages regenerate, the `/docs/reference/pkg/python/` URL should still serve the Hugo hub (Hugo runs second; takes precedence).

### Rollback

If something is broken, `git revert` this commit. The old per-package builds (with shared parent index) will resume producing output that links to the still-present shared assets.


CamSoper (approved):
stampy stamp

PR #19039, opened 2026-05-15
Add skip_auto_merge flag and no-changes guard to doc-gen workflows

jkodroff:
Every SDK/CLI doc-generation workflow now accepts a skip_auto_merge workflow_dispatch input (default false) so a maintainer can produce a PR without auto-merging it during testing. The three workflows that were repository_dispatch-only (esc-cli, pulumi-cli-dev-version, customer-managed-workflow-agent-cli) also gain workflow_dispatch triggers with the same version-resolution fallback already used by the .NET and Java SDK workflows.

Five workflows that previously ran `git commit` unconditionally now follow the canonical has_changes guard pattern used by the other seven, so re-running for an already-published version exits clean instead of failing the commit step. All twelve doc-gen workflows are now consistent.

### DIFF
--- a/.github/workflows/customer-managed-workflow-agent-cli.yml
+++ b/.github/workflows/customer-managed-workflow-agent-cli.yml
@@ -11,6 +11,15 @@ on:
   repository_dispatch:
     types:
       - customer-managed-workflow-agent
+  workflow_dispatch:
+    inputs:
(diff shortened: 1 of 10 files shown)


CamSoper (approved):
LGTM

PR #19062, opened 2026-05-15
Test: add Performance considerations section to assets-archives docs

CamSoper:
## Summary

Second test PR for the docs-review CI pipeline. Adds a multi-paragraph "Performance considerations" section (~52 line diff) to `content/docs/iac/concepts/assets-archives.md`, with two subsections, a configuration table, and a TypeScript code example.

This supersedes #19052, which got classified `review:trivial` because the prior change was a single long markdown line (4 diff lines total). This one is chunky enough to clear the trivial gate.

**Heads up: the entire added section is fabricated.** None of the following reflect real Pulumi behavior — they exist to give the fact-checker, code-examples specialist, and prose-pattern reviewer something to chew on:

- Version-introduction claims (`v3.84`, `v3.92`, `v3.95`)
- Cache directory `~/.pulumi/asset-cache/`
- Environment variables `PULUMI_ASSET_CACHE_SIZE`, `PULUMI_ASSET_CACHE_TTL`, `PULUMI_ASSET_CACHE_DIR`
- CLI subcommand `pulumi cache clear --type=assets`
- Default thresholds (512 MiB, 7-day TTL, 100 MiB streaming threshold, 40% memory reduction)
- TypeScript API `pulumi.asset.prefetchAll(...)` with `concurrency` / `verifyChecksums` options
- SHA-256 stack-state recording for assets

Will not be merged.

## Test plan

- [ ] Triage classifies as non-trivial and routes to the docs domain
- [ ] docs-review pinned comment posts with a full review (not just prose flagging)
- [ ] Fact-checker flags the fabricated version numbers, env vars, defaults, and CLI subcommand
- [ ] Code-examples specialist flags `pulumi.asset.prefetchAll` as a non-existent API

PR #19068, opened 2026-05-15
fix(claude-workflows): use PULUMI_BOT_TOKEN for checkout so commits trigger downstream workflows

CamSoper:
## Summary

Pushes made by claude-code-action were authenticated as \`github-actions[bot]\` because \`actions/checkout\` uses the default \`GITHUB_TOKEN\` when no \`token:\` is supplied. claude-code-action inherits that auth for \`git push\` regardless of the \`github_token\` input passed under \`with:\` (that input only configures the action's REST client). GitHub's well-known rule then blocks any push by \`GITHUB_TOKEN\` from triggering downstream workflows.

Caught on pulumi/docs#19062 — commit [a63d3c86](https://github.com/pulumi/docs/pull/19062/commits/a63d3c86ac7a2029e90cc3b1b0eef98f1f641618), made in response to an \`@claude #update-review\` comment, came in as \`github-actions[bot]\` / \`claude[bot]\` and the build / deploy / social-review workflows silently skipped it.

## Changes

For each affected workflow, ESC fetch is moved ahead of \`actions/checkout\` so the bot token is available when checkout configures \`.git/config\`. Then the checkout step gets \`token: \${{ steps.esc-secrets.outputs.PULUMI_BOT_TOKEN }}\`. After this, any \`git push\` claude-code-action makes goes out as pulumi-bot, which downstream workflows treat as a regular user push.

Applied to all four claude workflows:

- \`.github/workflows/claude.yml\` — ad-hoc \`@claude\` mentions
- \`.github/workflows/claude-update.yml\` — \`@claude #update-review\`
- \`.github/workflows/claude-code-review.yml\` — initial pinned review (no ESC step existed; added one)
- \`.github/workflows/claude-new.yml\` — kept consistent for safety; this workflow doesn't push today (only dispatches claude-code-review.yml via \`gh workflow run\`, which is already authenticated with PULUMI_BOT_TOKEN), but a future change could add a push path

## Test plan

- [ ] Merge, then \`@claude #update-review\` on a test PR
- [ ] Confirm the resulting commit shows pulumi-bot as committer (not github-actions[bot])
- [ ] Confirm Build and deploy, Run Example Code Tests, and Social Media Review all fire on the new commit

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -120,6 +120,15 @@ jobs:
       checks: write
 
     steps:
+      # ESC runs before checkout so the bot token can authenticate the
+      # checkout — pushes made by claude-code-action later in the workflow
+      # then go out as pulumi-bot rather than github-actions[bot],
+      # which is what lets downstream workflows (build-and-deploy, social
+      # review, etc.) fire on those commits.
+      - name: Fetch secrets from ESC
+        id: esc-secrets
+        uses: pulumi/esc-action@v1
+
       # Check out the PR head, not the base. workflow_run carries the
       # originating commit on the event payload; workflow_dispatch
       # doesn't, so claude-new.yml passes it through as an input.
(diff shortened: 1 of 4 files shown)


CamSoper:
@claude I made an update since you started your review. Please

#update-review

PR #19083, opened 2026-05-16
Fix local dev: make mise optional, drop Hugo from mise.toml

cnunciato:
Human Chris here: This change makes local dev work (including `make lint-prose`) on the Mac by making the recommended dev workflow use `mise` by default (having it install all required tooling) or, when `mise` isn't installed, falling back to allow devs to use their own self-managed binaries. 

This is local dev only, CI is unaffected.

--

## Why

`make ensure` was hard-failing on Mac for two reasons:

1. **Vale was a required check** in `scripts/ensure.sh` despite not being a build dep. Devs without Vale couldn't run the site at all.
2. **`mise install` itself fails on Mac** because Hugo's macOS distribution is a signed `.pkg` (no `.tar.gz` for mise's default aqua backend to unpack — Hugo did this intentionally for code signing / notarization).

The result: the documented setup path didn't work, and the "fix" wasn't obvious.

## What

- **Pin Hugo via mise on every platform** using OS-conditional tool entries. Linux uses the official aqua backend; macOS uses the [`NeoHsu/asdf-hugo`](https://github.com/NeoHsu/asdf-hugo) plugin (small third-party shell script that extracts Hugo's binary from the `.pkg`). Net result: `brew install mise && make ensure && make serve` works on any platform with no extra steps. If the asdf-hugo plugin ever breaks, `brew install hugo` is a clean fallback that `ensure.sh`'s version check handles.
- **Make mise optional.** The Makefile's `$(MISE)` variable expands to `mise exec --` when mise is present, empty string when not. **CI keeps working with zero workflow changes** — same Hugo via peaceiris, same Node via setup-node, etc.
- **Drop the Vale check from `ensure.sh`.** Vale is only needed for `make lint-prose`, where missing-binary errors are loud enough.
- **Fix `mapfile` in `lint-prose.sh`** — bash 4+ only, broken on macOS bash 3.2. Replaced with portable read loop.
- **README rewrite** to a single happy path: install mise, run `make ensure`, run `make serve`.

## Test plan

- [x] `mise install` succeeds on Mac (Hugo installs via asdf-hugo plugin)
- [x] `make ensure` completes on Mac end-to-end, no warnings
- [x] `make serve` starts Hugo on :1313, HTTP 200
- [x] `make lint-prose` runs (no `mapfile` error)
- [x] Simulated CI (no mise on PATH): `$(MISE)` empty, scripts run plain, identical to pre-PR behavior
- [x] Real CI passes on this PR


sicarul (approved):
LGTM

CamSoper:
@cnunciato Sorry about that!

cnunciato:
No worries! Hopefully this'll work and looks right — lemme know.

PR #19101, opened 2026-05-18
Remove stale auto-generated pulumi cloud api command docs

fnune:
The `pulumi cloud api` command was renamed to top-level `pulumi api`
in the CLI. PR #18887 renamed the custom docs and `llms.txt` to match,
and deliberately left the auto-generated command reference pages for
the next regen run to overwrite.

Upstream renamed the command rather than aliasing it, so the regen at
Pulumi@3.239.0 emitted new `pulumi_api*.md` pages but did not touch
the old `pulumi_cloud_api*.md` ones, leaving three orphan pages that
still document the obsolete `--format=...` flag (now `--output=...`).

Delete the orphans and strip the stale `api` subcommand mention and
SEE ALSO link from `pulumi_cloud.md`.

## Test plan

- `make build` and `make lint` clean
- `grep -r pulumi_cloud_api content/ data/ layouts/` returns no hits
- `/docs/iac/cli/commands/pulumi_cloud/` no longer has a dead link to
  `pulumi_cloud_api/`


jkodroff (approved):
CLI docs are generated by a GH workflow and are not intended to be changed manually. Approving b/c getting the right information on the site is more important, but worth noting the following:

1. The generated `pulumi cloud` command page does not have the CLI version number and has the "auto generated" tag - the GHA workflow fixes this.
2. This is a fix for a deeper issue: either the upstream CLI docs are wrong or the automated process has a bug.

fallimic (approved):

PR #19186, opened 2026-05-19
ci(cli-versions): drive table from data/versions.json, enable local refresh

### LINKED ISSUE #19183: Make CLI versions table data-driven from data/versions.json
## Problem

The Pulumi CLI versions/changelog table at `content/docs/install/versions.md` is maintained by a `sed` substitution in `.github/workflows/pulumi-cli-docs.yml` (line 114) that appends one row per workflow run. This approach has several issues:

- **Brittle path coupling.** When `versions.md` was moved from `content/docs/get-started/download-install/` to `content/docs/install/`, the workflow silently broke until #19181 repointed the `sed` path. Hugo aliases don't redirect filesystem writes.
- **Gaps on failure.** When the workflow fails (e.g., today's run for v3.242.0: https://github.com/pulumi/docs/actions/runs/26097909723), the table is missing the row and there's no backfill. Re-running the workflow with the older version regenerates command pages and overwrites `static/latest-version`, which can regress the site if newer versions have already shipped docs.
- **Non-idempotent re-runs.** `versions.md` currently has duplicate rows for 3.241.0 and 3.239.0 from re-runs that each \`sed\`-appended a fresh entry.
- **Date is workflow-run date, not release date.** Rows record \`\$(date +%Y-%m-%d)\` from the runner, not the GitHub release's \`publishedAt\`. Usually the same day, but not always.

## Proposal

Make the table fully derived from \`data/versions.json\`, which is already regenerated from GitHub release tags on every workflow run by \`scripts/get-versions.js\`. It contains the version string, ISO date, per-platform downloads, and checksums URL for every release — 602 entries today, going back to v0.1 (deeper than the 472 rows currently in \`versions.md\`).

Concretely:

1. Add a Hugo partial or shortcode (e.g., \`changelog-table-body\`) that iterates \`site.Data.versions\` and emits one \`<tr>\` per entry, reusing the same column layout as the existing \`changelog-table-row\` shortcode.
1. Replace the hand-maintained \`<tbody>...</tbody>\` block in \`content/docs/install/versions.md\` with a single call to that partial/shortcode.
1. Remove the \"Update version lists\" step (\`sed -e ...\`) from \`.github/workflows/pulumi-cli-docs.yml\`. \`scripts/get-versions.js\` already runs on every invocation; that becomes the only source of truth.

After the conversion, the table is regenerated wholesale from upstream release data on every CLI docs workflow run — no gaps, no duplicates, no manual append, and no filesystem-path coupling between the workflow and the markdown file.

## Caveats

- **Date shift.** Existing rows use the workflow-run date; the conversion will swap to GitHub's \`publishedAt\`. A handful of rows will shift by a day across timezones. Acceptable one-time visible diff.
- **More entries shown.** \`data/versions.json\` has 130 more historical releases than \`versions.md\` currently shows (back to v0.1). Decide whether to render all of them, cap at a minimum version, or paginate.
- **Removed flexibility.** The current shortcode supports per-row flags like \`exclude=\"windows\"\` (used on \`0.16.3\`). The data-driven renderer needs to either bake that into the data file (preferred — extend \`get-versions.js\`) or carve out hand-maintained overrides.
- **Pre-GitHub releases.** Any Pulumi CLI versions released before they were tagged in GitHub would not appear in \`data/versions.json\` and would need to be hand-maintained — either as additional entries in the data file or as static rows alongside the data-driven body. In practice all current \`versions.md\` entries (down to 0.12.2) are present in \`data/versions.json\`, so this is a hypothetical caveat for the proposal rather than a known migration gap.

## Related

- Today's failed run for v3.242.0: https://github.com/pulumi/docs/actions/runs/26097909723
- Path fix that revealed the brittleness: #19181 (commit 155be91f16)
- Duplicate-row evidence: \`content/docs/install/versions.md:47-51\`

jkodroff:
## Summary

- The CLI versions table at `content/docs/install/versions.md` is no longer hand-maintained by an in-workflow `sed` append; it renders at build time from `data/versions.json` via a new `{{< changelog-table-body >}}` shortcode (3.x releases only, pre-releases filtered out).
- `scripts/get-cli-versions.js` (renamed from `get-versions.js`) now works locally — switched off `@octokit/auth-action` (GHA-only) to a plain `GITHUB_TOKEN` env var and dropped `@octokit/graphql` in favor of native `fetch`. Added `make update-versions` for ergonomics. This means a missed release (e.g. today's failed `v3.242.0` run) can be backfilled from a laptop instead of waiting for the next workflow run; this PR's `data/versions.json` change is exactly that backfill.
- Removed the now-redundant "Update version lists" step from `.github/workflows/pulumi-cli-docs.yml` and the unused `changelog-table-row` shortcode.

Fixes #19183.

### Notable diffs

- `content/docs/install/versions.md` shrinks from 519 lines to ~40 (471 hand-maintained `<tr>` shortcode calls replaced by one).
- Rendered table now shows 329 rows (3.242.0 → 3.0.0) with no duplicates and no pre-release tags. Previously had 471 rows with two duplicate pairs (3.241.0, 3.239.0) and a gap at 3.242.0.
- Dates now come from GitHub's `committedDate`/`tagger.date` (i.e. the actual release date) instead of `$(date +%Y-%m-%d)` from the workflow runner. A handful of rows shift by a day across timezones — accepted one-time diff.

## Test plan

- [x] `make build` succeeds locally
- [x] `make lint` passes
- [x] Rendered table inspected: newest is 3.242.0, oldest is 3.0.0, arm64 date-gating still correct (3.0.0 has no `windows-arm64`; 3.241.0+ has all)
- [x] `make update-versions` runs locally with `GITHUB_TOKEN=$(gh auth token)` and produces same output as the workflow would
- [ ] CI green


CamSoper (approved):
Let the auto-review finish running to make sure there's nothing non-obvious, but other than that, LGTM!

PR #19352, opened 2026-05-22
docs: clarify Input<T> unwrapping and output declaration for dynamic providers

### LINKED ISSUE #12151: Suggestions for additional docs for dynamic resources
Some feedback from an otherwise happy user:

> Dynamic resources are great but it would be great to have a fully featured example closer to the top of the documentation which clearly shows how to access T for Input<T> types (there’s an example at the bottom with LabelInputs and LabelResourceInputs which is helpful but I missed this on first read as it’s so far down and not covered in the section on Inputs).
>
>The outputs on dynamic resources are not type safe. I need to manually declare the public read only properties which will exist at runtime but this is not the best DX. Ideally, there would be a “myCustomResource.outputs” property which is strongly typed using a generic on ResourceProvider.

jkodroff:
Addresses feedback in #12151 by improving the dynamic providers documentation:

- Expands "Dynamic Resource Inputs" to show both Input<T>-wrapped and plain-unwrapped interfaces, with an explanation of the resolution behavior
- Updates "Dynamic Resource Outputs" to explicitly note each output must be declared individually and initialized to undefined

Closes #12151

### DIFF
--- a/content/docs/iac/concepts/providers/dynamic-providers.md
+++ b/content/docs/iac/concepts/providers/dynamic-providers.md
@@ -371,6 +371,8 @@ The method is passed the `id` of the resource, as tracked in the cloud provider,
 
 The inputs to your `pulumi.dynamic.ResourceProvider`'s functions come from subclasses of `pulumi.dynamic.Resource`. These inputs include any values in the input arguments passed to the `pulumi.dynamic.Resource` constructor. This is just a map of key/value pairs however, in statically typed languages, you can declare types for these input shapes.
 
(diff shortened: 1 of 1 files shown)


pierskarsenbarg (approved):

jkodroff:
Closing as superseded by #19361, which split the dynamic providers page into a concepts page and an authoring guide. The new guide at `content/docs/iac/guides/building-extending/providers/dynamic-providers.md` already incorporates the improvements this PR proposed:

- The two-interface pattern (`MyResourceInputs` with `Input<T>` wrappers for the constructor; a separate un-wrapped interface for provider methods), with the explanation that Pulumi resolves all `Input<T>` values before invoking provider functions.
- The `declare readonly` outputs pattern, with the warning against `public readonly`/`!`, and seeding each output to `undefined` in the `super()` call.

Re-merging master would result in a no-op diff, so closing rather than merging.

PR #19402, opened 2026-05-27
content(iac): correct resource options classification for components

### LINKED ISSUE #19397: Resource options: verify and correct applies_to classifications for component resources
## Summary

The classification metadata in [`data/resource_options.yaml`](https://github.com/pulumi/docs/blob/master/data/resource_options.yaml) — which drives the per-option `applies_to` callouts on the [resource options docs pages](https://www.pulumi.com/docs/iac/concepts/resources/options/) — is incorrect for several entries with respect to component resources. The errors surfaced in a Slack thread when a user asked whether the `hooks` option works on component resources; the docs implied it didn't, but it does.

Additional suspect entries were flagged in the thread. The full set needs to be re-verified by actually running programs (not just reading p/p code).

## Known incorrect entries

- **`hooks`** — Docs currently say it does not apply to component resources. Engineering confirmed it does apply to components.
- **`deletedWith`** — Currently `applies_to: both`. The flag can be set on a component but the engine ignores it because Delete is never called on components — effectively a no-op for components.

## Entries to re-verify

These may now affect components and need explicit verification:

- `hideDiffs`
- `ignoreChanges`
- `replaceOnChanges`

A broader open question was also raised: **what do children of a component pick up automatically from the component's options?** That inheritance behavior should be reflected in the docs / metadata where relevant.

## Suggestions for a fix

1. For each option in `data/resource_options.yaml`, write a small Pulumi program that:
   - Sets the option on a component resource directly, and
   - Observes whether the engine actually honors it (vs. silently ignoring).
1. Update `applies_to` and `description` fields to match observed behavior. For options that are settable on components but ignored by the engine (e.g., `deletedWith`), make that explicit in the description rather than just claiming "applies."
1. Document which options a component's children inherit automatically.
1. Update the resource options landing page callout/note that prompted the original question so it matches the per-option metadata.

## Source

Internal Slack thread: https://pulumi.slack.com/archives/C9SGX9QA1/p1779800084665679

jkodroff:
Verifies and corrects the per-option `applies_to` classifications in `data/resource_options.yaml` against observed engine behavior, and documents which options propagate from a component to its children at resource registration time.

## Reclassifications
- `hooks`: `custom` → `both` (hooks fire on the component itself)
- `deletedWith`: `both` → `custom` (no-op on a component but inherited by children)
- `replacementTrigger`: `both` → `custom` (components are never replaced)
- `provider`: `custom` → `both` (propagates to children of the same package)

## What else changed
- New `inherited_by_children` field on every YAML entry, with a new `resource-options-inheritance-table` shortcode rendering the inheritance table on the options landing page from the YAML.
- Drops the unused `components_note` field (no template consumed it).
- Tweaks the scope-body partial to say "has no direct effect" instead of "silently ignored", so the boilerplate doesn't contradict the new inheritance notes.

Behavior was verified by running small Pulumi programs against the random provider on a local file backend and inspecting `pulumi stack export` for each option set on a custom component.

Closes #19397.

### DIFF
--- a/content/docs/iac/concepts/resources/options/_index.md
+++ b/content/docs/iac/concepts/resources/options/_index.md
@@ -23,8 +23,18 @@ Resource constructors accept the following resource options. Each option's refer
 
 ## Resource options and component resources
 
-Not all resource options apply to [component resources](/docs/iac/concepts/components/). Component resources are logical groupings that don't have a backing cloud provider, so options that affect provider behavior have no effect on the component itself (though they may affect child resources). The **Applies to** column in the table above shows which options apply to component resources; each option's reference page explains the behavior in detail.
+Not all resource options apply to [component resources](/docs/iac/concepts/components/). Component resources are logical groupings that don't have a backing cloud provider, so options that affect provider behavior have no effect on the component itself (though, as described below, several of them are inherited by the component's children). The **Applies to** column in the table above shows which options have a direct effect on the component itself; each option's reference page explains the behavior in detail.
 
 {{% notes type="info" %}}
-When you apply a resource option to a component that doesn't support it, the option is silently ignored. To apply options like `protect` or `ignoreChanges` to the child resources within a component, use the `transforms` option to modify child resources as they're created.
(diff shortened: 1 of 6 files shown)


jkodroff:
@claude #update-review

jeffmerrick (approved):

PR #19412, opened 2026-05-27
add a callout to use an explicit url in oidc

lichtie:
### Proposed changes

Add a callout to the OIDC login instructions that provide the solution to an error folks could run into.


cnunciato (approved):
This looks good! Thanks @lichtie.

PR #19426, opened 2026-05-28
Clarify apply error handling; normalize language tabs on apply page

### LINKED ISSUE #11684: docs that describe what exactly happens when an ApplyT() on an output returns an error
from community slack:
>Tried searching through the go documentation, the getting started, some blog posts, and github issues. I however cannot find anything that describes what exactly happens when an ApplyT() on an output returns an error (say like the example I wanted to return an error if a string does not contain the letter 'a'? Does ApplyT panic if the function passed in returns an error? Does it return the error via the main pulumi.RunErr() call? Is there something I need to do in order to capture and handle the error?

[community slack convo](https://pulumi-community.slack.com/archives/C84L4E3N1/p1650652700838039)

jkodroff:
## Summary

Addresses pulumi/docs#11684, which asked what happens when an `apply` callback returns/raises an error (the reporter guessed Go). I verified the behavior empirically across all five SDK languages by running real one-resource Pulumi programs against the local backend.

### Findings (verified)
- An error in an `apply` callback is reported as a **deployment failure** in every language — it does **not** panic/crash the process uncatchably.
- **Go is the outlier:** returning a non-nil `error` from `ApplyT` only fails the update if the resulting output is **consumed** (exported as a stack output or passed as a resource input). A discarded output's error is silently dropped. (Confirmed in SDK source: the apply goroutine `reject`s the output, and the `WorkGroup` tracks completion only, not errors.)
- TypeScript, Python, C#, and Java fail the update whether or not the output is consumed.

## Changes

- **New "Handling errors in apply" section** documenting the per-language behavior, with a Go-specific warning callout and the "no panic" clarification.
- **Removed JavaScript** from the page's choosers/example-programs. Choosers are synced site-wide by language; a stray JavaScript tab made JS the active language and blanked every section without a JS block. TypeScript represents the nodejs SDK in docs.
- **Added Java variants** for the three JSON example programs (`aws-s3-bucketpolicy-jsonstringify`, `-interpolate`, `aws-iampolicy-jsonparse`) so the Java tab is no longer blank, and added `java` to those `example-program` language lists.
- **Removed YAML** from the choosers (Pulumi YAML has no `apply` equivalent) and added a note near the top explaining why, pointing users to interpolation syntax.

## Testing

- `aws-iampolicy-jsonparse-java`: full `pulumi preview` **passes** locally (no cloud creds needed).
- The two S3 Java programs compile and run; locally they stop only at the AWS credentials check (no creds in my env) — identical to the existing JS/TS variants. The program-test CI job has AWS credentials and will run their previews.
- `make lint` passes; chooser/choosable shortcodes balance; rendered page verified (zero visible JavaScript/YAML tabs, Java tabs populated).

## Follow-up (separate)

While reviewing, I found a pre-existing site-wide CSS bug: multi-paragraph `{{% notes %}}` render at 16px vs 14px for single-paragraph notes. Filed separately (not in this PR).

Fixes #11684


smithrobs (approved):
LGTM; poor YAML, yay more info!

PR #19453, opened 2026-06-01
[docs] Document Terraform modules in the Pulumi Cloud registry

### LINKED ISSUE #20555: Document TF module to Pulumi package conversion in the registry docs and launch post
The docs and launch blog in [pulumi/docs#19453](https://github.com/pulumi/docs/pull/19453) describe hosting Terraform modules in the Pulumi Cloud registry, and the consumption path they teach is the interim one:

```
pulumi package add hcl module tf.pulumi.com/<namespace>/<name>/<system> [<version>]
```

They say nothing about the conversion, which by now is most of the story: publishing a module version automatically produces a Pulumi package, and the normal way to consume it is

```
pulumi package add <source>/<publisher>/<name>-<system> [<version>]
```

Missing from both the concept page and the post:

- **That conversion happens at all**, and that the package appears without the publisher doing anything.
- **The package's name.** `<module-name>-<system>`, under the module's namespace. Nothing outside the codebase states this rule, so a publisher cannot find their package without being told.
- **The states a conversion can be in** (`converting`, `converted`, `failed`, `unsupported`) and what each means for a consumer: what to do while it converts, and that `unsupported` is terminal while `failed` is not.
- **When the `hcl module` path is the right one.** The docs currently present it as *the* Pulumi consumption path. After conversion it is the fallback: correct while converting, and for modules that cannot convert, but not the default.
- **What a publisher should expect to see in the console**, once the package page and module page tell a coherent story.

The blog post has a framing question too: "your Terraform modules are hosted here" is a smaller claim than "your Terraform modules become Pulumi packages". Worth deciding which one the launch leads with before the post is finalised.

Depends on the surfaces settling first: #47000 (how a publisher discovers the package), #46996 and #47002 (what the package page shows for versions that did not convert). Writing the docs before those land risks documenting behaviour that changes.

fnune:
Docs for hosting Terraform modules in the Pulumi Cloud registry, and for the Pulumi package every published version converts into.

- **Concept page** (`docs/idp/concepts/terraform-modules.md`, new): authenticating with a Pulumi access token, publishing via go-tfe or the `hashicorp/tfe` provider, the module layout read at publish, migrating from HCP Terraform, what publishing produces, and consuming from a Pulumi program or from OpenTofu / Terraform.
- **Terraform module guide**: its "Using a module from Pulumi Cloud" section installs the converted package by name.

`pulumi package add <name>-<system>` is the documented path for a Pulumi program on both pages. `pulumi package add hcl module <address>` is the fallback for a version that has no package. Usage tracking follows the package, so a consumer reaching the module over the Terraform protocol reports no dependency; both pages say so.

The launch blog post was split out into #20648, which depends on this PR for its links and `canonical_url`.

Notes:

- The registry host is `tf.pulumi.com` (per pulumi/pulumi-service#44889), not `app.pulumi.com`, which is the console only.
- Module names reject underscores, unlike HCP Terraform. Covered under migration.
- Installing a converted package needs Pulumi CLI 3.248.0 or newer, verified by bisect: 3.247.0 fails the plugin handshake, 3.248.0 works.
- The UI these pages describe is the package detail page, not the Terraform module detail page being removed from the console.

Design doc: https://app.notion.com/p/Terraform-module-hosting-in-the-Pulumi-Cloud-registry-372fdbdf1cce801ea7a3f4946c1e4154

Fixes https://github.com/pulumi/pulumi-service/issues/44080
Fixes https://github.com/pulumi/docs/issues/20555

## Test plan

1. Automated checks
   - `./scripts/format.sh`, `node scripts/lint/lint-markdown.js`, and `vale --config=.vale.ini`: 0 errors
2. Manual checks
   - `make serve`, confirm the "Terraform Modules" entry renders under IDP Concepts
   - Click through the internal links on both pages


cnunciato:
@fnune Can you break out the blog into a separate PR so we can merge the docs independently? Then we can publish the post on or about launch day.

content/docs/idp/concepts/terraform-modules.md
```diff
@@ -0,0 +1,141 @@
...
+```go
+client, _ := tfe.NewClient(&tfe.Config{
+    Address: "https://tf.pulumi.com",
+    Token:   os.Getenv("PULUMI_ACCESS_TOKEN"),
+})
+```
+
+The `RegistryModules` surface (`client.RegistryModules.Create`, `CreateVersion`, `UploadTarGzip`, etc.) accepts the same payloads it accepts against `app.terraform.io`.
+
+### `hashicorp/tfe` Terraform provider
```
cnunciato, line 55: I'm guessing this is meant to show TF/Tofu users how to publish a module to P Cloud with HCL. Is that right? 

Both of these headings (go-tfe and this one) could use a brief one-line explanation to set context, IMO.

content/blog/terraform-modules-pulumi-cloud-registry/index.md
```diff
@@ -0,0 +1,97 @@
...
+At publish time the registry reads the [standard Terraform module layout](https://developer.hashicorp.com/terraform/language/modules/develop/structure). Root `.tf` files describe the module's inputs and outputs, `modules/<name>/` subdirectories become submodules, and `examples/<name>/` subdirectories are captured as examples. Submodules and examples are discovered automatically, so there is nothing extra to declare.
+
+## Every version becomes a Pulumi package
+
+Publishing does two things. The version becomes available over the Terraform module protocol, as it would on HCP Terraform. The version is also converted into a Pulumi package, automatically, with nothing extra to configure.
+
+The package takes its name from the module: `<name>-<system>`, in the same namespace. A module published as `acme-corp/vpc/aws` becomes a package called `vpc-aws`, and a Pulumi program installs it by that name:
+
+```bash
+pulumi package add vpc-aws 1.2.3
```
cnunciato: Needs updating. Should look more like this:

```
pulumi package add hcl module tf.pulumi.com/<org>/<module-name>/<system> <version>
```

... where "system" needs a brief explanation as well, because it isn't a term Pulumi otherwise uses.

content/blog/terraform-modules-pulumi-cloud-registry/index.md
cnunciato, line 1: Let's break the blog out into a separate PR so we can publish it separately.

content/docs/iac/guides/building-extending/using-existing-tools/use-terraform-module.md
```diff
@@ -84,6 +84,26 @@ pulumi package add terraform-module ./path/to/module localmod
 
 Any directory containing `.tf` files and optionally `variables.tf` and `outputs.tf` is considered a valid module.
 
+### Using a module from Pulumi Cloud
+
+If your organization publishes Terraform modules to the [Pulumi Cloud registry](/docs/idp/concepts/terraform-modules/), every published version is converted into a Pulumi package for you. Install it by package name, which is the module's name and system joined with a hyphen:
+
+```bash
+pulumi package add <name>-<system> [<version>]
```
cnunciato, line 92: Needs updating, should look more like this (pulled from the console, and also worked for me):

```
pulumi package add hcl module tf.pulumi.com/<org>/<module-name>/<system> <version>
```

... where "system" needs a brief explanation as well, because it isn't a term Pulumi otherwise uses.

content/docs/iac/guides/building-extending/using-existing-tools/use-terraform-module.md
```diff
@@ -84,6 +84,26 @@ pulumi package add terraform-module ./path/to/module localmod
 
 Any directory containing `.tf` files and optionally `variables.tf` and `outputs.tf` is considered a valid module.
 
+### Using a module from Pulumi Cloud
+
+If your organization publishes Terraform modules to the [Pulumi Cloud registry](/docs/idp/concepts/terraform-modules/), every published version is converted into a Pulumi package for you. Install it by package name, which is the module's name and system joined with a hyphen:
```
cnunciato: > converted into a Pulumi package for you

What is actually converted, and how? When I run `pulumi package add`, for example, it looks like the conversion is running (and the artifacts being saved and committed) locally. What actually happens remotely?

content/docs/iac/guides/building-extending/using-existing-tools/use-terraform-module.md
```diff
@@ -84,6 +84,26 @@ pulumi package add terraform-module ./path/to/module localmod
...
+
+```bash
+pulumi package add <name>-<system> [<version>]
+```
+
+A module published as `acme-corp/vpc/aws` installs as `vpc-aws`. This is the same as any other Pulumi package: you get a generated SDK in your language, an [API reference](/docs/idp/concepts/private-registry/#api-documentation) on the package's page, and [usage tracking](/docs/idp/concepts/private-registry/#usage-tracking) showing which of your stacks depend on it and which are behind the latest version. Installing a converted package requires Pulumi CLI 3.248.0 or newer; see [Download & Install Pulumi](/docs/install/) to upgrade.
```
cnunciato, line 95: > an API reference on the package's page

This doesn't line up with what I see in the UI. It reads like I should get language-specific API docs, but what I currently see is a list of the inputs and outputs of the module. That's good, but if we aren't actually generating SDKs and API docs, we should probably remove or adjust this.

content/docs/iac/guides/building-extending/using-existing-tools/use-terraform-module.md
```diff
@@ -84,6 +84,26 @@ pulumi package add terraform-module ./path/to/module localmod
...
+
+The package's page in Pulumi Cloud shows whether a given version has converted. The same conversion can also be run locally, against the module address rather than the package name:
+
+```bash
+pulumi package add hcl module tf.pulumi.com/<namespace>/<name>/<system> [<version>]
+```
+
+This runs the conversion at the moment you run it, using whatever version of the `hcl` provider you have, rather than using the package the registry produced. Use it while a version is still converting. A module the registry could not convert fails here for the same reason. The version is optional; omit it to resolve the latest published version. Self-hosted Pulumi Cloud installations use their own host (`<your-pulumi-host>/<namespace>/<name>/<system>`). Either way the CLI passes your Pulumi access token through to the provider, so you do not need to set a registry token by hand.
```
cnunciato: >This runs the conversion at the moment you run it, using whatever version of the `hcl` provider you have, rather than using the package the registry produced.

* How does the user know which version of the `hcl` provider they have? They don't (I don't think?) knowingly install this.
* I'm still unclear as to what the registry is actually producing and where to find/see it.

> Use it while a version is still converting.

What does this mean?

> A module the registry could not convert fails here for the same reason.

I'm not following what this refers to. 

>Either way the CLI passes your Pulumi access token through to the provider, so you do not need to set a registry token by hand.

Do we need to call this out, since it's just "how Pulumi works" (i.e., it reads and uses your locally-set credentials)? (If not, suggest we delete.)

content/docs/idp/concepts/terraform-modules.md
```diff
@@ -0,0 +1,141 @@
...
+  idp:
+    parent: idp-concepts
+    identifier: idp-concepts-terraform-modules
+    weight: 15
+aliases:
+  - /docs/terraform-modules/
+  - /docs/iac/using-pulumi/pulumi-cloud/registry/terraform-modules/
+---
+
+Pulumi Cloud hosts Terraform modules as a first-class registry resource alongside [packages](/docs/iac/concepts/packages/) and [templates](/docs/idp/concepts/organization-templates/). Teams migrating from HCP Terraform can publish their existing modules to Pulumi Cloud using the same tooling they already use (the [go-tfe](https://github.com/hashicorp/go-tfe) library or the [hashicorp/tfe Terraform provider](https://registry.terraform.io/providers/hashicorp/tfe/latest/docs)) by pointing those tools at `tf.pulumi.com` instead of `app.terraform.io`. Every module version you publish is also converted into a Pulumi package. The module's variables become typed inputs and its outputs become typed outputs, with a generated SDK in TypeScript, Python, Go, C#, Java, or YAML, an API reference on the package's page, and a record of which stacks depend on it. Existing `.tf` consumers are unaffected and keep resolving the module over the Terraform protocol.
```
cnunciato: > Every module version you publish is also converted into a Pulumi package. The module's variables become typed inputs and its outputs become typed outputs, with a generated SDK in TypeScript, Python, Go, C#, Java, or YAML, an API reference on the package's page, and a record of which stacks depend on it.

Need to confirm this. (I don't see any indication of it in the UI.)

> Existing `.tf` consumers are unaffected

Unaffected by what? If this is saying that uploaded Terraform modules that get converted into language-specific SDKs can still be consumed as HCL, I think we can just delete it. ("Every module version you publish is also converted into a Pulumi package" makes it pretty clear that's additive and not destructive/exclusive of HCL.)

content/docs/idp/concepts/terraform-modules.md
```diff
@@ -0,0 +1,141 @@
...
+1. You need a [Pulumi Cloud](https://app.pulumi.com) account on the Enterprise or Business Critical plan. Publishing is gated to those tiers; reading and listing modules is available on any plan, so you always keep access to modules you have already published.
+1. You need the [Pulumi CLI](/docs/install/) installed if you plan to consume modules from a Pulumi program.
+1. You need OpenTofu or Terraform installed if you plan to consume modules from a `.tf` file with `tofu init` / `terraform init`.
+
+## Authenticate
+
+Every surface authenticates with a [Pulumi access token](/docs/administration/access-identity/access-tokens/). It is the bearer token for everything Pulumi Cloud exposes over the HashiCorp protocol: the publish API, the state backend, and the module registry.
+
+- Publishing: the go-tfe client and the tfe provider take your Pulumi access token wherever they expect a TFE token today. See [Publish a module](#publish-a-module).
+- Consuming from a Pulumi program: run `pulumi login`. `pulumi package add terraform-module` passes the token through to the provider, so there is no separate registry login.
```
cnunciato:
```suggestion
- Consuming from a Pulumi program: run `pulumi login`. `pulumi package add hcl module` passes the token through to the provider, so there is no separate registry login.
```

content/docs/idp/concepts/terraform-modules.md
```diff
@@ -0,0 +1,141 @@
...
+- `modules/<name>/` subdirectories are parsed the same way as the root and can be consumed as submodules.
+- `examples/<name>/` subdirectories and the `README.md` are captured at publish.
+
+## Migrating from HCP Terraform
+
+If you publish from CI today, the move is a host change. Point your existing pipelines at `tf.pulumi.com`, supply a Pulumi access token, and your publish steps run unchanged. Reading and listing modules work on any plan, so you keep access to modules you have already published regardless of your plan.
+
+### Module names
+
+Pulumi Cloud uses the same `<namespace>/<name>/<system>` address form as HCP Terraform, where the namespace is your Pulumi organization. One rule is stricter: the module name must match `[a-z0-9][a-z0-9-]*`, so it starts with a letter or digit and underscores are rejected at publish. A module that HCP hosts under a name like `control_tower_account_factory` has to be renamed to `control-tower-account-factory` before you publish it. Uppercase in the name is lowercased automatically.
```
cnunciato: > One rule is stricter: the module name must 

Oh — this seems like a pretty big deal (in terms of migration), given module names are supplied by consumers. 😬 Is there a plan to support underscores at some point? (cc @djperlovsky)

content/docs/idp/concepts/terraform-modules.md
```diff
@@ -0,0 +1,141 @@
...
+
+## What happens when you publish
+
+Publishing a module version also converts it into a Pulumi package, with no extra step on your part. The package is named after the module: `<name>-<system>`, published under the same namespace and registry source, so a module published as `acme-corp/vpc/aws` produces a package called `vpc-aws`.
```
cnunciato, line 89: Just calling out that we may need to revise this depending on actual behavior.

content/docs/idp/concepts/terraform-modules.md
```diff
@@ -0,0 +1,141 @@
...
+
+Once a version has converted, install it by package name:
+
+```bash
+pulumi package add <name>-<system> [<version>]
+```
+
+The module is a [multi-language component](https://github.com/pulumi/pulumi-hcl/blob/master/docs/mlc.md): its `variable` blocks become typed inputs, its `output` blocks become typed outputs, and Pulumi generates an SDK in the language your project uses. The version you pass is persisted in `Pulumi.yaml`, so `pulumi install` regenerates the same pinned version.
+
+The resources the module creates appear individually in previews and in the resource graph rather than as one opaque unit. In Pulumi Cloud the package gets the same treatment as any other: an [API reference](/docs/idp/concepts/private-registry/#api-documentation) generated from the module's variables and outputs, and [usage tracking](/docs/idp/concepts/private-registry/#usage-tracking) recording which stacks depend on it and which of those are behind the latest version.
```
cnunciato, line 103: > API reference generated from the module's variables and outputs

Again just calling out that I don't this in the UI. (The link to private-reg documentation suggests we should see language-specific docs.)

content/docs/idp/concepts/terraform-modules.md
```diff
@@ -0,0 +1,141 @@
...
+
+Usage tracking only counts consumption through the converted package. A stack or workspace that consumes the module over the Terraform protocol does not report a dependency, so it does not appear in the usage columns or on the package's "Used by" tab.
+
+{{% notes type="info" %}}
+Installing a converted package requires Pulumi CLI 3.248.0 or newer. Older versions fail with a plugin handshake error. See [Download & Install Pulumi](/docs/install/) to install or upgrade.
```
cnunciato: Detail's unnecessary, can omit.

```suggestion
Installing a converted package requires Pulumi CLI 3.248.0 or newer. See [Download & Install Pulumi](/docs/install/) to install or upgrade.
```

content/docs/idp/concepts/terraform-modules.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Converting a module locally
+
+The same conversion can be run locally, against the module address rather than the package name:
+
+```bash
+pulumi package add hcl module tf.pulumi.com/<namespace>/<name>/<system> [<version>]
+```
+
+`hcl` is a parameterized provider. The `module` keyword selects module mode, followed by the module address and an optional version. Omit the version to resolve the latest published version; pass one to pin it.
```
cnunciato, line 119: > `hcl` is a parameterized provider. The `module` keyword selects module mode, 

This seems like an implementation detail, so irrelvant to the user. Suggest deleting unless they need to know this for some reason.

content/docs/idp/concepts/terraform-modules.md
```diff
@@ -0,0 +1,141 @@
...
+Reference the module in a `.tf` file:
+
+```hcl
+module "vpc" {
+  source  = "tf.pulumi.com/<namespace>/<name>/<system>"
+  version = "1.2.3"
+}
+```
+
+`tofu init` discovers the `modules.v1` endpoint on Pulumi Cloud's `.well-known/terraform.json`, lists available versions, and downloads the tarball using the token from `TF_TOKEN_tf_pulumi_com`.
```
cnunciato: > discovers the `modules.v1` endpoint on Pulumi Cloud's `.well-known/terraform.json`, lists available versions, and downloads the tarball using the token from `TF_TOKEN_tf_pulumi_com`.

Also seems like inside baseball (so suggest removing), unless we want to call out that this can be done with the REST API also.

cnunciato (commented):
Nice! Bunch of things for you to consider. Main thing is that I'd like to have us publish the blog separately, but I also a bunch of questions and suggestions for you to consider. Ping me when you're ready for a second look!

content/blog/terraform-modules-pulumi-cloud-registry/index.md
fnune: I believe the page shouldn't appear until the publishing date arrives, which would be tomorrow. But I'll split it.

content/docs/iac/guides/building-extending/using-existing-tools/use-terraform-module.md
fnune: `pulumi package add` is correct.

content/blog/terraform-modules-pulumi-cloud-registry/index.md
fnune: https://github.com/pulumi/docs/pull/19453#discussion_r3702298232

content/docs/iac/guides/building-extending/using-existing-tools/use-terraform-module.md
fnune: The job runs `https://github.com/pulumi-labs/pulumi-hcl` on the TF module to produce a Pulumi package with a correctly-typed schema.

content/docs/idp/concepts/terraform-modules.md
fnune: `pulumi package add <...args>` is public Pulumi CLI syntax, and it may be new to the user, because it's a parameterized package. It's public API, not an implementation detail IMO.

Thoughts @iwahbe?

content/docs/idp/concepts/terraform-modules.md
fnune: The UI this document talks about is the package detail page (under Registry or Private components). You may have been looking at the outdated TF module detail page we had for a while.

content/docs/idp/concepts/terraform-modules.md
fnune: https://github.com/pulumi/docs/pull/19453#discussion_r3702371525

content/docs/idp/concepts/terraform-modules.md
fnune: I believe this bullet point emphasizes `pulumi package add hcl <...args>` which is not what we want users doing. The happy path is `pulumi package add <your-converted-package>`.

content/docs/idp/concepts/terraform-modules.md
fnune: > Need to confirm this. (I don't see any indication of it in the UI.)

https://github.com/pulumi/docs/pull/19453#discussion_r3702371525

> Unaffected by what?

Yeah, it means unaffected in that you may continue to consume the TF module the same way after you've switched the host to `tf.pulumi.com`. I'll clarify/simplify.

content/docs/iac/guides/building-extending/using-existing-tools/use-terraform-module.md
fnune: > Do we need to call this out, since it's just "how Pulumi works" (i.e., it reads and uses your locally-set credentials)? (If not, suggest we delete.)

Hmm I agree I think.

fnune: > I'm still unclear as to what the registry is actually producing and where to find/see it.
> [...]
> I'm not following what this refers to.

I believe this is also because: https://github.com/pulumi/docs/pull/19453#discussion_r3702371525

content/docs/iac/guides/building-extending/using-existing-tools/use-terraform-module.md
fnune: https://github.com/pulumi/docs/pull/19453#discussion_r3702371525

content/docs/idp/concepts/terraform-modules.md
fnune: `_` would be unexpected in a Pulumi package name, I believe. Discussing here https://pulumi.slack.com/archives/C0B7AUEUWF7/p1785745290120049

content/docs/iac/guides/building-extending/using-existing-tools/use-terraform-module.md
fnune: See https://github.com/pulumi/docs/pull/19453#discussion_r3702371525

content/blog/terraform-modules-pulumi-cloud-registry/index.md
fnune: Split into #20648.

content/docs/idp/concepts/terraform-modules.md
cnunciato: Ah, ok thanks. I had no idea there was another place to see this!

That page (Terraform Modules) is definitely still there in prod though. Are we opening to remove it?

fnune: It's only enabled for your organization. I'll disable it for everyone now.

content/docs/idp/concepts/terraform-modules.md
cnunciato: Yep, and in a TF module also; looks like `-` is the convention there as well. So we should be okay.

cnunciato (approved):

PR #19509, opened 2026-06-03
fix: derive page "Updated" dates from git history, not build time

alexleventer:
## Problem

Every page's "Updated" date — and every `lastmod` in the sitemap — currently shows the **build timestamp** instead of the real last-modified date. On the live site, [what-is-pulumi](https://www.pulumi.com/what-is/what-is-pulumi/) reads "Updated June 3, 2026" (today), its JSON-LD `dateModified`/`datePublished` are pinned to the build second, and **every** URL in the production `sitemap.xml` carries the identical build timestamp.

## Root cause

`build-and-deploy.yml` checks out the repo with `actions/checkout@v6` and no `fetch-depth`, which defaults to a shallow clone (latest commit only). With no git history present, Hugo's `enableGitInfo` has nothing to read, so `.Lastmod` falls back to build time across the whole site.

## Changes

- **`.github/workflows/build-and-deploy.yml`** — set `fetch-depth: 0` so the production build has full history for `enableGitInfo`. Root-cause fix; also corrects the sitemap.
- **`layouts/partials/schema/collectors/article-entity.html`** — drop the `now` fallback that actively fabricated today's date into `datePublished`/`dateModified`. Dates now come only from front matter or git; when neither exists the fields are omitted rather than reporting a false "modified today" to search engines.

`layouts/what-is/single.html` is unchanged — its guard was already correct; it only rendered "today" because `.Lastmod` was wrong upstream.

## Verification

Local production build with full git history present:

| Page | Last commit | Rendered "Updated" | JSON-LD `dateModified` |
|---|---|---|---|
| what-is-pulumi | 2026-05-18 | May 18, 2026 ✓ | 2026-05-18T15:31:59 ✓ |
| what-is-devops | 2026-05-19 | May 19, 2026 ✓ | 2026-05-19T14:22:25 ✓ |

Sitemap now carries **378 distinct `lastmod` values** instead of a single build timestamp.

> Note: `fetch-depth: 0` adds some clone time to the deploy — the standard tradeoff for `enableGitInfo`. Takes effect on the next deploy after merge.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -33,6 +33,12 @@ jobs:
         uses: pulumi/esc-action@v1
 
       - uses: actions/checkout@v6
+        with:
+          # Full history (not the default shallow fetch-depth: 1) so Hugo's
+          # enableGitInfo can read each file's last commit date. Without this,
+          # .Lastmod falls back to build time and every page (and sitemap entry)
+          # reports "Updated <today>".
+          fetch-depth: 0
 
       - uses: actions/setup-node@v6
         with:
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -39,6 +39,10 @@ jobs:
           # .Lastmod falls back to build time and every page (and sitemap entry)
           # reports "Updated <today>".
           fetch-depth: 0
+          # Blobless clone: fetch the full commit graph and trees enableGitInfo
+          # needs, but skip downloading historical file blobs (the asset-heavy
+          # history of a docs repo). Current-tree blobs are fetched on demand.
+          filter: blob:none
 
       - uses: actions/setup-node@v6
         with:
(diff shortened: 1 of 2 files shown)


alexleventer:
@claude - take a look pls

alexleventer:
@claude - should i just proactively fix this: 
> The CI performance impact is real: a full clone of a mature docs repo on every deploy and every scheduled rebuild (3× daily) adds non-trivial clone time. Worth monitoring on the first post-merge run. If it becomes a bottleneck, --filter=blob:none (blobless clone) retains commit history while skipping large blobs and is [supported by GitHub Actions checkout](https://github.com/actions/checkout).

alexleventer:
@claude can you make this change pls: https://github.com/pulumi/docs/pull/19509#issuecomment-4617247846

CamSoper (approved):
LGTM. I'll keep an eye on the build and make sure nothing breaks.

PR #19522, opened 2026-06-04
docs(esc): note app-only self-rotation needs Application.ReadWrite.OwnedBy

pose:
## Summary

The `azure-app-secret` rotator's Permissions section presents being an **Owner** of the target app registration as a sufficient alternative to `Application.ReadWrite.All`. That is true for delegated logins, or identities that already carry directory read, but it does not hold for an **application-only** token, which gets no owner-based directory access on its own.

The most common rotation setup is an app rotating its own secret via Azure OIDC login, where the `login` identity and the rotated `clientId` are the same app. That path uses an application-only token, so the app must also be granted the `Application.ReadWrite.OwnedBy` Microsoft Graph permission with admin consent, in addition to self-ownership. `Application.ReadWrite.OwnedBy` is the least-privilege option because it scopes the rights to apps the caller owns.

This PR adds an info callout documenting the requirement, with links to Microsoft's app-only access overview and the guidance on adding an owner to an application.

## Verification

The note reuses the existing `{{% notes type="info" %}}` shortcode, matching sibling ESC docs. No em/en dashes, sentence-case heading unchanged, file ends with a newline, no trailing whitespace. Build and prose lint run in CI.


content/docs/esc/providers/rotators/azure-app-secret.md
```diff
@@ -101,6 +101,10 @@ The Azure identity used for rotation must have the following Microsoft Graph API
 
 Alternatively, the identity can be added as an **Owner** of the specific app registration whose secrets will be rotated.
 
+{{% notes type="info" %}}
+Owner access alone suffices only for **delegated** logins (a user identity) or identities that already have directory read. An **application-only** token must *also* be granted the `Application.ReadWrite.OwnedBy` Microsoft Graph permission with **admin consent**, on top of being an Owner of the target app. This is the common case when an app rotates its *own* secret via [Azure OIDC login](/docs/esc/providers/login/azure-login/), where the `login` identity and the rotated `clientId` are the same app. Scoped to owned apps only, `Application.ReadWrite.OwnedBy` is the least-privilege option for self-rotation. For background, see Microsoft's [app-only access overview](https://learn.microsoft.com/en-us/entra/identity-platform/app-only-access-primer) and [adding an owner to an application](https://learn.microsoft.com/en-us/troubleshoot/entra/entra-id/users-groups-entra-apis/add-owner-for-application-microsoft-graph).
```
seanyeh: maybe we can remove `or identities that already have directory read` as it's misleading (either a delegated login, or the app-only token needs write)

seanyeh (approved):

content/docs/esc/providers/rotators/azure-app-secret.md
pose: Done, removed in 5b1cd13.

Issue #16563, opened 2025-11-13
Make use of saved language from pulumi.com


kramhuber:
On pulumi.com + docs, we save the programming language the user is viewing docs for. We should make use of it to further improve the default user experience.

jkodroff:
@claude Is this still relevant?

jkodroff:
@claude Submit a PR to fix this.

PR #19624, opened 2026-06-12
Track docs choice selections in Segment

jkodroff:
Fire a Segment analytics event in `selectChoice()` whenever the user selects a language, OS, cloud, or k8s-language tab. Mirrors the pattern already used in `neo-mode.ts`.

Event: `docs-choice-selected` with payload `{ kind, choice }`

Collects the data required for #16563

### DIFF
--- a/theme/src/ts/chooser.ts
+++ b/theme/src/ts/chooser.ts
@@ -7,6 +7,11 @@ function getElemClasses(e: Element) {
 function selectChoice(kind, choice, extra?) {
     document.cookie = "pulumi_" + kind + "=" + choice + "; max-age=31536000; path=/";
 
+    const analytics = (window as any).analytics;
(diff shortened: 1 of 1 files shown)


lichtie (approved):

sicarul:
@jkodroff while this is useful i'd not agree this closes https://github.com/pulumi/docs/issues/16563
"We should make use of it to further improve the default user experience." I understand this means to close that issue we should use this to personalize the website. I agree with merging this PR, but i don't agree this closes the issue as fixed (However we could close as wontfix if we are not interested anymore)

CamSoper (approved):
Love it

jkodroff:
So what I'm hearing is that this just does the data collection part, but it's not used anywhere to improve UX. I'll just remove "fixes".

PR #19636, opened 2026-06-12
Remove standalone ESC CLI; standardize on `pulumi env`

### LINKED ISSUE #19666: ESC: Prefer the `pulumi` CLI over the `esc` CLI in all examples
We should be preferring the `pulumi` CLI commands to the `esc` CLI in the ESC docs b/c you're nearly always using the former when you're interacting with Pulumi.

This includes all setup instructions and examples in all pages throughout the ESC docs.

borisschlosser:
## What

The standalone `esc` CLI is being retired in favor of the integrated **`pulumi env`** command (which ships with the Pulumi CLI). This PR removes every reference to the standalone CLI — across docs, tutorials, blog posts, and build infrastructure — and points users at `pulumi env`.

## Changes

**Deleted the standalone ESC CLI docs + infra**
- Removed the entire `content/docs/esc/cli/` section: the generated command reference (~69 pages), the download/install page, shell-completion docs, and the overview (which held the old "ESC CLI vs Pulumi CLI" note).
- Removed the supporting build infra: the `esc-cli.yml` GitHub workflow, `static/esc/latest-version`, the `latest-version-esc` shortcode, the `esc-cli-commands` nav block in `menus.yml`, the menu-expansion case in `menu.html`, the S3-sync line, and the related `README.md` / `BUILD-AND-DEPLOY.md` entries.

**Redirects**
- Regenerated `scripts/redirects/esc-cli-redirects.txt` to map every old `/docs/esc/cli/*` and legacy `/docs/esc-cli/*` path — plus the orphaned `/docs/install/esc/` alias — to the equivalent `pulumi_env` command-reference page.

**Rewrote command usage to `pulumi env`**
- Docs, tutorials, `what-is`, pricing, and reference pages: `esc env <x>` → `pulumi env <x>`, `esc run`/`esc open` → `pulumi env run`/`pulumi env open`, `esc login` → `pulumi login`; install instructions and "ESC CLI" links/wording now point at the Pulumi CLI.
- **24 blog posts** updated the same way (commands, links, and "ESC CLI" prose). Publish dates were left untouched and edits kept surgical.

**GitHub Action**
- Pinned `pulumi/esc-action` to `@v3` (v3 installs the Pulumi CLI and uses `pulumi env`), and reworded the action's "installs the ESC CLI" copy.

**Master merges**
- Merged `origin/master` twice; it restructured the ESC docs (`docs/esc/environments/*` → `docs/esc/concepts/*`, extracted rotation best-practices, merged the IaC-integration steps into get-started). Resolved conflicts in favor of master's new structure and re-applied the `pulumi env` conversion to the new/restructured pages.

## Notes

- Upstream-generated `pulumi_env_*` reference pages still mention `esc env` in their cobra help text — that fix belongs in `pulumi/pulumi`, not here.
- One intentional leftover: an incoming-redirect **alias** `/docs/esc-cli/commands/` in `concepts/webhooks.md` frontmatter (keeps old URLs working; not user-facing content).

## Verification

Repo-wide scans confirm: zero standalone `esc` commands (including backtick-wrapped), zero links to `/docs/esc/cli` · `/docs/esc-cli` · `/docs/install/esc` (excluding frontmatter aliases), zero "ESC CLI" / "esc CLI" text, zero `pulumi esc` / `pulumi env env` / `pulumi pulumi` artifacts, and all `pulumi/esc-action` refs at `@v3`. Hugo renders all content cleanly (the only local build failures are a pre-existing, unrelated SCSS/webpack asset issue).

Fixes #19666


content/blog/pulumi-vscode-extension/index.md
```diff
@@ -68,7 +68,7 @@ You can run without debugging by selecting "Run Without Debugging" from the Run
 
 ## Using Pulumi ESC
 
-Pulumi ESC manages secrets and configuration for your applications and infrastructure within a centralized and composable secrets store. The environments are stored securely within Pulumi Cloud, and can be edited and opened from the Pulumi Cloud console or from the `esc` CLI or [REST API](/docs/pulumi-cloud/reference/cloud-rest-api).
+Pulumi ESC manages secrets and configuration for your applications and infrastructure within a centralized and composable secrets store. The environments are stored securely within Pulumi Cloud, and can be edited and opened from the Pulumi Cloud console or from the `pulumi` CLI or [REST API](/docs/pulumi-cloud/reference/cloud-rest-api).
```
fnune, line 72: In locations like this, say `pulumi env` subcommand instead of "the `pulumi` CLI"?

fnune (approved):
Some notes in case they give you ideas:

- The redirects LGTM, but do we need router-level redirects as well? In the infra (don't remember where the rules live at the moment)
- Do we want to edit _all_ the blog posts? Is there any old blog post actually introducing the CLI which we shouldn't change?

jkodroff:
@claude #update-review

content/blog/announcing-pulumi-esc-github-action/index.md
```diff
@@ -32,7 +32,7 @@ By packaging ESC functionality into a GitHub Action, we’re enabling teams to i
 
 At a high level, the Pulumi ESC GitHub Action can do three things:
 
-1. **Download the Pulumi ESC CLI** - By default, the action ensures the ESC CLI is installed in your GitHub Actions environment. If you provide a specific version, that version is installed.
+1. **Install the Pulumi CLI** - By default, the action ensures the Pulumi CLI is installed in your GitHub Actions environment. If you provide a specific version, that version is installed.
```
jkodroff, line 36: Check the style guide on this. I believe blog posts are supposed to be "moment in time" and the right move is to add a note, but not 100% sure as I don't often touch blog posts.

borisschlosser: Right. The policy is to not touch historic blog posts. But in this case the mentioning of the esc cli in the affected blog posts is unrelated and should be replaced with the new correct approach imo.

borisschlosser:
> * The redirects LGTM, but do we need router-level redirects as well? In the infra (don't remember where the rules live at the moment)

@fnune can you clarify what you mean?

> * Do we want to edit _all_ the blog posts? Is there any old blog post actually introducing the CLI which we shouldn't change?

There is no specific esc cli blog post and the mentioning of the esc cli is unrelated

borisschlosser:
Thanks for the review! Addressed in `64af6c2b73`:

### 🚨 Outstanding
- **[L82] `environments-secrets-configurations-management` — inaccurate `pulumi/esc` repo claim** — ✅ Fixed. Removed the sentence entirely (the Pulumi CLI lives in `pulumi/pulumi`, so rather than reword the repo description, we dropped it). The bullet now ends at "…a self-hosted service in Pulumi Cloud."

### ⚠️ Low-confidence
- **[L46-47] `esc-env-run-aws` — stale `pulumi login` output** — ✅ Fixed. Output now reads "Manage your Pulumi stacks by logging in." / "Run \`pulumi login --help\`". Also fixed the section heading ("Install and login to Pulumi ESC" → "Install and log in to the Pulumi CLI") and the "login to the CLI" grammar.
- **[L33] `esc-version-tagging` — stale first line** — ✅ Fixed → "Manage your Pulumi stacks by logging in."
- **[L114] `product-entities.html` — broadened ESC detection** — ℹ️ Working as intended; no change. This partial resolves `mainProduct` **by URL first**, and the content-keyword check (the line changed from `"pulumi esc"` → `"pulumi env"`) only runs `{{ if and (not $mainProduct) … }}`. The generated `…/docs/iac/cli/commands/pulumi_env*` pages live under `/docs/iac/`, so they're tagged Pulumi CLI by the URL rule and never reach the content check. The content fallback only affects pages with no product URL, where "pulumi env" is a legitimate ESC signal. So the suggested `/docs/iac/cli/commands/` exclusion is already effectively provided by URL precedence.
- **[L35, L63] `announcing-pulumi-esc-github-action` — `esc-action@v3` behavior** — ✅ Verified. The [`pulumi/esc-action` v3.0.0 release notes](https://github.com/pulumi/esc-action/releases/tag/v3.0.0) state: *"This action now installs the **Pulumi CLI**… The full Pulumi CLI (binary + language plugins) is installed."* The blog copy ("ensures the Pulumi CLI is installed") matches.

### 💡 Pre-existing
- **[L49] `esc-open-approvals` — "introdcued" typo** — ✅ Fixed (also prefixed the bare \`env open-request\` → \`pulumi env open-request\` for consistency).

@claude #update-review

static/esc/latest-version
cnunciato, line 1: @borisschlosser Will removing this break existing `esc` CLI users? (Assuming no, just doublechecking.)

seanyeh: I'm pretty sure this will break existing users of the `esc-action`. I don't think we can remove this

content/blog/announcing-pulumi-esc-github-action/index.md
cnunciato: Josh is right about the rule, but in this case it's okay, since `pulumi env` isn't "changing the story" in these posts (none of which are specifically advertising the standalone-ness of the `esc` CLI), and there's value in getting them all updated so the content gets picked up by the bots and LLMs.

static/esc/latest-version
borisschlosser: Good catch! Verified that it's used by esc-action@v1/v2. Restored file and code.

PR #19647, opened 2026-06-12
Scroll docs left nav to the active page and fix sidebar height

### LINKED ISSUE #11547: Left Nav does not scroll to the current page
### What happened?

When opening a page in the docs, the current page in the LH nav is not scrolled into view.

### Example

Open this page directly in the browser bar (or probably by clicking the link) https://www.pulumi.com/docs/pulumi-cloud/esc/

I have to scroll down to see the current page in both Firefox and Safari.

### Output of `pulumi about`

n/a

### Additional context

_No response_

### Contributing

jeffmerrick:
Human Jeff: Fixes the docs sidebar nav not scrolling on load, and also fixes a couple another annoyances with it,: 1. the height overflowing the viewport and 2. the short flash when loading and navigating between pages (mostly?).

---

The docs left nav didn't scroll the current page into view on load, so on short screens the active item was often out of sight. Fixes #11547.

## Changes
- **Scroll to active item** (`menu.html`): a pre-paint inline script centers the active nav item when it's out of view, and preserves the menu's scroll position across navigations (sessionStorage) so moving between sibling pages doesn't shift the nav.
- **No-flash render** (`menu.html`): expanded ancestors of the current page are now rendered server-side instead of being expanded by JS after load.
- **Sidebar height** (`docs-main.ts`, `_docs-main.scss`): the sticky sidebar is sized to the viewport space below the header (100vh − 64px at top, 100vh once the header scrolls away), and the scrollable menu area now extends to the viewport bottom.
- **Instant page transitions** (`head.html`): docs ↔ docs navigations opt into cross-document view transitions (Chrome 126+, Safari 18.2+) with snapshot animations disabled — the old page holds on screen until the new one is ready, then swaps instantly, eliminating the blank flash between page loads. Other browsers ignore it and navigate normally.

### DIFF
--- a/layouts/partials/docs/menu.html
+++ b/layouts/partials/docs/menu.html
@@ -81,7 +81,7 @@
             {{- with index site.Menus $menuName }}
             <li class="expandable expanded{{ if eq $currentSection $menuName }} expanded{{ end }}" data-menu-id="{{ $menuName }}-top">
                 <ul class="list-none">
-                    {{- partial "inline/menu/walk.html" (dict "page" $page "rootMenuID" $menuName "currentSection" $currentSection "menuEntries" . "allMenuEntries" . "forceTopLevelLabel" $label) }}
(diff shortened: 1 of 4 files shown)


CamSoper (approved):
Love it!

PR #19706, opened 2026-06-16
Fix syntax error in transforms.md

UnderKoen:
### Proposed changes

There was a missing closing parententis on an typescript example


CamSoper (approved):
Good catch! Thank you for your contribution!

PR #19721, opened 2026-06-16
Move content-review ledger from in-repo PRs to S3

CamSoper:
## Why

The automated existing-content review (#19642 / #19678 / #19684) maintains a **ledger** — one JSON record per reviewed page tracking when it was last reviewed — that `select-articles.py` reads to enforce review cooldowns and stale-lane ordering. Today that ledger is committed **through PRs**, which produces a daily approval ceremony. Two sources, only one is pure noise:

1. **Fix PRs** — content change + a `ledger/<slug>.json` entry in the same commit. We want to review these, but the ledger file is diff noise riding along.
2. **Ledger-only PRs** — for clean (zero-fix) articles, the day's entries are batched into one `content-review/ledger-<date>` PR. **Pure metadata, but still demands a daily human merge.** This is the ceremony being eliminated.

The ledger is just a key→JSON store; nothing about it requires git. The repo already has the exact precedent: `scripts/social/schedule-posts.py` keeps idempotency state ("which posts have gone out") in a versioned, private S3 bucket, never committed — and this very workflow already assumes the AWS role `…/ContinuousDelivery` and `aws s3 cp`s the traffic snapshot down.

## What

Relocate the ledger to a versioned, private S3 bucket:

- **`infrastructure/index.ts`** — new `content-review-ledger` bucket (versioned, private), mirroring `social-post-state`; export `contentReviewLedgerBucketName`.
- **Dispatcher** (`review-existing-content.yml`) — `aws s3 sync` the ledger into a local `.ledger-cache/` and pass `--ledger-dir`. Degrades gracefully like the traffic fetch.
- **Per-article worker** (`content-review-article.yml`) — add AWS creds + a deterministic ledger-write step that pushes one object per page (key = slug) to S3 from `.content-review-results.json`; `reviewed_at` is stamped by the workflow. Trim the model prompt so it no longer writes ledger files or opens ledger-only PRs.
- **Skill + docs** — record outcomes in the results file's `ledger[]` array instead of writing files; clean articles open **no PR at all**.

**`select-articles.py` is unchanged** — it already exposes `--ledger-dir` and reads whatever directory it's pointed at, so the workflow just points it at the synced cache.

## Net effect

The ledger-only PR disappears entirely, fix PRs become pure content diffs, and the daily merge obligation drops to "only when docs actually changed." S3 versioning provides the audit trail the git history used to.

## Config to set before enabling

A repo variable `CONTENT_REVIEW_LEDGER_S3_URI` (e.g. `s3://<bucket>/ledger/`), matching the existing `CONTENT_REVIEW_TRAFFIC_S3_URI` convention. The bucket comes from a `pulumi up` of `infrastructure/`. Until the var is set, both new steps no-op and selection falls back to tier + age (nothing in cooldown) — same graceful degradation as a missing traffic snapshot.

## Verification

- `python3 scripts/content-review/test_select_articles.py` — **35 passed** (the cooldown test already exercises the `--ledger-dir` read path the cache uses).
- Both workflow YAMLs parse; the ledger-write `jq` pipeline simulated locally produces correct per-slug keys and stamps `reviewed_at` for both fix and clean entries.
- `infrastructure/` is a line-for-line mirror of the `socialStateBucket` block; `tsc` not run here (infra deps not installed in the authoring env) — worth a local `tsc`/`pulumi preview` before merge.

> Note: this is a clean cutover — `scripts/content-review/ledger/` currently holds only its README (no entries yet), so there's nothing to migrate.

https://claude.ai/code/session_01MdpFjbwppAcxsKyMYnNhve

---
_Generated by [Claude Code](https://claude.ai/code/session_01MdpFjbwppAcxsKyMYnNhve)_


CamSoper:
@claude #update-review

Issue #11670, opened 2022-06-29
use of dark mode tools makes our website and docs difficult to use


susanev:
recently in a usability study a participant was using Dark Reader in Firefox which enables a dark mode, but unfortunately, it made our website and docs difficult to use because a number of things were totally invisible

later on a different community member mentioned they were using chrome dark mode and had a similar experience with components being totally invisible.

![dark-mode](https://user-images.githubusercontent.com/5489125/176333055-a3bf0dca-82a7-41a6-aef6-b8b529019e6b.png)

as more and more folks enable these dark mode solutions we should figure out if we can make our website/service work with them or if we should be shipping our dark mode

related: https://github.com/pulumi/registry/issues/244

XenoWad01:
You should really just add darkmode to your docs and this problem will go away.

ryanjjung:
The availability of tools like Firefox Reader View and dark mode browser extensions is not a good excuse to not provide a custom dark mode. Those tools are last-ditch attempts to fix accessibility problems with websites. The lack of a dark mode is an accessibility issue. I spend large portions of my day swapping between a terminal, a code editor, the AWS console, and Pulumi docs. Only one of those doesn't ship a dark mode theme. The blast of photons when I switch to a Pulumi docs tab is a physical discomfort.

Modern desktop environments ship with built-in light/dark theme selection options. Modern web browsers persist these settings down into their rendering engines. Modern stylesheets pay attention to those settings to deliver a consistent user experience regardless of the user's preference.

CamSoper:
No longer relevant due to #19750

PR #19750, opened 2026-06-17
docs: add dark mode to /docs

### LINKED ISSUE #19520: Dark mode for docs
Explore effort required in bringing dark mode to /docs. (Excluding marketing pages, registry and guides).

Related: https://github.com/pulumi/docs/issues/11670, https://github.com/pulumi/docs/issues/4526

jeffmerrick:
Was actually not as complicated as I thought it might be, would be good to get a few eyes clicking around to see if I missed anything.

---

### Proposed changes

Adds a **light / dark / system theme toggle** to the `/docs` section (Fixes #19520).

The approach is light-first — dark is a pure override on the existing light base, so light mode and all non-docs pages are unchanged:

- **Tokens** — semantic `--docs-*` tokens in the new `theme/src/scss/docs/_docs-theme.scss` are re-pointed under dark; raw color scales never invert.
- **Toggle + no-FOUC** — an inline script in `head.html` sets `data-theme` on `<html>` *only* on docs pages; `docs-theme.ts` owns the toggle and activates Algolia's dark theme.
- **Scoping** — enforced by both the docs-only inline script and `body.section-docs`-gated dark rules.
- **Assets** — sidebar theme-toggle partial, per-mark `-on-dark.svg` cloud/language/secrets-provider icons, and a light footer logo.

### Related issues

Fixes #19520, #11670


CamSoper (approved):
Image

cnunciato (approved):

PR #19831, opened 2026-06-24
Add HCL templates to template pages

cnunciato:
Adds **HCL** as a selectable language across the architecture template pages, paired with the new HCL templates in pulumi/templates#1141.

- Adds `hcl` as an option to the language chooser and partials
- Adds `- hcl` to each page's `template.languages` and to the `meta_desc`/index copy.
- Makes the config/outputsections HCL-aware: HCL templates use snake_case names with some different defaults, extra/dropped variables, and different outputs, so each diverging definition list, example command, and intro is wrapped in language choosables — non-HCL readers are unchanged.
- Adds an HCL walkthrough for the static-website/aws "Adding a custom domain" example and a cache-settings snippet.
- Corrects the kubernetes/aws `eksNodeInstanceType` default (`t2.medium` → `t3.medium`) to match the actual templates.

⚠️ **Do not merge before pulumi/templates#1141 merges.** The HCL templates these pages reference (`<prefix>-hcl`) won't exist in `pulumi/templates` until that PR lands.

Closes https://github.com/pulumi/marketing/issues/1771.

### DIFF
--- a/content/templates/_index.md
+++ b/content/templates/_index.md
@@ -1,6 +1,6 @@
 ---
 title_tag: 'Pulumi Templates: Deploy Infrastructure to AWS, Azure, & Google Cloud'
 title: Pulumi Templates
-description: Pulumi templates are starter projects for common cloud architectures on AWS, Azure, and Google Cloud, available in TypeScript, Python, Go, .NET/C#, and YAML.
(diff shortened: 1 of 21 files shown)


CamSoper (approved):
This looks great, @cnunciato!

PR #19844, opened 2026-06-24
Fix YAML frontmatter indentation rejected by stricter parsers

CamSoper:
## Problem

Eight content files have multi-line double-quoted YAML scalars in their front matter whose continuation lines are indented at (or below) the level of their key. The js-yaml version currently in the lockfile tolerates this, but newer/stricter parsers reject it as `deficient indentation`. When that happens, the markdown linter (`scripts/lint/lint-markdown.js`) — which parses every file's front matter with `js-yaml` — throws on these files and fails `make lint`.

This is what's blocking the Dependabot `js-yaml` 4.2.0 → 5.0.0 bump (#19803), but the affected files are technically malformed YAML regardless of the bump.

## Fix

Re-indent each continuation line so it sits deeper than its key, making the front matter valid under strict YAML. Because YAML strips leading whitespace from double-quoted multi-line scalars, the parsed string values are **unchanged** (verified by parsing every file before and after and diffing the result).

Affected files:

- `content/testimonials/_index.md`
- `content/blog/aws-cdk-on-pulumi-1.0/index.md`
- `content/blog/azure-native-v3/index.md`
- `content/blog/copilot-in-vscode/index.md`
- `content/blog/faster-secrets-management/index.md`
- `content/blog/java-1-0/index.md`
- `content/blog/pulumi-release-notes-114/index.md`
- `content/blog/python-uv-toolchain/index.md`

Two small extras in `aws-cdk-on-pulumi-1.0`: the `meta_desc` (which was the broken multi-line value) is collapsed to a single line, and an `amd` → `and` typo in it is corrected.

### DIFF
--- a/content/blog/aws-cdk-on-pulumi-1.0/index.md
+++ b/content/blog/aws-cdk-on-pulumi-1.0/index.md
@@ -1,8 +1,7 @@
 ---
 title: "Announcing the 1.0 release of AWS CDK on Pulumi"
 date: 2024-12-02T08:00:00-07:00
-meta_desc: "Enhanced support of AWS CDK constructs from within Pulumi. Combine Pulumi and AWS CDK
-resources amd use Pulumi Cloud Platform to manage CDK"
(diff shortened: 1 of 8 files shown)


CamSoper:
@claude #new-review

PR #19845, opened 2026-06-24
Add monthly updates section and JSON/RSS feeds to the releases page

jeffmerrick:
Adds `changelog` items that can be posted in addition to releases.  

Pulled from the blog, #product-updates Slack channel and the GitHub releases to build the items. These all need review to make sure they make sense and nothing is missing. We'll want to wait until July to merge so we can get any features between now and the end of the month in. 

In addition, to assist with https://github.com/pulumi/pulumi-service/issues/43714, I've added tier labels to the features, including on the May release. These also need review for accuracy!

---

### Proposed changes

Restructures `/releases` around per-item changelog pages and adds machine-readable feeds:

- **Per-item changelog pages** — each item is its own content file at `content/releases/changelog/<slug>.md` (title, date, optional `tier` badge in front matter, full markdown body) with a minimal detail page at `/releases/changelog/<slug>/`.
- **Auto-discovered list page** — no manifest to maintain. `/releases/` ranges over the section's pages, groups them by month, and renders full releases as large cards and changelog items as linked rows with their dates. The accordions are gone.
- **Modal navigation** — clicking a changelog row updates the URL via `pushState` and opens the detail page's content in a modal (`theme/src/ts/releases.ts`); back/forward buttons close and reopen it. Navigating directly to the URL (or clicking without JS) renders the full detail page.
- **Simplified feeds** — `/releases/index.json` and `/releases/rss.xml` emit one entry per changelog item and one per release (title, date, url, description, tier), linking to each page's own URL instead of flattening release cards into individual feed items.
- **Real ship dates** — each item is dated from its source: the blog post, the #product-updates Slack announcement, or the GitHub release.
- Populates **June 2026** (nine items) and **May 2026** (eight items that shipped outside the May release; items already on the release page are excluded as duplicates).
- **Product-tier badges** on changelog rows and detail pages, and on the May release cards.

Related: https://github.com/pulumi/pulumi-service/issues/43714


cnunciato (commented):
I dig this, but it could use a content scrub (happy to do that), and probably some dates, too (and a descending sort). I also feel like the section kinda wants to be a changelog of sorts, where each individual item has a bit more room to breathe (multiple short paragraphs maybe, possibly an image or inline video), and doesn't have to be expressed as a block of inline Markdown in YAML. (Individual files would be easier to author and track.) The clustered-up CLI release summaries are nice, too. ✨

jeffmerrick:
@cnunciato Are you thinking each individual update would live in it's own file or we'd move the groups of items for each month into their own file? My only push back on each in their own file is that they so often have an accompanying blog post or something to link to with more detail I wonder how often we will need to be adding much extra content into each item?

I'm down for restructuring how dates work so they're on the individual items, it would probably make the RSS feed more useful as well. I'd imagine for a big release we could just backdate or forward date the individual items even if they all get published at the same time.

cnunciato:
>Are you thinking each individual update would live in it's own file

@jeffmerrick I think so yeah. The more I think about it, the more it feels like these are just tiny blog posts, with their own Markdown files and folders, [like Buildkite's](https://buildkite.com/resources/changelog/), addressable by their own URLs (e.g., `/releases/changelog/some-slug`) that we'd surface here, maybe in the blog, in the Console, etc. The design you have here would just pull from this set. This way, changelog posts _could_ link to accompanying blogs, docs, etc., but for smaller features that don't call for a full post (like the head's up menu in the console), they'd be able to live here, with an image or video, get their own URLs (for sharing on social, SEO, etc.), and we could even give them a full page of their own (e.g., at `/releases/changelog`). (So in the Console, for example, you'd see the first few titles, each linked to its own page, and then maybe a "See all" link that'd point to the full list.) 

What do you think? I guess I don't see much _downside_ to this model, since the design we have here would be the same, but would love to hear your thoughts. I admit that I'm biased, as I saw this work well at Buildkite, but yeah. 😄

jeffmerrick:
@cnunciato Made the changes we discussed, ready for you to take a look at again!

cnunciato (approved):
I think we're good! Thanks @jeffmerrick — this is an awesome addition.

PR #19855, opened 2026-06-24
Improve discoverability of the LLM/Markdown widget in docs

### LINKED ISSUE #17557: Improve discoverability of the LLM/Markdown widget
It's great that we have this widget, but it's a little hard to find, and often gets pushed below the fold on pages with long TOC lists:

Image

One emerging convention is to situate this widget inline with the page title, floated right, which makes it much more discoverable:

Image

Image

Image

We should consider something similar to make ours equally easy to find and use.

jeffmerrick:
Adjusted some other things while I was in there, some colors and the scroll to top button.
Also removed the 'Edit this page' link, but left 'Request changes'.

---

### Proposed changes

Reworks the docs right-rail LLM/Markdown actions to make them more discoverable: the single "Copy Page" dropdown trigger becomes a split **Copy page** button (with a caret for the remaining LLM options) alongside a standalone **View as Markdown** button, both moved above the table of contents. Adds a reusable `.btn-split` style, restyles the scroll-to-top button as a sticky `.btn-outline` icon button anchored to the content column, and wires up dark-mode tokens for the affected headings. Also removes the legacy "Edit this Page" link and the now-unused `.btn-scroll-top` style.

### Related issues

Closes #17557

### DIFF
--- a/layouts/docs/list.html
+++ b/layouts/docs/list.html
@@ -47,6 +47,8 @@
                 {{ if .Params.openapi_schema }}
                     {{ partial "openapi/schema-component.html" . }}
                 {{ end }}
+
+                <pulumi-top-button></pulumi-top-button>
             </div>
             {{ if not .Params.norightnav }}
             <div class="docs-table-of-contents docs-toc-desktop">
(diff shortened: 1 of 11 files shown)


CamSoper (commented):
@claude #update-review

CamSoper (approved):

PR #19871, opened 2026-06-25
Generate branded OpenGraph meta images for /what-is and /docs

jeffmerrick:
Adds automatic og image generation. I'm not sure this is the best approach, so interested to hear feedback. ~~I went with committing everything instead of build-time generation for two reasons: 1. with the number of pages, could add a few minutes to build time and 2. new images will be able to be reviewed when they are added instead of being somewhat hidden.~~  See discussion below.

I started with /docs and /what-is as they're predictably structured and didn't have legacy og images to wrestle with.

**Follow up work:**
- All other marketing pages, need to audit old off brand meta images
- Events (may need some special treatment)
- Bring these into the blog and blog og skill with that redesign work

fn-concat

opengraph-docs

---

## Summary

Adds a Satori-based generator that renders branded **1200×628 OpenGraph/social cards** at author time (committed, not built) and wires them up for the **what-is** and **docs** sections — replacing placeholder/old-logo and generic shared meta images with on-brand, per-page cards. Zero build-time cost.

## Changes

- **New generator** `scripts/generate-meta-images.mjs` (Satori → resvg) with two templates: a centered **title** card (what-is) and a 4-field **`og-info`** card — section label, sub-section label, title, description — built from the Figma "Social assets — banners" design.
- **what-is:** per-page title cards; removed 54 old `meta_image` fields and deleted 53 placeholder PNGs (old logo / "PLACEHOLDER" images).
- **docs (all ~810 pages):** per-page `og-info` cards (badge "Docs" + nav-area label from `docs_menu_sections.yml` + page title + `meta_desc`). Removed the shared `docs-meta` from 378 frontmatters **and** the `cascade.meta_image` in `_index.md` (the real site-wide switch); `/docs` keeps a custom card (`static/images/docs/opengraph-docs.png`).
- **Shared precedence partial** `layouts/partials/meta-image-url.html`: path-based lookup; a page-level `meta_image` always overrides the generated card. Used by both `head.html` (OpenGraph/Twitter) and the JSON-LD schema collector, so they stay in sync.
- **Commit method:** cards live in `assets/images/generated/`; `make meta-images` regenerates, `make meta-images-check` (wired into `make lint`) fails on drift.
- **Deps:** `satori`, `@resvg/resvg-js`, `opentype.js` (title auto-fit). Monaspace Neon converted woff2→ttf for Satori.

## Notes

- Only `what-is` and `docs` are enabled; tutorials/events are the natural next sections (they'd reuse `og-info`).
- ~865 generated PNGs are committed by design (decided to commit rather than generate at build time).


CamSoper (changes requested):
Instead of tying it to lint, what if we instead had a `push` github workflow that generates and adds it automatically to any docs/what-is PR that doesn't already have it?

CamSoper (approved):
~~Automated rubber stamp approval applied :shipit:~~

Nooooope, nope, nope. Claude got too eager. 🤣

CamSoper (changes requested):
Re-marking "request changes" with my original comment:

> Instead of tying it to lint, what if we instead had a push github workflow that generates and adds it automatically to any docs/what-is PR that doesn't already have it?

jeffmerrick:
> Instead of tying it to lint, what if we instead had a push github workflow that generates and adds it automatically to any docs/what-is PR that doesn't already have it?

This is a bit outside my experience but it seems like that could complicate things, (especially as we might expand this to other parts of the site) so maybe actually generating them at build time would actually be the better approach that would also address your feedback? Committed that change - I was mostly worried about build time, but it looks like it only added around 38s even without the caching!

CamSoper (changes requested):
Oh wait! I just realized we can do better than that! We can cache the output using GitHub build caching and then we won't even add the 38s!!! We'll just need milliseconds to generate any images that don't already exist!

Mind if I make a few commits to this branch, @jeffmerrick?

jeffmerrick:
@CamSoper That should already be there! 🎉 

```
Generating meta images...

meta-images: 0 rendered, 865 cached, 0 failed | 21ms total
```

Check it out and make sure it looks right, but yes feel free to push up anything to the branch too!

CamSoper (approved):
Looks great!

cnunciato (approved):
Nice! This is awesome. ✨ Nice work on the caching bit, too — TIL!

PR #19879, opened 2026-06-25
soai: drop "What we found" lead-in and remove em dashes

alexleventer:
Follow-up polish to the merged [State of Agentic Infrastructure](/state-of-agentic-infrastructure/) landing page (#19821, now live).

## Changes
- Remove the **"What we found"** summary card that sat between the hero and the numbered sections, so the page goes straight from the intro into the survey sections.
- Replace em dashes throughout the prose with colons, semicolons, and commas. Number ranges (e.g. `201–5,000`) keep their en dashes.

No data or chart values changed.

### DIFF
--- a/content/state-of-agentic-infrastructure/_index.md
+++ b/content/state-of-agentic-infrastructure/_index.md
@@ -27,27 +27,19 @@ hero:
       label: The six-month outlook and sentiment
       anchor: the-six-month-outlook-and-sentiment
 
-findings:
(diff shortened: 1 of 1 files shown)


alexleventer:
Superseded by #19882, which targets a fresh branch instead of reusing the already-merged `jeffmerrick/agentic-infra-survey-page` branch (that reuse was causing the stale/conflicting diff).

PR #19882, opened 2026-06-25
soai: drop "What we found" lead-in and remove em dashes

alexleventer:
Follow-up polish to the live [State of Agentic Infrastructure](/state-of-agentic-infrastructure/) landing page (added in #19821).

## Changes
- Remove the **"What we found"** summary card that sat between the hero and the numbered sections, so the page goes straight from the intro into the survey sections.
- Replace em dashes throughout the prose with colons, semicolons, and commas. Number ranges (e.g. `201–5,000`) keep their en dashes.

No data or chart values changed. Single-file diff against current master.


alexleventer:
Superseded by #19883 (branch renamed to `soai-cleanup`).

PR #19904, opened 2026-06-26
Make Pulumi Mermaid theme a portable module with full diagram coverage

jeffmerrick:
Updates to the mermaid theme for most diagrams/charts (a couple minor ones aren't possible to style, but we aren't using them anyhow).

Next up is to add the theme to the brand guidelines and into the slidev theme. 

---

### Proposed changes

Extracts the Mermaid renderer out of `head.html` into a dedicated `mermaid.html` partial (loaded lazily, only on pages with diagrams) and moves the Pulumi-brand theme into a portable, dependency-free module (`static/js/pulumi-mermaid-theme.js`) that can be dropped into other codebases, slide decks, or notebooks. The theme is expanded from flowchart/sequence coverage to all Mermaid v11 diagram types (git, journey, timeline, xychart, radar, treemap, ER, kanban, packet, mindmap, etc.), with every color sourced from a Pulumi brand token. Each diagram now resolves its light/dark theme from the nearest `data-theme` ancestor and diagrams render in theme-grouped sequential batches to avoid Mermaid's shared-state races. Also fixes flex-based centering in `_code.scss` that broke Mermaid's container-width measurement (notably gantt) and stops global content `p` styling from bleeding into diagram labels.

screencapture-localhost-1313-mermaid-test-2026-06-26-12_59_54


CamSoper (approved):
This looks great! Someone was just asking me the other day about our Mermaid support.

PR #19917, opened 2026-06-27
Revise language around HCL

cnunciato:
A lot of our content frames HCL and "configuration languages" as less desirable than programming languages. This PR softens that rhetoric and repositions general-purpose languages as a flexible choice rather than categorically superior, while noting that HCL is now a supported peer language.

Closes https://github.com/pulumi/marketing/issues/1772.


djperlovsky (approved):
LGTM!

PR #19940, opened 2026-06-29
Redesign language chooser and add light/dark code theme

### LINKED ISSUE #19928: Update language chooser to handle more languages
Explore UI options to make 7+ languages fit well in the language chooser. Ideally the default view should still show the breadth of languages (so maybe not a dropdown).

jeffmerrick:
In addition to updates to the language chooser to make it fit more languages more easily, this updates all code blocks in /docs to be theme-aware instead of always dark mode.

~TBD: what icon to use for HCL when it's added, or consider if we want language icons at all.~

**Follow up work:** replicate these changes in the registry.

Screenshot 2026-06-29 at 2 44 42 PM
Screenshot 2026-06-29 at 2 44 47 PM

---

Resolves #19928.

Redesigns the language chooser into a bordered card with a left-label toolbar, and makes code blocks theme-aware in `/docs` (a new Min Light theme for light mode alongside the existing Min Dark default).

### Language chooser redesign
- Wraps language choosers (`type="language"`) in a bordered card with a gray toolbar: the active language's colored logo + full name on the left, underline-style abbreviation tabs (TS, PY, GO, …) on the right.
- Tabs now show uppercased file-extension abbreviations instead of full names, so 7+ language options fit on one row.
- Adds a `logo` (and optional `logoOnDark`) field to each `SupportedLanguage`; YAML ships a dark-mode wordmark variant swapped via the docs `.docs-logo-light`/`.docs-logo-dark` classes.
- Example-program choosers (code + GitHub footer) run flush to the card edges instead of double-padding; loose prose stays uniformly inset.
- Headless/inline choosers (`option-style="none"`) keep no chrome — the toolbar wrapper is hidden alongside the list.
- Updates the C# tech logo SVG to the full-color C# diamond.

### Theme-aware code blocks (/docs)
- Adds `theme/src/scss/docs/_code-light.scss` — the Min Light code theme (Violet 50 surface) for docs light mode, mirroring the Min Dark dark-mode block. Comments are bumped for contrast; everything else is the canonical Min Light palette.
- Imported unlayered from `main.scss` so it beats the layered component rules, scoped to `html:not([data-theme="dark"]) body.section-docs` so it can't touch non-docs pages or dark mode.

### Code-panel chrome refinements
- Copy button is now a transparent ghost chip with a single muted foreground that reads on both the dark panel and the docs-light surface (removes the old `code-mode-dark` overrides).
- Copy-button and example-program footer tooltips now point left of their trigger (they sit at the panel's top/bottom-right corners where an upward tooltip would overflow).
- Example-program footer matches the code block's background as its bottom lip in both modes; GitHub icon shrunk.
- Note/callout icon styling scoped to `div.icon-and-line` so it no longer leaks onto copy-button icons.

### Tests
- Adds e2e coverage for the new language toolbar (logo label + abbreviated tabs) and confirms non-language choosers still render plain full-name tabs with no toolbar.


CamSoper (approved):
I love it! You're a rockstar, Jeff.

cnunciato (commented):
Ok so first off, this looks fantastic. I love it. ✨ 

Still poking, but wanted to get you a first pass to work with some of the more noticeable things I've seen so far.

1. We use choosers to toggle blocks of content on the page, which currently produces what looks like an empty container ([new](http://www-testing-pulumi-docs-origin-pr-19940-741c5300.s3-website.us-west-2.amazonaws.com/docs/iac/get-started/aws/) vs. [current](https://www.pulumi.com/docs/iac/get-started/aws/)):
    
    image

1. Variations on the chooser look different. Should we change these too, or leave them the same?

    image

3. Many have extra padding — might be a quirk of Hugo rendering?
    
    image
    
4. I assume we're pulling this from the extension, but it'd be great to have the label here be `C#` rather than `CS`:

    image

5. The wrapping is a bit awkward on mobile. Should we maybe hide the full name and the icon below a certain width? 🤔

    image

jeffmerrick:
@cnunciato Good catches and feedback, thank you!

1. Fixed with an alternate style for the standalone ones.

Screenshot 2026-07-09 at 3 59 23 PM


2. I've updated these, the only downside is they can be used elsewhere on the marketing site and look a little weird there, but looks like currently only on /solutions (which probably could use some TLC anyhow.)

Screenshot 2026-07-09 at 3 59 30 PM


3. This was by design since they can contain a single code block or several code blocks plus text. BUT, I've made a change so if there is a single code block it will be flush.
Screenshot 2026-07-09 at 4 00 00 PM


5. Done! (For F# too.)
6. Fixed!
Screenshot 2026-07-09 at 4 02 49 PM

Unrelated update: I changed the info notes back to blue instead of violet - since the code background is violet in light mode now they were really blending in with eachother too much.

Also bonus: ran across and fixed a bug where tabs weren't working when there were mixed chooser options (clicking YAML here https://www.pulumi.com/docs/iac/concepts/stacks/#outputs doesn't work.)

cnunciato (approved):
Looks great! Will keep an eye out for any additional issues, but so far, everything looks solid. Well done!

PR #19947, opened 2026-06-29
Clarify that teams have both roles and entity access in RBAC intro

### LINKED ISSUE #19946: RBAC: Note that entity rules can be applied directly to teams without using roles
This is to allow team admins to grant access rules to entities under the team's control without the wide-ranging org level permission to create roles.

We need to run a test to ensure that we know the exact scope of the permissions that can be applied by the grantor - what entities are available for the grantor to grant access.

jkodroff:
Clarifies on the RBAC overview page that teams accumulate access two ways: through assigned roles and through entity access granted directly on stacks, environments, and Insights accounts (outside of any role).

## Changes
- Updated the Teams bullet under "Where roles apply"
- Reflected direct entity access in the permission-accumulation diagram
- Retitled the section to "Team roles and entity access grants" and noted direct grants there

Fixes https://github.com/pulumi/docs/issues/19946

### DIFF
--- a/content/docs/administration/access-identity/rbac/_index.md
+++ b/content/docs/administration/access-identity/rbac/_index.md
@@ -26,7 +26,7 @@ This model lets you:
 Roles apply to these kinds of principals in Pulumi Cloud:
 
 - **Users**: Each organization member has exactly one organization role (Admin, Member, Billing Manager, or a custom role).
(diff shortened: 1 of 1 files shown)


CamSoper (approved):

---

PR #19968, opened 2026-06-30
Tighten FUD-sounding language about DIY backend failure recovery

CamSoper:
### Proposed changes

Rewords the checkpoints paragraph in the [state and backends](https://www.pulumi.com/docs/iac/concepts/state-and-backends/#checkpoints) concept page.

The previous sentence read like FUD:

> ...but blob storage backends use a less transactional protocol that **may have more difficulty** recovering from partial failures.

Following the approach Joe proposed in the `#docs` Slack thread, this replaces the hedge with a definitive, concrete statement and gives the reader somewhere to go:

- "May have more difficulty" → DIY backends **cannot transparently recover from certain kinds of partial failures**, because they are fundamentally limited by the non-transactional protocols of blob storage.
- Names the concrete consequence — that limitation is why disabling checkpoints is the only blunt performance lever available on DIY backends — and links to [journaling](/blog/journaling/) (how Pulumi Cloud avoids the tradeoff) and the [`PULUMI_SKIP_CHECKPOINTS`](/docs/iac/cli/environment-variables/) environment variable.

Single-paragraph prose change; no menu, data, code, or image changes.

N/A

### Related issues (optional)

N/A

---
_Generated by [Claude Code](https://claude.ai/code/session_01TpUT8VsMoz6tokDytp6cgQ)_


jkodroff (approved):
This reads real good.

PR #19972, opened 2026-06-30
Refresh Organizing Projects & Stacks guide

### LINKED ISSUE #19381: IaC: Refresh Organizing Projects & Stacks Guide
https://pulumi.com/docs/iac/guides/basics/organizing-projects-stacks/

- We treat it like a binary choice: monolithic or microstacks. Instead, we should talk about monolithic and microstacks as possible solutions (and also include the canonical example of network/clusters/workloads) within the context of the design tradeoffs: blast radius, ownership (permissions), alignment with repositories, maintenance overhead (both wall time and realtime, e.g. the time it can take to deploy an application - Pulumi stacks with thousands of resources can be slow, and needlessly so if only a small number of resources in the stack have changed).
- If we are going to have examples, they need to cover all languages.
- Remove all images. Use mermaid diagrams where necessary to explain concepts and patterns.
- We should not be encouraging users to organize resources in files that align with cloud services. Instead, they should deal with entire workloads or aspects of the infrastructure, e.g., in a monolithic example: networking, clusters, workload(s).
- Configuration helpers are not a real world pattern (at least that I've seen).
- Whether to do app code with or not with infrastructure is a great topic. It usually works well for self-service models where app teams write Pulumi code, possible consuming components from Pulumi Private Registry. It does not work so well for full-fledged IDP where you want the IDP to own all the code in a repo since IDPs typically produce IaC, and having it mingled with app code will create ownership issues. (Expand upon all of this.)

jkodroff:
Refreshes the [Organizing Projects & Stacks guide](https://www.pulumi.com/docs/iac/guides/basics/organizing-projects-stacks/) per #19381.

## What changed

**Reframed around design tradeoffs instead of a binary.** The page no longer presents monolithic *vs* micro-stacks as a two-way choice. It opens with the four tradeoffs from the issue — **blast radius**, **ownership/permissions**, **repository alignment**, and **maintenance overhead** (including the point that large stacks are slow to deploy even when little changed) — and frames everything else as choosing a granularity along those axes.

**Granularity is now a three-tier spectrum:**

1. **Monolithic** — one project, everything together (where most teams should start).
2. **Layered stacks** — the recommended default beyond a starter project: split along major infrastructure layers (**networking → clusters → workloads**), connected by stack references. (This is the "medium granularity" we actually recommend; it is deliberately *not* called micro-stacks.)
3. **Micro-stacks** — repositioned as a **narrow, conditional** pattern for genuinely independent services (e.g. many Lambdas), with both its upside (highly targeted, fast deploys) and downside (per-stack maintenance overhead) called out.

**Other issue items:**

- **Removed both JPG diagrams**, replaced with Mermaid (which is automatically light/dark themed in `/docs`). No redirects needed — the images were referenced only by this page.
- **All six languages** now covered in examples (TypeScript, Python, Go, C#, Java, YAML), up from TypeScript + Go.
- **In-project code organization** rewritten to organize by workload/aspect (networking/clusters/workloads) rather than cloud-service-named files; the **"configuration helpers"** pattern was removed.
- **Expanded** the app-code-with-or-without-infrastructure discussion into the **self-service vs. full-IDP** ownership distinction.
- Trimmed the verbose multi-repo/stack-reference section and fixed internal links to current canonical paths.

## Testing

- `make build` succeeds; the page renders all 3 Mermaid diagrams and all 6 language tabs in each chooser.
- `make lint` passes (0 errors).
- No broken references to the removed images.

**Reviewer note:** please give the two new Mermaid diagrams a quick visual pass in docs **dark mode** (theme toggle at the bottom of the docs sidebar). They use the repo's automatic theming with no hand-set colors, so they should flip cleanly.

Fixes #19381


jkodroff:
@claude #update-review

CamSoper (approved):
Sick Mermaid diagram 🧜🏻‍♀️

CamSoper:
@claude Convert the instances of "it is" to "it's" (as per the style findings) and then #update-review

PR #19975, opened 2026-06-30
Fix deprecated "Pulumi Service" naming: flag-only Vale rule + ESO doc accuracy

CamSoper:
### Proposed changes

Follow-up to the merged #19973. Cleans up the retired **"Pulumi Service"** name (renamed to Pulumi Cloud in 2022), surfaced by the automated content review in #19963.

**1. Vale rule — stop auto-correcting toward a single product name**

#19973 redirected `pulumi service` → `Pulumi Cloud`. But the ESO page below shows "Pulumi service" is *ambiguous* — there it means a **Pulumi ESC environment**, not Pulumi Cloud generally. A blanket auto-replace just trades one wrong fix for another. So:

- Remove the `pulumi service` → `Pulumi Cloud` substitution from `Nomenclature.yml`.
- Add `styles/Pulumi/DeprecatedProductNames.yml`, an **existence** (flag-only) rule that warns on the deprecated name and asks the author to pick the correct current product (Pulumi Cloud / Pulumi ESC / Pulumi Deployments). This also catches the **capitalized** "Pulumi Service" the old lowercase-only substitution never saw, and deliberately does not match the plural "Pulumi Services."

**2. ESO doc — name the PushSecret destination precisely**

`PushSecret` writes into the **Pulumi ESC environment** configured in the `SecretStore` (`provider: pulumi` with `organization`/`project`/`environment`; `apiUrl` defaults to the ESC API at `https://api.pulumi.com/api/esc`) — consistent with the rest of the page ("Pulumi ESC secrets…", "This Pulumi ESC environment…"). The old text said the vague (and now deprecated) "pushed to the Pulumi Service."

(The old "With the latest release of Pulumi ESC…" qualifier is dropped: `PushSecret` for the Pulumi provider shipped in the same External Secrets Operator release as the provider itself — 0.10.0 — which the top of the page already states, so there's no version window worth calling out.)

#### Note on the new rule's scope

The flag-only rule is non-blocking (Vale `warning`; `lint-prose` never gates) and only surfaces on changed `content/docs` & `content/blog` files. It will warn on existing "Pulumi Service" / "Pulumi Service Provider" prose when those files are next edited — intended, since those should become Pulumi Cloud / the specific product / "Pulumi Cloud provider." No mass rewrite is performed here.

N/A

### Related issues (optional)

Follow-up to #19973; corrects content introduced by #19963.

### DIFF
--- a/content/docs/esc/integrations/kubernetes/external-secrets-operator.md
+++ b/content/docs/esc/integrations/kubernetes/external-secrets-operator.md
@@ -20,7 +20,7 @@ aliases:
 - External Secrets Operator supports multiple different sources, so you can use the same operator to manage secrets and configuration from different sources.
 - Take advantage of the advanced features of the secret provider, such as encryption of data at rest and scenarios like secret rotation.
 
-Since version `0.10.0` External Secrets Operator supports [Pulumi ESC as a secret provider](https://external-secrets.io/latest/provider/pulumi/).
+Since version `0.9.13` External Secrets Operator supports [Pulumi ESC as a secret provider](https://external-secrets.io/latest/provider/pulumi/).
 
 ## Authentication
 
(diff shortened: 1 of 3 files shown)


CamSoper:
@claude #update-review

PR #19997, opened 2026-07-01
Add event social-card generator and /event-meta-image skill

jeffmerrick:
This adds auto-generated meta images to events using their frontmatter. In addition, it adds a skill that can be used to add to the images for events with external speakers and generate all the sizes needed for different event platforms.

**How this works in practice:**
- For most events with only Pulumi hosts, the frontmatter will be sufficient to automatically generate a good image at build time, the skill would only need to be used to generate additional sizes if needed.
- For events with external hosts, if a speaker image is in the frontmatter, that will be used and generate a good enough image, but the company logo will be missing, the skill can then optionally be used to generate an image to be used as an override.
- The skill can be run without an accompanying event, if you just need images.

**What I've replaced:**
All events will now get an auto generated image, I ran the skill on only those that are `upcoming` or `on-demand` (assuming the older past events will not be linked to much anymore.)

Preview:
pr-clip-top

Full list: ([giant image](https://github.com/user-attachments/assets/495a711c-0836-45b9-bc5d-99855cd06b34))

---

### Proposed changes

Events can leave `meta_image` blank and get an on-brand social card generated at build time, plus a `/event-meta-image` skill for richer cards.

- **Build-time renderer** (`scripts/meta-images/events.mjs`, `render-event.mjs`, `lib.mjs`): shared Satori renderer that draws overline, title, presenter "With…" line, Pulumi logo, and presenter photos over new brand accent-line SVGs.
- **Square second card**: emits a 1:1 (628x628) card added as a second `og:image` in `head.html`; the landscape card stays the primary and `twitter:image`. Dims are only declared for the generated card, not custom overrides.
- **New partials**: `meta-image-key.html` centralizes the generated-image key lookup; `meta-image-square-url.html` resolves the square card (auto for events, or any page's `meta_image_square`). `meta-image-url.html` now uses the shared key partial.
- **Event archetype**: documents `meta_image`, `meta_image_square`, and `overline` override fields.
- **`/event-meta-image` skill** (`.claude/commands/event-meta-image/`): enriches cards with external co-presenters and partner logos and emits extra social sizes; ships a logo asset set and reference docs.

### Related issues (optional)


dirien (approved):
LGTM!

cnunciato (approved):

PR #20028, opened 2026-07-01
Adopt @pulumi/design-tokens JSON as Tailwind color source of truth

kimberleyamackenzie:
## Summary
- Add `@pulumi/design-tokens` to the Hugo theme and generate Tailwind v4 `@theme` color variables from `tokens/core/primitives.json` and `semantic.json`
- Replace the inline ~170-line color block in `_theme.scss` with generated output; docs-specific tokens (docs-* colors, breakpoints) remain in `_theme.scss`
- Typography is unchanged — still defined locally (not sourced from design-tokens yet)

## Test plan
- [ ] Run `yarn build:color-theme` in `theme/` and confirm no diff
- [ ] Compile theme SCSS (`sass src/scss/main.scss`) without errors
- [ ] Spot-check docs site light/dark mode for color regressions

Made with [Cursor](https://cursor.com)


jeffmerrick (approved):
Looks good!

PR #20040, opened 2026-07-02
Add what-is page: Is Kubernetes Infrastructure as Code?

joeduffy:
## Summary

Adds a new `/what-is/` page answering the evaluative question "Is Kubernetes Infrastructure as Code?" — a high-intent, high-GEO-value query (Tier 2/3 keyword per our SEO/GEO content plan) that LLMs are frequently asked directly.

- **Answer-first opening**: no, Kubernetes manifests are declarative config, not IaC on their own — a general-purpose tool like Pulumi or Terraform has to provision the cluster itself for the full stack to be infrastructure as code.
- **Core distinction**: config layer (Kubernetes API objects) vs. infrastructure layer (cluster, VPC, IAM), framed as a single litmus test ("does this tool provision the cluster, or assume it already exists?").
- **Verified code example** (TypeScript + Python, `{{< chooser >}}` shortcode): provisions an EKS cluster with `@pulumi/eks`, wires a `k8s.Provider` to `kubeconfigJson`, and deploys a Helm chart with the current `helm.v4.Chart` resource — checked against the live EKS guide and current Helm v4 docs.
- **Wiz proof point**: pulled verbatim from `/case-studies/wiz/` (thousands of clusters, hundreds of data centers, 1M+ resources, hundreds of thousands of daily updates) — corrected from the original board task's imprecise phrasing.
- **Helm/Kustomize/Terraform comparison sections**, each answering a direct evaluative question.
- **Six FAQ Q&As** structured as `### <Question>?` headings under a `## Frequently asked questions...` H2, which the site's schema auto-detection (see `SCHEMA.md`) picks up as `FAQPage` schema for `type: what-is` pages with question headings — no manual `schema_type` override needed.

## Content overlap with an existing page

`content/what-is/infrastructure-as-code-for-kubernetes.md` ("Kubernetes Infrastructure as Code: Tools and Best Practices") already covers this topic space and updated recently. I deliberately differentiated rather than duplicating:

- **This page** answers the yes/no evaluative question people (and LLMs) actually ask, with one proof-quality code example and a sharp definitional test.
- **The existing page** stays the deeper tools/best-practices roundup (GitOps, secrets, testing, migration, etc.).
- Added a reciprocal link in both directions — this page links to the existing one in "Learn more," and I added this page as the first "Related reading" entry on the existing page.

Flagging this explicitly in case the preferred move is to fold the new content into the existing page instead of shipping two pages — happy to consolidate if that's the better call, but shipped as a separate page per the board task's suggested URL and title.

## Internal links verified

All internal links (`/registry/packages/kubernetes/`, `/registry/packages/eks/`, `/docs/iac/comparisons/terraform/`, `/docs/iac/comparisons/helm/`, `/docs/iac/guides/clouds/aws/eks/`, `/case-studies/wiz/`, `/blog/kubernetes-chart-v4/`, `/blog/easily-create-and-manage-aws-eks-kubernetes-clusters-with-pulumi/`) were confirmed to exist as content files in this repo before use.

---


joeduffy:
Pushed a revision addressing a good catch from the human reviewer on the initial framing.

**What changed:** The first draft took too blunt a "no" stance on whether Kubernetes itself is infrastructure as code. That undersold a real point: Kubernetes controllers reconcile declarative YAML toward desired state continuously, which is the same declarative/convergent model IaC is built on. Calling that "just config" was inaccurate and would have weakened the page's credibility with the exact technical audience (and LLMs) it's trying to earn citations from.

**Why:** The page now takes a "partly" position instead of "no." It affirms that Kubernetes' reconciliation loop genuinely is IaC-like within the scope of the Kubernetes API, and locates the real gap precisely: that convergence loop stops at the cluster boundary, so cluster creation, node groups, VPCs, and IAM sit outside it. This is a more defensible and more citable claim than a flat "manifests aren't IaC."

**Where this landed:**
- Rewrote `meta_desc` and the opening answer-first paragraph to say "partly."
- Rewrote the "What is the relationship..." section to state the shared declarative/convergent model explicitly before drawing the boundary.
- Rewrote the "What's the difference..." section so the gap is framed as "the loop doesn't reach the infrastructure it depends on," not "manifests aren't code."
- Rewrote the first FAQ answer ("Is a Kubernetes YAML manifest infrastructure as code?") to answer "within scope, yes" before noting the boundary.
- Left the Helm/Kustomize sections as an unambiguous "no" — that distinction still holds, since neither of those tools has its own reconciliation/control loop; they're templating/patching layers on top of manifests, not an independent instance of the declarative model.

Sibling-scan: this is the only page in the diff, so no other file needed the same correction.

@claude #update-review

joeduffy:
Fixed the one outstanding item from the automated review: the claim that "Terraform's HCL is a purpose-built configuration language without general-purpose loops, functions, or a native test framework" was out of date. Terraform has shipped a native HCL-based test framework (`terraform test`, `.tftest.hcl` files with `run`/`assert` blocks) since v1.6 (October 2023).

Corrected the sentence to drop the test-framework claim entirely, since it wasn't essential to the comparison and the review's suggested rewrite (dropping vs. reframing) both work — dropping keeps the sentence focused on the point that actually matters: HCL is still a purpose-built configuration language rather than a general-purpose one, so it lacks general-purpose loops and functions. That distinction is accurate and unaffected by the test-framework fix.

Sibling scan: this claim only appeared once, in the new page's Terraform comparison section — no other file in the diff repeated it.

The other findings (3 low-confidence, 3 pre-existing) all sit in the pre-existing sibling page this PR only cross-links to, not in lines this PR changes, and were correctly flagged as non-blocking.

@claude #update-review

joeduffy:
Fixed the outstanding item from round 2: the reframe commit introduced a blanket "Kubernetes has no `pulumi preview`-style plan step" claim, repeated in three places (the impedance-mismatches paragraph, the YAML-vs-IaC difference section, and the closing FAQ answer). The automated review correctly caught this as the same pattern as the earlier Terraform test-framework miss — `kubectl diff` is a native, stable kubectl subcommand that does show a pre-apply diff against the live cluster.

Narrowed all three instances: `kubectl apply` itself has no built-in plan step (it reconciles immediately), and `kubectl diff` is a separate, less integrated command producing a raw object diff rather than a dependency-aware plan across resources the way `pulumi preview`/`terraform plan` do. The core point (no integrated, cross-resource preview workflow) still holds — only the "no preview mechanism at all" framing was wrong.

On the Terraform test-framework item marked "unresolved" in the round-2 review: I confirmed the fix from my previous push (commit 5a0646e81f) is present at that line and pushed to this branch before that review ran — the review's re-analysis appears to have raced against that push and picked up a stale snapshot. Re-verified locally; the current line reads correctly.

@claude #update-review

joeduffy:
Fixed the one remaining outstanding item flagged in the latest review: the "Is `kubectl apply` infrastructure as code?" FAQ answer still said `kubectl apply` "can't ... preview a plan before applying," repeating the same overclaim already fixed in the other three occurrences (L42, L48, L145). This is the fourth and last spot — narrowed it to say `kubectl apply` "has no plan step of its own," with `kubectl diff` credited as a separate, less integrated pre-apply diff mechanism, matching the phrasing used elsewhere in the page. A repo-wide grep confirms no other occurrence of the old "no preview/plan step" phrasing remains in this file.

@claude #update-review

CamSoper (approved):

---

CamSoper:
@claude 

The lint is failing:

```
/home/runner/work/docs/docs/content/what-is/is-kubernetes-infrastructure-as-code.md:
Line File Header: Meta description is too long. Must be shorter than 160 characters.
```

Replace the meta description then #update-review

PR #20070, opened 2026-07-02
Document plugin auto-install behavior for Automation API

### LINKED ISSUE #11727: Automation API and Pulumi Plugins - what they are and how to figure out which ones you need
### Problem description

Pulumi plugins are generally handled by CLI magic but with the Automation API  you are responsible for managing them.
Although, it's basically a matter of referencing the analogous sdk name and version, this is not obvious to all users.

So, adding some documentation on plugins would help.
Points that would likely be useful to cover:
- What are plugins? Why do they exist vis a vis the SDK?
- How to find the list of available plugins and/or deduce which plugins are needed based on SDKS being used.

This page should be referenced from the `pulumi plugin` CLI page(s) as well as from the automation API page(s) where applicable.

### Affected product version(s)

N/A

### Suggestions for a fix
See thoughts above. Not sure how best to structure this information.

jkodroff:
## Summary

Resolves #11727, which asked for docs on managing Pulumi plugins with Automation API. The issue's 2021 premise — that "with the Automation API you are responsible for managing them" — is **outdated**: Automation API drives the same engine as the CLI, so registry provider plugins auto-install on `up`/`preview`/`refresh`. The existing docs' unconditional `installPlugin()` calls made manual management look mandatory when it isn't.

## Changes

- **`content/docs/iac/concepts/automation-api.md`** — new `## Plugins` section explaining plugin-vs-SDK, that registry plugins auto-install, and the narrow cases where `installPlugin` is still needed (local/parameterized packages, version pinning, pre-fetch/air-gapped).
- **`content/docs/iac/guides/building-extending/automation-api.md`** — clarifies "Configure your provider plugins" so the `installPlugin` call reads as optional for registry providers, with cross-links to the concept doc, the local-packages section, and the `pulumi plugin` CLI reference.

## Verification

Ran an inline TypeScript Automation API program (`random` provider) against an **isolated empty plugin cache** with **no `installPlugin` call**:

```
>>> NOT calling installPlugin — relying on engine auto-install
 +  pulumi:pulumi:Stack ... Downloading provider: random
 +  random:index:RandomPet my-pet created
>>> up succeeded. petName = funny-man
```

The engine auto-downloaded the plugin and the update succeeded. A control run with an explicit `installPlugin` also succeeded. Tested with pulumi CLI 3.248.0 / SDK 3.250.0, `@pulumi/random` 4.21.0.

## Follow-up (not in this PR)

The issue also requested a link **from** the `pulumi plugin` CLI pages back to Automation API. Those pages (`content/docs/iac/cli/commands/pulumi_plugin*.md`) are **auto-generated** from `pulumi/pulumi` source, so that direction needs an upstream change and can't land here.

Fixes #11727

### DIFF
--- a/content/docs/iac/concepts/automation-api.md
+++ b/content/docs/iac/concepts/automation-api.md
@@ -79,6 +79,17 @@ Unlike traditional Pulumi programs, an inline program doesn't require a separate
 
 The program's lifecycle must be fully contained within the function, callback, or closure passed as the inline program. Performing actions outside the scope of the inline program function is unsafe and can lead to unpredictable behavior.
 
+## Plugins
+
+Pulumi providers are distributed as plugins that the engine loads at runtime, separately from the language SDK you import in your program. When you run the CLI, the engine automatically downloads any missing provider plugin before an operation. Automation API drives the same engine, so this behavior is identical: for providers published to the [Pulumi Registry](/registry/), you don't need to install plugins yourself---running `up`, `preview`, or `refresh` downloads them on demand, matching the SDK version your program references.
+
+A `Workspace` still exposes an explicit `installPlugin` method (`install_plugin`, `InstallPlugin`, `InstallPluginAsync`). You need it only in specific cases:
+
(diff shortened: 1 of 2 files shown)


CamSoper (approved):

---

PR #20077, opened 2026-07-02
Show blog post publish date in search results

### LINKED ISSUE #13801: Show the publish date in the preview for blog search results
We have a lot of blog posts in our blog. I was searching to find out whether we have a post announcing Java 1.0, and it came to me that it would make the results a lot more helpful if we could add a publish date for blog post search results (assuming our search results allows us to template results by page type).

Image

cnunciato:
## What

Adds the blog post publish date to blog results in the search dialog, as requested by folks searching the blog who wanted a date on each result.

Fixes #13801.

Flying a _little_ bit blind here until the `date` field gets added to Algolia (which'll happen on release), but this should look like this once deployed:

image

## How

Threads the date through all four layers of the search pipeline, scoped to **blog** hits only (docs/registry/etc. records are unchanged):

- **`layouts/index.json`** — Hugo's search-index JSON now emits a formatted `date` (`Jan 2, 2006`, matching the blog UI) for blog pages that have a non-zero date.
- **`scripts/search/page.js`** — carries `date` onto each Algolia record (`undefined` for non-blog records, keeping them clean).
- **`theme/src/ts/algolia/autocomplete.ts`** — renders the date under the description when a hit has one.
- **`theme/src/scss/_algolia.scss`** — styles it as small muted text. Uses literal Tailwind gray classes that the docs dark-mode block auto-remaps, so it reads correctly in both themes.

## Notes

- The dates appear in the live dialog only **after the Algolia index is regenerated**, which happens in the normal build/deploy pipeline (Hugo produces `public/index.json`, then records are re-uploaded). Existing records won't have the field until then.
- `date` is a display-only attribute — deliberately not added to searchable attributes.
- Verified the theme webpack build compiles cleanly.

### DIFF
--- a/layouts/index.json
+++ b/layouts/index.json
@@ -7,6 +7,10 @@
     {{- end -}}
 
     {{- if ne $page.File nil -}}
+        {{- $date := "" -}}
+        {{- if and (eq $page.Section "blog") (not $page.Date.IsZero) -}}
+            {{- $date = $page.Date.Format "Jan 2, 2006" -}}
+        {{- end -}}
         {{-
             $index = $index | append (
                 dict
@@ -16,6 +20,7 @@
                     "href" $page.RelPermalink
                     "params" $page.Params
                     "kind" $page.Kind
+                    "date" $date
                     "ancestors" (after 1 $ancestors)
             )
         -}}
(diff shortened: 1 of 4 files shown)


dirien (approved):
LGTM!

PR #20080, opened 2026-07-03
docs(import): refocus import guide on CLI-first and program-first (bulk) approaches

### LINKED ISSUE #15053: Refocus import docs to cover bulk scenarios better
[content/docs/iac/adopting-pulumi/import/_index.md](https://www.pulumi.com/docs/iac/adopting-pulumi/import/)

We currently describe imports as having 2 approaches: `pulumi import` CLI and the import resource option.

From a user's perspective there's really only 2 recommended approaches:
1. CLI first: 
  1. Call `pulumi import` to reference a specific cloud resouce, add the state and generate code to copy into their program. 
  2. User pastes the code into their program.
2. Program first (aka batch import):   1. Write all the infrastructure code, perhaps using components etc
  2. Run `pulumi preview --import-file <import-file-path>` to generate an import file for all new resources with the names, URNs, and types already filled in, and with blank id fields that need to be filled in.
  3. User fills in the id fields for their cloud resources.
  4. User runs `pulumi import --file <import-file-path>` to create the resource in state.

The use of the import resource option should be explained too for posterity, but has generally been superceded by the two approaches above.

We could also cross-link from the [components docs](https://www.pulumi.com/docs/iac/concepts/resources/components/) to highlight how components can be imported using the "program first" or "bulk import" approach.

See also:
- https://github.com/pulumi/pulumi/issues/19538

joeduffy:
Fixes #15053

## What changed

The import guide previously organized its two CLI-based workflows (single-resource import and bulk import from a JSON file) as if they were parallel-but-separate features, and buried the "generate an import file from a program" workflow — the one issue #15053 asks us to foreground — deep inside a "Bulk Import Operations" section, well after the code-based `import` resource option.

This restructures the page around the two approaches users actually reach for:

1. **CLI-first import** — run `pulumi import <type> <name> <id>` against a single resource, then copy the generated code into your program. Unchanged in substance, just retitled and reordered.
2. **Program-first (bulk) import** — write your program first (optionally using components), run `pulumi preview --import-file <path>` to generate an import file with names, URNs, and types already filled in, fill in the blank `id` fields, then run `pulumi import --file <path>` to import everything in one operation. This is now the lead workflow within the section, with the existing component-import walkthrough as its worked example.

The pre-existing "author the JSON file by hand" workflow and the resource/schema reference table are preserved verbatim, just moved under a new "Authoring an import file by hand" subheading as an alternative to generating the file with `pulumi preview --import-file`.

The `import` resource option keeps its full existing content (including the mismatched-state walkthrough) but is now explicitly framed as a third, older mechanism superseded by the two approaches above, per the issue.

I also found and fixed a broken internal anchor link (`#how-import-works`, which didn't exist) pointing it at the correct `#where-to-find` anchor, and added a short cross-link from the components concept page to the new program-first import section so readers building components know they can bulk-import them.

## Verification

I cross-checked the flag names against the authoritative, auto-generated CLI reference pages (`pulumi_import.md` and `pulumi_preview.md`) rather than inventing anything: `pulumi import --file`/`-f` and `pulumi preview --import-file` both match what's documented there. Front matter parses as valid YAML, Hugo shortcodes (`chooser`/`choosable`, `notes`) remain balanced exactly as in the original, and the file ends with a single trailing newline.

---

### DIFF
--- a/content/docs/iac/concepts/components/_index.md
+++ b/content/docs/iac/concepts/components/_index.md
@@ -361,6 +361,8 @@ This tree makes it clear that a single `awsx:ec2:Vpc` component encapsulates mul
 
 Resource options passed to a component resource do not always behave the same as they do for custom resources. For example, the `provider` option has no effect on a component—use `providers` instead to pass explicit provider configuration to a component's child resources. For a complete list of which options apply to component resources, see [Resource options and component resources](/docs/iac/concepts/resources/options/#resource-options-and-component-resources).
 
+If you have existing cloud infrastructure you'd like to bring under management as a component, see [program-first (bulk) import](/docs/iac/guides/migration/import/#approach-2-program-first-bulk-import), which generates an import file for a component and all of its child resources in one step.
+
 ## Authoring components
 
 You author a component by extending the `ComponentResource` class. The guides below walk through building, packaging, and testing components:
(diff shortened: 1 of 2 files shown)


CamSoper (approved):

---

PR #20105, opened 2026-07-07
Parallelize build-time meta-image rendering with worker threads

jeffmerrick:
### Proposed changes

Extracts the social-card templates and Satori/resvg rendering from `scripts/generate-meta-images.mjs` into `scripts/meta-images/render.mjs`, shared by the generator and a new `worker_threads` entry (`scripts/meta-images/render-worker.mjs`). Cold runs — a fresh checkout or an `OG_TEMPLATE_VERSION` bump, where every card re-renders — now fan out across a worker pool sized to the machine's cores (override with `OG_WORKERS`). Small incremental batches (fewer than 8 cards, the common case) and `OG_WORKERS=1` still render inline, skipping worker startup. The moved template code is byte-identical, so cache keys and rendered output are unchanged; verified both paths locally (0 failures).

n/a

### Related issues (optional)

n/a

### DIFF
--- a/scripts/generate-meta-images.mjs
+++ b/scripts/generate-meta-images.mjs
@@ -31,16 +31,15 @@
 import { readFileSync, writeFileSync, mkdirSync, readdirSync, existsSync, rmSync, statSync } from "fs"
 import { dirname, join, relative } from "path"
 import { createHash } from "crypto"
-import satori from "satori"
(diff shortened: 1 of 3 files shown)


CamSoper (approved):
LGTM

cnunciato (approved):
Claude had a few suggestions that seem reasonable, so pushed a commit to cover those. Nice work!

PR #20126, opened 2026-07-08
Add Floqer (RB2B) visitor identification, consent-gated, on high-intent pages

calon-pulumi:
## What

Adds the **Floqer (RB2B) visitor-identification** tag to 8 high-intent pages, **consent-gated** and **production-only**.

Floqer de-anonymizes individual site visitors, so it's deliberately gated on the site's existing **Marketing & Analytics** consent rather than loading for everyone.

## Pages (via `include_floqer: true` front matter)

- `/pricing`
- `/product`
- `/product/neo`
- `/product/infrastructure-as-code`
- `/product/secrets-management`
- `/docs/iac/comparisons/terraform`
- `/docs/iac/comparisons/terraform/opentofu`
- `/case-studies`

## How it works

- New partial **`layouts/partials/floqer.html`** holds the vendor tag (unchanged: `workers.floqer.com/signals/contact-tracker?sid=…`).
- Wired into **`head.html`** inside the existing `production` block, right after Common Room:
  `{{ if .Params.include_floqer }}{{ partial "floqer.html" . }}{{ end }}`
- **Consent gate:** the tag only loads inside `window.analytics.ready(...)`. That callback fires **only after the consent-managed Segment analytics initializes** — which `theme/src/ts/consent-manager` does only once the visitor's consent is satisfied (EU visitors must opt in via the banner; non-EU load by default). **Fail-closed:** if consent-managed analytics never loads, Floqer never loads.

## Reviewer notes / open questions

1. **Gate granularity.** `analytics.ready` fires whenever the consent-managed analytics loads (i.e. the visitor consented to *any* destination). It is **not** strictly limited to the *Marketing and Analytics* category — a visitor who consents to Advertising only would also get Floqer. If you want it limited to the Marketing & Analytics category specifically, that needs a small hook in `consent-manager` (e.g. emit a category-consent signal Floqer can subscribe to). Happy to follow up if preferred.
2. **Consistency with Common Room.** Common Room (a similar visitor-ID tool) currently loads **ungated** in production. Floqer is intentionally stricter here (consent-gated) per requirement — flagging in case you want to reconcile the two.
3. **Prod-only** — won't appear in preview/dev builds; verify on production (or a prod-env build) on the 8 pages, with consent granted.

Requested by GTM / marketing ops. Vendor `sid` is Floqer's account key.

### DIFF
--- a/content/case-studies/_index.md
+++ b/content/case-studies/_index.md
@@ -1,4 +1,5 @@
 ---
+include_floqer: true
 title: Case Studies
 description: See how our customers are using Pulumi to solve complex, real-world infrastructure challenges at scale.
 meta_desc: |
--- a/content/case-studies/_index.md
+++ b/content/case-studies/_index.md
@@ -1,9 +1,9 @@
 ---
-include_floqer: true
 title: Case Studies
 description: See how our customers are using Pulumi to solve complex, real-world infrastructure challenges at scale.
 meta_desc: |
     See how our customers are using Pulumi to solve complex, real-world infrastructure challenges at scale.
+include_floqer: true
 
 aliases:
     - /case-studies/lykke/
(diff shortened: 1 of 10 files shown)


cnunciato (approved):
LGTM!

PR #20134, opened 2026-07-08
Wrap blog category descriptions instead of truncating

jeffmerrick:
### Proposed changes

Blog category descriptions were being cut off with an ellipsis. On the category term-page header (`layouts/partials/blog/header.html`) the description now wraps to two lines via `line-clamp-2` instead of truncating to one — the two-line cap keeps it inside the fixed-height (`lg:h-26`) row so the layout doesn't shift. On the `/blog/category/` directory grid (`layouts/taxonomy/category.terms.html`) the `line-clamp-3` cap was removed so descriptions wrap fully. Only category pages pass a description into the shared header, so tag/author/series/main listings are unaffected.

### Related issues (optional)

n/a


CamSoper (approved):

---

PR #20140, opened 2026-07-09
Move blog series to a dedicated series taxonomy at /blog/series/<slug>/

jeffmerrick:
The main reason behind this is I believe `/blog/series` + `/blog/series/<slug>` should be better for SEO than having series pages live at `/blog/tag/<slug>`. Plus it's a little tidier!

---

### Proposed changes

Blog series were previously manufactured under the tags taxonomy — a member post needed both a `series:` key and a matching tag, and landing pages rendered at `/blog/tag/<slug>/` through a dispatcher in the tag layout. This registers `series` as its own Hugo taxonomy driven solely by the `series:` front-matter key: term pages render at `/blog/series/<slug>/` and the directory at `/blog/series/`, with head/header/meta-image/indexability partials treating series as a first-class blog term kind and meta-image generation covering the new pages. Every member post drops the redundant series tag (in-body `/blog/tag/<series>/` links updated, `lastmod` stamped where prose changed), and `make lint` now rejects an undefined `series:` value or a series slug used as a tag. S3 redirects map the 15 old `/blog/tag/<slug>/` URLs to their new homes, and `make-s3-redirects.js` learns to skip comment/blank lines so the redirect file can stay documented. AGENTS.md and BLOGGING.md are updated to describe the single-key workflow.

### Related issues (optional)


content/blog/platform-engineering-pillars-1/index.md
```diff
@@ -2,13 +2,13 @@
 title: "Platform Pillars: Build Platforms, Not Infrastructure"
 feature_image: feature.png
 date: 2025-02-26
+lastmod: 2026-07-09
```
cnunciato: I don't think this is right — lemme doublecheck. (I believe the param is last_updated or similar; for some reason Claude always seems to reach for `lastmod`, which is intended as more of a file-level thing I believe.

jeffmerrick: Do you think we should even mark these as edited? (Just a URL update.) Could also revert those URL changes and just let the redirect take care of it?

scripts/redirects/blog-series-redirects.txt
```diff
@@ -0,0 +1,18 @@
+# Blog series moved from the tags taxonomy (/blog/tag/<slug>/) to a
+# dedicated series taxonomy (/blog/series/<slug>/). Slugs come from
+# data/blog_series.yml. See layouts/taxonomy/series.html.
+blog/tag/neo-things/index.html|/blog/series/neo-things/
```
cnunciato, line 4: Good call on this, appreciate it!

content/blog/platform-engineering-pillars-1/index.md
cnunciato: Yeah, it's updated. Commit in progress, will push in a sec!

AGENTS.md
```diff
@@ -110,15 +110,22 @@ The reference pages under `content/docs/iac/concepts/resources/options/` show a
...
 
+### Dates: `updated` vs `lastmod`
+
+When you revise an existing blog post, use **`updated: YYYY-MM-DD`** — not `lastmod`. This is the established convention (the vast majority of revised posts use it) and the one wired to the UI: `layouts/blog/single.html` renders `.Params.updated` as the visible "Updated \<date\>" line beside the publish date. Leave the original `date` unchanged; set `updated` to the revision date. It's the same field documented in `BLOGGING.md`.
+
+**Do not reach for `lastmod`.** It's a Hugo built-in that only feeds the sitemap and schema.org `dateModified`, and the site already sets `enableGitInfo: true` (`config/_default/config.yml`), so Hugo derives `.Lastmod` from the commit date automatically. A hand-stamped `lastmod` is therefore invisible to readers *and* redundant with git. It's easy to default to because `lastmod` is the generic Hugo idiom for "last changed" — but on this site the reader-facing, canonical field is `updated`.
```
cnunciato, line 127: Added this to fix Claude constantly wanting to use `lastmod` (a Hugo built-in) instead of `updated`, which is what influences what appears on the page.

There's still a bug here in that `updated` should flow into the schema.org metadata (so not just be visual), but I'll submit a follow-up PR to fix that.

cnunciato (approved):
I love this — and yes, _way_ better for SEO, given our tag pages are suppressed. Nicely done. 🚀

Issue #20078, opened 2026-07-03
Evaluation: docs review automation (pre-merge review, daily content review, link checker)


CamSoper:
# Docs Review Automation: An Evaluation

*Date: 2026-07-03. Scope: the pre-merge AI docs review pipeline, the daily existing-content review, and the daily link checker — the workflows under `.github/workflows/`, the skills under `.claude/commands/`, and the scripts under `.claude/commands/docs-review/scripts/` and `scripts/content-review/` that implement them.*

This is a point-in-time assessment answering three questions: does the system accomplish what it sets out to do, what could it do better, and what other automation opportunities exist that would meaningfully raise documentation quality without generating slop or inconveniencing the humans in the loop.

---

## 1. What the system is (brief map)

**Pre-merge review** is a five-workflow state machine keyed on `review:*` labels and a re-entrant pinned comment (`<!-- CLAUDE_REVIEW N/M -->`):

- `claude-triage.yml` — deterministic classifier (`triage-classify.py`: domain routing, triviality, frontmatter-only), with a Haiku prose pass only for short-circuited PRs.
- `claude-code-review.yml` — the main review, chained off triage via `workflow_run`. Heavy deterministic pre-computation (Vale, URL prefetch, three-layer claim extraction, parallel claim verification, readthrough, editorial balance, cross-sibling discovery, frontmatter validation, conditional Hugo build) feeds `compose-review.py`, which assembles an ~80%-complete draft. Opus edits the draft; a validate → splice → re-validate → upsert chain (`validate-pinned.py`, `splicer.py`, `pinned-comment.sh`) publishes it.
- `claude-update.yml` / `claude-new.yml` / `claude.yml` — hashtag-routed refresh paths (`#update-review` re-entrant edit, `#new-review` regenerate, bare `@claude` ad-hoc chat), with documented precedence.

**Daily content review** (`review-existing-content.yml` → `content-review-article.yml`): deterministic selection (`select-articles.py`, weighted fair queuing: `score = importance × staleness`, tiers from `strategic-tiers.yaml`, traffic-weighted) fans out one worker per article. Each worker builds a synthetic whole-file diff, reuses the identical docs-review pre-step pipeline, applies **only five enumerated high-confidence fix categories**, opens one draft PR per article, passes a deterministic re-lint gate before draft→ready promotion (which fires the normal pre-merge review), and records the outcome in an S3 ledger. Human approval is still required to merge.

[...]

PR #20151, opened 2026-07-09
Code-enforce the content-review no_retire veto and fix-scope promotion gate

CamSoper:
### Proposed changes

Implements §3.1 of #20078: the daily content review's two instruction-only safety rails are now deterministic workflow gates. (Rebased onto the #20156 review/publish job split; the gates run in the credentialed publish job, before anything is pushed.)

**1. The `no_retire` veto is code-enforced.** New `scripts/content-review/check-retire-veto.py` (reusing `load_tiers`/`tier_for` from `select-articles.py`) answers "may this page be retired?" from `strategic-tiers.yaml` alone. The publish job's **Enforce guardrail gates** step runs it for any retirement verdict — judged against the publish job's own clean checkout of the tiers file plus the trusted `workflow_dispatch` `path` input, never the model-writable queue/verdict files. A veto fails the job before branch creation, so nothing is pushed.

**2. "High-confidence fixes only" is code-enforced.** New `scripts/content-review/verify-fix-scope.py` runs between the publish gate and branch creation: every hunk in the review job's exported patch (handed over as a zero-context `.review-changes.u0.diff`, matched on **old-side** line numbers — the same pre-fix numbering the pre-step artifacts use, so there's no line-drift bookkeeping) must fall entirely within recorded finding ranges:

- `claim` — `.verified-claims.json` verdicts `contradicted`/`mismatch`, ±2 lines
- `vale` — `deterministic_fix: true` entries, ±2 lines
- `readthrough` — `local_repair` findings, ±5 lines, with a move heuristic for reorders
- `frontmatter` — the frontmatter block, only when a violation was actually reported
- `link` — no artifact exists, so ranges come from the verdict's new `applied` entries, validated against the pre-fix blob (declared lines must carry a link construct, range capped at 3 lines)

Coverage is per-line, not per-overlap: an oversized hunk (whole-page rewrite/deletion) that merely intersects one finding fails. Out-of-range edits fail the gate before anything is pushed; the ledger still records the page incomplete (`if: always()`). Stubbed (errored) artifacts contribute no ranges, so fixes in that category fail closed to a human.

**Supporting changes:**

- The verdict sentinel gains a backward-compatible `applied` array (per-fix category/file/pre-fix lines/source) — the audit trail, and the sole range source for link fixes. `load_verdict()` is a bare `json.loads`, so nothing downstream breaks; the ledger record ignores the field for now.
- The review job snapshots the queue, four artifact JSONs, and `base-sha` into the `review-snapshot` run artifact **before** the model step (immutable from then on); the publish job's gates judge against that snapshot, so a model editing the workspace can't widen its own allowance.
- SKILL.md (steps 3/8, retirement section), the worker prompt, and the `strategic-tiers.yaml` header document the new contract.

Both scripts carry embedded `--self-test` suites (19 scope-gate cases, 9 veto cases); `record-review.py --self-test`, `render-gates.py --self-test`, `test_publish_gate.py`, and `test_select_articles.py` all pass, and `make lint` is green.

Deliberate policy choices (flagged in #20078 discussion): the scope gate is strict (shared-source edits from the rendered-content pass fail to a human); the standalone retirement push-guard workflow and ledger `applied` passthrough are deferred follow-ups.

N/A

### Related issues (optional)

Addresses §3.1 of #20078.

https://claude.ai/code/session_01YYxz6AgaHFqeunWsLY6yzA

### DIFF
--- a/.claude/commands/review-existing-content/SKILL.md
+++ b/.claude/commands/review-existing-content/SKILL.md
@@ -171,6 +171,14 @@ almost-made-the-cut record the human reviewer adjudicates. When you flag a
 `reconception`, set `clarity_flag: true` in the verdict sentinel (step 8) so the
 ledger carries the signal even on an otherwise-clean page.
 
+Record every fix you apply as an entry in the verdict sentinel's `applied`
+array (step 8): its category, file, **pre-fix** line range, and a pointer to
+the artifact finding it implements. The publish job deterministically verifies
+that every hunk in your exported changes falls within the line range of a
+recorded finding (`scripts/content-review/verify-fix-scope.py`); an edit
+outside the recorded findings fails that gate and nothing is pushed — so if a
+change doesn't trace to a finding above, don't make it.
+
 Editing guardrails:
 
 - Never rewrite prose beyond the specific correction.
@@ -293,10 +301,16 @@ the canonical ledger record, and uploads it to S3 keyed by slug.
 {
   "verdict": "fixed",
   "reason": "",
-  "fixes": 4,
+  "fixes": 2,
   "skipped_findings": 2,
   "retirement": false,
-  "clarity_flag": true
+  "clarity_flag": true,
+  "applied": [
+    { "category": "claim", "file": "content/docs/iac/concepts/stacks/_index.md",
+      "lines": [42, 43], "source": "verified-claims:c3" },
+    { "category": "link", "file": "content/docs/iac/concepts/stacks/_index.md",
+      "lines": [88, 88], "source": "dead link /docs/intro/concepts/state/" }
+  ]
 }
 ```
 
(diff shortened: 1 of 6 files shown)


CamSoper:
@claude #update-review

Rebased onto master after #20156 (the review/publish job split) merged. The gate integration moved accordingly -- worth re-verifying:

- The snapshot step now uploads a separate `review-snapshot` artifact before the model step (immutable pre-model copies), instead of stashing to $RUNNER_TEMP.
- The export step also emits a zero-context `.review-changes.u0.diff` twin of the patch for the fix-scope gate (context lines must not widen hunk ranges).
- "Enforce guardrail gates" now runs in the credentialed publish job between the publish gate and branch creation, so a veto or scope breach fails before anything is pushed -- the old branch-probing and PR-comment paths are gone (the model can no longer push branches).
- The retire veto reads the publish job's clean-checkout strategic-tiers.yaml + the trusted dispatch path input.
- The re-lint step keeps the split's branch derivation from gate outputs (dropped this PR's origin-probing variant).

CamSoper:
Re-verified the rewired integration (all five points you listed check out against the workflow file) and found one real breakage the static review missed, fixed in db42ffaf:

- **`verify-fix-scope.py` required `--branch`, but the publish job's invocation doesn't pass it** — every fix publish would have died on an argparse usage error (exit 2) at the gate, before pushing. Reproduced locally with the exact workflow invocation. `--branch` is now optional: required only in branch mode (no `--diff-file`), where it's needed to derive the diff. Side effect worth having: since the workflow routes retirement verdicts to the retire veto instead of this gate, a retirement-shaped patch smuggled under `retirement: false` now hits the scope gate and fails wholesale instead of triggering the old retire-branch skip.
- While adding the branch-less test cases, the whole-page-deletion fixture exposed that hunk matching used **interval intersection**, so one oversized hunk overlapping a single finding could ride that finding's allowance. Coverage is now per-line: every old-side line of a hunk must be individually granted.
- Also refreshed both module docstrings for the post-split mechanics (the pinned review's stale `$RUNNER_TEMP` finding) and the PR description.

Self-tests: 19 scope-gate + 9 veto cases pass, plus `test_publish_gate.py`, `record-review.py --self-test`, and `test_select_articles.py`. The exact publish-job invocation (`--diff-file .review-changes.u0.diff`, no `--branch`) now passes/fails correctly on covered/uncovered fixtures.

---

PR #20156, opened 2026-07-09
Split content-review worker into unprivileged review + gated publish jobs

CamSoper:
### Proposed changes

Implements §3.5 of #20078 ("Tighten the content worker's security posture"). The per-article content-review worker ran the review model with production credentials — the pulumi-bot push token persisted in `.git/config` by checkout, AWS env vars exported at job scope before the model step — and an unrestricted `Bash` allowlist, while consuming artifacts derived from fetched external URLs. The pre-merge review deliberately runs its model with no push credentials; this PR removes the asymmetry by splitting the worker into two jobs, rather than re-fighting the per-command allowlist (whose prefix matching rejected 6–21 benign compound commands per run):

- **`review` job — the model runs here, unprivileged.** `contents: read` + `id-token: write` only; checkout with the default token and `persist-credentials: false`; no `environment: production`, no ESC, no AWS. A permanent preflight step fails the job if AWS credentials or a git credential extraheader are detected. The model edits the working tree only — it cannot commit, push, or open a PR — and its changes leave as a patch in a run artifact. The staging pathspec excludes all dot-paths, so a patch can never carry `.github/` or `.claude/` changes. The existing-open-PR skip is now a deterministic pre-check that also saves the pre-compute/review spend on owned pages.
- **`publish` job — deterministic only, credentialed.** Before pushing anything it runs the new `scripts/content-review/publish-gate.py`, which enforces the verdict schema (`fixed` ⟺ non-empty patch), the diff scope (fix PRs: the queued article plus the shared render-time sources the skill's rendered pass may correct; retirements: `content/`, `scripts/redirects/`, and the docs menu data), and the `no_retire` veto from the selection queue — landing the first backstop from #20078 §3.1 as a side effect. The branch name is derived from the queue slug, never chosen by the model, and the scope check runs against the tree actually staged by `git apply`, not against patch text. The existing re-lint → promote → auto-merge and S3 ledger steps move here unchanged; a gate violation records the page `incomplete` so it retries under the existing attempt cap.

Also: deletes the temporary `summarize-denials.py` diagnostic (it existed to size an allowlist the split makes unnecessary), updates `review-existing-content/SKILL.md` for the no-write contract, and documents the privilege model in `BUILD-AND-DEPLOY.md`. `test_publish_gate.py` covers the gate (27 checks, self-contained like the sibling tests). Accepted residual risk, now stated in the workflow comment: the Anthropic API key the model inherently runs on.

Verified: gate unit tests pass alongside the existing `test_select_articles.py`/`test_compose_pr_body.py`; `make lint` passes; the export-pathspec → `git apply --index` handoff was exercised in a scratch repo (in-scope edits ride the patch; dot-path and gitignored files are excluded). The first production dispatch after merge should be watched end-to-end (preflight, artifact handoff, gate, PR-as-pulumi-bot, promotion, ledger).

n/a

### Related issues (optional)

Part of #20078 (§3.5, plus the `no_retire` half of §3.1)

https://claude.ai/code/session_018Xq8dwNYtWHehYAZt222Uv

---
_Generated by [Claude Code](https://claude.ai/code/session_018Xq8dwNYtWHehYAZt222Uv)_


CamSoper:
@claude #update-review

Addressed the 🚨 on the publish job's stale-branch push: b134728b now fetches `content-review/<slug>` immediately before `git push --force-with-lease`, so the lease exists and is fresh regardless of which refs the checkout step fetched.

---

PR #20162, opened 2026-07-09
Document customer-managed agent execution models

### LINKED ISSUE #13218: Document how the different flavors of customer-managed agents work
A customer had a lot of questions about the differences in how the k8s vs DinD customer-managed agents work. We don't have any documentation that specifically explains the internals of how these two configurations work, and the differences between them. Specifically, questions about scaling, the 1:1 relationship between agent and running job, etc. came up - which would be useful to cover.

jkodroff:
Adds an "Execution model" section to the customer-managed runners concept page explaining how the two deploy targets actually run a job, addressing customer questions about scaling and the agent↔job relationship.

## Changes
- New Execution model section: Docker (DinD) and Kubernetes deploy targets, per-job container/Pod lifecycle, one-job-per-runner relationship, and a "when to choose which" comparison table + Mermaid diagram
- Reciprocal cross-links between the concept and setup-guide pages
- Clarified that the runner credentials section is about cloud credentials

Fixes #13218

**Note:** The exact Docker internals (true nested DinD vs socket-mounted sibling containers) aren't documented anywhere I could find, so the Docker subsection stays within verifiable behavior. Worth an engineering confirmation before merge.

### DIFF
--- a/content/docs/deployments/concepts/customer-managed-runners.md
+++ b/content/docs/deployments/concepts/customer-managed-runners.md
@@ -1,7 +1,7 @@
 ---
 title: Customer-managed runners
 title_tag: Customer-managed runners | Pulumi Deployments
-meta_desc: Self-host Pulumi workflow runners on your own infrastructure — supplying credentials and the full configuration reference for customer-managed runner pools.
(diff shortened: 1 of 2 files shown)


lblackstone (approved):
LGTM!

PR #20195, opened 2026-07-10
Close superseded CLI regen PRs so they can't go stale-conflicted

CamSoper:
### Proposed changes

A few times a day, an automated "Regen dev version pulumi@…" PR ends up permanently conflicted and never auto-merges (e.g. #20185, #20094). Root cause: each `pulumi-cli-dev-version` dispatch opens a fresh PR off master rewriting the same one-line `static/latest-dev-version` file, and auto-merge takes ~15 minutes of CI. A dev build arriving inside that window branches without the in-flight PR's change, so the moment the earlier PR merges, the later one goes `dirty` forever — auto-merge stalls silently, later builds merge around it, and the orphan lingers until someone closes it by hand.

Since each new dispatch carries a strictly newer build, a still-open regen PR is superseded by definition. This PR fixes the mechanics at the source in both regen workflows (`pulumi-cli-dev-version.yml` and the release-driven `pulumi-cli-docs.yml`):

1. **Close superseded predecessors once the successor PR exists**: after each run opens its own PR (and arms auto-merge), it closes still-open predecessors with an explanatory comment + branch deletion. Ordering matters: closing only after a successful open means a failed run never leaves us with zero open regen PRs, and a no-change dispatch touches nothing. If a predecessor auto-merges in the brief window before it's closed, this run's PR goes conflicted instead — and the next run's sweep closes it, so nothing lingers. Scoped by title prefix (`Regen dev version` / `Generate Pulumi CLI docs`) since both workflows share the `automation/pulumi-cli-docs` label and must not close each other's PRs, and the run's own PR number is excluded. Best-effort: a failed list/close only leaves a conflicted PR for the next run to clean up.
1. **Serialize dispatches** with a per-workflow `concurrency` group (`cancel-in-progress: false`); a collapsed pending queue keeps the newest run, which carries the newest version anyway.
1. The supersede steps are skipped for `skip_auto_merge` test runs, and `publish_only` backfills never reach the pull-request job, so neither can close a live PR.

Verified: both workflows parse as YAML, and the `jq` title-prefix filters were dry-run against captured `gh pr list`-shaped JSON (each selects exactly its own workflow's PRs). Behavior changes only take effect once this merges to master, since `repository_dispatch` runs use the workflow file on the default branch.

N/A

### Related issues (optional)

Supersedes the manual cleanup of conflicted automation PRs such as #20185.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NEPa2ZyGq8T1Dc7SNvbRFM)_

### DIFF
--- a/.github/workflows/pulumi-cli-dev-version.yml
+++ b/.github/workflows/pulumi-cli-dev-version.yml
@@ -8,6 +8,13 @@ env:
   ESC_ACTION_EXPORT_ENVIRONMENT_VARIABLES: GITHUB_TOKEN=PULUMI_BOT_TOKEN
 
 name: Update latest-dev-version
+
+# Serialize runs: dev builds can dispatch minutes apart, and a queue of pending
+# runs collapses to the newest one, which carries the newest version anyway.
+concurrency:
+  group: pulumi-cli-dev-version
+  cancel-in-progress: false
+
 on:
   repository_dispatch:
     types:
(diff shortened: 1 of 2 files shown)


pgavlin:
Do we want to close predecessors up front? Seems like that could leave us without any open PRs if the workflow fails after closing but before opening. What's the downside of closing after a successful open?

CamSoper:
Good point — reordered in 389610a5 to close after a successful open.

The original up-front ordering was to guarantee the predecessor couldn't auto-merge under the new branch (which is the conflict race this PR fixes). But since every run now sweeps open predecessors, that race is self-healing under the new ordering too: if the predecessor merges in the couple-minute window before it's closed, this run's PR goes conflicted — and the *next* dispatch closes it and opens a fresh one, so nothing lingers. That worst case costs one build cycle, same as the failure mode you flagged, but it's rare (needs dispatches < ~15 min apart *and* the predecessor merging mid-run) rather than triggered by any run failure.

Close-after-open is also never worse than today's behavior in failure cases: a run that dies before opening its PR now touches nothing (previously it would have closed the predecessor and left zero open PRs), and a no-change dispatch no longer closes anything either. The close step excludes the run's own PR number and is skipped for `skip_auto_merge` test runs.

---

CamSoper:
@claude #update-review please.

@pgavlin Sorry for the bot generated reply. Robo-Cam had "Pat's got a good point, would you like me to make this change and reply?" queued up for me before I even read your reply. 😆 

-- Meat Cam

iwahbe (approved):

PR #20200, opened 2026-07-10
Hide docs sidebar toggle on desktop widths

cnunciato:
## What

The left-nav show/hide toggle tab (`.docs-nav-toggle`) is a mobile/tablet control, but it lingered as an inert "box" pinned to the left edge on narrow desktop windows:

image

## Why

The sidebar becomes a persistent 240px desktop sidebar at **≥1024px** (and the JS strips the toggle's collapse state above 1024px, so the toggle does nothing there), but the tab itself wasn't hidden until **≥1297px**. That left a dead zone of **1025–1296px** where both the full sidebar and the useless toggle showed.

## Fix

Hide `.docs-nav-toggle` at ≥1025px — matching where the sidebar goes persistent — and drop the now-redundant hide from the old 1297px block. Mobile/tablet behavior (≤1024px) is unchanged; dark mode is unaffected.

### DIFF
--- a/theme/src/scss/docs/_docs-main.scss
+++ b/theme/src/scss/docs/_docs-main.scss
@@ -231,6 +231,15 @@ body.section-registry {
             }
         }
 
+        // Once the sidebar becomes a persistent desktop sidebar (>1024px, matching
+        // the JS that strips the collapse state classes), the toggle tab is inert —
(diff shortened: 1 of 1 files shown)


jeffmerrick (approved):

CamSoper (approved):

---

PR #20236, opened 2026-07-13
docs: Neo usage limits — org-limit emails only

jkeiser:
We're removing limit notification emails when members cross a personal per-member limit threshold since they were added early (and erroneously)  and they have alternate ways to look at it (and it was causing emails that look like duplicates for small orgs). Org limits will still email the admin when thresholds are reached. Per-member *pausing* behavior is unchanged; this touches only the Email alerts section.


CamSoper (approved):

---

PR #20246, opened 2026-07-14
Restore redirects for 18 /docs URLs still 404ing after the IA restructure

### LINKED ISSUE #20242: Restore redirects for 18 /docs URLs still 404ing after the IA restructure
## Summary

A set of `/docs` URLs still return **404** following the information-architecture restructure. They are receiving steady, real traffic — from both humans and AI crawlers — and each one is a missing redirect rather than a page that should genuinely be gone.

All paths below were **verified live on 2026-07-13** and confirmed still 404 (following redirects).

## Data

Source: CloudFront access logs (`CLICKSTREAM.FCT_FRONTEND_CDN_PAGEVIEWS`), **1 May – 30 Jun 2026**, `HTTP_STATUS = 404`.

- **404s** = all requests to that path over the two months.
- **Agent 404s** = the subset from self-identified AI crawlers (ChatGPT/GPTBot, ClaudeBot, PerplexityBot, Googlebot-AI, Amazonbot, etc.). These matter disproportionately: a 404 teaches the crawler the URL is dead, so it can drop out of assistant answers entirely.

| Path | 404s | Agent 404s |
|---|---:|---:|
| `/docs/deployments/deployments/using/` | 276 | 22 |
| `/docs/get-started/azure/create-project/` | 257 | 23 |
| `/docs/clouds/aws/cloudfx/` | 216 | 11 |
| `/docs/guides/adopting/` | 202 | 16 |
| `/docs/iac/concepts/options/retainondelete/` | 201 | 91 |
| `/docs/intro/console/` | 177 | 7 |
| `/docs/get-started/azure/begin/` | 169 | 22 |
| `/docs/iac/guides/continuous-delivery/add-support-for-cicd-systems/` | 164 | 17 |
| `/docs/reference/tutorials/` | 164 | 4 |
| `/docs/reference/component-tutorial/` | 147 | 6 |
| `/docs/get-started/aws/review-project/` | 143 | 23 |
| `/docs/pulumi-cloud/access-management/oidc/client/github/` | 140 | 26 |
| `/docs/insights/policy/pre-built-packs/` | 132 | 6 |
| `/docs/iac/support/troubleshooting/` | 129 | 31 |
| `/docs/get-started/kubernetes/modify-program/` | 129 | 19 |
| `/docs/iac/crossguard/core-concepts/` | — | 53 |
| `/docs/iac/concepts/stacks/stackreference/` | — | 29 |
| `/docs/iac/cli/commands/pulumi_cloud_api_describe/` | — | 26 |

The last three sit below the reporting cutoff for total volume but are notable for how heavily *crawlers* hit them; counts shown are June-only agent 404s.

## Suggested fix

Add redirects (or restore the pages) for each path above. Most map cleanly onto an existing destination in the current IA — e.g. `/docs/iac/concepts/options/retainondelete/` → the current `retainOnDelete` resource-option page, and the `/docs/get-started/*` paths → their equivalents under the current getting-started flow.

## Notes / scope

- Already fixed, so **not** included: the entire `/docs/iac/guides/building-extending/*` cluster (packages, components, providers) plus `organizing-projects-stacks`, `concepts/plugins`, `resources/options/aliases`, and the `pulumi_stack_init` / `pulumi_plugin_ls` / `pulumi_plugin_rm` CLI pages. These 404'd in May–June but return 200 as of today, presumably resolved by the July provider-docs consolidation.
- Deliberately excluded as noise, not doc pages: legacy Java/.NET API-reference assets under `/docs/reference/pkg/*` (`*-search-index.js/zip`, fonts), `/docs/robots.txt`, `/docs/info/refs` (git clone probes), `/docs/.env` (vulnerability scanning), and malformed inbound links with a trailing `)` from external sites' broken markdown.

Separately, `/docs/robots.txt` alone drew ~688 crawler 404s in June; a redirect or an explicit `410` would quiet that.

jkodroff:
Fixes #20242.

Restores redirects for the 18 `/docs` paths in the issue. 17 are missing `aliases:` entries; one turned out to be a different bug entirely (below). Also adds `redirect_to:` stubs for the renamed `pulumi cloud api *` CLI commands.

## The retainOnDelete one wasn't a missing redirect — it was a casing bug

`/docs/iac/concepts/options/retainondelete/` already **had** an alias. It was spelled in camelCase:

```yaml
aliases:
  - /docs/iac/concepts/options/retainOnDelete/
```

Hugo lowercases page URLs derived from filenames, but writes aliases **verbatim**. So the build emitted the alias only at the camelCase key, while the lowercase URL — the one the world actually links to, and the one Hugo itself would have served back when the page lived there — had nothing behind it. S3 and CloudFront are case-sensitive, so it 404'd.

This survived because **macOS hides it**: APFS is case-insensitive, so the lowercase path resolves fine on a developer's machine and breaks only in production.

It was the *only* uppercase alias in the entire content tree (I swept all of `content/` to confirm), and it had the highest crawler-404 count in the issue by a wide margin — 91, roughly 3x the next path. That fits: crawlers were hitting the lowercase form and getting told the page was dead.

Both spellings are now aliased, so the camelCase URL that exists in production today keeps working rather than becoming a new 404.

## Redirect chain fixed

The cloudfx stub pointed at `/docs/iac/get-started/`, which has no `_index.md` — it is *itself* only an alias. That made it a three-hop client-side chain. Retargeted to `/docs/integrations/clouds/aws/`, a real page and a more useful destination for cloudfx traffic than the get-started hub.

## Scope note

The issue names `pulumi_cloud_api_describe`. The entire `pulumi cloud api` group was renamed to top-level `pulumi api`, so I also stubbed its two siblings (`pulumi_cloud_api`, `pulumi_cloud_api_list`), which 404 identically. These use `redirect_to:` frontmatter rather than `aliases:` because `content/docs/iac/cli/commands/` is regenerated from the CLI and the regen workflow deletes every page that lacks a `redirect_to:` — the same pattern as the existing `pulumi_history.md` stub.

## Judgment calls

Two paths are genuinely dead 2018-era pages with no true successor; destinations chosen with @jkodroff:

- `/docs/reference/tutorials/` (a tutorial hub) → `/tutorials/`
- `/docs/reference/component-tutorial/` (subclassing `ComponentResource`) → the `build-a-component` guide

## Verification

Built the site and resolved each path against the emitted output. The check compares path segments as exact strings against directory listings rather than using a filesystem existence test — on APFS, `test -e` answers the case question *wrong*, which is what let the original bug ship. All 20 paths emit a redirect to the intended destination.

`make lint` clean (1807 files, 0 errors).

## Not addressed: `/docs/robots.txt`

The issue's closing note suggests a `410` would quiet the ~688 crawler 404s. It wouldn't, so I left it alone.

RFC 9309 §2.3.1.4 names 404 and 410 together as "Unavailable" and prescribes *identical* crawler behavior for both. A crawler fetching robots.txt cannot act on the difference, and it re-fetches on its own cache TTL regardless. The 404-vs-410 distinction people have in mind is about *content indexing*, which doesn't apply — robots.txt is a directive fetch, not an indexable page.

The requests themselves are non-compliant: robots.txt is only defined at the root of an origin, so `/docs/robots.txt` is crawlers naively resolving the filename relative to the directory they're in. Suggest filtering it from the 404 reporting query as known noise.

### DIFF
--- a/content/docs/administration/_index.md
+++ b/content/docs/administration/_index.md
@@ -15,6 +15,7 @@ aliases:
   - /docs/administration/
   - /docs/pulumi-cloud/
   - /docs/pulumi-cloud/admin/
+  - /docs/intro/console/
 
 link_buttons:
   primary:
--- a/content/docs/administration/access-identity/oidc-issuers/github.md
+++ b/content/docs/administration/access-identity/oidc-issuers/github.md
@@ -13,6 +13,7 @@ aliases:
 - /docs/administration/access-identity/oidc-client/github/
 - /docs/pulumi-cloud/oidc/client/github/
 - /docs/pulumi-cloud/access-management/oidc-client/github/
+- /docs/pulumi-cloud/access-management/oidc/client/github/
 ---
 
 This document outlines the steps required to configure Pulumi Cloud to accept GitHub id_tokens and exchange them for Pulumi access tokens. Three token types are supported — [organization](#organization-tokens), [team](#team-tokens), and [personal](#personal-tokens) — subject to your [Pulumi edition](/docs/administration/access-identity/oidc-issuers/#token-types-by-edition).
--- a/content/docs/deployments/concepts/_index.md
+++ b/content/docs/deployments/concepts/_index.md
@@ -19,6 +19,7 @@ aliases:
 - /docs/pulumi-cloud/deployments/platform/
 - /docs/pulumi-cloud/deployments/using/
 - /docs/pulumi-cloud/deployments/using-deployments/
+- /docs/deployments/deployments/using/
 - /docs/deployments/deployments/reference/
 - /docs/pulumi-cloud/deployments/reference/
 - /docs/platform/deployments/reference/
--- a/content/docs/iac/cli/commands/pulumi_cloud_api.md
+++ b/content/docs/iac/cli/commands/pulumi_cloud_api.md
@@ -1,0 +1,3 @@
+---
+redirect_to: /docs/iac/cli/commands/pulumi_api/
+---
--- a/content/docs/iac/cli/commands/pulumi_cloud_api_describe.md
+++ b/content/docs/iac/cli/commands/pulumi_cloud_api_describe.md
@@ -1,0 +1,3 @@
+---
+redirect_to: /docs/iac/cli/commands/pulumi_api_describe/
+---
--- a/content/docs/iac/cli/commands/pulumi_cloud_api_list.md
+++ b/content/docs/iac/cli/commands/pulumi_cloud_api_list.md
@@ -1,0 +1,3 @@
+---
+redirect_to: /docs/iac/cli/commands/pulumi_api_list/
+---
--- a/content/docs/iac/concepts/resources/options/retainOnDelete.md
+++ b/content/docs/iac/concepts/resources/options/retainOnDelete.md
@@ -11,6 +11,7 @@ menu:
 aliases:
   - /docs/intro/concepts/resources/options/retainondelete/
   - /docs/concepts/options/retainondelete/
+  - /docs/iac/concepts/options/retainondelete/
   - /docs/iac/concepts/options/retainOnDelete/
 ---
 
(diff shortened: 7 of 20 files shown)


CamSoper (approved):

---

PR #20272, opened 2026-07-14
Slight tweak to footer copy

cnunciato:
Updates the copy to clarify free means for individuals.

### DIFF
--- a/data/footer.yml
+++ b/data/footer.yml
@@ -12,7 +12,7 @@ cta:
 
 brand:
   tagline_html: |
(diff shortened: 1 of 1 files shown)


jeffmerrick (approved):
👍 Looks good, I might tweak the width so it doesn't wrap to another line. But I want to fix some other stuff in the footer when I add it and the header over on /docs so can mess with it then!

jeffmerrick:
Scratch that, looks totally fine!

PR #20320, opened 2026-07-16
Document pulumi stack migrate

sicarul:
## Summary

- document `pulumi stack migrate` as the standard workflow for moving stacks between backends
- replace duplicated manual migration recipes with links to canonical guidance
- add generated CLI reference pages from [pulumi/pulumi#22902](https://github.com/pulumi/pulumi/pull/22902)
- preserve explicit safeguards for local config backups, target overwrites, and service-managed secrets

## Why

The CLI now handles checkpoint transfer, target stack creation, secret re-encryption, and URN rewriting. Docs should direct users to that safer workflow instead of asking them to coordinate export, login, stack creation, and import manually.

## Impact

Users get one canonical migration path with clearer warnings about files and state that can be overwritten. Cross-account GitLab guidance remains available for the case that `pulumi stack migrate` does not support.

🤖 Generated with OpenCode


sicarul:
@claude #update-review

sicarul:
The reported `pulumi stack` style findings are false positives in auto-generated CLI reference files.

- `pulumi_stack.md:67` and `pulumi_stack_migrate.md:79` are generated **SEE ALSO** link text.
- `pulumi_stack_migrate.md:3` is the generated frontmatter title.
- `pulumi_stack_migrate.md:6` is the generated metadata description.

Sibling CLI command pages use the same generated patterns, so no manual changes are needed.

sicarul:
This PR should only be merged after v3.254.0 is released

sicarul:
This is a valid but non-blocking editorial observation. The generated page mirrors the ordering in the `pulumi stack migrate` command's `cobra.Command.Long` description: both worked examples appear before the explanation that omitting `stack-name` opens an interactive selector.

No change is needed in this docs PR because the file is generated. If we want to improve the flow, the fix belongs in `pkg/cmd/pulumi/stack/stack_migrate.go` in `pulumi/pulumi`, followed by regenerating the CLI reference.

CamSoper:
@sicarul merge it whenever you're ready, I've labeled it do-not-merge in the meantime

sicarul:
@CamSoper can you approve please?

content/docs/iac/cli/commands/pulumi_stack.md
jkodroff, line 1: This file is auto-generated by a CLI release, but this is a harmless change and probably not worth correcting, so not gonna block on it.

jkodroff (approved):

PR #20326, opened 2026-07-16
Tint docs code chrome violet to match marketing-web

jeffmerrick:
Let me know your opinions here - I think it looks better with the violet chrome (the gray kept bugging me not matching the code background), but is it _too much_ violet?

Screenshot 2026-07-16 at 3 14 06 PM

---

### Proposed changes

Switches the `/docs` code-adjacent "chrome" — the language-chooser card border and toolbar, the standard (non-language) tab strip, and tab hover underlines — from gray to flat violet brand steps, matching the treatment marketing-web (apps/www) already ships so the two sites look consistent. The tint is centralized in four tunable `--chrome-*` custom properties on `body.section-docs` (violet-100/200/300/400) with a standard dark-mode inversion (violet-900/800/700/300), all tracing to `@pulumi/design-tokens`. Because docs' unlayered `*` border reset would otherwise flatten the violet borders back to gray in dark mode, the chrome borders are re-asserted in the dark block. Text on chrome, the active-tab brand underline, and non-docs choosers (which keep their gray fallback) are unchanged.

### Related issues (optional)

N/A

### DIFF
--- a/theme/src/scss/components/_chooser.scss
+++ b/theme/src/scss/components/_chooser.scss
@@ -1,24 +1,24 @@
 pulumi-chooser {
-    // Standard (non-language) chooser: a gray strip of small tabs matching the
-    // language chooser's toolbar — surface background, rounded top, 1px bottom
-    // border that the active tab's underline sits on. --docs-* tokens (with
-    // non-docs fallbacks); the dark rule in _docs-theme.scss re-asserts the
-    // border colors the unlayered `*` border reset would otherwise override.
+    // Standard (non-language) chooser: a flat violet chrome strip of small tabs
+    // matching the language chooser's toolbar — chrome background, rounded top,
+    // 1px bottom border that the active tab's underline sits on. --chrome-* tokens
+    // (with non-docs gray fallbacks); the dark rule in _docs-theme.scss re-asserts
+    // the border colors the unlayered `*` border reset would otherwise override.
     > ul {
         @apply m-0 mb-4 flex rounded-t-lg;
 
         padding: 0 !important;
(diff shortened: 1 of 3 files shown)


cnunciato (approved):
Yes! The violet looks much nicer, in both modes. Love it. ✨

cnunciato:
@jeffmerrick I will say there's a part of me that wonders whether the light-mode violet code background is a *touch* too dark, though, but that's a separate thing. Been meaning to ask you about it, but keep forgetting. 😄

jeffmerrick:
@cnunciato I don't necessarily disagree, but since `violet-50` is the lightest violet we have, we'd need to get into adjusting the opacity.

Here it is at 75%, but maybe too light when not bordered?

Screenshot 2026-07-16 at 3 58 19 PM

Another option would be to just switch to the page background for these and add a border around the standalone ones, I think this looks pretty clean (and could work with gray instead of violet chrome):

Screenshot 2026-07-16 at 4 01 10 PM

CamSoper (approved):

---

cnunciato:
@jeffmerrick I like it (with the lighter violet), but I see the point about the non-bordered version being a bit less distinct. 🤔 Would bordering those similarly look weird?

jeffmerrick:
@cnunciato I think the border looks good, we could bring in the background at 50% opacity maybe?

Screenshot 2026-07-17 at 11 23 12 AM

cnunciato:
@jeffmerrick Yeah, I think that works! What do you think?

jeffmerrick:
@cnunciato I like it!

PR #20337, opened 2026-07-17
Cover all Pulumi languages in the stack references tutorial

iwahbe:
The [stack references tutorial](https://www.pulumi.com/tutorials/building-with-pulumi/stack-references/) only showed the example in TypeScript and Python. This adds the same program in Go, C#, Java, YAML, and Pulumi HCL, so the tutorial covers every Pulumi language.

Notes:

- The HCL example uses the canonical `pulumi_stack_reference` resource form, and the setup step uses the `hcl` starter template added in https://github.com/pulumi/templates/pull/1141.
- The `langfile` shortcode gains an `hcl` entry (`main.tf`) so the "Add this code to the ... file" sentence renders the right filename. The chooser web component already supports the `hcl` key.
- The Java example uses `StackReference.output(String)`, which is the method on the current SDK ([StackReference.java](https://github.com/pulumi/pulumi-java/blob/main/sdk/java/pulumi/src/main/java/com/pulumi/resources/StackReference.java)); some older docs pages show a `getOutput` method that no longer exists.
- The explanation paragraph was previously duplicated verbatim per language; it is now a single language-neutral paragraph below the chooser.

Verification: the site builds cleanly with the pinned Hugo (0.157.0) and the rendered page contains both choosers with the full language set. The HCL example was run end-to-end against a local backend with two stacks; the consumer stack produced `shopUrl: "http://localhost:3002"` exactly as the tutorial shows. `pulumi new hcl -y` was verified to scaffold a project now that the template PR is merged.

### DIFF
--- a/content/tutorials/building-with-pulumi/stack-references/index.md
+++ b/content/tutorials/building-with-pulumi/stack-references/index.md
@@ -16,7 +16,7 @@ In this section, you'll create a new project and stack that'll use the stack out
 
 Let's start by making a new program in a new directory, alongside `my-first-app`:
 
-{{< chooser language "typescript,python" />}}
+{{< chooser language "typescript,python,go,csharp,java,yaml,hcl" />}}
 
 {{% choosable language typescript %}}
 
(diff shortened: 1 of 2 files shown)


content/tutorials/building-with-pulumi/stack-references/index.md
```diff
@@ -38,6 +38,62 @@ $ pulumi new python -y
...
+{{% /choosable %}}
+
+{{% choosable language hcl %}}
+
+```bash
+$ mkdir ../my-second-app
+$ cd ../my-second-app
+```
+
+There is no `pulumi new` template for Pulumi HCL yet, so create the project file by hand. Add a `Pulumi.yaml` with the following contents:
```
cnunciato: There actually is! My PR just needs to be merged: https://github.com/pulumi/templates/pull/1141 I'll do that this morning.

CamSoper (approved):

cnunciato (approved):

PR #20345, opened 2026-07-17
Make violet code chrome a site-wide default so every chooser gets it

jeffmerrick:
### Proposed changes

The recent violet code-chrome tint (chooser toolbars, tab strips, card borders) was scoped to `body.section-docs`, so choosers rendered outside `/docs` — template catalog pages like `/templates/container-service/azure/`, the Terraform-converter blog post, product pages — fell back to gray. This promotes the light-mode `--chrome-*` tokens to a site-wide default on `body` in `components/_chooser.scss` so every chooser reads violet regardless of section, and leaves the dark-mode inversion in `_docs-theme.scss` (docs is the only section with a dark toggle). The tint now has a single light source and a single dark source, with no duplicated declarations to drift.

### Related issues (optional)

N/A


cnunciato (approved):
🌟

PR #20347, opened 2026-07-17
Change CTAs to Get started / Download

cnunciato:
Updates our CTA buttons on the home page and in the nav:

* Secondary button on the home page changes to "Download open source" and points to https://www.pulumi.com/docs/install (as it did when this button was last labeled this way)
* Primary "Get started" CTA in the site-wide nav changes to point to https://pulumi.com/docs/iac/get-started
* Primary button on the home page remains unchanged, continues to point to https://app.pulumi.com/signup

### DIFF
--- a/config/_default/config.yml
+++ b/config/_default/config.yml
@@ -210,11 +210,10 @@ params:
     cta:
         primary:
             label: "Get started"
-            label_short: "Sign up"
-            href: "https://app.pulumi.com/signup"
+            href: "/docs/iac/get-started"
         secondary:
-            label: "Contact us"
-            href: "/contact/?form=sales"
+            label: "Sign up"
+            href: "https://app.pulumi.com/signup"
 
     canonicalURL: https://www.pulumi.com
     contentRepositoryURL: https://github.com/pulumi/docs
--- a/config/_default/config.yml
+++ b/config/_default/config.yml
@@ -212,8 +212,8 @@ params:
             label: "Get started"
             href: "/docs/iac/get-started"
         secondary:
-            label: "Sign up"
-            href: "https://app.pulumi.com/signup"
+            label: "Contact us"
+            href: "/contact/?form=sales"
 
     canonicalURL: https://www.pulumi.com
     contentRepositoryURL: https://github.com/pulumi/docs
--- a/content/_index.md
+++ b/content/_index.md
@@ -8,6 +8,10 @@ include_organization_schema: true
 sections:
   - type: hero
     layout: split
+    cta_primary_text: Get started
+    cta_primary_link: https://app.pulumi.com/signup
+    cta_secondary_text: Download open source
+    cta_secondary_link: /docs/install/
     badge_highlight_text: "New Release:"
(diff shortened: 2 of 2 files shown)


config/_default/config.yml
```diff
@@ -211,10 +211,10 @@ params:
         primary:
             label: "Get started"
             label_short: "Sign up"
-            href: "https://app.pulumi.com/signup"
+            href: "/docs/iac/get-started"
         secondary:
-            label: "Contact us"
-            href: "/contact/?form=sales"
+            label: "Sign Up"
```
cnunciato:
```suggestion
            label: "Sign up"
```

config/_default/config.yml
```diff
@@ -211,10 +211,10 @@ params:
         primary:
             label: "Get started"
             label_short: "Sign up"
```
sicarul: I don't know what label_short does but doesn't this need to change too?

cnunciato: It's unused — removed it.

config/_default/config.yml
```diff
@@ -210,11 +210,10 @@ params:
     cta:
         primary:
             label: "Get started"
-            label_short: "Sign up"
```
cnunciato, line 213: This is no longer used, so removing.

config/_default/config.yml
```diff
@@ -210,11 +210,10 @@ params:
...
-            href: "https://app.pulumi.com/signup"
+            href: "/docs/iac/get-started"
         secondary:
-            label: "Contact us"
```
cnunciato, line 216: Note that we still have the Contact button site-wide in the man nav: 

image

sicarul (approved):
Lgtm

alexleventer (changes requested):
I don't think this was the solution we landed on, this seems super confusing to me.

sicarul:
> I don't think this was the solution we landed on, this seems super confusing to me. 

What was the solution we came to? It was all a bit vague to me so i assumed Chris understood it better than me

cnunciato:
@alexleventer What was your take-away? I don't disagree that it looks a bit odd (it probably wouldn't be super clear to me how "Sign up" differs from "Get started"), but we did align on the two actions being (1) "Get started" changes to point to the docs path and (2) the other button changes from "Contact us" to point to the console. Would "Try Pulumi Cloud" be better? If not, what would you suggest? (Feel free to suggest a change to the label.)

cnunciato:
Chatted with @alexleventer, and here's what we'll do:

* Primary button on the home page doesn't change; it continues to point to https://app.pulumi.com/signup
* Secondary button on the home page changes to "Download open source" and points to https://www.pulumi.com/docs/install (as it did when this button was last labeled this way)
* Primary "Get started" CTA in the site-wide nav changes to point to https://pulumi.com/docs/iac/get-started

cnunciato:
Done, PR description updated. @alexleventer Please re-review and I'll merge.

alexleventer (approved):

PR #20359, opened 2026-07-18
Fix headless choosable bug when HCL is selected globally

cnunciato:
The language chooser threw a TypeError (`currentOptions[0] is undefined`) whenever its option set was momentarily empty, and selecting a special-purpose language like HCL persisted it as your global preference — so it silently broke get-started pages that don't support it, showing the wrong content while the tabs still read TypeScript. This guards the chooser against the empty case and treats HCL/OPA as page-scoped: they still display on the few pages that offer them, but are never written to the global preference (the serializer keeps your last real language, and stale values are coerced away on load), so they can't leak onto pages that don't offer them. It also ungates the TypeScript examples on the Terraform orchestrate page so they no longer hide for non-TypeScript readers. Note: the chooser and store changes are compiled Stencil code, so they need the asset build to ship.

### DIFF
--- a/content/docs/iac/get-started/aws/create-component.md
+++ b/content/docs/iac/get-started/aws/create-component.md
@@ -22,6 +22,8 @@ complexity and enable sharing and reuse. Instead of copy-pasting common patterns
 You will now create your first component that packages up your S3 website so you can stamp out
 entire websites in a few lines of code:
 
+{{< chooser language "typescript,python,go,csharp,java,yaml" / >}}
+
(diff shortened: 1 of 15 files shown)


jeffmerrick (approved):
Will we need to make similar fixes in /registry?

cnunciato:
@jeffmerrick I think we'll need to do this everywhere we build choosers, yeah. The issue is that it manifests as missing content on pages that don't have the chosen language:

image

PR #20371, opened 2026-07-20
add blog post for automatic logging

tgummerer:
Add a blog post for this new feature.


CamSoper (changes requested):
Leading with the good news: **the claim the pre-merge review couldn't verify is correct.** It flagged the rotation thresholds as unverifiable because pulumi/pulumi was outside its scope, so I went and checked. `pkg/logging/rotation.go`:

```go
const (
	defaultMaxAgeDays = 7
	defaultMaxTotalMB = 500
)
```

Seven days, 500 MB, exactly as written. The rest of the security claims hold up too:

| Claim | Source |
|---|---|
| gzip'd chunks encrypted with AES256-GCM | `pkg/engine/encryptedlog/writer.go` — `aes.NewCipher` + `cipher.NewGCM`; `format.go` has `keySize = 32` commented "the AES-256 key size in bytes"; chunks are independently gzip-compressed |
| Log files are owner-only | `0o600` |
| `pulumi logs decrypt` / `pulumi logs share` | both doc pages exist |

Requesting changes on one sentence, and asking about one more.

## "Encrypted on disk" is unconditional in the post, but not in the code

> These log files are encrypted on disk, using the relevant stack's secret manager, as they still contain secrets at this point.

The package doc comment in `pkg/logging/encrypted.go` says:

> Package logging provides automatic logging to disk, encrypted when a secrets manager is available, **plain gzip otherwise.**

`UpgradeToEncrypted` is documented as a "No-op if already encrypted or `sm` is nil," and if the encrypted writer fails to construct, it deliberately keeps logging unencrypted:

```go
encSink, err := encryptedlog.NewWriterFromKey(l.out, key)
if err != nil {
	l.sink = newGzipSink(l.out)
```

That sentence is doing double duty in the post — it's both the security guarantee and the acknowledgment that these files hold secrets. So a reader can reasonably take it as absolute and conclude `$PULUMI_HOME/logs` needs no further thought, when some files there may be plain gzip with secrets in them.

This isn't a sharing risk: `pulumi logs share` redacts and re-encrypts either way, so the sanctioned path stays safe. It's an at-rest expectation. One sentence naming the fallback would fix it, and I think it makes the post more credible rather than less — "encrypted whenever a secrets manager is available, plain gzip when there isn't one" is a more honest and more useful claim than an unqualified guarantee.

## Please confirm the v3.254.0 attribution

I scanned release notes for v3.252.0 through v3.255.0. `pulumi logs share` is there in v3.254.0, but I found no entry in any of the four for automatic logging, `pulumi logs decrypt`, or per-operation log files.

You wrote the feature so you're the authority here, and changelogs miss things — but the version leads the post and appears in all three social posts, so it's worth being sure.

## Two precision nits

1. **Both thresholds are defaults, not fixed values** — overridable via `PULUMI_LOG_ROTATION_MAX_AGE_DAYS` and `PULUMI_LOG_ROTATION_MAX_TOTAL_MB`. Readers thinking about either disk usage or how long secrets linger would want that.
2. **"after the log directory has reached 500 MB"** reads as though everything is purged at the cap. The implementation deletes oldest-first until the total is back under it (`for i := 0; i < len(remaining) && totalSize > maxTotalBytes; i++`).

Note that the Platform team is also a requested reviewer here — I'm not trying to route around that, just getting the findings recorded since this has been open since July 20.

tgummerer:
@CamSoper I updated the blog post according to the feedback.

content/blog/automatic-logging/index.md
```diff
@@ -0,0 +1,57 @@
...
+    - observability
+    - features
+category: product
+schema_type: auto
+
+# Social media copy — auto-posted to X, LinkedIn, and Bluesky when merged to master.
+# Character limits: X ~280, Bluesky 300, LinkedIn 3000. Leave blank to skip a platform.
+social:
+    twitter: |
+        When Pulumi fails, the fastest fix needs a log — but usually you had to reproduce the failure by hand to capture one. Pulumi v3.254.0 changes that: every operation is now logged automatically, encrypted on disk, and safe to share.
```
kramhuber: Any chance we could use the word error instead of 'fails'? For instance, when Pulumi hits an error...

content/blog/automatic-logging/index.md
```diff
@@ -0,0 +1,57 @@
...
+        Pulumi v3.254.0 fixes both. Every operation is now logged to an encrypted file automatically, and a new command re-encrypts and redacts it so you can share it over any channel, even a public GitHub issue.
+
+        We wrote up how it works.
+    bluesky: |
+        Debugging a Pulumi failure usually meant reproducing it by hand just to capture a log, then finding a safe way to share a file full of secrets. Pulumi v3.254.0 makes logs automatic and encrypted, safe to share even on a public GitHub issue.
+
+        Here's what changed.
+---
+
+Pulumi v3.254.0 introduces automatic logging: every operation is logged in an encrypted log file that can be shared with the Pulumi team for inspection. No more re-running commands just to get logs to the Pulumi team for debugging; instead you can share existing logs securely.
```
kramhuber:
```suggestion
Pulumi v3.254.0 introduces automatic logging: every operation is logged in an encrypted log file that can optionally be shared with the Pulumi team for inspection. No more re-running commands just to get logs to the Pulumi team for debugging; instead you can share existing logs securely.
```

kramhuber (approved):

kramhuber:
Do we expect users to take advantage of the automated logs? Currently the focus is on sharing logs with Pulumi; but, does this also make local debugging simpler?

content/blog/automatic-logging/index.md
tgummerer: Yeah, good idea, that does sound better!

tgummerer:
> but, does this also make local debugging simpler?

I wouldn't say it makes it much easier.  Logs are mostly meant for pulumi developers, rather than pulumi users, though I guess some power users may find it useful to have them in some cases?

tgummerer:
@CamSoper how is this looking to you now?

CamSoper (approved):
Looks good, thank you @tgummerer!

CamSoper:
@claude #update-review

PR #20386, opened 2026-07-20
Document context.pulumi.token properties for ESC

jkodroff:
## What

Documents the `context.pulumi.token` properties in ESC environments, and the two new OIDC subject attributes that came with them.

[pulumi/pulumi-service#45097](https://github.com/pulumi/pulumi-service/pull/45097) exposes the authenticating access token's `type`, `team`, and `name` under `context.pulumi.token`, and adds `pulumi.token.type` / `pulumi.token.team` to the subject attribute allowlist. None of it was documented.

Two pages change:

- **`content/docs/esc/concepts/interpolations-and-references.md`** — the `context` built-in property reference now lists all five properties, notes when each token property resolves to an empty string, and adds a short subsection on differentiating callers by token.
- **`content/docs/esc/guides/configuring-oidc/_index.md`** — the custom token claim attribute list gains the two new attributes, and the adjacent security note explains why `pulumi.token.name` is excluded.

## Why

For team and organization tokens, both `user.login` and `organization.login` resolve to the organization name, so neither distinguishes a team token from an individual user. This came up as a customer question about scoping AWS access per team over OIDC, and the token properties are the answer, but there was nothing to point at.

## Verification

- `make lint` passes.
- Wording was taken from the merged implementation in `cmd/service/services/environments/environments.go` rather than the PR summary, so the empty-value behavior described here matches the code (`token.name` empty for personal/web tokens, `token.team` empty for non-team tokens).
- Confirmed live against a real org: `${context.pulumi.token.type}` resolves to `personal` with a personal token, where it previously failed with `unknown property "token"`.

## Note for reviewers

I verified the personal-token path directly. I did **not** exercise a team-scoped token, so the `token.team` behavior described here comes from reading the implementation rather than from a live test. Worth a second look if you have a team token handy.

### DIFF
--- a/content/docs/esc/concepts/interpolations-and-references.md
+++ b/content/docs/esc/concepts/interpolations-and-references.md
@@ -107,12 +107,32 @@ In addition to referencing properties defined within an environment, references 
 
 ### context
 
-The `context` built-in property provides information about the user evaluating an ESC environment. It exposes the login of the requesting user at `context.pulumi.user.login` and the login of their organization at `context.pulumi.organization.login`.
+The `context` built-in property provides information about the user and the access token evaluating an ESC environment:
+
+* `context.pulumi.user.login`: the login of the requesting user
+* `context.pulumi.organization.login`: the login of their organization
+* `context.pulumi.token.type`: the type of the access token used to open the environment, such as `personal`, `team`, or `organization`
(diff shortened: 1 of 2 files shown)


CamSoper (approved):

---

PR #20433, opened 2026-07-21
docs: fix and complete ESC context object documentation

### LINKED ISSUE #20393: ESC `context` object documentation is stale and incomplete
The hand-written documentation for the ESC built-in `context` object has drifted from the engine in three places. These are independent of the automation work (see the companion issue on generating the built-in properties table from the API) and shouldn't wait on it — they're wrong today.

### 1. `currentEnvironment` / `rootEnvironment` are undocumented

`content/docs/esc/concepts/interpolations-and-references.md` (~line 106) is the canonical "built-in properties" reference. It documents the five `context.pulumi.*` fields but contains **zero occurrences** of `currentEnvironment` or `rootEnvironment`, even though both are injected into every evaluation and are interpolable.

Today the only place they're documented is `content/docs/esc/guides/configuring-oidc/_index.md`, which covers them solely in the narrow context of `subjectAttributes` — so a reader looking up "what can I interpolate?" can't find them at all.

### 2. The OIDC guide is a second, independent copy of the field list

`content/docs/esc/guides/configuring-oidc/_index.md` (~lines 86-95, 110-164) maintains its own hand-written list of context fields. Any new context field currently requires edits in both this and the interpolations page, and there's nothing linking them.

Worth preserving regardless of what happens to the list: this page also carries the **security rationale** for why `pulumi.token.name` is excluded from subject attributes (token names are user-chosen and could be crafted to forge a subject string). That reasoning isn't expressible in the schema and must survive any later automation.

### 3. A blog post documents a shape that never matches today's engine

`content/blog/esc-editor-enhancements/index.md` (~lines 35-43) shows:

```
context.user.login
context.organization.login
```

Both are missing the `pulumi.` infix. The real paths are `context.pulumi.user.login` and `context.pulumi.organization.login`. Either this predates GA or it's a straight error; either way it doesn't work if someone copies it.

### Also worth deciding

Two fields turned up in an audit with no documentation anywhere:

- **`context.pulumi.openDuration`** — produced by the service, undocumented, and with no consumer found outside one internal test. Document it or consider whether it should be public at all.
- **`context.pulumi.user.id`** — used by pulumi/pulumi's own ESC eval fixtures, but the docs only ever mention `user.login`. Worth confirming whether it's actually supported before documenting.

### Note

`pulumi/docs-private` mirrors these files at identical paths and will need the same fixes.

jkodroff:
Fixes #20393

The hand-written docs for the ESC built-in `context` object had drifted from the engine in three places. I verified every claim below against live ESC environments (created, probed, and cleaned up).

## Changes

**1. `currentEnvironment` / `rootEnvironment` are now on the canonical reference**

Both are injected into every evaluation and are interpolable, but `content/docs/esc/concepts/interpolations-and-references.md` had zero occurrences of either. They were only documented in the OIDC guide, under the narrow `subjectAttributes` framing — so a reader asking "what can I interpolate?" could not find them.

Verified the import semantics directly. Opening a parent that imports a child:

```json
{
  "parentView": { "current": "scratch/ctx-parent", "root": "scratch/ctx-parent" },
  "whoAmI":     { "current": "scratch/ctx-child",  "root": "scratch/ctx-parent" }
}
```

Rather than duplicate the OIDC guide's worked example, the new section states the rule and links to it.

I also confirmed the legacy-`default`-project behavior applies to **both** properties, not just `currentEnvironment.name` as the OIDC page implies — an env in `default` resolves both to the bare name with no project prefix.

**2. `context.pulumi.openDuration` is documented**

It is real and interpolable. Confirmed it reflects the lifetime requested per-open rather than a fixed environment property:

| open | resolves to |
|---|---|
| `esc env open <env>` | `2h0m0s` |
| `esc env open <env> --lifetime 30m` | `30m0s` |
| `esc env open <env> --lifetime 90m` | `1h30m0s` |

Since it is user-visible, interpolable, and has a legitimate use (aligning a credential's expiry with the environment that issued it), I documented it rather than proposing removal.

**3. The OIDC guide is cross-linked, not merged**

The issue flags the guide as a second copy of the field list. On reading, it is not quite duplication — it is the *restricted subset* permitted in subject claims, which deliberately excludes `token.name` and `openDuration`. Collapsing the two would lose that distinction. Instead the list now says it is a subset and links to the canonical reference. The security rationale for excluding `pulumi.token.name` is untouched.

**4. Blog post paths fixed**

`content/blog/esc-editor-enhancements/index.md` showed `context.user.login` and `context.organization.login`, both missing the `pulumi.` infix — neither works if copied. Fixed, plus a pointer to the canonical list so the post ages better. Stamped `updated:` per AGENTS.md since the snippet was broken for readers.

## On `context.pulumi.user.id`

The issue asks whether it is supported. **It is not** — the engine rejects it at edit time:

```
Error: updating environment definition: [0]
Diags: unknown property "id"
```

So despite its appearance in pulumi/pulumi's eval fixtures, it is not a valid public path and I left it undocumented.

## Notes

- `make lint` passes.
- All probe environments were deleted; `esc env ls` confirms none remain.
- No org-specific names appear in the diff.
- ⚠️ `pulumi/docs-private` mirrors these three files at identical paths and needs the same fixes — not done here.

### DIFF
--- a/content/blog/esc-editor-enhancements/index.md
+++ b/content/blog/esc-editor-enhancements/index.md
@@ -2,6 +2,7 @@
 title: "Introducing the Latest ESC Editor Enhancements for Improved Authoring"
 allow_long_title: true
 date: 2024-02-22T00:00:00-07:00
+updated: 2026-07-21
 draft: false
 meta_desc: "The new enhancements to Pulumi ESC Editor streamlines the authoring experience of environments for developers"
 authors:
@@ -34,14 +35,16 @@ We have also added support for accessing contextual information from within an E
 
 - `context.rootEnvironment.name`: the name of the root environment being evaluated
 - `context.currentEnvironment.name`: the name of the current environment being evaluated
-- `context.user.login`: the login of the user evaluating the environment
-- `context.organization.login`: the name of the user's organization
+- `context.pulumi.user.login`: the login of the user evaluating the environment
+- `context.pulumi.organization.login`: the name of the user's organization
 
 ```yaml
 values:
   rootEnv: ${context.rootEnvironment.name}
-  currentUser: ${context.user.login}
+  currentUser: ${context.pulumi.user.login}
 ```
+
+For the current, complete list of `context` properties, see [Built-in properties](/docs/esc/concepts/interpolations-and-references/#context).
 
 ## A Seamless Authoring Experience Awaits
 
--- a/content/docs/esc/concepts/interpolations-and-references.md
+++ b/content/docs/esc/concepts/interpolations-and-references.md
@@ -107,7 +107,18 @@ In addition to referencing properties defined within an environment, references 
 
 ### context
 
-The `context` built-in property provides information about the user and the access token evaluating an ESC environment:
+The `context` built-in property provides information about the environment being evaluated, the user requesting it, and the access token they used.
+
+#### Environment identity
+
+* `context.currentEnvironment.name`: the name of the environment in which the reference is written, including its project
+* `context.rootEnvironment.name`: the name of the environment that was opened, including its project
+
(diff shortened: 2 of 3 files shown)


jkodroff:
FWIW, I did read the changes line by line and LGTM.

CamSoper (approved):

---

PR #20437, opened 2026-07-21
Add OpenAI ad pixel to site head

alexleventer:
## What

Adds the OpenAI advertising pixel (`oaiq`) to the site.

## Where

Placed in the `production`-gated third-party script block in `layouts/partials/head.html`, alongside the existing Common Room and Floqer scripts. So it loads only in production, not in local/preview builds.

## Details

- **Deferred to `window.load`** to avoid blocking first paint, matching the Common Room pattern.
- Added a matching **`preconnect`** to `https://bzrcdn.openai.com` (production only), matching the Segment/Common Room preconnect pattern.
- **`debug:false`** — the snippet as supplied had `debug:true`, which would log to the console for every production visitor. Shipped with debug off; flip it if you actually want debug output live.

## Verification

- `npx prettier --check` passes on the changed file.
- Production Hugo build compiles the template cleanly (the two `make lint`/build errors present are pre-existing icon-sprite manifest issues, unrelated to this change).

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -476,11 +476,19 @@
             });
         </script>
 
-        <!-- OpenAI ad pixel (deferred to window.load to avoid blocking first paint) -->
+        <!-- OpenAI ad pixel (deferred to window.load to avoid blocking first paint) — Marketing & Analytics
+             consent-gated: loads only after the consent-managed Segment analytics is ready
+             (window.analytics.ready), same as the Floqer integration below. Fail-closed: if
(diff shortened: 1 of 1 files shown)


CamSoper (approved):

CamSoper:
@claude 

Per the low-confidence item, implement the environment checking as described, as well as wiring this for GDPR compliance. Then #update-review

CamSoper (changes requested):
Hold off on this before you merge it please

CamSoper (approved):
Okay, you're good now

PR #20438, opened 2026-07-21
esc: add environment composition patterns guide

### LINKED ISSUE #20430: ESC: Create a guide for how to size environments
We should encourage users to create more atomic environments so they don’t export secrets unnecessarily. Not sure where this belongs - probably under Guides.

We should provide the following examples:

- The general pattern of controlling secrets on the server-side: creating an environment mostly composed of imports and tagged versions, downstream consumers mostly import a single environment, secret rotations are seamless behind the scenes.
- Another pattern where consumers consume each small environment and rotate stuff on their own. Say when to use this pattern.
- How to break up (refactor) an environment seamlessly to downstream consumers.

Come up with some better names than "how to size environments".

jkodroff:
Adds a guide covering how granular to make an ESC environment, two topologies for composing them, and how to split an oversized environment without breaking downstream consumers. Fills a gap: the docs cover the mechanics of imports, versioning, and rotation, but not how to decide environment granularity.

## Changes
- New guide at `/docs/esc/guides/environment-composition-patterns/` (first under Guides, `weight: 5`)
- Sizing heuristic (owner / lifecycle / audience), then two patterns: centralized composition (platform team narrows output via `merge: false`) and direct consumption
- Output-preserving refactor walkthrough, since resolved output — not source YAML — is the contract with consumers

Closes #20430

### DIFF
--- a/content/docs/esc/guides/environment-composition-patterns.md
+++ b/content/docs/esc/guides/environment-composition-patterns.md
@@ -1,0 +1,170 @@
+---
+title_tag: Environment composition patterns | Pulumi ESC
+title: Environment Composition Patterns
+h1: Environment Composition Patterns
+meta_desc: How to size and compose Pulumi ESC environments — centralized composition, direct consumption, and refactoring an environment without breaking consumers.
+menu:
+  esc:
+    parent: esc-guides
+    identifier: esc-guides-composition-patterns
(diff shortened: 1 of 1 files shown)


CamSoper (approved):

---

PR #20445, opened 2026-07-22
Blog post: "Building Pulumi HCL With Two Sources of Truth"

iwahbe:
### Proposed changes

An engineering blog post about how we test Pulumi HCL's support. The full HCL blog post was getting unwealdy and I thought this section could stand on it's own. All feedback is appreciated.


iwahbe:
@claude #update-review

content/blog/pulumi-hcl-building-against-an-oracle/index.md
```diff
@@ -0,0 +1,187 @@
...
+
+The set of tests is [centrally maintained in pulumi/pulumi](https://github.com/pulumi/pulumi/tree/94536e530d770753b42087931c3e5c0b3c5a51b7/pkg/testing/pulumi-test-language), and then [each](https://github.com/pulumi/pulumi/blob/94536e530d770753b42087931c3e5c0b3c5a51b7/sdk/go/pulumi-language-go/language_test.go) [language](https://github.com/pulumi/pulumi-dotnet/blob/534fc3f5d74051bf644a9abbc1eb4bb3d0659073/pulumi-language-dotnet/language_test.go) [runs](https://github.com/pulumi/pulumi/blob/94536e530d770753b42087931c3e5c0b3c5a51b7/sdk/pcl/cmd/pulumi-language-pcl/language_test.go) [them](https://github.com/pulumi-labs/pulumi-hcl/blob/571fedb720fe7cd7d8f37ce01990c7c2df384658/cmd/pulumi-language-hcl/language_test.go). Because the test suite is language-agnostic, the cost of maintaining the test suite is shared across all languages, and a new Pulumi language can immediately take advantage of conformance tests to bootstrap.
+
+Let's walk through an example of a conformance test case to clarify what we mean. This is the `l2-resource-simple` test case, which verifies a language can define a resource with input properties:
```
cnunciato, line 64: >This is the `l2-resource-simple` test case

Should this be linked to some source code somewhere? It's unclear what "the l2-resource-simple" refers to (beyond of what's quoted below of course).  

If that isn't important, maybe it'd help if we said something like, "This test, which we'll call 'l2-resource-simple', verifies..."?

content/blog/pulumi-hcl-building-against-an-oracle/index.md
```diff
@@ -0,0 +1,187 @@
...
+res = simple.Resource("res", value=True)
+```
+
+Each language plugin knows how to generate code from PCL and can run the code they generate. This is the key to language-agnostic testing.
+
+The [assertion on this test](https://github.com/pulumi/pulumi/blob/94536e530d770753b42087931c3e5c0b3c5a51b7/pkg/testing/pulumi-test-language/tests/l2_resource_simple.go) is:
+
+```go
+func init() {
+	LanguageTests["l2-resource-simple"] = LanguageTest{
```
cnunciato, line 108: Does this just test the PCL, or the generated HCL (or whatever other language)? It sounds like it's meant to test a generated SDK, but the code may tuck that away somewhere. Is the code parameterized to accept a named SDK somehow? If so, could we make it a bit clearer where that happens?

content/blog/pulumi-hcl-building-against-an-oracle/index.md
```diff
@@ -0,0 +1,187 @@
...
+
+Let's walk through an example of a tfcompat test to make this clearer. The real test for the built-in `sum` intrinsic looks like this:
+
+```go
+// TestL1Sum exercises the `sum` built-in: both paths must accumulate with
+// arbitrary-precision numbers so large integers and decimal fractions stay
+// exact rather than being rounded through float64.
+func TestL1Sum(t *testing.T) {
+	t.Parallel()
+	tfcompat.RunCase(t, "l1_sum", tfcompat.Case{})
```
cnunciato, line 155: This left me wondering why we're testing numeric functions in an IaC context, but I think I get it now: the point isn't to show that we can calculate a sum, it's that because HCL has a `sum` function, Pulumi needs to support that function as well. Yeah?

If so, it might be worth mentioning something like "it's not just about the resources", and maybe call out other intrinsics like `count`, `foreach`, etc., just so it's clear why we've chosen this one.

content/blog/pulumi-hcl-building-against-an-oracle/index.md
```diff
@@ -0,0 +1,187 @@
...
+output "decimals" {
+  value = sum([0.1, 0.2])
+}
+
+output "mixed" {
+  value = sum([1, 2, 3, 4])
+}
+```
+
+Each test locks what it observes against OpenTofu's behavior, effectively making *correctness* a built-in property of the test framework. The challenge for tfcompat tests is *comprehensiveness* & *targetedness*. During development, we handle this by pairing each tfcompat bug fix & new feature with at least one tfcompat test. The test suite is not yet comprehensive, but we are getting there. Targetedness is, alas, left purely to human judgment, as it's a form of taste.
```
cnunciato, line 175: >Each test locks what it observes against OpenTofu's behavior, effectively making *correctness* a built-in property of the test framework.

What does this mean, "locks in"? Is something being generated somehow?

>by pairing each tfcompat bug fix & new feature with at least one tfcompat test

A link to an example would be great if we had one!

>as it's a form of taste

How so?

content/blog/pulumi-hcl-building-against-an-oracle/index.md
```diff
@@ -0,0 +1,187 @@
...
+
+#### Building out the tfcompat test corpus
+
+Because tests are correct by construction, we can ask our tireless LLM assistants to find failing tests. This is productive because we don't need to filter out false positives or negatives; the LLM can see if the test is correct or not. Here is my Claude prompt verbatim:
```
cnunciato, line 179: Does this mean the LLM is seeking out gaps in coverage and creating *new* tests, or somehow running or extending existing ones? How does it determine what a "bug" is? By creating random HCL programs and seeing what breaks? (Apologies if these are inane questions — I just don't understand this area of the product and codebase well enough to fill in the blanks myself, and I'm sure readers won't either.)

content/blog/pulumi-hcl-building-against-an-oracle/index.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Conclusion
+
+The key takeaway is that your tests inherently re-encode the desired behavior of your program, and you need to be careful to ensure that the encoding is *comprehensive*, *targeted* & *correct*. If you can encode correctness into the test harness itself instead of manually into each test case, writing tests becomes easier and much less error-prone. Like everything else in software engineering, LLMs improve velocity but not quality. In the worst case, they add useless tests locking in current behavior without concern for correctness. In the best case, you approach [property-based testing](https://hypothesis.works/articles/what-is-property-based-testing/) with an LLM as both [generator](https://www.cs.tufts.edu/~nr/cs257/archive/john-hughes/quick.pdf) & [reducer](https://users.cs.utah.edu/~regehr/papers/pldi12-preprint.pdf). Tests that are *correct* by construction let you unleash LLMs to move towards *comprehensive* coverage.
```
cnunciato, line 187: >The key takeaway is that your tests inherently re-encode the desired behavior of your program

Is this really the take-away we're aiming for, though? The post isn't (or doesn't seem to be) a broad lesson about the value and practice of testing; it's about (I think) the way _we test our new HCL support_. So I'm not sure there _needs_ to be any general take-away here other than something like, "All this testing has made us pretty confident that what we've shipped is about as close as you can get to full parity with Terraform and OpenTofu, and we'd love you to give it a try." (And here's how, file issues here, etc.) [This section of the brand guide](https://brand.pulumi.com/voice/writing-style/#drafting-revision-and-review) might help.

Also:

> Tests that are *correct* by construction

We've said this a few times now ("tests are correct by construction"), but TBH it's unclear what it means. Is it saying something more than just "tests are what we use to define correct behavior"? Maybe I missed some key detail somewhere?

content/blog/pulumi-hcl-building-against-an-oracle/index.md
```diff
@@ -0,0 +1,187 @@
...
+
+## OpenTofu compatible: tfcompat
+
+Unlike the Pulumi conformance test suite, the tfcompat test suite does not need careful human decision-making to figure out the correct behavior for a given OpenTofu program. There is a clear oracle: for all valid OpenTofu programs, `tofu apply` and `pulumi up` should do the same thing. For our purposes, we define "same thing" to mean that the invoked Terraform providers witnessed the same inputs and the program returned the same output. That gives us the definition of a tfcompat test: a Terraform config and a set of providers. We take that Terraform config and use *the same .tf file* to define both a Pulumi project & an OpenTofu workspace, then we run `tofu apply` & `pulumi up` against a set of recorded in-memory Terraform providers. If the providers saw the same thing for both engines, and the stacks output the same result, then our behavior matches & the test passes. If the providers received different inputs, then Pulumi's HCL implementation did not match `tofu`, and the test fails.
```
cnunciato, line 145: >the tfcompat test suite

This feels like it needs an introduction. (The reader won't know what this is, but we're jumping right in in a way that seems to assume they do.)

>There is a clear oracle: for all valid OpenTofu programs, `tofu apply` and `pulumi up` should do the same thing.

Couple things:
* The word "oracle" feels a bit strange to me here. I almost never hear this word used to refer to anything that isn't the database company. Is it idiomatic in this context somehow? Could we just delete "There is a clear oracle:" altogether and not lose anything? (I think we probably can.)
* What does "use *the same .tf file* to define both a Pulumi project & an OpenTofu workspace" actually mean here? I'm having a hard time visualizing this. How do you use a `.tf` file to define both a Pulumi project and an OpenTofu project, when it presumably contains only HCL? Or is this instead saying "the same file is tested in both Pulumi and OpenTofu projects"?

djperlovsky:
This generally looks real solid

content/blog/pulumi-hcl-building-against-an-oracle/index.md
```diff
@@ -0,0 +1,187 @@
+---
+title: "Building Pulumi HCL With Two Sources of Truth"
```
cnunciato, line 2: Maybe a tweak here since the post is more about testing than building?

```suggestion
title: "Testing Pulumi HCL With Two Sources of Truth"
```

(Or alternatively, "Testing Pulumi HCL: The Two Sources of Truth".)

content/blog/pulumi-hcl-building-against-an-oracle/index.md
```diff
@@ -0,0 +1,187 @@
...
...hard part of programming, any more than my penmanship is what's stopping me from writing the next great American novel. The hard part is making sure what you wrote is correct. This is especially true for a project like adding HCL support to Pulumi (in preview now, try it at [pulumi-labs/pulumi-hcl](https://github.com/pulumi-labs/pulumi-hcl)). Users already know exactly what they want their [Terraform modules](https://developer.hashicorp.com/terraform/language/modules) to do, and when they start using those modules in their Pulumi programs the semantics are clear. Terraform modules consumed from Pulumi should do exactly the same thing as Terraform modules consumed from Terraform. At the same time, if Pulumi supports HCL then HCL needs to be fully incorporated into the Pulumi ecosystem, and we need to ensure we can correctly express [Pulumi concepts](https://www.pulumi.com/docs/iac/concepts/) in HCL. The way we do both is exhaustive testing, but the test strategy for *ensuring Pulumi HCL is OpenTofu compatible* is different from the one for *ensuring HCL is Pulumi native*, and I want to talk through each. Since this is a blog post in 2026, this will touch on AI. You have been warned.
```
cnunciato, line 36: Minor, but suggest breaking the block up just a bit:

```suggestion
Typing out code has never been the hard part of programming, any more than my penmanship is what's stopping me from writing the next great American novel. The hard part is making sure what you wrote is correct. This is especially true for a project like adding HCL support to Pulumi (in preview now, try it at [pulumi-labs/pulumi-hcl](https://github.com/pulumi-labs/pulumi-hcl)). 

Users already know exactly what they want their [Terraform modules](https://developer.hashicorp.com/terraform/language/modules) to do, and when they start using those modules in their Pulumi programs, the semantics are clear. Terraform modules consumed from Pulumi should do exactly the same thing as Terraform modules consumed from Terraform or OpenTofu. At the same time, if Pulumi supports HCL, then HCL needs to be fully incorporated into the Pulumi ecosystem, and we need to ensure we can correctly express [Pulumi concepts](/docs/iac/concepts/) in HCL. 

The way we do both is exhaustive testing, but the test strategy for *ensuring Pulumi HCL is OpenTofu compatible* is different from the one for *ensuring HCL is Pulumi native*, and I want to talk through each. Since this is a blog post in 2026, this will touch on AI. You have been warned.
```

FWIW, though, I'd probably suggest swapping the order in which you present these. Here in the intro, it's "compatible" followed by "valid" — but the post's narrative actually starts with "valid" (PCL), then moves to "compatible" (tfcompat). Keeping that parallel structure would help make the piece a little clearer and more cohesive IMO.

content/blog/pulumi-hcl-building-against-an-oracle/index.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Targeted
+
+Targeted means that each test scenario is pointing towards a specific feature. A clear mapping between test cases and what broke is what allows you to see a broken test case and narrow that down to a bug. Consider this counterexample: building the [Linux kernel is a very broad test of a C compiler](https://www.anthropic.com/engineering/building-c-compiler), but if [it fails](https://github.com/anthropics/claudes-c-compiler/issues/1), that doesn't tell you what's broken in the compiler.
+
+### Correct
+
+Correct means that the test asserts the correct behavior. Incidental behavior should not be asserted on, since we don't want to update tests when something changes unless the desired behavior changes. [Snapshot testing](https://en.wikipedia.org/wiki/Characterization_test) has its place, but it's very easy to miss a regression in a 6,000-line snapshot diff. Naturally, the most important thing is that the test doesn't accidentally assert the wrong behavior.
```
cnunciato, line 52: Since these are just definitions, bullets would probably work a bit better:

```suggestion
The key question with putting together any test suite over a complex program is where to source the test corpus from. Writing tests, much like writing code, is easy. Assembling an effective test corpus is much harder. I'd argue that a good test corpus has three properties: it's _comprehensive_, _targeted_, and _correct_:

* **Comprehensive** means that your corpus covers your desired properties. When you change something by accident, a test will fail. If you have ever written a postmortem that ended with "not enough test coverage", this was your problem.

* **Targeted** means that each test scenario is pointing towards a specific feature. A clear mapping between test cases and what broke is what allows you to see a broken test case and narrow that down to a bug. Consider this counterexample: building the [Linux kernel is a very broad test of a C compiler](https://www.anthropic.com/engineering/building-c-compiler), but if [it fails](https://github.com/anthropics/claudes-c-compiler/issues/1), that doesn't tell you what's broken in the compiler.

* **Correct** means that the test asserts the correct behavior. Incidental behavior should not be asserted on, since we don't want to update tests when something changes unless the desired behavior changes. [Snapshot testing](https://en.wikipedia.org/wiki/Characterization_test) has its place, but it's very easy to miss a regression in a 6,000-line snapshot diff. Naturally, the most important thing is that the test doesn't accidentally assert the wrong behavior.
```

I will say that it's _slightly_ confusing to go from the "two things" mentioned in the title and intro (valid, compatible) directly into the _three_ things listed here (comprehensive, targeted, correct), since there's nothing clearly connecting them to what comes before or after them. I'm assuming these three characteristics are common to _both_ types of testing, though, which is fine — we should just sauy that up front, so the reader isn't left wondering how they fit into the bigger picture.

content/blog/pulumi-hcl-building-against-an-oracle/index.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Pulumi language: Conformance testing
```
cnunciato, line 54: I'm not sure the best way to say it, but since the post is about "the two ways we test this stuff", it might help if the headings were enumerated similarly — like "Category 1: Conformance testing" or "Method 1: Ensuring HCL programs are valid", or similar.

It might also help to use consistent language — e.g., in some places we say _valid_, in others _conformant_, etc.). (Little things like this can really help a piece feel clearer and more comprehensible.)

content/blog/pulumi-hcl-building-against-an-oracle/index.md
```diff
@@ -0,0 +1,187 @@
+---
+title: "Building Pulumi HCL With Two Sources of Truth"
+date: 2026-07-30
+draft: false
+meta_desc: "How we test that Pulumi HCL is both a valid Pulumi language and OpenTofu runtime."
```
cnunciato, line 5: Would this be more right?

```suggestion
meta_desc: "How we test that HCL is a valid Pulumi language, and that Pulumi is a compatible HCL runtime."
```

cnunciato (commented):
Nice! Some initial feedback for ya. Overall, it feels like we need to make a call on whether this is about testing in general or how we're testing our new stuff specifically. My suggestion would be the latter, but wonder what you think.

content/blog/pulumi-hcl-building-against-an-oracle/index.md
cnunciato, line 1: We should change the slug to align with the changes we make to the title as well.

content/blog/pulumi-hcl-building-against-an-oracle/index.md
iwahbe: It tests the generated program (in our case HCL) against the generated SDK. I'll try to make that clearer.

content/blog/pulumi-hcl-building-against-an-oracle/index.md
iwahbe: Exactly. I'll call that out more explicitly.

content/blog/pulumi-hcl-building-against-an-oracle/index.md
iwahbe: These are great questions.

> Does this mean the LLM is seeking out gaps in coverage and creating new tests, or somehow running or extending existing ones?

Both. Its not random, but it doesn't actually matter.

> How does it determine what a "bug" is? By creating random HCL programs and seeing what breaks?

Any HCL program whose behavior under `pulumi` diverges against `tofu`.

cnunciato:
Closing this in favor or #20876. (Thanks @iwahbe!)

PR #20449, opened 2026-07-22
docs: add "When to Build a Component" guide

jkodroff:
## Why

From [this thread in #product-questions](https://pulumi.slack.com/archives/C010K803WCA/p1783974772397919): *why should a customer use a proper component resource instead of a function that produces the same result?*

The docs couldn't answer it. The only rationale content was `build-a-component.md`'s "Why Write a Component?" — reuse, best practices, multi-language support — and two of those three are equally true of a plain helper function. Nothing named what registering the grouping with the engine actually buys you.

## What

Keeps two questions separate that were previously tangled:

**A. Component or function?** — a new guide, [When to Build a Component](/docs/iac/guides/building-extending/components/when-to-build-a-component/), at weight 5 so it sits ahead of "Build a Component". Sections:

- A component can be packaged for other languages later
- Resource options are inherited by child resources
- Aliases are inherited when a component is refactored
- The CLI treats the component as one resource
- **When a function is the better choice** — so it reads as guidance, not a pitch

**B. How do I distribute it?** — `packaging-components.md` opened with "you will probably want to package and distribute it," skipping past the local option. Adds "keep it local" as the zeroth option in the decision list, plus a note above the summary table so table and prose agree.

The components *concept* page is organized around consuming components, so it gets no rationale content — only the anchor fix below.

Reasons sourced from the Slack thread (Mitch, James, Levi) plus `data/resource_options.yaml`.

## Drive-by fixes

- `concepts/components/_index.md` — "source-based plugin package" linked to `#authoring-components`, the wrong section. Now points at `packaging-components/#source-based-plugin-packages`.
- `idp/concepts/private-registry.md` — components linked via the stale `/docs/iac/concepts/resources/components/` alias.

## Verification

- `make lint` and Prettier clean. `make lint-prose` reports one finding on the new page, a false-positive Oxford-comma nag on a two-item list.
- Every added link checked against a local build (all 200), including the `#urns` and `#options-inherited-from-a-component-to-its-children` anchors.
- Sidebar order confirmed: When to Build → Build → Testing → Packaging.

### Claims corrected during review

Three things were wrong in my drafts and are worth a reviewer's attention:

1. **`ignoreChanges` is not inherited** by component children (`data/resource_options.yaml:159`). Swapped for `deletedWith`.
2. **`dependsOn` on a component expands to *transitively reachable* resources**, not literally everything inside — the SDK walk stops at custom resources.
3. **`pulumi up --target` is not documented to cascade** from a component to its children. That claim was removed; the text now says only what the CLI help supports (URNs accepted directly, wildcards can select children by URN prefix). **If someone knows the cascade behavior is real and documented elsewhere, this is worth restoring.**

Also note the intro distinguishes Go (struct embedding `pulumi.ResourceState` + a `NewX` function) from the subclassing model used by TypeScript, Python, .NET, and Java.

## Follow-ups not in this PR

- `dependson.md` doesn't mention components at all, so "you can depend on a whole component" is documented only on this new page.
- Engin suggested a part 8 of the [IaC best practices series](https://www.pulumi.com/blog/series/iac-best-practices/) on this topic.

### DIFF
--- a/content/docs/iac/concepts/components/_index.md
+++ b/content/docs/iac/concepts/components/_index.md
@@ -118,7 +118,7 @@ Components distributed as Pulumi packages can be consumed in any language using 
 pulumi package add github.com/my-org/my-component@v1.0.0
 ```
 
-This pattern is common for components your organization publishes for internal consumption via a Git repository or the [Pulumi IDP Private Registry](/docs/idp/concepts/private-registry/). It is also how components from a [source-based plugin package](#authoring-components) are consumed across languages — the SDK is generated in your program's language regardless of the language the component was authored in.
+This pattern is common for components your organization publishes for internal consumption via a Git repository or the [Pulumi IDP Private Registry](/docs/idp/concepts/private-registry/). It is also how components from a [source-based plugin package](/docs/iac/guides/building-extending/components/packaging-components/#source-based-plugin-packages) are consumed across languages — the SDK is generated in your program's language regardless of the language the component was authored in.
 
 Under the hood, Pulumi fetches the package source (e.g. from GitHub), generates a [local SDK](/docs/iac/guides/building-extending/packages/local-sdks/) from the component's schema, and makes the generated SDK available for import in your program.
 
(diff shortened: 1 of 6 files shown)


CamSoper (approved):

---

PR #20474, opened 2026-07-23
Explain how Azure management groups map to Insights accounts

lblackstone:
## Why

Azure platform teams organize subscriptions under management groups and expect to grant one identity access at that scope and have it cover everything underneath. A customer engineer asked whether we have anything to send a customer on exactly this, and we don't — "management group" appears nowhere in the docs. The Connect cloud accounts page describes per-subscription role assignments but never says how the management-group model maps onto Insights, so the first question any Azure customer asks has no documented answer.

## What this adds

A new **Azure management groups** section on `/docs/insights/discovery/connect-cloud-accounts/`, plus a pointer to it from the Azure authentication step. It covers:

- Insights models one account per subscription; there is no management-group-level account and one account can't span subscriptions.
- What *is* shared tenant-wide: one app registration, one federated identity credential, one service principal, and one ESC environment that every subscription's Insights account references.
- The wizard assigns Reader/Contributor at subscription scope and does not detect a role inherited from a management group, so it needs Owner or User Access Administrator on each selected subscription. An existing management-group grant leads to a redundant (harmless) subscription-scoped assignment.
- The alternative for organizations that manage RBAC centrally: grant the role at management-group scope yourself and connect with **Connect using existing ESC credentials**, which takes one ESC environment per subscription and runs into Azure's cap of 20 federated identity credentials per app registration.

Behavior described here was verified against the wizard and `azuresetup` client in pulumi-service; no product change is involved.

## Checks

- `node ./scripts/lint/lint-markdown.js` — 0 errors
- `npx prettier --check` on the edited file — clean
- `make lint-prose` — 0 errors on the file; the 3 remaining warnings are pre-existing lines this PR doesn't touch

### DIFF
--- a/content/docs/insights/discovery/connect-cloud-accounts.md
+++ b/content/docs/insights/discovery/connect-cloud-accounts.md
@@ -67,7 +67,7 @@ Alternatively, you can connect using:
 
 Consent doesn't grant any roles on your subscriptions. The wizard assigns the **Reader** or **Contributor** role on each subscription later, after you select subscriptions and choose an access level.
 
-After you grant consent, the wizard lists the subscriptions in your tenant.
+After you grant consent, the wizard lists the subscriptions in your tenant, including subscriptions you can reach through a role assignment on a management group. If your subscriptions are organized under management groups, see [Azure management groups](#azure-management-groups).
 
 Alternatively, you can connect using:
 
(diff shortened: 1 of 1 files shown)


lblackstone:
@pulumi-eon /review

PR #20484, opened 2026-07-23
docs(rest-api): render named integer enums in REST API schema docs

### LINKED ISSUE #16180: Document teams -> environment permissions
### Problem description

Currently undocumented: one can add/edit team access to ESC environments via REST API endpoint

### Suggestions for a fix

Let's get this documented at some point soon as it can be tricky e.g.

```
PATCH https://api.pulumi.com/api/orgs/{ORG}/teams/{TEAM}

{addEnvironmentPermission: {envName: "{ENVNAME}", permission: "admin|read|open|write"}}
```

jkodroff:
## Problem

Integer-valued enums like `StackPermission` render in the REST API schema docs as bare wire numbers (`0, 101, 102, 103, 104`) with no names or descriptions. A reader hitting `TeamStackPermission.permission` has no way to know that `104` means `Creator`. Reported in #16180.

## Root cause

The metadata is **already in the OpenAPI spec** — it was just never rendered. `make openapi_all` inlines these enums as `integer` (the wire value is the number) and stashes the human-readable names/comments in the `x-pulumi-model-property` vendor extension, positionally aligned with `enum`:

```json
"permission": {
  "enum": [0, 101, 102, 103, 104],
  "type": "integer",
  "x-pulumi-model-property": {
    "enumFieldNames": ["None", "Read", "Write", "Admin", "Creator"],
    "enumFieldComments": ["StackPermissionNone provides no access.", "..."]
  }
}
```

Two Hugo partials rendered `enum` as a bare comma-separated list and ignored the extension.

## Fix (pulumi/docs only)

- **New shared partial** `layouts/partials/openapi/enum-values.html`: when the extension is present and its names line up with the values, render each as `<value> — Name` plus its description; otherwise fall back to the original bare list (string enums, etc.).
- Wire both `schema-component.html` and `property-rows.html` to the shared partial.
- Style `.enum-value-list` with theme-aware `--docs-*` tokens so it flips in dark mode.

This fixes **every** integer enum in the API docs at once (all `StackPermission` sites — `UserPermission`, `OrganizationMetadata.defaultStackPermission`, the `Update*` request bodies, etc.), not just `TeamStackPermission`. No `pulumi-service` / Java IDL change is required.

## Verification

- Full Hugo build clean.
- `TeamStackPermission` schema page now renders `104 — Creator` with its description (schema-component context).
- Stack-tags endpoint renders named enums in the property-row / request-body context too.
- Checked light and dark mode.

Fixes #16180

### DIFF
--- a/assets/fingerprinted/css/openapi.css
+++ b/assets/fingerprinted/css/openapi.css
@@ -170,6 +170,36 @@
     margin: 0.125rem 0;
 }
 
+/* Named integer enums (e.g. StackPermission): each opaque wire value is paired
+   with its name and description, rendered from the x-pulumi-model-property
+   vendor extension. Used in both the schema-component and property contexts. */
+
+.enum-value-list {
+    margin: 0.375rem 0 0;
+}
+
+.enum-value-list dt {
+    margin-top: 0.5rem;
+    font-weight: 600;
+}
+
+.enum-value-list dt:first-child {
+    margin-top: 0.25rem;
+}
+
+.enum-value-list dt code {
+    font-weight: 700;
+}
+
+.enum-value-list dd {
+    margin: 0.125rem 0 0;
+    color: var(--docs-fg-muted);
+}
+
+.enum-value-list dd p:last-child {
+    margin-bottom: 0;
+}
+
 /* Editorial intro prose rendered above the generated endpoint list.
    Sourced from assets/openapi/tag-intros/{slug}.md. */
 
--- a/layouts/partials/openapi/enum-values.html
+++ b/layouts/partials/openapi/enum-values.html
@@ -10,20 +10,35 @@
 
   Input: dict with "schema" — the schema or property object holding `.enum`.
 */}}
+{{/*
+  Descriptions to suppress: a few upstream enum comments are engineer-facing
+  implementation notes, not public API docs (e.g. "…is an implementation quirk…",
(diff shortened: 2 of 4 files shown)


jkodroff:
### Follow-up: guarded two engineer-facing enum comments

Rendering the enum descriptions surfaced that a couple of the upstream comment strings are internal implementation notes rather than public API docs. This partial renders only per-value comments (`enumFieldComments`), not the type-level `enumComments` blob — so the genuinely internal type-level text (StackPermission's legacy/bit-flags/database history, IdentityProvider's "Node backend / Angular" note, the "we clone it in case of a breaking change in the backend-type" asides) is **not** exposed by this PR.

Scoped to what actually renders, only two field comments read as internal, and both are now withheld via a small keyed denylist in `enum-values.html` (the value and its **name** still render; only the description is suppressed):

- `StackPermission.Creator` — "…implementation quirk… `StackPermissionAdminWhoAlsoHappenedToCreateTheStackToo`" (this value is also converted to `Admin` before it's ever returned to an API caller)
- `AppUpdateKind.Rename` — "NOTE: Do not remove this type - it is used by Pulumi Cloud code"

The proper long-term fix is to reword these at the source (the pulumi-service Java IDL, which generates the spec); the denylist entries can be removed once that lands. I'll file that follow-up separately.

jkodroff:
@claude #update-review

CamSoper (approved):

---

PR #20486, opened 2026-07-23
Restore GrowthBook A/B testing integration

alexleventer:
Brings GrowthBook back so we can resume A/B testing. This effectively reverts #18319, which removed it back in April to test PageSpeed impact.

## What's restored

- **`@growthbook/growthbook` dependency** in root + `theme/package.json`; both lockfiles regenerated (pulls `dom-mutator` back too).
- **SDK init + Segment tracking** in `theme/stencil/src/util/util.ts` — initializes against `cdn.growthbook.io`, fires `analytics.track("Experiment Viewed", …)` on exposure, and sets anon/user/UTM attributes.
- **Window keys + preconnect** in `head.html` (prod/non-prod SDK keys, `cdn.growthbook.io` preconnect).

## Things worth knowing before this merges

1. **"Latest" is 1.6.5 — the same version that was removed.** npm's current latest for `@growthbook/growthbook` is `1.6.5`, so there's no newer release to bump to; the init API is unchanged.
2. **This re-incurs the PageSpeed cost #18319 removed.** That PR pulled GrowthBook specifically to test load performance, *after* it had already been deferred and given a preconnect. Re-adding it brings that third-party request chain back. It's restored in its previously-optimized form (deferred init + preconnect), but the tradeoff is real — worth a Lighthouse check before marking ready.
3. **Ad-blocker exposure.** It calls `cdn.growthbook.io` directly. That host isn't on any current blocklist (I checked ~11, including Hagezi/OISD), but given past blocking pain, the durable fix is to **proxy it behind a first-party path** (like Segment's `evs.analytics.pulumi.com`) via a custom `apiHost`. Happy to do that as a follow-up — it's an infra change (subdomain/worker), not code-only.
4. **Verify the SDK keys are still valid** in the GrowthBook account — they're restored as-is from the April code and may have rotated.

## Deliberately NOT restored

- **The two April experiment files** (`cta-activations-direct-vs-docs`, `terraform-compare`): they were orphaned (imported by nothing), stale (Feb-2026 feature keys / old DOM selectors), and unnecessary — `util.ts` is imported by core Stencil components (`root`, `tooltip`, forms), so the GrowthBook init runs regardless. Define new experiments fresh (dashboard or new files).
- **`label-dependabot.yml` / `BUILD-AND-DEPLOY.md` references**: those sections were restructured since April; the original anchors are gone, so re-adding them would resurrect obsolete structure.

## Verification

- `@growthbook/growthbook@1.6.5` resolves in `node_modules`; lockfiles updated.
- Production Hugo build renders the prod SDK key + `cdn.growthbook.io` preconnect in `<head>`.
- Pre-commit (Prettier) passed.
- **CI should confirm the full theme/Stencil JS build** — I restored previously-working code with deps installed, but didn't run the complete bundle build locally.

Left as **draft** pending your review of the PageSpeed tradeoff and whether you want the first-party proxy folded in before this goes live.

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -3,19 +3,27 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=5.0" />
 
-    <!-- Segment keys -->
+    <!-- GrowthBook / Segment keys -->
     <script>
         {{- if eq hugo.Environment "production" }}
             var segmentWriteKey = "UK90Ofwacetj5VCPJ7cUgkbNcKLSHO3u";
             var snippetName = "C8Y429BDEy/06ujkUhzfL.min.js";
+            window.growthbook_sdk_key="sdk-psPofGr6jFV2ja9O";
+            window.growthbook_decrypt_key="9NsAVMNWaDS+Oky3rhLQ+A==";
+            window.growthbook_dev_mode=false;
         {{- else }}
             var segmentWriteKey = "bATvhu6XKpwi7Cs258MGsELkDpVOzvdY";
             var snippetName = "C8Y429BDEy/hpBKpR8Nx1yVCjKPB2nJWg.min.js";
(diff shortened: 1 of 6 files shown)


CamSoper (approved):

---

PR #20487, opened 2026-07-23
refactor(layouts): unify product and event pages on template-page layout

jeffmerrick:
Just some organization and refactoring of stuff that was bugging me. Event landing pages now use the same shared template system as product and parter pages. It's also renamed to `template-page` to make it's actual usage more clear (can be used for content other than product pages.)

To think about for future work: will be be creating more event landing pages in the future? If not, can the two we have be removed and redirected to standard event pages?

---

### Proposed changes

Renames the `product-page` layout to `template-page` and migrates the KubeCon and Google Cloud Next event pages from the bespoke `event-page` layout onto the shared `sections:` array system, deleting the old layout and its orchestrator partial. Generic partials move out of `template-partials/` (code-snippets, mini-cards, external-link-attrs, events/event-card) per the naming convention now documented in `template-page-content.html`, and the case-study TOC and whitepaper sidebar nav consolidate into a shared `section-toc.html` with a renamed `data-section-toc` scrollspy hook. Dead partials (`card.html`, `template-card.html`, `card-group-toggled.html`, `card-button-download.html`) are inlined into their only callers or deleted, with `proserv.html` rewritten to stand alone on shared `.card`/`.btn` styles, and the unused `icon_style: black` option removed. Padding-hack aspect ratios are replaced with Tailwind `aspect-*` utilities, and stale `i` selectors now target the `.ph-icon` sprite icons. AGENTS.md, the event page guides, the figma-to-template-page skill, and the case-studies archetype are updated for the new names; verified with a full local Hugo build, including production-parity checks on `/proserv/`.


CamSoper (approved):

cnunciato (approved):
>will be be creating more event landing pages in the future? If not, can the two we have be removed and redirected to standard event pages?

I doubt it, but let's hold onto these two just because they're our two largest, and having this extra marketing surface for them isn't necessarily hurting us.

PR #20428, opened 2026-07-21
Add cloud-agnostic getting-started overview and connect-cloud page

CamSoper:
### Proposed changes

Out of a UX-crit discussion ([Slack thread](https://pulumi.slack.com/archives/C0A84RUU2SH/p1784653681631779)): the getting-started funnel had no cloud-agnostic end-to-end explanation, and the "configure credentials" step is our biggest dropoff with only cloud-specific pages.

**1. Rebuilt `/docs/get-started/` landing page** (`content/docs/get-started/_index.md`)

- Light "Why Pulumi" value-prop section (3 bullets + link to `/what-is/what-is-pulumi/`)
- New "How getting started works" section: a Mermaid flowchart plus a numbered five-step journey (install → connect your cloud account → write a program → deploy → iterate), each step cloud-agnostic with links into the relevant docs
- Existing four cloud fan-out cards preserved verbatim (same `data-track` values, icons, and markup)
- Added a link to the previously orphaned [Pulumi for Terraform Users](https://www.pulumi.com/docs/iac/get-started/terraform/) flow and links to ESC/Deployments get-starteds under Additional resources
- Frontmatter, menu, and all six aliases unchanged (sharpened `meta_desc` only)

**2. New `/docs/get-started/connect-cloud/` page** (`content/docs/get-started/connect-cloud.md`)

A generic credentials explainer targeting the dropoff step. Leads with the key reassurance — Pulumi has no credential system of its own; it uses your cloud CLI's existing credentials — then:

- How authentication works (official SDK credential chains, per-stack provider config)
- An identity-check command table (`aws sts get-caller-identity`, `az account show`, `gcloud config list`, `kubectl cluster-info`) matching the commands the per-cloud configure pages use
- Card fan-out into each cloud's existing `configure` step, with **new** `data-track` values (`connect-*-configure`) so landing-page analytics stay clean
- ESC OIDC short-lived credentials best-practice pointer, mirroring the notes already on the per-cloud pages

Joins the left nav under Getting Started via frontmatter only — no `layouts/` or `data/` changes, no changes to the per-cloud flows. A follow-up PR will consolidate the near-identical per-cloud flow steps into shortcodes.

Verification: markdown lint passes (0 errors), prettier clean; per-cloud configure targets and stepper untouched.

PR #20530, opened 2026-07-28
Deduplicate get-started flows: shared fragments, os-command helper, destroy-stack proof

CamSoper:
### Proposed changes

First phase of consolidating the near-identical per-cloud get-started flow steps into shared content — the follow-up promised in #20428. The four flows under `content/docs/iac/get-started/{aws,azure,gcp,kubernetes}/` are hand-maintained copies (~29,400 words across 36 files) that have measurably drifted; this PR introduces the sharing machinery and proves it on the highest-identity pages.

**Machinery** (every shortcode ships an `.html` + `.markdown.md` pair):

- **`fragments/get-started/`** — shared tutorial prose as plain markdown, outside `content/` so Hugo doesn't build it as pages. Writers edit normal markdown (with shortcodes), not HTML templates.
- **`gs-include` shortcode** — renders a fragment via `RenderString`, preferring a per-cloud variant (`fragments/get-started/&lt;name&gt;/&lt;cloud&gt;.md`) when one exists. The cloud is derived from the calling page's path by the new `get-started/cloud-slug` partial, so the same call works verbatim on every cloud's copy. A missing fragment is a hard build failure with a descriptive message. The markdown twin emits shortcode-free fragments as **raw markdown** for full fidelity (the `cleanup.markdown.md` convention); shortcode-bearing fragments go through `RenderString` so nested calls resolve to their markdown variants.
- **`os-command` shortcode** — expands one command string into the standard linux/macos (`$`, bash) + windows (`&gt;`, powershell) choosable pair, replacing ~14 lines of boilerplate per command. Optional second argument for a divergent Windows form.

**Conversions:**

- **`destroy-stack`** (all four clouds): prose was already 100% identical in aws/gcp/kubernetes — each page body is now `{{&lt; gs-include "destroy-stack" &gt;}}` + stepper. The per-cloud CLI output listings live in cloud-scoped sub-fragments.
- **`next-steps`** (all four clouds): the verbatim 4× ESC section is one shared fragment. The per-cloud CTA lines stay in-page so their historical `data-track` refs (`gs-aws-esc`, …, `gs-k8-esc`) are preserved byte-for-byte for analytics continuity.

**Drift fixed while single-sourcing** — verified by building the site before/after and diffing the rendered output (both `index.html` and the content-negotiation `index.md`) of every page under `/docs/iac/get-started/`; every hunk maps to exactly one item below:

1. Azure `destroy-stack`'s two gratuitously reworded sentences return to the canonical aws/gcp/kubernetes phrasing.
1. ESC bullet canonicalized to "Developers can access" (aws/gcp had drifted to "can easily access").
1. Stale `/docs/concepts/...` links in aws/azure/kubernetes `next-steps` updated to `/docs/iac/concepts/...` (gcp was already current; `stack` → `stacks`, `state` → `state-and-backends`).
1. ESC CTA path normalized to `/docs/esc/get-started/` (gcp/kubernetes lacked the trailing slash).
1. Kubernetes `destroy-stack`'s console-output fence dropped its stray `bash` language tag (console output, not bash).
1. `pulumi stack rm` chooser values normalized `"macos,linux"` → `"linux,macos"` (attribute order only; the chooser treats values as a set).
1. The five aws pages missing `menu.iac.identifier` now set it like the other clouds (visible in the rendered diff as `data-menu-id` attributes appearing in the shared iac sidebar).
1. `auto-naming-note` links the current `/docs/iac/concepts/resources/names/#autonaming` path.

**Markdown-pipeline improvement:** the docs markdown output previously stripped links whose text is code (`&lt;a&gt;&lt;code&gt;` — e.g. [`pulumi stack rm`](/docs/iac/cli/commands/pulumi_stack_rm)) down to bare text, losing the href. The pipeline now converts them to `` [`text`](href) ``. This also restores lost links on five unrelated pages (`install`, `iac/cli/environment-variables`, `iac/concepts/secrets`, and two building-extending guides) — strict improvements, verified in the before/after diff.

**Verification:**

- Rendered-output diff over all 46 get-started pages × 2 output formats: byte-identical except the enumerated hunks above.
- Chooser semantic markers (`&lt;!-- chooser: os --&gt;` / `&lt;!-- option: ... --&gt;`) confirmed intact in the markdown output of converted pages.
- Intentional-typo test: a bad fragment name fails the build with `gs-include: fragment "destroy-stackk" not found (looked for ..., then ...; called from ...)`.
- `make lint` passes (lint-markdown: 0 errors; prettier clean). The generated-markdown lint gate reports the identical pre-existing error set (22) on baseline and branch builds.

**Follow-ups (phases 2–3, separate PRs):** a `gs` per-cloud token shortcode + `data/get_started.yaml` (deferred from this PR because Hugo can't nest shortcode calls in shortcode arguments, so phase 1 had no consumer for it), a `gs-pulumi-new` language-matrix helper and the `create-project` sandwich conversion, `deploy-stack` normalization, coarse `create-component` fragments, and an `os-command` sweep. `modify-program` and the `configure` bodies stay per-cloud by design (genuinely different content). Migrating inline code samples to CI-tested `static/programs/` is deliberately enabled by this structure but out of scope.

n/a

### Related issues (optional)

Follow-up to #20428 ("A follow-up PR will consolidate the near-identical per-cloud flow steps into shortcodes.")


jkodroff (approved):

CamSoper:
Upon further reflection, @jkodroff and I are shelving this (at least for now).

PR #20565, opened 2026-07-29
Identify signed-in users to Segment

alexleventer:
## What & why

The consent manager loads analytics.js but never told Segment **who** the signed-in user is — `identify()` only carried tracking-consent preferences. This wires up user identification so website sessions stitch to the user's Segment profile.

When a user is signed in (auth cookie present), we now call `analytics.identify(userId, traits)`:

- `userId` and any other fields are read from the `pulumi_web_user_info` cookie.
- Traits **merge** rather than replace, so a sparse cookie won't clobber richer traits (e.g. email) the app already set on that profile — and any fields the app later adds to the cookie flow through automatically.
- Called on normal page loads (`init()`) and when a fresh consent choice is made (`saveAndLoad()`).

## Consent-safe

`identify` is queued on the analytics stub and only flushed once `conditionallyLoadAnalytics()` actually calls `load()`. If consent is declined, nothing is sent.

## Assumptions this relies on

1. The app's `identify` uses the same Pulumi `userId` that lands in the `pulumi_web_user_info` cookie (same identity space → Segment stitches them).
2. Segment merges identify traits, so calling with sparse traits from the website is safe.

## Testing

- `yarn webpack --mode production` compiles cleanly; `consent-manager` bundle rebuilds.
- Prettier passes.

### DIFF
--- a/theme/src/ts/consent-manager/index.ts
+++ b/theme/src/ts/consent-manager/index.ts
@@ -14,11 +14,8 @@ function clearContainer() {
     if (container) container.innerHTML = "";
 }
 
-// Identifies the signed-in user to Segment, sending every field from the auth
-// cookie as traits. Calls queued on the analytics stub are replayed once
-// analytics loads, so this stays gated behind the consent flow: nothing is
-// sent unless conditionallyLoadAnalytics() has (or later does) call load().
-// extraTraits lets callers attach context (e.g. tracking preferences).
+// Identifies the signed-in user to Segment. Queued on the analytics stub, so
+// it stays consent-gated: only flushed once analytics actually loads.
 function identifyUser(extraTraits: Record<string, unknown> = {}) {
     if (!window.analytics) return;
     const user = getUserInfo();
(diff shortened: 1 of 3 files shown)


alexleventer:
@claude review this pls

sicarul (changes requested):
I have two problems, one technical and one methodology wise for this PR:

1) I opened the preview page and was unable to see the identify triggering, not sure if i'm doing something wrong, i do have the pulumi_web_user_info and i copied it from pulumi.com

2) The way this runs for every pageview (If i understood correctly) goes against Segment's best practices, not sure if we could break something depending on what we hook up into Segment, including the Intercom integration, because it'd be updating the intercom record on every pageview: https://www.twilio.com/docs/segment/connections/spec/identify

```
Segment recommends that you make an Identify call:

After a user first registers
After a user logs in
When a user updates their info (for example, they change or add a new address)
```

Alternatively, i could suggest either:

* Adding intercom just the library into app.pulumi.com without adding the widget so we can capture logins
* Setting some cookie to detect if we have already sent the latest user id into intercom so we don't repeat the identify call for each pageview

PR #20572, opened 2026-07-29
Document connecting SAML SSO to an existing account

djgrove:
Users who are already SAML-SSO-managed in one Pulumi organization had no clear path when they needed to access a second SAML organization: signing in directly returns an "Email already in use" screen that they can't clear from that screen. This documents the supported fix, connecting the second organization's SAML SSO identity to the existing account from Account Settings.

## Change Summary

- New "Connect SAML SSO to an existing account" section in the SAML admin overview: a three-step walkthrough plus a troubleshooting note on IdP app assignment and keeping the SAML `NameID` stable.
- Info note under the SCIM FAQ "Email already in use" resolution that points already-SSO-managed users to the connect flow and links the new section.

The guidance stays feature-agnostic and does not name any internal linked-organization mechanism, since that is not a public self-serve feature.

## Related Issues

Relates to pulumi/customer-support#2998

### DIFF
--- a/content/docs/administration/access-identity/saml/_index.md
+++ b/content/docs/administration/access-identity/saml/_index.md
@@ -33,7 +33,7 @@ If you're a member of a SAML-based Pulumi organization, you can sign in to [your
 
 ## Connect SAML SSO to an existing account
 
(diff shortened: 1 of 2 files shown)


smithrobs (approved):
Looks good to me! Thanks for adding this!

djgrove:
@claude #docs-review

PR #20001, opened 2026-07-01
Add Terraform/OpenTofu CLI switcher for code examples on terraform-state-backend page

joeduffy:
Adds a `tf-tool` chooser shortcode to all 12 bash code blocks on the [Store Terraform State in Pulumi Cloud](https://www.pulumi.com/docs/iac/get-started/terraform/terraform-state-backend/) page that invoke the Terraform CLI binary directly.

## Problem

OpenTofu users reading the Terraform-to-Pulumi Cloud state migration guide had to mentally translate every `terraform` command to `tofu`, since the CLI binary is named differently. Copying code blocks would yield `terraform` commands that do not work for OpenTofu users.

## Solution

Uses the same chooser pattern as the existing language and OS switchers (`{{< chooser tf-tool "terraform,opentofu" >}}`). When a user selects **OpenTofu**, all 12 code blocks switch from `terraform` to `tofu`. The selection persists across the page via `localStorage`, matching the behavior of language switchers.

### Commands converted

| Section | Command |
|---|---|
| Migrate from standard backend — step 1 | `state pull` |
| Migrate from standard backend — step 3 | `login` |
| Migrate from standard backend — step 4 | `init -migrate-state` |
| Migrate from standard backend — step 5 | `plan` |
| Migrate from HCP Terraform — step 1 | `state pull`, `state list` |
| Migrate from HCP Terraform — step 4 | `init` |
| Migrate from HCP Terraform — step 5 | `state push`, `state push -force` |
| Migrate from HCP Terraform — step 6 | `plan` |
| Update CI/CD pipelines | `init`, `plan` (GitHub Actions YAML) |
| Troubleshooting | `init -reconfigure` |

### Intentionally unchanged

- HCL `.tf` configuration blocks (`terraform { backend "remote" { ... } }`) — OpenTofu uses identical HCL syntax and the same `terraform {}` block name
- `pulumi stack export/import` commands — these are Pulumi CLI, not Terraform/OpenTofu
- `pulumi whoami` — same
- `rm -rf .terraform` — shell command, not tool-specific (both CLIs use `.terraform/`)

Prose in notes and troubleshooting sections updated to mention both `terraform` and `tofu` commands where inline references appear.

Fixes #18708

---

🧠 *This PR was created by [workprentice](https://github.com/workprenticebot) on behalf of @joeduffy.*

PR #20574, opened 2026-07-29
Make tf-tool a real chooser type, and fail lint on unknown chooser types

CamSoper:
> **tl;dr**: Joe/Workprentice submitted a doc change that used a chooser configuration that didn't exist. Joe missed it, I missed it, my automated tools missed it, and even if I had manually reviewed it from top to bottom... honestly? I'd probably have missed it without tools, too. I wouldn't have looked at the rendered preview, I'd probably have been looking at the diff.
>
> @nyobe noticed it when she went to make some similar documentation that uses `tf`/`tofu` choosers. (Thanks, Claire!)
> 
> This PR does two things: 
> 
> 1. Fixes the bug.
> 2. Builds a guardrail to blow up the build the next time an AI hallucinates a new chooser.

Fixes the Terraform/OpenTofu picker on the [Terraform state backend get-started page](https://www.pulumi.com/docs/iac/get-started/terraform/terraform-state-backend/), and adds a lint guard so this class of failure can't ship silently again.

## The bug

`tf-tool` was never a valid chooser type. The Stencil chooser resolves options through a closed switch in `mapOptions`, so an unrecognized type matched nothing: `currentOptions` stayed empty, the tab list rendered as `<ul></ul>`, and no selection was ever dispatched. Because a `pulumi-choosable` only reveals itself when its value matches the current selection, **the content the chooser wrapped stayed hidden too.**

Measured on production before this change: 12 choosers, all with zero tabs, and all 24 choosables at height 0 — **24 of the page's 29 code blocks were invisible.** "1. Back up your state" read:

> Before making any changes, create a backup of your current state file:

…followed by nothing. Anyone following the guide got prose with no commands.

This was never a regression. The switcher arrived in #20001 and that commit changed one file — the markdown. No component or store wiring ever accompanied it. Confirmed against history: zero commits have ever touched `tf-tool` under `theme/stencil/src` or `layouts/`, none of the 24 historical revisions of `chooser.tsx` contain it, and the only `ChooserType` unions that have ever existed are the three without it. There has never been a Terraform/OpenTofu chooser under any name (no `tofu`/`opentofu` commits either).

## The fix

Wire `tf-tool` through the same path as `pythontoolchain`: `TfToolKey` + `supportedTfTools` on the component, cases in `mapOptions`/`setChoice` and both store subscriptions, and a `tfTool` preferences slice (action type, creator, reducer, persisted-state hydration) so the selection persists across pages like every other chooser. No content changes — the markup on the page was already correct.

## The guard

Nothing mechanical caught this for four weeks, and the same silent failure applies to an unrecognized *option* key (`mapOptions` filters it out, leaving no tabs). `lint-markdown.js` now hard-fails on both.

Valid types and per-type option keys are **parsed out of `chooser.tsx`** — the `ChooserType` union, `mapOptions`' switch, and the `supported*` lists — rather than duplicated into a data file, so the guard can't drift from the component it guards. If that parse ever fails it throws rather than quietly passing, since a silently-empty registry would reproduce the exact failure mode the check exists to prevent.

## Verification

- **The fix, in a browser** (local Hugo, real site bundle): all 12 choosers hydrate with Terraform/OpenTofu tabs; visible code blocks go 5 → 17; clicking OpenTofu on one chooser switches all 12 and the command becomes `tofu state pull …`; `tfTool: "opentofu"` persists to localStorage and survives a reload.
- **Dark mode**: tab contrast 16.98/5.57 (light) and 16.26/10.41 (dark) — both well past WCAG AA.
- **The guard catches the original bug**: with `tf-tool` removed from the union again, it flags all 36 call sites on `terraform-state-backend.md`. PR #20001 would have failed lint.
- **No false positives**: clean across all content — 4,000+ `choosable` and 671 `chooser` call sites. Negative tests confirm unknown types and unknown option keys are both caught, with correct line numbers, for all four shortcode delimiter forms.
- `make lint` clean; prettier clean; `stencil build` typechecks.

## Notes for the reviewer

- The guard covers `content/`, not chooser markup emitted from `layouts/` templates.

- **Third commit is an unrelated one-character bug** in the same function, pulled in at Cam's request. `normalizeState` hydrated every preference from `persistedState.preferences` except `backend`, which it read from `persistedState.backend` — one level too high. That key never exists, so the `|| "service"` fallback always won and a saved backend choice silently reset on every page load. Verified as an A/B against production: seed `pulumi_state` with `backend: "self-managed"`, reload, and production comes back `"service"` while this branch comes back `"self-managed"`. (Seed and reload have to happen in the same tick — the store re-persists on a 3s interval, so a slower seed gets clobbered before the reload. My first attempt at this test hit exactly that and produced a false negative.)
- **The Stencil unit tests do not run**, on this branch or on master: the repo declares jest 30 while Stencil 4.43.2 supports only jest 29, so `stencil test` refuses to start. Two consequences worth knowing when reading the diff:
  - `store.spec.ts` is updated for the new `tfTool` slice but **unverified by execution**. The browser testing above covers the same ground end-to-end.
  - That spec's `toStrictEqual` was already asserting a preferences slice with no `backend` key, which `normalizeState` has always emitted — so it could never have passed. I added the missing key. Fixing the jest/Stencil mismatch is out of scope here but would have caught both this and, plausibly, the original `tf-tool` bug.

### DIFF
--- a/scripts/lint/lint-markdown.js
+++ b/scripts/lint/lint-markdown.js
@@ -906,6 +906,22 @@ const parseChooserRegistry = (function () {
             }
         });
 
+        // Fail loudly here for the same reason the union parse does above: an
+        // unresolved type leaves the option-key check with nothing to compare
+        // against, so it quietly passes everything. A regex that stops matching
+        // (a reformat moving the closing `];`, a renamed supported* list, a
+        // restructured mapOptions switch) would otherwise disable half the guard.
+        types.forEach(function (type) {
+            if (!optionsByType[type]) {
+                const listName = typeToList[type];
+                const where = listName ? `(list: ${listName})` : "(no matching case in mapOptions)";
+                throw new Error(
+                    `Could not parse option keys for chooser type '${type}' ${where} from ${CHOOSER_COMPONENT_PATH}. ` +
+                        `If the component was refactored, update parseChooserRegistry in this file to match.`,
+                );
+            }
+        });
+
         cached = { types, optionsByType };
         return cached;
     };
--- a/scripts/lint/lint-markdown.js
+++ b/scripts/lint/lint-markdown.js
@@ -840,6 +840,208 @@ function checkChangelogAssets() {
 }
 
 /**
+ * The Stencil chooser component, which is the single source of truth for which
+ * chooser types exist and which option keys each one accepts. The lint check
+ * below parses this file rather than duplicating its lists into a data file, so
+ * the guard can't drift out of sync with the component it's guarding.
+ */
+const CHOOSER_COMPONENT_PATH = path.resolve(__dirname, "../../theme/stencil/src/components/chooser/chooser.tsx");
+
+/**
+ * Option tokens that both chooser shortcodes rewrite before handing them to the
+ * component (layouts/shortcodes/{chooser,choosable}.html), and which therefore
+ * won't appear as keys in chooser.tsx.
+ */
+const CHOOSER_OPTION_ALIASES = ["nodejs"];
+
+/** Matches a chooser/choosable shortcode call, capturing its raw argument list. */
+const CHOOSER_SHORTCODE_REGEX = /\{\{[<%]\s*(chooser|choosable)\s+([^\n%>}]*?)\s*\/?\s*[%>]\}\}/g;
+
+/**
+ * Parses chooser.tsx into { types, optionsByType }: the set of valid chooser
+ * types (from the ChooserType union) and, for each type whose options we can
+ * resolve, the set of valid option keys (by following mapOptions' switch to the
+ * corresponding supported* list).
+ *
+ * Throws if the file can't be parsed into a non-empty type list. Failing loudly
+ * is deliberate: a silently-empty registry would turn this guard into a no-op,
+ * which is the exact failure mode it exists to prevent.
+ *
+ * @returns {{types: string[], optionsByType: Object<string, string[]>}}
+ */
+const parseChooserRegistry = (function () {
+    let cached;
+
+    return function () {
+        if (cached) {
+            return cached;
+        }
+
+        const src = fs.readFileSync(CHOOSER_COMPONENT_PATH, "utf8");
+
+        // export type ChooserType = "language" | "os" | ...;
+        const union = src.match(/export type ChooserType\s*=\s*([^;]+);/);
+        const types = union ? [...union[1].matchAll(/"([^"]+)"/g)].map(m => m[1]) : [];
+        if (types.length === 0) {
+            throw new Error(`Could not parse the ChooserType union from ${CHOOSER_COMPONENT_PATH}. ` + `If the component was refactored, update parseChooserRegistry in this file to match.`);
+        }
+
+        // case "language": options = this.supportedLanguages;
+        const typeToList = {};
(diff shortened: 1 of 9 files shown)


CamSoper:
@claude #update-review

nyobe (approved):
Thanks for the fix!

scripts/lint/lint-markdown.js
```diff
@@ -839,6 +839,208 @@ function checkChangelogAssets() {
...
+        const optionsByType = {};
+        types.forEach(function (type) {
+            const keys = listKeys[typeToList[type]];
+            if (keys && keys.length > 0) {
+                optionsByType[type] = keys;
+            }
+        });
+
+        cached = { types, optionsByType };
+        return cached;
```
nyobe, line 926: The type-union parse throws on failure, but this path doesn't — if the `supported*` regex stops matching (e.g. a reformat changes the indentation of `];`), `optionsByType` silently becomes `{}` and every option-key check hits `if (!valid) continue`. Half the guard becomes a no-op.

Suggest adding a throw here before caching:

```suggestion
        for (const [type, listName] of Object.entries(typeToList)) {
            if (!optionsByType[type]) {
                throw new Error(
                    `Could not parse option keys for chooser type '${type}' (list: ${listName}) from ${CHOOSER_COMPONENT_PATH}. ` +
                    `If the component was refactored, update parseChooserRegistry in this file to match.`
                );
            }
        }

        cached = { types, optionsByType };
```

— Claude, on behalf of @nyobe

scripts/lint/lint-markdown.js
```diff
@@ -839,6 +839,208 @@ function checkChangelogAssets() {
     return errors;
 }
 
+/**
+ * The Stencil chooser component, which is the single source of truth for which
+ * chooser types exist and which option keys each one accepts. The lint check
+ * below parses this file rather than duplicating its lists into a data file, so
+ * the guard can't drift out of sync with the component it's guarding.
+ */
```
nyobe, line 847: The reasoning is sound, but it trades drift risk for parse-fragility risk — the indentation-dependent regexes below are a concrete example. A plain `.js` data file exporting the type list and per-type option keys would be importable by both sides (`require()` from the linter, `import` from the component) and would let TypeScript derive `ChooserType` and the key unions via `as const` + indexed access. Not blocking — the throw guard covers the gap for now.

— Claude, on behalf of @nyobe

scripts/lint/lint-markdown.js
CamSoper: Good catch — implemented in #20589, which targets this branch.

One widening from the suggested block: it loops over the parsed `mapOptions` cases, so it catches a broken `supported*` regex but not a broken *case* regex — if that one stops matching, `typeToList` is empty and the loop iterates nothing, leaving the same silent no-op. The check now loops over the `ChooserType` union instead, so every declared type has to resolve to a non-empty key list. The message names the list when one was found, or says `(no matching case in mapOptions)` when the case regex was the thing that broke.

Verified both failure modes: reindenting a `supported*` list's closing `];` throws for `pythontoolchain`, and wrapping the `options = this.supportedTfTools;` assignment across lines throws for `tf-tool`. Full lint still clean on 1,832 files.

---

PR #20581, opened 2026-07-30
Add /extend-trial request page

calon-pulumi:
## What

Adds a dedicated **`/extend-trial`** page — the destination for the "Request an extension" link in the Pulumi Cloud console trial banner. A user near the end of their trial lands here, tells us what they're still evaluating, and submits a request that routes to the growth/CS team for a human to approve.

Duplicated the contact page as a starting point (thanks @<web-teammate>), then swapped the form for a HubSpot embed so we can prefill account context from the console link and show an in-place confirmation.

### Files
- `content/extend-trial/_index.md` — content + front matter
- `layouts/page/extend-trial.html` — layout (hero, form card, confirmation state, right rail)

## How it works
- **Header/footer** come from `baseof.html` via `{{ define "main" }}` — real site chrome, not approximations.
- **`noindex`** via `block_external_search_index: true` (same convention as `content/confirmation/`, `content/gads/`, event pages). This is a transactional page reached from the product; it shouldn't compete in search.
- **HubSpot form** "Trial Extension" (portal `4429525`, form `4b3db916-7d44-4178-9f86-1cdf94567bfa`), embedded with raw `hbspt.forms.create` — deliberately *not* the `<pulumi-contact-us-form>` component, because we need `onFormReady` prefill and the in-place confirmation swap.
- **Prefill from the console link** (visible, editable fields): `?email=` → `email`, `?org=` → `pulumi_organization_name_s_`. No params → fields are simply empty for the user to fill; nothing breaks. UTM capture is handled by the form's own hidden `last_utm_*` fields.
- **Right rail is dynamic** — pulls the next 3 upcoming events (same `sortable_date` + `unlisted` logic as `partials/events/list-section.html`) and the 4 most recent blog posts from the site's own content. Self-updating; nothing hardcoded.
- CSS is scoped with an `et-` prefix (no global changes) and also styles the HubSpot form fields at the page level — **set the form to "Unstyled" in HubSpot** so it doesn't inject competing CSS.

## Reviewer notes / follow-ups (not blockers)
1. **Inline `<style>`/`<script>` in the layout** — kept self-contained for a starting point. Happy to move CSS to the theme's pipeline and the embed script wherever you prefer.
2. **Console change is a separate dependency** (different repo): the trial-banner link needs to append `?org=<slug>&email=<url-encoded>&utm_source=console&utm_medium=trial_banner`. Without it the page still works; prefill is what makes it one-click.
3. **CI "test" failure is a fork artifact** — it fails at the "Fetch secrets from ESC" step because fork PRs are denied repo secrets. Running this branch in-repo (which also produces the deploy preview) will authenticate and run the real tests.

Opening as a **draft** for the web team to sanity-check conventions before it goes further.

### DIFF
--- a/content/extend-trial/_index.md
+++ b/content/extend-trial/_index.md
@@ -1,0 +1,8 @@
+---
+title: Extend your trial
+meta_desc: Request more time on your Pulumi trial. Tell us what you're still evaluating and we'll extend it. Most requests are approved within one business day.
+type: page
+layout: extend-trial
+# Transactional page reached from the product console trial banner. Keep it out of search.
+block_external_search_index: true
+---
--- a/layouts/page/extend-trial.html
+++ b/layouts/page/extend-trial.html
@@ -1,0 +1,231 @@
+{{ define "main" }}
+{{/*
+  /extend-trial — destination for the Pulumi Cloud console trial-banner
+  "Request an extension" link. Duplicated from the contact page as a starting
+  point (per web team), then swapped to a HubSpot embed so we can prefill
+  account context from the console link. Header/footer come from baseof.html.
+
+  HubSpot form: "Trial Extension" (portal 4429525, form 4b3db916-...).
+  Prefill (visible, editable fields):
+    ?email=  -> email
+    ?org=    -> pulumi_organization_name_s_  (the "Pulumi Organization Name(s)" prop)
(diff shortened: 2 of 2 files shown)


calon-pulumi:
Superseded by https://github.com/pulumi/docs/pull/20583 — same commits, moved in-repo (from a fork) so CI and the deploy preview run. Closing this one.

PR #20583, opened 2026-07-30
Add /extend-trial request page

calon-pulumi:
## What

Adds a dedicated **`/extend-trial`** page — the destination for the "Request an extension" link in the Pulumi Cloud console trial banner. A user near the end of their trial lands here, tells us what they're still evaluating, and submits a request that routes to the growth/CS team for a human to approve.

Duplicated the contact page as a starting point, then swapped the form for a HubSpot embed so we can prefill account context from the console link and show an in-place confirmation.

### Files
- `content/extend-trial/_index.md` — content + front matter
- `layouts/page/extend-trial.html` — layout (hero, form card, confirmation state, right rail)

## How it works
- **Header/footer** come from `baseof.html` via `{{ define "main" }}` — real site chrome.
- **`noindex`** via `block_external_search_index: true` (same convention as `content/confirmation/`, `content/gads/`). Also excluded from `sitemap.xml` by that flag, and nothing links to it — reached only from the console.
- **HubSpot form** "Trial Extension" (portal `4429525`, form `4b3db916-…`), embedded with raw `hbspt.forms.create` — deliberately *not* the `<pulumi-contact-us-form>` component, because we need `onFormReady` prefill and the in-place confirmation swap.
- **Prefill from the console link** (visible, editable fields): `?email=` → `email`, `?org=` → `pulumi_organization_name_s_`. No params → fields start empty; nothing breaks. UTM capture is handled by the form's own hidden `last_utm_*` fields.
- **Form styling is injected into the form's iframe** in `onFormReady`. HubSpot renders this form in a same-origin `srcdoc` iframe, so page-level CSS can't reach the fields; the field styling (labels, inputs, single-line org box, submit button) is applied inside the iframe document. Page-level CSS only covers the surrounding page (hero/cards/rail/confirmation).
- **Right rail is dynamic** — next 3 upcoming events (same `sortable_date`/`unlisted` logic as `partials/events/list-section.html`) and 4 most recent blog posts, pulled from the site's own content.

## Reviewer notes
1. **Inline `<style>`/`<script>` in the layout** — kept self-contained. Happy to move CSS into the theme pipeline / relocate the embed script if you'd prefer.
2. **The iframe styling relies on the form being same-origin** (it is today). If HubSpot ever changes the embed to a cross-origin iframe, the injection no-ops gracefully (wrapped in try/catch) and the form falls back to HubSpot's own theme.
3. **Console dependency (separate repo/team):** the trial-banner link needs `?org=<slug>&email=<url-encoded>&utm_source=console&utm_medium=trial_banner`. The page works without it; prefill is what makes it one-click.

CI is green (build + tests). Verified in the deploy preview.

### DIFF
--- a/content/extend-trial/_index.md
+++ b/content/extend-trial/_index.md
@@ -5,4 +5,43 @@ type: page
 layout: extend-trial
 # Transactional page reached from the product console trial banner. Keep it out of search.
 block_external_search_index: true
+
+overview:
+    eyebrow: Extend your trial
+    title: Need more time with Pulumi?
+    description: Tell us what you're still evaluating and we'll extend your trial. Most requests are approved within one business day, and you won't lose your stacks, state, or history.
+
+form:
+    hubspot_form_id: 4b3db916-7d44-4178-9f86-1cdf94567bfa
+    footnote: "Goes straight to the Pulumi team. Typical response: under 1 business day."
+
+confirmation:
+    title: Request received. We're on it.
+    description: We've notified the Pulumi team and sent a copy to your email. Your trial stays active while we review.
+    recap:
+        - label: Organization
+          field: pulumi_organization_name_s_
+        - label: Reason
+          field: trial_extension_reason
+    steps:
+        - title: Now.
+          description: Your request is with the Pulumi team, along with your account details.
+        - title: Within 1 business day.
+          description: Someone reviews it and either extends your trial or replies with a question.
+        - title: Confirmation.
+          description: You'll get an email with your new trial end date. Nothing in your account changes in the meantime.
+    cta:
+        label: Back to dashboard
+        url: https://app.pulumi.com
+
+help_links:
+    title: Need something else?
+    description: "If this isn't about your trial length, these get you there faster:"
+    links:
+        - label: Talk to sales about pricing
+          url: /contact/
+        - label: Open a support ticket
+          url: https://support.pulumi.com/
+        - label: Ask the community on Slack
+          url: https://slack.pulumi.com/
 ---
--- a/content/extend-trial/_index.md
+++ b/content/extend-trial/_index.md
@@ -1,0 +1,8 @@
+---
+title: Extend your trial
+meta_desc: Request more time on your Pulumi trial. Tell us what you're still evaluating and we'll extend it. Most requests are approved within one business day.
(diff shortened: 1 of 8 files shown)


jeffmerrick:
@calon-pulumi I refactored this a bit to reuse the existing Hubspot form component we already have (was able to fix some issues with it as well so that will help us with future embeds.) Please double check that everything is still working as expected! Summary of changes below:

---

The page now uses `<pulumi-hubspot-form>` instead of a direct `hbspt.forms.create()` embed, so it picks up the lazy load, loading/failure states, UTM + `iaid` capture, and Segment tracking that the rest of the site's forms have. All copy moved into front matter; the layout composes `.card`, `.btn`, `.badge`, the `heading-*`/`body-*` scale, and sprite Phosphor icons, with no page-level CSS.

**Why the styling needed work:** this form is built in HubSpot's newer form editor, which emits fields as bare `.hs-form-field` divs with no `<fieldset>`. The field rules in `_hubspot.scss` are all nested under `fieldset`, so none of them reached it — `/contact/` has 7 fieldsets, this form has 0.

### Changes

**`hubspot-form.tsx`** — two additions to the shared component:
- `prefill` prop: JSON map of HubSpot field name → query param, applied on `onFormReady`, so the console link can arrive as `?org=&email=`
- `hubspotFormSubmitted` event: lets a page render its own confirmation. Values are captured on `onFormSubmit`, before HubSpot replaces the form
- Plus an `event.data.id` guard, since HubSpot broadcasts callbacks on `window`

**`_hubspot.scss`** — three fixes that apply to all HubSpot forms:
- Field chrome extracted to a mixin with a second call site for fieldset-less forms, kept at the same specificity via `:where()`
- Validation messages now render red on `hs-form-fg-dark` forms (`.hs-form-fg-dark label` was winning over `.hs-error-msgs label` on source order)
- `.inputs-list` outer margins trimmed, so the global `main li { my-2 }` stops padding checkbox descriptions and error text

**Also:** new `theme/src/ts/extend-trial.ts` for the confirmation swap, and a `.card-highlight` variant added to `shared/_card.scss` (uses `var(--docs-bg-alt, violet-50)` so it works in docs dark mode).

jeffmerrick (approved):

PR #20663, opened 2026-08-03
Tier Vale findings in pre-merge reviews: blockers in 🚨, quieter nags, FP fixes

CamSoper:
### Proposed changes

Restructures how Vale prose-lint findings surface in the pinned pre-merge review. The tiering/FP changes are deterministic config/Python with **no new model calls**; the inline-suggestions feature adds a small amount of work to the existing Opus editorial pass (≤10 one-line rewrites) and one API call.

**1. Blocker tier — high-confidence Vale findings now block.** A new `blocker:` allowlist in `vale-deterministic-fixes.yaml` (`Pulumi.Substitutions`, `Pulumi.Nomenclature`, `Pulumi.PoliciesSingular`, `Google.Spelling`, `Pulumi.DeprecatedProductNames` — wrong or deprecated product names, banned terms, misspellings, agreement errors; near-zero false-positive rate) is stamped onto findings by `vale-findings-filter.py` (exempt from the per-file/total caps) and rendered by `compose-review.py` as `[style-blocker]` bullets under 🚨 Outstanding with standard `**[L<n>]**` anchors. They count in the 🚨 cell, drive the `review:outstanding-issues` label, participate in the auto-refresh gate on fix-pushes, and are outcome-classified by the review telemetry. The model may not delete or demote them. The triage lane surfaces them with a `[blocker]` prefix in the `TRIAGE_PROSE` comment.

Blocker membership is independent of `deterministic_fix`: `PoliciesSingular` and `DeprecatedProductNames` block even though their fixes need authoring ("Pulumi Service" is always wrong; the right replacement — Cloud/ESC/Deployments — is the author's call). `DeprecatedProductNames` was also missing from `RULE_CATEGORIES` and rendered as generic "style"; it now renders as "deprecated product name".

**2. Advisory findings quieted, but actionable.** Non-blocker style findings now always render collapsed (per-file `<details>`) and are excluded from the ⚠️ count-table cell — kept for the style-rule tuning loop, not reviewer burden. Additionally, the editorial pass triages the advisory findings and stages the ones whose rewrite is a clear, meaning-preserving improvement into `.style-suggestions.json` (cap 10; false positives and load-bearing hedges — "usually completes in five minutes" — are explicitly excluded). A new deterministic step (`post-style-suggestions.py`) validates every entry against the PR diff and the checked-out file (anti-hallucination: the anchor must be a PR-added line and the `original` text must be present verbatim), then posts one `event: COMMENT` review with a ```suggestion``` block per finding — one-click applyable, batchable into a single commit, and deleted-and-reposted per run so no stale comments accumulate. The script never fails the workflow.

**3. False positives fixed at the source.**
- `HeadingSentenceCase` gains multi-word proper-noun exceptions (`HashiCorp Vault`, `AWS Secrets Manager`, `Azure Key Vault`, `Google Cloud Secret Manager`, `1Password Secrets Automation`, `CyberArk Conjur`, `Doppler`, `Infisical`) — phrase entries so generic words like "Secrets"/"Manager"/"Key" still flag when capitalized outside a product name (verified against fixtures).
- `Google.OxfordComma` is disabled with a rationale comment: its regex can't distinguish serial lists from introductory clauses ("…the deciding factors, Doppler and…"), joining the nine other Google rules already disabled for FP rate.

Contract copy is updated everywhere it's duplicated: `output-format.md`, `ci.md` (including new editorial-pass step 10 for the suggestion sidecar), `SKILL.md`, the prompts in `claude-code-review.yml` and `claude-update.yml`, the validator rules (`count-table-matches-bullets`, `style-render-mode`, trail-match exemption), `scrape-review-outcomes.py`, and `STYLE-GUIDE.md` §Automated checks.

**Verification done:**
- Vale 3.14.1 fixture run: all six product-name headings pass; genuinely wrong headings ("Configuring Your Secrets", "The Manager Approves") still flag; the Oxford-comma sentence no longer fires.
- Filter test: blocker stamping + cap exemption; `DeprecatedProductNames` stamps `blocker: true`, `deterministic_fix: false`.
- Composer dry-run: blockers render in 🚨 with anchors and are counted, nags collapsed and uncounted; composer self-check clean.
- Validator: passes with a non-empty trail and `[style-blocker]` bullets that have no trail records (the exemption path), and still enforces counts.
- `auto-refresh-gate.py` fires on a push fixing a blocker's line.
- `post-style-suggestions.py`: 9 new tests covering validation (missing text, non-added line, duplicates, cap, payload shape) and the dry-run end-to-end path.
- Full scripts test suite: 105 passed.

N/A

### Related issues (optional)

N/A


CamSoper:
@claude #update-review

CamSoper:
@claude #update-review

Re-verify against 91237991f1. That commit acts on a separate quality pass over this branch: it fixes phantom ✏️ marks on split reviews (annotate_text now strips unconditionally), skips the delete-and-repost when the suggestion set is unchanged, restores the blocker tier on ci.md's composer-crash fallback path, and reconciles several contract docs. Nothing has reviewed it yet.

CamSoper:
@claude #update-review

Pushed d1ac01c66a. It makes the italic caption under `#### Style suggestions` workflow-written, the same way the ✏️ marks and the banner already are — the re-entrant lane renders freehand and was observed dropping the "✏️ marks one you can apply" legend on fork PRs #231 and #232, leaving marks on the page with nothing explaining them. `annotate_text` now reconciles the caption on both lanes, and a cross-file test pins it to what `compose-review.py` emits so the two can't drift.

CamSoper:
@claude #update-review

Pushed 174f5feec9. It keys the suggestion short-circuit on the replacement line rather than the whole comment body — the body embeds the model-authored `note`, which is rewritten every run, so the desired set never matched the posted one and the short-circuit added in 91237991f1 could never fire. Measured across three runs on fork #232. Confirming the fix live on a fresh fork pair now.

CamSoper:
@claude #update-review

Pushed 6e16b94fda. A refresh where the model omitted `.style-suggestions.json` was deleting live one-click buttons: `main()` read a missing file and an explicit `[]` identically, so the distinction both prompts ask for was discarded exactly where it mattered. Now `[]` stays authoritative and clears the buttons, while absent/unreadable/not-an-array means unknown and leaves them standing — with the marks reconciled against what is actually posted, so they still can't over-promise. Observed on fork #233 refresh 2.

CamSoper:
@claude #update-review

Pushed 417e50c9cb, acting on all three ⚠️ findings from the last review — all of them were in code added earlier tonight.

- `live_posted` conflated `fetch_prior_suggestions` returning `None` (listing failed) with `[]`, which is the same unknown-vs-confirmed-empty conflation the previous commit removed, reintroduced one helper over. It now returns `None` and the caller touches nothing.
- `_reconcile_caption` sniffed for single asterisks, so a bold-wrapped caption got a second caption inserted above it and the forbidden column-0 bold line survived. Replaced with a positional rule.
- claude-triage.yml's comment still named the pre-rename `[blocker]` prefix.

PR #19777, opened 2026-06-20
docs: consolidate voice/writing-style guidance into the brand MCP

cnunciato:
**Draft** — opened primarily to verify the brand MCP server connection in CI.

## What this does

Makes the Pulumi brand guide (public **`pulumi-brand`** MCP server at `https://brand.pulumi.com/mcp`) the source of truth for general **voice + writing-style**, and has the docs `STYLE-GUIDE.md` and content skills consult it rather than restating it. Adds a **precedence rule**: wherever the brand guide overlaps anything in this repo, the brand guide wins.

## Verifying the MCP connection

The brand MCP is wired into `claude-code-review.yml` and `content-review-article.yml` via `mcp_config`, pointing at the already-live public server (no auth). Note: the auto-review fires via `workflow_run`, which uses **master's** copy of the workflow — so this PR's new `mcp_config` is exercised by **manually dispatching the review on this branch**, not by the auto-trigger.

## Out of scope

SEO/AEO and social-copy skills stay in this repo untouched for now; they'll be handled in the brand guide separately. The brand-side changes (voice/writing-style enrichments) are a separate `pulumi/marketing-web` PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #20664, opened 2026-08-03
docs: consolidate voice/writing-style/terminology guidance into the brand MCP (tooling only)

CamSoper:
### Proposed changes

**Tooling only — no content files change.** Makes the Pulumi brand guide (the public `pulumi-brand` MCP server) the source of truth for voice, writing style, and terminology, and teaches the offline Vale mirror the brand guide's **Names & terminology** section.

Supersedes #19777. The content migration these rules surface is tracked in **#21133** and is deliberately out of scope.

## What's in it

| Change | Why |
| --- | --- |
| **Vale terminology mirror re-tiered.** `Substitutions` (replacement right in every context), `RetiredNames` (right in most, so the message hedges), `DeprecatedProductNames` (flag-only). All three block. | The mirror was stale and, in one case, *inverted*: `Nomenclature` was auto-canonicalizing prose toward "Pulumi Insights", a retired name. |
| **`ThirdPartyNames` (new, blocker).** Github, Gitlab, Javascript, Typescript, Cloudformation, Postgresql, Wordpress, VMWare, Nodejs, Hashicorp, Oauth. | The brand guide says to match each project's own casing. The dictionary accepts "Github", so spell-check can't see it. |
| **`Spelling` (new, advisory).** Hunspell plus an 816-word hand-triaged vocabulary. | `Google.Spelling` is not a spell-checker despite the name — four tokens for British spellings. Nothing checked real typos. |
| **All lookbehinds removed.** | Vale silently drops a negative lookbehind that starts its scoped window, so identical prose flags or doesn't depending on unrelated text length. Lookaheads are stable. |
| **Sentence case at every heading level**, front-matter titles included via `scripts/lint/frontmatter-title-case.py`. | Vale reads front matter, but the sentence-case rule is heading-scoped and front matter isn't parsed as a heading. Advisory and changed-files-scoped, so the backlog surfaces one page at a time. |
| **Weekly brand-sync job** (`brand-style-sync.yml` + the `brand-vale-sync` skill), gated on a reachability probe. | An offline mirror drifts. The probe runs before the model starts, because a sync written from memory is worse than no sync. |
| **Brand MCP wired into all 8 `claude-code-action` workflows.** | Agents writing or reviewing content should consult the guide directly rather than a local copy. |
| **Redundancy removed** from `STYLE-GUIDE.md` and `AGENTS.md`; the duplicated terminology table replaced with a pointer. | The brand guide leads; the repo follows. Two copies of a naming table is one too many. |

## Exemptions

- **`content/blog/`** is exempt from the four Pulumi-name rules — the brand guide says dated posts should remain as they are, and ~105 archived posts would otherwise flag. It is **not** exempt from `ThirdPartyNames`/`Spelling`: "Github" records no history, so fixing it falsifies nothing.
- **`content/docs/iac/cli/commands/`** — generated from `pulumi gen-docs`; fixes belong upstream.
- **`azure-devops-integration.md`** — its "Pulumi Service" is the literal name of a group the integration creates.

## Notes

- **`Google.Spelling` is on neither tier list.** 3 of its 6 corpus findings are proper nouns it can't distinguish ("RAI Convention Centre", "Organise imports"), which fails the blocker list's near-zero bar.
- **Backlog: 335 findings across 194 files** (#21133). None of it blocks anyone — findings are diff-scoped, so a blocker only fires on a line an author is writing today.
- Regression tested on the fork, plus `make lint` and `make test-review-pipeline`. Record: pulumi/docs-review-benchmarks#8.


CamSoper:
@claude #update-review

All findings addressed in `1bf8ada`. Summary of what changed and what I verified, finding by finding.

## 🚨 `self-managed backend` swap (outstanding) — fixed

**Confirmed it belongs in the mirror.** Pulled `get_guidelines({section: "terminology"})`; the retired-names table has the row verbatim: `| self-managed backend | DIY backend | This is different from _self-hosted_ (i.e., customer-managed, on-prem) deployments of Pulumi Cloud. |` So the row stays.

**Backlog cleared rather than downgraded.** All 53 occurrences checked in context — every one means the DIY backend (S3 / Azure Blob / local filesystem), and the repo already preferred that term 95:53, with `circleci.md` linking to `#using-a-diy-backend`. Migrated the 40 non-blog occurrences; the 13 in `content/blog/` stay (see below). Non-blog count is now **0**, so the row keeps `level: error` with no backlog behind it.

**The token was also broken.** `\bself-managed backend\b` never matched the plural — `\b` doesn't split `backend`/`backends`, so it silently missed 23 of the 40 non-blog occurrences while flagging the 17 singular ones. Both forms are now covered.

## New: an advisory tier, because some of these swaps aren't safe to automate

Auditing the rest of my added rows against real content turned up two cases where a blocker-tier auto-fix is actively wrong — and the brand guide's own preamble predicts both:

> *"Some replacements can't be automated."* … *"Historical usage is fine. Prose that intentionally describes past behavior … is acceptable."*

- `content/case-studies/unity.md:50` — "Pulumi Neo is our AI platform engineer, **replacing Pulumi Copilot** with expanded capabilities." The swap turns this into "Neo … replacing Neo". Line 95 also carries a `{#copilot}` heading anchor a rename would break.
- "Insights" resolves to the Discovery *product* in most sentences but to the *"Discovery & governance"* category in others — its own separate row in the brand table — and the docs IA still lives under `/docs/insights/`.

So renamed-product rows moved to **`styles/Pulumi/RetiredNames.yml`** (`level: warning`, off the `blocker:` list, still on `deterministic_fix:` so `review-existing-content`'s meaning-preserving gate applies). Rows graduate to `Substitutions.yml` once their backlog reaches zero — exactly what `self-managed backend` just did. `BRAND-SYNC.yaml` records the tier rule so the weekly sync job files new entries correctly.

## ⚠️ `DeprecatedProductNames` lookbehind tokens — fixed, and the finding was right

Reproduced a genuine defect: **Vale silently drops a negative lookbehind that starts its scoped window**, so identical prose flags or doesn't depending on unrelated preceding text. With `(?<!VS Code )\bCopilot\b`, this line

```
GitHub Copilot works well. Microsoft Copilot exists. VS Code Copilot too.
```

is clean under an 8-character H1 and flags `VS Code Copilot` under a 17-character one. Lookaheads are stable across the same sweep (30 padding lengths, 0 variation).

The rule now has **no lookbehinds**. Bare `Insights` and bare `Copilot` are dropped and documented as intentionally unmirrored — both needed a lookbehind to avoid double-flagging the `Pulumi Insights` / `Pulumi Copilot` that `Substitutions` already handles, and bare `Copilot` collides with the GitHub / Microsoft / VS Code products this site legitimately references. `Pulumi Service` keeps its two lookaheads.

## ⚠️ Heading case (L39) — verified, and the four sibling files are fixed

The verification step couldn't reach the brand page; I pulled it through the MCP server. It says, under **Headings and titles**: *"Always use sentence case for headings (e.g., 'This is a heading', not 'This Is a Heading')."* No H1 or page-title carve-out anywhere in the section. The terminology section is more explicit still: *"Pulumi has no title-case convention: headings are sentence case, and names keep their own casing wherever they appear."* And empirically, brand.pulumi.com's own H1s are sentence case ("Writing style", "Names & terminology", "Voice & tone"). So yes — every level, page title included.

All four contradicting siblings updated: `new-changelog.md:57` and `:73`, `archetypes/case-studies.md:11`, `new-doc/references/questions-index.md:219`. I left `create-event/references/event-page.md:11` alone, per your read that it's a marketing-template requirement rather than a restatement of the heading rule.

## ⚠️ Code samples section (L142) — verified

`get_guidelines({section: "writing-style"})` returns a `## Code samples` section under that exact name, covering indentation (four spaces; two for YAML, tabs for Go), quoting (double outer, single inner), comments, and line-splitting for long CLI commands. The pointer is correct.

## ❌ Glossary path — fixed

Good catch. `/docs/iac/concepts/glossary/` is only an alias on `content/docs/reference/glossary.md`; both references now use the canonical `/docs/reference/glossary/`.

## Also fixed: two pre-existing false positives the audit surfaced

- **`language-native package`** fired on "language-native package **managers** (npm, PyPI, NuGet, Maven, Go modules)" — a different meaning entirely. That was **13 of 13 occurrences**, i.e. a 100% false-positive rate at blocker tier. Now excluded with a lookahead.
- Generic **"Pulumi deployments"** / **"Pulumi Service"** prose corrected where unambiguous. The one legitimate "Pulumi Service" — the literal group name the integration creates in Azure DevOps — is exempted per-file rather than rewritten, since changing it would misdocument the actual UI string.

**Net effect, measured:** running master's rule set and this PR's rule set over identical content, master produces **46** blocker-tier findings and this PR produces **31**. The PR reduces blocker noise rather than adding it.

Remaining findings in that set are pre-existing `Nomenclature` casing hits (`Pulumi UI`, `Pulumi Console`, generic `Pulumi deployments`) that fire on master today and are unrelated to these changes — they belong to the content-migration PR in recommendation 1, not here.

## Verification

`make lint` clean (1850 files, 0 errors, Prettier clean) · `make test-review-pipeline` all green (pytest + standalone harnesses + every `--self-test`) · rule behavior validated with vale 3.9.4 against positive and negative fixtures and a padding-stability sweep · blog exemption confirmed (0 findings on a post containing "Pulumi Insights").

One note on the blog exemption, since it's the largest behavioral change here: without it, the new terminology rows would have turned **105 archived blog posts** into blocker findings on any PR touching them. The brand guide is explicit that *"dated blog posts should remain as they are"*, so `content/blog/` is now exempt from the three terminology rules — the same treatment it already gets for `FutureTense` and `NarrativeWe`.

Rebased onto @CamSoper's `18968db` (`|| true` on the title-case helper) — kept as-is.

---

CamSoper:
@claude #update-review

Scope change: **this PR is now tooling-only.** The 41 content files it was editing (48 terminology swaps) are backed out in `edc3e3e` — `git diff origin/master...HEAD -- content/` is empty. Those docs edits want their own reviewed PRs, batched by tree, and are inventoried in #21133 (assigned).

One honest consequence for the rules, since it changes the resolution of the earlier outstanding finding:

`self-managed backend` was promoted to `Substitutions.yml` (error, blocker) **only because** this branch had migrated its 40 occurrences. With the content restored it has a live backlog again, so both singular and plural forms move to `RetiredNames.yml` (warning, advisory) alongside the other rows awaiting migration. That is the same remedy this review originally proposed for it — the advisory tier — now applied because the backlog is real rather than because the swap is unsafe.

The upside is that `Substitutions.yml` is now exactly what its header claims: terms that are wrong in every context **and** carry no backlog. Adding a row there can never land a blocker on an unmigrated page. `RetiredNames.yml` and `BRAND-SYNC.yaml` record the per-term backlog and point at #21133 as the graduation path.

**No tooling behavior is lost.** Unchanged from the previous review pass: the rule set, the advisory tier, the lookbehind removal (with the reproducer documented in the rule file), the `language-native package` false-positive fix, the blog and generated-CLI exemptions, the `azure-devops-integration.md` carve-out, the sentence-case H1 work, and the weekly sync workflow.

**Re-measured on unmodified master content** (the cleaner comparison, now that the branch touches no content): master's rule set produces **54** blocker-tier findings, this PR's produces **39**.

`make lint` clean (1852 files, 0 errors) · `make test-review-pipeline` all green.

CamSoper:
## Fork regression test — passed

Ran on `CamSoper/pulumi.docs` against a synthetic fixture built to trip every new rule at once. Record: pulumi/docs-review-benchmarks#8 (`2026-08-25-vale-brand-mcp`), which extends the `2026-08-03-vale-tiering` campaign — same fork, same fixture file, harness reused.

Six arms across three surfaces, each with a same-day control.

| check | control (master) | this branch |
|---|---|---|
| validator violations | 0 | 0 |
| Vale blockers rendered | 4 | **17** |
| retired names missed | 7 | **0** |
| miscased project names missed | 4 | **0** |
| brand MCP | not wired | **connected** |
| update lane refresh | complete | complete, identical counts |
| allowed-tools lost in the `claude_args` rewrite | — | **0 across 7 workflows** |

The 17/7 blocker/advisory split matched the local vale 3.14.1 prediction exactly, which is what makes a single rep worth anything here. The new `Pulumi.Spelling` rule routes to the uncounted advisory block as intended — it's the first `extends: spelling` rule in the set, so that mapping was untested until now.

### Four things the test changed in this branch

1. **`brand-style-sync.yml` hand-downloaded Vale 3.9.4 while `mise.toml` pins 3.14.1** — the workflow that exists to catch drift was drifting. Now installs via `mise-action`. (Rule behavior is identical across the two versions; verified with a 28-finding probe, so this is hygiene, not a fix.)
2. **`Conjur`, `CyberArk`, `supergraph`** were false positives in the new spelling rule; added to the vocabulary.
3. **`BRAND-SYNC.yaml` still described `RetiredNames` as advisory** — stale from before it moved to the blocker list. The weekly sync job reads that file to tier new rules, so the wrong claim would have mis-tiered the next one.
4. **The sync skill's tier list omitted `RetiredNames` entirely** — a file this PR adds. Both documents now agree on all four tiers.

### One finding left open on purpose

Reading the live brand table through the MCP shows retired-name rows the mirror doesn't cover yet: the `Pulumi Console / Pulumi Cloud Console / Pulumi UI` row, the `Discovery & governance` category row, `Policy as Code` (as a product name), and `Insights Policy`. All look safe to mirror. I left them for the weekly `brand-style-sync` job rather than widening this PR — it's the mechanism this PR ships, and a first run with real drift to find is a better test of it than an empty one. `runtime code` → `function serialization` stays deliberately unmirrored; the brand guide itself warns the word has too many live meanings.

### Worth knowing before reading too much into the MCP

The ad-hoc lane control answered "what is Pulumi Insights called now?" correctly **without** the MCP, by grepping the repo. The MCP didn't make the answer right — but the control sourced it from a docs page that still carries the retired label "Discovery & Governance" and repeated it, where the MCP arm rendered the canonical form. The value is provenance, not accuracy, and the record says so.

CamSoper:
## Overnight review pass — four reviewers, nine defects, all fixed

Ran four independent reviews over this branch (Vale rule correctness, workflow/CI safety, pipeline integration, docs accuracy), fixed everything real they found, and re-ran the fork regression against the final SHA.

**Readiness summary:** https://claude.ai/code/artifact/17963dd2-c156-49e4-8b2c-4fd8ca3423c0
**Rule reference (refreshed):** https://claude.ai/code/artifact/69d26995-b1be-4ec5-8342-e098ec2781f1

### What was actually wrong

**Three live false positives on shipping pages.** `Nomenclature` read `pulumi policy` and `pulumi neo` — real CLI command groups — as miscased product names, on `content/docs/insights/policy/cli.md` and `content/docs/ai/neo/pulumi-cli/_index.md`. A subcommand allowlist can't rescue those; the bare group name is equally valid prose. The three rows that collide with command groups came out, matching the precedent the file's own header already set for `pulumi deployments`. Separately, `RetiredNames` is `ignorecase`, so `Account Discovery` fired on the heading "View account discovery progress" — now case-pinned.

**The weekly sync job's core guarantee was fictional.** "If the brand MCP is unreachable, exit non-zero" was a bullet in a prompt. A model can't set the job's exit code, and `claude-code-action` reports success for any completed session — so a brand.pulumi.com outage would have produced a confident sync written from memory and gone green. That's precisely the failure the design exists to prevent. Now there's a `curl` probe that fails the job before Claude starts. The same job was also missing `make ensure` (its `make lint` step could never have run), a timeout, a concurrency group, and PR de-duplication — it would have opened up to 52 conflicting draft PRs a year.

**Both new blocker rules rendered as `_style_`.** `RULE_CATEGORIES` was never updated, and `category` is the only rule identity a reader ever sees. Now confirmed rendering as `_retired product name_` (9) and `_third-party name_` (4) in live CI.

**One repeated word could consume a file's entire advisory budget.** Measured on a real post: `superintelligence` recurred 27 times and took 8 of 10 slots, silently dropping every heading-case finding. `cap()` now de-duplicates before capping.

**Plural terms would have been rewritten to singular.** Eight rows missed the plural. The obvious one-row fix would have rewritten "Insights accounts" → "cloud account", since these carry a fixed replacement. Paired rows instead.

Plus: archetypes still generated Title Case (a page scaffolded by `new-changelog` lint-dirty on creation), glow-up's MCP hard-stop silently aborted the unattended lane without its ledger sentinel, and "Vale can't see front matter" was false in eight places — which mattered, because 15 findings in the #21133 backlog *are* front matter.

### One judgment call worth your veto

I removed **`Google.Spelling` from both tier lists.** Despite the name it isn't a spell-checker — four tokens for British spellings — and 3 of its 6 corpus findings are proper nouns it can't distinguish: "RAI Convention **Centre**" (the Amsterdam venue, twice) and "**Organise** imports" (a literal VS Code command label). A 50% false-positive rate fails the blocker list's own near-zero admission criterion. It was also on `deterministic_fix` while supplying no replacement to apply — `styles/Google/Spelling.yml` is `extends: existence` with no `action:`. Easy to revert if you disagree.

### Re-test

Same byte-identical fixture as the first run. **17 blockers across two reps**, matching the local Vale 3.14.1 prediction exactly; correct category labels; `pulumi-brand: connected`; validator violations **0**.

One honest note: the first of those two reps raised a `trail-verdict-bucket-promotion` violation that the second did not, with no code change in between. The paths I touched can't reach trail-verdict placement, so it reads as claim-lane variance — but it's n=2 and it's written into the record as a caveat rather than dismissed.

### New tests

50 assertions that didn't exist before. The suite would previously have passed with all three new rules misspelled in the tier list. Now: every tiered rule file exists, every blocker has a category and is `level: error`, blockers bypass the caps, the dedup actually prevents starvation, and all 8 workflows wire the brand MCP identically — that last one verified against injected drift.

Backlog is **335 findings across 194 files** (#21133 updated). None of it blocks anyone; findings are diff-scoped.

CamSoper:
@claude #update-review

Worked both reviews — the pinned pre-merge review and the Unblocked bot — against current HEAD (`3379b59`). Most findings predate the overnight pass and were already fixed; two were not, and are now.

**🚨 Outstanding — both resolved**

1. **`RetiredNames` `%s` order inverted.** Fixed. Live output now reads `Use 'Pulumi Discovery' — 'Pulumi Insights' is a retired name.` — replacement first, matching the house form in the other substitution rules.
1. **`CrossGuard` in `Substitutions.yml` with a live backlog.** Took your option 1: both `Pulumi CrossGuard` and `CrossGuard` moved to `RetiredNames.yml`. Worth noting the *reasoning* has since changed — backlog size was explicitly rejected as a tiering criterion, because the PR review is diff-scoped so a blocker only fires on a line an author writes today. All three files now block; the file a row lives in decides only how much its message hedges.

**⚠️ Low-confidence**

1. **Census counts (`Pulumi Copilot 2`, `Policy Violations 3`).** Moot — the census was the graduation gate under the backlog-tiering design and was removed with it.
1. **`Policy Violations` matching the lowercase generic phrase (~30 occurrences).** Fixed with an inline `(?-i)`, so the capitalized console page name fires and "Fewer policy violations" doesn't. Verified both directions. Good catch — that one was a real regression I'd introduced by moving the row under `ignorecase: true`.
1. **`DeprecatedProductNames` header still naming `Substitutions.yml`.** *Was still open.* Fixed now — an earlier pass missed it because it anchored on different line wrapping. You were right that this is a mis-filing risk rather than a comment nit: `brand-vale-sync` reads these headers to decide where a new row goes.
1. **Blog exemption applies to new posts, not just the archive.** *Was still open.* Kept, but now recorded as a deliberate decision rather than an inherited one, with its compensating control: blog PRs are classified `domain:blog` and route to `claude-code-review`, which this PR wires to the brand MCP — so a new post's terminology is checked by a model reading the **live** retired-names table, which is stronger than the offline mirror it replaces. Noted to revisit if that lane ever stops covering blog.

**Unblocked bot** flagged three files where "Pulumi Cloud" had been replaced with the retired "Pulumi Service" (`greenpark-sports.md`, both `kubernetes-helm-*` tutorials). All three are moot: the PR is tooling-only now and touches zero content files. The underlying occurrences are pre-existing in master, and all three are caught by the new rules and counted in the #21133 backlog.

Since that review ran, four independent review passes found nine further defects — three live false positives, a workflow whose fail-loudly guarantee was only a prompt bullet, and a category-label bug that rendered both new blocker rules as "style". All fixed and regression tested on the fork.

jkodroff (approved):

PR #20682, opened 2026-08-04
add docs for new automatic logging feature

tgummerer:
Since this is the new way of sharing logs, add more docs to the troubleshooting guides.


content/docs/iac/operations/debugging/logging.md
```diff
@@ -17,7 +17,24 @@ aliases:
 - /docs/iac/troubleshooting/debugging/logging/
 ---
 
-Pulumi provides two main types of logging to help with debugging and troubleshooting: CLI logging for diagnosing Pulumi engine operations and program logging for emitting custom diagnostics from your Pulumi programs.
+Pulumi provides several ways to capture logs for debugging and troubleshooting: automatic logging, which records every operation to an encrypted file you can share securely with the Pulumi team; CLI verbose logging for diagnosing Pulumi engine operations on demand; and program logging for emitting custom diagnostics from your Pulumi programs.
```
julienp: This makes it sound like users should use verbose logging, and that automatic logs are just for the pulumi team. Is that what we want or should we encourage users to use automatic logs for their own debugging too?

content/docs/iac/operations/debugging/logging.md
julienp: Mention automatic logging here

content/docs/iac/operations/debugging/logging.md
tgummerer: That was not the intention.  We can probably just drop the CLI verbose logging.  Mostly I see the logging stuff as a developer tool and not so much as a end user facing tool most of the time, but we can probably just cut out the mention of verbose logging here completely, as automatic logging should be at equivalent in almost every case.

julienp (approved):

PR #20665, opened 2026-08-03
Surface recent blog posts and the next event on the homepage

jeffmerrick:
Screenshot 2026-08-03 at 1 59 54 PM

### Proposed changes

Adds a "What's new at Pulumi" section near the bottom of the homepage: a three-tile row of the latest blog posts (excluding the `general` category), with the next upcoming event in the last slot and a fallback to three posts when nothing is coming up. It reuses the existing blog and event card partials rather than inventing new tiles — which meant making the event card portable, so it now derives its own link and date from the event page and renders just the `<article>`, with the filterable `<li>` moved to `events/event-list-item.html` and the "what's upcoming" query moved to `events/upcoming.html` so the events list and the homepage can't drift apart. `blog/feed.html` now also returns its non-`general` post pool, and the new `blog/card/contained.html` boxes the medium card in a `.card` so it sits comfortably beside the event tile.

The rendered `/events/` page was diffed against `master` and is byte-identical apart from one intentional tweak: the event card's internal `gap-16` is now `gap-6`, which only affects the card in a narrow column (`justify-between` already pins the date chip right when there's room). The countdown badge renderer in `resources.ts` is no longer gated on the events list, so the "in X days" chip also shows on the homepage.

### Related issues (optional)

Fixes #20584

PR #20662, opened 2026-08-03
Add a copy-prompt agent CTA to the homepage hero

jeffmerrick:
Screenshot 2026-08-03 at 12 09 31 PM

### Proposed changes

Groups the homepage hero's **Get started** / **Download open source** buttons into a bordered "For humans" card and adds a "For agents" card beside it, whose single button copies a one-line onboarding prompt (`Fetch https://www.pulumi.com/onboard.md and follow its instructions…`) for pasting into a coding agent; the prompt's target ships here as a static `/onboard.md`. The hero's CTA row moves into a new shared `partials/hero-ctas.html` used by both the split and centered layouts, gated behind optional front matter (`cta_card_label`, `cta_agent_prompt`, `cta_agent_prompt_label`, `cta_agent_label`) so the other 12 `type: hero` pages render as before — except that the arrow icon is now dropped from the primary CTA everywhere, which is intentional but site-wide. Also adds a generic `[data-copy-text]` clipboard module and fixes a pre-existing hero bug where the left column was sized from its content, letting a wide child overflow the viewport and making the column's width shift whenever child text changed.

### Related issues (optional)

Fixes #20327

PR #20698, opened 2026-08-04
Add a /learn hub page and collapse nav and footer onto it

jeffmerrick:
Closes the docs half of pulumi/marketing#1792: a single front door at `/learn` that points at every learning resource — the Dev Hub, Pulumi Academy, the get-started path, workshops, and the blog's series and tutorials.

> [!IMPORTANT]
> **This PR stacks on two open PRs and should not merge before them.** The diff below includes their commits.
> - **#20665** (`jeffmerrick/issue-20584-latest-on-homepage`) — hard dependency. Provides the portable `events/event-card.html`, `events/upcoming.html`, and `renderEventCountdowns()` running outside `.template-event-list`.
> - **#20662** (`jeffmerrick/issue-20327-homepage-agent-cta`) — soft dependency. Provides `static/onboard.md` and the generic `theme/src/ts/copy-text.ts` clipboard module that the "For agents" half uses. If it doesn't land, dropping the `agents:` key from `content/learn/_index.md` removes that half and nothing else changes.
>
> Once both are on `master`, this branch just needs a rebase.
>
> **It also should not go live before [pulumi/marketing-web#82](https://github.com/pulumi/marketing-web/pull/82).** That's the Dev Hub itself, in a different repo, so it isn't a merge-order dependency in git terms — but `/dev-hub/` 404s until it ships, and this PR links it from the spotlight card *and* the site footer. If that PR slips, either hold this one or drop the two Dev Hub links and add them back later.

## What's here

**`/learn` was an alias.** It was declared on `content/tutorials/_index.md`, so `/learn` 302'd to `/tutorials/`. That entry is removed — Hugo would otherwise emit a duplicate-target-path warning and pick an order-dependent winner. The 32 deeper `/learn/<tutorial>/` aliases on child pages sit below the new page's path and are unaffected (verified: `/learn/embedding-pulumi/` still redirects to `/tutorials/embedding-pulumi/`).

**A bespoke section page**, modeled on State of Agentic Infrastructure rather than the `template-partials` system. All content lives in `content/learn/_index.md` frontmatter; `layouts/learn/list.html` dispatches each top-level key to a partial in the private `layouts/partials/learn/` namespace, so a section disappears by dropping its key. Styling is inline Tailwind over the shared primitives (`.card`, `.card-highlight`, `.btn`, `.badge`, the `heading-*`/`body-*` scale) — no new SCSS file. `/learn` is outside `/docs`, so no dark-mode pass applies.

Page composition:

1. **Hero** — the centered `h1` + description pattern from `/case-studies/`.
1. **Spotlight** — Dev Hub and Pulumi Academy as two dark violet-950 cards with centered copy and bottom-aligned CTAs.
1. **New to Pulumi?** — For humans and For agents, split across one card. The human half is four cloud links over a Get started link; the agent half is the onboarding prompt and a copy control sharing one outlined box, plus links to Agent Skills and the Pulumi MCP server. Per `AGENTS.md` §"AI and agent positioning", it names Claude Code, Cursor, and Codex as first-class rather than presenting Neo as the only option.
1. **Workshops and events** — upcoming events (capped at half the grid) then curated on-demand recordings.
1. **Blog series** — four curated series, beginner-first.
1. **Latest tutorials** — the four newest posts in the `tutorials` category.

**`layouts/partials/events/on-demand.html`** is new: the mirror of `events/upcoming.html` (past, has a `youtube_url`, newest-first). The events grid uses it to back-fill any curated slug that no longer resolves, so a deleted or renamed event can never leave a hole.

**Nav and footer collapse onto the new page.** In the header's "For engineers", **Templates**, **Tutorials**, and **Pulumi guides** become one **Learn** entry. Its icon is `compass`, not `graduation-cap` — that one reads as Pulumi Academy. In the footer's Resources column, **Templates**, **Tutorials**, and **Guides** become **Learn**, with **Dev Hub** and **Pulumi Academy** listed under it.

**Meta image** is generated on-brand at build time (`titleSection("learn", false)` in `scripts/generate-meta-images.mjs`); no `meta_image` in frontmatter.

## Changes outside /learn

Three shared-partial fixes fell out of building the page. Each is a strict improvement on its own and affects the blog:

- **`blog/post-meta.html` — bylines no longer wrap oddly at any card width**, without the caller having to know how wide its grid is. Each bullet is grouped with the segment it introduces, so a separator can't be stranded at the end of a line (the original bug: a dangling `•`). Reading time drops out below a `22rem` container query, which is roughly where the full byline stops fitting. Below that the author name truncates, since it's the only variable-width segment. **This is the theme's first container-query use** — Tailwind v4 ships them in core, no plugin needed.
- **`shared/_badge.scss` — `.badge` gets `font-normal`.** The weight was inherited before, so a badge composed onto a heading or inside a `<strong>` rendered bold and stopped looking like a badge.
- **`blog/avatars.html`** — `min-w-0` and a truncating label, so the name is what gives in a narrow card.

## Verification

- `make build` — no duplicate-target-path warning for `/learn`, no fingerprinted-asset panic, and the meta-image generator picks up the new section.
- `make lint` — clean.

Worth a look on a local server: `/learn/` end to end, `/tutorials/` with its child aliases, the copy-prompt button writing to the clipboard, and the blog's own card grids to confirm the byline change reads correctly there.


jeffmerrick:
Rather than having a separate page here, we're going to roll the /learn homepage into `marketing-web`

PR #20713, opened 2026-08-05
Exclude queue wait from the slow-build alert

CamSoper:
### Proposed changes

docsbot has been firing runs of "Build and deploy took *N*m, longer than usual" in #docs-ops with steadily climbing durations (21m, 23m, 25m, 28m, 31m…). That climb is the signature of a queue backing up, not of a build getting slower.

The cause: the "Build and deploy" step doesn't only build and deploy. Partway through `make ci_push`, `scripts/await-in-progress.js` parks and polls once a minute until any older run of the same workflow on the same branch finishes (`scripts/ci-push.sh:12`). When pushes to `master` stack up, each run waits longer than the one before it. The duration alert timed the whole step — `CI_BUILD_START_EPOCH` is stamped just before `make ci_push` and read after it — so that wait counted as build time and tripped the 20-minute threshold.

This change makes the alert measure only the build and deploy work:

- **`scripts/await-in-progress.js`** records the seconds it spent waiting to `.build-queue-wait-seconds` (gitignored). Recording it required restructuring the `setTimeout` recursion into a `while` loop so the poller has a terminating frame to report from; as a side effect the workflow ID is now resolved once instead of on every poll.
- **`scripts/ci-build-duration-alert.sh`** subtracts that wait from the step's wall-clock time and compares the remainder against `BUILD_DURATION_THRESHOLD_MINUTES`. When the wait is non-zero, the Slack message notes how much was excluded ("Excludes 14m spent waiting for an earlier run to finish."). A missing or malformed record falls back to timing the whole step, so nothing regresses if the file is never written.
- **`.github/workflows/build-and-deploy.yml`** clears a stale record before stamping the start time, so a reused workspace can't mask a genuinely slow build.

The threshold stays at 20 minutes; only what's being measured against it changes. Genuine regressions still alert — a 26m build behind a 5m queue is still reported as 26m.

Verified by running the alert script against synthetic start times and queue records (queued-but-fast stays quiet; slow-with-queue and slow-without-record both alert; malformed, absent, newline-less, and larger-than-elapsed records all handled), and by running the poller against a stubbed Octokit to confirm the loop terminates and writes the wait it actually observed.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -119,7 +119,7 @@ jobs:
         run: |
           # Clear any stale queue-wait record so it can't mask a slow build on a
           # reused workspace; await-in-progress.js rewrites it during the build.
-          rm -f .build-queue-wait-seconds
+          rm -f "${CI_BUILD_QUEUE_WAIT_FILE:-.build-queue-wait-seconds}"
           echo "CI_BUILD_START_EPOCH=$(date +%s)" >> "$GITHUB_ENV"
 
       - name: Build and deploy
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -116,7 +116,11 @@ jobs:
             hugo-resources-
 
       - name: Record build start time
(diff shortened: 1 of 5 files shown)


CamSoper:
@claude Implement the L20 suggestion and #update-review

PR #20728, opened 2026-08-05
Speed up the per-PR Lighthouse audit

jeffmerrick:
Trying to speed up the build a bit!

---

## Changes

**One Chrome instead of six.** `scripts/run-lighthouse-pr.sh` shelled out to `npx lighthouse` once per page/device combination, paying a cold Chrome boot each time. `scripts/lighthouse/run-audits.mjs` drives the Node API instead and runs all six against a single launched browser. Lighthouse still calls `newPage()` per run, closes it after, and clears storage, so the audits stay independent.

**No screenshots.** The report only reads the performance score plus FCP/LCP/TBT/CLS/SI, so every captured screenshot was encoded and thrown away. The expensive one is the full-page screenshot, and it can't be dropped via `skipAudits` — `full-page-screenshot` is a gatherer id, not an audit id, and `FullPageScreenshot` is in Lighthouse's `filterResistantArtifactIds`, so it runs regardless of which audits survive filtering. `disableFullPageScreenshot` is the only lever. The two real screenshot audits are skipped alongside it.

**One page list.** The audited pages move to `scripts/lighthouse/pages.json`, read by both the runner and the renderer, so the two can't drift.

`chrome-launcher` becomes a direct devDependency since `run-audits.mjs` imports it; it was already present as a Lighthouse transitive dep, and `yarn.lock` needed no change.

## Numbers

Measured on `ubuntu-latest` against real preview buckets, bounded by the audit's own log markers so old and new are measured identically.

- **135s → 82s**, ~39% faster, straight off `buildSite`'s critical path.
- Roughly three-quarters of that came from sharing the Chrome instance (135s → 97s); the rest from dropping the full-page screenshot (97s → 82s).

Single runs, not medians — see Follow-up.

## Verified

The audit ran green on this PR with all six rows scoring normally and none falling back to an `Error` row, using a throwaway `theme/` touch to trip the path filter (reverted; the diff above is unaffected). The `disableFullPageScreenshot` behavior was confirmed directly against Lighthouse 12.8.2 by resolving both the mobile and desktop configs through `initializeConfig`: the `FullPageScreenshot` artifact is present before the change and absent after, with the performance category and desktop form factor intact.

## Considered and dropped: a separate `lighthouse` job

Moving the audit to its own job made nothing faster in absolute terms: the fresh job re-paid ~55s of setup it had been inheriting from `buildSite`, so total runner time went up and only the gating check's report time improved. At an 11% trigger rate (13 of the last 120 merged PRs), that bought 82s on 1 in 9 PRs in exchange for an artifact hand-off, a duplicate ESC fetch, and an npm install that scraped `yarn.lock` and symlinked `node_modules`. Cleanly additive later if it's ever worth it.

## Follow-up

Median-of-3 is what would make these numbers trustworthy — the same page scored 89, 51, 45 and 34 across runs here. It's ~12 lines in `run-audits.mjs`, but it triples the audit load, so it belongs in its own PR.


jeffmerrick:
@claude #update-review

sicarul (approved):
Nice!

CamSoper (approved):
Looks great, looking forward to this

PR #20746, opened 2026-08-06
Add cost calculator to /pricing/

jeffmerrick:
Screenshot 2026-08-06 at 3 50 00 PM

### Proposed changes

Replaces the Stencil pricing calculator on `/pricing/` with a Hugo partial plus a plain TypeScript module, driven by a new `calculator:` block in `data/pulumi_pricing.yaml`.

* The estimator sits below the comparison table, since "what would this cost me" is a question you have after seeing the cards. Each edition card links down to it with "Estimate your cost" - on the editions the calculator prices - plus a "View all features" link to the table
* Those links carry the edition in the hash, so `/pricing/#calculator-enterprise` opens the estimator on Enterprise. Plain `#calculator` still works, and the estimator's heading links to it so the URL is easy to copy
* The volume-pricing note fades both ways now, via `@starting-style` and `transition-behavior: allow-discrete`. Browsers without them show and hide it instantly
* Meter ceilings are sized for large-scale usage - 250K IaC resources, 50K ESC secrets, 5B Neo tokens, 1M workflow minutes, 2M Insights resources. That's a wide enough range that the sliders are a cubic curve rather than linear, and the number input beside each one takes anything up to 10x its ceiling
* Past 3,000 IaC resources the estimate greys back and a note offers volume pricing
* `layouts/page/pricing.md` gets a "Cost estimation" section with the per-unit rates and the credit model, since an agent can't move a slider
* Adds a shared `.form-range` slider primitive to `theme/src/scss/shared/_forms.scss`
* Fixes `pulumi-tooltip` so the bubble sets its own typography instead of inheriting the trigger's
* The pricing cards' usage tooltip reads its rates from the `calculator:` block now - note: Neo tokens show as `3.00` rather than `3`
* Deletes the `pulumi-pricing-calculator` and orphaned `pulumi-value-calculator` Stencil components, plus the dead `price-toggle` and `pricing-trial` modules
* The comparison table's on-demand rows still carry these rates by hand so they can drift from the `calculator:` block. Collapsing them would mean templating the availability cells, so I've left that for another PR
* Screenshot above predates the meter changes and the move below the comparison table

N/A

### Related issues (optional)

N/A


layouts/page/pricing.html
```diff
@@ -100,6 +100,33 @@ <h2 class="text-2xl font-semibold tracking-tight m-0">{{ $edition.name }}</h2>
...
+                    <a href="#calculator-{{ $edition.id }}" data-track="pricing-card-{{ $edition.id }}-estimate" class="inline-flex items-center gap-1.5 text-violet-primary hover:underline">
+                        {{ partial "icon.html" (dict "name" "phosphor/calculator" "class" "size-4") }}
+                        Estimate your cost
+                    </a>
+                    {{ end }}
+                    <a href="#compare" data-track="pricing-card-{{ $edition.id }}-compare" class="inline-flex items-center gap-1.5 text-violet-primary hover:underline">
+                        {{ partial "icon.html" (dict "name" "phosphor/arrow-down" "class" "size-4") }}
+                        View all features
+                    </a>
+                </div>
```
cnunciato:
```suggestion
                <div class="mt-auto pt-2 flex flex-row md:flex-col lg:flex-row xl:flex-col xxl:flex-row items-start gap-4 text-sm">
                    {{ if $priced }}
                    <a href="#compare" data-track="pricing-card-{{ $edition.id }}-compare" class="inline-flex items-center gap-1.5 text-violet-primary hover:underline whitespace-nowrap">
                        {{ partial "icon.html" (dict "name" "phosphor/arrow-down" "class" "size-4") }}
                        View all features
                    </a>
                    <a href="#calculator-{{ $edition.id }}" data-track="pricing-card-{{ $edition.id }}-estimate" class="inline-flex items-center gap-1.5 text-violet-primary hover:underline">
                        {{ partial "icon.html" (dict "name" "phosphor/calculator" "class" "size-4") }}
                        Estimate your cost
                    </a>
                    {{ end }}
                </div>
```

cnunciato (approved):
Love it. My only (very non-blocking, thus approving) suggestion would be to consider positioning features first (just 'cause value > cost), and maybe fitting both links on one line, and stacking when necessary — something like this, maybe:

https://github.com/user-attachments/assets/8d87efb3-a6ae-4033-ab7f-4b261bf76503

Left this as a suggestion, but up to you. Thanks again for knocking this out!

layouts/page/pricing.html
jeffmerrick: @cnunciato Good call, that looks way better! I just made them `flex-wrap` so they'll stay side-by-side when there is enough room.

PR #20761, opened 2026-08-07
Let authors embed event and blog post cards in a post body

### LINKED ISSUE #20699: Allow embedding events/posts in blog posts
Rather than using CTAs, allow authors to embed event (and other blog post?) cards directly in blog posts.
Depends on: https://github.com/pulumi/docs/pull/20665

jeffmerrick:
Screenshot 2026-08-07 at 9 57 05 AM

### Proposed changes

Adds a `blog/card` shortcode so authors can embed the real event and blog post tiles anywhere in a post body, instead of hand-writing a `blog/cta-card` that restates a title, date, and blurb and then goes stale — a card derives everything from the target page, so an embed picks up a retitled session on its own and an event card flips Register → Watch once a recording lands. A card is always full width and takes exactly one path, which is its only parameter. Blog posts render as the wide card (text left, square image right), which moved out of `blog/card/list-row.html` into a shared `blog/card/wide.html` so the homepage feed's card view and in-body embeds can't drift — rendered homepage output is unchanged. Converts the two event CTA cards in the yolo-mode sandboxing post to real event cards, and documents the shortcode for authors in `BLOGGING.md` and for agents in `AGENTS.md`.

### Related issues (optional)

Fixes #20699


cnunciato (approved):
Nice!

PR #20764, opened 2026-08-07
List related events alongside the blog posts on release pages

jeffmerrick:
Screenshot 2026-08-07 at 11 19 58 AM

### Proposed changes

The card at the foot of a release page can now carry a second list — related events — under its own overline, in the same compact row style as the posts above it, and the August 2026 release page uses it to point at the two upcoming HCL/Terraform workshops. Rows come from a new `events/list-row.html` (the events counterpart of `blog/card/list-row.html`, sharing its columns so the two lists line up) and carry the cards' Register/Watch CTA as a quieter ghost-primary button. An event stays listed while it's still upcoming — per `events/upcoming.html`, the same query the events list and homepage use — and afterwards for as long as it has a recording, in which case its row flips itself to Watch; one that ran with nothing to watch drops off, so the card can't advertise a date that has passed. Since the card is no longer blog-only, `releases/blog-list.html` becomes `releases/feed-list.html` and the frontmatter key `blog_section` becomes `feed_section` (`title` → `posts_title`) on both release pages; the `#from-the-blog` anchor is unchanged.

N/A

### Related issues (optional)

N/A


cnunciato (approved):
Nice! Looks great.

PR #20795, opened 2026-08-10
Generate the ESC context built-in properties reference from the API schema

### LINKED ISSUE #20392: Generate the ESC `context` built-in properties reference from the API schema
> [!IMPORTANT]
> **Blocked — on a different thing than originally stated.** `GET /api/esc/context/schema` has since shipped ([pulumi/pulumi-service#46402](https://github.com/pulumi/pulumi-service/issues/46402)) and is live on `api.pulumi.com`, so the 404 risk is gone. But it returns property names and types with **no per-property descriptions** — only a single top-level one. The hand-written lists this issue would retire *do* carry a description per field, so generating from the schema as it stands would replace good docs with an empty-description property list.
>
> Blocked on [pulumi/pulumi-service#47609](https://github.com/pulumi/pulumi-service/issues/47609), which adds them. The same gap blocks step 3's `configuring-oidc` half: that page's `subjectAttributes` allow-list is already an enum in `data/esc_schemas.json` (identical across all six login providers, and it matches the page exactly), but the enum supplies attribute **names** only — the descriptions would have to come from the context schema. One upstream fix unlocks both renders.

### Background

The ESC built-in `context` object (`context.pulumi.user.login`, `context.pulumi.organization.login`, `context.pulumi.token.{type,name,team}`, `context.pulumi.openDuration`, plus the evaluator-injected `currentEnvironment.name` / `rootEnvironment.name`) is currently documented by hand, in two independent places. Those lists have drifted from the engine — see the companion content-fix issue.

The Pulumi Cloud API is gaining `GET /api/esc/context/schema`, which returns a JSON schema for the object. It's derived server-side from the same declaration that produces the values, so it can't fall out of date.

We already have the machinery to consume this: `scripts/fetch-esc-schemas.sh` pulls every provider and rotator schema from the API, and `.github/workflows/esc-update-schemas.yml` runs it nightly and opens a PR when `data/esc_schemas.json` changes. This issue extends that to the context schema so the built-in properties reference maintains itself.

### Work

1. **Fetch the context schema** in `scripts/fetch-esc-schemas.sh`.

   Note this does *not* drop into the existing loop. That loop iterates `for kind in providers rotators`, fetching a list and then `{kind}/{name}/schema` per entry. The context is a **singleton** — there is no list endpoint and no name — so it needs its own `fetch` call and its own key in the output document, e.g.:

   ```json
   {
     "providers": { ... },
     "rotators": { ... },
     "context": { ... }
   }
   ```

   One wrinkle confirmed against the live endpoint: unlike the provider/rotator endpoints, which return the schema document directly, this one wraps it as `{"schema": {…}}`. Unwrap `.schema` before assigning it to the `context` key so all three keys are shaped alike, and fail loudly if that key is missing rather than defaulting to `{}` — a silent empty object would render an empty reference.

2. **Render the built-in properties table from it**, similar to how `layouts/partials/esc/property-rows.html` renders provider schemas today.

3. **Retire the hand-written lists** in `content/docs/esc/concepts/interpolations-and-references.md` and `content/docs/esc/guides/configuring-oidc/_index.md` once the generated table covers them. Note that `configuring-oidc` also carries security rationale prose (why `pulumi.token.name` is excluded from subject attributes) that is **not** in the schema and must be kept.

### One thing to decide — answered

The fetch script runs authenticated, as `pulumi-bot` via OIDC. The endpoint is specified to return a **static** shape — every field that can appear, not the subset the calling credential produces — so what the bot captures should be canonical. Worth a quick sanity check on the first run that the response doesn't vary by caller; if it somehow does, we're snapshotting pulumi-bot's context rather than the real reference, and the endpoint needs revisiting rather than the script.

**Checked, and it holds.** Called with a personal token, the response still includes `pulumi.token.team`, `pulumi.token.name`, and `pulumi.openDuration` — fields a personal token doesn't populate. The shape is the full declaration, not the caller's subset, so whatever pulumi-bot snapshots nightly is the real reference. No script change needed for this.

### Why

Adding a context field today means coordinated hand-edits across two docs pages and two test fixtures in three repos, and the evidence is that people miss them. After this, the docs half updates itself through a pipeline that already exists.

jkodroff:
The ESC `context` object was documented by hand in two places that had drifted from the engine. `GET /api/esc/context/schema` now returns a description on every property, so the reference generates itself.

## Changes

- **Fetch** the context schema into `data/esc_schemas.json` alongside the provider and rotator schemas, so the existing nightly workflow keeps it current.
- **New Built-in Properties concepts page** holding `context`, `environments`, and `imports`, with `context` rendered from the schema. Aliases moved with it.
- **OIDC page**: the `subjectAttributes` allow-list is generated from the shared enum, each attribute linked to its description; the default token claims now name the context property each is built from. The `pulumi.token.name` security rationale is unchanged.
- **Determinism fix**: `fetch-esc-schemas.sh` sorts the schema's string arrays, so the nightly refresh stops opening no-op PRs when the API returns them in a different order. This is why the data diff is larger than the new key alone.

Two build-time drift alarms warn if the login providers stop agreeing on the allow-list, or if an attribute leaves the context schema.

Closes #20392


jkodroff:
@claude #update-review

CamSoper (approved):

PR #20745, opened 2026-08-06
Clarify that creator grants apply to machine tokens

jkodroff:
The RBAC docs described creator grants as something only a user receives, so it was ambiguous who holds the automatic Stack Admin grant when a stack is created with an organization or team token.

## Changes

- `rbac/_index.md` — creator grants now cover machine tokens, not just people; adds how to remove the grant or change the stack owner
- `rbac/roles.md` — defines "creator grant" where **Stack permissions** used the term bare
- `access-tokens.md` — disambiguates "the stack creator" for a machine token

## Scope

This addresses the creator-grant accuracy portion of #20493 only. The service principal documentation requested in that issue is deliberately out of scope: the feature is gated behind a default-off LaunchDarkly flag (`46739-service-principals`), so it is not yet observable by customers. Details in a comment on the issue.

Refs #20493

PR #19944, opened 2026-06-29
Note that team tokens carry org-wide default permissions

jkodroff:
## Description

Documents that **team tokens also carry the organization's default Member permissions** — the baseline access configured in [Organization-wide role settings](/docs/administration/access-identity/rbac/roles#organization-wide-role-settings) — on top of the roles assigned to their team.

Previously the docs described a team token's permissions only as *the union of the roles assigned to that team*, which could mislead a reader scoping automation to a single team's resources (e.g. a team token can read every stack in the org when the org default stack access is set to **Read**).

## Changes

- `content/docs/administration/access-identity/access-tokens.md` — add an `info` note to the "What team tokens can do" section with a concrete example.
- `content/docs/administration/access-identity/rbac/_index.md` — extend the "Team tokens" principal bullet in "Where roles apply" to stay consistent.

## Notes

- Content-only change; `make lint` passes.
- Standard `{{% notes %}}` callout, dark-mode safe.

Fixes #19942

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #20798, opened 2026-08-10
Document the creator-grant caveat for machine tokens

jkodroff:
A token that creates a stack holds a permanent, additive creator grant of Stack Admin on it, so a long-lived automation token accumulates admin access to every stack it has created — and no role, org-wide setting, or tag-based rule takes that back. This is intended behavior, but it surprised a customer who expected an ABAC rule to confine a team token to matching stacks.

## Changes

- `rbac/_index.md` — Creator grants: the accumulation problem for machine tokens, and the recommended hand-off.
- `rbac/roles.md` — a warning callout under Tag-based rules (ABAC) that tag rules only *grant* and cannot confine a principal to the resources they match, plus a new "Stack ownership in automation" best practice.
- `access-tokens.md` — extends the existing creator-grant security bullet with persistence and the hand-off.
- `least-privilege/_index.md` — new "Hand off ownership of stacks your automation creates" subsection under CI/CD, and the Summary bullet.

Scoped to stacks only; makes no claim about ESC environments or Insights accounts. Complements #19944 and #20745.

Refs pulumi/pulumi-service#45326, #20493


CamSoper (approved):
Looks right, approved (or you can wait for @djgrove )

PR #20799, opened 2026-08-10
Fix outcome scraper decoding gh --jq output twice

CamSoper:
## The bug

`fetch_pinned_bodies()` in `scrape-review-outcomes.py` assumed `gh api --jq '... | @json'`
double-encodes and decoded twice:

```python
obj = json.loads(json.loads(line))   # gh emits ONE single-encoded object per line
except (json.JSONDecodeError, TypeError):
    continue                          # every line lands here
```

`gh` emits a single-encoded JSON object per line, so the inner `json.loads` returns a
`dict`, the outer one raises `TypeError`, the bare `except` swallows it, and the function
returns `[]` for **every** PR. `scrape_pr()` reads an empty body list as "this PR has no
pinned review", so the failure renders as `status: no_review_data` rather than as an error.

## Blast radius

Present since the file shipped in #20153 (2026-07-09) — **32 days / 5 weekly digests**
(2026-07-13, 07-20, 07-27, 08-03, 08-10; all six scheduled runs went green).

Everything downstream of the scraper has been reporting zeros:

| consumer | effect |
|---|---|
| `scripts/weekly-digest/digest.py` → `collect_review_outcomes()` | `available: True` with an all-zero aggregate. It only degrades to `available: False` on a *crash*, and this never crashed. |
| Weekly digest (`.github/workflows/weekly-digest.yml`, Mondays 14:00) | 5 digests carried "0 fixed / 0 conceded / 0 disputed" as if the reviews had no outcomes. |
| Quarterly `--stats` tuning report (docs-review `SKILL.md`) | Same scraper, so any run in this window produced an empty per-verdict fix/concede/ignore table. |

What those digests should have carried, re-scraped with the fix over the same window:

- **137** review-labeled PRs with pinned reviews (19 genuinely had none)
- **333** classified human findings — 47 fixed, 10 conceded, 17 ignored-outstanding, 155 ignored-low-confidence, 104 unconfirmed-at-merge
- **190** style findings, **10** dispute events, **6** PRs merged with outstanding findings

Nothing was corrupted and no state was written — this is lost observability, not lost data.
Re-running the scraper over any past window now reconstructs it, because the pinned comment
is the ledger.

## Why the tests didn't catch it

`test_fetch_pinned_bodies_filters_and_orders` built its fake `gh` output with
`json.dumps(json.dumps(row))` — the same wrong assumption as the decoder. The test
exercised the bug and passed. It's the only test that touches this boundary, and it
pinned the defect in place.

## The fix

1. `_decode_gh_json_line()` decodes once (gh's real shape) and falls back to the
   double-encoded form, so a future wrapping `gh`/`jq` wouldn't regress.
2. The test fixture now emits gh's actual single-encoded output, with the double-encoded
   form kept as a separate compatibility case.
3. **Stop conflating "nothing decoded" with "no review data."** A payload where every line
   fails to decode now raises instead of returning `[]`. A PR with no pinned review still
   legitimately yields zero bodies; a wire-format change now surfaces as a failed digest
   section instead of a tidy row of zeros. That conflation is the reason this ran silently
   for five weeks, and it's the part worth keeping.

## Verification

- `pytest test_scrape_review_outcomes.py` — 15 passed (3 new)
- `--self-test` — passes
- Live: `--pr 20240` now scrapes 1 fixed + 1 ignored_low_confidence; previously `no_review_data`
- Live: `--closed-since 2026-07-09` scrapes 137 PRs; previously 0

Found while building a benchmark of the `#update-review` lane, which needed real dispute
base rates from this scraper.

### DIFF
--- a/.claude/commands/docs-review/scripts/scrape-review-outcomes.py
+++ b/.claude/commands/docs-review/scripts/scrape-review-outcomes.py
@@ -39,7 +39,11 @@ like any outstanding finding, so they ARE outcome-classified.
 This is a telemetry READER, never a gate: unparseable or legacy comment
 formats degrade to `parse_confidence: "low"` (counts-only) or
 `status: "no_review_data"`, and the aggregate reports how often that happened
-so silent degradation stays visible.
+so silent degradation stays visible. One case is NOT a degradation: a comment
+payload where nothing decodes at all means the wire format moved, not that the
+PR went unreviewed. That is `status: "decode_failed"` with its own aggregate
+column (`prs_decode_failed`) — a window scrape still returns every other PR's
+data, and `--pr` exits nonzero.
 
 Usage:
   scrape-review-outcomes.py --pr 20123 [--repo owner/repo]
@@ -169,6 +173,15 @@ def _decode_gh_json_line(line: str) -> dict | None:
     return obj if isinstance(obj, dict) else None
 
 
+class CommentDecodeError(RuntimeError):
+    """No line of a PR's comment payload decoded — the wire format moved.
+
+    Distinct from "this PR has no pinned review", which is an ordinary,
+    expected result. Raised per PR; `scrape_pr` converts it into a
+    `decode_failed` record so one unreadable PR can't abort a window scrape.
+    """
+
+
 def fetch_pinned_bodies(repo: str, pr: int) -> list[str]:
     """Return the bodies of every CLAUDE_REVIEW N/M comment, ordered by N.
 
@@ -196,10 +209,10 @@ def fetch_pinned_bodies(repo: str, pr: int) -> list[str]:
     # payload where NOTHING decoded is a different thing entirely -- the wire
     # format moved -- and must not be reported as "this PR has no review data".
     # That conflation is exactly how this went unnoticed: the decoder assumed
-    # the wrong encoding, every line raised, the bare except swallowed it, and
-    # 100% scrape failure rendered as a tidy row of zeros.
+    # the wrong encoding, every line raised, the except clause swallowed it,
+    # and 100% scrape failure rendered as a tidy row of zeros.
     if lines and undecodable == len(lines):
-        raise RuntimeError(
+        raise CommentDecodeError(
             f"could not decode any of {len(lines)} comment records for {repo}#{pr}; "
             f"`gh api --jq '... | @json'` output is not in a recognized form. "
             f"First line: {lines[0][:120]!r}"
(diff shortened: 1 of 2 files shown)


CamSoper:
@claude #update-review

Both findings addressed in e7e5aa6. Detail on each, including one place I went a different way than suggested.

**🚨 [L201] — `RuntimeError` aborts the whole `--closed-since` window.** Correct, and the sharper version of the point is that it reproduces the exact failure the PR is fixing one level up: `check=True` in `digest.py` turns any nonzero exit into `{"available": False}`, so a single unreadable PR would mute the outcomes section entirely. The raise itself stays — "nothing decoded" really isn't "no review data" — but it's now a typed `CommentDecodeError` that `scrape_pr` converts into a `decode_failed` record and keeps going.

It also gets its own `prs_decode_failed` column rather than being counted with `prs_no_review_data`, which is the part I'd have gotten wrong following the suggestion literally: `aggregate` treats *any* non-`scraped` status as no-review-data, so returning the new status without a new column would have rebuilt the same conflation in a different spot.

**One deliberate divergence.** The suggestion was to keep the raise on the `--pr` path so a traceback surfaces there. I split it by what each path has to lose instead: `--pr` exits nonzero (the record still prints, with `detail` — no partial result worth protecting), and the window path exits **zero** with a stderr warning. A nonzero exit on the window path would discard 130-odd good records to report that one was unreadable, and `digest.py` would mute the section — which is the finding again. The count rides in the aggregate where the digest can render it.

**⚠️ [L199] — "bare except".** Right, and it's not a nitpick: the deleted lines are `except (json.JSONDecodeError, TypeError):`. Calling it bare misattributes the cause, and the cause is the interesting part — the handler was targeted at exactly the exception the wrong assumption produced, which is why it looked deliberate and survived review. Comment reworded, slightly differently from the suggestion so it says that rather than just dropping the word.

**💡 [L225] — nothing in CI runs these tests.** Confirmed, and it's the finding with the longest tail: the original fixture double-encoded, so it validated the bug rather than `gh`'s output, and nothing would have caught that even if someone had noticed. Leaving it out of scope here, but I added the `decode_failed` case to the script's embedded `--self-test` as well as to pytest, since `--self-test` needs no test runner and is the more likely thing to get wired up first.

Verified live after the change: `--pr 20240` scrapes clean, and `--closed-since 2026-08-01` returns 59 scraped / 7 no-review-data / **0 decode-failed** across 66 PRs. 17 tests pass, `--self-test` passes, `make lint` clean.

PR #20801, opened 2026-08-10
Hold an evidence-spine floor on review refreshes, and move the lane to opus-5 medium

CamSoper:
Makes it impossible for a review refresh to publish without its evidence spine, and moves the re-entrant lane to `claude-opus-5 --effort medium`.

The floor is the larger half. The model change reduces how often the failure happens; the floor stops it happening at all — and it's what makes a later move back to Sonnet safe to evaluate on its merits rather than on which model drops the trail least.

## The bug

`pinned-comment.sh upsert` is a transport. It splits the body it's handed, stamps markers, and PATCHes the existing comments in place. It fetches those comments' **IDs, never their bodies** — so nothing in the system compares what's about to be published against what's already there. A refresh that re-renders the review without its 🔍 Verification trail publishes cleanly and the job goes green.

On `claude-update.yml` that loss is permanent. The lane does a fresh shallow checkout and runs only Vale, so there are no claims artifacts to rebuild from, and the comment it just overwrote was the only copy. Later refreshes merge from the damaged body, so it compounds.

Campaign [`2026-08-10-update-lane-model-config`](https://github.com/pulumi/docs-review-benchmarks) measured it at **1 in 6 chained refreshes** on the shipping configuration and **3 in 6** at `--effort low` — one of those collapsing a 40-line trail to 11 lines and then holding exactly 11 through the next refresh, because the floor it was checked against had moved down with it.

## The fix

`splice-spine.py` restores the three sections that describe the review's own *investigation* rather than the findings' *state*, when a re-render shrank them:

| section | shrink test |
| :--- | :--- |
| 🔍 Verification trail | fewer parsed trail records than the prior body |
| Investigation log | fewer of the 8 mandatory bullets |
| 📊 Editorial balance | under half the prior section's bytes |

Growth is still allowed. This enforces the render contract's existing **"may add, never drop"** instead of hoping for it — and it closes a hole in the old check, which was relative to the *prior* body and so scored an already-collapsed trail as passing.

**It runs inside `cmd_upsert`,** because that's the only point in the system holding both bodies at once, and it runs *before* `split_body` — so it never sees the continuation-`<details>` artifacts that make a naive fetch → concatenate → re-upsert compound them (the trap documented at `claude-update.yml:506`).

**Unconditionally.** The model composes its own `pinned-comment.sh upsert` command and the Bash allow-list is a prefix match, so a `--spine-floor` flag could simply be omitted — the same reasoning that already makes the ✏️ marks workflow-written. `SPLICE_SPINE=0` opts out and doesn't match the allow-list pattern, so it's reachable by a human or a workflow step and not by the model.

**Scoped by capability, not lane name.** `cmd_upsert_validated` delegates to `cmd_upsert`, so this also runs on the composer lane — where it can be *wrong*: that lane recomposes the trail from `.verified-claims.json` against the current diff, so if an author force-pushes a smaller change and re-requests review, a shorter trail is correct there. Restoring the old one would inject records for lines that no longer exist. The gate tests for claims artifacts: present → the caller owns the trail, stand down; absent → the pinned comment is the only copy, hold.

**Safety posture.** Every ambiguous case no-ops rather than guessing, and any internal failure exits 0 and publishes as rendered. This runs on a body nobody has seen yet, so a wrong repair would corrupt a good review, and a repair pass must never be why a review fails to publish.

## It also measures

A **Report evidence-spine floor** step prints the verdict and raises a `::warning::` on any restore.

That's not decoration. The floor runs from the model's own Bash tool, and claude-code-action doesn't echo tool output into the job log — so without this step a run where the floor held and a run where it never executed look identical. (That bit me during the fork test: correct output, no way to tell what produced it.)

More usefully: **every warning is a live instance of the model dropping the evidence spine** — the rate the campaign had to measure synthetically. The lane now measures it on real traffic for free, and a run of warnings is the argument for revisiting the model config rather than leaning harder on the repair.

## The model change

`claude-update.yml` ran `--model claude-sonnet-5` with no `--effort` — not a measured configuration, but residue: the 2026-07-24 battery shipped `--effort low` on a cost/latency win, and a 2026-07-30 fork A/B reversed the flag six days later when it turned out to strip the spine. Models had never been compared on this lane.

4 arms × 3 reps × 2 chained refreshes, scored deterministically:

| arm | spine intact | trail lines (of 40) | validator viol. | output tok |
| :--- | ---: | ---: | ---: | ---: |
| sonnet-5, no `--effort` (**what ran here**) | 0.833 | 40 | 1.833 | 355,907 |
| opus-5, no `--effort` | 0.833 | 39.8 | 1.167 | 220,511 |
| **opus-5 `--effort medium`** | **1.0** | 40.7 | 2.333 | 154,561 |
| sonnet-5 `--effort low` | 0.5 | 29.5 | 5.667 | 137,513 |

Cost: **$8.14 vs $6.54** — 1.25×, about **$0.27 per refresh**, permanent. (An earlier version of this description said the premium would invert to a ~17% saving on 2026-08-31 when Sonnet 5's introductory rate lapsed. That rate is now Sonnet 5's standing price, so the crossover never happens.) At this lane's observed ~30 runs/week that's on the order of $400/year.

**The quality signal is one cell: 6/6 against 5/6, at n=6.** It isn't statistically separated and no affordable number of reps would separate it. The case is the asymmetry, not the significance — and with the floor in place the asymmetry mostly goes away, which is the point of shipping both together.

`claude.yml` (the ad-hoc `@claude` lane) moves to the same model, so a contributor gets the same capability whichever way they mention `@claude`. Flagged in the comment as an unmeasured job shape — it tracks the decision, it isn't justified by it.

Also resolves the campaign's biggest open caveat: `claude-code-action` does nothing with `--effort` (it passes `claude_args` straight through, and the strings `high`/`xhigh` appear nowhere in its source), and the Agent SDK it pins documents `'high' — Deep reasoning (default)` twice. **Unset was the API's `high`**, so the control was the shipping configuration.

## Verification

| check | result |
| :--- | :--- |
| Unit tests (`test_splice_spine.py`) | 13/13 |
| Full docs-review script suite | 167 passed |
| **40 real published bodies, healthy** | 40/40 untouched (13 multi-part) |
| **40 real published bodies, gutted trail** | 40/40 restored exactly |
| `validate-pinned.py` on repaired vs original | identical status, 15/15 |
| End-to-end `upsert` with stubbed `gh` | 0 → 40 published, caller's file untouched, re-split to 2/2 and round-trips |
| Capability gate | fires without claims artifacts, stands down with them |
| `make lint` | clean |

Multi-part bodies matter most — they exercise reassembly of the continuation-`<details>` artifacts `split_body` inserts.

### Fork smoke test (CamSoper/pulumi.docs#234)

| # | scenario | result |
| ---: | :--- | :--- |
| 1 | Initial review (composer lane) | Posted; 19 records, balance, 8 log bullets |
| 2 | Refresh at `opus-5 --effort medium` | Spine 19→19, no spurious edits |
| 3 | Refresh at `sonnet-5 --effort low` | Spine held |
| 4 | Refresh, trail inflated to 44 | Floor ran, reported `noop` correctly |
| 5 | Refresh with a forced trail drop | **Restored 0 → 44, warning fired** |

Run 5's report, from the real job log:

```json
{ "restored": [ { "section": "🔍 Verification trail",
                  "prior": 44, "new": 0, "unit": "trail records" } ],
  "status": "restored" }
```

Published body: 44 records, all 25 injected records byte-preserved, pointer prose gone, balance and log intact. Runs 2–4 are the negative cases, and they matter more — they show it doesn't touch healthy reviews.

`cam-fork/master` has been reset to the incumbent.

## Follow-ups this doesn't do

- **Doesn't fix disputes.** All arms were indistinguishable across 56 dispute cells, and the ledger agrees: 137 PRs, 10 dispute events, 10 conceded, **0 held**. That's the lane conceding whenever nothing settles the question — an `update.md` Case 2 change, not a model choice.
- **Doesn't fix the other contract failures.** ✏️ marks carried over in 6/6 cells; correct ✅ moves land 0.167–0.667; quotes survive verbatim under a third of the time. No arm fixes those.
- **Doesn't touch the composer lane's `--effort low`** — that was a measured 2026-07-24 decision, and `#new-review` routes through it (`claude-new.yml` has no model of its own; it dispatches `claude-code-review.yml`).
- **Worth re-running the compose surface with the floor in place**, to see whether the arms converge enough to drop back to Sonnet as its own one-axis change.

### DIFF
--- a/.claude/commands/docs-review/scripts/pinned-comment.sh
+++ b/.claude/commands/docs-review/scripts/pinned-comment.sh
@@ -367,7 +367,18 @@ cmd_upsert() {
     # Never fatal: splice-spine.py exits 0 on any internal failure and leaves
     # the body as rendered. A repair pass must not be the reason a review fails
     # to publish. Operates on a COPY so a caller's file is never mutated.
-    if [[ "${SPLICE_SPINE:-1}" != "0" ]]; then
+    # Scoped by CAPABILITY, not by lane name. The floor is only sound where the
+    # caller could not have re-derived the trail: claude-update.yml does a fresh
+    # shallow checkout and runs only Vale, so a shrunken trail there is always a
+    # loss. The composer lane (claude-code-review.yml, which reaches this
+    # function through cmd_upsert_validated) recomposes from
+    # `.verified-claims.json` against the CURRENT diff — if the author force-
+    # pushed a smaller change and re-requested review, a shorter trail is
+    # CORRECT there, and restoring the old one would inject records for lines
+    # that no longer exist. So: claims artifacts present => the caller owns the
+    # trail, stand down. Absent => the prior comment is the only copy, hold.
+    if [[ "${SPLICE_SPINE:-1}" != "0" ]] \
+       && [[ ! -f .verified-claims.json && ! -f .candidate-claims.json ]]; then
         local script_dir splicer
         script_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
         splicer="$script_dir/splice-spine.py"
--- a/.claude/commands/docs-review/scripts/pinned-comment.sh
+++ b/.claude/commands/docs-review/scripts/pinned-comment.sh
@@ -350,6 +350,43 @@ cmd_upsert() {
         fi
     fi
 
+    # Evidence-spine floor. This is the only point in the system that holds the
+    # old body and the new one at the same moment: everything above fetches
+    # comment IDs, not bodies, so nothing else can notice that a re-render
+    # dropped the 🔍 Verification trail. Measured on 2026-08-10, the update
+    # lane dropped it in 1 of 6 chained refreshes and published green.
+    #
+    # UNCONDITIONAL, deliberately. The model composes its own `pinned-comment.sh
+    # upsert` command and the Bash allow-list is a prefix match, so a
+    # `--spine-floor` flag could simply be omitted — the same reason the ✏️
+    # marks are workflow-written rather than model-written. The escape hatch is
+    # the env var, which does NOT match the allow-list pattern (that requires
+    # the command to begin `bash .claude/…`), so it is reachable by a human or
+    # a workflow step and not by the model.
+    #
+    # Never fatal: splice-spine.py exits 0 on any internal failure and leaves
+    # the body as rendered. A repair pass must not be the reason a review fails
+    # to publish. Operates on a COPY so a caller's file is never mutated.
+    if [[ "${SPLICE_SPINE:-1}" != "0" ]]; then
+        local script_dir splicer
+        script_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+        splicer="$script_dir/splice-spine.py"
+        if [[ -f "$splicer" ]]; then
+            local prior_file spliced_file
+            prior_file=$(mktemp)
+            spliced_file=$(mktemp)
+            fetch_pinned_bodies "$repo" "$pr" >"$prior_file" 2>/dev/null || true
+            cp "$body_file" "$spliced_file"
+            python3 "$splicer" \
+                --prior "$prior_file" \
+                --body "$spliced_file" \
+                --in-place \
+                --report "${SPLICE_SPINE_REPORT:-/tmp/splice-spine.json}" || true
+            body_file="$spliced_file"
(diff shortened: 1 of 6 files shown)


CamSoper:
@claude #update-review

PR #20810, opened 2026-08-11
Keep the 📋 Triaged section's <details> wrapper

CamSoper:
The 📋 Triaged verifier findings section is meant to render collapsed, behind a summary line that tells a reader what it *is* — findings the reviewer checked and concluded were verifier noise, not things to act on:

```markdown
### 📋 Triaged verifier findings

<details>
<summary><em>I double-checked these and realized they weren't real findings — click to expand</em></summary>
…
</details>
```

`compose-review.py:render_triaged` always emits that. The **editorial pass** then swaps the `_No triaged findings._` placeholder for the real `**Spurious:**` / `**Mis-sourced:**` bullets — and can take the wrapper out with it. The section then renders expanded and unexplained, so triaged **non**-findings sit open on the page looking exactly like the real findings above them. That's the opposite of what the triage was for.

## Why nothing caught it

- **No rule.** `validate-pinned.py` had nothing covering the wrapper — it only counted the bullets.
- **`splicer.py`'s canonical block only lands when the section is missing *entirely*** (via `mandatory-h3-order`), not when it's present-but-unwrapped.
- **`strip-empty-triaged.py`** keys on the `_No triaged findings._` placeholder — the exact thing the editorial pass had just removed.

Found on pulumi/docs#20781, which has rendered that way since its first review on 2026-08-09.

## The fix

`triaged-details-wrapper` (schema v20) plus a surgical splicer, so the composer lane repairs it before publish instead of soft-flooring.

**The splice is purely additive.** It rebuilds the wrapper around the bullets and never reflows, reorders, or drops any of them — including the struck-through "this triage was wrong" form that isn't a bucket bullet. The triage prose is the valuable part; only the furniture was missing. It defers to the fallback lane if a `<details>` is already open in the section, because a partial wrapper is ambiguous and guessing would be worse than not fixing.

**The rule fires only when the section exists *and* has bullets.** 📋 is optional and absent from `MANDATORY_H3_SECTIONS`, and an empty one is `strip-empty-triaged.py`'s business. A false positive here would soft-floor a good review over cosmetics — worse than the defect.

## Verification

| check | result |
| :--- | :--- |
| New rule tests | 8/8 |
| Splicer tests | 25/25 |
| Full docs-review suite | 183 passed |
| **40 real published reviews** | 0 flagged, 27 correctly wrapped, 13 no such section |
| **Full validator vs `origin/master` on those 40** | byte-identical exit codes, 40/40 |
| Real defective body (#20781) | flagged → spliced → 0 remaining, 0 content lost |
| Workflow chain replicated at the CLI | validate → fix-me JSON → splicer → re-validate, emoji `line_ref` survives the JSON round-trip |

### Fork smoke test (CamSoper/pulumi.docs#235)

1. **Normal composer run** — review published, validator `rc=0`, no 📋 section (nothing was triaged, so `strip-empty-triaged.py` correctly removed the placeholder). No regression.
2. **Fault-injected run** — a fork-only prompt instruction forced an unwrapped section with two `**Spurious:**` bullets. Job log:

   ```
   splicer.py: applied surgical fixes for rule(s): triaged-details-wrapper
   ```

   Published body: wrapper present, both bullets preserved, `<details>` balanced, validator `rc=0`.

`cam-fork/master` has been reset to the incumbent.

## One note on how rare this is

I first measured it as 2 of 28 published reviews. That was wrong, and the second "case" is instructive: pulumi/docs#19777 is a docs-review meta PR whose own findings *quote* the review format, so a substring search for the heading matched in a body that has no such section. `find_section` requires a real column-0 H3. The real rate is **1 of 28**, and there's a regression test for exactly that trap.

### DIFF
--- a/.claude/commands/docs-review/scripts/splicer.py
+++ b/.claude/commands/docs-review/scripts/splicer.py
@@ -88,6 +88,7 @@ EMPTY_FORMS = {
 # bucket-bullet prefix prepends run last so trail records are stable.
 APPLY_ORDER = [
     "mandatory-h3-order",
+    "triaged-details-wrapper",
     "internal-link-existence",
     "shortcode-existence",
     "verified-claims-trail-faithful",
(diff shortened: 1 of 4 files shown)


CamSoper:
@claude #update-review

Both findings fixed in `f379a5f`.

**🚨 [L621] — the three splicer tests never ran.** Correct, and it was worse than a missing registration: they were appended *after* the `if __name__ == "__main__"` guard, so `main()` executed before their definitions were bound. The standalone runner reported the same count as before the PR. Moved the definitions above the `TESTS` list and registered them — standalone and pytest now both report 29 (was 25 vs 28).

**⚠️ [L359] — the guard was one-sided.** Also correct, and this is the one I'd have most regretted shipping. Guarding only on the opening tag meant a section with an orphan `</details>` took the wrap path and produced a block that closes early, leaves the trailing bullets expanded, and then *passes* re-validation because a wrapper and a summary are both present. Silent corruption is strictly worse than the cosmetic defect the rule exists to catch, and it's the opposite of the deferral posture the docstring claims. Widened to `startswith(("<details>", "</details>"))` as suggested, plus the fourth test — which also asserts the body comes back byte-identical when it defers, since "doesn't corrupt" is the property that matters there.

I left `check_triaged_details_wrapper` alone deliberately. It correctly raises on that orphan case already (no opening tag ⇒ violation), so the malformed body still gets flagged and now routes to the model lane instead of the splicer. Adding balance-checking to the validator would widen its false-positive surface on an optional section for no gain.

Re-verified: 184 tests pass, 0 of 40 real published reviews flagged, and #20781 still repairs to 3 bullets with the wrapper restored.

PR #20811, opened 2026-08-11
Fix three silently-dead review lanes and sweep pipeline tech debt

CamSoper:
Maintenance pass over the pre-merge review pipeline and the daily content review. Most of it is small, but three components turned out to be reporting green while doing nothing.

## Three silently-dead components

**`blog-review-index` and `claims-reverify` have never run on schedule.** Both gate on `vars.<NAME>_COUNT != '0'`. An unset variable is `null`, and GitHub coerces a Null-vs-String comparison to numbers (`null` → 0, `'0'` → 0), so the gate evaluates `0 != 0` = false. "unset -> on", the default both workflow headers document, actually meant **off**.

It looked healthy because `workflow_dispatch` short-circuits the gate on its first clause — every S3 object under `blog-review/` came from a manual dispatch. The three lanes are otherwise structurally identical (same gate shape, same `environment: production`, same holiday job); the only difference is that `CONTENT_REVIEW_COUNT` happens to be set, which is why that one works.

Verified on a fork with no org variables, so repo config is ruled out:

| gate | variable | job |
|---|---|---|
| `vars.X != '0'` | unset | **skipped** |
| `format('{0}', vars.X) != '0'` | unset | runs |
| `format('{0}', vars.X) != '0'` | `'0'` | skipped |
| `format('{0}', vars.X) != '0'` | `'5'` | runs |

`format()` forces a string, so the unset case becomes a real `'' != '0'` compare. Applied to all three lanes — `review-existing-content` only works today by luck, and deleting its variable would silently stop it.

**The Hugo build pre-step has never run on any PR.** `templating_changed` grepped `FILES_LIST` — the human-readable `"  - path (+n/-m)"` rendering — with a `^`-anchored path pattern that its two-space bullet prefix can never match. Now built from a separate bare path list.

**The oversized-skip label cleanup was unreachable**: it required `has_write_access == 'true'`, but the step producing that output only runs when `skip_reason` is empty. A stale `review:error` beside `review:oversized` was never cleared.

## Correctness

- `select-articles` `BOT_AUTHORS` was missing `"Pulumi Bot"` and `"workprentice[bot]"`, so bot commits reset the page staleness clock — **637 of 996** `content/docs` paths.
- `reverify-claims` wrote a 2-field stub over the full ledger record when the local cache was empty, destroying `status`/`reviewed_at`/`attempts`/`tier`/`score`.
- `claude-update` ran three post-model steps on issue mentions, gating on `pr_number != ''` where `is_pr == 'true'` was meant.
- The content-review compose-failure fallback dropped the `<!-- LINT-RESULT -->` placeholder, so the lint result never reached the PR body.
- `blog-review` overwrote `index/_summary.json` after a zero-object sync; `record-findings` could fail green.
- Three comment lookups read only the first page of 30.
- `FILE_COUNT` used `.files | length`, which the API caps at 100 (verified against #20787: 256 files, 100 returned); now `.changedFiles`.

## Observability and hygiene

- Six pre-step scripts have no `safe_main` wrapper, so `2>/dev/null` discarded the only trace of a crash.
- 14 generated artifacts were missing from `.gitignore`; the update lane runs with `contents: write` and grants the model `Bash(git:*)`.
- Removed the dead `upsert-validated` subcommand (88 lines) — the workflow drives validate/splice/re-validate/upsert itself.
- Dropped `content/learn/` (doesn't exist) from routing, regexes and docs; fixed stale saml/scim paths, a `Sonnet 4.6` label, two hardcoded runner paths, and the oversized label description.
- Reconciled `ci.md` ↔ `output-format.md`: the stub-promotion mapping was stated twice with a *different* verdict missing from each copy. Added the live `📋 Triaged` section and the readthrough artifact to the inventories that omitted them.

## Test coverage

20 test suites existed and were green, but nothing ran them — `make test` covers only the example programs. Added `scripts/test-review-pipeline.sh` (pytest + standalone harnesses + every `--self-test`), a `make test-review-pipeline` target, and a path-filtered workflow. Also fixed `test_select_posts.py`, which had been erroring under pytest collection.

The new suite is negative-tested both ways: the runner goes red on an injected failure, and the new `BOT_AUTHORS` regression test fails on the pre-fix set, naming both missing identities.

## Operational changes made alongside this PR

Because the gate fix makes unset genuinely mean on, both revived lanes needed an explicit decision rather than a default. Set in `pulumi/docs` variables:

- **`BLOG_REVIEW_COUNT=0`** — parked. Fixing the gate would otherwise start ~5 Opus post-reviews/day on a lane that has never run, and the noindex decision process this index feeds doesn't exist yet.
- **`CLAIMS_REVERIFY_COUNT=25`** — on. Dry-run against the live claims index first (86 snapshots, 63 volatile entities), and the day rotation verified to cover 63/63 in exactly 3 nights with no gaps and a clean cycle repeat.

Both recorded in the workflow headers so the values aren't mystery state later.

## Deliberately not included

- Copying the `markdown-syntax-findings` merge into the update and triage Vale lanes. Real gap, but it changes what findings those lanes produce — wanted separately from a maintenance PR.
- `references/update.md`'s `gh pr diff --range`, which is not a real flag, so the delta-diff path silently falls back to a full diff on every run. Behavior change, needs a decision.
- The five duplicated ledger-bucket resolve blocks and the ~266 byte-identical lines between the two selectors. Both want a shared helper, which is refactoring rather than maintenance.
- ~15 unreferenced step `id:` attributes: pure diff noise on a safety-critical PR.

## Verification

`make lint`, `actionlint` (no new findings), and all 21 suites green. Every behavioral claim above was checked against live data or a live run rather than inferred.


CamSoper:
@claude #update-review

Pushed `fc1ab2722a` closing both ⚠️ findings from the last pass:
- `BUILD-AND-DEPLOY.md` §Makefile Targets now carries a `test-review-pipeline` row.
- `CONTRIBUTING.md` §oversized now names both budget axes (>15K lines **or** >150 files).

Please re-verify and move them to ✅ Resolved.

PR #20825, opened 2026-08-11
Reconcile the Terraform onramp comparison pages with the HCL reference

CamSoper:
### Proposed changes

Implements the recommendations from a content audit of the Terraform onramp corpus. The audit's finding was that the comparison pages describe Pulumi HCL **more confidently than the HCL reference docs do** — they were updated from launch messaging rather than from the reference page that owns the behavior. Each item below is a contradiction inside the corpus or against a primary source.

**P0 — the `cloud` block.** `comparisons/terraform/_index.md` said `backend`, `cloud`, and `required_version` are all "accepted and ignored". Per `languages-sdks/hcl/_index.md`, a `cloud` block is an *error*. A `cloud` block means the reader is on HCP Terraform — exactly the migrator this onramp targets — so the page promised their config runs unchanged while the engine hard-fails on the one block that defines their setup. Corrected and linked to the compatibility section.

**P1 — version floor.** `terraform-cloud.md` and `opentofu.md` said Pulumi CLI 3.235.0; the reference says 3.256.0. 21 releases apart, and a reader on 3.240 would set `runtime: hcl` and fail.

**P1 — absolute compatibility claims.** Replaced "superset of Terraform HCL", "no syntactic differences", and "fully OpenTofu compatible" with the reference's hedge plus a link to the documented exceptions. These absolutes are contradicted by our own reference (`cloud` errors, Terraform state files neither read nor written, provisioner `connection` blocks SSH-only) and are trivially falsifiable by a competitor. The audit flagged 3 instances; there were 9, across `terraform-cloud.md`, `opentofu.md`, `tacos.md`, and `spacelift.md`.

**P2 — `pulumi import --from`.** The migration guide used both `--from terraform` and `--from hcl` without explaining either, and nothing in the corpus documented which to use. **Verified against the CLI (3.256.0):** both work, and they differ meaningfully — `--from terraform` reads the state file on its own, while `--from hcl` also parses the `.tf` files in the project directory (it fails with "No Terraform files found" without them). That is the same split `convert-hcl.md` already documents for `pulumi convert`, so the existing usages were each contextually correct; the fix is to state the rule. The 10 existing `--from hcl` call sites are correct and unchanged.

**P2 — stale tense on a competitor claim.** "HashiCorp *is retiring*… organizations *are being migrated*" described an event that completed on March 31, 2026 (verified against HashiCorp's EOL notice), on a page framed around "right now". Recast to past tense, which reads stronger. Also date-stamped the RUM-pricing and 500-resource-cap assertions with `<!-- verified: 2026-08 -->` — competitor pricing is the fastest-rotting content here, and this marks what the next sweep should re-check. **This comment convention is new; happy to drop it if the team would rather not adopt it.**

**P3 — the language list.** The supported-language set was written six different ways across `comparisons/**`, dropping JavaScript in places and alternating `C#`/`.NET`; `terraform/_index.md` contradicted itself three paragraphs apart. This is the failure mode for agent-consumed content — a model retrieving three of these pages gets three different answers. Added a `pulumi-languages` shortcode (canonical set taken from `languages-sdks/_index.md`) and replaced 25 freehand instances.

### Verification

- `node ./scripts/lint/lint-markdown.js` — 1849 files, 0 errors (it caught a too-long `meta_desc` mid-change, since fixed)
- `yarn prettier --check` — clean on all changed files
- Shortcode rendering verified in an isolated Hugo 0.157.0 site across all three call forms, in prose and inside table cells, for both the HTML and markdown output variants
- All four heading anchors introduced were confirmed against Hugo's `github` auto-ID scheme

A full `hugo` build could not complete in this environment — the theme asset pipeline fails on a CSS fingerprint step, identically on a clean tree at this commit's base, so it's pre-existing and unrelated. All content parsed successfully (the build reaches the render phase), and the 25 shortcode call sites use only the three verified forms.

### Related issues (optional)

None.


CamSoper:
@claude 

Implement the L24 suggestion that doesn't pin a version, and the pre-existing L70 suggestion, then #update-review

CamSoper:
@claude implement your suggestion for L46, and let's just call L20 as settled as a non-load-bearing claim.

#update-review

cnunciato (approved):
Nice! Thanks.

PR #20831, opened 2026-08-12
fix `pulumi do` docs

tgummerer:
These docs were seemingly written with `--stateless` in mind, and are no longer correct.  Update them with stateful mode as default in mind.

### DIFF
--- a/content/docs/iac/cli/direct-resource-operations.md
+++ b/content/docs/iac/cli/direct-resource-operations.md
@@ -26,6 +26,11 @@ The `pulumi do` command provides direct operations on cloud resources through th
 - **Provider functions**: Read-only queries against cloud APIs (e.g., looking up a VPC, fetching an AMI).
 - **Resource operations**: Create, read, patch (update), and delete cloud resources.
 
+### Modes
+
+- Stateful: This is the default mode for `pulumi do`.  Resources created/updated using `do` in this mode will be recorded as snippets in the state file of your current project, and their lifetime will be tracked. If you are not currently in a project, they will be recorded in a global root project that's automatically created for `pulumi do`.
+- Stateless: This mode can be enabled using the `--stateless` flag. In this mode resources are not recorded anywhere, so it is usually good for one off operations, or testing.
+
 ### Command syntax
 
 ```
@@ -33,7 +38,7 @@ The `pulumi do` command provides direct operations on cloud resources through th
 pulumi do <package:module:function> [flags]
 
 # Resource operations
-pulumi do <package:module:type> <operation> [<id>] [flags]
+pulumi do <package:module:type> <operation> [<name>] [flags]
 ```
 
 The package, module, and type/function segments come directly from the provider schema. Pass `--help` at any level of the command tree to discover available subcommands.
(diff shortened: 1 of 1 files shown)


i-am-tom (approved):
Eng ✅

content/docs/iac/cli/direct-resource-operations.md
```diff
@@ -26,16 +26,23 @@ The `pulumi do` command provides direct operations on cloud resources through th
...
 ```
 # Provider functions
 pulumi do <package:module:function> [flags]
 
 # Resource operations
-pulumi do <package:module:type> <operation> [<id>] [flags]
+pulumi do <package:module:type> <operation> [<name>|<id>] [flags]
 ```
 
+The positional argument depends on the operation: `create`, `delete` and `patch` take the resource name, while `read` takes the provider-assigned resource ID. With `--stateless`, `create` takes no argument and `delete` and `patch` take the provider-assigned ID instead of the name.
```
cnunciato:
```suggestion
The `create`, `delete` and `patch` operations take the resource's [logical name](/docs/iac/concepts/resources/names), while the `read` operation takes the provider-assigned [physical ID](/docs/iac/concepts/resources/names). With `--stateless`, `create` takes no argument and `delete` and `patch` take the provider-assigned ID instead of the name.
```

content/docs/iac/cli/direct-resource-operations.md
```diff
@@ -26,16 +26,23 @@ The `pulumi do` command provides direct operations on cloud resources through th
 - **Provider functions**: Read-only queries against cloud APIs (e.g., looking up a VPC, fetching an AMI).
 - **Resource operations**: Create, read, patch (update), and delete cloud resources.
 
+### Modes
+
+- Stateful: This is the default mode for `pulumi do`. Resources created or updated in this mode are recorded as snippets in the state file of your current project, and their lifetime is tracked. Because they live in ordinary Pulumi state, they get the same benefits as program-managed resources: policy enforcement, drift detection with `pulumi refresh`, and references between resources. If you are not currently in a project, they are recorded in a global project named `default-global-project` (stored in your Pulumi home directory, with a stack named `default`) that's created automatically on first use.
+- Stateless: This mode can be enabled using the `--stateless` flag. In this mode resources are not recorded anywhere, so it's a good fit for one-off operations or testing.
```
cnunciato:
```suggestion
- **Stateful**: This is the default mode for `pulumi do`. Resources created or updated in this mode are recorded as snippets in the state file of your current project, and their lifetime is tracked. Because they live in ordinary Pulumi state, they get the same benefits as program-managed resources: policy enforcement, drift detection with `pulumi refresh`, and references between resources. If you are not currently in a project, they are recorded in a global project named `default-global-project` (stored in your Pulumi home directory, with a stack named `default`) that's created automatically on first use.

- **Stateless**: This mode can be enabled using the `--stateless` flag. In this mode resources are not recorded anywhere, so it's a good fit for one-off operations or testing.
```

content/docs/iac/cli/direct-resource-operations.md
```diff
@@ -108,10 +115,11 @@ Resource operations let you create, read, update, and delete cloud resources dir
 
 ### Create
 
-Creates a new cloud resource. Pass inputs via an input file. The CLI prompts for confirmation before creating.
+Creates a new cloud resource. Pass inputs via an input file, or set scalar inputs directly with per-input command-line flags. The CLI prompts for confirmation before creating.
```
cnunciato, line 119:
```suggestion
Creates a new cloud resource. Pass inputs via an input file, or set input values directly with per-input command-line flags. The CLI prompts for confirmation before creating.
```

cnunciato (commented):
A few suggestions, but leaving an approval to keep you unblocked. 

One thing to note, though, is that the CLI version I have does not seem to align with these docs, which seem to suggest this should work:

```bash
➜ pulumi do aws:s3:Bucket create "my-bucket"
error: get stack configuration: could not detect project stack path: no Pulumi.yaml project file found (searching upwards from /tmp/do). If you have not created a project yet, use `pulumi new` to do so: no project file found
```

content/docs/iac/cli/direct-resource-operations.md
tgummerer: Hrm this one isn't correct. We can't set e.g. maps/arrays through flags directly, only strings/numbers/bools.

PR #20837, opened 2026-08-12
add getting started guide

tgummerer:
Add a getting-started guide for `pulumi do` to explain the basic functionality.

### DIFF
--- a/content/docs/iac/cli/direct-resource-operations/get-started.md
+++ b/content/docs/iac/cli/direct-resource-operations/get-started.md
@@ -21,7 +21,7 @@ This guide takes you from zero to your first resource operations, and shows you 
 
 ## Prerequisites
 
-1. [Install the Pulumi CLI](/docs/install/), version v3.257.0 or later. Because `pulumi do` is in research preview and evolving quickly, we recommend the latest version.
+1. [Install the Pulumi CLI](/docs/install/), version 3.257.0 or later. Because `pulumi do` is in research preview and evolving quickly, we recommend the latest version.
 1. Credentials for a cloud provider. The examples use AWS, but any [Pulumi provider](/registry/) works the same way.
 
 ## Create your first resource
@@ -55,9 +55,18 @@ By default, `pulumi do` runs in stateful mode: the bucket is recorded as a snipp
 
 A bucket wouldn't be very useful without anything to put into it.  Let's create a bucket object in the bucket.
 
-To do that, we need to reference the newly created bucket.  We can do that either by creating a resource-file (in JSON format), and passing that via the `--resources-file` option.  The easiest way to create that file is by using the output of `pulumi do show-resources --json`. The names `pulumi do show-resources` shows can also be used to reference the resource directly, as we'll do in the example below.
+The object needs to reference the bucket you just created. `pulumi do` auto-assigns every tracked resource an identifier, derived from its name with hyphens converted to underscores — so `my-bucket` becomes `my_bucket`. Run `pulumi do show-resources` to list them:
 
-Further, inputs can be passed to `pulumi do` either via `--<input>` flags for scalar inputs, or through a YAML formatted file, that's passed with `--input-file`. In the following example we'll pass the bucket name through the YAML file, while passing the content through an `--<input>` flag.
+```bash
+$ pulumi do show-resources
+
+NAME       URN
+my_bucket  urn:pulumi:default::default-global-project::aws:s3/bucket:Bucket::my-bucket
+```
+
+You can also supply your own identifiers by passing a JSON file that maps identifiers to resource URNs with `--resources-file`; its entries take precedence over the auto-assigned ones.
+
+Inputs can be passed to `pulumi do` either via `--<input>` flags for scalar inputs, or through a YAML formatted file that's passed with `--input-file`. Inside an input file, reference another resource by its identifier with `${...}`. The following example passes the bucket reference through the YAML file, while passing the content through an `--<input>` flag:
 
 ```yaml
 # object.yaml
@@ -82,7 +91,7 @@ Created my-object (snippet 08de6783-0957-415d-98f8-e304399a1d09)
 
 ## Read, update, and delete
 
-Read the current state of any resource by its cloud provider ID. The id is shown in the outputs, or can be read using `pulumi state get <name>`.
+Read the current state of any resource by its cloud provider ID, which is shown in the `id` output when the resource is created.
 
 ```bash
 $ pulumi do aws:s3:Bucket read my-bucket-f998a37
@@ -151,6 +160,8 @@ $ pulumi do aws:ec2:getVpc --default
   ...
 }
 ```
+
+`--default` isn't a `pulumi do` flag — it's the `default` input of `getVpc`, passed with the same `--<input>` form used for resource inputs.
 
 ## Use it in scripts
 
--- a/content/docs/iac/cli/direct-resource-operations/get-started.md
+++ b/content/docs/iac/cli/direct-resource-operations/get-started.md
@@ -1,0 +1,170 @@
+---
+title_tag: "Get Started with pulumi do | Pulumi CLI"
+meta_desc: "Get started with pulumi do: create, query, update, and delete cloud resources directly from the Pulumi CLI without writing a program."
+title: Get Started with pulumi do
+h1: Get started with pulumi do
+menu:
+    iac:
+        name: Get Started
+        parent: iac-cli-do
+        identifier: iac-cli-do-get-started
+        weight: 1
+---
+
+{{% notes type="info" %}}
+`pulumi do` is in **research preview**. The command interface may change based on feedback.
+{{% /notes %}}
+
+The `pulumi do` command gives you direct create, read, update, delete, and query access to cloud resources from the terminal. It works with every Pulumi provider, so one command shape covers thousands of resource types across every cloud.
+
+This guide takes you from zero to your first resource operations, and shows you what you can do with `pulumi do`.
+
+## Prerequisites
+
+1. [Install the Pulumi CLI](/docs/install/), version v3.257.0 or later. Because `pulumi do` is in research preview and evolving quickly, we recommend the latest version.
+1. Credentials for a cloud provider. The examples use AWS, but any [Pulumi provider](/registry/) works the same way.
+
+## Create your first resource
+
+`pulumi do` picks up provider credentials the same way the provider itself does — from standard environment variables and credential files (for AWS, for example, `AWS_PROFILE`, `AWS_REGION`, or `~/.aws/credentials`). You can also pass explicit provider configuration in a file with `--provider-file`.
+
+Create an S3 bucket, giving it a name:
+
+```bash
+$ pulumi do aws:s3:Bucket create my-bucket
+```
+
+The provider plugin is installed automatically on first use. The CLI shows you the planned changes and asks for confirmation (which can be skipped by passing `--yes`), then creates the bucket:
+
+```
+ +  aws:s3:Bucket my-bucket created
+Outputs:
+    arn   : "arn:aws:s3:::my-bucket-f998a37"
+    bucket: "my-bucket-f998a37"
+    id    : "my-bucket-f998a37"
+    ...
+
+Created my-bucket (snippet c7901139-b263-4572-9a57-6e393c2eae2c)
+```
+
+Note that the bucket name uses the same provider defaults and [auto-naming](/docs/iac/concepts/resources/names/#autonaming) rules as a full Pulumi program.
+
+By default, `pulumi do` runs in stateful mode: the bucket is recorded as a snippet in the state file, and its lifecycle is tracked. You can bypass this tracking by passing the `--stateless` flag.
+
+## Add an object to the bucket
+
+A bucket wouldn't be very useful without anything to put into it.  Let's create a bucket object in the bucket.
+
+To do that, we need to reference the newly created bucket.  We can do that either by creating a resource-file (in JSON format), and passing that via the `--resources-file` option.  The easiest way to create that file is by using the output of `pulumi do show-resources --json`. The names `pulumi do show-resources` shows can also be used to reference the resource directly, as we'll do in the example below.
+
+Further, inputs can be passed to `pulumi do` either via `--<input>` flags for scalar inputs, or through a YAML formatted file, that's passed with `--input-file`. In the following example we'll pass the bucket name through the YAML file, while passing the content through an `--<input>` flag.
+
+```yaml
+# object.yaml
+bucket: ${my_bucket}
+```
+
+```bash
+$ pulumi do aws:s3:BucketObject create my-object --input-file object.yaml --content "Hello from pulumi do" --yes
+```
+
+```
+ +  aws:s3:BucketObject my-object created
+Outputs:
+    arn        : "arn:aws:s3:::my-bucket-f998a37/my-object"
+    contentType: "application/octet-stream"
+    etag       : "a7008c67f9dcdd81b4d0bde99ebf21c2"
+    id         : "my-bucket-f998a37/my-object"
+    ...
+
+Created my-object (snippet 08de6783-0957-415d-98f8-e304399a1d09)
+```
+
+## Read, update, and delete
+
+Read the current state of any resource by its cloud provider ID. The id is shown in the outputs, or can be read using `pulumi state get <name>`.
+
+```bash
+$ pulumi do aws:s3:Bucket read my-bucket-f998a37
+$ pulumi do aws:s3:BucketObject read my-bucket-f998a37/my-object
+```
+
+Update a resource with `patch`, which reads the current state, merges your changes, and shows you a diff to confirm. For example, to tag the bucket:
+
+```yaml
+# tags.yaml
+tags:
+    environment: dev
+```
+
+```bash
+$ pulumi do aws:s3:Bucket patch my-bucket --input-file tags.yaml
+```
+
+And when you're done, delete both resources by the names you gave them. The object first, because a bucket must be empty before it can be deleted:
+
+```bash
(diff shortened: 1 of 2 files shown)


tgummerer:
@claude review this?

i-am-tom (approved):

tgummerer:
> patch example contradicts the reference page

Fixed the example page in #20831

> pulumi state get <name> does not appear to exist

This is about to be merged.  Will exist in 3.257.0

jkodroff:
@claude #new-review

tgummerer:
> requires installing the Pulumi CLI version 3.257.0 or later." — verdict: contradicted; framing

No.  stateless is only in 3.257.0, and explaining all the difference between the different versions is not worth it. Just install a new CLI if necessary.

> the scripting example drops the resource-name argument that every other

Yes, because it's using the stateless version.  This is fine.

tgummerer:
@claude #update-review

tgummerer:
@claude #update-review v3.257.0 is released now.

content/docs/iac/cli/direct-resource-operations/get-started.md
```diff
@@ -0,0 +1,181 @@
...
+        parent: iac-cli-do
+        identifier: iac-cli-do-get-started
+        weight: 1
+---
+
+{{% notes type="info" %}}
+`pulumi do` is in **research preview**. The command interface may change based on feedback.
+{{% /notes %}}
+
+The `pulumi do` command gives you direct create, read, update, delete, and query access to cloud resources from the terminal. It works with every Pulumi provider, so one command shape covers thousands of resource types across every cloud.
```
cnunciato: Has an AI ring to it:

```suggestion
The `pulumi do` command gives you direct create, read, update, delete, and query access to cloud resources from the terminal. It works with every Pulumi provider, so covers thousands of resource types across every cloud.
```

content/docs/iac/cli/direct-resource-operations/get-started.md
```diff
@@ -0,0 +1,181 @@
...
+
+Note that the bucket name uses the same provider defaults and [auto-naming](/docs/iac/concepts/resources/names/#autonaming) rules as a full Pulumi program.
+
+By default, `pulumi do` runs in stateful mode: the bucket is recorded as a snippet in the state file, and its lifecycle is tracked. You can bypass this tracking by passing the `--stateless` flag.
+
+## Add an object to the bucket
+
+A bucket wouldn't be useful without anything to put into it. Create a bucket object in the bucket.
+
+The object needs to reference the bucket you just created. `pulumi do` auto-assigns every tracked resource an identifier, derived from its name with hyphens converted to underscores — so `my-bucket` becomes `my_bucket`. To check the identifiers Pulumi assigned, run `pulumi do show-resources`:
```
cnunciato, line 58: >with hyphens converted to underscores

Curiosity: Why? (I don't think I've seen this before, unless it's been an internal detail up to now.)

content/docs/iac/cli/direct-resource-operations/get-started.md
```diff
@@ -0,0 +1,181 @@
...
+
+```bash
+$ pulumi do show-resources
+
+NAME       URN
+my_bucket  urn:pulumi:default::default-global-project::aws:s3/bucket:Bucket::my-bucket
+```
+
+You can also supply your own identifiers by passing a JSON file that maps identifiers to resource URNs with `--resources-file`; its entries take precedence over the auto-assigned ones.
```
cnunciato, line 67: >with `--resources-file`;

We should include an example of what this file looks like. (I don't think it's covered anywhere else, and I don't think I see it mentioned in the CLI help, either.)

Separately, though: Why would you do this? And how? Presumably you wouldn't always know what the URNs are in advance, since some are computed — right? Or am I misunderstanding? (I totally might be.)

content/docs/iac/cli/direct-resource-operations/get-started.md
```diff
@@ -0,0 +1,181 @@
...
+
+Inputs can be passed to `pulumi do` either via `--<input>` flags for scalar inputs, or through a YAML formatted file that's passed with `--input-file`. Inside an input file, reference another resource by its identifier with `${...}`. The following example passes the bucket reference through the YAML file, while passing the content through an `--<input>` flag:
```
cnunciato, line 69: I'm honestly pretty confused here. 

The task at hand seems to be to add an object to a bucket. Ordinarily this would be done with `aws s3 cp s3://my-bucket-name ./my-file.txt`, so I'd expect something similarly simple for `pulumi do` — and according to these instructions (and the [provider docs](https://www.pulumi.com/registry/packages/aws/api-docs/s3/bucketobject/#constructor-example)), that _should_ be possible — yet:

```bash
➜ pulumi do aws:s3:BucketObject create "my-object" --bucket "my-bucket-c058965" --content "Hello, world"
error: unknown flag: --bucket
```

The example quoted below, with the input file, _does_ work — but it's not clear to me why `--content` works but `--bucket` doesn't, and conceptually, the example is also quite a bit more complex in that you have to create this additional YAML file, which itself contains a mapping expression (that also seems to require the Pulumi YAML provider), and none of this is explained, so you're sort of left wondering whether these input files are required, what `my_bucket` actually is, why that works but `my_bucket.id` (the actual property we're presumably referencing when creating the object) doesn't, and so on. 

Since this is a getting-started guide, we should keep it as conceptually simple and copy-paste friendly as possible. If we could get something like this to work:

```bash
pulumi do aws:s3:BucketObject create "my-object" \
  --bucket "my-bucket-c058965" \
  --content "Hello, world"
```

... that'd be ideal. We can always link off to the docs on using input files as a level-two concept.

(And if this is just totally me doing something wrong, let me me know! As always, that could well be the case. 😄)

content/docs/iac/cli/direct-resource-operations/get-started.md
```diff
@@ -0,0 +1,181 @@
...
+Update a resource with `patch`, which reads the current state, merges your changes, and shows you a diff to confirm. For example, to tag the bucket:
+
+```yaml
+# tags.yaml
+tags:
+    environment: dev
+```
+
+```bash
+$ pulumi do aws:s3:Bucket patch my-bucket --input-file tags.yaml
```
cnunciato, line 110: Similarly here, I'd expect to be able to do this inline by passing `--tags` — or if that doesn't work (e.g., if inline values can't/don't handle nested/JSON-like structures by design), then _this_ is the place to introduce `--input-file`.

content/docs/iac/cli/direct-resource-operations/get-started.md
```diff
@@ -0,0 +1,181 @@
...
+And when you're done, delete both resources by the names you gave them. The object first, because a bucket must be empty before it can be deleted:
+
+```bash
+$ pulumi do aws:s3:BucketObject delete my-object --yes
+$ pulumi do aws:s3:Bucket delete my-bucket --yes
+```
+
+## Explore a provider from the terminal
+
+Every `pulumi do` command addresses a resource or function by its type token, in the form `<package:module:type>` — for example, `aws:s3:Bucket` is the `Bucket` resource in the `s3` module of the `aws` package. For single-module packages, you can omit the module segment.
```
cnunciato:
```suggestion
Every `pulumi do` command addresses a resource or function by its [type token](/docs/iac/concepts/resources/names/#types) in the form `<package:module:type>`. For example, `aws:s3:Bucket` is the `Bucket` resource in the `s3` module of the `aws` package. For single-module packages, you can omit the module segment.
```

content/docs/iac/cli/direct-resource-operations/get-started.md
```diff
@@ -0,0 +1,181 @@
...
+
+You don't need to know the token up front. Pass a partial token and `pulumi do` lists what's available at that level:
```
cnunciato:
```suggestion
You don't need to know the token up front; you can pass a partial token and `pulumi do` will list what's available at that level:
```

content/docs/iac/cli/direct-resource-operations/get-started.md
```diff
@@ -0,0 +1,181 @@
...
+  aws:s3:getBucketObject
+  ...
+
+Resources:
+  aws:s3:AccessPoint
+  aws:s3:Bucket
+  ...
+```
+
+`--help` works at every level of the command tree, generated from the provider's schema. On a resource, it shows the available operations plus every input and output with its type and documentation:
```
cnunciato: This seems like an implementation detail that doesn't need to appear in a getting-started guide.

```suggestion
`--help` works at every level of the command tree. On a resource, it shows the available operations plus every input and output with its type and documentation:
```

content/docs/iac/cli/direct-resource-operations/get-started.md
```diff
@@ -0,0 +1,181 @@
...
+
+```bash
+$ pulumi do aws:s3:Bucket --help
+```
+
+## Query the cloud with provider functions
+
+Providers also expose read-only functions for querying cloud APIs. Invoke them directly by name, with inputs as flags or an `--input-file`:
```
cnunciato:
```suggestion
Providers also expose read-only functions for querying cloud APIs. You can invoke them directly by name, with inputs as flags or with an optional `--input-file`:
```

content/docs/iac/cli/direct-resource-operations/get-started.md
```diff
@@ -0,0 +1,181 @@
...
+{
+  "arn": "arn:aws:ec2:us-west-2:123456789012:vpc/vpc-d7b311af",
+  "cidrBlock": "172.31.0.0/16",
+  "enableDnsHostnames": true,
+  "id": "vpc-d7b311af",
+  ...
+}
+```
+
+`--default` isn't a `pulumi do` flag — it's the `default` input of `getVpc`, passed with the same `--<input>` form used for resource inputs.
```
cnunciato:
```suggestion
`--default` isn't a `pulumi do` flag — it's the `default` input of `getVpc`, passed with the same `--<input>` form used for resource inputs. When passed as CLI inputs in this way, Boolean values like this one are interpreted as `true`. 
```

content/docs/iac/cli/direct-resource-operations/get-started.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Use it in scripts
+
+Pass `--output json` to get machine-readable output for resource operations, which lets you compose `pulumi do` with tools like `jq`:
+
+```bash
+$ pulumi do aws:s3:Bucket create --stateless --yes --output json | jq -r '.id'
+bucket-9f7f27f
```
cnunciato, line 172: Since this is a getting-started guide, I'd include the corresponding `destroy`:

```suggestion
$ pulumi do aws:s3:Bucket create --stateless --yes --output json | jq -r '.id'
bucket-9f7f27f

$ pulumi do aws:s3:Bucket delete bucket-9f7f27f --stateless --yes
```

content/docs/iac/cli/direct-resource-operations/get-started.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+This same predictable command shape and output contract across every provider is what makes `pulumi do` work well for coding agents too — see [Pulumi CLI for agents](/docs/ai/cli-for-agents/).
```
cnunciato:
```suggestion
Keeping this same command shape and output format across all Pulumi providers makes `pulumi do` work especially well for coding agents, too. See [Pulumi CLI for agents](/docs/ai/cli-for-agents/) to learn more.
```

content/docs/iac/cli/direct-resource-operations/get-started.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Next steps
+
+1. Read the [direct resource operations reference](/docs/iac/cli/direct-resource-operations/) for the full command syntax, flags, and provider configuration options.
+1. Browse the [Pulumi Registry](/registry/) to see the providers and resources you can operate on.
+1. When a one-off resource grows into real infrastructure, graduate to [Pulumi IaC](/docs/iac/get-started/) for state tracking, dependency management, and repeatable deployments.
```
cnunciato: Since it's already "real" infra — and with stateful as the default, also also tracked and essentially managed:

```suggestion
1. When it's time to collect the resources you've created with `pulumi do` into a new Pulumi project, you can do so by graduating to [Pulumi IaC](/docs/iac/).
```

More importantly though: How does one actually do this? (The Getting Started guide probably isn't the right place to send them.) Do we have docs on this yet?

cnunciato (commented):
Bunch lil' things for ya!

content/docs/iac/cli/direct-resource-operations/get-started.md
tgummerer: Hrm, maybe with stateful you don't really need to do this at all :thinking: For stateless you would import the resources using `pulumi import`. Maybe we can just drop this point completely.

content/docs/iac/cli/direct-resource-operations/get-started.md
tgummerer: Because `-` is not a valid as part of identifiers in some languages, e.g. YAML.  So auto naming replaces them so the identifier is always valid and usable by the user.

content/docs/iac/cli/direct-resource-operations/get-started.md
tgummerer: You use this, because you're not using URNs in expressions directly, but rather need to use a name.  The `resources-file` is an alternative to the auto assigned name, and can be used in addition to auto names. Not sure how to best describe this in a getting started guide without things getting too complicated.

content/docs/iac/cli/direct-resource-operations/get-started.md
tgummerer: The problem here is that `--bucket` is an expression, not just a string/int/bool, that's allowed as regular flags.  Those input flags are a bit more complicated, and currently require an `--input-file` (similar to arrays and maps).  So I think what you want is currently not possible with the command.

content/docs/iac/cli/direct-resource-operations/get-started.md
tgummerer: Similar to above, the problem is that this is a list, and thus requires the `--input-file` argument.

content/docs/iac/cli/direct-resource-operations/get-started.md
cnunciato: Ah — because it takes a complex type (`aws:s3:Bucket`)? If so, that'd make sense ... although in this case, it looks like `bucket` also takes plain `string`: 

https://www.pulumi.com/registry/packages/aws/api-docs/s3/bucketobject/#bucket_nodejs

This kind of feels like a bug to me. If you think a case could be made for making this work, I'd be happy to file one. It definitely seems reasonable to expect that for something that _can_ (according to the docs) be passed as a plain value, passing a plain value should work.

content/docs/iac/cli/direct-resource-operations/get-started.md
cnunciato: This also seems pretty doable — happy to file for this one as well if you agree! https://www.pulumi.com/registry/packages/aws/api-docs/s3/bucketobject/#tags_nodejs

```bash
pulumi do aws:s3:Bucket patch my-bucket --tags '{"key": "value"}'
```

or 

```bash
pulumi do aws:s3:Bucket patch my-bucket --tags "key1=value1" --tags "key2=value2"
```

content/docs/iac/cli/direct-resource-operations/get-started.md
cnunciato: Asked this internally, and it sounds like first-class support for this is up next, so I'm okay with having the path be `import` and linking to that. Will update my suggestion!

cnunciato:
```suggestion
1. When it's time to collect the resources you've created with `pulumi do` into a new or existing Pulumi [project](docs/iac/concepts/projects/), you can import them using their type tokens and resource IDs. See [Importing resources](/docs/iac/guides/migration/import/) to learn more.
```

cnunciato (approved):
Thanks @tgummerer! Let me know what you think about filing those issues and I'll take care of 'em.

content/docs/iac/cli/direct-resource-operations/get-started.md
cnunciato: If you could point me to a working/usable example of what this `resources-file` looks like, I'd be happy to follow up on this one.

content/docs/iac/cli/direct-resource-operations/get-started.md
tgummerer: I'm a bit confused by this one, because it says it takes `str | str` :thinking:

But yes, in general it's because it takes a complex type, which we can't do on the CLI yet.  Making it work with a plain input when possible does seem like the right thing to do though.

content/docs/iac/cli/direct-resource-operations/get-started.md
tgummerer: Yeah, an issue would be good for this.  I agree that it should be doable, the trickiest bit will be deciding on a syntax.

content/docs/iac/cli/direct-resource-operations/get-started.md
tgummerer: So as mentioned in the docs, you can genenrate a `resources-file` using `pulumi show-resources --output=json`.  So in this example, you would do

```
pulumi show-resources --output=json >resources-file.json
pulumi do aws:s3:BucketObject create my-object --input-file object.yaml --content "Hello from pulumi do" --yes --resources-file resources-file.json
```

That is as long as you leave `object.yaml` the same as it currently is.  You can rename the resource in the `resources-file`, and then rename the corresponding thing in `object.yaml` as well.

tgummerer:
> Let me know what you think about filing those issues and I'll take care of 'em.

Thank you!  I've replied in the comment threads, but all of those things seem reasonable to me to open issues for them.

PR #20851, opened 2026-08-12
Make the nightly claims re-verify check facts, not itself

CamSoper:
The claims re-validator went live on Aug 11. Its first two sweeps checked 50 volatile entities and found 0 stale — and both halves of that result were misleading, in the two ways `claims-reverify.yml`'s own header had flagged as worth watching. This fixes both.

## 22 of 35 `verified` verdicts were sourced from our own docs

16 of them cited the published copy of the very page making the claim:

```
content/docs/iac/get-started/aws/modify-program.md
  verified high ← https://www.pulumi.com/docs/iac/get-started/aws/modify-program/
```

`www.pulumi.com` is this repo rendered, so that check confirmed the page against itself. It can never go stale, and a `contradicted` from the same place would be equally meaningless. The worst case restated a marketing figure — "5-10X improvements" in the onboarding guide — as `verified high` by quoting itself.

`source_is_own_corpus` now demotes any decided verdict whose every cited source is a `www.pulumi.com` URL or a `content/` file to `unverifiable`: reported, never counted fresh, never written as a stale marker. Positive evidence only — a source naming nothing identifiable (no URL, no path) is left alone, because unrecognized is not the same as circular.

Replayed over both nights' reports, this demotes 22 of 35 and leaves 13 real verifications — against AWS documentation, `pulumi/pulumi` source, `.goreleaser.yml`, and the `data/policy_pack_policies/*.json` files that generate the tables being checked.

| | Aug 11 | Aug 12 |
|---|---|---|
| `fresh` as reported | 16 | 19 |
| demoted (own-corpus evidence) | 11 | 11 |
| `fresh` under this change | 5 | 8 |

## 48 of the 50 entities were typed `numerical` — and many were never volatile

`ALWAYS_VOLATILE_TYPES` made every `numerical` claim volatile unconditionally, so counts of a page's own example output rode along: *"this tutorial takes 15 minutes"*, *"the example update creates 3 resources"*, *"the preview output shows 2 changes"*. Those figures are properties of the prose — they can only change when the page changes, so a nightly check can never find drift in them. Eight came back `not-a-claim` from the verifier saying exactly that.

`SELF_DESCRIBING_RE` excludes them from `volatile`. They stay keyed and indexed, and are still re-checked whenever the page itself is reviewed. The policy-pack row counts share some of that vocabulary ("this page lists all 139 policies…") and deliberately stay volatile — those do drift, and a test pins the distinction.

`reverify-claims.py` re-derives the flag through `entity_key.is_volatile` rather than trusting what each snapshot stored, so the narrower rule reaches the whole index the night it ships instead of page by page as each page is re-reviewed. It falls back to the stored flag if the module can't load.

## The report artifact

The Slack summary has always said "the evidence lives in the report artifact"; no step ever uploaded one. That matters more than it sounds — Actions masks every occurrence of the org name, so source URLs in the run log read `www.***.com`, which is precisely the field this PR turns on.

## Deliberately partial

Neither fix stops the verifier from reaching for the site in the first place; they stop it from being *recorded as confirmed* on that basis. Prompting it toward independent sources means editing `verify-claims.py`, which the pre-merge review lane shares, so that wants its own change rather than a ride-along here.

Not addressed either: near-duplicate keys burning one verifier call each (three separate entities for one ELB SSL policy fact, four for one policy-table count). Real waste, but it's a key-normalization change with a different blast radius.

## Verification

`make lint` and `make test-review-pipeline` (21 suites) green. The new assertions in both suites were negative-tested — they go red against the pre-fix code and name the specific behavior that regressed. The demotion counts above come from replaying the two live reports through the new code, not from an estimate.

Nothing here changes the pre-merge review lane: `volatile` is read only by the nightly job.

### DIFF
--- a/.claude/commands/docs-review/scripts/entity_key.py
+++ b/.claude/commands/docs-review/scripts/entity_key.py
@@ -70,6 +70,13 @@ PRICING_LIMIT_RE = re.compile(
 # except the page it came from (which is what makes those checks circular —
 # see reverify-claims.py's own-corpus demotion). Excluded from `volatile`;
 # still keyed, still indexed, still re-checked whenever the page is reviewed.
+#
+# PRICING_LIMIT_RE vetoes the exclusion, because the two overlap on the one
+# phrasing where this rule would otherwise do damage: "the example stack costs
+# $12/month" is self-describing in form but the figure is an outside-world
+# price that drifts on its own. Keeping a stray example count in the nightly
+# pool costs one verifier call and a demoted verdict; dropping a real price
+# means never checking it again.
 SELF_DESCRIBING_RE = re.compile(
     r"\b(?:examples?|samples?|tutorials?|walkthroughs?|screenshots?|"
     r"meta[- ]description|"
@@ -165,7 +172,8 @@ def derive(claim: dict) -> tuple[str | None, bool]:
     volatile = ctype in ALWAYS_VOLATILE_TYPES or (
         ctype == "entity-spec" and bool(PRICING_LIMIT_RE.search(text))
     )
-    if ctype == "numerical" and SELF_DESCRIBING_RE.search(text):
+    if (ctype == "numerical" and SELF_DESCRIBING_RE.search(text)
+            and not PRICING_LIMIT_RE.search(text)):
         volatile = False
 
     if ctype not in KEYED_TYPES:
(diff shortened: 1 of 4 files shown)


CamSoper:
@claude #update-review

Both ⚠️ findings addressed in 52639de.

**L367-370 (health lane vs. demotion)** — real, and the diagnosis was the broken part, not the alarm. I did not take the suggested `n_inconclusive > n_demoted` guard: it returns `"ok"` for a night that detected no drift at all, which is exactly what this signal exists to notice. Instead the alert stays and gets accurate: `observe_reverify` carries `n_demoted` into the detail, and the `CONSEQUENCES["reverify"]` text now separates the two causes — a demotion count means the verifier's source routing, anything else means plumbing (S3 sync, API key, gh lane). Your point about the docstring was right and applied to two places, not one: signal-health's signal index and `reverify-claims.py`'s meta-block comment both described all-inconclusive as the broken-verifier tell.

**L168-169 (self-describing beats price/limit)** — correct, took the fix as suggested. `PRICING_LIMIT_RE` now vetoes the exclusion, so "the example stack costs about $12 per month" stays in the nightly pool. The asymmetry is the argument: a stray example count costs one verifier call and a demoted verdict, a dropped price is never checked again. Three self-test cases pin it (price, free-tier limit, seat count).

Both new suites negative-tested — the veto cases and the demotion detail go red against the pre-fix code. `make lint` and `make test-review-pipeline` (21 suites) green.

PR #20896, opened 2026-08-14
Add HCL to the language lists it was missing from

CamSoper:
> [!WARNING]
> I've turned on auto-merge for this! 

### Proposed changes

HCL shipped as a supported language with its own section under Languages & SDKs, but a number of pages that enumerate the supported languages were never updated. Three of them are now **incorrect** rather than merely incomplete: the project, project-file, and plugin runtime tables omit `hcl`, which is a valid `runtime` value in both `Pulumi.yaml` and `PulumiPlugin.yaml`.

**Language lists**

- `iac/_index.md` — add HCL to the section description and the Languages card grid, so the IaC landing page matches its own child hub. Uses the existing `hcl-color-32-32` icon.
- `concepts/projects/_index.md` — add `hcl` to the supported-runtime list, add an HCL tab to the `Pulumi.yaml` example chooser, and note the `.tf` entrypoint convention alongside the other languages'.
- `concepts/projects/project-file.md` — add `hcl` to the `runtime` row.
- `concepts/plugins.md` — add `hcl` to the `PulumiPlugin.yaml` `runtime` row. A directory of `.tf` files with a `PulumiPlugin.yaml` declaring `runtime: hcl` is precisely how an HCL multi-language component is defined — see the [HCL component reference](https://www.pulumi.com/docs/iac/languages-sdks/hcl/hcl-component-reference/).
- `integrations/clouds/{aws,azure,gcp,kubernetes}/_index.md` — add HCL to the shared "Pulumi IaC lets you define cloud infrastructure using…" sentence. Left as prose rather than swapped for `{{< pulumi-languages >}}`, because these are *or*-lists in a different order and the shortcode emits a fixed *and*-list that doesn't read correctly in this sentence.

**Card grid fix (second commit)**

Adding the seventh language card exposed a latent layout bug: the `.languages` grid was sized for six cards, so the seventh wrapped and `flex: 1` stretched it across the full container width. The cards now take a fixed basis at each breakpoint, sized so a full row still fills the container — two per row on mobile, three at tablet, seven on one row at desktop — and a short final row sits left-aligned with every card the same size. `.languages` is used only by `content/docs/iac/_index.md`, so the blast radius is this one page.

**Adjacent fix on the same pages**

- `comparisons/cdktf.md` — point four `use-terraform-module` links at the canonical path rather than the pre-move alias, per the internal-link rule in `AGENTS.md`. The canonical path is already used in 21 other places.

**Not included**

An earlier revision added three bullets to the "What you'll learn" list on `get-started/terraform/_index.md`, on the theory that the list covered five of the section's eight pages. That was wrong and has been reverted: those five bullets are a one-to-one mirror of the five numbered items in the "Overview of examples" section below them, describing the progressive ECS/web-app example arc. The first-look, state-backend, and remote-execution pages are standalone and aren't taught through that arc. Those three pages genuinely aren't mentioned anywhere in the section intro, but fixing that means restructuring the intro — a content decision rather than a mechanical one.

This is the mechanical, verifiable slice of a larger gap. HCL is still absent from the four cloud getting-started guides, most of the core concepts pages, and `static/programs/`. Those need example programs and a product decision first, and aren't in scope here.

### Test plan

- `make lint`: 1850 files parsed, 0 errors; Prettier clean.
- Built and rendered locally with Hugo. Verified the new card, the description, the HCL chooser tab, and both runtime tables render as intended.
- Card grid checked in **both light and dark mode** at 390, 1024, and 1440 px — seven across at desktop, no full-width orphan at any width, no horizontal overflow at mobile.
- Canonical link target verified to exist on `master`; the old path confirmed to be an alias on the destination page.
- `runtime: hcl` on plugins confirmed against `pulumi/pulumi` master and our own HCL component reference.

### Related issues (optional)

None — no open issue tracks the HCL docs surface-area gap.

---

### DIFF
--- a/content/docs/iac/_index.md
+++ b/content/docs/iac/_index.md
@@ -10,7 +10,7 @@ menu:
         weight: 1
 meta_desc: Learn how to create, deploy, and manage infrastructure on any cloud using Pulumi's open source infrastructure as code SDK.
 h1: Infrastructure as Code
-description: <p>Define and manage cloud infrastructure using familiar programming languages. Pulumi's <a href="https://github.com/pulumi/pulumi" target="_blank">open source</a> infrastructure as code SDK supports TypeScript, Python, Go, .NET, Java, and YAML.</p>
+description: <p>Define and manage cloud infrastructure using familiar programming languages. Pulumi's <a href="https://github.com/pulumi/pulumi" target="_blank">open source</a> infrastructure as code SDK supports TypeScript, Python, Go, .NET, Java, YAML, and HCL.</p>
 link_buttons:
   primary:
     label: Get Started
@@ -55,6 +55,9 @@ sections:
   - label: YAML
     icon: icon-32-32 yaml-color-32-32
     link: /docs/iac/languages-sdks/yaml/
+  - label: HCL
+    icon: icon-32-32 hcl-color-32-32
+    link: /docs/iac/languages-sdks/hcl/
 - type: button-cards
   heading: Resources
   cards:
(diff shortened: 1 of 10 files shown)


jeffmerrick (approved):
Looks good!

PR #20901, opened 2026-08-14
Add HCL to the AWS, Azure, GCP, and Kubernetes get-started guides

joeduffy:
Pulumi HCL is already a first-class language, but the four core get-started guides still offered only TypeScript, Python, Go, C#, Java, and YAML. This adds HCL alongside them at every step of each guide.

## Changes

| Page | AWS | Azure | GCP | K8s |
|---|:-:|:-:|:-:|:-:|
| `_index.md` — chooser + prerequisites | ✅ | ✅ | ✅ | ✅ |
| `create-project.md` — `pulumi new <cloud>-hcl` + program | ✅ | ✅ | ✅ | ✅ |
| `deploy-stack.md` — snake_case `pulumi stack output` | ✅ | ✅ | ✅ | n/a |
| `modify-program.md` — website / blob / IAM resources | ✅ | ✅ | ✅ | ✅ |
| `create-component.md` — module-as-component | ✅ | ✅ | ✅ | ✅ |

Plus `layouts/shortcodes/compfile.html` gains an `hcl` entry (`website/main.tf`); `langfile` already had one.

## Design decisions worth a reviewer's attention

**Components map to HCL `module` blocks.** A module instance *is* a Pulumi component resource, so children are parented automatically and the "set each resource's `parent`" step in the other languages drops out. The alternative — packaging the directory as a multi-language component and consuming it via `pulumi package add` — resolves `fileasset()` relative to the module directory rather than the program, which breaks the `index.html` upload these tutorials are built around.

**Modules prefix their children's Pulumi names** (`my-website.my-bucket`), which is fatal on three of four clouds: an Azure storage account name would exceed 24 characters and contain a dot, a dotted GCS bucket triggers domain verification, and a Kubernetes Service name can't contain a dot. The module code sets `pulumi { name = ... }` on each resource, explained once per page.

**HCL exports snake_case**, so it shares Python's choosable blocks wherever a guide prints a `pulumi stack output` command. Sample terminal output that differs only by output name is left shared — matching how these guides already treat Python.

## Verification

Every HCL program here — initial, modified, and component form, across all four clouds — was verified with `pulumi preview` against real provider schemas (Pulumi 3.257.0, local backend, dummy credentials). That caught a real error: `pulumi convert --language hcl` emits `index_document = { ... }` for `aws_s3_bucket_website_configuration`, which the schema rejects; it must be a block. The `components:index:Website` type tokens and child resource names in the new preview output blocks are copied from actual runs, not invented.

`make lint` passes. All 20 pages render at HTTP 200 under `make serve` with no shortcode leakage.

## Incidental fixes

Three pre-existing errors in blocks already being edited: a typo (`ino` → `into`), and sample output naming a component `my-site` when the code creates `my-website` (AWS and Azure).

## Out of scope

`/docs/iac/get-started/terraform/` — its `first-look.md` already offers HCL, and the remaining pages are about converting *away* from HCL, where it isn't a meaningful target language.


joeduffy:
@claude #update-review

Fix response for the two low-confidence findings, plus three I caught in my own pass. All in 99c4ca7.

**[L1040 / L983 / L934 / L1081] "component's type is derived from the directory name" — accepted, and verified before accepting.**

You're right, and I checked it rather than taking it on faith: I ran a module out of a `my-site-thing` directory and got `components:index:MySiteThing`. So the `components:index:` prefix is fixed and only the trailing segment derives from the directory — title-cased, which the original sentence also didn't say. Reworded in all four guides:

> The `Website` segment of the component's type, `components:index:Website`, is the title-cased name of the directory the module lives in; the `components:index:` prefix is fixed.

**Style suggestions — both taken, in all four guides.** "It'll be mostly empty to start with" → "It starts out with just its input declared"; "There is no step to reparent the resources" → "Nothing needs reparenting". The remaining `mostly empty` hit on the Kubernetes page is pre-existing prose in the non-HCL paragraph that this PR only wrapped in a `choosable`, so I left it alone.

**Three more from my own review pass**, found by rendering each page as a reader with HCL selected actually sees it and scanning for prose that leaks across languages:

1. The module code's comment claimed the explicit `pulumi { name = ... }` values "keep the names the program used before" — true for AWS, but Azure's component deliberately renames to `my-group`/`myaccount`, and GCP's bucket object shifts `index-html` → `index.html`. It now states what it does instead of making a claim that only holds in some guides.
2. AWS and GCP prose called `pulumi/aws` and `pulumi/gcp` the "native" Pulumi providers. Both are bridged from their upstream Terraform providers — only `azure-native` and `kubernetes` are native in that sense. Now "the Pulumi AWS provider from the Pulumi Registry". Also fixed "Azure resource manager" → "Azure Resource Manager".
3. `gcp/modify-program.md` told every reader to use "Pulumi's `FileAsset` class" — no such class in HCL (it's the `fileasset()` function), and YAML has the same mismatch with `fn::fileAsset`. Now points at the language-neutral [asset](/docs/iac/concepts/assets-archives/#assets) concept.

**Triaged findings — agreed on all of them.** Worth confirming the `gcp_storage_bucket_i_a_m_binding` one independently: I hit that name by running `pulumi convert --language hcl` over the equivalent YAML program and then previewing it, so it's what the toolchain actually emits, not a guess.

**Pre-existing items — leaving all three.**

- `gcp/create-project.md` `.url` returning a `gs://` URI rather than a DNS name: agreed that fixing only the HCL tab would make it the odd one out. Worth a separate pass across all seven tabs.
- The dangling `menu.clouds.parent` entries on the Azure and Kubernetes `_index.md` files: untouched frontmatter, out of scope here.

Every HCL program in this PR — initial, modified, and component form, across all four clouds — was verified with `pulumi preview` against real provider schemas on Pulumi 3.257.0.

joeduffy:
@claude #update-review

Three of the four accepted and extended; one disputed. All in b15c75d.

**[L88 gcp] / [L303 aws] / [L237 azure] — accepted, and swept rather than patched.**

You were right that the GCP fix was half done, and the framing ("name the concept, not the language's type") is the right generalization. So instead of fixing the three flagged lines, I ran the HCL reader's view of all four guides through a grep for backticked type names appearing outside any `choosable` and fixed the whole class:

| File | Was | Now |
|---|---|---|
| `gcp/modify-program.md` L88 | `` `BucketObject` `` | bucket object |
| `gcp/modify-program.md` L223 | `` `BucketObject` `` | bucket object |
| `gcp/modify-program.md` L225 | `` `BucketObject` `` | bucket object |
| `aws/modify-program.md` L303 | ``[`BucketObject`](…)`` | [bucket object](…) |
| `azure/modify-program.md` L237 | `` `FileAsset` `` + ``[`Blob`](…)`` | file asset + [blob](…) |
| `azure/modify-program.md` L323 | `` `staticEndpoint` `` export | a static endpoint export |

The last one you didn't flag, and it's the one I'd rank highest of the six: it names an output the reader types by hand on the next page, and HCL spells it `static_endpoint`. The sentence also said "add it to your return statement", which is TypeScript/C#-shaped.

**Deliberately left, so the boundary is explicit rather than silent:** the numbered "you will add these resources" lists (`BucketWebsiteConfiguration`, `StorageAccountStaticWebsite`, `BucketIAMBinding`, …), plus `dependsOn` at `aws/modify-program.md` L84 and `websiteEndpoint` at L99. Those name canonical Pulumi types and options, they deep-link to the registry/option docs those names title, and they match what six of the seven languages literally write — YAML included. Rewriting them would cost precision for six tabs to gain recognizability for one, and the HCL block sits directly below in every case.

**[L814-815 kubernetes] — disputed.**

The rationale wasn't traded away; it's stated in prose immediately below the code block, which is where the other three guides put their per-cloud reason:

- `kubernetes/create-component.md` L847 — "Naming the service explicitly matters here: Kubernetes service names must be DNS labels, and the module-prefixed name would contain a dot."
- `azure/create-component.md` L767 — the 24-character/lowercase-alphanumeric constraint.
- `gcp/create-component.md` L715 — the dotted-bucket domain-verification behavior.

So the DNS-compatibility reason is present, once, in the same slot its siblings use. Restoring it to the code comment would state it twice on the one page and re-break the symmetry across the four. If you'd still rather have it in the comment, I'd move it out of the prose sentence rather than duplicate it — happy to do that if you disagree.

**Style — taken.** `just` → `only` on all four `create-component` pages.

`make lint` clean; `choosable` open/close balanced across all 19 pages.

joeduffy:
@claude #update-review

One accepted and extended, one declined with evidence. In f41696e.

**[L83 azure] — accepted. Fair hit: my boundary was applied by eye, not mechanically.**

You're right that it fits neither set-aside category, and it's structurally the L237 sentence I'd already fixed. Taken as suggested. Re-running the sweep afterward turned up one more of the same shape that neither of us had flagged:

- `aws/modify-program.md` L530 — *"We prepend `http://` using a helper because `websiteEndpoint` is an output property…"*. `websiteEndpoint` is a bare reference to a property in the reader's own program (HCL: `website_endpoint`), so same class. The rewrite also drops "using a helper", which was already untrue for YAML before HCL existed. Now: *"We prepend `http://` because the bucket's website endpoint is an output property…"*.

The class is now closed mechanically rather than by eye — the sweep over all four guides' HCL view returns only the two categories set aside: the numbered resource lists (`aws` L14/16/18, `azure` L12/14, `gcp` L14/16) and `dependsOn` (`aws` L84).

**[L482 + 9 siblings] — declining, and here's the evidence rather than just a preference.**

This one predates the PR and isn't HCL-specific. On `master`, before any of my commits:

| | declares | shared transcript prints |
|---|---|---|
| `aws/create-project.md:287` | `pulumi.export('bucket_name', …)` | `bucketName` (`deploy-stack.md:73`) |
| `azure/create-project.md:322` | `pulumi.export("storage_account_name", …)` | `storageAccountName` (`deploy-stack.md:37,67`) |
| `gcp/create-project.md:304` | `pulumi.export("bucket_name", …)` | `bucketName` (`deploy-stack.md:37,64`) |

Python has had exactly this mismatch on every one of those pages for as long as they've existed. HCL joins an existing club rather than creating one, and the part that would actually strand a reader — the `pulumi stack output` command they type — was already language-scoped for Python and is now scoped `"python,hcl"`.

So fixing it properly means splitting ~10 transcripts × 2 blocks each **and** correcting Python at the same time, which is a cross-cutting editorial change to six languages' worth of sample output that has nothing to do with adding an HCL tab. That's its own PR. Doing it for HCL alone would leave Python as the odd one out — the same trap we agreed to avoid on the `.url`/`gs://` item.

Recording it as a deliberate decision, not an oversight, which is what you asked for.

`make lint` clean.

joeduffy:
@claude #update-review

Taken. In 177cfb3.

**[L527 gcp]** — you're right that this is the tail of my own reasoning, and I stopped one page short. Applied as suggested. Confirmed it was the last one: `grep -rn "using a helper" content/docs/iac/get-started/` now returns nothing, and Azure's counterpart at L449 never carried the phrase, so the three guides agree.

That clears the last open item on this PR. `make lint` clean.

CamSoper:
@joeduffy I actually started on this same task this afternoon. I'll reconcile what I've done with what you've done, test end-to-end, and push any changes that fall out of that to this PR.

CamSoper:
@joeduffy — I ran this end to end on real clouds before sign-off: every HCL program in the PR (initial, modified, and component form) deployed live on AWS, Azure, GCP, and an AKS cluster — `pulumi new <cloud>-hcl` → `up` → website modify → `curl` → module component → `curl` → `destroy`. The module-block component design held up everywhere: program-relative `fileasset()`, working `pulumi { name }` overrides, `components:index:Website` tokens, and all four per-cloud naming rationales matched observed behavior.

I pushed four commits on top:

1. **k8s `ip` export fix** — the HCL tabs exported `spec.cluster_ip` only. On AKS with `isMinikube false` the Service gets a public LoadBalancer IP but the guide's own `curl $(pulumi stack output ip)` times out against the internal 10.x address. Now conditional with `coalesce(ingress[0].ip, hostname)`, matching the TypeScript/Python tabs; verified serving nginx through both the flat program and the component. (The Java tab has the same pre-existing flaw — left for a follow-up.)
2. **Chooser persistence** — the theme store never persisted `hcl` as the language preference, so picking HCL on a guide index reverted to TypeScript on every stepper click. Dropped `hcl` from `specialPurposeLanguages` (OPA stays page-scoped), plus spec cases and chooser fixes for two tutorial pages whose leading choosables would otherwise render blank under a persisted HCL preference.
3. **Function casing in the HCL language reference** — the Pulumi-specific functions table said `fileAsset`/`pulumiResourceName`; the runtime registers these all-lowercase (calling `fileAsset` fails with "no function named"). Also added the missing `pulumiresourceurn` row.
4. **Fact-check fixes** — unprefixed sources resolve from the OpenTofu registry, not "exactly as in Terraform" (and `hashicorp/azure-native` / `hashicorp/gcp` don't exist there, so that advice only held on AWS); the dotted `"index.html"` resource labels in the AWS/Azure programs are Terraform-invalid and unreferenceable, so they now use `index-html` with an explicit `key`/`blob_name` like the GCP tab; and two GCP transcripts showed an autoname suffix the HCL program never produces.

Everything else in the prose checked out against pulumi-hcl source: the 3.256.0 floor, the snake-casing derivation, the `components:index:` token formation, the module name-prefix behavior, and the `pulumi config set isMinikube` → `var.isMinikube` claim. All cloud resources from testing are destroyed.

CamSoper:
@claude #update-review

cnunciato:
>Chooser persistence

Nice work @CamSoper — this was the main thing I was concerned about. I made these non-persistent at launch because we have several places in the docs (including in the guide) that toggle inline content based on the globally selected language, so when that language is HCL, that content won't be there. Taking a pass now to check for any instances of those, as it's been a while. (I'm nearly positive there are at least a handful of them in the Concepts docs, etc.)

cnunciato:
Ok yeah. Here's what the unit testing page look like on the PR when HCL's been selected:

image

Here's what that section should look like (with fallback to TypeScript):

image

Have a fix, verifying locally, will push a commit.

cnunciato (approved):
All right! Pushed up a couple of additional commits to fix the hidden-content issue as well as a few typos and grammar things. Also ran through the end-to-end flow myself and all went well.

CamSoper:
@claude #update-review

CamSoper:
@claude #update-review

PR #20903, opened 2026-08-14
Add HCL to the AWS Get Started guide

CamSoper:
### Proposed changes

Adds an HCL tab to every page of the AWS Get Started guide that offers language tabs, closing the gap where a reader who chooses Pulumi for its HCL support has no path through the main quickstart.

- **`_index.md`** — HCL in the language chooser, with prerequisites (an AWS account; like YAML, no local runtime is needed).
- **`create-project.md`** — `pulumi new aws-hcl`, the generated `main.tf`, and a note on `pulumi/` provider sources.
- **`deploy-stack.md`** — HCL joins the snake_case `bucket_name` output group alongside Python.
- **`modify-program.md`** — the three website resources, the `BucketObject` with `fileasset()` and native `depends_on`, and the `url` output.
- **`create-component.md`** — a full HCL narrative: components are authored as multi-language components (a `website/` module with `PulumiPlugin.yaml`), registered with `pulumi package add ./website`, and instantiated as `website_aws_s3_website`.

Every command and snippet was verified against a live AWS deployment end to end (`pulumi new` → `up` → website modify → `curl` → component refactor → `curl` → `destroy`), including the module-context details: attribute-map syntax for nested properties, `fileasset` path resolution relative to the module directory (hence `abspath`/`basename`), and automatic parenting.

### Notes for review

- Selecting HCL does not yet persist across stepper pages — the theme store deliberately never persists it. A follow-up PR removes that guard once all four cloud guides have HCL tabs; until then the tab behaves like the existing HCL tab on the Terraform onramp's first-look page.
- #20360 rewrites this guide and does not add HCL; whichever lands second pays for the collision. Happy to rebase this onto it if it lands first.

### DIFF
--- a/content/docs/iac/get-started/aws/_index.md
+++ b/content/docs/iac/get-started/aws/_index.md
@@ -27,7 +27,7 @@ Complete this step-by-step tutorial to deploy an AWS S3 bucket-based website usi
 
 First, choose your language and ensure you've performed any prerequisites:
 
-{{< chooser language "typescript,python,go,csharp,java,yaml" / >}}
+{{< chooser language "typescript,python,go,csharp,java,yaml,hcl" / >}}
 
 {{% choosable language "typescript" %}}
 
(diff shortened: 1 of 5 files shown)


CamSoper:
Superseded by #20901

PR #20914, opened 2026-08-16
Make self-hosted Pulumi Cloud a first-class story

joeduffy:
Self-hosted Pulumi Cloud is a real product with seven supported installations, and the website has been hiding it behind a marketing page written in 2022, a navigation entry that didn't exist, and a compliance page that claimed less than we can actually defend. This fixes the surface around it and corrects a dozen factual errors in the docs.

## Marketing and trust

- **`/product/self-hosted/` rebuilt** on the template-partials system every other product page uses; the bespoke 146-line layout is deleted, URL unchanged. The old page said "Cloud Engineering Platform", offered "Azure Active Directory" as an identity option, **omitted AWS ECS entirely** despite it being our most complete installer, and routed every docs link through the retired `/docs/pulumi-cloud/self-hosted/` path so each click was a redirect.
- The VMware tile is gone. It linked to a contact form with no docs behind it and implied a vSphere installer that doesn't exist.
- The identity copy was wrong on substance, not just naming: there is no Azure AD OAuth connector — Entra ID reaches self-hosted through SAML.
- **`/security/` now states what we don't hold** (ISO 27001, HITRUST, PCI DSS) alongside the SOC 2 Type II attestation, and separates our compliance *policy packs* from our own certification status — a distinction a procurement reader could easily conflate. `/trust/` and `/trust-center/` alias here.
- **FedRAMP gets an equal-weight column** headed "No FedRAMP authorization" on the product page. No ATO, no Marketplace listing, said outright, followed by what self-hosting actually changes: who owns the boundary. A reviewer who can see immediately what we don't hold will trust what we do claim.
- **No FIPS claim.** Nothing in this repo, `pulumi/pulumi`, or `pulumi-self-hosted-installers` substantiates one. It can land later once engineering documents precisely what's true.
- **The `/security/` "Read Whitepaper" button has been dead since January 2026** — `.Resources.GetMatch` on a branch bundle never matches a content page, so it rendered `href=""`. Fixed with `site.GetPage` and an `errorf` guard; I proved the guard fires by deleting the whitepaper and watching the build exit 1.

## Docs corrections

Every fix verified against installer source, not reasoned about:

| Claim | Reality |
|---|---|
| `PULUMI_DATA_PATH` defaults to `/tmp/pulumi-ee/data` | `${HOME}/pulumi-self-hosted-installers/data` |
| `PULUMI_LOCAL_DATABASE_NAME` | `run-ee.sh` reads `PULUMI_LOCAL_DATABASE_HOST` |
| Docker network `pulumi-ee` | `pulumi-self-hosted-installers` |
| "Your sales contact should provide the migrations folder" | Compose pulls `pulumi/migrations` from Docker Hub |
| "The container image repository is private" | `pulumi/service` is public with 220,518 pulls; `pulumi/console` with 114,034 |
| EKS on Kubernetes 1.18–1.21, `npm run install -- update --` | Describes `eks-hosted-deprecated`; current installer is v4.0 |
| Deploy air-gapped "using Helm" | No Helm chart exists anywhere |
| `db.r5.large` is "the installers default" | True for EKS; ECS defaults to `db.t3.small` |

`PULUMI_LICENSE_KEY` causes a hard `exit 1` and was never mentioned — the most common first-run failure.

## Two safety gaps

- **The V2 engine-events schema migration.** All four modern installer READMEs open with *"DO NOT USE THIS VERSION OF THE INSTALLER FOR AN EXISTING INSTALL. CONTACT PULUMI SUPPORT TO MIGRATE THE DB FIRST."* It appeared nowhere in the docs. Now a shortcode on the upgrades guide and all six production deployment pages.
- **First signup becomes the administrator**, including the SAML administrator, and enabling SAML doesn't close email signup. Documented only in the AKS README. Now on every deployment page.

## New and restructured

- `system-requirements` and `troubleshooting` pages, both built from facts that already existed but were scattered. Troubleshooting is indexed by symptom — the string `ALGORITHM=INPLACE is not supported` is what a failed migration actually prints, and it previously lived only inside the operations database page.
- **ECS install guide rewritten** with real commands, config tables, and a verification sequence taken from the quickstart's own integration tests. This is the pattern; the other six still hand off to their READMEs.
- Telemetry split out of `api.md` (510 → 376 lines), keeping the env vars and the `#opentelemetry` anchor in place so inbound links — including from the installer READMEs, which we don't control — keep working.
- Revives and corrects the unmerged June 17 `self-hosted-first-class` branch: its install front door, quickstart rewrite, and Terraform Enterprise migration guide. Its self-serve framing is rewritten, because the evaluation-license mechanism it assumed still doesn't exist — readers were told they could try it themselves and then hit a sales gate.

## Verification

`make lint` and `make build` clean. No broken links introduced. Rendered output checked for every claim above.

## Known follow-ups

- Six deployment guides still delegate to READMEs; ECS sets the pattern.
- `components/console.md` says 1 CPU / 1 GB while `operations/compute-sizing.md` says 0.5 vCPU / 512 Mi. They may be measuring different things (host minimum vs. container request), so I left both rather than delete one on a guess. **This needs an answer before the duplicate requirements tables can be collapsed.**
- Pre-existing, not touched: `/guides/` 404s from both the header nav and footer; the JSON-LD FAQ schema emits raw shortcode markup into `acceptedAnswer.text`; the 2024 post `self-hosted-search-and-deploy` links to `components/api#deployments`, which has always anchored as `#pulumi-deployments`.


joeduffy:
@claude please review this PR.

PR #20930, opened 2026-08-17
Contact page: rebuild the sidebar and swap the form chooser to radio buttons

jeffmerrick:
### Proposed changes

This cleans up the work in https://github.com/pulumi/docs/pull/20867 and adds some additional enhancements and fixes.

- Design cleanup on sidebar, address moved to sidebar
- So the page doesn't start visually unbalanced, the first option is now selected by default so the page loads with a form visible.
- Changes the options from a select to radio-style buttons so users can see all the options and will be more likely to choose the correct one.
- Switching between reasons no longer discards what's been typed. The chooser reads the outgoing form before it swaps and carries first/last name, email, company, and message into the next one. Carried values fill only fields that are still empty, so HubSpot's own prefill of a known contact is left alone; a `?param=` prefill still overwrites both, since an explicit link is the strongest signal of intent. `pulumi-hubspot-form` gains a `getFieldValues()` method and a `carryOverValues` prop, both built on the `collectFieldValues`/`applyPrefill` helpers already there for submit tracking and query-param prefill.
- Two fixes in `_hubspot.scss` that reach every HubSpot form on the site, not just this page:
    - Validation messages carried an extra 1.25rem of top margin, since HubSpot marks each one up as a `<label>` and it picked up the `mt-5` meant to space a field label off its control - pre-existing.
    - The two-column gutter on a `form-columns-2` fieldset is padding on the leading field, so it stayed put once the columns stacked and left the first field ending short of the second.
- The carry-over list assumes the four forms share HubSpot's standard internal field names (`firstname`, `lastname`, `email`, `company`, `message`). A form that names a field differently just won't carry that one - worth checking against the live forms.

**Not covered here**

- The events pages have a related problem, but not the same one: they render every session's form up front and toggle `hidden` on the panel wrappers, so nothing is destroyed - values just don't carry to the session you switch to. That fix belongs in the panel-toggle code rather than in this component.

### DIFF
--- a/content/contact/_index.md
+++ b/content/contact/_index.md
@@ -4,14 +4,14 @@ meta_desc: Contact Pulumi for any general inquiries, including requests for pric
 type: page
 layout: contact
 quick_links:
-    title: Looking for something quicker?
+    title: Quick links
     items:
-        - label: Start a trial
+        - label: Ready to try it?
           description: No sales call needed. Create an org and ship infrastructure in minutes.
           cta_label: Start a trial
           url: https://app.pulumi.com/signup?create-organization=1
 
-        - label: See it in action
+        - label: Want to see it first?
           description: Get a guided walkthrough from our team before you commit to anything.
           cta_label: Request a demo
           url: /request-a-demo/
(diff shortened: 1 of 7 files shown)


cnunciato (approved):
Nice work @jeffmerrick, and apologies for the surprise on this one. 

This looks great. The only thing worth mentioning (which is neither specific to this PR nor to the last one) is that because we replace the form when you switch between "reasons", it's easy to start filling something out, make another selection, and then lose everything and have to start over. If it were possible to capture the contents of an existing form before a new one is loaded, and then reapply whatever field values could be kept — first, last, company, email, and message, looks like — that'd be awesome. Definitely could be done as a follow-up (and FWIW, we have the same issue on the events pages now, since you can toggle between sessions), but would be nice if we could do it.

jeffmerrick:
@cnunciato That was an easy add so went ahead and did it!

cnunciato:
@jeffmerrick Amazing thank you!

PR #20933, opened 2026-08-17
Clear Segment's stale user identity on www when signed out of Pulumi Cloud

danbiwer:
### Proposed changes

Fixes the five **"User ID mismatch"** Intercom installation errors on www.pulumi.com — the blocker for enabling **Enforce Messenger Security**, which the Intercom customer portal requires.

Segment persists its signed-in identity in apex-domain (`.pulumi.com`) cookies plus a www-origin localStorage mirror, and its device-mode Intercom destination pushes that id into the Messenger unsigned on every page view. When the id is stale — the visitor signed out of the console, or switched accounts — it collides with the JWT-verified identity attached by `intercom-identity.ts` (#20547).

One file, `theme/src/ts/intercom-identity.ts`, clears Segment's user keys (`ajs_user_id`, `ajs_user_traits`; cookies and localStorage) at module evaluation, which runs before the consent manager loads analytics.js:

- **No console session** (hint cookie absent): clear any persisted Segment user.
- **Account switch** (persisted Segment id differs from the session's user id — the console rewrites the apex cookie on sign-in, but analytics.js reads the www-origin localStorage copy first): clear it; the current user re-attaches on their next console-driven identify. Both ids are `apitype.User.id`, so the comparison is sound.
- **Dead session** (hint cookie present, settings fetch 401s): the same cleanup runs in `signedOutCleanup`, additionally clearing analytics.js's in-memory user id.

Deliberately untouched: `ajs_anonymous_id` (pre-signup tracking), signed-in event attribution, and Segment configuration. Disabling the legacy device-mode Intercom destination in Segment is deferred to a follow-up with the Segment owners.

Rollout: stale browsers self-heal on their next www visit. When the install-error log (7-day window) drains to zero, enforcement can be enabled.

### DIFF
--- a/theme/src/ts/intercom-identity.ts
+++ b/theme/src/ts/intercom-identity.ts
@@ -2,8 +2,8 @@
 // backend-signed Messenger Security JWT fetched from app.pulumi.com. Rides the
 // consent-managed Segment lifecycle (analytics.ready, same as the head
 // partial's ad pixel): anonymous or non-consented visitors cost zero requests.
-// Segment boots the widget as usual — this only attaches (or clears) verified
-// identity.
+// Segment boots the widget as usual — this attaches (or clears) verified
+// identity and drops Segment's stale persisted user.
 
 // Local docs dev (make serve) pairs with a local console (devtool.py start);
 // everywhere else the console lives at app.pulumi.com.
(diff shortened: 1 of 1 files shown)


lucascrespo88 (approved):
LGTM

danbiwer:
@claude #update-review

PR #20951, opened 2026-08-18
Document the SCIM feature surface and fix the stale 40-char team-name limit

### LINKED ISSUE #20943: SCIM docs don't explain the feature surface, and repeat a stale 40-char team-name limit
## Problem

The SCIM docs are organized as three provider setup walkthroughs with no shared statement of what SCIM actually does. A reader can't answer basic questions ("can SCIM assign roles?", "does deprovisioning delete the user?", "will my IdP work?") without reading all three pages and inferring. Several cross-provider facts are stated on exactly one page, which makes them read as provider-specific quirks when they're universal.

Verified against the implementation in `pulumi/pulumi-service` (`cmd/service/api/scim/`, `cmd/service/api/{users,teams}_scim*.go`, routes at `cmd/service/main.go:1652-1692`).

Pages in scope:

- `/docs/administration/access-identity/scim/` (`_index.md`)
- `/docs/administration/access-identity/scim/okta/`
- `/docs/administration/access-identity/scim/entra/`
- `/docs/administration/access-identity/scim/onelogin/`
- `/docs/support/faq/scim/`

## Key context: SCIM is provider-agnostic

Unlike VCS integrations, there is **one** SCIM 2.0 implementation. Every IdP hits the same routes, schemas, attributes and limits. There is no per-provider config, no capability matrix, and no supported-provider enum in the codebase. Only the IdP-side setup differs.

The only two places an IdP affects server behavior are compatibility accommodations available to everyone: `GET /Users` is registered for search because some IdPs only search via GET, and a `userName` reactivation quirk is tolerated leniently.

Practical implication: any spec-compliant SCIM 2.0 IdP should work, not just the three we document. The docs currently read as if those three are the only options. (Flagging that this is a support-commitment statement — worth product sign-off on how strongly to word it.)

## Bugs to fix

- [ ] **Stale team-name limit.** All three provider pages warn that team names "must not exceed 40 characters". The real limit is **100** (`maxDefaultNameLength = 100`, enforced in `teams_scim.go:549`; column widened 40→100 by `migrations/328_bump_scim_display_name_size.up.sql` in Nov 2024). Stale for ~21 months. Locations: `okta.md:143`, `entra.md:33`, `onelogin.md` (group provisioning section).
- [ ] **Broken anchor.** `onelogin.md:20` links to `#known-limitations`, but the OneLogin page has no such section. Okta and Entra both do.
- [ ] **FAQ attribute list is incomplete.** The "Unknown path" answer lists `userName`, `displayName`, `givenName`, `familyName`, `active` — it omits `emails[type eq "work"].value`, which is supported and which the Entra guide tells you to mark as required.
- [ ] **Entra intro undersells scope.** It says "provisioning/deprovisioning of your users"; Okta and OneLogin say "users and groups". Entra supports groups, and the page already has a group-provisioning section.

## Add a capability section to the SCIM index page

`_index.md` is 32 lines and documents no feature surface at all. Proposed content:

### Supported

| Capability | Notes |
|---|---|
| User provisioning | Create, read, update (PUT/PATCH), search |
| User deprovisioning | **Soft only** — `active: false`. There is no `DELETE /Users`. |
| Group → Pulumi Team sync | Create, read, update, **delete**, membership add/remove/replace |
| PATCH | Supported for both Users and Groups |
| Filtering | `userName eq` (Users) and `displayName eq` (Groups) only; `eq` is the only operator |
| Pagination | **Users only.** Max 100/page. Group search returns all results in one response. |

### Supported attributes — this is the complete list

- **Users:** `userName` (immutable after creation), `displayName`, `name.givenName`, `name.familyName`, `emails[type eq "work"].value`, `active`
- **Groups:** `displayName`, `members`

Anything else returns `400 invalidPath`.

### Not supported

| | Notes |
|---|---|
| Role / admin provisioning | No `roles` or `entitlements`. Org role and team role can't be set via SCIM; SCIM users are always created as members and team members are always added as plain members. |
| `externalId` | Accepted on Users but never stored; Groups have no `externalId` field at all. This is why the Entra guide says to delete the `objectId`→`externalId` group mapping. |
| Bulk operations | `bulk: false` |
| Sorting | `sort: false` |
| ETag | `etag: false` |
| Password sync | `changePassword: false` |
| Enterprise User extension | Core 2.0 schemas only |
| Secondary emails | Only the primary work email is used |

### Editions

Business Critical (Pulumi Cloud) and self-hosted Enterprise.

## Promote single-page facts to the shared page

These are universal but currently appear on one page each:

- **Deprovisioning never deletes.** OneLogin's "Suspend (*DO NOT set to Delete*)" is the only place this appears, but it's true for every IdP — there is no `DELETE /Users` route at all. Okta and Entra readers never learn this.
- **`userName` is immutable after creation.** Okta's page explains the "Create only" mapping and Entra's explains "Only during object creation"; neither says *why*, and OneLogin doesn't mention it.
- **Group members must already be provisioned and active** before the group is pushed, or the request returns 400. Currently only in the Okta page and the FAQ.
- **`externalId` on groups is ignored** — documented only on the Entra page today, universal in fact.

## Also worth a warning

`GET /Groups` returns *all* teams matching the display-name filter, including Pulumi-local and GitHub-backed teams, not only SCIM-provisioned ones (`cmd/service/model/teams.go:515-523`). Since the docs explicitly encourage running local teams alongside SCIM teams, an IdP reconciling group state may see teams it doesn't own.

Group member validation also caps at 3000 SAML identities per query, which is an undocumented practical ceiling on team size.

## Related

- #20796 (Refresh Administration section of the site)

jkodroff:
Fixes #20943.

The SCIM docs were three IdP walkthroughs with no shared statement of what SCIM actually does, so a reader couldn't answer basic questions ("can SCIM assign roles?", "does deprovisioning delete the user?") without reading all three pages and inferring. Every claim was verified against `pulumi-service` @ `a5402c5c31`.

## Changes

- **`scim/_index.md`**: new `Capabilities`, `Supported attributes`, and `What SCIM does not manage` sections, plus `Behavior to plan for` — which promotes four facts that each previously lived on a single provider page (soft-delete-only deprovisioning, `userName` immutability, group members must be provisioned first, group `externalId` ignored). Also states up front that SCIM provisions users → org members and groups → teams.
- **Team-name limit 40 → 100** on all three provider guides. The column was widened in Nov 2024 (`maxDefaultNameLength = 100`, `validation.go:32`); the docs were ~21 months stale. Entra's warning also moved into its group-provisioning section, where Okta's and OneLogin's already were.
- **Dead `#known-limitations` anchor** on the OneLogin page fixed. The duplicated `Known Limitations` sections on Okta/Entra now link to the shared reference instead.
- **FAQ**: completed the "Unknown path" attribute list (added `emails[type eq "work"].value`, corrected to the `name.givenName` / `name.familyName` paths the server actually matches), and added a "Display name is too long" answer — the one documented failure mode with no FAQ coverage.
- **OneLogin**: added the SAML SSO prerequisite and the `userName` immutability guidance the other two guides already had.
- **Rendering fix**: indented `{{% notes %}}` blocks inside Okta's ordered list were splitting one six-step procedure into three lists (`</ol>…<ol start="3">`) with callouts orphaned between them. Converted to the `{{< notes >}}` form per AGENTS.md.

## Notes for reviewers

- The provider-agnostic framing is deliberately **descriptive** ("Pulumi implements a single SCIM 2.0 endpoint… only the IdP-side setup differs") rather than a support commitment. The stronger statement — that any spec-compliant SCIM 2.0 IdP should work — needs product sign-off.
- The issue's **self-hosted Enterprise** availability line was dropped as unverifiable; no self-hosted doc mentions SCIM. Happy to add it if someone can confirm.
- No hand-written editions section: `pulumi_cloud_feature: scim` already renders the Business Critical callout.
- `layouts/shortcodes/sso-scim-limits-info.*` also affects the **SAML index page** — intentional ("each Pulumi Cloud organization"), rendering verified on both.
- Removing the `Known Limitations` headings means external `#known-limitations` deep links will land at page top.

### DIFF
--- a/content/docs/administration/access-identity/scim/_index.md
+++ b/content/docs/administration/access-identity/scim/_index.md
@@ -36,7 +36,7 @@ Pulumi implements a single SCIM 2.0 endpoint. Every identity provider uses the s
 | Group-to-team synchronization | Yes (create, read, update, delete, and membership changes) |
 | `PATCH` | Yes, for both users and groups |
 | Filtering | `userName eq` for users and `displayName eq` for groups, and nothing else. Any other attribute or operator, such as `sw` or `co`, fails with a `400 invalidFilter` response |
-| Pagination | Users only, at a maximum of 100 per page. A request for a larger `count` fails with a `400` response. A group search is unpaginated and returns every match in one response |
+| Pagination | Users only, at most 100 per page. A request for a larger `count` fails with a `400` response. A group search is unpaginated and returns every match in one response |
 
 ## Supported attributes
 
--- a/content/docs/administration/access-identity/scim/_index.md
+++ b/content/docs/administration/access-identity/scim/_index.md
@@ -16,17 +16,89 @@ pulumi_cloud_feature: scim
 
 The [Pulumi Cloud](https://app.pulumi.com/signin) supports System for Cross-domain Identity Management (SCIM) 2.0 integration with different identity providers. SCIM enables you to manage your users and groups centrally in your Identity Provider (IdP) and then synchronize those users and groups to the Pulumi Cloud.
 
-    {{% notes type="info" %}}
-If desired, in addition to the SCIM-managed teams, one can also configure and manage Pulumi-local teams in the Pulumi Cloud. See [Teams](/docs/administration/organizations-teams/teams/) for how to configure teams in the Pulumi Cloud.
-    {{% /notes %}}
+SCIM provisions two things in Pulumi Cloud:
+
+- **Users** become [members of your Pulumi organization](/docs/administration/organizations-teams/organizations/), able to sign in through your identity provider.
+- **Groups** become [teams](/docs/administration/access-identity/rbac/teams/), and group membership becomes team membership. Grant those teams access to stacks, environments, and other entities with [RBAC](/docs/administration/access-identity/rbac/).
+
+Pulumi implements a single SCIM 2.0 endpoint. Every identity provider uses the same routes, schemas, and attributes, so only the IdP-side setup differs from one provider to the next. The guides at the end of this page cover popular providers, and everything documented on this page applies to each of them.
 
 {{% notes type="info" %}}
 {{< sso-scim-limits-info idp="your Identity Provider" >}}
 {{% /notes %}}
+
+## Capabilities
+
+| Capability | Support |
+|---|---|
+| User provisioning | Yes (create, read, update, and search) |
+| User deprovisioning | Yes, but soft only (see [Deprovisioning never deletes a user](#deprovisioning-never-deletes-a-user)) |
+| Group-to-team synchronization | Yes (create, read, update, delete, and membership changes) |
+| `PATCH` | Yes, for both users and groups |
+| Filtering | `userName eq` for users and `displayName eq` for groups. `eq` is the only supported operator |
+| Pagination | Users only, up to 100 results per page. A group search returns every match in a single response |
+
+## Supported attributes
+
+The lists below are complete. Pulumi implements the core SCIM 2.0 schemas only, so a request that adds or updates any attribute not listed here fails with a `400 invalidPath` response. See [Unknown path](/docs/support/faq/scim/#unknown-path) in the SCIM FAQ.
+
+For users:
+
+- `userName` (immutable after creation, see [Usernames cannot change](#usernames-cannot-change))
+- `displayName`
+- `name.givenName`
+- `name.familyName`
+- `emails[type eq "work"].value`
+- `active`
+
+For groups:
+
+- `displayName`
+- `members`
+
+## What SCIM does not manage
+
+| Feature | Notes |
+|---|---|
+| Roles and administrator status | Pulumi does not implement the `roles` or `entitlements` attributes. A user provisioned through SCIM is always created as an organization member, and a group member is always added to the team as a plain member. Promote users to admin in the Pulumi Cloud console |
+| `externalId` | Accepted on users but never stored, and not part of the group schema at all. Remove any `externalId` mapping for groups in your IdP |
+| Bulk operations | Not supported. Provision resources one request at a time |
+| Sorting | Not supported |
+| ETags | Not supported |
+| Password synchronization | Not supported. Credentials stay in your IdP |
+| Enterprise User extension | Not supported. Only the core SCIM 2.0 user and group schemas are implemented |
+| Secondary emails | Not supported. Only the primary work email is used |
+
+## Behavior to plan for
+
+These apply to every identity provider, regardless of which guide you follow.
+
+### Deprovisioning never deletes a user
+
+Pulumi has no endpoint for deleting a user through SCIM. Deprovisioning sets `active` to `false`, which removes the user's access to the organization while preserving their account and its history. Configure your IdP to suspend or deactivate users rather than delete them.
+
+Teams are different: a group deleted in your IdP does delete the corresponding Pulumi team.
+
+### Usernames cannot change
+
+A Pulumi username is immutable once the account exists, because it identifies the user across every organization they belong to. If your IdP pushes a changed `userName` on an update, the request fails with a `400 immutability` response. Map `userName` so that it applies only when the account is created, not on later updates. Each provider guide covers where to set this.
+
+### Group members must already be provisioned
+
+Pulumi validates every member of a group before creating or updating the corresponding team. If any member has not yet been provisioned into your organization, or has been deactivated, the entire request fails with a `400` response and no membership changes are applied. Provision users before pushing the groups that contain them.
+
+Member validation reads up to 3000 identities per request, which is a practical ceiling on the size of a SCIM-managed team.
+
+{{% notes type="warning" %}}
+A group search returns every team in your organization whose display name matches the filter, not only the teams that SCIM created. Pulumi-local teams and teams backed by a GitHub organization appear in the results alongside SCIM-provisioned ones. If your IdP reconciles group state, it may see teams it does not own.
+{{% /notes %}}
+
+## Next steps
 
 To set up synchronization between Pulumi and your SAML 2.0 identity provider, refer to one of our example guides:
 
 - [Microsoft Entra ID (formerly Azure Active Directory)](/docs/administration/access-identity/scim/entra/)
 - [Okta](/docs/administration/access-identity/scim/okta/)
 - [OneLogin](/docs/administration/access-identity/scim/onelogin/)
-- [FAQ](/docs/support/faq/scim/)
+
+For the provisioning errors you are most likely to hit and how to resolve them, see the [SCIM FAQ](/docs/support/faq/scim/).
--- a/content/docs/administration/access-identity/scim/_index.md
+++ b/content/docs/administration/access-identity/scim/_index.md
@@ -35,8 +35,8 @@ Pulumi implements a single SCIM 2.0 endpoint. Every identity provider uses the s
 | User deprovisioning | Yes, but soft only (see [Deprovisioning never deletes a user](#deprovisioning-never-deletes-a-user)) |
 | Group-to-team synchronization | Yes (create, read, update, delete, and membership changes) |
 | `PATCH` | Yes, for both users and groups |
-| Filtering | `userName eq` for users and `displayName eq` for groups. `eq` is the only supported operator |
-| Pagination | Users only, up to 100 results per page. A group search returns every match in a single response |
+| Filtering | `userName eq` for users and `displayName eq` for groups, and nothing else. Any other attribute or operator, such as `sw` or `co`, fails with a `400 invalidFilter` response |
+| Pagination | Users only, at a maximum of 100 per page. A request for a larger `count` fails with a `400` response. A group search is unpaginated and returns every match in one response |
 
 ## Supported attributes
 
@@ -60,7 +60,7 @@ For groups:
 
 | Feature | Notes |
 |---|---|
-| Roles and administrator status | Pulumi does not implement the `roles` or `entitlements` attributes. A user provisioned through SCIM is always created as an organization member, and a group member is always added to the team as a plain member. Promote users to admin in the Pulumi Cloud console |
+| Roles and administrator status | Pulumi does not implement the `roles` or `entitlements` attributes, so no SCIM request can set one. A user is always created as an organization member, and a group member is always added to the team as a plain member. Change either afterwards in the Pulumi Cloud console; SCIM does not overwrite a role you set there |
 | `externalId` | Accepted on users but never stored, and not part of the group schema at all. Remove any `externalId` mapping for groups in your IdP |
 | Bulk operations | Not supported. Provision resources one request at a time |
 | Sorting | Not supported |
@@ -75,9 +75,20 @@ These apply to every identity provider, regardless of which guide you follow.
 
 ### Deprovisioning never deletes a user
 
-Pulumi has no endpoint for deleting a user through SCIM. Deprovisioning sets `active` to `false`, which removes the user's access to the organization while preserving their account and its history. Configure your IdP to suspend or deactivate users rather than delete them.
+Pulumi has no endpoint for deleting a user through SCIM. Deprovisioning sets `active` to `false`, which removes the user's access to the organization while preserving their account and its history, so the change is reversible by reactivating the user in your IdP. Configure your IdP to suspend or deactivate users rather than delete them.
 
-Teams are different: a group deleted in your IdP does delete the corresponding Pulumi team.
+Teams do not work this way. See [Deleting a group deletes the team](#deleting-a-group-deletes-the-team).
+
+### Deleting a group deletes the team
+
+Unlike users, teams are deleted outright. Deleting a group in your identity provider deletes the corresponding Pulumi team, and the access you granted that team goes with it: the team's roles, and its permissions on stacks and environments. Members themselves are not deleted and remain in your organization.
+
+Two things to plan for:
+
+- The deletion is not recorded in your organization's audit log, because a SCIM request has no acting user to attribute it to.
+- A delete request identifies the team by id, and Pulumi does not check that the id belongs to a SCIM-provisioned team. Any team in the organization can be deleted this way, including one created in the Pulumi Cloud console.
+
+To keep a team while removing its members, empty the group rather than deleting it.
 
 ### Usernames cannot change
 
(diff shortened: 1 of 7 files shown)


jkodroff:
@claude #update-review

Both 🚨 items are confirmed against `pulumi/pulumi-service` @ `a5402c5c31`. Sources below. One of the two findings is right about a defect I introduced — the error string — and verifying the second turned up an exception that isn't documented yet.

## 1. The 100-character team-name limit — confirmed

| What | Where |
|---|---|
| The constant | `cmd/service/api/validation.go:32` — `maxDefaultNameLength = 100` |
| Enforced on group create/replace | `cmd/service/api/teams_scim.go:549` — `if len(displayName) > maxDefaultNameLength { return scim.BadRequest(scim.ErrorTypeInvalidValue, scimTeamDisplayNameTooLong) }` |
| Enforced on `PATCH` | `cmd/service/api/teams_scim.go:615` — same check on the patched `displayName` |
| The message | `cmd/service/api/teams_scim.go:43` — `scimTeamDisplayNameTooLong = fmt.Sprintf("Display name is too long. It must be %d characters or less", maxDefaultNameLength)` |

The 40 → 100 change was the `SCIMTeams.display_name` column widening in `migrations/328_bump_scim_display_name_size.up.sql`, landed 2024-11-14 in pulumi/pulumi-service#23999. The docs have been stale since.

### You were right about the error string — I had the format wrong

Good catch, and it matters for exactly the reason you gave. The message text is correct, but I wrapped it in a plain-text envelope (`Bad Request: …`) copied from the neighbouring FAQ entry. The API does not emit that. `scim.BadRequest` (`cmd/service/api/scim/util.go:165`) goes through `toSCIMError` (`:179`) and serializes `ErrorResponse` (`:139-150`), so the wire format is JSON:

```json
{
    "status": "400",
    "response": {
        "schemas": ["urn:ietf:params:scim:api:messages:2.0:Error"],
        "status": "400",
        "scimType": "invalidValue",
        "detail": "Display name is too long. It must be 100 characters or less"
    }
}
```

Note `status` is a JSON **string**, not a number (`// NB. This is a string in the RFC.`), and `scimType` is `invalidValue`. I'll fix the FAQ entry to this shape so it matches the other four answers on that page.

## 2. Group deletion cascading to team deletion — confirmed, and it's a hard delete

`DELETE /Groups/{groupID}` is registered at `cmd/service/main.go:1691` → `DeleteSCIMTeamHandler` (`cmd/service/api/teams_scim.go:185`) → `deleteTeamImpl` (`cmd/service/api/teams.go:705`) → `DeleteTeamByName(..., queueCleanup: true)` → `cmd/service/model/teams.go:1183`, which is a real SQL `DELETE` against the `Teams` table. No soft-delete, no tombstone.

What goes with it:

- **Team roles** — CASCADE-deleted with the team row; the team's inline ABAC role is deleted explicitly (`model/teams.go:1194`).
- **Team-owned services** deleted, and the team is removed from other services' member lists (`model/teams.go:1201-1218`).
- **Stack and environment grants** — a `TeamPermissionsBackfillCleanup` background activity is queued (`model/teams.go:1220-1232`), so access granted through that team goes away.
- **Users are not touched.** They stay org members; only the team and its grants go. So the asymmetry the page describes is real: users soft, teams hard.

One operational detail worth knowing: **SCIM team deletion writes no audit log entry.** `DeleteSCIMTeamHandler` passes `nil` for the audit logger (`teams_scim.go:201-202`), commented "There's no user associated with the SCIM requests, so we skip writing the audit log event." A team can therefore vanish with no audit-log trace.

## 3. Exceptions — including the non-SCIM-member case

**A non-SCIM member does not block deletion.** `DeleteSCIMTeamHandler` never looks at membership. The only guard between the request and the `DELETE` is `validateTeamBelongsToOrg` (`teams_scim.go:560-580`), which compares `team.OrgID` to the requesting org and nothing else. The team goes regardless of who is in it.

**Can a non-SCIM user even be in a SCIM team?** Not through Pulumi Cloud: `cmd/service/api/teams.go:1064-1068` rejects add/remove for any non-Pulumi team — *"Adding and removing team members is only supported for Pulumi teams"* → `400 Unable to add/remove members for this type of Team.` And not through SCIM either, since members must already have an active SAML identity in the org (`teams_scim.go:127-181`) or the request fails `400 invalidValue`. So membership is SAML-identity-scoped from both directions.

**But there is a real exception to "SCIM users are always plain members":** role promote/demote *is* permitted on SCIM teams through the console (`teams.go:1076-1099`, gated on `PulumiFeatureTeamMemberRoleManagement`) — only GitHub-backed teams are excluded. SCIM itself never sets a role, but an admin can change a SCIM team member's role in Pulumi Cloud afterwards. My "always added as plain members" row is accurate about SCIM and misleading about the end state; I'll qualify it.

**The exception that actually worries me** — and I think this deserves to be in the docs rather than just this thread:

> `DELETE /Groups/{groupID}` does not check that the target team is SCIM-provisioned.

`scimGroupPreamble` resolves the group with a bare `GetTeamByID` (`teams_scim.go:510`), and `validateTeamBelongsToOrg` only compares org IDs. There is no team-kind check anywhere on the delete path. Combined with `GET /Groups` returning **every** team whose display name matches the filter — `ListTeamsByDisplayNameAndOrganizationID` (`cmd/service/api/scim/groups_read.go:137`) filters on display name and org only, so Pulumi-local and GitHub-backed teams come back too — an IdP that reconciles group state can discover a Pulumi-local team it does not own and hard-delete it.

That is the missing mitigation you asked for on the `[L93]` callout. I'll restructure per your suggestion:

- Split the team-deletion cascade out of the "Deprovisioning never deletes a user" section into its own `### Deleting a group deletes the team`, saying plainly that it is a hard delete, that team grants go with it, and that no audit-log entry is written.
- Extend the group-search warning with the concrete mitigation — scope IdP reconciliation to SCIM-provisioned teams, or name SCIM-managed teams distinctly — and state the consequence, which is deletion of a team the IdP does not own, not merely visibility.

## On the ⚠️ low-confidence items

The source for all of them is the service implementation, not a spec — that is what I checked each claim against. For the record, with citations, so a future reviewer can re-verify:

- **`eq` is the only operator, on `userName` / `displayName`** — `cmd/service/api/scim/users_read.go:274-297`, `validateSupportedUsersFilter` / `validateSupportedGroupsFilter`. The list is exhaustive, and it is enforced rather than ignored: an IdP that sends `sw` or `co` gets `400 invalidFilter` with *"Unsupported filter operator: sw"*, and an unsupported attribute gets *"Unsupported filter attribute: …"*. I'll add that to the table, since you're right that the reader currently gets no hint.
- **User pagination: 100 is a hard maximum, not a default.** `cmd/service/api/scim/discovery.go:28` `maxResultsPerPage = 100`; `users_read.go:329-330` rejects a larger `count` with *"Invalid value for 'count'. Maximum results per page is 100."* You were right to ask — "up to 100" is ambiguous and I'll say "maximum".
- **Group search is unpaginated** — `groups_read.go:129-131`, verbatim: *"We currently do not support the pagination part of the spec for now. We are essentially ignoring the pagination parameters and just returning all the teams associated with the organization."*
- **bulk / sort / etag / changePassword all `false`; patch and filter `true`** — `discovery.go:49-81`, the literal `ServiceProviderConfig`.
- **Core schemas only, no Enterprise User extension** — `discovery.go:16-23` registers only the core 2.0 User and Group URNs.
- **`externalId` accepted on users, never stored** — `pkg/apitype/scim/users.go:47` parses it; nothing persists it. Groups have no `externalId` field at all, which is why the Entra guide removes that mapping.
- **The six user attributes** — `pkg/apitype/scim/core.go:61-66`; anything else hits `Unknown path` at `users_scim_mapping.go:447`.
- **3000-identity member validation** — `teams_scim.go:124-125`, `maxSAMLIdentitiesPerQuery = 3000`, with the TODO referencing pulumi/pulumi-service#30222 to paginate it. On what the reader sees when crossing it: members beyond the window are not silently dropped — they fail validation and the whole request returns `400 invalidValue` with `Cannot add invalid members to team. Invalid member ids: …` (`teams_scim.go:85-86`). I'll say that instead of "a practical ceiling".

Pushing the doc changes shortly.

jkodroff:
@claude #update-review

Pushed `95802ec` with the corrections. Every outstanding and low-confidence item is now either fixed or answered below with a source. Sources are `pulumi/pulumi-service` @ `a5402c5c31`, which is the origin for every behavioral claim on the page — that's the answer to the standing "name the source you wrote this from" ask.

## 🚨 Outstanding — both resolved

**[L135] The 100-character limit.** Confirmed: `cmd/service/api/validation.go:32` (`maxDefaultNameLength = 100`), enforced at `teams_scim.go:549` and `:615`, message at `teams_scim.go:43`. Raised from 40 by `migrations/328_bump_scim_display_name_size.up.sql`, 2024-11-14, pulumi/pulumi-service#23999.

You were right about the error string, and it was wrong in a way that mattered — I'd copied a plain-text envelope from the answer above it. `scim.BadRequest` → `toSCIMError` (`cmd/service/api/scim/util.go:165,179`) serializes `ErrorResponse` (`:139-150`), so it is JSON, `status` is a **string**, and `scimType` is `invalidValue`. Fixed. The neighbouring "Cannot add invalid members" answer had the same invented format, so I corrected that one too rather than leave the page showing two shapes for one API.

**[L80] Group deletion cascading to team deletion.** Confirmed, and it is a hard delete: `main.go:1691` → `DeleteSCIMTeamHandler` (`teams_scim.go:185`) → `deleteTeamImpl` (`teams.go:705`) → `model/teams.go:1183`, a SQL `DELETE`. Team roles CASCADE, team-owned services are deleted, and a `TeamPermissionsBackfillCleanup` activity is queued (`model/teams.go:1194-1232`), so stack and environment grants go with it. Users are untouched.

Restructured as you suggested, into its own `### Deleting a group deletes the team`. Verifying it also turned up two things worth documenting that neither of us had:

- **No audit-log entry.** `teams_scim.go:201-202` passes `nil` for the audit logger — *"There's no user associated with the SCIM requests, so we skip writing the audit log event."*
- **The delete path does not check that the target is a SCIM team.** `scimGroupPreamble` resolves it with a bare `GetTeamByID` (`teams_scim.go:510`) and the only guard is `validateTeamBelongsToOrg` (`:560-580`), which compares org ids. Any team in the org can be deleted by id — including one created in the console.

## ⚠️ Low-confidence — answered

- **[L37-39] Filtering.** The operator list is exhaustive and enforced, not advisory: `validateSupportedUsersFilter` / `validateSupportedGroupsFilter` (`scim/users_read.go:274-297`) reject anything but `eq`, and any attribute but `userName` / `displayName`, with `400 invalidFilter`. Your point stands that the table gave no hint — the row now says an `sw` or `co` filter fails, rather than leaving the reader to infer it.
- **[L39] Group search pagination.** `scim/groups_read.go:129-131`, verbatim: *"We currently do not support the pagination part of the spec for now. We are essentially ignoring the pagination parameters and just returning all the teams associated with the organization."* Stated as unpaginated, next to the 3000-identity note.
- **[L39] User pagination.** A hard maximum, not a default — you were right that "up to" was ambiguous. `maxResultsPerPage = 100` (`scim/discovery.go:28`); a larger `count` is rejected at `users_read.go:329-330` with *"Invalid value for 'count'. Maximum results per page is 100."* Row now says maximum, and that a larger request fails.
- **[L64] `externalId`.** `pkg/apitype/scim/users.go:47` parses it on users; nothing persists it. Groups have no `externalId` field at all, which is the asymmetry you flagged as the actionable half — that's exactly why the Entra guide deletes the `objectId` → `externalId` group mapping.
- **[L64-70, L66, L67, L69] Enterprise User extension, sorting, ETags.** All from the literal `ServiceProviderConfig` at `scim/discovery.go:49-81`: `Bulk`, `ChangePassword`, `ETag`, and `Sort` are `featureNotSupported`; `Filter` and `Patch` are supported. Core User and Group URNs only (`discovery.go:16-23`).
- **[L78 ×2] Soft deprovisioning.** There is no `DELETE /Users` route at all (`main.go:1671-1691` registers one for Groups and none for Users), so `active: false` is the only path. Added the reversibility half you singled out — reactivating in the IdP restores access.
- **[L90] The 3000 figure.** `teams_scim.go:124-125`, `maxSAMLIdentitiesPerQuery = 3000`, with a TODO referencing pulumi/pulumi-service#30222 to paginate it. On your question about what the reader sees: not silent truncation. Members outside the window fail validation and the whole request is rejected `400 invalidValue` (`teams_scim.go:85-86`). Said so, instead of leaving "a practical ceiling" ambiguous.
- **[L93] The group-search warning.** Confirmed — `ListTeamsByDisplayNameAndOrganizationID` (`scim/groups_read.go:137`) filters on display name and org only. Added the mitigation you asked for, and sharpened the consequence: because the delete path doesn't check team kind either, the risk is deletion of a team the IdP doesn't own, not just visibility.
- **[L44] entra.md, SCIM token retrieval.** Pre-existing text this PR only re-indented, matching `okta.md` verbatim. No change, as you noted.
- **[L32-34] okta.md, `SCIM_PROVISIONING` gating.** Okta-side vendor process, pre-existing, only re-indented. Left alone — I have no better source than Okta's own docs, and inventing certainty here would be worse than leaving it.
- **[L126-133] The "Display name is too long" answer.** Covered by the 🚨 item above; number and error string were confirmed and corrected together.

## One claim I'm revising rather than defending

The "Roles and administrator status" row said team members are *always* plain members. True of SCIM, but misleading about the end state: `teams.go:1076-1099` permits promote/demote on SCIM teams through the console (only GitHub-backed teams are excluded), gated on `PulumiFeatureTeamMemberRoleManagement`. Add/remove is still rejected for SCIM teams (`teams.go:1064-1068`). The row now says SCIM can't set a role and won't overwrite one set in the console.

## Style suggestions

Taken: the two `click` → `select` on entra.md L61 and the `prior to` wordiness on the FAQ (both via the suggestions Josh accepted), plus the `as shown below` directional reference on the same entra line, and the trailing period on the FAQ heading — I removed it from all three sibling `### A failure occurred…` headings rather than leave the new one inconsistent with its neighbours. Hugo generates identical anchors with or without the period, and nothing links to them, so no redirects are needed.

One I'm not taking: Vale flags `maximum` as wordy in the pagination row. It's there deliberately — "up to" was the ambiguity you asked me to remove.

jkodroff:
@claude #update-review

Fixed in `694fd9c`. You were right that the page contradicted itself, and right that I caused it.

**Resolved differently than you proposed, though**, on the outer `status`. Your suggestion was to quote the inner field and keep the outer bare on the theory that the outer wrapper is this page's own annotation of the HTTP status rather than part of the payload. It isn't — it's a real serialized field:

```go
type ErrorResponse struct {
	Status   string                   `json:"status"` // NB. This is a string in the RFC.
	Response scimErrorResponseDetails `json:"response"`
}
```

`cmd/service/api/scim/util.go:139-142`, and `toSCIMError` (`:179-192`) fills both from the same `fmt.Sprintf("%d", status)` and returns `rest.ClientErrorAsJSON(status, resp)` — the entire `ErrorResponse`, outer field included, is the response body. So both are strings on the wire, and the fix is to quote all twelve rather than eight.

The RFC comments in the struct are the reason: [RFC 7644 §3.12](https://datatracker.ietf.org/doc/html/rfc7644#section-3.12) defines the error `status` as a string, and the second field carries `// NB. Yes, this is redundant... as per the RFC.` The redundancy is deliberate spec compliance, not a documentation wrapper — which is exactly why the four pre-existing blocks were wrong in the same direction, not merely inconsistent.

All six now agree, and they agree with the wire format. Your underlying point stands and was the useful part: readers paste these into log matchers, so `"400"` vs `400` inside one page costs someone real time — it just needed correcting toward the string in both positions.

jkodroff:
@claude #update-review

Fixed in `694fd9c`. You were right that the page contradicted itself, and right that I caused it.

**Resolved differently than you proposed, though**, on the outer `status`. Your suggestion was to quote the inner field and keep the outer bare on the theory that the outer wrapper is this page's own annotation of the HTTP status rather than part of the payload. It isn't — it's a real serialized field:

```go
type ErrorResponse struct {
	Status   string                   `json:"status"` // NB. This is a string in the RFC.
	Response scimErrorResponseDetails `json:"response"`
}
```

`cmd/service/api/scim/util.go:139-142`, and `toSCIMError` (`:179-192`) fills both from the same `fmt.Sprintf("%d", status)` and returns `rest.ClientErrorAsJSON(status, resp)` — the entire `ErrorResponse`, outer field included, is the response body. So both are strings on the wire, and the fix is to quote all twelve rather than eight.

The RFC comments in the struct are the reason: [RFC 7644 §3.12](https://datatracker.ietf.org/doc/html/rfc7644#section-3.12) defines the error `status` as a string, and the second field carries `// NB. Yes, this is redundant... as per the RFC.` The redundancy is deliberate spec compliance, not a documentation wrapper — which is exactly why the four pre-existing blocks were wrong in the same direction, not merely inconsistent.

All six now agree, and they agree with the wire format. Your underlying point stands and was the useful part: readers paste these into log matchers, so `"400"` vs `400` inside one page costs someone real time — it just needed correcting toward the string in both positions.

CamSoper (approved):
I read it all, I don't have anything to add. :shipit:

CamSoper:
@claude #update-review

PR #20274, opened 2026-07-14
Generate pre-built policy pack pages from the Pulumi Cloud API

jkodroff:
The 16 policy listing pages under /docs/reference/pre-built-policy-packs/ were hand-maintained Markdown tables that drifted as soon as a pack was republished -- cis/aws.md alone was a 48 KB file with a 115-row table pasted in by hand. Pulumi Cloud already stores full metadata for every published pack, so the pages are now generated from it.

## Changes

- `scripts/fetch-policy-packs.js` pulls each pack's policies from the Pulumi Cloud REST API into `data/policy_pack_policies/<pack>.json`
- `data/policy_packs.yaml` is the editorial allowlist: which packs get a page, what they're called, and a required `source:` repo path so an internal-use pack can't be documented by accident
- `content/docs/reference/pre-built-policy-packs/_content.gotmpl` generates one page per entry -- 23 generated pages, 2394 policies
- `.github/workflows/update-policy-packs.yml` refreshes the data nightly and opens an auto-merging PR when a pack is republished

Pages now carry severity, enforcement level, per-control specifications, and full remediation steps, and state the exact pack version they describe. Adds pages for the 7 published packs that had none (CMMC/AWS, HITRUST/AWS Native, ISO 27001 Azure + GCP, NIST Azure, PCI DSS Azure + GCP), and fixes three drifts the migration surfaced.

All 16 existing page URLs are preserved, so no aliases or redirects are needed.

PR #20990, opened 2026-08-19
content-review: unstarve the reverify rotation; make the self-describing exclusion actually exclude

CamSoper:
The 2026-08-19 nightly claims re-verify checked **one** entity, that one came back demoted, and `signal-health` read 1-of-1 inconclusive as "no conclusive results" for the whole lane. Replaying the run against the S3 state as it stood at 07:21 reproduces it exactly — 54 volatile entities, 3 marked, **51 eligible**, `count=25`.

## What was wrong

**The rotation starves on the remainder.** `tonight_chunk` took a contiguous slice, so the last chunk of a rotation is whatever `len % count` happens to be. `ceil(51/25) = 3` chunks sized **25 / 25 / 1** — one night in three spends 4% of the budget and reports a sample of one. The slice is strided now, so sizes differ by at most one (51 → 17/17/17). Coverage and determinism are unchanged; only the grouping differs, and each entity is verified independently so nothing downstream cares.

**A sample of one became a lane-wide verdict.** `observe_reverify` degraded on `n_inconclusive == n_checked` with no floor. An all-inconclusive night under `MIN_INCONCLUSIVE_SAMPLE` (5) is now no observation at all, leaving the prior state standing — the same treatment a missing report already gets. `skipped` / `no_snapshots` are untouched: a lane that could not run says so directly and never depends on sample size.

**The report couldn't explain itself.** Added `meta.n_marked` and `meta.n_eligible`. Without them a small `n_due` is ambiguous between "the pool is nearly empty" and "the rotation handed tonight a short chunk", and the 2026-08-19 report could not distinguish those.

## The self-describing exclusion had never excluded anything

`#20851` added `SELF_DESCRIBING_RE` to keep page-internal counts out of the nightly pool. Against the live index it matched **0 of 53** volatile entities, for two independent reasons:

- `PRICING_LIMIT_RE` vetoed it on the word **"plans"** in ``` `pulumi up` plans to create 2 resources ``` — Pulumi's CLI vocabulary, not an edition. `plans?` now carries a negative lookahead for the infinitive, separating the verb and noun senses. The veto keeps its full breadth otherwise, so the existing `limit` / `free` / `seat` cases still veto (their tests are unchanged and green).
- The `output <verb>` alternation listed only reporting verbs, so ``` the `storageAccountName` output has the value ... ``` sailed through. Added the copulas.

## What I deliberately did *not* do

- **No page-artifact rule.** Matching "the table" / "this page" as a subject would also reclassify `This page lists all 139 policies in the NIST SP 800-53 pack`, which the suite pins as `policy-pack count stays volatile`. That's a deliberate decision encoded in a test and not mine to reverse silently.
- **No `version` escape.** The index does carry mis-typed `version` entities that burn calls for nothing (`running-git` is a `git tag` sequence, `error-tag` is an error message, `range-matches` is a syntax explainer). But a regex tuned to catch them also drops `the Go example imports .../sdk/v7/...` and `requires applying a semver Git tag (for example, v1.2.3)`, where the pinned major is a real outside-world fact — and `source_hint` is empty on **every** version claim in the index, so there's nothing to discriminate on by the time `entity_key` sees them. That fix belongs in claim extraction.

## Verification

Replayed against the live claims index:

| | entities | eligible | chunk sizes | `n_due` 08-19 |
|---|---|---|---|---|
| before | 54 | 51 | 25 / 25 / 1 | **1** |
| after | 51 | 48 | 24 / 24 | **24** |

53 → 50 volatile entities on the current index. The three drops are all example-output claims — ``bucketName output is set to``, ``storageAccountName output has the value``, and the ``plans to create 2 resources`` preview transcript. No product facts lost: the real pack counts stay, and so do the page-self-describing twins (see "what I deliberately did not do").

`make test-review-pipeline` passes (all 6 harnesses, all 17 `--self-test` suites). New cases pin both regex bugs and the starved night.

---

## Second commit: generated-page findings route upstream

The pre-built policy-pack tree stopped having editable files on 2026-07-16 (#20274 swapped the `.md` files for a content adapter). Its claims stayed in the nightly index — last reviewed 2026-07-15, **one day** before the conversion.

So the three findings the lane produced there in August marked pages that cannot be edited, and `already_marked` then dropped those entities from the rotation permanently. The three most valuable results this lane has produced — real user-facing misstatements, now filed as `policy-packs-internal#204` — became the only three facts in the system guaranteed never to be re-checked.

**This does not retire the tree from the sweep.** `update-policy-packs.yml` refetches those pages nightly, which keeps them *fresh*, not *correct* — it republishes an upstream error faithfully forever. This lane is the only thing checking those descriptions against AWS's own documentation.

Findings on a page with no markdown source now:

- are **never marked** and never boosted, so they stay in the rotation and get re-checked every cycle;
- are **reported every run** (`meta.n_upstream`, `meta.upstream_entities` with the tracking issue, plus a run-log line each);
- **notify only on a change.** `references/upstream-claims.yaml` records what is already filed upstream. A known, still-contradicted finding is silent. A finding nobody has filed pages #docs-ops — and so does one that *starts verifying clean*, which is the upstream fix landing. That last event was structurally unreportable before, because a marked entity was never checked again.

The existing stale-summary Slack line promised *"affected pages are boosted into the next content-review sweep"*, which was never true for these. They get their own line naming the upstream repo.

`n_inconclusive` is now computed against all contradicted results rather than only the markable ones — otherwise every upstream finding would book as inconclusive and eventually degrade the lane's health signal for working correctly.

Tier 0 for `content/docs/reference/pre-built-policy-packs/`: the entry that should have landed with #20274. It only stops the *fix* lane spending a slot on an uneditable page, and the note says explicitly that the fact-check must keep running.

### Ledger state repaired alongside (S3, versioned)

- Three dead markers cleared — those entities re-enter the rotation, first re-check within 2 nights.
- `attempts` reset on `migrating-to-pulumi.md`. All three of its failures (Jul 20/21/22) were the same `verify-fix-scope` rejection — `uncovered hunk ...:L38`, same line each time — which #20970's pre-flight self-check addresses.

Live state now: 0 markers, 0 capped pages, 0 permanently-excluded entities.

### A gap this turned up

Generated trees are invisible to the fact-check by construction: tier 0 → never selected → no claims snapshot → never verified. Measured against the live index:

| tree | claims snapshots |
|---|---|
| `iac/cli/commands/` (248 pages) | **0** |
| `esc/cli/commands/` | **0** |
| `reference/pkg/` (SDK API ref) | **0** |
| `reference/cloud-rest-api/` (OpenAPI) | **0** |
| `pre-built-policy-packs/` | 12 |

The policy-pack tree is in the pool *only* by the one-day accident above — and it is the sole source of every confirmed misstatement the lane has found. Not addressed here; worth its own issue.

---

## Later commits: the seams, not the parts

Everything above came out of one starved night. Working back from it turned up a pattern — this pipeline's failures live in its seams and nothing was positioned to see them — so these close the seams.

**Classifiers now measured against real data** (`test_corpus_classifiers.py`). Every other test here feeds them invented strings. That is how three separate changes shipped inert: #20851's exclusion matched 0 of 53 entities for a month, the `vars.X != '0'` gate kept whole lanes dark, and `upload-artifact` silently uploaded nothing. All three passed their tests. This runs the classifiers over a checked-in sample of the live index (34 pages, 683 claims) and asserts on distributions. Verified by reintroducing the #20851 bug — unit tests pass, this fails.

**Silence is now an alarm.** Every signal is reported *by* a job, and nothing reports on a job that never ran, so a dead lane sits at whatever it last said — usually `ok`. Each signal now declares its cadence, and any run flags a signal nobody has written to in too long. It deliberately does not override an existing degradation: that finding is more specific and already on the alert clock.

**Findings persist as JSON** (`record-page-findings.py`, new `findings/` prefix). They previously survived only as a ledger integer and PR-body prose, which `build-glowup-backlog.py` recovered by parsing markdown back into data — making the PR body load-bearing, failing silently on a renamed heading, and reaching only the latest PR because branches are reused. Nothing new is asked of the model: the findings are already the pre-step artifacts (reused via `compose-pr-body.collect()`, so they cannot drift), the disposition is already `applied[]`, and deferred is the difference. The prose scrape still merges in for the model's reason text.

**#20970's missing half.** Its message describes two guards against re-reviewing a just-fixed page; only one shipped. `STALE_BOOST_COOLDOWN_DAYS` is the other — narrower than the message implies, because a marker the review *saw and left* is real drift that must keep boosting. Only a marker written *after* a completed review is the echo.

**Glow-up throughput** (Cam): 2/night, cap 5 → 10. The cap test now reads the constant instead of pinning a literal.

Plus: dropped a tier-0 rule protecting `content/docs/esc/cli/commands/`, a directory that does not exist.

### Live state changed alongside

Attempt counter reset on `migrating-to-pulumi.md`; three dead markers cleared; `GLOWUP_COUNT` set to `2`. All reversible — the bucket is versioned.

### Follow-up filed

#20996 — report-only mode, so the 248 CLI command pages get fact-checked at all.


CamSoper:
@claude #update-review

All three findings actioned — one resolved as a real bug, two resolved with changes, the triaged batch refuted.

**🚨 [L871] `signal-health.py` — self-test gate placement. Resolved; you were right, and it was the most embarrassing possible place to have this bug.**

Verified before fixing: breaking one of the new assertions exited **0**. After moving the gate below every block, the same mutation exits **1**. So those eight assertions genuinely could not have failed CI, in the PR whose stated purpose is catching logic that ships inert.

Your read on it being a slip rather than a pattern is correct — I appended the blocks with a text anchor that matched the success print, which put them past the gate. `reverify-claims.py` has the same shape with its gate last.

**⚠️ [L237] `claims-reverify.yml` — two definitions of "new". Resolved, and worth more than the low-confidence label.**

The point that the workflow was leaning on a check in a different file to keep its own two numbers agreeing is the part that matters, independent of whether they can diverge today. Rather than filtering on a second derived property, the report now carries an explicit `new` flag computed from the same `known_upstream` membership test that produces `n_upstream_new`, and the jq filters on that. One field, one definition, no cross-file dependency. Pinned with a test.

**⚠️ [L65] `upstream-claims.yaml` — three entries, one issue. Resolved as documentation.**

Your first reading is the correct one: `policy-packs-internal#204` is an umbrella filed for the batch, not a placeholder. The three defects share a root and an owner. The removal rule behaves exactly as you describe.

Added the note you suggested, plus an explicit statement of what the file actually guarantees — `--self-test` checks that an entry cites *an* issue, not a distinct or open one, so the honest claim is "a human has taken responsibility for this finding somewhere" rather than anything stronger.

**📋 Triaged verifier findings — refuted, no action.**

All of them are claims *inside* `testdata/claims-sample.json`, being re-verified as though this PR asserted them. It doesn't — that file is a checked-in sample of the live claims index, and its contents are deliberately verbatim so the classifiers get measured against real text rather than invented strings. The `iam-user-unused-credentials-90` entry flagged at L703/L5318 is a genuine docs defect, which is why it appears in `upstream-claims.yaml` in this same PR and is tracked upstream.

The review classified these spurious/mis-sourced itself, so this is agreement rather than a dispute. Worth noting for future reviews of this file: a fixture of real claims will keep producing this class of finding.

`make test-review-pipeline` and `make lint` pass.

PR #20173, opened 2026-07-10
Retire the /cf2pulumi converter page (#11521)

jkodroff:
Retires the `/cf2pulumi` interactive converter page. cf2pulumi is a non-functional proof of concept with no roadmap, so the docs page is being removed.

## Changes
- Delete the `/cf2pulumi` page and its layout
- Redirect `/cf2pulumi` (and the `/migrate/cf2pulumi` alias) to the CloudFormation migration guide
- Drop the cf2pulumi self-serve link from the migrate hub
- Remove "cf2pulumi converter" claims from the AWS/CloudFormation ad pages

The sibling converters (tf/kube/arm) and the shared convert component are untouched. Backend teardown (the `convertCFN` handler in `pulumi/home`) is a cross-repo follow-up, related to #11523.

Fixes #11521

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #20998, opened 2026-08-19
Prune 27 retired /gads landing pages and refresh the 5 that remain

jkodroff:
Prunes the `/gads/` Google Ads landing pages and refreshes the ones that remain. Follows up on [pulumi/marketing#1791](https://github.com/pulumi/marketing/issues/1791) with fresher, wider data.

## The data changed the recommendation

That audit used a 90-day window, which made 26 of the 32 pages look like they had never been used. Over **12 months** the picture is different: roughly 20 of them ran real campaigns until Google Ads consolidated in mid-April 2026 onto `prospecting_infraascode` (→ `/gads/infrastructure-as-code/`) and `conquest_terraform_alternatives` (→ `/gads/terraform/`). The `google ads competitor-related brand campaign` that fed ~15 competitor pages stopped in mid-March.

These are **retired campaign assets, not pages nobody ever visited** — which matters for how we talk to marketing about deleting them.

Two things also changed since the issue was filed on 2026-07-14:

- `/gads/infrastructure-as-code/` is spending harder — 636 paid clicks in 30d, still live.
- `/gads/terraform/` **went dark on 2026-08-06**. We are now actively advertising to one page, not two.

Portfolio-wide over 12 months: **~12,700 paid clicks produced 12 conversions (0.09%)**. `/gads/terraform/` took 2,110 clicks and converted zero.

## What this PR does

**Deletes 27 pages** — 8 that took no paid click in 12 months (`akuity`, `aws`, `azure`, `chkk`, `google-cloud-deployment-manager`, `kubernetes`, `system-initiative`, `terraform-migration`) and 19 dark since the March/April consolidation. S3 redirects in `scripts/redirects/gads-redirects.txt` land them all on the homepage.

**Keeps 5:**

| Page | Last paid click | Lifetime paid (365d) | Conversions |
|---|---|---|---|
| `infrastructure-as-code` | 2026-08-19 (live) | 2,051 | 3 signups + 1 form |
| `terraform` | 2026-08-06 | 2,110 | 0 |
| `backstage` | 2026-05-30 | 3,010 | 1 signup |
| `platform-teams` | 2026-06-06 | 2,603 | 1 signup |
| `crossplane` | 2026-07-29 | 163 | 0 |

**Content fixes on the survivors:**

- Removes the retired **Crosswalk** feature card (4 pages + the template).
- Rewrites the native-provider card. It claimed full API coverage for AWS, Azure, Google Cloud, and Kubernetes; Azure Native and the Kubernetes provider are the genuine differentiators, so the card now says only that.
- **Adds Neo**, which was missing from every surviving page including our only live one. Per `AGENTS.md` these are marketing pages, so they lead with Neo while noting Claude Code, Cursor, and Codex work with Pulumi too.
- Drops the CloudFormation converter claim on `infrastructure-as-code` — `cf2pulumi` was retired in #20173.
- Points the Starburst card at `/case-studies/starburst/` instead of a blog post, and replaces "up to 3x faster" with the case study's actual finding (two weeks → three hours).
- Adds "per organization" to the HCP Terraform 500-managed-resource claim. Verified against HashiCorp's docs: still accurate. The *legacy user-based* free plan retired 2026-03-31; the 500-resource tier survived and now has unlimited users. Worth knowing because HashiCorp's pricing page no longer shows a free card at all.
- Replaces the verbatim boilerplate subheading shared by `backstage`, `crossplane`, and `platform-teams` with page-specific copy.
- Resolves `/gads/backstage/` arguing against Backstage while advertising our own Backstage plugin.

**Frontmatter and template cleanup:**

- Deletes ~185 lines of `key_features_below` YAML that `hide_platform_details` suppressed entirely, plus the flag. The template now renders that section when the data is present rather than reading a separate boolean, so the dead-block/dead-flag pair can't recur.
- Removes `key_features.title` and `color:`, neither of which any template reads.
- Moves `content/gads/gads-template.md` → `archetypes/gads.md`. It was a *published* page whose `/gads/gads-template` alias nine files claimed at once; as an archetype it stays an authoring aid without a URL.
- Removing the Crosswalk card left a two-item row in a `lg:grid-cols-3`, rendering with an empty third cell. The grid now goes two across for two-item rows.

## Open question for reviewers

**Should we go further and delete everything except `infrastructure-as-code` and `terraform`?**

This PR keeps `backstage`, `platform-teams`, and `crossplane` because all three took paid traffic *after* the April consolidation, so they read as paused rather than retired. But none has run since June, and the case for deleting them is not weak:

- `backstage` — 3,010 lifetime clicks, **1 signup** (0.03%)
- `platform-teams` — 2,603 lifetime clicks, **1 signup** (0.04%)
- `crossplane` — 163 lifetime clicks, **0 conversions**; its single click in the last 30 days is a straggler from the campaign that otherwise died in March

If marketing has no plan to bid on Backstage, platform engineering, or Crossplane again, these should go too and the section drops to the two pages we actually fund. Deleting them later is cheap — same pattern, three more redirect lines. **Keeping a page we never advertise on is the expensive option**, since every one of these is `block_external_search_index: true` and earns no organic traffic by design, so it exists only to be kept accurate.

## Note for reviewers

`/gads/azure/` owned the **root-level** aliases `/azure-resource-manager/` and `/azure-infrastructure/`, which deleting the file would have broken. Both redirect to `/docs/integrations/clouds/azure/` rather than the homepage — the one deliberate exception to the redirect target.

The S3 redirects can't be exercised locally; `make serve` knows nothing about `scripts/redirects/` and those become 301s at deploy time. Verified instead: format matches the existing redirect files, all 30 source keys map to pages actually deleted here (none points at a survivor), and both Azure destinations resolve. No static asset was orphaned. `make lint` and `make build` pass.


cnunciato (approved):

jkodroff:
Intentionally ignored the finding on the # of providers. It changes so often that it's not something that we generally advertise IME, so I don't want to keep chasing the correct number.

PR #21001, opened 2026-08-19
content-review: report-only mode, so the generated trees get fact-checked

### LINKED ISSUE #20996: content-review: add a report-only mode so generated trees get fact-checked
## The gap

Marking a tree tier 0 is meant to say *"don't edit this — a generator owns it."* It is implemented as *"never select this,"* and selection is the only thing that ever writes a page's claim list. So:

> tier 0 → never selected → no claim list → **never fact-checked, ever**

"Don't edit" and "don't look" are different statements, and merging them hides 30% of the docs corpus from the nightly claims re-verify.

Measured against `content/docs/`:

| | pages |
|---|---|
| total | 859 |
| reviewable (the sweep reaches all of these, ~6-month lap) | 605 |
| **tier 0 — permanently excluded, not slow** | **254** |

248 of those 254 are the CLI command reference.

Claims snapshots in the live index, by generated tree:

| tree | snapshots |
|---|---|
| `iac/cli/commands/` (248 pages) | **0** |
| `reference/pkg/` (SDK API ref) | **0** |
| `reference/cloud-rest-api/` (OpenAPI) | **0** |
| `pre-built-policy-packs/` | 12 |

## Why this is not hypothetical

The policy-pack tree is in the index **by accident**. Those pages were last reviewed **2026-07-15**; #20274 converted them to generated content on **2026-07-16**. One day of overlap.

That one day is the only reason we know about the three real, user-facing misstatements now tracked in `policy-packs-internal#204` — including docs describing an AWS Config rule as being about credentials *rotated* every 90 days when it is about credentials left *unused*.

Nightly regeneration (`update-policy-packs.yml`) does not help: it keeps those pages **fresh**, not **correct**. It republishes an upstream error faithfully forever. The claims lane is the only thing checking those descriptions against the upstream source of truth.

We have never looked at the 248 CLI pages. Not "we checked and they're fine."

## Proposal

A third worker mode alongside `fix` and `glowup`: **review, record the claim list, change nothing.** No fix attempt, no PR, no wasted edit against a file the generator owns.

Then lift the tier-0 gate for that mode only, and let the nightly lane take it from there. Findings route through the upstream lane that already exists (#20990) — never marked, always re-checked, reported every run, with a prefilled upstream issue in the Slack line.

### Scope

- **In: `content/docs/iac/cli/commands/`** — 248 committed `.md` files, so the selector can already find them. Full of hand-written prose descriptions that can be wrong exactly the way the policy-pack ones were.
- **Out: `reference/cloud-rest-api/`** — assembled at request time from a schema, nothing on disk to review. Needs a different intake.
- **Out: `reference/pkg/`** — generated from the machine-readable API definition, which *is* the source of truth. The one place "it's generated" really does mean "it's correct."

### Also needed

Split the tier-0 flag into its two real questions (`editable` vs `reviewable`), so this doesn't have to be a mode-specific gate override.

### Cost

One-time pass over 248 pages to build the claim lists; after that it is only the nightly check. Worth sizing the model spend before committing.

---
Context: pulumi/docs#20990, `policy-packs-internal#204`.

CamSoper:
### Proposed changes

Tier 0 was two statements wearing one flag. It was meant to say *"a generator owns this — don't edit it."* It was implemented as *"never select this,"* and selection is the only thing that ever writes a page's claim list. So `tier 0 → never selected → no claim list → never fact-checked, ever` — 254 pages, 30% of `content/docs/`, 248 of them the CLI command reference. Not "we checked and they're fine."

This splits the flag into its two real questions and adds the lane the split makes possible.

#### The split

`strategic-tiers.yaml` gains two keys, both defaulting to true, with `tier: 0` kept as shorthand for "neither":

| key | question |
|---|---|
| `editable` | may a PR change this file? |
| `reviewable` | may the pipeline read this page and record what it claims? |

`policy_for()` replaces `tier_for()` as the full answer (the old signature stays for the scoring-only callers). Every consumer that used to test `tier == 0` now asks `editable`: the glow-up selector, the snippet sweep, the retire veto, and — the one that matters most — `reverify-claims.fix_route`, so a tree the report lane reads still routes its contradictions upstream rather than to a stale-claims marker no PR here could ever retire.

Two guards fall out of that and are pinned by tests: the CLI reference is now tier **3**, so the retirement veto has to come from `editable` (it does), and the snippet sweep would otherwise have started editing 248 generated files the day their tier stopped being 0 (it doesn't — verified against the real tiers file).

#### The lane

`select-articles.py --mode report` selects the other half of the corpus — reviewable-but-not-editable pages — with the same scoring and the same staleness lap. The two lanes partition the corpus; they never overlap. It makes no `gh` calls, because it opens nothing to collide with.

The worker's third mode runs the claim pipeline **and nothing else**. There is no model step: with no fix to apply and no PR body to write, an agent turn buys nothing and can only produce an edit the gate then has to reject. Vale, frontmatter, cross-sibling, and readthrough are skipped for the same reason — they exist to produce fixes for someone to apply. What's left is exactly what the claims index is built from, which is the whole point of visiting these pages.

The workflow writes the `reported` verdict itself, and **only when verification actually produced verdicts**. A degraded run writes no sentinel, so the ledger records `incomplete` and the page stays due — stamping it `reported` would advance the staleness clock on a page nothing had checked, which is the exact claim this lane exists to stop making.

#### Reporting the findings

`report-claims-findings.py` is new. Recording to the index is not enough on a page's **first** pass, and the first pass is where the findings are: the nightly re-verify only re-checks *volatile* entities, so a contradicted behavioral claim would otherwise sit in the index forever with nobody told, and there is no model on this lane to read the verdicts and act on them.

Routing is the nightly lane's own, imported rather than reimplemented (`fix_route`, `load_known_upstream`, `file_issue_url`). A contradiction reaches #docs-ops with a prefilled issue against the repo that owns the text — `pulumi/pulumi` for this tree, already in `upstream-claims.yaml`'s `repos:` map — and one a human has already filed stays quiet. `file_issue_url` gained an `origin` argument so the issue body says which lane found it: for these pages it is the first check they have ever had, which the reader should know.

`publish-gate.py` fails closed on a `reported` verdict with a non-empty patch or with a retirement.

#### Scope

- **In: `content/docs/iac/cli/commands/`** — 248 committed files whose descriptions, synopses, and flag help are hand-written prose that can be wrong exactly the way the policy-pack ones were.
- **Out: `reference/pkg/`** — stays tier 0, and the entry now says why: it renders from the machine-readable API definition, which *is* the source of truth. Checking it would be verifying the schema against itself.
- **Out: `reference/cloud-rest-api/` and the pre-built policy packs** — not a policy call but a mechanical one: the selector globs `content/docs/**/*.md` and there are no files there to find. Both need an intake that can render the adapter's pages first. The policy-pack entry now records that, so the next person doesn't read its tier 0 as a judgment about its content.

#### Cost

The lane ships **default-off** — `REPORT_REVIEW_COUNT` unset means off, deliberately inverting the default-on convention the other lanes use — because the issue asks for the spend to be sized before committing, and an estimate is not a size.

What can be measured now: 248 pages, 754 KB, median page ~690 tokens. Extraction is two Sonnet calls per page; verification is one agent loop per claim, and claims-per-page for this tree is the unknown that dominates. Taking the checked-in corpus sample's ~20 claims/page as an upper bound and a mid-range per-claim loop, a page lands around **$0.30–$0.60** at Sonnet 5 rates, so the one-time pass is **order $100–150**, not $10 and not $1,000. Steady state after that is only the nightly volatile rotation these pages join.

So rather than argue the estimate, the lane measures itself: every run sums the pre-steps' own reported token usage into a `cost` block in the findings artifact and prints it. Turn it on at `REPORT_REVIEW_COUNT=5` for a week and the real per-page number is in the artifacts; the lap time follows directly (5/working day ≈ 10 weeks for the tree, 10/day ≈ 5).

n/a

### Related issues (optional)

Fixes #20996. Follows #20990, whose upstream lane this reuses wholesale; the three findings tracked in `policy-packs-internal#204` are the evidence that this tree is worth checking.

### Verification

- `make test-review-pipeline` — all suites green (7 harnesses, 18 `--self-test`s). New cases: the two lanes partition the corpus and carry their mode; a reported page yields to its unreported sibling on the next lap; the shipped tiers file routes each generated tree to the lane its entry claims; the CLI tree is retirement-vetoed by `editable` while tier 3; `fix_route` sends a reviewable-but-not-editable page upstream; the gate rejects `reported` + patch and `reported` + retirement; the ledger records `reported` as a completed status. Plus `report-claims-findings.py --self-test` (15 checks).
- `make lint` — 1845 files parsed, 0 errors; Prettier clean.
- Simulated a full report-mode run locally end to end with a fabricated verdicts artifact (everything except the API calls): queue → verdict → publish gate (`publish=false`) → claims index → ledger (`status=reported`) → findings report → Slack line. Confirmed the fix lane's own selection is byte-identical apart from the two new queue keys.
- Confirmed the snippet sweep skips the CLI tree (0 pages scanned) and still scans a normal page, since its tier is no longer 0.

Not verified here: a real end-to-end run, which needs the API key and the S3 ledger. The first live dispatch is the remaining check — and it is the same dispatch that produces the cost number.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UD98cAjgUXZhWtKgb6ZyhR)_


CamSoper:
@claude 

Fix the blocking issue with your recommendation, and then #update-review

PR #21002, opened 2026-08-19
content-review: stop the ledger erasing the backlog a glow-up failed to execute

CamSoper:
### Proposed changes

Unit 1 of 3 from the #20984 investigation. This one stops the data loss; the other two make the backlog recoverable and stop a glow-up poisoning its own findings record.

**What happened.** [#20984](https://github.com/pulumi/docs/pull/20984) opened a glow-up announcing `No banked findings reachable (no prior review PRs reachable; run the taxonomy-only sweep) — taxonomy-only glow-up` for a page whose ledger entry recorded `skipped_findings: 17, clarity_flag: true` — which is precisely why the glow-up selector picked it (score 17.6166). Nothing was unreachable. The pointer to the page's last review PR had been erased three weeks earlier, so `fetch_pr()` was never called at all. [Run 32263128501](https://github.com/pulumi/docs/actions/runs/32263128501) logs `0 banked finding(s) from 0 PR(s)` beside `skipped_findings: 17`, and the dispatched queue carried `source_pr_number: null`. The findings it needed were sitting in [#19885](https://github.com/pulumi/docs/pull/19885) the whole time.

**Why the pointer was gone.** `build_record()` rebuilds the record from a blank sheet every review and only fills `pr_number` in the `fixed`/`glowup`-with-a-PR branch, so any review that opens no PR blanks it. That is not a rare path: a review applying zero fixes writes `clean`, `publish-gate.py` sets `publish=false`, and no PR is opened however many findings it banked. Same carry-forward class as #20968.

**Why it then got worse.** The degraded glow-up wrote its own zeros over the counters — the sentinel's `skipped_findings` is the *declined* count (0, the backlog was empty) and it carries no `clarity_flag` at all. `select-glowup.py` excludes on `banked <= 0 and not clarity_flag` and cools the page down for 90 days, so the lane deleted the debt it had just failed to collect.

Changes:

- **`load_prior()`** reads the record back from S3 before this run overwrites it. From S3 rather than the queue because it also covers a `--paths` manual dispatch and keeps `queue_json` — already the dispatch's largest input — from growing. Returns `None` on any failure, so the degradation is "no carry-forward", never "no ledger write". `--prior` injects it for tests, mirroring `--pr-json`.
- **`last_pr` / `last_pr_number`** carry the pointer forward, chaining through the prior record's own carried value — without the chain the reference survives exactly one PR-less review. `pr_number` keeps its meaning ("the PR *this* review opened") so `reconstruct-ledger-history.py`'s CSV stays comparable across history.
- **`clarity_flag` is three-state**: an absent key carries the prior flag, an explicit `false` clears it. The glow-up sentinel never carried the key, so every glow-up silently cleared the flag it had just been selected for. The sentinel schema now asks for it explicitly — `SKILL.md` and the workflow prompt in the same commit.
- **A glow-up that executed nothing and declined nothing** preserves the prior banked count and sets `glowup_degraded`; `glowup_cooldown_active()` skips a degraded entry. The flag is load-bearing because the counters cannot express it: a degraded run carries 17 forward, which reads identically to a run that declined 17.

Carry-forward is an explicit three-field read, never a merge — a blanket merge would leak a stale `status` or `fixes` forward and make one record describe two reviews. There are tests for that directly, plus the #20984 regression case and `load_prior`'s failure paths.

One deviation from the approved plan, flagged for review: the plan proposed keying the cooldown on `fixes == 0 and skipped_findings == 0` with no new ledger field. That is self-contradictory once the same change preserves the counters — after preservation `skipped_findings` is 17, not 0. Hence the single explicit `glowup_degraded` boolean.

Verification: `make test-review-pipeline` and `make lint` both pass (1845 files parsed, 0 errors; Prettier clean).

### Related issues (optional)

Follows up on #20984. Related: #20990 (structured findings record, forward-only), #20970 (glow-up lane), #20968 (the stale-claim-marker carry-forward this mirrors).

---

### DIFF
--- a/.claude/commands/review-existing-content/SKILL.md
+++ b/.claude/commands/review-existing-content/SKILL.md
@@ -478,8 +478,13 @@ frontmatter) are also present and are your evidence base.
    .self-check.u0.diff --article <path> --article-blob <pristine-copy> --out
    .self-check-report.json` (copy the article aside before your first edit).
 1. **Verdict sentinel**: `{"verdict": "glowup", "fixes": <executed count>,
-   "skipped_findings": <declined count>, "retirement": false}` — no
-   `applied[]` array; the glow-up gate replaces the per-hunk check. If the
+   "skipped_findings": <declined count>, "clarity_flag": <bool>,
+   "retirement": false}` — no `applied[]` array; the glow-up gate replaces
+   the per-hunk check. State `clarity_flag` explicitly: `false` once you have
+   resolved the page's readthrough reconception, `true` while one still
+   stands. Omitting it carries the page's existing flag forward unchanged —
+   which is the right default, since the flag is usually why the page was
+   selected, but it means only you can put it down. If the
    queue article carries `stale_claim_markers`, they are must-address
    findings here exactly as in a fix review: resolve each (fix it, or
    establish the flag was wrong) and list its `entity_key` in the sentinel's
--- a/.github/workflows/content-review-article.yml
+++ b/.github/workflows/content-review-article.yml
@@ -512,9 +512,12 @@ jobs:
               BEFORE your first edit for the --article-blob input). Fix any
               violation and re-run until it passes.
             - Verdict sentinel: `{"verdict": "glowup", "fixes": <executed
-              count>, "skipped_findings": <declined count>, "retirement":
-              false}` — no `applied[]` needed; the glow-up gate replaces the
-              per-hunk check.
+              count>, "skipped_findings": <declined count>, "clarity_flag":
+              <bool>, "retirement": false}` — no `applied[]` needed; the
+              glow-up gate replaces the per-hunk check. State `clarity_flag`
+              explicitly: `false` once you have resolved the page's
+              readthrough reconception, `true` while one still stands.
+              Omitting it carries the page's existing flag forward.
             - Your PR will NOT auto-merge: it opens for human review (the
               sweep assigns reviewers). Write the body tables for that
               reader.
--- a/scripts/content-review/record-review.py
+++ b/scripts/content-review/record-review.py
@@ -30,7 +30,8 @@ genuinely "incomplete" and stays due for retry.
 
 Canonical record (every field always present):
   { path, slug, lane, status, pr, pr_number, last_pr, last_pr_number, head_sha,
-    fixes, skipped_findings, glowup_degraded, retirement, note, attempts,
+    fixes, skipped_findings, glowup_degraded, glowup_degraded_runs, retirement,
+    note, attempts,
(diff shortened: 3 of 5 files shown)


CamSoper:
@claude #update-review

All three findings fixed in 8ec0d27 — none disputed.

**🚨 The durable pointer was written and never read.** Correct, and the reader-facing half of #20984 was indeed still open. `select-glowup.py` now falls back:

```python
"source_pr_number": int(entry.get("pr_number")
                        or entry.get("last_pr_number") or 0) or None,
```

That is the load-bearing one — it is what the unprivileged worker is handed, and without it the worker got `None` for exactly the pages with the largest backlogs. `build-glowup-backlog.py`'s reader is in the stacked #21004, in the same form you suggested (`for n in (entry.get("pr_number"), entry.get("last_pr_number"), article.get("source_pr_number")):`). The `record-review.py` comment no longer claims a reader this tree lacks — it now names both consumers and what each uses it for.

**⚠️ The degraded exemption was unbounded.** Right, and worse than "may not converge": with a systematically unrecoverable page (no PR ever + no findings record, which is the pre-2026-08-19 population) it never converges. Added `glowup_degraded_runs`, counting consecutive degraded runs, with the exemption stopping at `GLOWUP_DEGRADED_ATTEMPT_CAP = 2`; a glow-up that executes or declines anything resets it. Your read on `attempts` was the key detail — the glowup path zeroes it, so it could never have served as the guard.

**⚠️ A failed `aws s3 cp` reported as "first review".** Fixed. Only a genuine `404`/`Not Found`/`NoSuchKey` stays quiet; anything else warns, names the consequence ("may reset this page's PR pointer and banked counters"), and includes the aws stderr. Tests assert the two paths by their logs, not just by the `None` return.

Also new: tests for the counter accruing and resetting, both sides of the cap, the pointer fallback plus this-review's-own-PR still winning, and both ledger-read failure paths.

CamSoper:
@claude #update-review

Fixed in e105385 — not disputed. You were right on both clauses: `build-glowup-backlog.py` at this HEAD neither reads `last_pr_number` nor has a head-ref query to fall back *from*. Both arrive in the stacked #21004, and the comment was describing that tree rather than this one.

It now names only what is true here — `select-glowup.py` folds the pointer into the `source_pr_number` it stamps on the queue, which is the only form the unprivileged worker gets it in and what `build-glowup-backlog.py` reads there — and puts the dispatcher-side ledger read in future tense against #21004.

Worth the second pass rather than an argument: a comment asserting behaviour the code lacks is precisely how this pipeline shipped the bug the PR is fixing.

PR #21003, opened 2026-08-19
content-review: take a glow-up's findings disposition from its own sentinel

CamSoper:
### Proposed changes

Unit 2 of 3 from the #20984 investigation. **Stacked on #21002** (it edits the same glow-up sentinel spec, so it targets that branch; GitHub will retarget to `master` when #21002 merges).

`record-page-findings.py` decides `applied` by overlapping each finding's line range against the verdict's `applied[]`. A glow-up verdict has no `applied[]` — by design, since `verify-glowup-scope.py` replaces the per-hunk check with whole-page bounds and there is no per-fix attribution to record. So `spans` is empty, every finding is written `applied: false`, and the record now claims the work the glow-up just finished is still outstanding.

`build-glowup-backlog.banked_from_findings()` banks exactly the `applied: false` items, so the next glow-up re-proposes work that already landed — which that function's own docstring names as the thing that erodes trust in the lane.

The disposition already exists and did not need inferring. The model partitions `.glowup-backlog.json` into the Backlog executed and Backlog declined tables, and those items carry stable ids; the sentinel now reports that partition as `executed_ids` / `declined_ids`, and `mark_from_backlog()` maps them back onto findings (`findings-<id>` → `<id>`). A pr-body-sourced backlog item is prose recovered from a PR description with no counterpart on the record, so it can never mark one — there's a test pinning that.

A glow-up sentinel carrying neither list writes **no record for the run** rather than an all-false one. The previous record stands — stale but true, refreshed by the next fix-lane review from its own artifacts — which beats a fresh record that is wrong in the one direction that costs reviewer trust. That means a glow-up run's findings `reviewed_at` can lag the ledger's; the docstring says so explicitly so nobody "fixes" it later.

Rejected: synthesizing `applied[]` from the glow-up diff. It would recreate the per-hunk attribution the glow-up gate deliberately avoids, and guessing is how the first cut of this file shipped the substring-matching bug (fixed in #20990).

The sentinel schema change lands here in both places it's specified — `SKILL.md` and the workflow prompt — or the model never emits the lists and the skip path fires on every glow-up. The fix-lane path is untouched, with a regression test asserting it still uses line overlap.

Verification: `make test-review-pipeline` and `make lint` pass (1845 files parsed, 0 errors; Prettier clean).

### Related issues (optional)

Stacked on #21002. Follows up on #20984; the line-overlap matching this special-cases came from #20990.

---

### DIFF
--- a/.claude/commands/review-existing-content/SKILL.md
+++ b/.claude/commands/review-existing-content/SKILL.md
@@ -486,8 +486,15 @@ frontmatter) are also present and are your evidence base.
    .self-check-report.json` (copy the article aside before your first edit).
 1. **Verdict sentinel**: `{"verdict": "glowup", "fixes": <executed count>,
    "skipped_findings": <declined count>, "clarity_flag": <bool>,
-   "retirement": false}` — no `applied[]` array; the glow-up gate replaces
-   the per-hunk check. State `clarity_flag` explicitly: `false` once you have
+   "executed_ids": [...], "declined_ids": [...], "retirement": false}` — no
+   `applied[]` array; the glow-up gate replaces the per-hunk check.
+   `executed_ids` and `declined_ids` are the `id` values from
+   `.glowup-backlog.json` for the rows you put in the Backlog executed and
+   Backlog declined tables — the same partition, reported as data. Every
+   banked id belongs in exactly one of them; both lists are empty only when
+   the backlog itself was empty. Without them the findings record cannot
+   tell what you executed from what you left, so it records nothing at all
+   rather than filing your completed work as still outstanding. State `clarity_flag` explicitly: `false` once you have
    resolved the page's readthrough reconception, `true` while one still
    stands. Omitting it carries the page's existing flag forward unchanged —
    which is the right default, since the flag is usually why the page was
--- a/.github/workflows/content-review-article.yml
+++ b/.github/workflows/content-review-article.yml
@@ -632,11 +632,18 @@ jobs:
               violation and re-run until it passes.
             - Verdict sentinel: `{"verdict": "glowup", "fixes": <executed
               count>, "skipped_findings": <declined count>, "clarity_flag":
-              <bool>, "retirement": false}` — no `applied[]` needed; the
-              glow-up gate replaces the per-hunk check. State `clarity_flag`
+              <bool>, "executed_ids": [...], "declined_ids": [...],
+              "retirement": false}` — no `applied[]` needed; the glow-up
+              gate replaces the per-hunk check. State `clarity_flag`
               explicitly: `false` once you have resolved the page's
-              readthrough reconception, `true` while one still stands.
-              Omitting it carries the page's existing flag forward.
+              readthrough reconception, `true` while one still stands;
+              omitting it carries the page's existing flag forward.
+              `executed_ids`/`declined_ids` are the `id` values from
+              `.glowup-backlog.json` for the rows you put in the Backlog
+              executed and Backlog declined tables — every banked id in
+              exactly one list. Without them the findings record would file
+              the work you just did as still outstanding, so it records
+              nothing for the run instead.
             - Your PR will NOT auto-merge: it opens for human review (the
               sweep assigns reviewers). Write the body tables for that
               reader.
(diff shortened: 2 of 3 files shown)


CamSoper:
@claude #update-review

Fixed in 7cf369c — not disputed. This was the sharpest finding across the three PRs.

You were right that the two id spaces are not pinned to the same record: `build-glowup-backlog.banked_from_findings` mints `findings-<id>` from the **previous** run's record, and this record renumbers positionally over freshly collected artifacts. The expected case between glow-ups is exactly the one that breaks it — an intervening fix-lane review applies a fix, its finding disappears, everything after it shifts by one, and `executed_ids: ["findings-f2"]` marks the wrong finding.

Took the durable fix rather than the interim guard. Ids now locate the backlog **item**, and the finding is resolved by **content**: `banked_from_findings` writes each item's text as `label` (plus ` — detail`), so the label is a prefix of the text and the match is exact rather than fuzzy or hashed. Each finding can be claimed once. An id that resolves to no finding warns and leaves everything deferred — deliberate asymmetry, since re-proposing finished work is irritating and visible while marking the wrong item applied loses real work silently.

Two test-side notes, both deliberate:

- The existing `"a missing backlog does not crash the id mapping"` check asserted the ordinal fallback (`findings-f2` marking the second finding with nothing to resolve it against). That fallback **is** the guess this removes, so the check now asserts the safe behaviour instead. Changed because it was wrong, not to make the suite green.
- The fixture carried backlog text `banked_from_findings` cannot produce (`"Vale filler (L48)"` against a longer label). It now matches the real shape — a test of a shape the lane never emits proves nothing.

New checks cover a renumbered finding set resolving correctly, a finding that disappeared leaving everything deferred, an id absent from the backlog, the ` — detail` suffix not breaking the match, and two ids being unable to claim the same finding.

CamSoper:
@claude #update-review

Fixed — not disputed. The two guards did disagree, and the consequence was identical to the one the first guard exists to prevent.

Both take the `return None` exit now. One refinement on the suggested `if executed and not applied_idx`: that would have skipped the write for a glow-up that legitimately executed **only pr-body items** — prose recovered from a PR description, which has no counterpart on this record and is *expected* to resolve to nothing. Skipping there discards an accurate record. So the guard counts only ids that should have resolved: `findings-record`-sourced ones, plus ids absent from the backlog entirely (ambiguous — either the snapshot is gone or the sentinel named something never banked, and both should have resolved). An existing test caught it, which is why it is a distinct case rather than a footnote.

Partial resolves still write, as you said they should.

Four checks moved from asserting an all-False record to asserting the skipped write — all in cases where nothing resolved, so the behaviour they pinned is exactly what this replaces. Two new ones cover the partial resolve and a pr-body id riding alongside a resolving one.

Also note this PR was rebased (no content change) after #21002 took another commit, so the previous review was stale against the new head.

CamSoper:
@claude #update-review

Both fixed — neither disputed. Both were drift my own previous commit introduced, which is the reason they were worth catching rather than waving through.

**Docstring.** You are right that the module header covering it is not enough: the function docstring is what a caller reads, and this change's own argument is that the reasoning lives there so nobody deletes the guard later. Someone reading "Returns None when the sentinel carries neither list" could reasonably have concluded the new guard contradicted the documented contract. Took your wording, with one addition — pr-body ids explicitly do not count toward "should have resolved", since that distinction is the whole reason the guard is scoped the way it is.

**Caller warning.** Also right, and worse than imprecise: it pointed debugging at the model's sentinel when the actual cause is a missing backlog snapshot. The callee already logs the specific reason on both paths, so the caller now states the outcome and defers to it.

No behaviour change in either — both are the record telling the truth about itself, which is the point of the series.

CamSoper:
@claude #update-review

Both fixed — neither disputed. The outstanding one was my own regression from the previous commit, and your read of the trade is exactly right: swapping a wrong message on the rare path for a useless one on the common path is not an improvement.

`mark_from_backlog`'s first exit did return silently, so `(see above)` pointed at nothing — and that path fires whenever the model omits both keys, which is the likelier of the two. Took your fix: the silent exit warns, which makes the invariant the caller's wording already assumed ("every `None` return warns first") true rather than aspirational.

You noted no self-test asserts on warn text. That was part of why it broke, so I pinned it — two checks capturing stderr across both `None` paths. Nothing else would catch its removal, and the caller's message now depends on it.

Header line reflowed too.

CamSoper:
@claude #update-review

Fixed — not disputed, both parts.

The orphaned comment was a real readability trap: "Without the backlog there is nothing to resolve..." sat fifteen lines above its check with two checks that pass `BACKLOG` in between. Moved the warn-invariant block below the neighbouring nothing-resolved check so both comments sit over their own assertions.

On the imports — you were right that mine were the only function-local ones in the file, so they were a lone inconsistency rather than a convention. Hoisted. `--self-test` runs in the same process, so keeping them local bought nothing.

Comments and layout only; assertions unchanged, still 36 passing.

PR #21004, opened 2026-08-19
content-review: recover a page's banked backlog by head ref, and stop lying when it fails

CamSoper:
### Proposed changes

Unit 3 of 3 from the #20984 investigation. **Stacked on #21003** (which is stacked on #21002); each retargets automatically as its parent merges.

**The false premise.** `build-glowup-backlog.py` asserted that reused branch names made older review PRs undiscoverable, so it read only whatever pointer the ledger happened to carry. That is not true. `gh pr list --head <branch> --state all` returns every PR a branch has ever carried, closed and merged included — `content-review/docs-iac-concepts-providers` returns #20953, #20927 **and** #20503, and the branch #20984 needed returns [#19885](https://github.com/pulumi/docs/pull/19885), whose "Findings not applied" section held the very items its taxonomy sweep then rediscovered by luck.

Discovery now queries all three canonical heads — fix, retire, glow-up — with the ledger/queue pointer kept only as defence in depth. The glow-up head earns its call on its own: a prior glow-up's "Backlog declined" rows are the debt `record-review.py` counts into `skipped_findings`, and the selector re-picks the page on that number, so the backlog has to be able to name it or the counter is unbacked. Per the approved decision those rows are **re-banked, tagged** `glowup-declined` and rendered `#20984 (declined)` — dropping them makes debt vanish silently, which is the bug class this file exists to close, while re-proposing them blind invites a decline loop nobody can see forming.

Section extraction is lane-aware, keyed off `headRefName` rather than guessed from the body: a glow-up body uses different headings, and reading fix-lane headings out of one recovers nothing.

**The collapsed message.** Four distinct outcomes were reported as one sentence, `no prior review PRs reachable` — which #20984 printed for the single case where *no lookup was attempted at all*, sending its reader after an API failure that never happened. There are now four states (`recovered` / `no_prior_prs` / `gh_unavailable` / `prs_carried_nothing`), each with its own note, all recorded in a `recovery` block naming the heads queried and the PRs read.

**The silent contradiction.** `skipped_findings: 17` with `banked: []` is impossible — those counters *are* the selection criterion — and it passed through three programs without a word. It now sets `degraded`, emits a `::warning::`, and the composer renders a `[!WARNING]` naming the count, replacing the sentence that read as "this page had nothing outstanding".

Fail-open by default (decision D5): the review job is unprivileged, a red X there reads as "the lane is broken" when the true state is "the evidence is out of reach", and #21002 now preserves the backlog either way, so a degraded run costs a slot rather than the data. `--strict` exists so the workflow can be flipped fail-closed without another change here.

`pull-requests: read` is added to the review job — read-only, and it still holds no push token. Worth knowing: that job has only ever had `contents: read` + `id-token: write`, so the pre-existing `gh pr view` path likely never succeeded; nothing surfaced it because the pointer was always empty by the time it was called.

**One thing I could not verify locally:** `gh` isn't installed in this sandbox, so `gh pr list --json body,headRefName` is unconfirmed end-to-end. `body` is a documented `pr list` field and a failure would be loud (non-zero exit → `gh_unavailable` → the new warning, not a silent empty backlog), and a PR arriving without a body is hydrated individually via the existing `fetch_pr`. Still worth a glance at the first glow-up run's logs.

The `--pr-json` fixture path still performs zero subprocess calls — now with a test that asserts it, since that property is what the whole self-test suite rests on.

Verification: `make test-review-pipeline` and `make lint` pass (1845 files parsed, 0 errors; Prettier clean). New self-tests cover the four recovery states, the lane-specific section split in both directions, declined-item tagging, and the #20984 regression in both the backlog builder and the composer.

### Related issues (optional)

Stacked on #21003 and #21002. Fixes the root cause behind #20984; corrects a claim repeated in #20990's commit message.

---

### DIFF
--- a/.github/workflows/content-review-article.yml
+++ b/.github/workflows/content-review-article.yml
@@ -85,6 +85,9 @@ jobs:
     permissions:
       contents: read
       id-token: write
+      # build-glowup-backlog.py reads every prior review PR for the page by
+      # head ref. Read-only, and the job still holds no push token.
+      pull-requests: read
     outputs:
       claude_outcome: ${{ steps.claude.outcome }}
     steps:
--- a/scripts/content-review/build-glowup-backlog.py
+++ b/scripts/content-review/build-glowup-backlog.py
@@ -16,17 +16,33 @@ backlog rather than a cold read:
                  "source_pr": 123, "text": "<the row/bullet>"}, ...],
      "notes": ["<degradation notes, if any>"]}
 
-Sources: the page's ledger entry (`--ledger-dir`) carries the latest review's
-`pr_number`; earlier PRs for the same page are discovered via `gh pr list`
-over the page's canonical branch names is NOT possible (branches are reused),
-so the backlog reads the PRs the ledger knows about — currently the latest —
-plus any extras passed via `--pr` (the workflow may pass several). Section
-extraction is heading-based and tolerant: a PR body without the section, an
-unreachable PR, or no PRs at all degrade to an empty `banked` list with a
-note — the skill then runs its taxonomy-only sweep.
+Sources, in order of authority:
 
-`--pr-json <file|->` injects PR view fixtures for tests (a JSON list of
-{number, url, body}), replacing every `gh` call.
+1. The structured findings record (`record-page-findings.py`), when the page has
+   one. Forward-only: pages last reviewed before 2026-08-19 have none.
+2. Every prior review PR for the page, found by HEAD REF. An earlier version of
+   this file asserted that reused branch names made older PRs undiscoverable, and
+   so read only the pointer the ledger happened to carry. That was wrong, and it
+   is why #20984 recovered nothing: `gh pr list --head <branch> --state all`
+   returns every PR that branch has ever carried, closed and merged included
+   (`content-review/docs-iac-concepts-providers` returns #20953, #20927 and
+   #20503). All three canonical heads are queried — fix, retire and glow-up —
+   because a prior glow-up's "Backlog declined" rows are debt too: they are what
+   `record-review.py` counts into `skipped_findings`, and the selector re-picks
+   the page on that number, so the backlog has to be able to name it.
+3. The ledger/queue PR pointer and any `--pr` extras, as defence in depth.
+
+Section extraction is heading-based, tolerant, and lane-aware — a glow-up PR body
+uses different headings from a fix PR's. Recovery never fails the run: the
+`recovery` block records which of four outcomes happened (`recovered`,
+`no_prior_prs`, `gh_unavailable`, `prs_carried_nothing`) and the skill falls back
+to its taxonomy-only sweep. When the ledger says findings were banked and none
+could be recovered, the backlog is stamped `degraded` and a `::warning::` is
+emitted — a silently empty backlog reads exactly like a clean page, which is how
+#20984 shipped a PR announcing it had nothing to do while 17 findings waited.
+
+`--pr-json <file|->` injects the discovered PR set for tests (a JSON list of
+{number, url, body, headRefName}), replacing every `gh` call.
 
 Self-contained smoke checks: `python3 build-glowup-backlog.py --self-test`.
 """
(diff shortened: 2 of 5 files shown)


CamSoper:
Verified end to end now that I have \`gh\`.

**The unverified risk is closed.** \`gh pr list --head <branch> --state all --json number,url,body,headRefName,createdAt\` works exactly as this depends on — \`content-review/docs-iac-concepts-providers\` returns #20953, #20927 and #20503 with full bodies (8678 / 8269 / 6067 chars).

**And the real thing:** running the builder against the actual page from #20984, with the same empty pointer that run had:

```
build-glowup-backlog: 9 banked finding(s) from 2 PR(s) [recovered]
```

It finds #19885 and #20984 by head ref and recovers #19885's four unverifiable claims and four Vale nags — including the `a clean slate` cliché, the two `all of` wordiness nags and the `mostly` weasel word that #20984's taxonomy sweep rediscovered by luck. `degraded: false`, `state: recovered`.

**One fix that live run earned** (pushed as f8ead52): the first pass recovered 13 items, five of which were the composer talking to itself. Exact-line noise matching was never exercised before, because the pointer was always empty by the time anything read it — and every pre-fill continues past its phrase:

- `No images. The page source references no screenshots, diagrams, ...`
- `Skipped — the page source uses only render-safe chrome (...)`
- `None — the backlog was empty (`banked: []`; ...)`

None matched `{"no images.", "skipped.", "none."}`, so all were banked for the model to execute or decline in the PR tables. Same for the routing trailer (`... or run /glow-up <path> to work them now`) and the empty-deferrals placeholder.

Prefix-matched now, with the trailer caught by the invariant that it points at the `/glow-up` command rather than describing a finding — its wording changed mid-lane, so matching either phrasing would rot. 13 items → 9: the eight real deferrals plus #19885's rendered-content observation, which is genuine prose a human may want to skim and is left alone rather than filtered by guessing at what "clean" reads like.

`make test-review-pipeline` and `make lint` still pass.

CamSoper:
@claude #update-review

Both findings fixed in 4c74ec3 — neither disputed, and both confirmed against live data.

**⚠️ Partial head-query failure reported as "never reviewed".** Correct, and it was the same collapse this PR exists to prevent, one level down. `elif failed_heads:` now, and `_recovery_note` names `recovery["prs_unreachable"]` rather than all of `heads_queried`, so the sentence stays true when only some heads failed. Tests assert a partly-failed query never says "has ever used" and that the note names only the unanswered head.

**⚠️ No de-duplication across discovered PRs.** Right, and measurably worse than the fix-PR-plus-decline case you described. Ran it against `content-review/docs-iac-concepts-providers`:

```
raw banked lines across the 3 PRs: 48
distinct findings: 43
collapsed by dedup: 5
   x3  vale filler l48 don t start a sentence with there are
   x3  vale wordiness l84 all of is too wordy
   x2  vale future tense l63 future tense the pulumi cli will ins
```

The same unactioned Vale nag sits in all three fix PR bodies — in three different wordings — so the gap widens by one row per review forever, before any glow-up decline is involved.

One refinement on your suggestion: a **normalized-text** key would have collapsed almost nothing here, because the reason clause differs every time (`Style nag; a prose judgment` vs `` `write-good` heuristic; a style call``). Keyed on the **label** instead — every shape the lane emits is `<label> — <reason>`, so splitting at the em-dash gives the finding's identity and the match becomes exact. Collision preference is as you suggested for the declined variant (it carries the decline history), and otherwise the later PR wins so the reason shown is the most recent judgment rather than the oldest.

Re-verified end to end after both changes: the #20984 page still recovers the same 9 items (distinct labels, nothing collapsed), the providers page goes 48 → 43 with no key banked twice.

CamSoper:
@claude #update-review

No new findings to address here — both from the last round were fixed in the previous push and the review came back clean (0 outstanding, 0 low-confidence).

This is a rebase-only refresh: #21002 and #21003 each took another review-fix commit, so this branch was rebased onto them and the previous review is stale against the new head. No changes to this PR's own diff.

CamSoper:
@claude #update-review

Fixed — not disputed, and a good catch on a PR titled "stop lying when it fails."

You are right that both halves are false at this head: `build-glowup-backlog.py` already iterates `(pr_number, last_pr_number, source_pr_number)`, so it does not "still key on `pr_number` alone", and "#21004 teaches it" points at the PR the reader is standing in.

This is the intrinsic cost of describing cross-PR state in a stack — the sentence is true in exactly one of the two trees. #21002's version is accurate *there*, and had to be: its own review rejected an earlier comment for claiming a reader that tree lacks. So as you suggested, it lands here, the last commit that can make it accurate: both consumers named, both described as preferring `pr_number` and falling back, and no PR numbers left in the prose to go stale the moment this merges.

Verified both consumers actually exist at this head before rewriting the claim about them. Comment only, no behaviour change.

CamSoper:
@claude #update-review

Rebase-only refresh — #21003 took another review-fix commit, so this branch was rebased onto it. No change to this PR's own diff; the last review of this head came back clean (0 outstanding, 0 low-confidence).

CamSoper:
@claude #update-review

Rebase-only refresh again — #21003 took a comment-and-layout commit. No change to this PR's own diff.

PR #20899, opened 2026-08-14
Add a copy-prompt agent CTA to /docs/get-started/

jeffmerrick:
### Proposed changes

Adds a copy-prompt component to the Get Started page.

* `{{< copy-prompt >}}` takes the prompt as inner content, with optional `label` and `track` params. `copy-prompt.markdown.md` is the twin `layouts/docs/single.md` needs since it renders via `.RenderShortcodes`, and emits a fenced `text` block.
* `static/onboard.md` is a byte-identical copy of the same file in #20662, and `theme/src/ts/copy-text.ts` differs only by the analytics call below. That PR will need to drop its copies on rebase.
* `tracking.ts` registers `document.querySelectorAll("a")` only, so a `data-track` on a button reaches nothing. `copy-text.ts` fires its own `copy-text` event on a successful copy, named by the control's `data-track`, the way `copybutton.ts` emits `copy-code-block`. The button on this page is `docs-copy-onboarding-prompt`. #20662's copy button carries `data-role="cta-agent-prompt"` and is currently untracked.
* Dark mode uses explicit `dark:` variants for the grays, and the automatic flip for `text-violet-primary` on the confirm icon.
* AGENTS.md §Dark mode listed `text-gray-950` / `bg-white` / `border-gray-200` / `bg-gray-50` among the utilities auto-remapped in the dark block, but no such remap exists - `--color-violet-primary` is the only `--color-*` scale `_docs-theme.scss` re-points. That bullet now lists the three flips that do happen, including the `border` vs `border-gray-200` specificity case.
* "Choose a cloud provider to get started:" promoted to an H2, so the new "Or start with a coding agent" section reads as the alternative to it rather than a sibling of "Learn more".
* The copy button needs an assets rebuild to work locally (`make serve-all`), since `copy-text.ts` is new to the bundle.

### Related issues (optional)

Fixes #20898
Relates to #20327

Screenshot 2026-08-14 at 3 03 53 PM

PR #21017, opened 2026-08-20
Reland the copy-prompt agent CTA on the homepage hero

### LINKED ISSUE #20327: Add a copy-prompt CTA to the home page for agent onboarding
Add a CTA to the home page that copies this one-line prompt to the clipboard, for pasting into a coding agent:

```
Fetch https://pulumi.com/onboard.md and follow its instructions to get me started with Pulumi.
```

Two notes:

- The hosted `onboard.md` ships separately. If it lands at a different URL, the line changes to match.
- Spec with the full flow and the current `onboard.md` draft (internal Notion): [Onboarding Prompt](https://app.notion.com/p/399fdbdf1cce80a4a95fcd3ae6e9dc8e)

jeffmerrick:
Screenshot 2026-08-20 at 9 13 53 AM

### Proposed changes

Relands #20662, which was reverted in #21016 for an unintended merge rather than for a problem with the change. The commit is a revert of that revert, so the four files match their state at the original merge exactly.

* Groups the homepage hero's buttons into a "For humans" card and adds a "For agents" card whose button copies an onboarding prompt, driven by four optional front matter keys (`cta_card_label`, `cta_agent_label`, `cta_agent_prompt`, `cta_agent_prompt_label`).
* The CTA row moves into `partials/hero-ctas.html`, shared by the split and centered hero layouts. Without those keys it renders the plain button row as before, so the other 12 `type: hero` pages are unchanged.
* The split hero's columns go from `xl:w-1/2` on both to `xl:w-[55%]` / `xl:w-[40%]`, plus `w-full min-w-0` on the left column below xl. Applies to all 7 split heroes, not just the homepage - it fixes a content-sized left column that could overflow the viewport and shift width whenever its own text changed.
* `static/onboard.md` (the prompt's target) and `theme/src/ts/copy-text.ts` came in with #20899 and were not reverted, so nothing here re-adds them.
* `.hero-cta-card-label` breaks the card border with its own white background, so the cards only read correctly on a white section background.

### Related issues (optional)

Fixes #20327

### DIFF
--- a/content/_index.md
+++ b/content/_index.md
@@ -10,6 +10,10 @@ sections:
     layout: split
     cta_secondary_text: Download open source
     cta_secondary_link: /docs/install/
+    cta_card_label: For humans
+    cta_agent_label: For agents
+    cta_agent_prompt: Fetch https://www.pulumi.com/onboard.md and follow its instructions to get me started with Pulumi.
+    cta_agent_prompt_label: Copy prompt
     badge_highlight_text: "Latest release:"
     badge_text: "Full support for Terraform and HCL"
     badge_link: /releases/terraform-state-backend-modules-hcl/
--- a/layouts/partials/hero-ctas.html
+++ b/layouts/partials/hero-ctas.html
@@ -1,0 +1,70 @@
+{{/*
+    Hero CTA row — shared by the split and centered branches of template-hero.
+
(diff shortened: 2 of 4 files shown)


CamSoper (approved):
This still looks great to me :)

PR #21055, opened 2026-08-21
Correct pre-built policy pack coverage, and enforce the allowlist against the registry

CamSoper:
The list that decides which pre-built policy packs get a documentation page, `data/policy_packs.yaml`, is hand-maintained and was never checked against what Pulumi actually publishes. It had drifted in both directions: frameworks we ship for three clouds documented as AWS-only, and packs published privately to the org carrying public reference pages.

This corrects the drift and adds a check so it can't recur silently.

## What's here

| What | Why |
| --- | --- |
| **Corrected pack coverage** on the pre-built packs table, the Policies and Insights overviews, and the per-cloud AWS/Azure/Google Cloud pages | The generated reference pages under `/docs/reference/pre-built-policy-packs/` have been correct all along; the hand-maintained summary pages understated them. A reader comparing the two would conclude we ship less than we do. |
| **Removed `hitrust-awsnative`** from the table, the AWS integrations page, and the allowlist | Flagged in review by @danbiwer: it is private to the org, not a shipped pack, and docs automation had generated a public reference page for it. Since deleted from the org, which is why the nightly `Policy Packs - Update Policies` job currently fails on `master` — merging this clears it. |
| **Repointed CIS Kubernetes GCP** from `cis-kubernetes-gcp` to `cis-kubernetes-google-cloud` | Same defect, opposite direction. `policy-packs-internal` publishes under `cis-kubernetes-gcp`, but that name is private, so the documented page was built from a pack customers cannot enable. The public pack is `cis-kubernetes-google-cloud`. Confirmed with @danbiwer, who is renaming it upstream so releases land on the public name. Both carry the same 50 policies and the slug is unchanged, so no page content or URL moves. |
| **Registry check** in `scripts/fetch-policy-packs.js` | The three fixes above are the same bug three times, found by hand each time. `/api/orgs/{org}/policypacks` — which this script already reads — returns products and privately published packs indistinguishably. The registry listing distinguishes them with `source: "pulumi"`, so the build can now enforce what a comment used to ask for. |
| **Tests for that check**, plus a repair to `scripts/run-unit-tests.sh` | The check guards what reaches the public site and had no regression coverage. Two of the 15 cases fail against the previous commit. The runner turned out to be broken and unreferenced — it ran `yarn --cwd components test` against a directory that no longer exists — so it is repaired and wired into `make test`. |
| **Two prose corrections** in `setting-up-for-success.md` | Pre-existing findings on untouched lines, surfaced by the pre-merge review: an unsourced compliance claim, and enforcement levels named "warning or error" rather than Pulumi's `advisory` / `mandatory` / `remediate`. |

## Coverage corrections

| Framework | Table said | Actually ships |
| --- | --- | --- |
| ISO/IEC 27001:2022 | AWS | AWS, Azure, Google Cloud (238 + 158 + 153 = 549 policies) |
| PCI DSS v4.0.1 | AWS | AWS, Azure, Google Cloud |
| NIST SP 800-53 | AWS, Google Cloud | AWS, Azure, Google Cloud |
| CMMC 2.0 | *absent* | AWS |

## How the check works

Every entry in `data/policy_packs.yaml` must come back from `/api/registry/policypacks?orgLogin={org}` with `source: "pulumi"`, and every pack that does must have either a page or a recorded exemption. Both directions fail the build.

- **Not `publisher`.** It is `"pulumi"` for every pack in the org, including the private ones, so filtering on it would admit exactly what this excludes. It is the field a reasonable person reaches for first, so there is a comment saying not to.
- **`source` is set by an admin command**, not by whoever publishes the pack, which is what makes it safe to fail a build on.
- **Fails closed.** An unreadable, changed, or paginated response stops the job rather than falling back to publishing unverified pages. Reading one page of a paginated listing is worse than not reading it: the forward check would fail packs that are fine, and the reverse check would silently miss the undocumented product it exists to catch.
- **`undocumented:`** records products that deliberately get no generated page. Both `pack:` and `why:` are required, so the list cannot be used to quietly silence the check.

`data/policy_packs.yaml` remains the editorial layer — slugs, titles, meta descriptions, menu placement, framework grouping. The registry carries none of that (`displayName` is empty on every record). What changed is that it is no longer the source of truth for *which* packs are documented.

## Verification

Run against the live listing for the `pulumi` org: 36 packs, 23 with `source: "pulumi"`. All 22 documented packs plus the one recorded exemption account for exactly those 23, with no products left undocumented. `make lint`, `make test-unit` (15/15), and `prettier --check` pass.

## Relationship to #21120

#21120 adds an **Edition** column to the same table and per-section `feature:` keys to `data/policy_packs.yaml`. That PR notes it left out the packs missing from the table pending confirmation they are GA — the registry query above is that confirmation, and those packs are what this PR adds. The two conflict on `pre-built-packs.md` and `policy-packs/_index.md`; this branch rebases onto #21120 once it lands, and the resolved table is this PR's coverage with that PR's column.

## Follow-ups, not in this PR

- Several pages still carry the pre-correction framework list and will read as incomplete once this lands: the "Policy as code" row across `content/docs/iac/comparisons/*.md`, and `content/what-is/what-is-policy-as-code.md` (L199).
- pulumi/policy-packs-internal#211 records four packs that failed to publish in the 2026-07-09 release batch and were never retried, one of them a shipped pack. Unrelated to this diff.


content/docs/integrations/clouds/aws/_index.md
```diff
@@ -89,6 +89,9 @@ For AWS:
 - [NIST 800-53 for AWS](/docs/reference/pre-built-policy-packs/nist/aws/)
 - [PCI DSS for AWS](/docs/reference/pre-built-policy-packs/pci-dss/aws/)
 - [HITRUST CSF for AWS](/docs/reference/pre-built-policy-packs/hitrust/aws/)
+- [HITRUST CSF for AWS Native](/docs/reference/pre-built-policy-packs/hitrust/aws-native/) — for the AWS Cloud Control (`aws-native`) provider.
```
danbiwer: This isn't actually published.  It exists in our repo privately.  I noticed it's appearing in [references](https://www.pulumi.com/docs/reference/pre-built-policy-packs/hitrust/aws-native/), too.  Would you mind removing that in this PR?  

I think some docs automation might have picked it up @jkodroff

jkodroff: Good catch (sorry for the Claude-speak). I'll hit you up on Slack to figure out the source of truth on what should actually be published.

CamSoper: Done in 93d5904 — removed from the table, this page, and `data/policy_packs.yaml`, which is what was generating the reference page. Also dropped the fetched policy data and its lastmod entry, and added an S3 redirect from `/docs/reference/pre-built-policy-packs/hitrust/aws-native/` to the HITRUST AWS page, since the old URL is live today.

On how it got there, since it's relevant to the source-of-truth question: `data/policy_packs.yaml` is a hand-maintained allowlist, and its rule read "only document packs that are built from the policy-packs-internal repo, **i.e.** the ones we ship to customers" — treating those two as one condition. `hitrust-awsnative` satisfies the first and not the second, and nothing available to the docs side distinguishes them: it's built there, it published cleanly in the 2026-07-09 batch, and the Pulumi Cloud API returns it like any other pack, which is how the nightly fetch picked up 114 policies for it. So the automation wasn't working from a bad signal — it was working from the only signal it has, which doesn't carry the private/public bit at all.

I've reworded the rule as two independent conditions and recorded the exclusion at the bottom of that file, so an audit that diffs the org listing against the repo doesn't re-add it. But that's a comment telling a human to go ask, not a check — if there's a durable way to mark a pack private in `policy-packs-internal` (a `private: true` in its `package.json`, a naming convention, anything the fetch script could read), that'd be a much better fix than trusting the next person to read a comment. Happy to wire that up on the docs side once you and @jkodroff land on what the source of truth should be.

Worth a look while you're in there: the same batch had four packs fail to publish and never get retried, one of them `aws-organizations-tag-policies` v2.0.1 — pulumi/policy-packs-internal#211. Found by an agent, unverified by a human, so treat it accordingly.


---

CamSoper:
@claude #update-review

Two pushes since the last review:

- **acd985ff** — merged `master` to clear a conflict. Upstream restructured `content/docs/administration/`, moving `setting-up-for-success.md` from `onboarding-guide/` to `get-started/`; git followed the rename and carried this branch's edit with it. Nothing resolved by hand. That also relocates the two pre-existing findings you flagged (the "AWS-reviewed" claim and the "warning or error" enforcement levels) — both are still there, still on untouched lines, now at `content/docs/administration/get-started/setting-up-for-success.md`.
- **93d5904d** — removed the HITRUST **AWS Native** pack after @danbiwer confirmed in review that it's private and not shipped to customers. Note this reverses your triage of the `[L92-94]` finding: the verifier called it contradicted, you (and I) overruled that on the grounds that `data/policy_packs.yaml` defines the pack and the data file exists — and the verifier's conclusion was right for a reason neither of us could see from this repo. The pack really is built, really does publish, and really is returned by the API; it just isn't public. Worth noting because the reasoning that overruled it looked sound and was still wrong.

The removal spans the table, the AWS integrations page, `data/policy_packs.yaml`, `data/policy_pack_policies/hitrust-awsnative.json`, and the lastmod ledger, plus an S3 redirect for the now-removed `/docs/reference/pre-built-policy-packs/hitrust/aws-native/`. The remaining nine added pack links are unchanged and still resolve against the allowlist.

Not disputing anything else. The out-of-scope note about `content/docs/iac/comparisons/*.md` and `content/what-is/what-is-policy-as-code.md` is recorded in the PR body as a follow-up.


---

data/policy_packs.yaml
```diff
@@ -251,6 +247,12 @@ sections:
 #                                   guide, not a policy listing. Stays hand-maintained.
 #   cis-kubernetes-google-cloud     A stale earlier publish, superseded by
 #                                   cis-kubernetes-gcp. See the cis-kubernetes section.
+#   hitrust-awsnative               Built here and resolvable through the API, but NOT a
```
jkodroff: It's resolvable through the API b/c someone published it specifically to the `pulumi` org. We either need an allowlist on the docs end, or a "clean" org for reading things from API endpoints.

jkodroff (approved):
Approving, since this fixes the most important mistake, but it's not a full solution.

CamSoper:
@jkodroff I'm rolling in the allowlist idea

data/policy_packs.yaml
CamSoper: Built the allowlist option, in `cd00d9a2`, using the `source` field you and @danbiwer landed on in Slack.

`data/policy_packs.yaml` was already an allowlist — it's how we document 22 of the packs that repo carries rather than all of them. What it didn't have was any check on its own entries, so a human adding a wrong row was enough. `fetch-policy-packs.js` now requires every entry to come back from `/api/registry/policypacks?orgLogin=<org>` with `source: "pulumi"`, and fails naming the offenders and their actual source.

Verified against the two records Dan pasted, including both failure modes from this incident — a `source: "private"` pack, and a pack missing from the listing entirely, which is `hitrust-awsnative`'s state now that you've deleted it.

Two things from your thread are captured as comments in the code, because both are easy to get wrong later:

- **Not `publisher`.** It's `"pulumi"` for everything in our org, so filtering on it admits the private packs too. It's the more obvious-looking field, so there's a comment saying don't switch to it.
- **Your spoofing question is what made a hard failure defensible.** Dan's answer — `source` is admin-set, not set by whoever publishes — is why the check throws rather than warns. If a passing engineer could set it, gating the build on it would be theater. That reasoning is in the comment too, so it survives.

The check fails closed: an unreadable or changed response stops the job rather than publishing unverified pages.

I couldn't exercise the live call — both endpoints need a token this environment doesn't have — so the parsing is written against Dan's records and unit-tested, and the envelope key is accepted loosely since only the record shape was confirmed. First real run will settle it, and it's on the nightly, so that's tonight.

Which brings up something time-sensitive: deleting the pack from the org means `master` is now broken for that job. `fetch-policy-packs.js` throws when an allowlisted pack isn't in the org, and `master` still carries the entry — so the 07:30 UTC run fails and pings #docs-ops. This PR is the fix.

One more, tangential: Dan's `aws-organizations-tag-policies` record reads `version: 1.0.0`, while the repo is at 2.0.1 and there's a GitHub release for 2.0.0. If that was a live listing rather than an illustrative paste, the published version is lagging further than pulumi/policy-packs-internal#211 estimates and I should correct that issue. If it was just a paste, ignore me.


---

CamSoper:
@claude #update-review

One new commit since the last refresh request, plus scope this PR didn't have before — it now touches `scripts/fetch-policy-packs.js`, so the infra lane applies where it previously didn't.

**`cd00d9a2` — enforce the allowlist against the registry listing.** `data/policy_packs.yaml` decides which packs get a generated reference page, but its rule was prose and nothing validated its entries. `/api/orgs/<org>/policypacks`, which the script already reads, returns products and privately published packs indistinguishably — that's how `hitrust-awsnative` got a public page. The script now additionally requires every allowlisted pack to come back from `/api/registry/policypacks?orgLogin=<org>` with `source: "pulumi"`, and throws naming any that don't.

Three things worth pointing your infra checks at:

- **The live call is unrun.** Both endpoints need a token this environment doesn't have. Parsing is written against two real records @danbiwer pasted in Slack and unit-tested against them; the record shape (`name`, `source`) is confirmed, the envelope key is not, so it accepts a bare array, `policyPacks`, or `items`. If you think that tolerance is wrong — that it should pin one shape and fail loudly — say so, that's a genuine judgment call I'd rather have checked.
- **It fails closed by design.** Unreadable or changed response stops the nightly rather than falling back to publishing unverified pages. That's deliberate, but it does mean an upstream API change takes out the sync until someone fixes the parser. Worth confirming that tradeoff reads right to you.
- **`publisher` is deliberately not used.** It's `"pulumi"` for every pack in our org, so filtering on it would admit the private ones. There's a comment saying so, since it's the field a reasonable person would reach for.

Also note this reverses your earlier triage of the `[L92-94]` finding, which I flagged last refresh and which now has a second-order consequence: the pack is gone from the diff entirely, and `data/policy_packs.yaml` is down from 23 documented packs to 22. Any claim you extract about pack counts should come from the current file, not the earlier review.

Not disputing anything outstanding. The two pre-existing findings in `setting-up-for-success.md` are still there, still on untouched lines, now under `content/docs/administration/get-started/` after the upstream move. The out-of-scope note about `content/docs/iac/comparisons/*.md` and `content/what-is/what-is-policy-as-code.md` is recorded in the PR body as a follow-up.


---

data/policy_packs.yaml
CamSoper: > All you have to do is look for `source:pulumi`. Don't need a static allowlist.

Agreed on the gate, and `e7794493` makes that true in both directions. One wrinkle on deleting the file outright, though.

`data/policy_packs.yaml` isn't only deciding *which* packs — it's also carrying everything the page generator needs and the API doesn't have: URL slug, page title, H1, meta description, menu parent, weight, and the framework grouping that puts `cis-aws` / `cis-azure` / `cis-google-cloud` under one "CIS 8.1" section. Worth noting `displayName` is `""` on both records Dan pasted, so we can't even derive titles from the listing.

So the file stays as the editorial layer and stops being the gate. `source: pulumi` now decides membership, checked both ways:

- a documented pack that isn't a product → fail (this is what shipped in `cd00d9a2`)
- **a product nobody documented → fail** (new in `e7794493`)

The second one is the same bug this PR was opened for, just pointing the other way: we shipped ISO 27001 for three clouds and the docs said AWS. Under the old design a new product pack could ship and docs would stay silent indefinitely. Now the nightly fails until someone either adds an entry or records why not.

"Why not" is now a machine-readable `undocumented:` list rather than a comment, holding the two genuine omissions — `aws-organizations-tag-policies` (hand-written prose page, not a policy listing) and `cis-kubernetes-google-cloud` (the stale publish superseded by `cis-kubernetes-gcp`). If either drops out of that list the build fails, so they're enforced rather than remembered.

On your other message — yes, `publisher: pulumi` was the trap. It's `"pulumi"` for every pack in the org including the private ones, so it would have admitted exactly what we were trying to exclude. The check uses `source` only, with a comment saying don't switch it, since `publisher` is the more natural-looking field.

Last night's run failed exactly as expected: `pack "hitrust-awsnative" is in data/policy_packs.yaml but is not published to the "pulumi" org` ([run](https://github.com/pulumi/docs/actions/runs/32822995187/job/97724931434)). `master` still carries the entry you deleted from the org, so that repeats nightly at 07:30 UTC until this merges.


---

CamSoper:
@claude #update-review

One new commit, `e7794493`, extending the registry check in `scripts/fetch-policy-packs.js` from one direction to two.

Previously it only caught a documented pack that isn't a published product. It now also fails when a pack the registry returns as a product has no reference page — the same defect this PR exists to fix, pointing the other way (we shipped ISO 27001 for three clouds; the docs said AWS). Deliberate omissions moved from a prose comment into a machine-readable `undocumented:` list in `data/policy_packs.yaml`, holding `aws-organizations-tag-policies` (hand-written page) and `cis-kubernetes-google-cloud` (superseded publish).

For the infra lane: both directions were exercised against a simulated listing built from @danbiwer's real records — happy path, undocumented new product, documented private pack, documented deleted pack, and exclusions removed. The live API call is still unrun here (no token in this environment), unchanged from the last refresh.

One thing worth your judgment: the reverse check is a hard failure, so a pack published as a product with no docs entry breaks the nightly until someone acts. That's deliberate — silence is how the original bug survived — but it does mean the policy-packs team can break the docs build by publishing, which is a real coupling. If you think that should be a warning instead, say so.

`data/policy_packs.yaml` also gained a top-level `undocumented:` key. `content/docs/reference/pre-built-policy-packs/_content.gotmpl` only iterates `.sections`, so generation is unaffected, but flag it if you see a consumer I missed.


---

CamSoper:
@claude #update-review

Ran the `curl` you asked for, against the live `pulumi` org — 36 packs, 23 `source: "pulumi"`, 13 `"private"`. It settled three findings and found a fourth. All eight items dispositioned in `13587ab`.

**Fixed**

- **L156-165** — you were right twice. The envelope is `policyPacks` and nothing else, so the bare-array/`items` tolerance was covering an unknown that isn't one; `continuationToken` is real and this script ignored it. Your read of the asymmetry is the part I'd missed: the forward check fails closed on a partial read while the reverse check fails *open*, silently missing exactly the undocumented product it exists to catch. Now pins the envelope and refuses a paginated response outright. (`continuationToken` is null at 36 packs today, so this is defensive, not currently firing.)
- **L166-172** — both. Rows-but-no-products now throws a shape error rather than surfacing as "22 documented packs are privately published". `nonProducts` stores `"(no source field)"` so present-without-`source` stops being reported as absent.
- **L224-236** — `undocumented:` now requires both `pack:` and `why:`, rejects a pack that is also documented, and warns on a stale exemption whose pack is no longer a product.
- **L266-283** — the 12 asserted non-products all come back `source: "private"`. Reverse check clean, no undocumented products. On your warning-vs-fail question: **keeping the hard failure**, and I took your point about the header — THE RULE now carries a third bullet, that the file is a claim about the whole product set rather than just an allowlist.
- **L177** — 21 of 22 documented packs confirmed `source: "pulumi"`. And you caught me squarely on the tests: that commit message described simulations I never committed. Now 15 real cases on `node:test`, covering both gate directions and every shape branch; two of them fail against the previous commit. `scripts/run-unit-tests.sh` turned out to be broken *and* unreferenced — it ran `yarn --cwd components test` against a directory that no longer exists — so it's repaired and wired into `make test`.
- **L253-255** — true as written now that the validation landed.
- **Pre-existing L23/L27** — fixed both while in the file.

**The fourth thing.** 21 of 22, not 22 of 22. `cis-kubernetes-gcp` is `source: "private"`, and we generate a public page from it — same class as `hitrust-awsnative`, still live. The product is `cis-kubernetes-google-cloud` (`source: "pulumi"`, v1.0.0); the comment at `data/policy_packs.yaml:100-102` has it backwards. So the gate currently throws on this branch, correctly. That's a product decision, raised with @danbiwer on #21120 and not something to fix in code.

**Not addressed, deliberately.** #21120 rewrites the same packs table with an Edition column and is in review; this branch rebases onto it once it lands, and the table resolution is this PR's coverage × that PR's column. No point resolving it twice.

---

CamSoper:
@claude #update-review

Three commits since the last review (`13587ab`). All five findings dispositioned; `--require-clean` exits 0.

**Fixed**

- **L92** — this is the one worth calling out, because it made the previous round's fix cosmetic. Wiring the tests into `make test` did nothing for PRs: its only caller is `scheduled-test.yml`, where `Run the tests` is gated on `static/programs/` having changed, so this content-only PR reported that check green in seconds without running a case. `make test-unit` now runs on every PR from `pull-request.yml`'s build job. I took your second option rather than reshuffling the scheduled workflow — that job is already checked out with Node, so it needs no second checkout.
- **L241-247** — hoisted the `undocumented:` integrity checks above the forward registry comparison. Your framing is the right one: validate our own inputs, then compare them against the world. Added a test that fails under the old ordering.
- **L155** — Makefile comment no longer describes the suite the same commit removed.

**Fixed in `ea95624`, which landed after this review ran**

- **L103 / L104-106** — both the gate failure and the inverted comments. I went with your option 2, and @danbiwer confirmed why it's the right one: `cis-kubernetes-google-cloud` is the public pack and was manually uploaded, which is why it sits at 1.0.0. `policy-packs-internal` publishes under `cis-kubernetes-gcp`, but that name was never made a product, so every automated release landed on a private pack while the public one went stale. He's renaming it upstream so releases reach the public name.

  Both packs carry the same 50 policies — 1.0.1 was a dependency bump — and `slug: google-cloud` is unchanged, so no page content or URL moved. The `undocumented:` entry was deleted rather than reworded, since the pack now has a page. You were right that it was load-bearing and inverted; it would have been the thing someone deleted while tidying stale publishes.

  You were also right that #21120 wasn't the resolution. It's merged and this branch is rebased onto it, but that only settled the table, not which pack is a product.

**Merge resolution** — #21120 conflicted on two files. `pre-built-packs.md` takes this branch's rows and coverage with that PR's Edition column; CMMC 2.0 gets Business Critical like the other compliance frameworks. `policy-packs/_index.md` takes this branch's framework list with that PR's edition-split sentence. `data/policy_packs.yaml` auto-merged with all 8 `feature:` keys intact.

**Verification** — gate dry-run against the live registry passes (36 packs, 23 products, all accounted for by 22 pages plus 1 recorded exemption); 16/16 unit tests; lint and prettier clean.

---

CamSoper:
@claude #update-review

One commit since the last refresh, `2145a5e`, fixing the CI wiring I added for **L92**.

The step ran before anything was installed and this module required `js-yaml` at load, so the suite died on `MODULE_NOT_FOUND` before executing a case — the build failed, and the coverage was still effectively absent.

Installing first would have been wasted work (`make ensure` begins with `clean`, so anything installed ahead of `make ci_pull_request` is discarded), and moving the tests after the build would surface a failure only once the site had deployed. `js-yaml` is used in exactly one place, reading the allowlist inside `main()`, and none of the exported helpers touch it — so it's required there, and the module is now importable with no `node_modules`. That's what the workflow comment already claimed; it just wasn't true yet.

Verified the way CI does it: hid `node_modules/js-yaml`, ran the suite, 16/16. `main()` still parses the allowlist. Build is green and the job log shows all 16 cases executing on this PR, which was the point of the finding.

No other findings affected; the worklist was clean before this push and the change touches only the require site and a workflow comment.

---

jkodroff (approved):

CamSoper:
@claude #update-review

Both fixed in `723be65`, comment-only.

- **L254-260** — you're right, and it's the ironic one: hoisting the integrity checks carried the reverse-direction paragraph along with the code, leaving text about `undocumentedProducts` sitting above `staleExemptions`. Moved back to the check it describes, blank line restored.
- **L283-301** — recorded `cis-kubernetes-gcp` with the reason. Your argument for why is the right one: it's the plausible-looking half of the pair and the exact mistake this change just fixed, and no check can catch it, since a private pack nothing references is invisible to both directions of the gate.

  On the count — the header was wrong in a way the suggestion didn't quite capture. It claimed every name in that block comes back `source: "private"`, but `hitrust-awsnative` isn't private, it's gone from the org entirely. Reworded to say which 13 are private (`cis-kubernetes-gcp` plus the 12 named below) and to call out the exception. Verified against the live listing: 13 private, all 13 accounted for, none unlisted.

Re-ran the real fetch against the live API afterward: exit 0, and `data/policy_pack_policies/` and the lastmod ledger are still byte-identical to what the API returns. 16/16 tests, lint and prettier clean.

Worklist is clean. Merging once this comes back.

---

PR #21059, opened 2026-08-21
Add Pulumi Context API launch blog post and changelog entry

lblackstone:
The Pulumi Context API enters public preview on 2026-08-26. This adds the launch blog post and changelog entry.


.claude/commands/blog-feature-image/assets/icons/graph-fill.svg
cnunciato, line 1: @jeffmerrick Tagging you just to make sure this all looks good!

content/blog/pulumi-context-api/index.md
```diff
@@ -0,0 +1,111 @@
...
+---
+
+Every platform team fields the same questions: What is running? What breaks if we change this? What can we safely delete? The answers exist, but they're scattered across state files, cloud consoles, and the memories of whoever set things up. Today we're launching the Pulumi Context API, which connects what Pulumi knows about your infrastructure into a single graph you can query. It's designed agent-first: [Pulumi Neo](/product/neo/), our infrastructure agent, uses it out of the box, and the API itself teaches any other agent everything it needs in one request. It's available in preview for organizations on the Enterprise and Business Critical editions.
+
+<!--more-->
+
+## Answers that span all your infrastructure
+
+Pulumi already holds a detailed picture of your infrastructure. It knows the resources your programs manage, the dependencies between them, how stacks consume each other's outputs, and which resources in your cloud accounts [Pulumi Insights](/docs/insights/) has discovered outside IaC entirely. Each of those views is useful on its own. The Context API connects them, so you can ask the questions that cut across all of them:
```
cnunciato: Insights → Discovery:

```suggestion
Pulumi already holds a detailed picture of your infrastructure. It knows the resources your programs manage, the dependencies between them, how stacks consume each other's outputs, and which resources in your cloud accounts [Discovery](/docs/insights/discovery/) has found outside IaC entirely. Each of those views is useful on its own. The Context API lets you ask questions that cut across all of them:
```

content/releases/changelog/2026-08-26-context-api.md
```diff
@@ -0,0 +1,16 @@
...
+date: 2026-08-26
+meta_desc: "The Pulumi Context API is now in preview: one queryable graph spanning IaC state, stack dependencies, and cloud resources discovered outside IaC."
+authors:
+    - levi-blackstone
+editions:
+    - enterprise
+    - business-critical
+---
+
+The [Pulumi Context API](/docs/insights/context-api/) is now available in preview. It connects everything Pulumi knows about your infrastructure, the resources your programs manage, the dependencies between them, how stacks consume each other's outputs, and the resources [Pulumi Insights](/docs/insights/) discovers in your cloud accounts outside IaC, into a single graph you can query.
```
cnunciato: Insights → Discovery:

```suggestion
The [Pulumi Context API](/docs/insights/context-api/) is now available in public preview. It connects everything Pulumi knows about your infrastructure, the resources your programs manage, the dependencies between them, how stacks consume each other's outputs, and the resources [Discovery](/docs/insights/discovery/) finds in your cloud accounts outside IaC, into a single graph you can query.
```

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,670 @@
...
+}
+```
+
+{{< /details >}}
+
+`target.absent: true` inverts the traversal, returning the stacks from which the walk reaches no consumer at all, the candidates for cleanup. Treat this as a lead, not a final verdict: absence is judged only against what the calling identity can see, so a consumer the caller lacks permission to see won't disqualify a stack from appearing here. Only trust the result when `meta.resultMode` is `exact`; a `truncated` answer means the check isn't reliable, so narrow the query and re-run before acting on it.
+
+## How much of each cloud account is covered by IaC, versus discovered separately?
+
+How much infrastructure did Pulumi Insights find in each cloud account that no Pulumi program declared?
```
cnunciato: Terminology: "Pulumi Insights" → **Discovery** (retired name).

```suggestion
How much infrastructure did Discovery find in each cloud account that no Pulumi program declared?
```

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,670 @@
...
+    "resultMode": "exact",
+    "schemaVersion": "2026-08-21",
+    "visibility": "complete"
+  }
+}
+```
+
+{{< /details >}}
+
+`in_stack` only exists for resources a Pulumi program deployed, so walking it outward from a resource returns a stack if, and only if, that resource is Pulumi-managed. A resource discovered by cloud scanning produces an empty `stacks` frontier every time, because a scan deploys nothing and creates no such edge. This is the reliable way to distinguish IaC-managed resources from ones Insights found on its own.
```
cnunciato: Terminology: "Insights" → **Discovery** (retired name).

```suggestion
`in_stack` only exists for resources a Pulumi program deployed, so walking it outward from a resource returns a stack if, and only if, that resource is Pulumi-managed. A resource discovered by cloud scanning produces an empty `stacks` frontier every time, because a scan deploys nothing and creates no such edge. This is the reliable way to distinguish IaC-managed resources from ones Discovery found on its own.
```

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,118 @@
+---
+title: Context API
+title_tag: "Context API | Discovery & Governance"
```
cnunciato: Lowercase "g" here. (See the [Brand guide](https://brand.pulumi.com/voice/names-and-terminology/#discovery--governance)) — which now has an [MCP server](https://brand.pulumi.com/mcp-server/)! Cam's working on getting it integrated into the review process.)

```suggestion
title_tag: "Context API | Discovery & governance"
```

.claude/commands/blog-feature-image/assets/icons/graph-fill.svg
jeffmerrick: ✅ Looks good to me!

content/blog/pulumi-context-api/index.md
```diff
@@ -0,0 +1,111 @@
...
+
+        Which stacks break if we upgrade this provider? How much of our infrastructure lives outside IaC? What has no dependents and can be retired? Each one is a single query.
+    bluesky: |
+        We just launched the Pulumi Context API: everything Pulumi knows about your infrastructure as one queryable graph, built agent-first. Any agent can learn it in one request, and Neo uses it out of the box.
+
+        We wrote up what it can answer.
+---
+
+Every platform team fields the same questions: What is running? What breaks if we change this? What can we safely delete? The answers exist, but they're scattered across state files, cloud consoles, and the memories of whoever set things up. Today we're launching the Pulumi Context API, which connects what Pulumi knows about your infrastructure into a single graph you can query. It's designed agent-first: [Pulumi Neo](/product/neo/), our infrastructure agent, uses it out of the box, and the API itself teaches any other agent everything it needs in one request. It's available in preview for organizations on the Enterprise and Business Critical editions.
```
cnunciato:
```suggestion
Every platform team fields the same questions: What is running? What breaks if we change this? What can we safely delete? The answers exist, but they're scattered across state files, cloud consoles, and the memories of whoever set things up. Today we're launching the [Pulumi Context API](/docs/insights/context-api/), which connects what Pulumi knows about your infrastructure into a single graph you can query. It's designed agent-first: [Pulumi Neo](/product/neo/), our infrastructure agent, uses it out of the box, and the API itself teaches any other agent everything it needs in one request. It's available in preview for organizations on the Enterprise and Business Critical editions.
```

content/blog/pulumi-context-api/index.md
```diff
@@ -0,0 +1,111 @@
...
+
+Pulumi already holds a detailed picture of your infrastructure. It knows the resources your programs manage, the dependencies between them, how stacks consume each other's outputs, and which resources in your cloud accounts [Pulumi Insights](/docs/insights/) has discovered outside IaC entirely. Each of those views is useful on its own. The Context API connects them, so you can ask the questions that cut across all of them:
+
+- **Impact**: Which stacks are affected if we upgrade this provider? If this stack changes, what consumes its outputs?
+- **Coverage**: How much of our infrastructure lives outside IaC, and in which accounts?
+- **Cleanup**: Which stacks have no dependents and can be retired?
+
+Each of these is a single query. The graph covers your resources (both Pulumi-managed and discovered), your stacks, and the relationships that connect them, from dependencies and containment to providers and cross-stack references.
+
+A query names its starting nodes (`anchor`), walks the relationships between them (`traverse`), and returns the connected results. Here's one that finds every resource still managed by an AWS provider older than 7.0:
```
cnunciato: Feels like we're missing some connective tissue here. We also seem to be missing the part that explains what the query syntax is, which I believe is based on an open standard.

Also, how does one _submit_ a query? Do I just POST some similarly structured JSON to the endpoint you mention below? Can I use the CLI? Something else? Would be good to address this in the post somehow.

content/blog/pulumi-context-api/index.md
```diff
@@ -0,0 +1,111 @@
...
+    }
+  },
+  "traverse": [
+    { "edgeTypes": ["provided_by"], "direction": "in", "depth": { "min": 1, "max": 1 }, "alias": "managed" }
+  ],
+  "return": { "select": ["anchor", "managed"] }
+}
+```
+
+The response includes the matching nodes, the edges between them, and, when you ask for them, the paths that connect them, so an answer shows its evidence. Each response also reports whether it is complete. That means an empty answer to "does anything depend on this?" is one you can act on.
```
cnunciato: Any reason not to show an example response as well? Doesn't have to be here (could be down below where the requests are), but feels like something we should show (since the data is the valuable part).

content/blog/pulumi-context-api/index.md
```diff
@@ -0,0 +1,111 @@
...
+
+or over plain HTTP:
+
+```bash
+curl -H "Accept: text/markdown" \
+    -H "Authorization: token $PULUMI_ACCESS_TOKEN" \
+    https://api.pulumi.com/api/insights/my-org/graph/schema
+```
+
+Paste the primer into a chat, or put the fetch command in your `AGENTS.md` or `CLAUDE.md` so your agent pulls a fresh copy when it needs one (the primer evolves with the API, so a saved copy goes stale). Your agent can then compose queries from plain-English questions and run them with the same CLI or token access shown above. Validation errors list the legal values and point back at the schema endpoint, so agents recover from their own mistakes without help.
```
cnunciato: Knowing how the underlying machinery works is nice, but it also feels like this is something we shouldn't actually be asking a human user to do. Are we pulling this into a skill somehow? Seems like a good opportunity both to mention that we have skills (so promote/link to them) and to suggest that the reader try this now by installing the skill and submitting a prompt. What do you think?

> Paste the primer into a chat

It's unclear what "the primer" means here. Are we saying that you should run `pulumi api GetGraphSchema` and then paste the response of that call into a new chat?

content/blog/pulumi-context-api/index.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Available in preview
+
+The Context API is available now for every organization on the Enterprise and Business Critical editions. The `pulumi api` commands shown here need [Pulumi CLI](/docs/install/) v3.243.0 or later. Access uses the same permission as [Resource Search](/docs/insights/discovery/search/), so if you can search resources in the console today, you can query the graph. Results are trimmed to the caller's permissions at every step of a query, so a more restricted caller sees less. While the API is in preview the contract can still change; any breaking change will be announced in the [changelog](/releases/).
```
cnunciato:
```suggestion
The Context API is available now for every organization on the Enterprise and Business Critical editions. The `pulumi api` commands shown here need [Pulumi CLI](/docs/install/) v3.243.0 or later. Access uses the same permission as [Resource Search](/docs/insights/discovery/search/), so if you can search resources in the console today, you can query the graph. Results are trimmed to the caller's permissions. While the API is in preview the contract may change; any breaking changes will be announced in the [Changelog](/releases/).
```

content/blog/pulumi-context-api/index.md
```diff
@@ -0,0 +1,111 @@
...
+}
+```
+
+The response includes the matching nodes, the edges between them, and, when you ask for them, the paths that connect them, so an answer shows its evidence. Each response also reports whether it is complete. That means an empty answer to "does anything depend on this?" is one you can act on.
+
+## Ask Neo, or bring your own agent
+
+We expect AI agents to be the primary users of the Context API, and it's designed for them. [Pulumi Neo](/product/neo/) uses it out of the box: ask Neo what breaks if a stack changes, and it queries the graph on your behalf, with the permissions of the user who invoked it.
+
+Any other agent, whether that's Claude Code, Cursor, Codex, or your own automation, can learn the API in one request. The schema endpoint answers in Markdown by default with a short primer covering the node and edge vocabulary, the query grammar, and worked examples. That document is the entire onboarding. Fetch it with the Pulumi CLI:
```
cnunciato: What does "worked examples" mean? Should that be "working"?

content/blog/pulumi-context-api/index.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Get started
+
+If your organization uses [Neo](/product/neo/), just ask it a question. For any other agent:
+
+1. Fetch the primer: `pulumi api GetGraphSchema -F orgName=<your-org>`. The primer is the reference documentation, served by the API itself, so it's always in sync with the deployment answering your queries.
+1. Hand it to your favorite agent.
+1. Ask a question: tell your agent in plain English what you want to know, or compose a query from the primer's examples and run it with `pulumi api GraphQuery -F orgName=<your-org> --input query.json`.
+
+The [Context API documentation](/docs/insights/context-api/) covers access, permissions, and worked query recipes if you want more than the primer.
```
cnunciato: "Worked" seems like the wrong word to be using here somehow — I can't quite tell how. I think what it's trying to say is "working examples", but I think we can assume that if they're here, they also work – right?

```suggestion
The [Context API documentation](/docs/insights/context-api/) covers access, permissions, and query recipes if you want more than the primer.
```

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,118 @@
...
+  insights:
+    parent: insights-home
+    identifier: insights-context-api
+    weight: 30
+pulumi_cloud_feature: context-api
+---
+
+The Context API connects everything Pulumi knows about your infrastructure into a single queryable graph: the resources your programs manage, the dependencies between them, how stacks consume each other's outputs, and the resources [Discovery](/docs/insights/discovery/) has found in your cloud accounts outside IaC entirely. Instead of piecing an answer together from stack state, cloud consoles, and tribal knowledge, you ask the graph directly. Questions like which stacks break if you upgrade a provider, how much of an account lives outside IaC, or which stacks have no dependents and are safe to retire, are each a single query.
+
+The API is designed agent-first. [Pulumi Neo](/product/neo/) queries it out of the box, and any other agent, from Claude Code to Cursor to your own automation, can learn the whole query language in one request by fetching the API's own schema. The rest of this page covers the graph model, how to call the API, and the fetch-the-primer workflow that gets an agent using it without any docs beyond this page.
```
cnunciato: We're in a docs context, so we should probably be linking to the docs page instead. The content here is pretty verbose, too — we could probably trim a bunch to ease the burden on the reader without losing any real detail.

```suggestion
The API is designed agent-first. [Pulumi Neo](/docs/ai/neo/) uses it automatically, and other agents can learn the query language by fetching the API's self-documenting schema. This page covers the graph model, how to call the API, and how to fetch the schema to use it in your prompts.
```

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,118 @@
...
+meta_desc: Query IaC state, stack dependencies, and discovered cloud resources as one graph, built for AI agents.
+menu:
+  insights:
+    parent: insights-home
+    identifier: insights-context-api
+    weight: 30
+pulumi_cloud_feature: context-api
+---
+
+The Context API connects everything Pulumi knows about your infrastructure into a single queryable graph: the resources your programs manage, the dependencies between them, how stacks consume each other's outputs, and the resources [Discovery](/docs/insights/discovery/) has found in your cloud accounts outside IaC entirely. Instead of piecing an answer together from stack state, cloud consoles, and tribal knowledge, you ask the graph directly. Questions like which stacks break if you upgrade a provider, how much of an account lives outside IaC, or which stacks have no dependents and are safe to retire, are each a single query.
```
cnunciato: This feels a bit wordy (and also a tad markety for docs):

```suggestion
The Context API connects everything Pulumi knows about your infrastructure into a single queryable graph, including the resources your programs manage, the dependencies between resources and stacks, and the unmanaged resources [Pulumi Discovery](/docs/insights/discovery/) finds in your cloud accounts. Rather than piece together an answer from multiple data sources on your own, you can instead ask the Context API directly, and get a focused response that minimizes the volume of information that's pulled into your session context.
```

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,118 @@
...
+
+| Relationship | Connects | Meaning |
+|---|---|---|
+| `pulumi:reference` | resource to resource | one resource depends on another |
+| `parent` | resource to resource | containment, such as a component and the resources it creates |
+| `provided_by` | resource to resource | a resource and the provider that manages it |
+| `in_stack` | resource to stack | a resource belongs to a stack |
+| `consumes_outputs_of` | stack to stack | one stack reads another stack's outputs |
+
+A query names a starting point (`anchor`), walks relationships from there (`traverse`), and returns the connected nodes and edges (`return`). The wire format is JSON, not a query language. The full grammar, including field predicates, aggregation, and pagination, lives in the schema the API serves itself, described below.
```
cnunciato: We just called it a query language above, though. Should we change that? 

```suggestion
A query names a starting point (`anchor`), walks relationships from there (`traverse`), and returns the connected nodes and edges (`return`). The query format is JSON. The full grammar, including field predicates, aggregation, and pagination, lives in the schema the API serves itself, described below.
```

Still, I could swear you said this was based on a standard of some kind. "GQL" was it?

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+To run a query, with a selector saved to `query.json`:
+
+```console
+$ pulumi api GraphQuery -F orgName=<your-org> --input query.json
+```
+
+```console
+$ curl -X POST https://api.pulumi.com/api/insights/<your-org>/graph/query \
```
cnunciato: Nothing on this page shows what an actual query looks like. Can we provide a sample?

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,118 @@
...
+    -H "Authorization: token $PULUMI_ACCESS_TOKEN" \
+    -H "Content-Type: application/json" \
+    --data @query.json
```
cnunciato: Nothing on this page shows what an actual query looks like. Can we provide a sample, and then link to the recipes page at the end of this section (a couple of lines down)? I get the disclaimer about the API being fluid, but folks will be wondering this, and they may not think to scroll down the page to read the "Next Steps" section.

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+A response reports whether it's complete: `meta.resultMode` is `exact` or `truncated`. Only an `exact` response can answer a question about absence, such as "nothing depends on this resource." A `truncated` response means the query hit a size limit before it finished, so narrow the anchor or scope and run it again rather than treating a partial answer as final. Each response also carries `meta.schemaVersion`, a date that advances whenever the API's observable behavior changes.
```
cnunciato: This would be much clearer and easier to parse if it were presented as a table that explained what these fields and values mean.

lblackstone:
@pulumi-eon /review

content/blog/pulumi-context-api/index.md
```diff
@@ -0,0 +1,111 @@
...
+
+## The graph will grow
+
+This launch is the first step in a larger effort to make everything Pulumi knows about your infrastructure available as context, wherever you need it. The graph vocabulary is designed to grow without breaking existing consumers. In the future, we plan to bring in more of what Pulumi knows, including [Pulumi ESC](/docs/esc/) environments (trace which stacks a configuration or secret change reaches), teams and roles (who owns what), cloud accounts, and service catalog concepts from [Pulumi IDP](/docs/idp/), our internal developer platform.
+
+## Get started
+
+If your organization uses [Neo](/product/neo/), just ask it a question. For any other agent:
```
lblackstone: Applied, in `ca832c8`.

🤖 AI-assisted (Claude Code)

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,128 @@
...
+| Relationship | Connects | Meaning |
+|---|---|---|
+| `reference` | resource to resource | one resource depends on another, as Pulumi recorded it |
+| `inferred_reference` | resource to resource | one resource may depend on another, inferred by matching a property value against another resource's identifier |
+| `parent` | resource to resource | containment, such as a component and the resources it creates |
+| `provided_by` | resource to resource | a resource and the provider that manages it |
+| `in_stack` | resource to stack | a resource belongs to a stack |
+| `consumes_outputs_of` | stack to stack | one stack reads another stack's outputs |
+
+The two reference types differ in how certain they are. A `reference` edge is a dependency Pulumi recorded as fact. An `inferred_reference` edge is a heuristic lead that may be wrong, because an identifier that doubles as an ordinary value can attract edges that aren't dependencies. Traverse `reference` when you'll act on the answer, and add `inferred_reference` when you want leads worth checking.
```
cnunciato: >an identifier that doubles as an ordinary value can attract edges that aren't dependencies

I don't understand what this means.

> Traverse `reference`

What does "traverse" mean here? Are we really just saying "use" or "look at"?

lblackstone:
@claude #update-review

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Access and permissions
+
+The Context API is read-only. It uses the same permission model as [Resource Search](/docs/insights/discovery/search/): if you can search for a resource in the Pulumi Cloud console, you can reach it through a graph query. Results are trimmed to the caller's permissions at every step of a query, not just at the anchor, so a more restricted caller sees a smaller graph, not an error.
+
+The Context API is in preview, and its contract can still change. Any breaking change will be announced in the [changelog](/releases/).
+
+## Next steps
+
+- [Query recipes](/docs/insights/context-api/recipes/): worked selectors for common questions like provider-upgrade impact, blast radius, and IaC coverage.
```
cnunciato: I really don't get this phrase "worked selectors" (or really just "worked"). Where did it come from? (I'm honestly just curious at this point.)

```suggestion
- [Query recipes](/docs/insights/context-api/recipes/): queries for common questions like provider-upgrade impact, stack relationships, and IaC coverage.
```

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,128 @@
...
+| `in_stack` | resource to stack | a resource belongs to a stack |
+| `consumes_outputs_of` | stack to stack | one stack reads another stack's outputs |
+
+The two reference types differ in how certain they are. A `reference` edge is a dependency Pulumi recorded as fact. An `inferred_reference` edge is a heuristic lead that may be wrong, because an identifier that doubles as an ordinary value can attract edges that aren't dependencies. Traverse `reference` when you'll act on the answer, and add `inferred_reference` when you want leads worth checking.
+
+A query names a starting point (`anchor`), walks relationships from there (`traverse`), and returns the connected nodes and edges (`return`). The wire format is JSON, not a query language. The full grammar, including field predicates, aggregation, and pagination, lives in the schema the API serves itself, described below.
+
+## Calling the API
+
+The Context API exposes two operations, both documented in the [Pulumi Cloud REST API reference](/docs/reference/cloud-rest-api/). Call either one through the [Pulumi CLI](/docs/install/) (v3.243.0 or later) or with a plain HTTP request.
```
cnunciato:
```suggestion
The Context API exposes two operations, both documented in the [Pulumi Cloud REST API reference](/docs/reference/cloud-rest-api/). You can call either one through the [Pulumi CLI](/docs/install/) (v3.243.0 or later) or with a plain HTTP request.
```

Guessing we don't have the actual URL for the REST API docs, because generated, but would be nice to link directly to that once we do.

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,128 @@
...
+
+To fetch the schema:
```
cnunciato: Here is the place to explain both why one would _want_ to fetch the schema (if only briefly, with a link to the section below) and that it returns two (quite different) kinds of data — Markdown for usage, JSON for shape.

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+A response reports its fidelity through two independent fields:
+
+- `meta.resultMode` is `exact` or `truncated`. `truncated` means the answer was clipped before it finished, either at a size limit or by a partial answer from the search backend. Narrow the anchor or scope and run it again rather than treating a clipped answer as final.
+- `meta.visibility` is `complete` or `trimmed`. Read `trimmed` as "the engine can't claim this answer is whole" — usually because the caller's permissions hid part of the graph the query touched. It's absent from an aggregation response, which reports `resultMode` alone.
+
+Both need to be clean before you can answer a question about absence, such as "nothing depends on this resource," since a size limit and a permission trim can each hide a real dependent. Neither speaks to freshness, though: queries answer from the search index, which trails the source of record, so confirm an absence against current stack state before acting on it destructively.
+
+Each response also carries `meta.schemaVersion`, a date that advances whenever the API's observable behavior changes.
```
cnunciato:
```suggestion
Responses contain two fields that indicate the fidelity of the result: 

- `meta.resultMode`: `exact` or `truncated`. `truncated` means the answer was clipped before it finished, either by a size limit or by a partial answer from the search backend. In this case, you should the anchor or scope and run it again rather than treat the result as final.
- `meta.visibility`: `complete` or `trimmed`. `trimmed` means "the engine can't claim this answer is whole" — usually because the caller's permissions hid part of the graph the query touched. It's absent from an aggregation response, which reports `resultMode` alone.

For queries meant to confirm the absence of data (e.g., "nothing depends on this resource"), results should be both `exact` and `complete`. Also note that queries are run against the search index, which may trail sources of record, so be sure to confirm using current stack state before taking destructive actions.

Responses also contain `meta.schemaVersion`, a date that advances as the API's observable behavior changes.
```

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,128 @@
...
+
+The primer is the canonical reference for the query language, not this page. It's generated from the deployment that will actually answer your queries, so it's always in sync. Re-fetch it before composing new queries rather than relying on a saved copy or on examples you've seen before, since the vocabulary and caps can change between schema versions.
+
+That makes onboarding an agent a single step: fetch the schema, hand it to the agent, and let the agent compose and run queries from there. If a query is rejected, the error message usually names exactly what would have been accepted — an unsupported `aggregate.groupBy` field, for instance, comes back with the supported set listed — so an agent can correct itself without a person in the loop. Two rejections don't enumerate: an unrecognized edge type points back at the schema endpoint instead, and an unrecognized field in `match.fields` reports only that the field is unknown.
```
cnunciato: This just feels like too many words, and the copy feels LLM-generated, so it's cumbersome to read and understand. I think all we need to say here is that 

```suggestion
## Fetching and using the primer

In addition to the schema itself, the schema endpoint returns a primer document that helps agents understand how to use the Context API. The primer is written in Markdown, and contains the API's node and edge vocabulary, full query grammar, examples, and any usage caps currently in effect. 

By default:

* The Pulumi CLI returns the primer document, which you can hand directly to an agent. You can request the JSON schema instead by passing `--output json`:

    ```bash
    pulumi api GetGraphSchema -F orgName=<your-org> --output json
    ```
 
* The REST API returns the JSON schema. You can request the Markdown document instead by passing an `Accept: text/markdown` header:

    ```bash
    curl -H "Authorization: token $PULUMI_ACCESS_TOKEN" \
      -H "Accept: text/markdown" \
      https://api.pulumi.com/api/insights/veridian/graph/schema
    ``` 

The schema and primer should be considered the canonical reference for the API at any given moment. Both are kept in sync with the API, so should be fetched before submitting new queries rather than relying on saved copies or existing examples that may be out of date.
```

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Wire it into your agent's instructions
```
cnunciato:
```suggestion
### Wiring the primer into your agent's instructions
```

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,128 @@
...
+Don't guess at fields, edge types, or caps, and don't reuse a schema fetched in an
+earlier session. If a query is rejected, read the error message and correct the
+selector against the schema it names.
+````
+
+Replace `<your-org>` with your Pulumi organization name. See the [query recipes](/docs/insights/context-api/recipes/) for worked examples covering common questions.
+
+## Access and permissions
+
+The Context API is read-only. It uses the same permission model as [Resource Search](/docs/insights/discovery/search/): if you can search for a resource in the Pulumi Cloud console, you can reach it through a graph query. Results are trimmed to the caller's permissions at every step of a query, not just at the anchor, so a more restricted caller sees a smaller graph, not an error.
```
cnunciato:
```suggestion
The Context API is read-only and uses the same permission model as [Resource Search](/docs/insights/discovery/search/). If you can find a resource in the Pulumi Cloud console, you can reach it through a graph query. Results are also trimmed to the permissions of the caller, so a more restricted caller would get a smaller graph rather than an error.
```

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,128 @@
...
+
+The Context API is in preview, and its contract can still change. Any breaking change will be announced in the [changelog](/releases/).
```
cnunciato:
```suggestion
The Context API is in public preview, so its features and behavior may change. Breaking changes will be announced in the [Changelog](/releases/).
```

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,641 @@
...
+h1: Query Recipes
+meta_desc: Curated Context API query recipes for provider-upgrade impact, cross-stack dependencies, and infrastructure coverage.
+menu:
+  insights:
+    parent: insights-context-api
+    weight: 10
+pulumi_cloud_feature: context-api
+---
+
+These recipes translate common infrastructure questions into [Context API](/docs/insights/context-api/) selectors. Each one answers a specific question a platform team runs into repeatedly, such as scoping a provider upgrade or finding resources that live outside IaC.
```
cnunciato: Rings markety. First sentence is enough.

```suggestion
These recipes translate common infrastructure questions into [Context API](/docs/insights/context-api/) selectors. 
```

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,641 @@
...
+
+{{% notes type="info" %}}
+The schema returned by `pulumi api GetGraphSchema -F orgName=<org>` is the canonical reference for the query vocabulary and caps, and it can change between schema versions. Fetch it before composing new queries rather than relying on the examples below verbatim.
+{{% /notes %}}
+
+The selectors below use an example organization (`acme-corp`) whose projects include `payments`, `invoicing`, `reporting`, and `audit`, each with `prod` and `staging` stacks. Substitute your own organization, project, and stack names. Each recipe includes the response the example organization gets back, collapsed under the query.
+
+## What does an outdated provider still manage?
```
cnunciato: There's no explicit notion of "outdated" in the query that I can tell, and it reads a bit oddly. Suggest instead:

```suggestion
## Which resources are managed by older versions of a provider?
```

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,641 @@
...
+    "schemaVersion": "2026-08-25"
+  }
+}
+```
+
+{{< /details >}}
+
+Only resources discovered by cloud scanning carry an `account` field, so this anchor isolates them from Pulumi-declared ones, and grouping by `account` counts them per account. Compare those counts against your own stack resource counts to gauge how much of each account you've brought under management.
+
+Two caveats. Buckets come back in account-name order, not by count, so the list isn't a ranking. And if `meta.resultMode` is `truncated`, don't act on the totals at all: either whole buckets were dropped at a result cap, leaving the returned counts exact but the account list short, or the search backend answered partially, in which case a returned count can be low too. The response doesn't distinguish the two, so narrow the anchor and re-run.
```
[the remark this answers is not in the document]
lblackstone: Good catch, and the fix went the other way: the prose was right and the example was wrong.

The service sorts aggregation buckets ascending by the `groupBy` value, in two places — the backend issues a composite aggregation with no order set, so the composite default (`asc`) applies, and the response layer re-sorts explicitly before serializing. Nothing sorts by count.

The example's order came from a golden fixture, which is the trap. The golden comparison re-sorts buckets by a `fmt.Sprint` over a Go map, where `/` (0x2F) precedes `]` (0x5D); the wire sort compares the JSON array rendering of the key values, where the closing `"` (0x22) beats `/`. The two agree for ordinary values and disagree exactly when one key is a prefix of another — which is this fixture. So the golden corpus is structurally unable to catch a regression here, and copying its order into the docs inherited an artifact rather than the behavior.

The buckets now read `acme-legacy-aws` (1), `acme-prod-aws` (4), `acme-prod-aws/payments-legacy-cfn` (2): true ascending key order, and non-monotonic in `found`, so it still demonstrates that the list isn't a ranking.

Two things fell out of checking it. `account` needed defining — it holds the Pulumi Insights account name, not the cloud account's identifier, and a value with a slash is a discovered child account whose resources the parent's bucket excludes. And the recipe's stated purpose didn't survive: an imported resource keeps its scanned node, so this count can't measure IaC coverage. It now claims only a census of what the scan found, and points at `managed` for the coverage question.

🤖 AI-assisted (Claude Code)

content/docs/insights/context-api/_index.md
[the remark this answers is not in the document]
lblackstone: Added, and auditing the class turned up three more sites.

Confirmed against the implementation rather than inferred: aggregation queries run through the same permission-filtered search as everything else, but the probe that detects hidden neighbors is wired only into the traversal executors, never the aggregate path — deliberately, so anchor enumeration can't be used as an existence oracle. So the counts really are caller-relative and really are unsignalled. The recipe now says a restricted caller gets quietly lower totals, or loses a bucket entirely, with both responses reading `exact`, and that a total is a floor unless the caller sees the whole organization. The landing page's `meta.visibility` bullet carries the same clause.

The rule behind this is the landing page's own — both fidelity fields must be clean before answering a question about absence — and three other recipes broke it:

- **What does an outdated provider still manage?** cited only `meta.resultMode` for a completeness claim.
- **Which stacks consume this stack's outputs?** made an absence claim citing only `meta.resultMode`, with the permission half hand-waved as "at least none the caller can see."
- **Is this resource managed by a Pulumi stack?** stated an if-and-only-if with no fidelity caveat at all.

All four now name both signals, or name the one that isn't there.

One correction in the opposite direction while auditing the same class: the depth caveat on the decommission recipe was vacuous and has been removed. Any transitive `consumes_outputs_of` chain necessarily has a direct consumer at depth 1, so `target.absent` over depth 1–3 is equivalent to depth 1–1 and raising `depth.max` cannot reveal a hidden consumer. It does apply to the recipe that enumerates consumers, where the returned list genuinely truncates, so it moved there.

🤖 AI-assisted (Claude Code)

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,641 @@
...
+
+{{% notes type="info" %}}
+The schema returned by `pulumi api GetGraphSchema -F orgName=<org>` is the canonical reference for the query vocabulary and caps, and it can change between schema versions. Fetch it before composing new queries rather than relying on the examples below verbatim.
+{{% /notes %}}
+
+The selectors below use an example organization (`acme-corp`) whose projects include `payments`, `invoicing`, `reporting`, and `audit`, each with `prod` and `staging` stacks. Substitute your own organization, project, and stack names. Each recipe includes the response the example organization gets back, collapsed under the query.
+
+## What does an outdated provider still manage?
+
+Before upgrading a provider, which resources does the old version still manage?
```
cnunciato: These read better as statements that expand on the question. (Sounds repetitive otherwise.)

```suggestion
Before upgrading a provider, you can ask which resources would be affected:
```

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,641 @@
...
+    "visibility": "complete"
+  }
+}
+```
+
+{{< /details >}}
+
+The anchor matches provider resources below version 7.0.0 (versions compare as semantic versions, not as strings, so `10.0.0` counts as newer than `7.0.0`); walking `provided_by` inward from there returns everything each of those providers manages. Check `meta.resultMode` before treating the returned list as complete: `truncated` means more managed resources may exist that this response didn't reach, so narrow the query (a tighter `scope`, or a more selective anchor) and re-run.
+
+## What's the full blast radius of upgrading this provider?
```
cnunciato:
```suggestion
## What is the impact of upgrading this provider?
```

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,641 @@
...
+
+If I upgrade this provider, which stacks could be affected, and by what chain of dependencies?
```
cnunciato:
```suggestion
In addition to resources, you can also ask which stacks would be affected, along with the chain of dependencies:
```

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,641 @@
...
+
+A blast-radius answer is only useful if it's complete, and three separate things can leave it short:
+
+- **The engine clipped it.** Treat `meta.resultMode: "truncated"` or `meta.visibility: "trimmed"` as disqualifying: narrow the scope and re-run, or re-run as a caller with broader access.
+- **Your own depths cut the walk short.** A dependency chain longer than the four hops this selector asks for is left out *without* setting either flag, since no engine cap was exceeded. Raise `depth.max` if your graph is deeper than the example's.
+- **Heuristic dependencies weren't asked for.** The `dependents` step walks `reference` only, so it covers what Pulumi recorded and stays silent about inferred edges. Add `inferred_reference` beside it to widen the search, and treat what that adds as leads to verify rather than confirmed impact.
```
cnunciato: Scrubbing more LLMisms. Also, what does "final frontier" mean here? It sticks out a bit. Is there another more colloquial way to put this?

```suggestion
This chains four hops: provider, to what it manages, to what depends on those resources, to the stacks that contain them, to every stack consuming those stacks' outputs. `return.select` names only the final frontier, so the answer is the affected stacks; `return.paths: true` then returns one evidence trail per stack, showing the chain of resources that connects it back to the provider. Name the intermediate aliases in `return.select` too if you want the affected resources themselves, not just the stack list and its evidence.

Answers like these should be checked for completeness. Pay attention to the following fields and adjust your query as necessary:

- **The result was trimmed or truncated.** Treat `meta.resultMode: "truncated"` or `meta.visibility: "trimmed"` as disqualifying: narrow the scope and re-run, or re-run as a caller with broader access.
- **The query exceeded the specified depth.** A dependency chain longer than the four hops this selector asks for is left out without setting either flag, since no engine cap was exceeded. Raise `depth.max` if your graph is deeper than the example's.
- **Heuristic dependencies weren't requested.** The `dependents` step walks `reference` only, so it covers what Pulumi recorded and omits inferred edges. Add `inferred_reference` beside it to widen the search, and treat what that adds as leads to verify rather than confirmed impact.
```

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,641 @@
...
+
+## Which stacks consume this stack's outputs?
+
+If I change this stack, which other stacks read its outputs and could break?
```
cnunciato:
```suggestion
You may also want to determine how a change to the outputs of a given stack might affect its downstream consumers:
```

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,641 @@
...
+}
+```
+
+{{< /details >}}
+
+The anchor is the stack, or stacks, you're changing; walking `consumes_outputs_of` inbound up to three hops catches both direct consumers and stacks that consume a consumer. `return.paths` distinguishes the two: a one-edge path is a direct dependency, a longer one is transitive. An empty `downstream` frontier with `meta.resultMode: "exact"` means no stack currently consumes this one's outputs, at least none the caller can see.
+
+## Which stacks have no downstream consumers?
+
+Which stacks are decommissioning candidates because no other stack reads their outputs?
```
cnunciato:
```suggestion
You can also query for stacks that have no downstream consumers:
```

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,641 @@
...
+    "resultMode": "exact",
+    "schemaVersion": "2026-08-25",
+    "visibility": "complete"
+  }
+}
+```
+
+{{< /details >}}
+
+`target.absent: true` inverts the traversal, returning the stacks from which the walk reaches no consumer at all, the candidates for cleanup.
```
cnunciato:
```suggestion
`target.absent: true` inverts the traversal, returning the stacks from which the walk reaches no consumer at all.
```

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,641 @@
...
+
+Read the result as a shortlist, never as permission to delete. It answers one narrow question — does another stack read this stack's outputs — and three things can make a "no" wrong:
+
+- **It's the only question asked.** A stack whose resources are depended on some other way, or used by something outside the graph entirely, still appears here.
+- **Absence is relative to the caller.** A consumer the caller lacks permission to see won't disqualify a stack, so check `meta.resultMode` is `exact` and `meta.visibility` is `complete`, and re-run with broader access if not.
+- **The index trails reality.** Queries answer from the search index, so a stack reference added moments ago may not show yet. Confirm against current stack state before decommissioning.
```
cnunciato: I think we can delete all of this. The query isn't asking "which stacks can I delete", it's asking "which stacks don't have consumers". So the extra context doesn't really apply. 

I'll leave it to you to decide whether it's worth mentioning `resultMode` and `visibility` again, but I think we've pretty much established them at this point.

```suggestion
```

lblackstone:
@pulumi-eon /review

content/docs/insights/context-api/recipes.md
cnunciato: I suggest we delete this too. It doesn't seem to be telling me anything I can't  figure out either by looking at the data or recalling the meaning of `resultMode`, which has already been explained.

```suggestion
```

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,641 @@
...
+
+## How much infrastructure did a cloud scan find in each account?
+
+How much of what Pulumi knows about each cloud account came from scanning it rather than from deploying it?
```
cnunciato: The phrasing of this question confuses me. Based on the explanation below, I _think_ this what it's trying to say: 

```suggestion
## How much of my cloud account has been brought under management?

You can use the Context API to tell you which resources were scanned and imported versus deployed with a stack:
```

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,641 @@
...
+    "resultMode": "exact",
+    "schemaVersion": "2026-08-25",
+    "visibility": "complete"
+  }
+}
+```
+
+{{< /details >}}
+
+`in_stack` only exists for resources a Pulumi program deployed, so walking it outward from a named resource returns a stack if, and only if, that resource is Pulumi-managed. A resource discovered by cloud scanning produces an empty `stacks` frontier every time, because a scan deploys nothing and creates no such edge. To sort a whole population rather than one resource, anchor on `account: { "op": "present" }` instead: that selects exactly the scan-discovered set, every member of which comes back with an empty `stacks` frontier.
```
cnunciato: "Every time" feels odd (and LLM-y), since it's presumably a determininstic API. Suggest removing.

```suggestion
`in_stack` only exists for resources a Pulumi program deployed, so walking it outward from a named resource returns a stack if, and only if, that resource is Pulumi-managed. A resource discovered by cloud scanning produces an empty `stacks` frontier, because a scan deploys nothing and creates no such edge. To sort a whole population rather than one resource, anchor on `account: { "op": "present" }` instead: that selects exactly the scan-discovered set, every member of which comes back with an empty `stacks` frontier.
```

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,641 @@
...
+
+{{< /details >}}
+
+Only resources discovered by cloud scanning carry an `account` field, so this anchor isolates them from Pulumi-declared ones, and grouping by `account` counts them per account. Compare those counts against your own stack resource counts to gauge how much of each account you've brought under management.
+
+Two caveats. Buckets come back in account-name order, not by count, so the list isn't a ranking. And if `meta.resultMode` is `truncated`, don't act on the totals at all: either whole buckets were dropped at a result cap, leaving the returned counts exact but the account list short, or the search backend answered partially, in which case a returned count can be low too. The response doesn't distinguish the two, so narrow the anchor and re-run.
+
+## Is this resource managed by a Pulumi stack, or was it only discovered?
+
+How can I tell whether a given resource is under Pulumi management or was found by a cloud scan?
```
cnunciato:
```suggestion
## Is this resource managed by a stack, or only discovered?

Similarly, you can tell whether a given resource is under management or was found by a cloud scan:
```

content/releases/changelog/2026-08-26-context-api.md
```diff
@@ -0,0 +1,16 @@
...
+
+[Pulumi Neo](/product/neo/) uses the Context API out of the box. Any other agent, including Claude Code, Cursor, and Codex, can onboard itself; the API's schema endpoint serves a Markdown primer covering the query grammar and worked examples (the `pulumi api` CLI fetches it by default; over raw HTTP, send `Accept: text/markdown`), so an agent learns the whole system in one request and stays current as the API evolves.
+
+Read the [Context API documentation](/docs/insights/context-api/) to get started, browse the [query recipes](/docs/insights/context-api/recipes/) for common questions, or read the [announcement blog post](/blog/pulumi-context-api/) for the full story.
```
cnunciato:
```suggestion
[Pulumi Neo](/product/neo/) uses the Context API out of the box. Any other agent, including Claude Code, Codex, Cursor and others can use it by fetching a Markdown document that tells the agent how to use it.

Check out the [announcement blog post](/blog/pulumi-context-api/) for the full story, or read the [Context API documentation](/docs/insights/context-api/) to get started and browse [common query recipes](/docs/insights/context-api/recipes/).
```

content/docs/insights/context-api/_index.md
```diff
@@ -0,0 +1,128 @@
...
+
+The two reference types differ in how certain they are. A `reference` edge is a dependency Pulumi recorded as fact. An `inferred_reference` edge is a heuristic lead that may be wrong, because an identifier that doubles as an ordinary value can attract edges that aren't dependencies. Traverse `reference` when you'll act on the answer, and add `inferred_reference` when you want leads worth checking.
+
+A query names a starting point (`anchor`), walks relationships from there (`traverse`), and returns the connected nodes and edges (`return`). Each traversal step takes an `alias`, and a node's `frontier` lists the aliases it was reached through; an empty `frontier` marks a node that's present only as a waypoint on a returned path. That's how you tell an answer from the evidence for it. An optional `scope` narrows which stacks a query draws its anchors from; it bounds where a query starts, never where the traversal can reach, so a scoped query routinely returns nodes from outside its own scope. The wire format is JSON, not a query language. The full grammar, including field predicates, aggregation, and pagination, lives in the schema the API serves itself, described below.
```
cnunciato: I feel like this block of text would be better represented and more scannable as a bullet list. Until now (late in my review, second pass), I completely missed what "frontier" means, as you'll see from some of my other comments. It's essentially a glossary of terms, right? If so, I'd suggest a reformat.

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,655 @@
...
+
+These recipes translate common infrastructure questions into [Context API](/docs/insights/context-api/) selectors. Each one answers a specific question a platform team runs into repeatedly, such as scoping a provider upgrade or finding resources that live outside IaC.
+
+{{% notes type="info" %}}
+The schema returned by `pulumi api GetGraphSchema -F orgName=<org>` is the canonical reference for the query vocabulary and caps, and it can change between schema versions. Fetch it before composing new queries rather than relying on the examples below verbatim.
+{{% /notes %}}
+
+The selectors below use an example organization (`acme-corp`) whose projects include `payments`, `invoicing`, `reporting`, and `audit`, each with `prod` and `staging` stacks, and whose cloud scans run under the accounts `acme-prod-aws` and `acme-legacy-aws`. Substitute your own organization, project, and stack names. Each recipe includes the response the example organization gets back, collapsed under the query.
+
+## What does an outdated provider still manage?
```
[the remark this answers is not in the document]
lblackstone: Fixed in 15fd1222c3. Confirmed the mechanism before changing anything: `layouts/partials/schema/graph-builder.html` gates `$qualifiesForFaqSchema` on `.IsPage` plus `.Type == "docs"` with no opt-in flag, and in `faq-entity.html` the answer-accumulation branch (`else if $inQuestion`) carries no `$inCodeFence` guard, unlike the two heading-detection branches above it. So the JSON selectors and example responses really were being appended verbatim to `acceptedAnswer.text`.

All six H2s are now declarative sentence case:

- Find what an outdated provider still manages
- Map the blast radius of a provider upgrade
- Find the consumers of a stack's outputs
- Find stacks with no downstream consumers
- Count scan-discovered resources by account
- Check whether a resource is managed or discovered

Two things that made this safe. Each recipe already restated its question as the body line directly beneath the heading, so the question phrasing survives on the page for readers and for search; it just stops being a heading. And these pages are new in this PR, so no anchors were ever published: a repo-wide search for links to the old slugs and for `context-api/recipes/#` returns nothing, and no other file quotes a recipe by its question wording.

Swept the rest of the diff for the same shape rather than just the site you flagged. The blog post and changelog entry have no `?`-terminated headings, and neither would qualify anyway: the changelog cascades `type: changelog`, and the two `_index.md` files are section pages, so `.IsPage` is false for them.

🤖 AI-assisted (Claude Code)

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,655 @@
...
+
+`return.select` names only the final frontier, so the answer is the affected stacks; `return.paths: true` then returns one evidence trail per stack, showing the chain of resources that connects it back to the provider. The resources those trails pass through come back too, marked with an empty `frontier` to set them apart from the answer, and `pageInfo.resultCount` counts them too, so it runs ahead of the number of stacks you selected. Name the intermediate aliases in `return.select` too if you want the affected resources themselves, not just the stack list and its evidence.
+
+A blast-radius answer is only useful if it's complete, and three separate things can leave it short:
```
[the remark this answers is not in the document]
lblackstone: Confirmed and fixed in 15fd1222c3, and the audit turned up more sites than the two flagged.

Ground truth: `paging.go:372-373` mints `continuationToken` only when `end < len(keys)`, so it's present exactly when pages remain and omitted, not null, on the last one. `COOKBOOK.md:1081-1083` demonstrates truncation and paging firing together, and `limits_test.go` has a subtest asserting both signals in one response. Defaults are `defaultPageSize = 200` / `maxPageSize = 1000` (`types.go:76-78`), clamped by `min()`, so the recipe's `pageSize: 500` is honored as written.

Rather than add one bullet, I inventoried every completeness and absence claim in the bundle by reading the four files end to end. Eleven sites, not two. The overview now states the mechanics once (token presence, cumulative merge, per-page fidelity, and that a drain is not a snapshot), and each recipe making such a claim says so at its own site. Aggregations page their bucket stream through the same cursor (`cursorSectionBuckets`, `bucketKey`), so the coverage rollup needed it too.

Two neighbouring claims turned out wrong rather than merely incomplete: `pageInfo.resultCount` is `len(nodes)` over the post-window slice (`assemble.go:126`), so it counts the page and not the answer; and the `account: present` walk needed a full drain before "every member comes back" holds.

One deliberate divergence. You also asked for this under the decommissioning recipe's "three things can make a 'no' wrong" list, and I left that at three. `target.absent` runs against a scratch graph and inverts during traversal (`service.go:388`, `finishAbsentStep`), well before `assemble` cuts the page window, so an unread page can shorten that shortlist but cannot make a returned "no" wrong. That list is specifically about what makes a returned "no" wrong, so the paging caveat would be misfiled there. The shortlist's own completeness is covered one paragraph earlier, where the list is produced.

🤖 AI-assisted (Claude Code)

lblackstone:
@pulumi-eon /review

lblackstone:
@claude #update-review

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,657 @@
...
+
+Read the result as a shortlist, never as permission to delete. It answers one narrow question — does another stack read this stack's outputs — and three things can make a "no" wrong:
+
+- **It's the only question asked.** A stack whose resources are depended on some other way, or used by something outside the graph entirely, still appears here.
+- **Absence is relative to the caller.** A consumer the caller lacks permission to see won't disqualify a stack, so check `meta.resultMode` is `exact` and `meta.visibility` is `complete`, and re-run with broader access if not.
+- **The index trails reality.** Queries answer from the search index, so a stack reference added moments ago may not show yet. Confirm against current stack state before decommissioning.
+
+## Count scan-discovered resources by account
+
+How much has cloud scanning turned up in each of our accounts?
```
cnunciato: This one seems to have snuck in during my review, so I didn't have time to revise it. Therefore leaving it for now (late here), but would be good to revise it similarly to the others here. (Also happy to do this either in the morning or as a follow-up.)

content/docs/insights/context-api/recipes.md
```diff
@@ -0,0 +1,657 @@
...
+- **The buckets page like nodes do.** An aggregation slices that same ordered stream into pages, so a long account list arrives over several responses and a page boundary can land between a parent and its children. `pageInfo.resultCount` counts the buckets on the page, not the accounts in your organization, so drain before summing.
+- **`truncated` disqualifies the totals.** Either whole buckets were dropped at a result cap, leaving the returned counts exact but the account list short, or the search backend answered partially, in which case a returned count can be low too. The response doesn't distinguish the two, so narrow the anchor and re-run.
+- **Nothing here reports permission trimming.** An aggregation carries `meta.resultMode` alone; `meta.visibility` is absent by design. Counts are always relative to the caller, so two callers with different permissions get different totals for this query and both come back `exact`. Read a total as a floor unless the caller can see the whole organization.
+
+## Check whether a resource is managed or discovered
+
+How can I tell whether a given resource is under Pulumi management or was found by a cloud scan?
```
cnunciato: Same with this one.

cnunciato (commented):
Excellent work. I've left a lot here, but hopefully it's all useful stuff.

Couple bits of meta-feedback:

* Super-respectfully (💜), a lot of this feels pretty LLM-generated to me. I've done a bunch to try to scrub some of that away, but in general, we shouldn't be submitting LLM-generated content meant for human consumption without a humanization/voice pass. The brand guidelines [try to make this clear](https://brand.pulumi.com/guidelines/generative-ai/#never), but it's up to us to apply them, since there isn't really a reliable way to detect LLM-generated copy with LLMs. 

* After reading all this, I still can't say I feel like I fully understand the altitude at which I'm supposed to understand and use this stuff. If the goal (as I think it is) is to have folks interact with it through their agents, then there may be too much detail here than is necessary — or at least that belongs in a doc outside of the Reference section. We can come back to this after we launch, but I suspect a cleaner distinction between _usage_ docs and _reference_ docs would make this feel much more digestible. (I'd also love to hear more about whether we're planning to pull the context stuff into a skill, as that'd seem like the ideal way to use this — but I'm sure there's more to it than that!)

lblackstone:
> Excellent work. I've left a lot here, but hopefully it's all useful stuff.
> 
> Couple bits of meta-feedback:
> 
>     * Super-respectfully (💜), a lot of this feels pretty LLM-generated to me. I've done a bunch to try to scrub some of that away, but in general, we shouldn't be submitting LLM-generated content meant for human consumption without a humanization/voice pass. The brand guidelines [try to make this clear](https://brand.pulumi.com/guidelines/generative-ai/#never), but it's up to us to apply them, since there isn't really a reliable way to detect LLM-generated copy with LLMs.
> 
>     * After reading all this, I still can't say I feel like I fully understand the altitude at which I'm supposed to understand and use this stuff. If the goal (as I think it is) is to have folks interact with it through their agents, then there may be too much detail here than is necessary — or at least that belongs in a doc outside of the Reference section. We can come back to this after we launch, but I suspect a cleaner distinction between _usage_ docs and _reference_ docs would make this feel much more digestible. (I'd also love to hear more about whether we're planning to pull the context stuff into a skill, as that'd seem like the ideal way to use this — but I'm sure there's more to it than that!)

Thanks, and I agree. The blog is the main content that's intended for human consumption, and I've iterated on that quite a bit. I'm trying to figure out what other public docs make sense, and I don't think it's quite there yet. As you mention, a lot of this content is geared toward agents. I was trying to give a sense of the types of questions that could be answered, but I wouldn't expect most people to understand all of the API nuances. It occurs to me that a conceptual explanation would be a more useful reference doc. Possibly a lot of the recipe content might be better served directly through the API.

What do you think about dropping most of the doc content except for the blog post and then we can follow up with some better human-focused reference docs?

content/blog/pulumi-context-api/index.md
lblackstone: Worked, as in, with an example response corresponding to the query. I'll reword.

content/blog/pulumi-context-api/index.md
lblackstone: I'll elaborate further. The query syntax is documented through the schema endpoint. It's complex, so I thought that might be beyond the scope of an announcement post. It is based on an open standard, but I didn't want to muddy the waters trying to explain the connection since we don't accept that syntax in the v1 API.

> Also, how does one submit a query?

Yes, POST that JSON snippet to the query endpoint. I'll make that clearer.

lblackstone:
I've reworked the blog post and changelog with your input and removed the other docs for now. Thanks again for your thorough review, and for bearing with me as I try to explain a complex subject to a broad audience!

lblackstone:
@claude #update-review

Note that I've removed most of the previous docs content and squashed the commits.

lblackstone:
@claude #update-review

cnunciato:
>What do you think about dropping most of the doc content except for the blog post and then we can follow up with some better human-focused reference docs?

That sounds great, and I'm happy to hop in and help shape the docs on this too!

cnunciato (approved):
🚀🚀🚀! I added a bit to pin the post as the featured one to keep it in place for a few days as others go out.

lblackstone:
@pulumi-eon merge changes from master into this branch

PR #21060, opened 2026-08-21
Cut PR build time: stop deleting the Hugo image cache, drop GOGC=3

CamSoper:
### Proposed changes

Successful `Pull Request` runs have been landing at 14–17 minutes. Profiling [run 32500502599](https://github.com/pulumi/docs/actions/runs/32500502599) (17m12s): "Build and deploy" is 15m45s of it, and the Hugo build alone is **9m05s**.

Two causes, plus two missing guardrails. Three runs on this branch isolate each fix, because the middle one ran cold *with* the GC change:

| | Total | Hugo | `blog/feature-image.html` |
|---|---|---|---|
| Cold cache + `GOGC=3` (before) | 17m12s | 9m05s | 994ms/call |
| Cold cache + `GOMEMLIMIT` ([15484](https://github.com/pulumi/docs/actions/runs/32531633733)) | 12m22s | 4m56s | 561ms/call |
| Warm cache + `GOMEMLIMIT` ([15480](https://github.com/pulumi/docs/actions/runs/32528763915)) | **6m54s** | **48.8s** | **458µs/call** |

The GC change is worth ~4m50s, the cache fix ~5m28s, and they're close to independent. `make ensure` also dropped 1m53s → 1m05s from the preserved yarn cache.

#### 1. The Hugo image cache was deleted 0.2s after it was restored

`make ci_pull_request` → `make ensure`, `ensure` depends on `clean`, and `scripts/clean.sh` ran `rm -rf resources` — the exact directory `actions/cache` had just restored:

```
15:59:59.20  Cache restored from key: hugo-resources-342b83e3... (89 MB)
15:59:59.37  ./scripts/clean.sh
```

Hugo then re-encoded every blog feature image from scratch on every run. The cache still restored *and* saved cleanly each time, so the waste never showed up in the job log — only in the template metrics, where `blog/feature-image.html` averaged **994ms per call** across 718 calls.

`clean.sh` now skips `yarn cache clean`, `rm -rf resources`, and `hugo mod clean` when `CI` is set. The rest are no-ops on a fresh clone, and local behavior is unchanged. `hugo mod clean` tested clean on the pinned 0.157.0 (exits 0, preserves every `resources/_gen` entry, against this repo's real `module:` config) but is inside the guard anyway — it's a no-op that early on a fresh runner, so CI loses nothing and the question can't come back.

Ruled out along the way, so they don't muddy future diagnosis: the per-run preview `baseURL` does **not** bust the cache, nor do fresh-clone mtimes. And `Processed images │ 2755` counts `.Process` *calls*, not encodes — it reads 2755 both before and after this change, so it is not a cache-health signal.

#### 2. `GOGC=3`

`build-site.sh` ran Hugo under `GOGC=3`, collecting once the heap grows 3% over live heap. Measured on a 6,549-page Hugo 0.157.0 harness with a warm cache:

| Config | wall | peak RSS |
|---|---|---|
| `GOGC=3` (before) | 9.4–10.0s | 143 MB |
| `GOGC=100` | 5.6–5.8s | 223 MB |
| `GOGC=100` + `GOMEMLIMIT` | 6.1s | 224 MB |

**1.7–3× slower** for ~35% peak RSS saved, and the penalty grows with heap size. `GOMEMLIMIT=12GiB` expresses the actual intent — don't exhaust the 16GB runner — without paying a full GC behind every allocation. Set **only under CI**, since `build-site.sh` is also the local build path (`make build`, `scripts/laptop-deploy.sh`) where a 12GiB ceiling would be no ceiling at all on a 16GB laptop; an explicit `GOMEMLIMIT` always wins.

`--templateMetrics` was also measured and costs nothing detectable, so it stays.

#### 3 & 4. Two guardrails this workflow was missing

- **Concurrency group with `cancel-in-progress`.** Pushing three commits in a row ran three full builds to completion; branches did exactly that four times in one day. Already demonstrated on this PR — [run 15483](https://github.com/pulumi/docs/actions/runs/32530978253) was cancelled when the next commit landed. Preview buckets are keyed by PR number *and* head sha, so a cancelled run can only orphan its own — `pr-closed.yml` deletes every `*-pr-<num>-*` bucket on close and `bucket-cleanup-testing.yml` sweeps the `www-testing-*` previews daily. Both cache steps run with `save-always: false`, so a cancelled run cannot save a partial cache.
- **`ci-build-duration-alert.sh`,** which `build-and-deploy.yml` already has and PR builds did not. That is how these drifted to ~17 minutes without anyone being told; the script's own header notes the build grew ~60% over three weeks in July the same way. Threshold 15m — it reported `Build and deploy took 10m (655s) ... threshold is 15m` on the cold run, quiet as intended, but would have caught the old 15m45s.

#### 5 & 6. Cache hardening (added after review)

Making the cache actually live activates two latent issues that didn't matter while it was being nuked. Both verified, both fixed here:

- **The Hugo version is now in the cache key.** Hugo 0.156.0 reuses 0.157.0's cached output byte-for-byte, so after an upgrade images stay encoded by whatever version first produced them. Exactly four workflows cache `path: resources` — `pull-request.yml`, `build-and-deploy.yml`, `testing-build-and-deploy.yml`, `pulumi-cli-docs.yml` — and all four now key and restore on `hugo-resources-0.157.0-` and pin `hugo-version: '0.157.0'`. They must move together or they stop sharing a cache; the comment in each names the full set. Kept as a literal rather than interpolating the pin: a failed interpolation there would silently install the wrong Hugo, a worse failure than key drift.
- **`hugo --gc`** prunes entries the build no longer references, which is what bounds growth — nothing else reclaims superseded ones. Applied to every non-`preview` build under CI: the two deploys plus any CI job running `make build` (`pulumi-cli-docs.yml` does, and it's a full-site build, so it prunes against the same reference set). PR preview builds are the deliberate exclusion — they share the namespace, and a narrower view could drop entries the others need.

#### Blast radius

`build-and-deploy.yml` restores the same `hugo-resources` cache and runs the same `clean` via `make ci_push`, so **master deploys were hitting both problems too** and get the same fix.

⚠️ **Expect one slow build per workflow after this merges.** Nothing has written a `hugo-resources-0.157.0-*` entry yet, so the first run of each of the four repopulates from cold. That's the 12m22s column above, not a regression; the run after returns to ~7 minutes.

#### New profile

Hugo is no longer the bottleneck (48.8s, 12% of the run). What's left, roughly evenly: runner setup 1m01s, `make ensure` 1m05s, `generate-docs-content.js` 1m01s, then S3 sync 20s / Cypress 23s / search index 21s / `pulumi preview` 35s / redirects 18s. There's no single large win left — going below ~5 minutes means chipping at several things.

N/A

### Related issues (optional)

N/A

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -111,9 +111,14 @@ jobs:
         uses: actions/cache@v6
         with:
           path: resources
-          key: hugo-resources-${{ github.sha }}
+          # The version prefix must match the hugo-version pinned above, and be
+          # identical in pull-request.yml and build-and-deploy.yml or PRs stop
+          # inheriting master's cache. Hugo does not put its own version in the
+          # cache key, so without this a Hugo upgrade keeps serving images
+          # encoded by the previous version indefinitely. Bump both on upgrade.
+          key: hugo-resources-0.157.0-${{ github.sha }}
           restore-keys: |
-            hugo-resources-
+            hugo-resources-0.157.0-
 
       - name: Record build start time
         run: |
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -112,10 +112,12 @@ jobs:
         with:
           path: resources
           # The version prefix must match the hugo-version pinned above, and be
-          # identical in pull-request.yml and build-and-deploy.yml or PRs stop
-          # inheriting master's cache. Hugo does not put its own version in the
-          # cache key, so without this a Hugo upgrade keeps serving images
-          # encoded by the previous version indefinitely. Bump both on upgrade.
+          # identical in every workflow that caches resources/ -- pull-request.yml,
+          # build-and-deploy.yml, testing-build-and-deploy.yml, and
+          # pulumi-cli-docs.yml -- or they stop sharing a cache. Hugo does not put
+          # its own version in the cache key, so without this a Hugo upgrade keeps
+          # serving images encoded by the previous version indefinitely. Bump them
+          # together on upgrade.
           key: hugo-resources-0.157.0-${{ github.sha }}
           restore-keys: |
             hugo-resources-0.157.0-
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -103,6 +103,22 @@ jobs:
           restore-keys: |
             meta-images-
 
+      # Derive the cache key from the Hugo installed above rather than a literal,
+      # so revving hugo-version updates the key on its own. Fails loudly on an
+      # unparseable version rather than silently collapsing every build into a
+      # single unversioned namespace, which is the bug this exists to prevent.
+      - name: Resolve Hugo version for the cache key
+        id: hugo-version
+        run: |
+          hv="$(hugo version)"
+          # -n/p so a non-match yields empty rather than echoing the input back,
+          # and check before appending -extended so the guard can't see a value
+          # that is only the suffix.
+          ver="$(printf '%s' "$hv" | sed -nE 's/^hugo v([0-9]+\.[0-9]+\.[0-9]+).*/\1/p')"
+          [ -n "$ver" ] || { echo "could not parse hugo version from: $hv" >&2; exit 1; }
+          case "$hv" in *+extended*) ver="$ver-extended" ;; esac
+          echo "version=$ver" >> "$GITHUB_OUTPUT"
+
       # Persist Hugo's processed-image cache (resources/_gen). Blog templates
       # process feature images to WebP at several sizes via .Process; without
       # this cache every CI run re-encodes ~2,400 images from scratch (~4-5
@@ -111,16 +127,16 @@ jobs:
         uses: actions/cache@v6
         with:
           path: resources
-          # The version prefix must match the hugo-version pinned above, and be
-          # identical in every workflow that caches resources/ -- pull-request.yml,
-          # build-and-deploy.yml, testing-build-and-deploy.yml, and
-          # pulumi-cli-docs.yml -- or they stop sharing a cache. Hugo does not put
-          # its own version in the cache key, so without this a Hugo upgrade keeps
-          # serving images encoded by the previous version indefinitely. Bump them
-          # together on upgrade.
-          key: hugo-resources-0.157.0-${{ github.sha }}
+          # Keyed on the Hugo resolved above, not a literal, so the four
+          # workflows that share this cache (pull-request.yml,
+          # build-and-deploy.yml, testing-build-and-deploy.yml,
+          # pulumi-cli-docs.yml) stay in step through an upgrade without anyone
+          # remembering to bump them. Hugo does not put its version in its own
+          # cache entries, so without this an upgrade keeps serving images
+          # encoded by the previous version indefinitely.
+          key: hugo-resources-${{ steps.hugo-version.outputs.version }}-${{ github.sha }}
           restore-keys: |
-            hugo-resources-0.157.0-
+            hugo-resources-${{ steps.hugo-version.outputs.version }}-
 
       - name: Record build start time
         run: |
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -98,10 +98,12 @@ jobs:
         with:
           path: resources
           # The version prefix must match the hugo-version pinned above, and be
-          # identical in pull-request.yml and build-and-deploy.yml or PRs stop
-          # inheriting master's cache. Hugo does not put its own version in the
-          # cache key, so without this a Hugo upgrade keeps serving images
-          # encoded by the previous version indefinitely. Bump both on upgrade.
+          # identical in every workflow that caches resources/ -- pull-request.yml,
+          # build-and-deploy.yml, testing-build-and-deploy.yml, and
+          # pulumi-cli-docs.yml -- or they stop sharing a cache. Hugo does not put
+          # its own version in the cache key, so without this a Hugo upgrade keeps
+          # serving images encoded by the previous version indefinitely. Bump them
+          # together on upgrade.
           key: hugo-resources-0.157.0-${{ github.sha }}
           restore-keys: |
             hugo-resources-0.157.0-
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -89,6 +89,22 @@ jobs:
           restore-keys: |
             meta-images-
 
+      # Derive the cache key from the Hugo installed above rather than a literal,
+      # so revving hugo-version updates the key on its own. Fails loudly on an
+      # unparseable version rather than silently collapsing every build into a
+      # single unversioned namespace, which is the bug this exists to prevent.
+      - name: Resolve Hugo version for the cache key
+        id: hugo-version
+        run: |
+          hv="$(hugo version)"
+          # -n/p so a non-match yields empty rather than echoing the input back,
+          # and check before appending -extended so the guard can't see a value
+          # that is only the suffix.
+          ver="$(printf '%s' "$hv" | sed -nE 's/^hugo v([0-9]+\.[0-9]+\.[0-9]+).*/\1/p')"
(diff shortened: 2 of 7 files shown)


CamSoper:
@claude #update-review

All four findings addressed in dda8e001, plus the two follow-ups the PR body had listed as out of scope. I verified each finding against the repo before changing anything — all four were correct.

**🚨 [L13] `pull-request.yml` — wrong sweeper. Fixed.** Confirmed: `bucket-cleanup.yml` is `environment: production` (L15), `bucket-cleanup-testing.yml` is `environment: testing` (L16), and PR preview buckets are `www-testing-*` created by this workflow's `environment: testing` job. Comment now names `bucket-cleanup-testing.yml`.

**🚨 [L37] `build-site.sh` — stale `GOGC` row. Fixed.** `BUILD-AND-DEPLOY.md` L4010 now reads:

```
| **GOMEMLIMIT** | Go soft memory ceiling for Hugo (CI only) | `12GiB` | build-site.sh |
```

**⚠️ [L37] `GOMEMLIMIT` leaks into local builds. Fixed — good catch.** `make build` (`Makefile` L79) and `scripts/laptop-deploy.sh` L54 both call `build-site.sh`, so a ceiling sized for the CI runner was being applied to laptops. Now gated on `CI`, and written as `${GOMEMLIMIT:-12GiB}` so an explicit value always wins. Local builds fall back to Go's defaults.

**⚠️ [L20] Does `hugo mod clean` undo the guard? Tested, then made moot.** With a populated `resources/_gen` on the pinned 0.157.0 it exits 0 and preserves every entry — in a bare harness and against this repo's real `module:` config. Rather than rest on that, I moved it inside the guard as you suggested. It's a no-op on a fresh runner anyway (nothing is in the module cache that early), so CI loses nothing and the question stops being askable.

**Follow-ups now included:**

- **Hugo version stamped into the cache key.** Verified 0.156.0 reuses 0.157.0's cached output byte-for-byte, so without this an upgrade keeps serving images encoded by the old version indefinitely. Applied identically in both workflows — they must match or PRs stop inheriting master's cache. Kept as a literal rather than interpolating the `hugo-version` pin: a failed interpolation there would silently install the wrong Hugo, which is a much worse failure than key drift.
- **`hugo --gc`** to prune entries the build no longer references, since nothing else reclaims superseded ones. Deploy path only and CI only, so a PR build can't evict entries master still needs.

**⚠️ Expect one slow build.** The new cache key means the next run finds no `hugo-resources-0.157.0-*` and repopulates from cold — roughly a 17-minute build before the ~7-minute baseline returns. Not a regression.

Verification: I extracted the real conditional block from `build-site.sh` and ran it against a stub `hugo`. PR path gets no `--gc`; both deploy paths get it; local builds get neither `--gc` nor `GOMEMLIMIT`; a preset `GOMEMLIMIT` survives. `clean.sh` under `CI` preserves `resources/` and runs neither `yarn cache clean` nor `hugo mod clean`; with `CI` unset it still does all three.

On the `facts: MEDIUM` note — correct that the numbers aren't reproducible from CI, but the headline ones now are: [run 32528763915](https://github.com/pulumi/docs/actions/runs/32528763915) took 6m54s against a 17m12s baseline, with Hugo at 48,847 ms vs 545,004 ms and `blog/feature-image.html` at 458µs/call vs 994ms/call.

---

CamSoper:
@claude #update-review

Both findings from the second pass addressed in 9437eb14. The outstanding one was a real miss on my part, not a comment problem.

**🚨 [L105] Version stamp covered two of four workflows. Fixed properly.** Confirmed exactly four workflows cache `path: resources` — `pull-request.yml`, `build-and-deploy.yml`, `testing-build-and-deploy.yml` (L91/L93), `pulumi-cli-docs.yml` (L151/L153) — and all four pin `hugo-version: '0.157.0'`. All four now use `hugo-resources-0.157.0-`. You're right about both consequences: the two I missed kept the stale-encoder bug, and their saves were landing in a namespace the updated two no longer restore from, so that cache work was written and never read. I've used your suggested comment naming the full set, in all four, so the next upgrade bumps them together.

**⚠️ [L48] `--gc` comment claimed a narrower scope than the code has. Fixed.** Also correct — the guard is on the non-`preview` `else`, so it covers every non-preview CI build, and `pulumi-cli-docs.yml` L246 does run `make build`. I checked whether that's a bug before rewriting the comment, and I don't think it is: that workflow runs a *full* `make build` (its own cache comment at L145 says "for the full `make build` below"), so it references the same set of entries as a deploy and its prune is equivalent. That matters more now that all four share one namespace — a partial-site build pruning here would have been a genuine problem. The comment now describes the predicate the code tests:

```
# The guard below is on the non-preview branches, so this covers every
# non-preview build under CI: the two deploy workflows, plus any CI job that runs
# `make build` (pulumi-cli-docs.yml does). That is fine because all of them build
# the full site and so reference the same set of entries. PR preview builds are
# the ones deliberately excluded -- they share the same cache namespace, and a
# build pruning against a narrower view could drop entries the others still need.
```

Verification: re-ran the extracted conditional against a stub `hugo`, including the exact case you identified. `CI` set with `DEPLOYMENT_ENVIRONMENT` unset takes the production branch and gets `--gc`; PR preview gets none; local builds get neither `--gc` nor `GOMEMLIMIT`. All four workflows parse and their pins match the new prefix.

Thanks for the empty-array check on `"${hugo_gc[@]}"` — I'd relied on `nounset` being off without saying so, and it's worth having that written down for the macOS bash-3.2 case.

Still expect one cold build after this lands, now across all four workflows rather than two, since nothing has written a `hugo-resources-0.157.0-*` entry yet.

---

jkodroff (approved):
I like faster builds. LGTM.

CamSoper:
@claude #update-review

Pushed ff95472e, replacing the literal `hugo-resources-0.157.0-` prefix with one derived from the installed Hugo.

The literal moved the drift risk rather than removing it — revving Hugo meant remembering eight more places (four keys plus four restore-keys), and missing one silently reintroduces exactly the bug the stamp exists to prevent. `peaceiris/actions-hugo` already runs well before the cache step in all four workflows, so the key can come from the binary that will actually produce the cache:

```yaml
      - name: Resolve Hugo version for the cache key
        id: hugo-version
        run: |
          hv="$(hugo version)"
          ver="$(printf '%s' "$hv" | sed -nE 's/^hugo v([0-9]+\.[0-9]+\.[0-9]+).*/\1/p')"
          [ -n "$ver" ] || { echo "could not parse hugo version from: $hv" >&2; exit 1; }
          case "$hv" in *+extended*) ver="$ver-extended" ;; esac
          echo "version=$ver" >> "$GITHUB_OUTPUT"
```

Revving `hugo-version` now updates the key on its own, and the four workflows can't drift apart because none of them states the version. This deliberately does **not** interpolate into the `hugo-version:` pins — a failed interpolation there would silently install the wrong Hugo, which is worse than a stale encoder.

**One thing worth a close look, since it's the load-bearing part and I got it wrong first.** My initial guard was `sed -E` plus `[ -n "$ver" ]`. `sed -E` echoes its input back unchanged on a non-match, so `ver` was non-empty garbage and the guard passed — a build with an unrecognizable version banner would have keyed every workflow into one unversioned `hugo-resources--` namespace, precisely the failure being defended against. Fixed by using `sed -nE ... p` (non-match yields empty) and checking emptiness *before* appending `-extended`, so the guard can't pass on a value that is only the suffix.

Verification: step ordering asserted programmatically from the parsed YAML in all four (`hugo@3 < resolve@9 < cache@10` in pull-request.yml; 4<10<11 in build-and-deploy and testing-build-and-deploy; 9<13<14 in pulumi-cli-docs). The snippet was executed against real Hugo 0.157.0 and 0.156.0 binaries (both → `<semver>-extended`), a garbage banner, and empty output — the last two exit 1.

Also moved the pre-existing "Persist Hugo's processed-image cache" comment back down to the cache step, since the new step had come between it and what it describes.

**Not fixed here, and worth knowing:** this does not make revving Hugo a one-line change. `0.157.0` is still hardcoded in ~17 places — `mise.toml` twice (Linux and macOS pins) and 15 `hugo-version:` pins across workflows, several of which this PR never touched (`check-search-urls.yml`, `scheduled-test.yml`, the SDK docs workflows). Collapsing those onto a single source of truth is a separate piece of work; this change just stops the cache keys from being part of that problem.

---

PR #21061, opened 2026-08-21
Add /address-review: work every pre-merge review finding to a decision

CamSoper:
### Proposed changes

The review pipeline is good at finding things and has no way to make anyone look at them. `scrape-review-outcomes.py` carries an `ignored_low_confidence` column precisely because authors clear 🚨 and stop reading. This adds the author-side counterweight: a skill that works a PR's pinned review to zero, and the meta-file changes that make doing so the expectation rather than an option.

**New skill — `/address-review`** (`.claude/commands/address-review/`)

Watches for the pinned review to land after a PR goes ready for review, then walks **every** item with the author — 🚨 blockers, ⚠️ low-confidence, and the ✏️ inline style suggestions — one at a time, with a proposed fix for each. Fixes batch into a single push (keeping the shape `auto-refresh-gate.py` recognizes), disputes and fix-responses go out in one `@claude #update-review` mention, and the skill won't call a PR done while anything is undecided. It's also explicitly pushy: it offers to watch at ready-for-review without being asked, and says once — with a count — when a merge would leave items open. It never blocks a merge the user has decided on; "just merge it" becomes `accepted` with the user's reason on each open item, so the post-merge scrape reads the truth.

Two references: `watching.md` (event subscription where available, bounded polling otherwise, plus the short-circuit labels that mean no review is coming at all) and `dispositions.md` (the closed set — `fixed` / `refuted` / `deferred` / `accepted` / `not-applicable`, the last three note-required).

**New script — `review-worklist.py`** (`.claude/commands/docs-review/scripts/`)

The deterministic half. Parses the pinned body through `validate-pinned.py`'s helpers so the comment format keeps exactly one parser, merges in the inline one-click suggestions, assigns stable per-finding ids (`outstanding:L40-50`, `style:content/docs/a.md:L88`), and answers "is anything still open?" via `--require-clean`. Splits the ✏️ style block out of the ⚠️ section so advisory polish isn't miscounted as low-confidence findings, and never reports an unparseable body as an all-clear. State lives in a gitignored `.review-worklist-<PR>.json`, so a lost session resumes instead of restarting.

**Meta changes**

- `AGENTS.md` §PR Lifecycle — shipping a PR includes finishing its review; names the skill and the `--require-clean` check. (Also completes a sentence that had been left truncated mid-clause.)
- `CONTRIBUTING.md` — new §Working the review to zero: the five outcomes, and why a fix that never refreshed the review scrapes as ignored.
- `/shipit` — new mandatory Step 9 hands off to the review loop (8 → 9 steps, counters renumbered). PRs now open as **drafts**, matching CONTRIBUTING §Draft-first, which is what Step 9's ready-for-review offer assumes.
- `/pr-review` — Step 6 now discloses what an approval would be merging over; disclosure, not a gate, and the maintainer's own PRs route through `/address-review` first.
- `/docs-review` — one line pointing at the post-push half.

**Testing**

`./scripts/test-review-pipeline.sh` passes; the runner's `--self-test` discovery picks up `review-worklist.py` automatically (no change needed there). Also verified against the real `pr20079-pinned-review.md` fixture and confirmed `/address-review` is discovered by `docs-tools/scripts/scrape-metadata.py`. `make lint` is unaffected — nothing here is under `content/`, and prettier ignores `*.md`.

### Related issues (optional)

None.

---
_Generated by [Claude Code](https://claude.ai/code/session_017e8zEPDdXm3xuDfLWBfbP5)_


jkodroff (approved):

CamSoper:
@claude #update-review

All four findings fixed in 7cb09b9c — no disputes.

**[L127-144] 🚨** — you were right, and the docstring was writing a check the code didn't cash. `fetch_inline_suggestions` now returns `(suggestions, ok)`: empty stdout is failure (with `--jq '[...]'`, gh prints at least `[]` per page on success), and bytes-that-decode-to-nothing is failure too, tracked by a `decoded_any` flag rather than a silent `continue`. `ok` gates `clean` alongside `parse_confidence`, so a failed fetch can't present as a finished worklist. Two consequences worth noting beyond the suggested shape: the markdown render grows a warning banner when the fetch failed, and `_remaining_label` no longer says "0 item(s) still open" in that state — it says "every listed item decided — but the list is incomplete", since "0 open" was the exact false signal the finding is about. `--body-file` with no `--pr` stays `ok=True`: that caller opted out of the suggestions lane deliberately.

**[L188] ⚠️** — took the first option rather than rewording the docstring. New `_bullet_blocks()` groups each column-0 bullet with its continuation lines (stopping at the next bullet or a nested heading), so `text` now carries the fix prose and any suggested patch. One block per bullet, so counts still agree with `extract_bucket_bullets`. The style loop folds continuations into the bullet it's under while still tracking the `##### <path>` headings. Your own blocker bullet was the worked example — its `python` block is precisely what was being dropped.

**[L96] ⚠️** — dropped the duplication rather than the rows. `_split_low_confidence` reads its heading from `BUCKETS["low"][0]`, both the low and style loops read their `blocking`/`optional` from the table, and a new `PLAIN_BUCKETS` tuple names the two buckets that sit directly under their own H3. The table comment now explains why `low` and `style` share a heading.

**[L326] ⚠️** — fixed as suggested: "🐿️ Ship it! Your changes are pushed and the PR is open as a draft."

Self-test extended to cover all three code fixes — multi-line bullet capture, block boundaries, and a failed fetch blocking `clean`. `./scripts/test-review-pipeline.sh` green.

---

CamSoper:
@claude #update-review

Both round-2 findings fixed in 7f158321 — no disputes.

**[L153-164]** — right, and the comment I wrote (`# Bytes came back but nothing decoded`) described a stricter guarantee than the code delivered. `decoded_any` latched on the first page that parsed, so a decode failure on page 2 returned a short list with `ok=True`. Took the suggested shape: `failed_any` is set by both a `JSONDecodeError` and a valid-JSON-but-not-a-list document, and `ok` is now `decoded_any and not failed_any`. The short list is still returned alongside `ok=False` — a partial result is worth showing, it just isn't worth calling complete.

**[L44-48]** — took the second option, and it was more crowded than the finding suggested: that regex existed in **four** places, not two. `validate-pinned.py` defined it as a local inside both `extract_bucket_bullets` (L404) and `extract_finding_paragraphs` (L2391), `scrape-review-outcomes.py` had its own copy at L105 explicitly commented as "mirroring extract_bucket_bullets", and `_bullet_blocks` added a fourth. Hoisted it to a module-level `validate_pinned.FINDING_START_RE` with a comment naming its consumers; all four sites now reference that one object, and the self-test asserts identity (`is`, not equality) so a future local copy fails the suite rather than drifting quietly. Pointing `scrape-review-outcomes.py` at it is one line beyond what the finding asked for — I'd rather not create a shared constant and leave a known duplicate sitting next to it, but say the word and I'll revert that hunk.

The docstring now describes what actually happens instead of claiming a reuse that had stopped being true: same recognition rule, wider capture — `extract_bucket_bullets` returns a bullet's first line, which is right for counting and useless for working, so `_bullet_blocks` walks the same sections with the same rule and keeps the continuation lines.

Self-test adds four cases for the fetch's completeness signal (all-pages-decode, silent gh failure, error object, partial failure), stubbing `run` rather than gh. `./scripts/test-review-pipeline.sh` green and 184 pytest cases pass — worth noting since the last review couldn't execute the interpreter in its sandbox.

---

PR #21068, opened 2026-08-22
Fixed instruction providing two options where is just the same one

tehsis:
### Proposed changes

Removed duplicated option. I guess this is a remanet from when `esc` was its own command.


CamSoper:
@claude correct the outstanding issue and #update-review

CamSoper (approved):

PR #21110, opened 2026-08-25
Document organization-managed users and fix adjacent identity gaps

### LINKED ISSUE #21103: Administration: Document organization-managed users
## Summary

Pulumi Cloud has a distinct account type — a user whose account is **managed by an organization** (created through SAML JIT provisioning or SCIM). These accounts are subject to several restrictions that ordinary accounts are not. None of this is documented anywhere under `content/docs/administration/`, and searching the docs tree for "org-managed", "managed user", or equivalent returns nothing.

This matters because the restrictions are user-visible, produce hard errors, and are currently unresolvable from the documentation. A user who hits one has nothing to read.

## Behaviors that need documenting

Observable product behavior, verified against the service:

1. **An org-managed user can only belong to the organization that manages them**, plus organizations related to it through a parent/child multi-org relationship. Invites to unrelated organizations are rejected.
2. **An org-managed user cannot connect additional identity providers.** The exception is a user managed by a multi-org, who can connect SAML identities for related organizations only — no OAuth (GitHub/GitLab/Atlassian/Google) identities in either case.
3. **An org-managed user cannot create organizations.**
4. **Becoming org-managed is destructive.** The migration removes the user from their other organizations and deletes their non-SAML linked identities. This is closely related to the existing warning in [SAML admin](https://www.pulumi.com/docs/administration/access-identity/saml/saml-admin/) — "When a user stops being a SAML admin, they will automatically lose all other login methods" — which is currently the only place in the docs where anything like this appears.
5. **Org-managed status is not the same as SAML org membership.** An account *created by* logging into a SAML org is org-managed. A pre-existing Pulumi account that later joins a SAML org by invite is not — it stays an ordinary account with an added SAML identity. There is a separate opt-in migration in Account settings to convert. The docs never distinguish these, and the distinction determines whether restrictions 1–3 apply.

## Suggested placement

A new page under `content/docs/administration/access-identity/` (sibling to `saml/` and `scim/`), cross-linked from:

- `organizations-teams/accounts.md` — the [Adding New Identities](https://www.pulumi.com/docs/administration/organizations-teams/accounts/#adding-new-identities) section presents identity linking as unconditional; it should note that org-managed accounts cannot do this.
- `access-identity/scim/_index.md` — SCIM is how most org-managed users get created.
- `access-identity/saml/_index.md`.

## Related gaps found in the same area

Filing together since they surfaced from one investigation and touch adjacent pages. Happy to split into separate issues if preferred.

### A. Inaccurate statement in the Pulumi Cloud FAQ

[`content/docs/support/faq/pulumi-cloud.md`](https://www.pulumi.com/docs/support/faq/pulumi-cloud/), "How do I link an existing Pulumi account to my company's organization?":

> To join your company's organization, you must sign in with the identity provider that organization is backed by (for example, GitHub, GitLab, SAML/SSO, or email).

This is **incorrect for the OAuth-backed organizations** (GitHub, GitLab, Atlassian). For those, you must have the identity *linked to your account*; the identity provider you actually sign in with is irrelevant. Someone who signs in with email and has an Atlassian identity linked has full access to a Bitbucket-backed organization.

It is true for SAML, where authenticating through the IdP is what establishes the identity — so the fix is to distinguish the two cases rather than delete the sentence. The following sentence already gets it right ("connect that identity provider, then accept the organization invite"), so the two currently contradict each other.

Separately, the fallback advice in that answer — "If this fails, delete your account, then accept the organization invite" — is drastic for what is usually just an unlinked identity, and deleting an account has consequences the note below it describes. Worth reconsidering.

### B. Nothing documents that identity linking can be unavailable

[Adding New Identities](https://www.pulumi.com/docs/administration/organizations-teams/accounts/#adding-new-identities) describes connecting additional identities as always available. In practice the controls can be absent — org-managed status is one cause; a deployment without that provider's OAuth credentials configured is another. A short "if you don't see these options" subsection would resolve a class of support questions.

### C. Bitbucket: the admin requirement is understated

[Bitbucket identity provider](https://www.pulumi.com/docs/administration/organizations-teams/organizations/#bitbucket-identity-provider) says a workspace admin must grant the OAuth app read access. It doesn't say that the person performing the change in Pulumi must themselves be an **admin or owner** of the Bitbucket workspace — a contributor or plain member gets a hard error. Worth stating explicitly, alongside the fact that a personal Bitbucket account with no workspace won't work.

### D. Backend membership is not the same as Pulumi organization membership

On per-member billing plans, being a member of the backing GitHub org / GitLab group / Bitbucket workspace does not by itself make you a member of the Pulumi organization — it makes you eligible, and an organization admin still has to add you. Older plans behave differently and grant membership directly.

[Bitbucket identity provider](https://www.pulumi.com/docs/administration/organizations-teams/organizations/#bitbucket-identity-provider) gestures at this ("the admin can see a list of Bitbucket workspace members that they can add or invite") but never names the two-tier model, and it isn't stated for GitHub or GitLab at all. This is a common source of "I'm in the GitHub org, why can't I see anything?"

### E. Switching away from SAML is destructive and undocumented

[Disconnecting identity providers](https://www.pulumi.com/docs/administration/organizations-teams/organizations/#disconnecting-identity-providers) correctly explains that you remove SAML by selecting a different identity provider. It does not warn that doing so discards the organization's SAML identities, its SAML member roster, and its SCIM access token — meaning SCIM must be reconfigured with a newly issued token and users re-provisioned if you switch back. Given the SCIM docs are otherwise thorough, this deserves a warning note on both pages.

### F. Converting an organization to SAML can fail for the admin doing it

The SAML setup guides walk the reader to **Membership Requirements → Change requirements** and stop. The change is rejected if the admin performing it belongs to other, unrelated Pulumi organizations. This produces a confusing error at the last step of an otherwise well-documented setup flow, and should be listed as a prerequisite.

### G. Terminology is inconsistent across surfaces

The docs say "identity provider" and "Membership Requirements"; the console uses both that phrasing and "organization backend"; the API calls it the organization backend. For Bitbucket specifically, the docs and console alternate between "Bitbucket" and "Atlassian" (the connect control is labeled **Connect Atlassian**, and a linked identity appears as **Atlassian**) without ever stating they're the same thing. A short glossary note, or simply saying "Atlassian (Bitbucket)" on first use, would help.

jkodroff:
Adds a concepts page for organization-managed Pulumi Cloud accounts — accounts an organization creates and controls through SAML JIT provisioning or SCIM — and closes the seven adjacent documentation gaps found in the same investigation.

## Changes

**New page** — `concepts/org-managed-users.md`: how an account becomes org-managed, why an administrator wants it, how it differs from SAML organization membership, the three restrictions, and the destructive **Migrate to Org-Managed Account** migration. Cross-linked from Accounts, SAML, SCIM, and SAML admin.

**Adjacent fixes**

- FAQ: the answer on linking an existing account conflated *linking* an OAuth identity with *signing in* through it — now split by provider kind
- Accounts: new "If the identity options aren't there" section
- Organizations: Bitbucket workspace admin requirement; backing membership vs. Pulumi membership; destructive SAML disconnect; multi-org prerequisite for SAML conversion; Atlassian/Bitbucket naming; trimmed the roles table that duplicated RBAC
- SCIM: provisioned users are org-managed; identity-provider change revokes the SCIM token
- New `saml-conversion-prereq` shortcode, used in all six SAML guides

Concepts nav weights shifted to seat the new page after Accounts.

Fixes #21103

### DIFF
--- a/content/docs/administration/concepts/_index.md
+++ b/content/docs/administration/concepts/_index.md
@@ -19,6 +19,7 @@ How Pulumi Cloud models your organization and who can do what inside it. Read th
 
 - [Organizations](/docs/administration/concepts/organizations/) — the top-level container that owns your stacks, environments, and settings.
 - [Accounts](/docs/administration/concepts/accounts/) — individual user accounts, profiles, and identity providers.
+- [Organization-managed users](/docs/administration/concepts/org-managed-users/) — accounts an organization creates and controls through SAML or SCIM, and the restrictions that come with them.
 - [Agent accounts](/docs/administration/concepts/agent-accounts/) — accounts for AI agents and automation acting on your organization's behalf.
 - [Billing managers](/docs/administration/concepts/billing-managers/) — the role that delegates billing access without granting admin rights.
 - [Access tokens](/docs/administration/concepts/access-tokens/) — personal, team, and organization tokens for authenticating the CLI, CI/CD, and the REST API.
--- a/content/docs/administration/concepts/access-tokens.md
+++ b/content/docs/administration/concepts/access-tokens.md
@@ -7,7 +7,7 @@ menu:
   administration:
     name: Access tokens
     parent: administration-concepts
-    weight: 5
+    weight: 6
 aliases:
 - /docs/administration/access-identity/access-tokens/
 - /docs/intro/pulumi-service/organization-access-tokens/
--- a/content/docs/administration/concepts/accounts.md
+++ b/content/docs/administration/concepts/accounts.md
@@ -68,6 +68,15 @@ Connecting these additional identities will enable you to join Pulumi organizati
 
 In order to be invited as a member of a Pulumi organization, you must connect your account with the organization's backing identity provider. Once your account is linked to your third-party identity, you will then show up on the list of users that the organization admin can invite.
 
+Connecting an identity doesn't add you to the organizations that identity backs. An organization admin still has to invite or add you. See [Backing membership doesn't grant Pulumi membership](/docs/administration/concepts/organizations/#backing-membership).
+
+#### If the identity options aren't there {#identity-options-missing}
+
+The controls for connecting identities aren't available to every account. Two things remove them:
+
+* **Your account is organization-managed.** An account that an organization created through SAML or SCIM can't connect additional identity providers. See [Organization-managed users](/docs/administration/concepts/org-managed-users/).
+* **The provider isn't configured for your deployment.** In [self-hosted Pulumi Cloud](/docs/administration/self-hosting/), a provider only appears if the deployment has been configured with OAuth credentials for it. Ask whoever administers your deployment.
+
 ## Password Reset
 
 For users who signed up with email, you may change your password by:
--- a/content/docs/administration/concepts/agent-accounts.md
+++ b/content/docs/administration/concepts/agent-accounts.md
@@ -7,7 +7,7 @@ menu:
     administration:
         name: Agent Accounts
         parent: administration-concepts
-        weight: 3
+        weight: 4
 aliases:
   - /docs/administration/organizations-teams/agent-accounts/
 ---
--- a/content/docs/administration/concepts/audit-logs.md
+++ b/content/docs/administration/concepts/audit-logs.md
@@ -7,7 +7,7 @@ menu:
   administration:
         name: Audit Logs
         parent: administration-concepts
-        weight: 7
+        weight: 8
         identifier: administration-concepts-audit-logs
 aliases:
 - /docs/administration/security-compliance/audit-logs/
--- a/content/docs/administration/concepts/billing-managers.md
+++ b/content/docs/administration/concepts/billing-managers.md
@@ -7,7 +7,7 @@ menu:
   administration:
     name: Billing managers
     parent: administration-concepts
-    weight: 4
+    weight: 5
 aliases:
 - /docs/pulumi-cloud/access-management/billing-managers/
 - /docs/intro/pulumi-service/billing-managers/
--- a/content/docs/administration/concepts/customer-managed-keys.md
+++ b/content/docs/administration/concepts/customer-managed-keys.md
@@ -6,7 +6,7 @@ h1: Customer Managed Keys
 menu:
   administration:
     parent: administration-concepts
-    weight: 8
+    weight: 9
     identifier: administration-concepts-customer-managed-keys
 aliases:
   - /docs/pulumi-cloud/customer-managed-keys/
--- a/content/docs/administration/concepts/org-managed-users.md
+++ b/content/docs/administration/concepts/org-managed-users.md
@@ -52,7 +52,7 @@ Migrating an ordinary account to an organization-managed one is permanent and de
 Migrating an account deletes your personal organization, removes you from every organization other than the one taking it over, and deletes every linked identity except the managing organization's SAML identity. Any authentication method that depended on a deleted identity, such as GitHub or Google, stops working. Single sign-on through the managing organization becomes the account's only authentication method.
 {{% /notes %}}
 
-Pulumi rejects the migration unless all of the following are true:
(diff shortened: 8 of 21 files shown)


jkodroff:
@claude #update-review

Worked the full list. Several claims you flagged as unverifiable I verified against the `pulumi-service` source rather than leaving open — two of them changed the docs, two confirmed your doubt was misplaced, and one showed the doc was outright wrong.

**Fixed**

- **L203 (outstanding)** — replaced the hand-written SAML-conversion prerequisite in `organizations.md` with `{{< saml-conversion-prereq >}}`, keeping a narrative lead-in so it doesn't become a callout mid-section. You were right that two editable copies would drift.
- **L47 / L61-62** — you were right that there's a confirmation step. `cmd/console2/src/app/users/settings/saml/org-managed-migration.dialog.html` is a Cancel/Migrate dialog, so "takes effect immediately" is now "takes effect as soon as you confirm it," and the step list gained the dialog. The control label is confirmed character-for-character (`account-profile.component.html:97` → `Migrate to Org-Managed Account`, hyphenation as written), so no label change. Verifying it surfaced a bigger problem you didn't flag: the "Before you migrate" list was wrong. `validateUserCanBeOrgManaged` (`cmd/service/api/users.go`) enforces four *hard rejections* — no stacks in the personal org, no environments in the personal org (not transferable, only deletable), not an admin of any unrelated org, and a SAML identity for exactly one org — and migration deletes the personal organization outright. Rewritten accordingly.
- **L41** — the multi-organization language is gone from all three sentences that carried it. `MultiOrgList` (`cmd/service/model/organizations.go:201`) is a hardcoded map of org UUIDs that engineering edits when sales files a request (`.ai/skills/add-multi-org-exception`). It really does gate invites (`organizations_members_invites.go:897 areOrgsRelated`) and identity connection (`api/users.go:1513,1557`), so the behavior is real — but it has no self-service surface, so it isn't a documentable product concept and shouldn't debut in a public concepts page.
- **L81** — good instinct, and the answer is that the sentence was **wrong**. SCIM *is* conditionally retroactive: `users_scim_upsert.go:474` calls `makeUserOrgManaged` when a SCIM PATCH flips the user's `active` state (`users_scim_mapping.go:143`) and `validateSamlOmu` passes (empty personal org, no unrelated org memberships). The conversion is silent and not opt-in. Rewritten on the SCIM page, cross-referenced from the concept page's table.

**Disputing — please re-check with this in view**

- **L83-85** — "Pulumi has no endpoint for deleting a user through SCIM" is correct as written. `cmd/service/main.go:1685-1704` registers `POST /Users`, `PUT /Users/{userID}`, `PATCH /Users/{userID}`, and a `DELETE` only for `/Groups/{groupID}`. There is no `DELETE /Users` route. That absence is also what makes the adjacent "unlike users, teams are deleted outright" contrast true. The third-party directory asserting a user DELETE endpoint is simply wrong.
- **L120** — the roster half you singled out is confirmed, so no wording change. `replaceOrganizationBackend` (`model/organizations.go:1700`) deletes the `SAMLOrganizations` row; `scim_access_token` is a column *on that table* (`migrations/097`), and `SAMLIdentities` carries `fk_samlids_org_id ... REFERENCES SAMLOrganizations(org_id) ON DELETE CASCADE` (`migrations/089`), so the member roster cascades away with it. Switching back inserts a fresh row with `scim_access_token DEFAULT ""`. "Re-provision your users" is accurate, not overstated.

**Accepted, no change**

- **L35 / L36** — confirmed; this was verified against the service in the investigation recorded in #21103 (point 5).
- **pre-existing L36 (`saml-admin.md`)** — the one-admin limit is accurate, and your "the plural could be forward-looking" reading is the right one. `organizations_saml.go:263` reads the single nullable `samlOrgInfo.OwnerID` and wraps it in the `SAMLAdmins` slice; `UpdateSAMLAdmin` replaces rather than appends.

`make lint` passes.

CamSoper (approved):

PR #21119, opened 2026-08-25
Rename the Pulumi Cloud editions and update the pricing table

jeffmerrick:
### Proposed changes

Renames the Pulumi Cloud editions for the September 2026 launch and applies the card and comparison table changes in GRO-40.

| Old | New |
| --- | --- |
| Individual | Free |
| Team | Essentials |
| Enterprise | Pro |
| Business Critical | Enterprise+ |

* Edition **ids** in `data/pulumi_pricing.yaml` are renamed alongside the display names, so `editions:` values from before the rename now fail `make lint`. The three changelog entries carrying the key are updated.
* `/releases/index.json` publishes edition display names in its deprecated `tiers` field, so those values change with the rename.
* Docs availability callouts derive their edition names from the same file, so this shouldn't merge before the September launch - every marked page flips at once.
* Neo is dropped from the Free edition per GRO-40, so the Neo `AggregateOffer` on `/neo/` now spans only the editions that carry it, $40-$400 rather than $0-$400.
* The Pro card bullet reads "Automated audit log export" rather than GRO-40's "Audit logs and automated export", since audit logs now start at Essentials and the card sits under "Everything in Essentials, plus:".
* FAQ prose on `/pricing/` is renamed along with the cards - note: GRO-40 covers the cards and the table only.
* The data file header and AGENTS.md carried a rule against ever writing "Free" or "Pro" for an edition. Both now describe the new names, including "the legacy Pro SKU" for the pre-2023 grandfathered one and a rule against bare "Enterprise", which is ambiguous between Pro and Enterprise+.
* Hand-written edition mentions across docs, product, tutorials, and what-is prose are renamed too, including the stragglers a review pass caught afterward: the "free Individual tier and paid plans" boilerplate on ten `/docs/iac/comparisons/` pages, the access-token edition lists, the GitHub Enterprise Server cell in the Neo permissions table, the Pulumi Cloud concepts page, and the Cloudflare ESC guide. Blog posts are left as historical.

Open questions

* `neo-cli` and `neo-integrations` are still available from Free while the `neo` row is not, so the table shows Pulumi Neo unavailable on Free with two Neo rows checked below it. GRO-40 doesn't list those rows. If Free really loses Neo entirely, `neo-cli` at least should probably move to Essentials; the MCP server may genuinely be free.
* Related: the "Free and Essentials editions" section of `docs/administration/get-started/choose-edition.md` still says readers "can also make use of the Pulumi Neo" - true for Essentials, no longer for Free. Left as is pending the row question above.
* Commitment pricing moved to Essentials for IaC resources but stayed at Pro for ESC secrets, as GRO-40 names only one of the two rows. The prepay FAQ no longer enumerates editions as a result.
* Free advertises up to 2,000 managed resources while Essentials at $40/month includes 500, in the card note, the calculator, and the FAQ.

Out of scope (other repos)

* Generated artifacts still carry the old names and need upstream regeneration after launch: `data/openapi-spec.json` API descriptions ("available on Business Critical edition", etc., from pulumi-service) and the generated CLI docs (`pulumi_insights_resource_search.md` says "Enterprise plans only", from pulumi/pulumi help text).
* `data/newsroom.toml` keeps the 2022 Business Critical press-release title as historical record, same as blog posts.

### Related issues (optional)

[GRO-40](https://linear.app/pulumi/issue/GRO-40/pricing-table-update)


jeffmerrick:
@kramhuber @cnunciato Should we make all the changes needed elsewhere on the site in this PR or keep it scoped to just the pricing page?

cnunciato:
Are these new plans in effect today? I was under the impression it was still in progress. Here's what the console shows today when I go to upgrade: 

image

jeffmerrick:
@cnunciato Nope, I don't think anything will go live until September, just getting the work ready to go here.

cnunciato:
>Should we make all the changes needed elsewhere on the site in this PR

That'd make sense to me!

jeffmerrick:
Looks to be handled here https://github.com/pulumi/docs/pull/21509 so closing this one.

PR #21134, opened 2026-08-25
Replace homepage hero code selector with GSAP agent-loop animation

jeffmerrick:
### Proposed changes

Replaces the homepage hero's code selector composite with a GSAP-driven inline SVG that plays the agentic loop from design's storyboards (Figma "Storyboards R2", node 752:19223): a prompt goes to an agent, the agent writes the VPC program in a chosen language, CI checks pass, a platform agent runs `pulumi up` under policy, and the AWS VPC diagram grows in. One ~23s pass, then it loops with a fresh agent and language.

- Every pass (page load and each loop) randomizes both leads. Agents are weighted 66/15/8/4/3.5/3.5 across Claude Code, Codex, Cursor, opencode, Copilot, and Neo; languages by usage share with brand-new HCL boosted above YAML: TypeScript 51, Python 29, Go 7.5, C# 5, HCL 3.5, YAML 3, Java 1. Both tables live in `hero-animation.ts` (`AGENT_WEIGHTS`/`LANG_WEIGHTS`). The static no-JS/reduced-motion frame always shows Claude Code with no language chrome.
- The SVG is authored in its finished state, so no-JS and `prefers-reduced-motion` render the final frame statically with no layout shift; the animation pauses offscreen and on hidden tabs.
- The seven editor examples (one per language, adapted from the classic code-selector hero) and the terminal stream are plain-text sources at the top of `data/hero_agent_loop.yaml` — editing the content means editing those and running `yarn --cwd theme gen:hero-text`, which re-highlights them with shiki (min-light, each language's own grammar plus bash) and rewrites the generated sections underneath. shiki is a devDependency used only by that script; nothing highlight-related runs at build or page-load time. The generator fails if an example outgrows the panel's 13 rows - note: adding or removing terminal rows means shifting `TERM_SCROLL_END` by 18px per row.
- The generated runs anchor every text segment to an explicit column instead of carrying leading spaces, since `hugo --minify` (preview/production only) collapses whitespace inside the tspans.
- `template-hero.html` gains `hero_animation`, plus an explicit `hero_visual` key ("animation", "code", "video", "image") so alternate visuals can stay fully configured while one is active; pages without it keep the existing precedence. The homepage keeps its full `code_snippets` block configured, so `hero_visual: code` restores the classic selector in a one-word change. The composite itself is untouched and still used by `/product/infrastructure-as-code`, `/gcp`, `/azure`, and `/kubernetes`.
- New `hero-animation` webpack entry loaded only by the homepage, ~31KB gzipped including the new `gsap` dependency (core only).
- The Monaspace preload no longer skips the homepage, since the hero now types monospace above the fold.
- Strokes use `vector-effect: non-scaling-stroke` except the dash-drawn rects, since Chrome measures their dashes in screen units under that effect.


cnunciato (approved):
This looks great! Made a few tweaks to some copy and adjusted the measurements to make everything fit. 

The only thing I noticed (which we can definitely address as a follow-up) is that at narrower widths, the animation is slightly off center, and vertical whitespace feels like it could be tightened. So we may want to make a few tweaks to address that: 

https://github.com/user-attachments/assets/c67997b0-9cdd-42ec-ad0c-cd3670a065cb

Otherwise, this is good to go. I'll merge and we can come back to it once I verify all's well in a few other browsers.

PR #21165, opened 2026-08-27
infrastructure: use the project no-cache policy for the support redirect

cnunciato:
Follow-up to #21161, which unblocked the master deploy by resolving the AWS-managed `Managed-CachingDisabled` policy by name.

This switches the support-redirect distribution to the project's own `noCacheKeyPolicy` (`cacheKeyPolicy("no-cache", 0)`) instead, so it references an infrastructure-owned cache policy by `.id` — the same pattern every other CDN behavior in `index.ts` already uses — rather than reaching for a managed policy by name.


CamSoper (approved):
This LGTM!

PR #21183, opened 2026-08-27
Render Pulumi Cloud edition banners on blog posts

jeffmerrick:
### Proposed changes

- `pulumi_cloud_feature:` front matter now renders on blog posts, at the foot of the post, through the same partial `/docs` uses. The section-level `{{< pulumi-cloud >}}` shortcode already worked in posts, so only the whole-page form needed wiring.
- `layouts/partials/docs/cloud-availability.html` moved to `layouts/partials/cloud-availability.html`, as it is no longer docs-only.
- `context-api` added to `data/pulumi_pricing.yaml`, available from Enterprise and hidden, so the callout links to `/pricing/` rather than a row anchor.
- Marked on the Context API post, replacing its two hand-written edition sentences. The intro sentence no longer names the editions, so neither does the post's excerpt on `/blog/`.
- No `updated:` stamp on that post, since it published 2026-08-26 and the editions it states are unchanged.

**Open questions:**
- One overall consideration: blog posts are historical and the banner could change without the blog post getting an updated stamp. Is this ok?
- Do we want the banner on the latest Context API blog post since it's a feature in preview? I did it as a test mostly.
- If so, should we keep the hand-written sentences? (Knowing they may drift if availability changes in the future.)
- Do we want to try to add to any older (but recent) feature announcement blog posts?

Screenshot 2026-08-27 at 9 31 29 AM


cnunciato (commented):
I'm actually not sure on this one. I do see the benefit, and this of course works really well in the docs, but I'm not sure it works quite as well for the blog. A few things:

* There isn't always a 1:1 relationship between the topic of a post and a single product feature. Launch posts, for example, typically cover several. So that could get a bit awkward: "this feature" (which one?), posts that include (or even just mention) features available to all tiers, and so on. I can imagine it being somewhat confusing.
* The inline prose explanations are also kind of nice in that they give you way to work the tiers into the prose organically. They're a little more tedious to maintain this way for sure, but agents also make that fairly easy nowadays. 
* In docs pages, it's usually clear what "this feature" refers to, because the whole page is about that one thing. Blogs aren't always so focused, though. I'd almost want the widget to say something like, "The features in this post are available..." — but there again, you'd still run into the whole multiple-features-spanning-multiple-tiers problem.

So I guess on the whole, I'm kinda thinking maybe we don't totally need this? I suppose it could work if the template were able to accommodate multiple features, and the widget were tweaked to use the "features in this post" phrasing, but yeah — I'm also wondering whether just having the review bot call out missing tier info wouldn't ultimately solve the problem and be more flexible. (Boring as that may be. 😇)

jeffmerrick:
@cnunciato All good points, 100% defer to you on this. The existing shortcode used on docs could be inserted anywhere in a post before this PR, so it's something we could just leave up to authors.

jkodroff:
I am also on the fence about this one and because the blog not my wheelhouse and also I am a coward, I'm avoiding being the one to make a call on this.

cnunciato:
> The existing shortcode used on docs could be inserted anywhere in a post before this PR, so it's something we could just leave up to authors.

Ah right! That might be the best way to go, yeah. That, plus an addition to the review skill (probably to nag rather than block) ought to do it.

PR #21188, opened 2026-08-27
Homepage hero animation: dark code panel, looping, and panel roll-ups

jeffmerrick:
### Proposed changes

* Code panel uses shiki's `min-dark` on a violet-950 background. `dark_code` in `data/hero_agent_loop.yaml` selects the treatment, and `theme/scripts/gen-hero-text.mjs` copies the flag into the generated file next to the syntax colors it picked, so the partial paints the panel from the same flag. Flipping it needs `yarn --cwd theme gen:hero-text`.
* The timeline loops. The finished frame rests for 10 seconds with the cubes bobbing, then fades out, and the next pass picks a fresh agent and language.
* Code and terminal panels roll up as their shell collapses rather than fading out. Each has its own clip window now, so one can close without touching the other.
* Cube geometry taken from the storyboard export - the base traces only the two front-facing edges, and each label sits inside its cube's group.
* `normalizeColor` in the text generator stripped a trailing `FF` from any hex, turning `min-dark`'s `#79B8FF` into `#79B8`. It now only drops a fully opaque alpha from an 8-digit value.
* `theme/src/ts/hero-animation.ts` gained a header covering the rules the loop depends on, since breaking any of them still renders a correct first pass. AGENTS.md has a section for the four files the animation spans and the regen step.


cnunciato (approved):
Terminal scroll-end remeasured for the two new `Loading policies...` rows so the closing Duration line clears the fold. LGTM.

PR #21191, opened 2026-08-27
Triage: route theme/ to domain:infra, add a domain:other fallback

CamSoper:
### Proposed changes

Two gaps in triage's path map, both surfaced by [#21164](https://github.com/pulumi/docs/pull/21164) — a change touching only `theme/src/scss` and `theme/src/ts` that came out of triage with no `domain:` label. Triage ran fine ([run](https://github.com/pulumi/docs/actions/runs/33036291604)); it just had nothing to say.

1. **`theme/` now routes to `domain:infra`.** It matched no rule in `classify_path()` and fell through to `None`. It's the same kind of path as `layouts/` and `assets/`, which already route to infra. This also gives those files the infra *review criteria*, which they weren't getting — the review lane is path-based off `domain-routing.md`, and `theme/` was landing in its catch-all.
1. **A PR where nothing matches now gets `domain:other`**, so an unlabeled PR unambiguously means triage never ran. It fires only when nothing else matched, so it can't flip `mixed` on a PR that isn't.

`domain:mixed` is the tempting reuse for that fallback, but it's derived from `len(domains) > 1` and means "touches more than one domain" — pointing it at a single unmatched file asserts the opposite of what's true and collapses the label's filter value. Swapping the two is one string in `FALLBACK_DOMAIN` if the tradeoff reads differently to you.

Also: `domain-routing.md` now matches the classifier (its table sent `layouts/`, `assets/`, and `static/` to the shared-criteria catch-all while the code labeled them infra); `labels-pr-review.md` gains `domain:other` and `domain:website` (the latter has been applied by triage all along, just never listed); routing tests added, 8 checks → 26. Both test functions now assert on recorded failures so they fail under pytest, not just standalone — previously a broken rule printed `FAIL` lines and still reported a green suite.

### Before merging

`domain:other` doesn't exist in the repo yet and needs creating, or it silently won't apply — triage batches its whole ADD set into one `gh pr edit --add-label a,b,c`, which `gh` rejects entirely on an unknown name, and the workflow's `|| true` swallows the error.

```bash
gh label create "domain:other" --color ededed --description "PR touches no domain-specific path; reviewed under shared criteria only"
```

### Testing

- `make test-review-pipeline` green. Replaying #21164's file set through the fixed classifier gives `domain:infra`.
- `make lint` doesn't cover this diff — markdownlint scans only `content/`, and `.prettierignore` excludes `*.md` and `scripts`. It was not run locally: `make ensure` needs hugo and Node 24.

### Related issues

Follows from [#21164](https://github.com/pulumi/docs/pull/21164).

https://claude.ai/code/session_01ALUobedhegnGSPKD1W24zb

### DIFF
--- a/.claude/commands/docs-review/references/domain-routing.md
+++ b/.claude/commands/docs-review/references/domain-routing.md
@@ -9,13 +9,15 @@ Each changed file routes to **exactly one** domain by path. Apply the rules in o
 
 | Order | Domain | Applies when the file path matches |
 |---|---|---|
-| 1 | `docs-review:references:programs` | `static/programs/**` (includes every nested file in a program directory: `Pulumi.yaml`, `package.json`, `requirements.txt`, source files) |
+| 1 | `docs-review:references:programs` | `static/programs/**` (includes every nested file in a program directory: `Pulumi.yaml`, `package.json`, `requirements.txt`, source files) and `scripts/programs/**` |
 | 2 | `docs-review:references:blog` | `content/blog/**`, `content/case-studies/**` |
 | 3 | `docs-review:references:docs` | `content/docs/**`, `content/tutorials/**`, `content/what-is/**` |
 | 4 | `docs-review:references:website` | Any other `content/**.md` (pricing, legal, `vs/`, `why-pulumi/`, `about/`, `careers/`, etc.) |
-| 5 | `docs-review:references:infra` | `.github/workflows/**`, `scripts/**` except `scripts/programs/**`, `infrastructure/**`, `Makefile` (repo root), `package.json` (repo root only), `webpack.config.js`, `webpack.*.js` |
-| 6 | `docs-review:references:shared-criteria` only | Anything else (`layouts/`, `assets/`, `data/`, etc.) |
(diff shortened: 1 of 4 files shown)


CamSoper:
@claude

L43 - I've added the label manually. We're fine.

Implement your suggestions for L17 and L24 and then #update-review

CamSoper:
@claude #update-review

Both patches applied locally in d0af0189 (the earlier run couldn't write under `.claude/**`).

**L24** — confirmed the bug before fixing: breaking only `OVERSIZED_TOTAL_FILES` failed *both* tests under pytest, reporting the second as "test_domain_routing: 1 check(s) failed" for a check it doesn't own. Each test now captures `len(_failures)` on entry and `assert_clean` asserts on the slice it recorded itself, with the baseline required rather than defaulted. Verified in both directions afterwards: breaking the oversized threshold fails `test_oversized_threshold` alone; breaking the `theme/` rule fails `test_domain_routing` alone with its own count of 5.

**L17** — applied at both sites verbatim, `domain-routing.md` rule 6 and `labels-pr-review.md`'s `domain:other` row.

`make test-review-pipeline` green.

---

PR #21192, opened 2026-08-27
Tune the content-review glow-up and fix lanes

CamSoper:
### Proposed changes

Three tuning changes to the content-review lanes, from an audit of the live ledger (146 entries) on 2026-08-25.

1. **Glow-ups no longer run on backlogs they can't read.** The selector queues on a counter (`skipped_findings`); the worker fetches the items from a different store. 50 of 115 eligible pages have neither a findings record nor a readable PR body — 43% of slots executed nothing. `select-glowup.py` now checks recoverability before queueing, skipped when `--findings-dir` is absent or it would strand the whole corpus.

2. **The freed slot repairs the data.** Unrecoverable pages ride a new `repairs[]` array and go to the **fix** lane, whose review writes the findings record a real glow-up needs. One per run. Retires the two-retry loop (`GLOWUP_DEGRADED_ATTEMPT_CAP`) and wires `--strict` as a clean skip rather than a red X — the skip verdict carries the banked count forward, so recording it can't delete the debt it just failed to collect.

3. **One fix-lane slot reserved for the oldest never-reviewed page.** Tier 3 was 0/288 reviewed; `score = importance × staleness` and importance spans ~22×, so cold pages never surface. Simulated: first full pass 862 → 556 days, tier 3 at one year 12/288 → 151/288. Fix lane only, disabled at `count 1`.

Also adds `backfill-findings-records.py` — a one-time hand-run repair for the 46 stranded entries whose PR body is still readable. Reuses the existing scraper; marks records `synthetic: true`. Deliberately not wired to a workflow.

**Verification:** `make test-review-pipeline` green, with new cases for each change. The `--strict` skip and both new jq fragments were exercised end to end against a synthetic ledger. `make lint` covers neither `.py` nor `.github/**`, so it's a no-op here. Worth one `dry_run=true` dispatch after merge to watch the repair path against the real S3 ledger.

N/A — tooling only.

### Related issues (optional)

Unrelated to #20664, the Vale-rules work this audit came out of.


CamSoper:
@claude #update-review

Both findings addressed in 97500ea3. The auto-refresh gate declined (`fire=false — hunk .github/workflows/review-existing-content.yml:340-346 is outside outstanding finding lines`), so this is the manual refresh.

**Fixed [L249] — the fail-open guard could never fire.** Confirmed exactly as described: `mkdir -p .findings-cache` runs unconditionally before the sync, the sync failure is swallowed by `|| echo`, so `[ -d .findings-cache ]` always passes and `check_recoverable` was `True` against an empty directory. Took option 1 — the guard now tests directory contents rather than existence:

```python
check_recoverable = findings_dir is not None and any(findings_dir.glob("*.json"))
```

Absent, empty, and failed-sync all take the skip path, and the invariant stays in the script that depends on it.

Two of the existing tests had used a *nonexistent* `--findings-dir` to mean "check runs, finds nothing" — which the new predicate correctly treats as skip — so they now point at a directory seeded with one unrelated record. Added the case that was missing: an existing-but-empty directory filters nothing and routes nothing to a repair. Verified it fails against the old predicate (only the PR-pointer page survived selection), so it genuinely covers the regression.

**Fixed [L421] — BUILD-AND-DEPLOY.md drift.** One sentence each added to the §Scheduled workflows bullet for the glow-up lane's fix-lane repair dispatch (`GLOWUP_REPAIRS_PER_RUN`) and the fix lane's reserved first slot (`NEVER_REVIEWED_RESERVE`).

**One hunk outside the flagged lines**, which is what stopped the gate and is worth a look on refresh: the dispatcher's bare `echo` on a failed findings sync is now a `::warning::`, so the degradation reaches the run summary rather than only the log. Same "no silent darkening" goal as the blocker; not required by either finding.

Also caveated the `NEVER_REVIEWED_RESERVE` day-counts in-place — they were simulated at 3 fix slots/day and this repo currently runs `CONTENT_REVIEW_COUNT=5`, so they are directional rather than a forecast. The starvation they describe is not sensitive to the count.

CI is green on the new head, including the review-pipeline suite.


---

PR #21193, opened 2026-08-27
Add deprecation note and link to crd2pulumi docs

guineveresaenger:
As of pulumi-kubernetes v4.34.0 (pending), the extended Kubernetes provider can do everything crd2pulumi can do, better, backwards, and in heels.
This contains a link to unpublished how-to, so must be merged after https://github.com/pulumi/registry/pull/12267.

### DIFF
--- a/content/docs/integrations/clouds/kubernetes/crd2pulumi.md
+++ b/content/docs/integrations/clouds/kubernetes/crd2pulumi.md
@@ -13,6 +13,8 @@ aliases:
 - /docs/iac/clouds/kubernetes/crd2pulumi/
 ---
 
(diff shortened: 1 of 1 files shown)


guineveresaenger:
Both the review findings are true and are pointed out in the PR message. :|

CamSoper:
@claude Note @guineveresaenger's comments and #update-review

@guineveresaenger --For future reference, you just need to verify you know what you're doing using the right format. I tried to edit your comment the way I would have done it, but editing a comment doesn't trigger workflows the same way a new comment does.

CamSoper (approved):

PR #21227, opened 2026-08-28
Generate the audit log events reference from the Pulumi Cloud API

### LINKED ISSUE #21111: Audit log events list has drifted: 87 emitted events undocumented, 3 documented events not emitted
Move the list of Audit Log events from inline at https://www.pulumi.com/docs/administration/concepts/audit-logs/#list-of-audit-log-events to a reference page:

- The page should go under Reference in this docs section
- The page should be automatically updated in a manner similar to ESC schemas. There is now an API endpoint that should be live in production from which to pull the values
- Ideally, we would categorize the events to allow them to be scanned via headings in the nav, but this needs to be manageable via the metadata provided by the API endpoint or a maintainable, deterministic script-type process
- The page should be styled similar to other reference docs (pick the most appropriate example)

jkodroff:
Fixes #21111

The hand-maintained audit log events table documented 63 of the 153 events Pulumi Cloud emits — three of them under names the API no longer uses — and `content/docs/esc/administration/audit-logs.md` kept a second, staler copy. Both are replaced by a reference page regenerated nightly.

- **`scripts/fetch-audit-log-events.sh`** + **`.github/workflows/update-audit-log-events.yml`** — nightly at 07:15 UTC, modeled on the ESC schemas pipeline. Opens an auto-merging PR only when the data actually changes.
- **`data/audit_log_event_categories.yaml`** — the 13 categories and the exact-ID-beats-longest-prefix rules that fill them, since the API returns no category. An unmapped event fails the job and pings `#docs-ops` rather than landing in an "Other" bucket.
- **New page** at `/docs/administration/reference/audit-log-events/` — one `<h2>` per category, three columns (Event / Event ID / Description), permission requirements as inline notes. Adds the event IDs the old table lacked.
- **Old tables** replaced with Learn more sections. The ESC page is now a signpost; its screenshot and duplicated console steps are gone.

**Known issue:** 11 events carry "Pulumi Insights" in their API strings. Rendered verbatim so the page matches what the log actually prints; fix filed upstream at pulumi/pulumi-service#48765, and the nightly job picks it up once it lands.

**Before merge:** dispatch the workflow once with `skip_auto_merge: true`. Against freshly committed data it should report `has_changes=false` — confirming the bot's org token can reach the endpoint, which is the one thing not verified locally.

Verified: fetch script byte-identical across runs; unmapped-event path exits non-zero naming the IDs; `make lint` passes; full Hugo build exits 0 with no errors. Note that `make lint` does **not** validate data-file YAML — only the Hugo build catches a malformed `data/*.yaml`.

https://claude.ai/code/session_01Cn8MUHQ3r3RxBdzVXmciRa


CamSoper:
@claude Implement both of the suggestions for the low-confidence findings and #update-review

CamSoper (approved):

PR #21229, opened 2026-08-28
Document passkey sign-in for Pulumi Cloud accounts

### LINKED ISSUE #21108: Administration: Passkeys are not documented
We need to add a short section that we support passkeys, under what circumstances (whether there are any org-level restrictions an admin could put in place to require or disallow them), and how to generate a passkey.

jkodroff:
Passkeys shipped in July 2026 but have no coverage under `content/docs/`. This adds a **Signing in with a passkey** section to the accounts concept page, beside the existing MFA and password sections.

Fixes #21108

## Changes

`content/docs/administration/concepts/accounts.md`:

- Lists passkeys among the supported sign-in methods, linking to the new section.
- New `### Signing in with a passkey` subsection: what a passkey is, who can register one, the registration procedure, rename/remove behavior, signing in, and that a passkey does not replace password sign-in.
- Info callout on the passkey/MFA interaction.
- Note that self-hosted deployments only show the section when configured for passkeys.

No frontmatter, nav, or pricing-marker changes. Prose only.

## Ground truth

Written against `pulumi/pulumi-service` at `7446d9882f` rather than from the announcement post, because two blog claims no longer match the code.

**1. TOTP MFA does not prompt on a passkey sign-in.** `finalizePasskeyLogin` (`cmd/service/api/passkey.go:328`) mints a login token directly; `PerformSecondFactorAuthentication` is called only from the two password paths (`cmd/service/api/login.go:690`, `:1396`). This looks deliberate — `services/passkey/passkey.go:44-52` requires user verification on every ceremony so a passkey is possession + biometric/PIN on its own. The [blog post](https://github.com/pulumi/docs/blob/master/content/blog/passkey-support-in-pulumi-cloud/index.md) says the opposite at line 77 ("will still prompt for your second factor").

**2. Any account can register a passkey**, not just email/password accounts. `PasskeyRegisterBeginHandler` (`passkey.go:136`) checks only for a session, and `<app-passkey-settings />` renders unconditionally (`user-settings.component.html:29`). The blog scopes the feature to password users at line 45.

The docs describe the code. **The blog post is untouched here** — worth a Cloud team confirmation of intent before anyone edits it.

## Answering the issue's org-restriction question

The issue asks whether an admin can require or disallow passkeys. **No such control exists.** The only gates are a global LaunchDarkly flag (`passkeys-enabled`, evaluated with a no-org context, `services/featureflag/flags.go:35-42`) and the `PULUMI_PASSKEY_CEREMONY_KEY` env var, which self-hosted deployments must set or the routes 404 (`config/env.go:80-86`). `model/auth_policies.go` governs OIDC token exchange, not human sign-in. Per review discussion this is recorded here rather than stated on the page.

## Worth a second opinion

A SAML- or SCIM-managed account can register a passkey and then sign in with it, bypassing the organization's identity provider. That reads like a policy gap rather than a docs gap, so it is not documented as intended behavior.

## Testing

- `make lint` — 1859 files, 0 errors; Prettier clean.
- `make lint-prose` — nags only.
- Rendered locally: both anchors resolve, callout renders, left nav unchanged.

### DIFF
--- a/content/docs/administration/concepts/accounts.md
+++ b/content/docs/administration/concepts/accounts.md
@@ -168,9 +168,9 @@ To disable multi-factor authentication, select **Reset authentication method** i
 
 ### Signing in with a passkey
 
-A passkey lets you sign in with Touch ID, Face ID, Windows Hello, or a hardware security key instead of typing a password. It is a public-key credential: the private key stays on your device or in your password manager, and Pulumi Cloud stores only the public half. Passkeys are built on the [WebAuthn](https://www.w3.org/TR/webauthn-3/) standard, which every major browser and operating system supports.
+A passkey is a public-key credential that lets you sign in with Touch ID, Face ID, Windows Hello, or a hardware security key instead of typing a password. The private key stays on your device or in your password manager, and Pulumi Cloud stores only the public half. Passkeys are built on the [WebAuthn](https://www.w3.org/TR/webauthn-3/) standard, which current versions of every major browser and operating system support.
 
-Any Pulumi account can register a passkey. Passkeys are available in every edition at no additional cost.
+Passkeys are available in every Pulumi Cloud edition at no additional cost.
 
 To register one:
 
@@ -179,11 +179,11 @@ To register one:
 1. In the Passkeys section, select **Register a passkey**.
 1. Complete your device's prompt: Touch ID, Face ID, Windows Hello, or your hardware key.
 
-Pulumi Cloud names the passkey after the authenticator it recognizes, such as "Windows Hello" or "Chrome on Mac". Select **Rename** to give it a name of your own, or **Remove** to delete it. Removal takes effect immediately, and a deleted passkey can no longer sign you in.
+Pulumi Cloud names the passkey after the authenticator it recognizes, such as "Windows Hello" or "Chrome on Mac". Select **Rename** to give it a name of your own, or **Remove** to delete it. Removal takes effect immediately, and a deleted passkey can no longer sign you in. It does not end sessions that are already signed in, so sign out of those separately if you have lost the device.
 
 You can register as many passkeys as you want. One per device is common, as is one synced credential plus a hardware key as a backup.
 
-To use one, select **Sign in with a passkey** on the sign-in page. If your browser supports passkey autofill, the email field also offers your registered passkeys as suggestions.
+To use one, select the passkey option on the sign-in page. If your browser supports passkey autofill, the email field also offers your registered passkeys as suggestions.
 
 Registering a passkey does not disable any other way of signing in. Your password, if you have one, keeps working, so losing every registered passkey does not lock you out of your account.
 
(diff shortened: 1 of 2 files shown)


jkodroff:
@claude #update-review

Every flagged claim was verified against the `pulumi-service` source (`7446d988`), which is the ground truth the review's web searches couldn't reach. Pushed `23506ab`.

**Fixed:**

- **L173 (🚨)** — dropped "Any Pulumi account can register a passkey." You were right that it was an overclaim, though not for the stated reason: `PasskeyRegisterBeginHandler` applies *no* eligibility gate, and nothing in the org-managed migration path deletes `passkey_credentials`, so passkeys do survive migration. Rather than document what may be unintended service behavior, the sentence is gone; the line now states only edition/cost availability. `org-managed-users.md` is untouched.
- **L171** — scoped to "which current versions of every major browser and operating system support." Agreed the spec can't source an adoption claim; kept the link, which supports "built on the WebAuthn standard."
- **L182** — you were right to ask about sessions. `DeletePasskeyHandler` is a synchronous credential delete and does not invalidate tokens already minted, so added: "It does not end sessions that are already signed in, so sign out of those separately if you have lost the device."
- **L186** — good catch on the button label. It is *not* stable: `email-login-form.component.html:57` says "Sign in with a passkey" while `unified-auth-form.component.html:38` says "**Continue** with a passkey", selected by `unifiedSigninEnabled` in `welcome.component.html`. Replaced the bolded string with a description of the control.
- **L194** — named the real setting: `PULUMI_PASSKEY_CEREMONY_KEY` on the API service (`cmd/service/config/env.go:86`), base64 of 32 bytes.
- **Vale L171** — folded the wordy "It is a public-key credential" into the opening sentence.

**Disputing, with source evidence — please mark these verified rather than unverifiable:**

- **L191 (🚨) and its L191 companion** — both halves confirmed. `services/passkey/passkey.go:55` sets `UserVerification: protocol.VerificationRequired` on the WebAuthn engine, with a code comment giving the same reasoning the callout does ("Requiring UV means a verified passkey is genuine possession + biometric/PIN"). For the OTP half: `api/login.go:690` calls `PerformSecondFactorAuthentication` on the password path, while `api/passkey.go:255` goes straight to `finalizePasskeyLogin`, which mints a login token with no second-factor call. The callout is accurate as written.
- **L182 (Rename/Remove)** — verbatim in `passkey-settings.component.html`: "Register a passkey" (L13), "Rename" (L59), "Remove" (L62).
- **L182 (default names)** — "Windows Hello" and "Chrome on Mac" are both literal values in `wellKnownAAGUIDs` (`services/passkey/aaguid_names.go`). Real, not illustrative.
- **L184 (no cap)** — no per-account limit exists anywhere in the API or console.
- **L188 (×3, password fallback)** — passkey registration never touches password or identity rows, so the fallback holds. Declined the extra clause for password-less users: L86 already enumerates every sign-in identity.
- **L173 (pricing)** — confirmed ungated. The passkey code paths contain no edition or pricing check; the only gates are the `passkeys-enabled` LaunchDarkly flag and the ceremony key. No `pulumi_pricing.yaml` entry added, since the feature is available at the Individual edition and correctly carries no `pulumi_cloud_feature` marker.

One note for a follow-up, out of scope here: `agent-accounts.md:48` enumerates sign-in identities and omits passkeys. The claim flow uses the standard sign-in path, so that list has likely drifted — worth its own PR.

jkodroff:
@claude #update-review

Fixed the carried-over **L86** finding in `87527413`, and answered your author question from the service source.

**Can a user complete an agent-account claim with a passkey? Yes.** `claim/components/claim-unauthenticated.component.html:121` renders the same `<app-email-login-form>` the sign-in page uses, and that component gates its passkey button on the `passkeys-enabled` flag alone (`email-login-form.component.ts:106-117`) — there is no claim-context or route check. Your alternative reading (that a fresh claim needs an identity provider because there's no account to hold a passkey yet) doesn't hold: the claim link is followed by an *existing* Pulumi user signing in to absorb the placeholder account, so they may well already have a passkey registered.

So the two lists had genuinely drifted, and `agent-accounts.md` was the stale copy. Added "or a passkey" to the identity list at L48. Good catch carrying this one over — the original blocker's fix did supersede it without touching it.

That clears the last open item. `make lint` passes (1859 files, 0 errors).

CamSoper (approved):

CamSoper:
@claude #update-review

PR #21295, opened 2026-09-01
Context API guide: let the CLI supply the organization

lblackstone:
The guide told readers to pass `-F orgName=my-org` on every `pulumi api` example. That flag *overrides* the organization `pulumi api` already resolves on its own, so the placeholder was the only value readers ever sent, and the copy-paste failed with `Organization 'my-org' not found`.

`pulumi api` fills `{orgName}` from the currently selected stack, falling back to the CLI's default organization (`pulumi org get-default`).

This drops the flag from all five examples and adds a prerequisite covering the resolution order, plus the `402 Payment Required` a personal-organization default produces.

Verified against api.pulumi.com: every edited command runs as written.


lblackstone:
@pulumi-eon /review

PR #21308, opened 2026-09-01
Claim extractor: net diff, fence seeding from the checkout, distinct stances survive merge

CamSoper:
Three recall gaps in the pre-merge claim pipeline, all traced from PR #21291 (the `convert-hcl.md` glow-up). The agent added "`pulumi convert` is the fastest path for most configurations, and it's where to start" and the pre-merge review never put that sentence in front of the verifier. Review comment r3908440407 caught it by hand.

## What changed

**Net diff everywhere.** Six pre-steps fetched `gh pr diff <pr> --patch`, which is a per-commit mailbox, not the net diff. A sentence one commit adds and a later commit removes was still extracted and verified ("contradicted" against a file that no longer contains it). `extract-claims.py`, `extract-claims-llm.py`, `extract-urls-and-fetch.py`, `readthrough.py`, `vale-findings-filter.py`, and `post-style-suggestions.py` now take the plain `gh pr diff`. These hunks are copied verbatim from commit 72a3c34d710 on `CamSoper/review-v3` so the later rebase is clean; `test_no_script_uses_per_commit_pr_patch` locks it at the source level.

**Fence state is seeded from the checkout.** `iter_added_lines()` reset `in_fence = False` at every hunk boundary. In run 33518039058 the `@@ -1217` hunk opens inside a code block (21 fence markers precede line 1217 at that PR's head), so the first fence it met was a *closing* marker read as an opener: lines 1220-1260 were tagged in-code, the real prompt block at 1263-1267 as prose, and every `prose_only` matcher was suppressed. Only `url` fired on the "fastest path" line. Each hunk now seeds its fence state by counting markers above `new_start` in the file at `--repo-root` (default: cwd, which is the checkout in both workflows). The diff-only guess survives only as the fallback when the file isn't there. The old docstring's "FP-tolerant" claim is gone; it was a false-negative generator.

**Merge keeps a distinct stance.** The regex `positioning` record's text is the whole added line, so it overlapped 0.93 with an LLM `capability` restatement of the *other* sentence on that line (threshold is 0.34). The LLM text won as representative and "fastest" vanished. `merge-claims.py` now emits a regex `positioning`/`comparison` record as its own claim when the cluster's representative carries a different type. `TOKEN_OVERLAP_THRESHOLD` and `RANGE_WINDOW` are unchanged.

## Evidence and fixtures

- `testdata/pr21291-fence-seed.diff` is the real `@@ -1217` hunk from run 33518039058's `review-handoff` artifact (`.review-changes.patch`), real line numbers kept.
- `testdata/pr21291-head/.../convert-hcl.md` is the page at PR #21291's head (`article-base.txt` from the `review-snapshot` artifact with the handoff patch applied), so the seeding has a file to count.
- `test_fixture_pr21291_hunk_opening_inside_a_fence` asserts: a `positioning` claim is emitted for L1226; the hunk's `bash` and `text` code blocks yield nothing; pointing `--repo-root` at an empty dir reproduces the old behavior (url only).
- `test_merge_keeps_a_distinct_regex_stance` builds the regex/LLM pair from the real line (overlap > 0.9, different sentences) and asserts both claims survive with their types intact, and that two `positioning` records still collapse to one.

Both run under `python3 test_extract_claims.py` and under pytest via `make test-review-pipeline`; the harness now has an `assert_clean` so pytest can't report green on recorded FAILs.

## Verified offline vs. still needs a live run

- Offline: `make test-review-pipeline` and `make lint` green; the fixture reproduces the exact suppression and its fix.
- Live: the next pre-merge review on any docs PR exercises the net-diff fetch and the seeding against a real checkout. Nothing here is gated by a workflow change.

Not in scope: `scripts/review-v3/`, `compose_v3`, and the v3 surface stay on `CamSoper/review-v3`.

---


CamSoper:
@claude #update-review

Fixed L191 in 48d5c97: fence state is now CommonMark-style in both the seed and the diff walk. `fence_toggle` tracks the open marker (char + length) and a fence closes only on a same-char marker at least as long as its opener, so a ``` nested inside a ```` block is content rather than a toggle. `fence_open_at` walks the lines above the hunk with the same rule instead of counting parity. New regression test `test_fence_seeding_survives_nested_fences` covers a ```` block containing ``` lines and a ~~~ block whose body mentions ~~~, for both the seed and the walk.

Accepting L1220 as-is: that line is inside the frozen fixture copy of an already-published page (`testdata/pr21291-head/...`), exercised only by the test suite. Nothing new is asserted to readers, so re-running the verdict wouldn't change the PR.

---

CamSoper:
@claude #update-review

Fixed L171 in debc687: deleted the callerless `fence_open_at` wrapper and rewrote the `fence_state_at` docstring so it no longer points at a second consumer. `git grep fence_open_at` over the scripts directory now returns nothing; the harness and pytest are green on this head.

---

CamSoper:
Review status: three findings raised across two rounds; all resolved. L191 (whole-file parity inversion on nested fences) fixed in 48d5c97 with CommonMark fence matching; L171 (dead wrapper) removed in debc687; L1220 accepted as-is (frozen fixture of a published page). Zero open.

**Verified offline**
- `make test-review-pipeline` and `make lint` green on debc687.
- The real `@@ -1217` hunk from run 33518039058 plus the page at that PR's head: the seeded walk emits the "fastest path" positioning claim, the hunk's code blocks yield nothing, and an empty repo root reproduces the old url-only result.
- Nested-fence regression: a ```` block containing ``` lines, and a ~~~ block whose body mentions ~~~, in both the seed and the diff walk.
- The distinct-stance merge case from pre-merge run 33519246857 (overlap 0.93, different sentences): both records survive.
- Source-level lock that no pre-step fetches `gh pr diff --patch`.

**Needs a live run**
- The next pre-merge review on any docs PR exercises the net-diff fetch and the seeding against a real checkout. Nothing here is workflow-gated.

Note for whoever merges: #21312 is stacked on this branch and should be retargeted to `master` once this lands.

---

PR #21309, opened 2026-09-01
Triage: edits inside existing code fences are not trivial

CamSoper:
`triage-classify.py`'s `has_code_block_change` only fired when a *changed line was itself a fence marker*. Ten lines of Java/Go/TypeScript edits inside existing fences in one docs file therefore classified `review:trivial`, and the review was skipped entirely. Found on master on 2026-09-01 while tracing the PR #21291 recall gaps.

## What changed

Ported from commit 72a3c34d710 on `CamSoper/review-v3`, detection only (master has no `classify_mechanical`, and that stays on the v3 branch):

- `_CODE_LINE_RE` / `_looks_like_code` / `_hunk_looks_like_code`: a deliberately narrow "does this line read as code" heuristic (leading declaration keywords, `{`/`;` terminators, `=>`, `:=`, comment openers). A prose line ending in `)` or containing `=` is not code.
- Per-hunk `in_fence` tracking across context lines, and `~~~` fences count too. Any `+`/`-` line inside a fence sets `has_code_block_change`.
- A hunk that opens mid-fence (no marker in view) is caught when 60%+ of its non-blank lines are code-shaped (minimum 3 lines).

The heuristic fails toward "code" on purpose. A wrong call in that direction costs one review run; the wrong call in the other direction skipped the review. That includes a prose edit whose first context line is a closing fence -- it reads as an opener and the PR gets reviewed. The docstring says so.

## Test

`test_code_inside_existing_fence_is_not_trivial` in `test_triage_classify.py`, against `classify_pr(...)["trivial"]` the way the workflow consumes it:

- fence marker visible in context, Java edit inside -> not trivial
- hunk starting mid-fence with Java-shaped lines, no marker in view -> not trivial
- prose-only hunk with a parenthesis at a line end -> still trivial

`make test-review-pipeline` and `make lint` are green. The live check is the next docs PR that edits only code inside existing fences: it should come out of triage without `review:trivial`.

---

### DIFF
--- a/.claude/commands/docs-review/scripts/test_triage_classify.py
+++ b/.claude/commands/docs-review/scripts/test_triage_classify.py
@@ -185,8 +185,24 @@ def test_code_inside_existing_fence_is_not_trivial() -> None:
     before = len(_failures)
     path = "content/docs/iac/concepts/x.md"
 
-    # Fence marker visible as context: the edit is inside the block.
+    # Fence marker visible as context, PROSE-shaped body: only the marker can
+    # carry the signal, so this pins the fence-tracking half of the fix on
+    # its own (the review of PR #21309 noted a Java-bodied fixture here was
+    # also caught by the shape heuristic, leaving fence tracking untested).
+    # It inverts cleanly against the `prose` fixture below, which is the same
+    # shape with no marker and is asserted trivial.
     with_marker = [
+        " ```text",
+        " Set the region before you deploy.",
+        "-Use us-west-2 for the walkthrough.",
+        "+Use us-east-1 for the walkthrough.",
+        " Then run the deploy command.",
+    ]
+    r = run_classify(_pr(1, 1, [path]), _md_diff(path, with_marker))
+    check(r["trivial"] is False, f"a prose edit under a visible fence marker is not trivial; got {r}")
+
+    # Both signals at once: marker visible AND code-shaped body.
+    with_marker_code = [
         " ```java",
         " class AwsS3Website extends ComponentResource {",
         "-    public AwsS3Website(String name) {",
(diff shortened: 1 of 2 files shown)


CamSoper:
@claude #update-review

Fixed L198 in b3ffd5a: the visible-marker fixture is now prose under a ```text marker (the shape you proposed), so the marker is the only thing that can flip the classification, and it inverts against the marker-less `prose` fixture. The Java-bodied case stays as a separate both-signals check. I verified the isolation by removing the `or in_fence` clause locally: the prose fixture then classifies trivial, so the assertion now fails without fence tracking.

---

CamSoper:
Review status: one finding raised (L198, the fixture didn't isolate fence tracking), fixed in b3ffd5a and confirmed resolved by the refreshed review. Zero open.

**Verified offline**
- `make test-review-pipeline` and `make lint` green on b3ffd5a.
- The prose-bodied `with_marker` fixture fails when the `or in_fence` clause is removed from `triage-classify.py` and passes with it, so fence tracking is pinned independently of the shape heuristic; the Java-bodied case remains as the both-signals check.

**Needs a live run**
- The next docs PR whose only change is code inside existing fences should leave triage without `review:trivial`. Nothing else here is workflow-gated.

---

CamSoper:
CI note: `Install deps and build site` failed on b3ffd5a, but not on anything in this PR. The job died in `aws-actions/configure-aws-credentials` before Hugo ran -- `Could not assume role with OIDC: connect ETIMEDOUT` against `sts.us-west-2` on all 12 retries between 23:59 and 01:21 UTC ([job log](https://github.com/pulumi/docs/actions/runs/33573319196/job/100071774773)). This PR touches two Python files under `.claude/commands/docs-review/scripts/` and nothing the build or the AWS step reads; the same job passed on the previous head (8676a30) and on the sibling PRs #21308 and #21310 during the same window. There is no fix to port; it's an STS connectivity outage on the runner side.

I don't have a way to re-run the job from this session. A re-run of the failed job should go green; if it fails the same way twice, that's a runner/network problem worth a look, not this PR's.

---

PR #21310, opened 2026-09-01
Glow-up lane: reconcile the banked backlog against the fresh run before the model sees it

CamSoper:
Two glow-up PRs a human reviewer rejected on 2026-09-01 executed banked findings the fresh run had already overruled. Nothing compared the bank against the fresh artifacts; the model got both, fused, and did what the older one said.

## What went wrong (from the workflow artifacts)

**PR #21291** (`convert-hcl.md`, run 33518039058, review comment r3908440407). The backlog carried `pr20004-findings-4`: a July `contradicted (medium)` verdict on "Using the Pulumi MCP server is the recommended approach", fused with the July agent's decline reason, which contained the aside "the page frames `pulumi convert` as primary". The September run re-verdicted the same sentence `not-a-claim (high)` -- as c37 at L1228, the id having drifted from c23 and the line from L1115. The glow-up executed the July item anyway and promoted the aside into a new "fastest path" claim. Same run: c7 and c32 were fresh `contradicted` verdicts and never surfaced, because `compose_glowup()` wrote only a verdict count.

**PR #21293** (`providers/_index.md`, run 33518035360 -- the handoff cited 33518021209, which turned out to be the yaml-component-reference fix run). The backlog carried `pr20503-findings-5`, a July readthrough "self-redundancy" finding the July agent had declined as editorial. The fresh readthrough pass raised three findings, none of them this one (its self-redundancy is a different one, seven lines away). The glow-up executed it and collapsed the parameterized-provider definition.

## What changed

- **Reconcile before the model sees it** (`compose-pr-body.py`). Every banked item is matched to this run's `.verified-claims.json` / `.readthrough-findings.json` by claim-text token similarity (reusing `merge-claims.py`'s scorer, threshold 0.5) inside a 200-line window -- never by id or exact line. Each item gets a `fresh_verdict` stamp. A claim re-verdicted `not-a-claim`/`verified`/`matches`, or a readthrough finding the fresh pass did not re-raise, is pre-declined "superseded by re-verification" and leaves the work list. Fresh `contradicted`/`mismatch` verdicts are stubbed as work rows (`fresh-<claim id>`), shaped like the fix lane's "Fixes applied" stubs. The composer writes the reconciled backlog back in place (`--reconciled-backlog`) so the snapshot, the publish gate and the findings record all see the same list.
[...]

PR #21312, opened 2026-09-01
Pre-merge review: editorial stances get a home, and rewrites cite what they borrow

CamSoper:
## Why

`positioning` / `comparison` claims ("the fastest path", "the recommended approach", "unlike Terraform") were extracted, sent to the verifier, landed `not-a-claim` every time (a page's own framing has no external ground truth -- the July 2026 hard rules are right about that), and that verdict was then banked as a finding for the fix and glow-up lanes to act on. PR #20004's `contradicted (medium)` on "the recommended approach" is the item PR #21291 executed. Meanwhile an agent-written "`pulumi convert` is the fastest path ... and it's where to start" had no place in the pre-merge review at all: no trail line, no bucket, nothing for a human to notice.

## What changed

- **Routing** (`merge-claims.py` schema v2). Positioning/comparison records are written to a separate `stances` list. `claims` stays the verifier's input and the `candidate-claims-coverage` floor. `verify-claims.py` also filters the two types defensively, so an older artifact can't produce a verdict on one.
- **Surface** (`compose-review.py`). A `#### Editorial stances introduced by this PR` block under ⚠️ Low-confidence, before `#### Style suggestions`: one `- L \`file\` -- *"text"* -- positioning (found by regex)` row per record. No verdict emoji, no verdict word, no trail record, not counted in the ⚠️ cell (the rows start with `- L`, not `- **[L`, so `extract_bucket_bullets` skips them). A pre-v2 artifact renders nothing; an empty list renders the explicit-empty form so "checked, none" is distinguishable from "didn't check".
- **Validator** (`validate-pinned.py` schema v21). `editorial-stances-coverage` mirrors `candidate-claims-coverage`: every record has a row whose line ref overlaps it, every row answers to a record, and no row carries a verdict marker. Non-surgical; it surfaces the gap loudly.
- **Never banked** (`build-glowup-backlog.py`). A stance row recovered from an older PR body is dropped with a log line. Only the finding half of the row is tested against the extractor's own `POSITIONING_RES`/`COMPARISON_RES` (imported by path), never the reviewer's reason -- "the page frames X as primary" in a decline reason must not turn the finding into a stance.
- **Provenance on rewrites** (`output-format.md`, `shared-criteria.md`). When a suggestion block or proposed rewrite draws wording from another page, cite it inline ("per `any-terraform-provider.md` L23-24"). Cross-sibling consistency is a mismatch detector, not ground truth. PR #21293's `@claude implement the suggested rewrite` applied a review rewrite verbatim (5ab96c0) and inherited two clauses from the sibling, one drifted.
- **`@claude implement` guidance** (`update.md` Case 1). Re-check the implemented rewrite against the sibling pages the finding cited before pushing; confirm each borrowed clause is true of this page's subject. This lives in the skill and nowhere else. An earlier revision of this PR also injected an abbreviated copy into `claude.yml` as an `--append-system-prompt` line; that is reverted in f5065277. Guidance for a review lane belongs where the lane is maintained, and `claude.yml` was the wrong workflow regardless: it is the ad-hoc chat handler, and its own condition skips any comment containing `#update-review` or `#new-review`, so the flag was absent from precisely the runs that act on a review.
- **One extractor fix**: `COMPARISON_RES` only matched lowercase `unlike`, so the common sentence-initial "Unlike Terraform, ..." was never extracted. Now case-insensitive on the keyword only (the proper noun after it still has to be capitalized).
- `claim-extraction.md` and `fact-check.md` say the routing plainly: stances are extracted and surfaced, not verified.

## Fixture

`test_editorial_stances.py` (standalone and pytest) runs a diff introducing "`pulumi convert` is the fastest path for most configurations, and it's where to start" through `extract-claims.py` -> `merge-claims.py` and asserts: no positioning/comparison record in `claims`, exactly one positioning stance in `stances` with its anchor and provenance, the line's url record still reaches the verifier, the composed ⚠️ block lists it with no verdict marker and the bucket-bullet counter ignores it, and the validator passes on a faithful body and fails on a dropped block, a row moved off its record, and a row given a verdict. `test_extract_claims.py`'s distinct-stance merge test now reads `stances`. `build-glowup-backlog.py --self-test` covers the never-bank filter and the reason-vs-finding boundary.

## Merge history

Opened stacked on #21308, retargeted to `master` once that landed. Two merges of `master` since, one conflict each:

- `merge-claims.py` / `test_extract_claims.py` (after #21308): kept this branch's `STANCE_TYPES`, which already folds in master's rationale comment, and master's fixture materialization plus this branch's schema-v2 stance assertions. Also dropped `testdata/pr21291-head/`, which the merge had kept alongside master's renamed payload.
- `build-glowup-backlog.py` (after #21310): master's finding/prior-disposition self-test checks and this branch's editorial-stance checks landed at the same insertion point. They test different things, so both are kept.

## Verified offline vs. still needs a live run

- Offline: `make test-review-pipeline` and `make lint` green on this branch, after both merges.
- Live: the next docs PR with a superlative in its added lines is the check that the block renders and the validator accepts the composed form. This PR's own pinned review has already rendered the block and passed the validator, so the composer/validator pair is exercised in CI and not only in the fixture.

Not in scope: `compose_v3`, `scripts/review-v3/`, `apply-update.py`, `build-evidence.py`.

---


CamSoper:
@claude #update-review

All three fixed in 2cfc716:

- L1615: the verdict-marker scan now reads only the bullet's metadata tail after the closing `*` of the quoted text (`ln.rsplit("*", 1)[-1]`, whole line when there is no italic span). New fixture in `test_editorial_stances.py`: a stance whose text is `| Unlike Terraform | ✅ | ❌ |` passes the rule.
- L1704: the unreachable `if not stances and bullets: pass` branch is gone; its note now sits on the unbacked-bullet loop that already covers the case.
- L473: `safe_main()`'s crash-fallback `meta` carries `"stances": 0` like the success path; a source-level check in the test locks it.

For the record, the six rows under "Editorial stances introduced by this PR" on this review are the feature doing its job on its own reference docs: the extractor matched the example phrases the taxonomy table quotes ("the recommended approach", "unlike Terraform", ...). They're accepted as-is; no page copy changes.

---

CamSoper:
Review status: three findings raised, all fixed in 2cfc716 and confirmed resolved by the refreshed review. Zero open. The six "Editorial stances" rows are the extractor reading its own taxonomy examples; accepted as-is.

**Verified offline**
- `make test-review-pipeline` and `make lint` green on this branch (which carries #21308).
- `test_editorial_stances.py`: the "fastest path" diff yields one positioning record in `stances` and none in `claims`; the composer lists it with no verdict marker and the bucket-bullet counter ignores it; the validator passes on a faithful body and fails on a dropped block, a row moved off its record, and a verdicted row; a `| Unlike Terraform | ✅ | ❌ |` stance passes.
- `build-glowup-backlog.py --self-test`: a stance row is never banked; a stance word inside the reviewer's reason does not make the finding a stance.

**Verified live, incidentally**
- This PR's own pinned review rendered the new "Editorial stances introduced by this PR" block and the validator accepted it, so the composer/validator pair works in CI, not just in the fixture.

**Needs a live run**
- The `claude.yml` `--append-system-prompt` line only exercises on the next `@claude implement ...` mention.

Merge order: this PR's base is `CamSoper/claim-extractor-recall` (#21308). Retarget to `master` after that merges.

---

CamSoper:
@claude #update-review

The pinned review above is stale (last updated 00:11:30Z, three pushes ago) and still describes a `claude.yml` change that no longer exists. Nothing outstanding was left open; what changed is scope and base.

- **f5065277** removes the `--append-system-prompt` line from `.github/workflows/claude.yml`, reverting that file to `master` byte for byte. The rule it carried was already written in `docs-review/references/update.md` Case 1, so the workflow copy was duplicated guidance in the wrong place. `claude.yml` is also the wrong workflow for it: its own trigger condition excludes any comment containing `#update-review` or `#new-review`, so the flag was absent from exactly the runs that act on a review. The rule is unchanged in `update.md`.
- **143d8516** merges `master` after #21308 landed, and the PR is retargeted from `CamSoper/claim-extractor-recall` to `master`. Conflicts in `merge-claims.py` and `test_extract_claims.py`: kept this branch's `STANCE_TYPES` (it already folds in master's rationale comment, so taking master's would define it twice), took master's fixture materialization, kept this branch's schema-v2 stance assertions. The merge had also kept `testdata/pr21291-head/` alongside master's renamed payload, which would have re-registered the 1288-line article as an agent skill; removed.
- **0be8c6e8** merges `master` after #21309 and #21310 landed. One conflict in `build-glowup-backlog.py`'s self-test, where master's finding/prior-disposition checks and this branch's editorial-stance checks landed at the same insertion point. Both kept, master's block verbatim and in place.

`make test-review-pipeline` and `make lint` are green after both merges, and the PR body is updated to match. The six "Editorial stances introduced by this PR" rows are still the extractor matching the example phrases its own taxonomy table quotes, accepted as-is with no page-copy change.

---

Issue #21329, opened 2026-09-02
Dedupe Any Terraform Provider content between the concepts page and the get-started page


jkodroff:
## Problem

Two pages document the Any Terraform Provider feature, and they overlap substantially:

- `content/docs/iac/concepts/providers/any-terraform-provider.md` → [/docs/iac/concepts/providers/any-terraform-provider/](https://www.pulumi.com/docs/iac/concepts/providers/any-terraform-provider/) — the canonical page, just rewritten in #TBD.
- `content/docs/iac/get-started/terraform/terraform-providers.md` → [/docs/iac/get-started/terraform/terraform-providers/](https://www.pulumi.com/docs/iac/get-started/terraform/terraform-providers/) — a thinner treatment of the same feature.

The get-started page is the one that holds the legacy alias `/docs/iac/using-pulumi/pulumi-packages/terraform-provider/`, so it inherits inbound traffic that arguably belongs on the concepts page.

Several pages also link to the get-started page when they mean the concepts page. Two were fixed in the rewrite PR (`local-sdks.md`, `providers/_index.md`); `content/docs/iac/comparisons/cdktf.md:62` still points at the get-started page for on-demand SDK generation.

## Suggested work

1. Decide which page owns the feature. The concepts page is the natural canonical home; the get-started page should stay a short quickstart within the Terraform get-started path and link out rather than restate.
2. Trim the duplicated explanation from the get-started page.
3. Audit remaining inbound links (`grep -rn "get-started/terraform/terraform-providers" content/`) and repoint the ones that mean "explain this feature."
4. Consider whether the legacy alias should move to the concepts page (with the usual alias/redirect care).

## Context

Split out of the Any Terraform Provider page rewrite so that PR stays reviewable.

jkodroff:
This has been fixed.

PR #21330, opened 2026-09-02
docs: refocus the Any Terraform Provider page on the capability

jkodroff:
Reorients the page around the Any Terraform Provider capability and the providers it makes available, rather than around local SDKs.

- **Positioning.** Any TF is for when there's no bridged Pulumi provider and there is a Terraform/OpenTofu one. Dropped "evaluating a provider before committing to it."
- **Languages.** Every language uses `pulumi package add`; YAML just doesn't get a generated SDK. HCL doesn't need `pulumi package add` at all.
- **`pulumi install`.** Removed from the walkthrough, since `pulumi package add` already generates and links the SDK. Kept for the clone/CI case, where it installs whatever is missing: provider binaries and SDKs.
- **Config and property names.** Plainer language: provider config keys are always camelCase; resource inputs and outputs are cased for your language. Removed the `pulumi package get-schema` "discovering names" procedure.
- **Local SDKs.** Delegated to the [Local SDKs](https://www.pulumi.com/docs/iac/guides/building-extending/packages/local-sdks/) page instead of restating it.
- **Registry.** Rewritten around discoverability and the badge on Any TF packages.
- **Removed** the Limitations and Best practices sections.

Also repoints two links that went to the get-started page or the Registry package instead of this page. Deduping the get-started page is #21329.

https://claude.ai/code/session_01A8PbULLXfvPBb3TMr1fh4h


jkodroff:
@claude #update-review

Addressed all five findings in 401a810.

**Fixed:**

- **L23** (🚨) — you were right that the intro contradicted the callout six lines
  below it. Now reads: "The Any Terraform Provider feature works with every Pulumi
  language. In every language except Pulumi HCL, `pulumi package add` is how you add
  a provider; what then differs between them is whether you also get a generated SDK."
- **L41** — softened to "mirrors its providers, so in practice a provider published
  to either one is available." The overlap between the two registries is large, but
  the page no longer asserts an identical catalog.
- **L304** — scoped to provider configuration keys: "Pulumi renders provider
  configuration keys in camelCase instead." That stops a reader who stops after the
  first paragraph from carrying away a rule the page contradicts four lines later.
- **L320** — added one sentence pointing at the authoritative answer: "To see a
  provider's exact configuration keys, run `pulumi package get-schema` and inspect
  `config.variables`." Deliberately not restoring the removed "discovering names"
  jq subsection, since plainer language was the point of the rewrite.

**Disputing L338** (the Registry badge): the badge does ship in Registry search
today — confirmed by the page author. Note also that the sentence describes what the
badge conveys rather than quoting a literal label, so it stays accurate even if the
label's exact wording differs. Please re-check with that in view.

For the record on the gate: the auto-refresh declined with `fire=false — hunk
content/docs/iac/concepts/providers/any-terraform-provider.md:38-44 is outside
outstanding finding lines`, since it scopes only against the single outstanding
finding at L23. Hence the explicit mention.

CamSoper (approved):
This makes perfect sense, huge improvement.

PR #21340, opened 2026-09-02
Docs review v3: deterministic machinery and the two-card composer surface (dark)

CamSoper:
First slice of the docs-review v3 rollout (`REVIEW_V3_COMMENTS` / `REVIEW_V3_SENTINEL` stay unset, so nothing a contributor sees changes).

**What lands**

- `scripts/review-v3/`: routing config + parser, `REVIEW_STATE` and `/resolve` contracts, the evidence trio (validate / record / render), the Sentinel evaluator, the SLA sweep, the mechanical-bar replay tool, and their tests. `README.md` there is the map.
- `.claude/commands/docs-review`: `compose-review.py --surface v3` (author card + reviewer brief + evidence base), `build-evidence.py`, `apply-update.py`, validate-pinned schema v23 with the v3 rules, the tightened mechanical bar in `triage-classify.py`, `pinned-comment.sh --role`, the consumer scripts reading the v3 shape, references, footers, fixtures.
- Four new workflow files, all inert until their variable is set: `review-sentinel.yml` (tri-state `REVIEW_V3_SENTINEL`: unset = never runs, `report`, `1`), `review-resolve.yml`, `review-sla-sweep.yml`, `staging-deploy-pr.yml`.
- Harness + CI path-filter wiring.

**Why dark works:** the default composer surface is golden-pinned byte-identical; no existing workflow calls the new scripts; every new workflow job has an explicit `if:` on a literal variable value (unset is never truthy).

**Verification:** `make test-review-pipeline` (45 suites) and `make lint` green. Fork battery on CamSoper/pulumi.docs: waves A–F in pulumi/docs-review-benchmarks#9 (flag-off control identical to master; v3 initial / update / infra rounds; dark-state probe with every gated job skipped).

Stacked: the lane rewiring and the downstream consumers follow as separate PRs based on this branch.


CamSoper:
@claude #new-review

CamSoper:
@claude #update-review

Fixed the two blockers; the low-confidence question is already covered by a test.

- L116: the step now stamps `SINCE` before dispatching and selects the first `workflow_dispatch` run on the branch created after it, polling up to 60 s, erroring if none appears. A push-triggered or earlier `/deploy-staging` run can no longer be watched by mistake.
- L140: the `cancelled()`-guarded step is gone (a displaced pending run never starts, so it could never fire). Instead, when a deploy is already in progress the job posts an up-front "queued behind <url>" comment that says a newer request would drop this one silently, and the header note describes the real queue behaviour.
- L233: yes — `test_triage_classify_mechanical.py::test_fence_parity_seeded_from_file` reproduces exactly that shape (a hunk whose leading context line is the closing fence of the block above) and asserts the diff-only tracker calls it code while the file-seeded pass does not, plus the mid-fence and wrong-checkout fallback cases. The helper's docstring now points at the test.

Also in this push, at its source: `scrape-review-outcomes.py`'s `bulk_accepted` counts only bulk answers that classified as `author_accepted` (its documented meaning), which is the fix for the >100% rate #21342's review found in the digest.

CamSoper:
@claude #update-review

All four fixed:

- L23-30: you're right that every job in this workflow sits in the `staging-stack` group, so no notice can fire from inside it. Dropped the busy check. The header note now states that a displaced request is silent by design, and the 🚀 "started" comment (the one thing a requester reliably sees when their run does start) says a request that never gets that comment was displaced and should be re-run once the current deploy finishes.
- L129: `SINCE` is stamped 2 seconds back, keeping "nothing created before this job" while not excluding the run by its own second.
- L138: `if [ -n "$RUN_ID" ]; then break; fi`, so the timeout diagnostic runs.
- L602: the fixture's F1 is now a bulk `fixed`; the test asserts `bulk_accepted` stays 1 (the old counter would say 2).

Also in this push: the `/resolve` listener honours the new `surface:v3` label (the initial lane derives the surface from `REVIEW_V3_COMMENTS` OR that label, in #21341), plus the label's row in `labels-pr-review.md` and a sentence in the README's rollout section.

CamSoper:
@claude #update-review

L51: fair — the row now says which half ships where: the `/resolve` listener reads the label in this PR, the initial review lane in the lane-rewiring PR stacked on it, and until that merges the label changes nothing on a fresh review.

CamSoper:
@claude #update-review

One follow-up from #21344's re-review that lives here: `test_sentinel.py`'s standalone harness now binds its test list inside `run_standalone()` (the same latent hazard the SLA sweep's harness had), skips fixture-taking tests by name, and counts a crash as a failure. No behaviour change under pytest.

PR #21342, opened 2026-09-02
Docs review v3: downstream consumers read the evidence record

CamSoper:
Third slice of the docs-review v3 rollout, stacked on #21340. Every consumer prefers the v3 evidence object when a PR has one and falls back to today's shape otherwise, so nothing changes until the surface flips.

- **weekly-digest**: `collect_v3_ops()` (waive rate, SLA breaches by lane, bulk-accept rate) and the synthesis prompt's v3 section.
- **review-admin**: a `pr-review` domain loader for the evidence ledger.
- **content-review**: the glow-up backlog gains a source — a review PR's still-open, accepted-as-is, or held findings and every pre-existing row from `pr-review/<pr>/latest.json`. `select-glowup.py --pr-review-dir` stamps trimmed records on the queue article (the worker has no AWS), and the dispatcher syncs only `*/latest.json`.

Verification: harness + lint green on the stack; builder self-test (77 checks) and selector test (68) cover the new source, dedupe, and the queue stamp.

### DIFF
--- a/.claude/commands/review-existing-content/SKILL.md
+++ b/.claude/commands/review-existing-content/SKILL.md
@@ -458,7 +458,10 @@ except as amended here.
 `compose-pr-body.py`): the ledger's `skipped_findings` / `clarity_flag`
 counters plus every banked finding extracted from the page's prior review
 PRs' "Findings not applied", "Screenshot check", and "Rendered content"
-sections, each with a stable `id` and its `source_pr`. The pre-step
+sections, each with a stable `id` and its `source_pr` — plus, for PRs
+reviewed on the v3 surface, what the pre-merge *reviewer* found and the page
+still carries (`source: pr-review`: findings left open, accepted as-is, or
+held over a dispute, and every pre-existing issue it filed). The pre-step
 artifacts (claims, Vale, readthrough, frontmatter) are also present and are
 your evidence base.
 
--- a/.github/workflows/content-review-article.yml
+++ b/.github/workflows/content-review-article.yml
@@ -997,15 +997,32 @@ jobs:
             python3 scripts/content-review/compose-pr-body.py \
               --replace-notice judgment --body-file .pr-body-draft.md
           fi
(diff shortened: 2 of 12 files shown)


CamSoper:
@claude #update-review

Head moved once under the first review (rebased onto the trimmed machinery branch). Fixed all three:

- L306: `weekly-digest.yml` now assumes the ContinuousDelivery role (best effort, `continue-on-error`) before collecting, so `collect_v3_ops()` can discover and sync the ledger bucket.
- L313: `pr-review/state/` is no longer synced; only `runs/` is load-bearing, `waives/` stays optional.
- L363: fixed at the source rather than clamped — in the scraper (machinery PR #21340, same stack), `bulk_accepted` now counts only bulk answers that classified as `author_accepted`, which is the meaning its key docs already state, so the rate nests and can't exceed 100%.

CamSoper:
@claude #update-review

One addition since the last round: while the `REVIEW_V3_BOT_PRS` repo variable is `1`, the content-review and glow-up publish step opens its PRs with `--label surface:v3` (on `gh pr create` itself, since the initial lane reads labels off the `opened` event), retrying unlabeled if the label is missing. Bot review PRs become v3 test traffic without any human author seeing a change.

CamSoper:
@claude #update-review

- L1002: the step comment and the README now say the initial lane reads the label once the lane-rewiring PR (stacked on #21340) has merged, and that until then it only reaches the `/resolve` listener.
- L1009: both now state the order of operations — run the `gh label create` line for `surface:v3` first, then flip `REVIEW_V3_BOT_PRS` — and what happens if not (unlabeled fallback with a `::warning::`).

CamSoper:
@claude #update-review

README only: the rollout section is now byte-identical to #21344's copy (same wrapping), so whichever of the two merges second is a clean merge — verified with `git merge-tree` in both orders. The `REVIEW_V3_BOT_PRS` sentences are unchanged; the SLA paragraph is #21344's text carried verbatim for that reason.

PR #21341, opened 2026-09-02
Docs review v3: wire the review lanes to the two-card surface (behind REVIEW_V3_COMMENTS)

CamSoper:
Second slice of the docs-review v3 rollout, stacked on #21340. With `REVIEW_V3_COMMENTS` unset every lane behaves exactly as it does today; the flag flip is the user-facing event, not this merge.

- **`claude-code-review.yml`**: behind the flag, the model edits the author card + reviewer brief drafts, and one credentialed publish job validates both cards, builds and records the evidence object (S3 under the ledger scripts' degradation contract, workflow artifact always), upserts brief-then-author, labels, and pokes the Sentinel while `REVIEW_V3_SENTINEL` is set.
- **`claude-update.yml`**: the compose-based update lane — a structured patch from the model, `apply-update.py` re-renders both cards and merges `REVIEW_STATE` against the live card.
- **`claude-triage.yml`**: `synchronize` re-classification and matrix reviewer routing behind `REVIEW_V3_ROUTING`.
- CONTRIBUTING / AGENTS paragraphs describing the flagged surface.

Verification: harness + lint green on the stack; fork battery waves A–F (pulumi/docs-review-benchmarks#9) ran these exact workflow files with the flags on and off.

PR #21343, opened 2026-09-02
infrastructure: public review-evidence bucket for the v3 evidence pages

CamSoper:
The one piece of the docs-review v3 rollout that touches infrastructure: a stable public static-website bucket (modeled on the PR preview buckets, never deleted) that the v3 publish job writes `/<pr>/<sha>.html` and `/<pr>/latest.html` into.

Exposure is unchanged from today: this is the verification trail that currently ships inside the pinned PR comment. Kept separate from the code PRs (#21340, #21341, #21342) because merging it implies a stack update, which is a deliberate step — until the bucket exists, the publish job degrades to uploading the HTML as a workflow artifact.

Type-checks; not deployed.

### DIFF
--- a/infrastructure/index.ts
+++ b/infrastructure/index.ts
@@ -354,6 +354,60 @@ new aws.s3.BucketPublicAccessBlock("content-review-ledger-public-access-block", 
     restrictPublicBuckets: true,
 });
 
+// Bucket for the pre-merge review evidence pages: one self-contained HTML
+// page per (PR, head SHA) plus a latest.html pointer, rendered by
+// scripts/review-v3/render-evidence-html.py and linked from the two pinned
+// review comments. Public by design — the content is the same verification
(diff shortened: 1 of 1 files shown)


CamSoper:
@claude #update-review

Fixed all four:

- L368: dropped the static-website configuration and the `reviewEvidenceWebsiteEndpoint` export. The stack now exports `reviewEvidenceBaseUrl` (the bucket's HTTPS regional REST URL) and both review lanes link `<base>/<pr>/latest.html` (updated in #21341 and on the source branch). The TLS-only statement stays.
- L383: the public-read policy now `dependsOn` the public-access block, same ordering as the uploads bucket.
- L366: added `BucketVersioning`; the comment now states why.
- L357: the comment notes the renderer path lands with the machinery PR (#21340).

PR #21344, opened 2026-09-02
Docs review v3: SLA sweep (dark behind REVIEW_V3_SLA)

CamSoper:
Phase 5 machinery of the docs-review v3 rollout, split out of #21340 to keep that PR inside the 15k-line review budget. Stacked on the machinery branch.

`scripts/review-v3/sla-sweep.py` + `review-sla-sweep.yml`: clocks derived fresh each sweep from the GitHub timeline (reviewer-stalled → one escalation per epoch; author-stalled → warn at 14 days, close at 21 only if warned ≥ 7 days earlier; legacy PRs never closed). The job runs only while `REVIEW_V3_SLA == '1'`; a manual dispatch defaults to `dry_run`. Merging changes nothing.

Verification: `--self-test` + pytest (business-day math, canned timelines, idempotency, never-close-unclassifiable); harness + lint green on the stack.

### DIFF
--- a/.github/workflows/review-sla-sweep.yml
+++ b/.github/workflows/review-sla-sweep.yml
@@ -110,7 +110,7 @@ jobs:
           if [ -n "$BUCKET" ]; then
             echo "evidence_uri=s3://$BUCKET/pr-review" >> "$GITHUB_OUTPUT"
           else
-            echo "::warning::state bucket output unavailable; sla-sweep.py will degrade to local-only state"
+            echo "::warning::state bucket output unavailable; sla-sweep.py forces DRY-RUN this cycle (mutations without durable state would repeat every sweep)"
           fi
 
       # Comment-noise safety: a manual dispatch always honors its own
--- a/.github/workflows/review-sla-sweep.yml
+++ b/.github/workflows/review-sla-sweep.yml
@@ -1,0 +1,147 @@
+name: Review SLA sweep
+
+# The v3 review workflow's clocks: machine-enforced reviewer-wait SLAs
+# (one-shot escalation) and author-staleness warn/close. scripts/review-v3/
+# sla-sweep.py is the deterministic evaluator (no model, no PR code
+# checkout — it reads PR state through the GitHub API only); this workflow
+# gathers credentials, resolves the S3 evidence-layer bucket, and runs it.
+# See scripts/review-v3/README.md and .github/review-routing.yml (the `sla:`
+# and `author_staleness:` blocks) for the policy this enforces.
+#
+# SINGLE SWITCH — the REVIEW_V3_SLA repo variable gates BOTH whether the
+# scheduled job runs at all AND whether it's allowed to mutate anything:
+#   unset / anything but '1' -> the schedule no-op-skips the job entirely
+#     (job-level `if:`), so an unconfigured repo spends nothing here.
+#   '1' -> the schedule runs for real (comments, labels, closes, escalates).
+# workflow_dispatch always runs regardless of the variable (the testing
+# path), and always honors its own `dry_run` input rather than the
+# variable — a manual run never needs the switch flipped to preview safely.
+#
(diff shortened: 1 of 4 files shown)


CamSoper:
@claude #update-review

Fixed all five:

- L477: `classify_waiting_state` now carries the newest `CHANGES_REQUESTED` timestamp out in `extra`; `process_author_time` clocks idle time from `max(last author activity, that review)`, and the nudge says "a reviewer has requested changes" when there is no finding count. New test covers the CHANGES_REQUESTED-only path (2 idle days, not 20; then the warn fires 13 days later with the right copy).
- L328: committer date first, author date as fallback, with a test.
- L483: `process_author_time` / `process_reviewer_time` run inside the same per-PR `except Exception` guard the reads use — one `gh` failure skips that PR (state unsaved, error in the run record) and the rest of the sweep continues; test with a flaky `add_label`. The README now says to create `review:author-stalled` before flipping the switch.
- L618: the stale-warn cleanup also runs on the `kind is None` branch, so the label can't leak on a PR that resolved to mechanical; test added.
- L104: README paragraph on `REVIEW_V3_SLA`, the `pr-review/state/` and `pr-review/runs/` prefixes, and the label step, right beside the Sentinel switch.

CamSoper:
@claude #update-review

Both new items fixed:

- L504: label add/remove now go through `_try_label` (best-effort, logged); comments and closes stay fatal for that PR only. The warn record is therefore always persisted alongside the comment that earned it. `test_mutation_failure_does_not_abort_the_sweep` now asserts exactly one comment on PR 1 across two consecutive sweeps while `add_label` keeps failing (and no retry storm), and a new `test_comment_failure_is_fatal_for_that_pr_only` pins the other half.
- L117: the two sentinel sentences are back at the end of the `REVIEW_V3_SENTINEL` paragraph; the SLA paragraph now ends at the label step.

CamSoper:
@claude #update-review

L746 fixed: `resolve_mutation_mode()` forces dry-run whenever `PR_REVIEW_EVIDENCE_URI` is unset and `--allow-local-state` (a developer-machine override, never used by the workflow) isn't given, and the run record carries `forced_dry_run` with the reason. The workflow's bucket-unresolved warning now says the cycle runs dry. Tests: the mode table, plus `main()` end to end with a stalled PR and no URI posting nothing.

CamSoper:
@claude #update-review

Both fixed:

- L433: `save_state` now returns whether the durable copy landed; the sweep writes `state_persisted: false` onto the action (with a warning) instead of the run record claiming the warn stuck. One layer earlier, `durable_state_writable()` round-trips a probe object through the state prefix before any non-dry run, and `resolve_mutation_mode` forces dry-run when the bucket resolved but the write didn't land — so a runner that can't persist never mutates in the first place. Tests: a failing upload is recorded on the action; the probe gates real runs both ways.
- L435: the standalone harness binds its list inside `run_standalone()` (it was running 12 of 19), skips fixture-taking tests by signature with a note, and counts a crash as a failure. `--self-test` now reports 19 ok + 2 skipped.

CamSoper:
@claude #update-review

L682: the no-clock branch now emits the same `warn` as the changed-PR path when its state write doesn't land, so `state_persisted` means one thing everywhere. The adjacent note about `test_sentinel.py`'s module-level test list is being fixed where that file lives, in #21340.

CamSoper:
@claude #update-review

No code change: rebased onto the current machinery tip to clear the README conflict. This branch's README now carries the full rollout paragraph as it will read once the whole stack lands (the one-sentence `REVIEW_V3_BOT_PRS` note belongs to #21342; it is here verbatim so that whichever sibling merges second merges cleanly).

CamSoper:
@claude #update-review

L115: right, and thank you for running it. Fixed on the other side: #21342's README is now byte-identical to this branch's (both equal the union text, same wrapping), and `git merge-tree` on the live objects is clean in both landing orders. No change on this branch.

PR #21345, opened 2026-09-02
Fix empty tutorial collection pages with section-only members

cnunciato:
The `/tutorials/learn-pulumi/` collection page currently renders no tutorials: 

image

This is happening because the rendering logic uses `.Pages`, which excludes `_index.md` section members once the sitemap `.RegularPagesRecursive` lastmod rollup added in #21036 evaluates on the term first — so any collection whose members are all sections (learn-pulumi) came up empty. Iterating the `collections` taxonomy object instead retains section members and orders them by `collections_weight`. 

Fortunately the pages themselves are still there, so there shouldn't be any significant impact to SEO here (since they're all already indexed and crawlable); this is just a temporary blip in reachability from overview pages.

Also restores the `esc-github` member that was silently missing from `/tutorials/pulumi-esc/` for the same reason.

### DIFF
--- a/layouts/taxonomy/collection.html
+++ b/layouts/taxonomy/collection.html
@@ -32,7 +32,15 @@
                     <div class="mt-8 mb-12">
                         <ul class="flex flex-col lg:flex-row lg:flex-wrap list-none p-0 rounded-md">
                             {{ $perRow := 3 }}
-                            {{ range $index, $item := .Pages }}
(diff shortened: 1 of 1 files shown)


CamSoper (approved):
Automated rubber stamp approval applied :shipit: -- [What is this?](https://app.notion.com/p/3adfdbdf1cce81bc946be3b4ffacd28d)

---

PR #21346, opened 2026-09-02
docs: self-hosted search supports OpenSearch 3.x

aacotroneo:
## What

The self-hosting docs told operators the search requirement was a **v2.x cluster**, "tested with OpenSearch v2.9 and v2.11". Pulumi Cloud now runs **OpenSearch v3.7**, so that guidance was both stale and needlessly narrow — it steered self-hosted users away from a version we run in production ourselves.

| | Before | After |
|---|---|---|
| `components/search.md` | `v2.x cluster` — "tested with v2.9 and v2.11" | `v2.x or v3.x cluster` — "Pulumi Cloud runs v3.7. We recommend the latest patch release of either the v2.19.x or the v3.x line." |
| `operations/architecture.md` | `OpenSearch 2.x or Elasticsearch 7.x` | `OpenSearch 2.x or 3.x, or Elasticsearch 7.x` |

Also added the supported range to the intro paragraph, where a reader first learns a cluster is needed.

## Notes

- **Recommends version *lines*, not exact patches.** Latest at time of writing is 2.19.6 and 3.8.0; naming those pins the doc to a release cadence it can't keep up with, so it points at `v2.19.x` / `v3.x` instead.
- **`network.md` deliberately untouched** — it only documents port 9200 and makes no version claim.
- The client genuinely supports both majors: Pulumi Cloud has been running 3.7 on the same OpenSearch client that self-hosted installations use, so this records demonstrated behaviour rather than widening a support promise.

### DIFF
--- a/content/docs/administration/self-hosting/components/search.md
+++ b/content/docs/administration/self-hosting/components/search.md
@@ -20,7 +20,7 @@ pulumi_cloud_feature: self-hosting
 To manage your state with a self-managed backend, such as a cloud storage bucket, see [State and Backends](/docs/iac/concepts/state-and-backends/).
 {{< /self-hosting-trial-note >}}
 
-Pulumi Cloud provides resource search features that require an OpenSearch cluster. Self-hosted installations require a compatible OpenSearch cluster and changes to the API container configuration to enable these features.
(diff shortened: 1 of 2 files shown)


CamSoper (approved):
Automated rubber stamp approval applied :shipit: -- [What is this?](https://app.notion.com/p/3adfdbdf1cce81bc946be3b4ffacd28d)

---

PR #21355, opened 2026-09-03
Add /dev CloudFront cache-invalidation workflow

cnunciato:
Receives a `repository_dispatch` (`dev-deployed`) from pulumi/marketing-web at the end of a production Dev Center deploy and invalidates just the `/dev*` paths on the www.pulumi.com distribution, which otherwise caches them for 30 minutes (`thirtyMinuteCachePolicy`). Auth scaffolding mirrors `build-and-deploy.yml`; a `workflow_dispatch` escape hatch allows manual/override invalidations.

### DIFF
--- a/.github/workflows/invalidate-dev-cache.yml
+++ b/.github/workflows/invalidate-dev-cache.yml
@@ -1,9 +1,10 @@
 name: Invalidate /dev cache
-# Triggered by pulumi/marketing-web at the end of a production Dev Center deploy.
-# The www.pulumi.com distribution proxies /dev* to marketing-web's own CDN and
-# caches it for 30 minutes (thirtyMinuteCachePolicy, see infrastructure/index.ts),
-# so a fresh Dev Center deploy is otherwise invisible here for up to half an hour.
-# This flushes just the /dev surface — nothing else on the site is touched.
+
+# Invalidates the Dev Center cache, which lives at `/dev` behind the CloudFront CDN managed by pulumi/docs. 
+# We do this with a workflow dispatch because the Dev Center deploys on its own schedule (so its cache needs
+# to be cleared independently of pulumi/docs deployments), and it makes more sense for the Dev Center to 
+# notify pulumi/docs of this event (so it can handle the event on its own), rather than for us to reach into its 
+# live, running infrastructure from the outside.
 on:
   repository_dispatch:
     types: [dev-deployed]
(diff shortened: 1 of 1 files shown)


.github/workflows/invalidate-dev-cache.yml
jkodroff, line 1: Can we explain why this lives here? (I would assume it's b/c pulumi/docs owns the entirety of pulumi.com's infrastructure.)

cnunciato: Yep, that's right. We need to invalidate the cache that's managed by pulumi/docs (as it's the forward-facing one), and that cache lives in a different AWS account than the one we deploy the Dev Center into, so we don't have the creds to invalidate it directly. 

But even so, it also makes sense for us to tell pulumi/docs, "Hey, we just deployed — can you invalidate our cache?" and then leave it up to pulumi/docs to decide whether and how to do that. (Rather than reach into its live infra directly from the outside and mess with it.)

jkodroff: Can we add a sentence to the top of the workflow to explain this?

cnunciato: Absolutely!

cnunciato: @jkodroff This work? 👇🏻

cnunciato:
>If leaving /assets* out is deliberate (e.g. those objects are fingerprinted and don't need flushing), a one-line comment saying so would keep the next reader from re-deriving it.

Exactly — they are fingerprinted, but it's also fine to clear them as well. I'll do that. (It's also worth noting that the contents of `/assets` are shared across more than just `/dev` (which is the only path served by marketing-web we expose publicly as of now), so this workflow will also expand to include more than just `/dev` as we pull more top-level content away from pulumi/docs and into pulumi/marketing-web.)

.github/workflows/invalidate-dev-cache.yml
```diff
@@ -53,8 +53,12 @@ jobs:
         env:
           PULUMI_ACCESS_TOKEN: ${{ steps.esc-secrets.outputs.PULUMI_ACCESS_TOKEN }}
           PULUMI_STACK_NAME: ${{ vars.PULUMI_STACK_NAME }}
-          # A dispatch caller may narrow/override the paths via client_payload.paths.
-          PATHS: ${{ github.event.client_payload.paths || github.event.inputs.paths || '/dev /dev/*' }}
+          # Default flushes both surfaces the distribution proxies to the Dev Center:
```
cnunciato, line 56: This was added separately to address https://github.com/pulumi/docs/pull/21355#issuecomment-5530261731.

.github/workflows/invalidate-dev-cache.yml
jkodroff: Yes.

jkodroff (approved):

PR #21377, opened 2026-09-03
deps: group security updates, fix label detection, and add gated auto-merge

CamSoper:
### Proposed changes

Cutting Dependabot PR volume, and the human toil that comes with it. **Nothing auto-merges until you set a repo variable** — the automation ships default-off.

#### The problem

`.github/dependabot.yml` was already dialed to the floor: monthly, `open-pull-requests-limit: 1`, grouped, majors ignored. It was working fine. The volume came from somewhere else.

Of the **59 Dependabot PRs opened between 2026-07-02 and 2026-09-02, only 10 came from that config.** The other **49 were security updates**, which bypass nearly everything it sets:

| Configured option | Applies to security updates? |
|---|---|
| `schedule: monthly` | No -- fires the moment an advisory lands |
| `open-pull-requests-limit: 1` | No -- "not subject to this limit and do not count toward it" |
| `groups:` | No -- groups default to `applies-to: version-updates` |

Unbatched, that's one PR per affected package per manifest: 9 on 07-23, 8 on 08-03, 11 on 09-01. `js-yaml` alone produced 9 PRs.

#### 1. `.github/dependabot.yml` — batch the security updates

- **Add an `applies-to: security-updates` group to every entry.** An advisory day now produces one PR per ecosystem instead of one per package.
- **Cover `/infrastructure/versioned-docs` and `/scripts/snippet-sweep`** — neither had an entry, so their security PRs arrived unprefixed with default naming (#20772).
- **Repoint the pip entry.** It pointed at `/scripts`, which contains no Python manifest of any kind — it had never matched anything. The real Pipfiles are `/scripts/python`, `/scripts/image-borders`, `/tools/pydocgen`.
- **Add `devcontainers`.**
- **Add `/static/programs` entries** (npm, gomod, pip, nuget, maven, docker) with `open-pull-requests-limit: 0`. These never produce version updates — `scripts/programs/upgrade.sh` owns that. They exist purely to batch security PRs GitHub already opens against those 444 sample dirs with zero config coverage today (#21304, #20771). Adding entries can't increase discovery; the dependency graph is repo-wide and independent of this file.

⚠️ The two `/static/programs` invariants (`open-pull-requests-limit: 0`, `applies-to: security-updates` on every group) are what stop this becoming monthly version updates across 444 directories. Called out in a file header comment and again at the block.

#### 2. `label-dependabot.yml` — fix the risk labels

Both risk labels were broken, and broken **specifically on grouped PRs** — the only PRs that can be bulk.

The old extractor matched `` Bumps [name] ``, but a grouped body reads `Bumps the all-dependencies group ... with 7 updates:` followed by a markdown table and `` Updates `name` from x to y `` lines. No bracket in either. It returned **zero** dependencies for every grouped PR. Verified against #21285's real body:

| | old | new |
|---|---|---|
| dependencies extracted | **0** | **7** |
| `deps-lambda-edge-risk` | not applied | applied |
| `deps-bulk-update` | not applied | applied |

**#21285 bumped `webpack` and `css-loader` and merged with neither label.** Also confirmed on #21282 ("with 10 updates") and #20711 ("with 12 updates").

Changes: source dependency names and advisory state from `dependabot/fetch-metadata`; replace the `isSecurity` substring match (207 Dependabot PRs match Dependabot's boilerplate use of the word "security"; only 52 reference a real advisory, so a label reading "merge immediately" was firing on routine bumps); extend the Lambda@Edge list to `@pulumi/`, `esbuild`, `rollup`, `terser`; lower the bulk threshold 10 → 5 (the 09-01 batch had groups of 2, 3, 7, 9, 10); guard the job on `head.repo`; fix a crash on the `workflow_dispatch` path.

#### 3. Auto-merge, behind `DEPS_AUTO_MERGE` (default off)

When the variable is `true`, the workflow approves as `pulumi-bot` and arms GitHub's native auto-merge, which waits for the build check itself. A PR qualifies **unless** it carries `deps-lambda-edge-risk` or `deps-bulk-update`.

Two things here are load-bearing:

- **Why approve at all.** Auto-merge waits for *every* branch protection requirement, not just checks, and `master` requires an approving review — an unapproved PR with a green build sits armed forever. That requirement is currently met by a routine stamping every Dependabot PR under a human's name, which manufactures a trail showing someone reviewed 59 bumps nobody read. `pulumi-bot` is a machine user with write access, so it can approve directly. A bot approving a bot-authored PR gated on CI states the real policy instead of faking a human one.
- **Why `PULUMI_BOT_TOKEN`, never `GITHUB_TOKEN`.** `build-and-deploy.yml` triggers on push to `master`, and `GITHUB_TOKEN` events don't trigger workflows. A `GITHUB_TOKEN` merge would land and never deploy — and the three daily scheduled rebuilds would hide that for hours rather than surface it.

**Not routed through `auto-approve-for-auto-merge.yml`**, despite that being the established pattern for the other 16 bot workflows. Those attach `automation/merge` *at PR creation* with a PAT, so the `opened` event already carries the label. Dependabot PRs open without it, and a label applied with `GITHUB_TOKEN` does not trigger workflows — the `labeled` event would never fire and the approve job would silently never run. Putting the label in `dependabot.yml`'s `labels:` doesn't rescue it either: that's unconditional, and the whole point is that risky PRs *don't* get approved.

### Testing

`.github` and `*.md` are both in `.prettierignore`, and `scripts/lint/lint-markdown.js` only scans `content/`, so neither linter covers any file in this diff and `make lint` wouldn't exercise it. I did not run a full `make ensure`. Instead:

**The token path is confirmed, not assumed.** A temporary probe workflow ran on this PR ([run 33815532057](https://github.com/pulumi/docs/actions/runs/33815532057)):

```
event=pull_request actor=CamSoper
RESULT: ESC token acquired on a pull_request run with no environment.
Authenticated as: pulumi-bot
```

ESC over OIDC works with no `environment:` set, and the token is `pulumi-bot`. An earlier draft of this PR said the rollout had to wait for a real Dependabot PR to prove this — **that was wrong.** The OIDC subject claim does not include the actor: a `pull_request` run presents `repo:pulumi/docs:pull_request` whichever account opened it, so this was always testable here. The probe workflow has been deleted and the finding recorded in `BUILD-AND-DEPLOY.md`.

**Label logic**, executed against mocked payloads with stubbed `github`/`context`/`core`:

| scenario | labels | `auto_merge` | comment |
|---|---|---|---|
| grouped, 7 deps incl `webpack` | bulk + lambda-edge | `false` | held |
| single safe dep | — | `true` | armed |
| single dep with GHSA | security-patch | `true` | armed |
| switch unset | — | `true` (unused) | checklist |
| 3 deps incl `@pulumi/aws` | lambda-edge | `false` | held |

Rows 1 and 5 produced **no labels at all** before this branch.

Both YAML files parse; the embedded `github-script` body passes `node --check`. The `fetch-metadata` path can only be exercised on a real Dependabot event.

### Rollout

1. Merge this. Nothing changes about merging — `DEPS_AUTO_MERGE` is unset.
2. Set `DEPS_AUTO_MERGE=true`. Watch the first Dependabot PR end to end: approval within a minute, merge ~10 min later when the build goes green.
3. Retire the rubber-stamp routine — **last**. It currently satisfies the approval requirement; removing it before step 2 means nothing merges at all.

Rollback at any point: unset the variable. Already-armed PRs release with `gh pr merge --disable-auto <n>`.

### Related issues (optional)

None.

https://claude.ai/code/session_01J3vuWejcQDE3xEGbKKC2SC


CamSoper:
@claude #update-review

Worked all six low-confidence findings. Four fixed, two verified as already correct.

**Fixed:**

**[L83]** Good catch, and worse than reported. `theme/package.json` carries `@pulumi/design-tokens`, so the `'@pulumi/'` prefix would have held every monthly theme bump as a Lambda@Edge bundle risk -- a design-token package with no bundling relevance at all. Narrowed to the two exact packages that actually get bundled, `@pulumi/aws` and `@pulumi/pulumi`. Retested against mocked payloads: a theme group carrying `design-tokens` now auto-merges, an infra group carrying `@pulumi/aws` is still held, and the 7-package webpack group is still held on both flags.

The residual over-capture you identified is accepted rather than fixed: `/infrastructure/versioned-docs` carries the same two packages and a name-only match can't distinguish it from `/infrastructure`. Failing closed is right for a merge gate. Documented in the comment above the array and in `dependabot-labels.md`, per your suggestion.

**[L187]** Confirmed. `/scripts/python` and `/scripts/image-borders` pin every package as `"*"` and ship no `Pipfile.lock`; `/tools/pydocgen` is the only one with real constraints (`sphinx >=7.1,<7.2`) and a lock. So the entry is one-third live, and you're right that it's the same failure as the `/scripts` bug this PR fixes. Rather than claim three working directories, both `dependabot.yml` and `BUILD-AND-DEPLOY.md` L3442 now say only `/tools/pydocgen` is expected to produce updates, and that committing locks for the other two is what would make them live.

**[L3496]** You're right and the sentence was wrong. Its subject was the monthly grouped PRs, which are exactly what the gates hold. Took your rewrite and added the missing half: it's the off-schedule security PRs, usually one or two packages, that mostly merge unattended.

**[L24]** Fair distinction, and I'd overstated it. The identical OIDC subject shows ESC's trust policy will match on a Dependabot-opened PR; it doesn't show GitHub mints a token at all on that path, since Dependabot-triggered runs carry their own event-level policy. Softened the comment to claim only the former, restored a token-availability report line so the first real Dependabot run settles it, and made reading that line a step before setting `DEPS_AUTO_MERGE`.

**Verified, no change:**

**[L3448]** Numbers hold. `js-yaml` returns 10 matching PRs; 9 fall inside the stated 07-02..09-02 window (#20006 is 07-01, outside it). The daily counts come from enumerating the full listing: 9 on 07-23 (#20462-20468, #20483, #20492), 8 on 08-03, 11 on 09-01.

**[L3456]** Split confirmed exactly: search `total_count` for `author:app/dependabot created:2026-07-02..2026-09-02` is **59**. Ten came from the monthly schedule (five on Aug 1, five on Sep 1), forty-nine were security updates. Also applied the snapshot framing you suggested -- the 5:1 ratio tracks advisory volume, so it isn't a constant.

**Not addressed, deliberately:** the four items under "Editorial stances introduced by this PR" are workflow comments and an internal tooling reference, not user-facing docs prose. Happy to revisit if you disagree.

---

CamSoper:
CI is red here and it is not this PR's failure. Recording why rather than leaving it looking unattended.

**Failing check:** `Install deps and build site` ([run 33922772096](https://github.com/pulumi/docs/actions/runs/33922772096/job/101202742442)), in the "Build and deploy" step.

```
error: Running program '/home/runner/work/docs/docs/infrastructure/' failed:
Error: Required output 'cloudAiAppDistributionDomain' does not exist
       on stack 'pulumi/pulumi-ai-app-infra/prod'.
```

**Cause:** `infrastructure/index.ts:110-111` holds a `StackReference` to `pulumi/pulumi-ai-app-infra/prod` and calls `requireOutput('cloudAiAppDistributionDomain')`. That output is no longer present on that stack, so the `pulumi preview` inside `make ci_pull_request` aborts. Because it is `requireOutput` rather than `getOutput`, there is no soft-fail path — the Airflow reference twelve lines below uses `getOutput` and degrades gracefully; this one does not.

**Why it isn't this PR:** the diff is four files -- `.github/dependabot.yml`, `.github/workflows/label-dependabot.yml`, `BUILD-AND-DEPLOY.md`, and `.claude/commands/pr-review/references/dependabot-labels.md`. Nothing under `infrastructure/`, and no path from any of them to a StackReference resolving against a stack in another project.

Not a flake either: the run is already `run_attempt: 2` and failed identically both times.

One caveat on the evidence, so nobody over-reads it. Every other PR run today passed, but all of them ran before 19:21 and this one ran at 23:14 -- so the pattern is equally consistent with this simply being the first build after the output disappeared. The mechanism is what rules this PR out, not the timing. The testable prediction is that **the next PR pushed to this repo will fail the same way**; if it doesn't, this diagnosis is wrong.

**No fix ported, because none exists.** I searched open issues and PRs and found nothing tracking it. The repo-side fix would be softening that `requireOutput` to a `getOutput` with a fallback, but that changes CloudFront routing behavior for the AI app in production and does not belong in a Dependabot-configuration PR. The better fix is on the other side: restoring the `cloudAiAppDistributionDomain` output on `pulumi-ai-app-infra/prod`.

Not re-running again -- the one re-run that would confirm a transient has already been spent. This PR is otherwise ready: pre-merge review is at zero open findings, and nothing in it changes merge behavior until `DEPS_AUTO_MERGE` is set.

---

CamSoper:
@claude #update-review

---

CamSoper:
@claude #update-review

Rebased onto master (now includes #21414, the dead-stack-reference fix that was breaking every build), plus one hardening commit from a pre-merge safety pass.

**Since the last review:**

- **`continue-on-error` on the `fetch-metadata` step.** That step is new in this branch, and without the guard a transient failure would have put a red check on every Dependabot PR where previously the labelling job had no way to fail. With it, outputs come back empty, `metadataAvailable` is false, and the script falls through to the body parser. Verified the degraded path still fails safe: a grouped 7-package body containing `webpack` parses to 5 dependencies and lands `deps-bulk-update` + `deps-lambda-edge-risk` with `auto_merge=false`.
- **Comment on the checkout step** recording that it is load-bearing -- `gh pr review` and `gh pr merge` infer the repository from its git remote. It reads as unused because nothing touches the working tree, and removing it would break the arming step only once `DEPS_AUTO_MERGE` is on.

**Config validation** (relevant to the earlier low-confidence items): confirmed against GitHub's Dependabot options reference that all eight `package-ecosystem` values here are valid (`npm`, `pip`, `gomod`, `nuget`, `maven`, `docker`, `github-actions`, `devcontainers`), and that no two entries share an ecosystem+directory pair. Both matter more than they look: an invalid config is the one failure mode that stops Dependabot silently rather than announcing itself.

All six low-confidence findings from the previous round remain dispositioned as described in the earlier comment -- four fixed, two verified against the repo's own history.

---

CamSoper:
@claude #update-review

All five low-confidence items fixed in `b8ce01dd`. **[L49] was correct and important** -- thank you for actually replaying the regexes against the real bodies rather than taking the commit message's word for it.

**[L49] — conceded, and the finding understates how neatly the two failures line up.** I reproduced it independently: #20711 is titled "with 12 updates", its body is exactly **65,535** characters, and the fallback recovers 10. My previous commit message claimed the degraded path "fails safe"; it does not. The parser under-counts, and because `dependencies.length` feeds *both* `isBulk` and `hasLambdaEdgeRisk`, truncation moves both inputs toward merging simultaneously. Alphabetical ordering puts `rollup`, `terser` and `webpack` -- three of the nine risk patterns -- exactly where the truncation lands.

Took the suggested fix verbatim: `autoMergeEligible = metadataAvailable && !hasLambdaEdgeRisk && !isBulk`. The fallback may label, but may never clear the merge gate. Verified: the truncated-body case now yields `auto_merge=false` with `held_reason=metadata-unavailable`, while the clean paths are unchanged (safe single dep still arms, `@pulumi/aws` still held).

You're also right that my cited evidence was wrong -- #21285 parses to 7, not 5, and since its body fits under the cap it demonstrated the clean path while I described it as the degraded one. The conclusion happened to be right for the wrong reason.

**[L144] — fixed.** Both comments corrected. Added the distinction the finding points at: the `isSecurity` under-report is tolerable because `deps-security-patch` is informational and gates nothing, whereas the dependency count gates two things, which is exactly why the fallback is now refused at the gate.

**[L353] — fixed.** The triage comment now says when metadata was missing, why the labels are less trustworthy, and that auto-merge is withheld regardless of flags -- plus a line distinguishing a permanent failure from a flaky one. Mirrors what the sibling ESC step already does.

**[L3469] — fixed.** Glossary now reads "Bundler, `@pulumi/aws` / `@pulumi/pulumi`, or AWS SDK updates", matching the matcher. (L3503/L3528 still gloss it as "webpack/bundler/AWS SDK" but those predate this PR, as you noted.)

**[L358] — fixed.** `github.actor` moved to `env: ACTOR` and referenced as `${ACTOR}`. Agreed it isn't exploitable behind the job's actor guard; taking it as a safe-by-construction consistency fix.

---

CamSoper:
@claude #update-review

All four fixed in `4135dbcd`.

**[L298] — fixed.** The held-for-review line now mirrors the `held_reason` ternary. You're right that the old `else` was actively misleading on the one case the previous commit exists to handle: it cited `deps-bulk-update` on a PR that never got the label, quoting the very count that is untrustworthy in that state, two lines below a paragraph giving the real reason. Verified all three arms: metadata-missing states the true cause, lambda-edge and bulk still name their flags.

**[L3481] — fixed in both places.** `BUILD-AND-DEPLOY.md` and the label reference each now say a PR is also withheld, with no flag set, when metadata was unavailable. Your framing of why it matters is the right one — those two files are what someone consults when a PR sits unmerged with nothing on it to explain why, which is precisely the state the new branch creates.

**[L236] — conceded, and I should have caught it.** I adopted your earlier phrasing verbatim without checking it, which is on me. Replayed against #20711: the two lost packages are **`beasties` and `simple-icons`**, at alphabetical positions 3 and 10 of 11 — not the tail. The body is cut at 65,535 characters with `_Description has been truncated_` appended, and what's lost is whatever falls after the cut in *body* order, with release notes interleaved between entries. Comment rewritten to say "anything after the cut" rather than "anything late in the alphabet", since the old model would send a reader looking in the wrong place.

One thing that fell out of checking it, recorded in the comment and deliberately not acted on: the header table rows name all 12 packages and sit near the top, so they survive truncation. Matching those would make the degraded path's *labels* accurate. I left it — the gate already refuses that path, so it's regex surface for no safety gain, and this branch has already had one self-inflicted bug in exactly this area.

**[L38] — fixed on both halves.** The block no longer says the fallback exists "for that path only", and no longer characterizes the degradation as just the `isSecurity` under-report. It now names both losses and points at why only one of them reaches the gate.

---

CamSoper:
@claude #update-review

**[L242] — conceded and fixed in `6c0f017d`.** Verified independently: counting `| ---` separator rows gives exactly one table in #20711, #21282 and #21285, and #20711 carries a single `Bumps the …` line. "More than one table block" was wrong.

Your reading of what I meant is right, and it's the better sentence. #20711's twelve rows run ten direct dependencies alphabetically, then `beasties` and `simple-icons` appended — one table, two blocks of rows. That's exactly why the two lost packages sort early in the alphabet but sit at rows 11 and 12, and last is what the cut takes. The comment now says that, and keeps the operative reason the body outruns 65,536 characters (the per-package release-note `<details>` blocks), which was already correct.

No functional change. That should be the last of it — the gate logic has been unchanged since `4135dbcd` and the remaining rounds have all been comment accuracy.

---

CamSoper:
@claude #update-review

**[L251] — conceded, fixed in `675ebd24`, and fixed by deletion rather than correction.**

You're right on both counts. Sorted against all twelve names the pair is 4th and 11th, and reproducing "3 and 10 of 11" requires silently dropping `@aws-sdk/client-s3` — row 1 of the table. The figures came from copying 0-based indices out of a scratch script without re-reading them.

The claim is gone rather than corrected. It's the least load-bearing part of the comment and has now produced three consecutive rounds of findings on its own, which is a reasonable signal that a coordinate nobody needs isn't worth carrying. What remains is the part a reader can actually replay: 12 named, 10 recovered, and the two lost are rows 11 and 12 — the appended indirect block. That they sort mid-alphabet is now stated as *the point* (position in the table decides, not the name) rather than as a number to check.

Gate logic unchanged since `4135dbcd`; the last three commits have been comment accuracy only.

---

PR #21381, opened 2026-09-03
Reviewer brief: empty ⚠️ sentinel acknowledges stances; 'the first <ordinal>' is not a stance

CamSoper:
Two things Cam spotted on the v3 reviewer brief for pulumi/docs#21369:

- **Contradictory sentinel.** The ⚠️ section read "Nothing needs a human eye beyond the rubber-stamp list below" directly above an editorial-stances block the reviewer is asked to confirm. The empty-table sentinel now comes from one helper that knows whether the stances H4 follows, used by all three emitters (composer, `build-evidence.py` collapse, `apply-update.py` re-render).
- **False-positive stance.** "for the first deployment, create it" matched the positioning regex's `the first`. Ordinal/temporal nouns after "the first" no longer match; "the first IaC tool" still does.

Regression tests for both.

https://claude.ai/code/session_01HgQPx7TyE73KH1Bk4TBfHq

### DIFF
--- a/.claude/commands/docs-review/scripts/apply-update.py
+++ b/.claude/commands/docs-review/scripts/apply-update.py
@@ -530,7 +530,12 @@ def _render_doc(body: str, rows: dict[str, dict], resolved_rows: list[str], doc:
             new_lines = _table(resolved_rows) if resolved_rows else [RESOLVED_PLACEHOLDER]
         else:
             bucket_rows = by_bucket.get(bucket)
-            new_lines = _table(bucket_rows) if bucket_rows else [SECTION_EMPTY.get(bucket, "")]
+            empty = SECTION_EMPTY.get(bucket, "")
+            if bucket == "reviewer-check":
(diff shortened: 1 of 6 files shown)


CamSoper:
@claude #update-review

Fixed both: L354 now uses the positive-lookahead shape you suggested (`the first X to/that <verb>`) with first-class, plural, and numeric ordinals added to the test; L2019 drops `V3_EMPTY_CHECKS_SENTINELS` so `empty_checks_sentinel()` is the only source.

CamSoper:
@claude #update-review

Fixed L356 with the intersected pattern you proposed: ordinal nouns are rejected inside the `the first X to/that` frame, nouns accept hyphens, and `the first step to take` / `the first thing to do` are pinned in the test.

CamSoper:
@claude #update-review

Fixed L359: `s?` before the `\b` in the in-frame denylist, with `the first steps to take` and `the first lines that match` pinned in the test.

PR #21393, opened 2026-09-03
Make pulumi/docs authoritative for Pulumi Registry submission

### LINKED ISSUE #21389: Update Publishing to the Pulumi Registry with docs from pulumi/registry
https://www.pulumi.com/docs/iac/guides/building-extending/packages/publishing-packages/ should be the authoritative source for guidance on contributing to the Pulumi (public) Registry.

We have updated information in pulumi/registry that we need to port over here, and then we need to remove the duplicated content from pulumi/registry (and point to the docs website).

The general guideline is that we outline the process in pulumi/docs and more niche topics (e.g. previewing changes locally) can live in pulumi/registry.

jkodroff:
Fixes #21389. Makes `/docs/iac/guides/building-extending/packages/publishing-packages/` the authoritative source for contributing to the public Pulumi Registry, porting the process from `pulumi/registry` (`docs/adding-a-new-package.md` and `docs/overview-page.md`, both rewritten 2026-09-02). Repo-local mechanics stay in `pulumi/registry`.

## Two instructions that fail a contributor's PR today

- Required an `installation-configuration.md`. Only `docs/_index.md` is required.
- Told authors to commit docs into `themes/default/content/registry/packages/`. CI rejects a PR touching anything outside the package list.

## Added

One-entry package-list flow · fact-sheet / `/check` / `/preview` loop · publisher registration in `publisher-names.json` · `v`-prefixed release and schema-version rules · the User-Agent ask · the full overview-page standard · instructions for getting an Any Terraform Provider listed.

## Verified by running it, not by reading

- `pulumi package add` generates a local SDK with no published SDK, so per-language SDKs are now framed as **recommended, not required**.
- Both generated and published SDKs carry `pluginDownloadURL`, so the "tell readers to `pulumi plugin install`" advice is gone.
- HCL resolves packages via `required_providers` + `pulumi install`, **not** `pulumi package add`. The registry's own guidance and ~45 package pages have this wrong.
- Bridge token mapping is automatic — the boilerplate's own comment says "You shouldn't need to override anything" — so the "your work is mapping tokens" framing is corrected.

## Also

- `content/docs/iac/concepts/packages/_index.md`: softened an inaccurate "the Registry requires package contributors to generate SDKs in all languages" claim to a recommendation.
- `styles/Pulumi/ignore/tech-vocabulary.txt`: added `Gradle`, `Logfire`, `shortcodes`, `wordmarks`.

`#publish-your-package` is preserved — three inbound links depend on it. `#publish-the-documentation` was renamed; it was only self-referenced.

**Follow-up (not in this PR):** a `pulumi/registry` PR trimming the now-duplicated copies to pointers, once this is live.

https://claude.ai/code/session_0165YLLPveFcTpZQoLAPeR4S


jkodroff:
@CamSoper Need an actual human review here b/c I can't stare at this page anymore and need a new set of eyes.
@iwahbe Tagging you in for a Pulumi CLI/Registry technical knowledge sanity check
@tatcoo-pulumi Tagging you for anything process-related (that is, ensuring this guidance is in alignment with our strategy for external public registry contributions)

Labeling `do-not-merge` b/c I want the AI review w/o actually merging.

jkodroff:
@claude #update-review

**Fixed (pushed in eb477efd47):**

- **L148** (blocker) — dropped the "worked examples" citation entirely. You were right that Logfire's `docs/_index.md` opens with `# Logfire Provider` and so contradicts the rule at line 184. Rather than swap in ImprovMX, the author's call is that neither repo is exemplary enough to hold up as the standard, so the page now states the standard without pointing at an implementation of it.
- **L128** — the claim is correct (see below), but I took your point that "fails" was doing a lot of unexplained work. The line now names the error: `expected file ... to start with YAML front-matter`, and notes it's the most common reason a package's docs fail to publish.
- **L375** — dropped the Google Form link. I could not find that URL anywhere in `pulumi/registry`, and it was new in this PR, so it isn't something I can source. The individual/company split itself *is* documented — `pulumi/registry` `docs/adding-a-new-package.md` maintainer checklist: *"If the provider is community maintained (maintained by a person, not a company), then a GitHub handle is sufficient. If the provider is maintained by a company, Pulumi needs a contact person at the maintaining company."* The sentence now says exactly that, which also settles the requirement-vs-"please" mismatch you flagged: it's a maintainer checklist item, not an automated gate.
- **L96** (style) — "several upstream resources" → "more than one upstream resource".

**Disputing the remaining low-confidence findings.** All seven verified against primary sources. Several are near-verbatim from `pulumi/registry`, which is the upstream this PR is absorbing into `pulumi/docs` — that repo is the authority for the whole page.

- **L114** — bridged provider User-Agent. `pulumi/pulumi-terraform-bridge`, `pkg/vendored/opentofu/httpclient/useragent.go`: `DefaultApplicationName = "OpenTofu"`, and `OpenTofuUserAgent(version)` returns `OpenTofu/<version>`. Also stated verbatim in `pulumi/registry` `docs/adding-a-new-package.md` step 2.
- **L116** — no `ProviderInfo` field. I enumerated every field of `info.Provider` in `pkg/tfbridge/info/info.go` (48 fields: `P`, `Name`, `ResourcePrefix`, … `SkipDefaultFixups`). There is no user-agent or application-name field. Leaving the negative unhedged: it matches `pulumi/registry`'s own wording, and the sentence already routes a stuck reader to Slack, so it degrades gracefully if a field ever lands.
- **L144** — `# WARNING:` and `edit_url:`. Confirmed against a live generated page, `pulumi/registry` `themes/default/content/registry/packages/logfire/_index.md`, which begins:

  ```yaml
  ---
  # WARNING: this file was fetched from https://raw.githubusercontent.com/pydantic/pulumi-logfire/v0.1.19/docs/_index.md
  # Do not edit by hand unless you're certain you know what you are doing!
  edit_url: https://github.com/pydantic/pulumi-logfire/blob/v0.1.19/docs/_index.md
  title: Logfire
  ```

- **L159-165** — the doc lint covers all three patterns, not two. `pulumi/registry` `scripts/ci/community-package/doc_lint.py`, `_PATTERNS`: three regexes for `relative-image` (`<img src="` with `"`, with `'`, and ``), one for `raw-relative-link` (`<a href="./…`), one for `md-suffixed-link` (`](…​.md)` / `](…​.md#…)`). The table promises nothing the lint doesn't catch.
- **L330** — schema `version` vs release tag. Enforced by `_schema_version_matches()` in `scripts/ci/community-package/verify_entry.py`, which feeds the `green` gate. `fact_sheet.py` renders it as a `schema version ❌` row plus: *"The registry publishes the schema straight from the release tag and rejects a version that does not match. Either omit the `version` key entirely, which is what bridged providers do so the registry takes the version from the publish request, or stamp the real version at release time."* That is the sentence on line 330.
- **L353** — publisher action. `pulumi/pulumi-package-publisher` `action.yml` declares the `sdk:` input (default `all`, supports `nodejs,python`, `all,!nodejs`), with steps *Nodejs — Publish to NPM*, *.NET — Publish to Nuget*, *Python — Publish to PyPi*, *Java — Publish to Maven*. There is no Go step, which is the Go-tag exception.

**Accepted as-is:** the two remaining weasel-word flags. L90 "mostly" — the following sentence covers the per-resource overrides that make it not-all, so dropping the hedge would overstate. L389 "usually" — the key and value differ when a publisher already ships under an existing slug, which the next sentence covers.

CamSoper (approved):

content/docs/iac/concepts/packages/_index.md
```diff
@@ -167,7 +167,7 @@ You can author a Pulumi package in any language, create a hand-authored schema,
...
 
 - If the component is intended for internal use and your organization has security policies that restrict the ability of developers to install software on their devices (specifically, a required runtime for your package), writing your component in Go and publishing it as a binary with published SDKs hosted in an internal package feed will make it easier for consumers to use your package.
-- If you are intending to publish your component(s) in the Pulumi Registry for general public consumption, you should write your component in Go, and publish it as a binary with published SDKs hosted in the standard public package feeds (i.e., npm, PyPI, etc.). Note that the Pulumi Registry requires package contributors to generate SDKs in all languages Pulumi supports.
+- If you are intending to publish your component(s) in the Pulumi Registry for general public consumption, you should write your component in Go, and publish it as a binary with published SDKs hosted in the standard public package feeds (i.e., npm, PyPI, etc.). We recommend publishing SDKs for all the languages Pulumi supports, so your package is usable by the whole Pulumi community.
```
iwahbe, line 170: This is a policy change. I like the policy change, but I want to make sure we are doing this intentionally.

CC Whoever owns the registry. I guess that's @tatcoo-pulumi by default.

content/docs/iac/guides/building-extending/packages/publishing-packages.md
```diff
@@ -23,150 +23,373 @@ aliases:
...
+
+Bridging it into a package of your own buys you control over the package name, the documentation, and the release cadence, at the cost of a package you have to keep current with every upstream release.
+
+Popular Terraform providers appear in the Registry as Any Terraform Provider listings, so readers can find and evaluate them there the same as any other package. [Honeycomb](/registry/packages/honeycombio/) and [Supabase](/registry/packages/supabase/) are examples.
+
+To have a Terraform provider listed this way, open a [New Package issue](https://github.com/pulumi/registry/issues/new?template=new-package.yml) with the provider's name and repository. These listings are generated automatically and carry no logo by default; to add one, contact [Pulumi support](/support/new/) with a link to a web-accessible SVG. Wordmarks are preferred, with all surrounding whitespace removed.
```
iwahbe, line 55: We should always grab an appropriate wordmark or logo here. I'd have the submitter find one and submit it with the issue.

If we are going to host the docs for it, we should bother to get the logo. We should backfill all our existing TF providers with these logos as well.

content/docs/iac/guides/building-extending/packages/publishing-packages.md
```diff
@@ -23,150 +23,373 @@ aliases:
...
 
-## Create a repository and author your package
+## Create and name your repository
 
-To get started, create a repository for your Pulumi Package. We recommend hosting your Pulumi Package in a public repository on GitHub. We also recommend following the naming conventions below to help the community find the source code for your packages.
+Host your package in a public GitHub repository named `pulumi-<name>`, where `<name>` is the name you'll publish under. It's normally the name of the cloud or service the package configures, and it has to be unique in the Registry.
 
-### Select a template
+- Bridging a Terraform provider: reuse the upstream provider's name, replacing `terraform-provider-` with `pulumi-`. `terraform-provider-auth0` becomes `pulumi-auth0`.
+- A native provider: name it after the service it manages, as in [`pulumi-aws`](https://github.com/pulumi/pulumi-aws) and [`pulumi-kubernetes`](https://github.com/pulumi/pulumi-kubernetes).
```
iwahbe, line 73: Let's use native providers for this example.
```suggestion
- A native provider: name it after the service it manages, as in [`pulumi-kubernetes`](https://github.com/pulumi/pulumi-kubernetes) and [`pulumi-pinecone`](https://github.com/pinecone-io/pulumi-pinecone).
```

content/docs/iac/guides/building-extending/packages/publishing-packages.md
```diff
@@ -23,150 +23,373 @@ aliases:
...
 
-## Write documentation
+A component package can also declare custom resources of its own, so you aren't limited to composing what other providers offer. If your components need a resource type no provider manages, you can implement it in the same package.
 
-We recommend writing documentation to help others in the Pulumi community use your package. In your repository, there should be a `docs/` folder containing markdown files (the templates include a few suggested pages). The files should correspond to the various tabs on a package page in Pulumi Registry (like the [Azure Native](/registry/packages/azure-native/) package). Use the guidance in the following sections to author content in these pages.
+Components are most often distributed as [source-based plugin packages](/docs/iac/guides/building-extending/packages/source-based-plugin/), which the public Registry doesn't support. To list a component package in the public Registry, build it as an [executable plugin package](/docs/iac/guides/building-extending/packages/executable-plugin/).
```
iwahbe, line 108: AFAIK the public registry neither knows or cares if a plugin is source-based or compiled. They both have schemas, and that's all we need. 

I'm not aware of an imposed rule requiring compiled plugins in the registry, though we might want to note that a source-based component requires the supporting tool-chain on downstream consumer's devices.

content/docs/iac/guides/building-extending/packages/publishing-packages.md
```diff
@@ -23,150 +23,373 @@ aliases:
...
+---
+```
+
+| Key | Notes |
+|---|---|
+| `title` | The package display name. Should match `displayName` in your `schema.json`. Rendered as the page's heading. |
+| `meta_desc` | One sentence, used as the page's meta description. Include the package name. |
+| `layout` | Use `package`. |
+
+The generator adds a `# WARNING:` comment and an `edit_url:` key of its own when it publishes; you don't write those.
```
iwahbe, line 144: We only write that when the file was sourced externally. We don't write it to a file the user writes. The user should never need to know or care about these.

content/docs/iac/concepts/packages/_index.md
jkodroff: We already have some packages in there that don't e.g. have Java SDKs (including one of the ones we previously cited as "exemplary"), which is why I made this change.

iwahbe: I think its a positive change, but it is a policy change that we should make consciously. Since we now support local packges, we can drop the requirement for SDKs to be published.

---

In the far future:

The registry should know which SDKs are present for which languages, and it should maintain SDKs in the remaining languages in a `@pulumi-registry/...` namespace (or `Pulumi.Registry`, whatever is language appropriate). That is not the case now.

CamSoper:
@claude #update-review

---

content/docs/iac/guides/building-extending/packages/publishing-packages.md
jkodroff: Cut. It's true for this flow (the docs page is externally sourced, so the generated page does carry both), but your point stands that it's noise the author never acts on. The front-matter table right above already tells them the three keys they do write.

content/docs/iac/guides/building-extending/packages/publishing-packages.md
jkodroff: Agreed. The paragraph now asks the submitter to include a web-accessible SVG link in the New Package issue itself, with the support contact demoted to a fallback for listings that are already live.

The issue template has no logo field yet (inputs are PackageType, PackageName, RepositoryUrl, Connection, OtherInfo), so the docs are currently ahead of it. Filed pulumi/registry#12420 for the template field and the backfill of existing Terraform provider listings.

jkodroff:
@claude #update-review

Pushed three fixes from @iwahbe's review, all in `publishing-packages.md`:

- **L108** — the claim that the public Registry doesn't support source-based plugin packages was wrong. It accepts either kind; rewritten as a runtime/reach trade-off. Verified against `pulumi/registry` `scripts/ci/community-package/verify_entry.py` (checks release tag, `schemaFile`, generated docs, schema-version match — nothing about plugin kind) and `docs/adding-a-new-package.md` (names Component a first-class PR path).
- **L55** — the logo ask moved from a post-hoc support contact into the New Package issue itself. Template field and backfill filed as pulumi/registry#12420.
- **L144** — dropped the `# WARNING:` / `edit_url:` sentence. Accurate, but the generator writes those keys and the author never types them.

Note this supersedes your earlier `low:L144` finding, which I'd refuted on accuracy grounds — the line is gone for relevance, not correctness.

jkodroff:
One more push: reframed the SDK line in `packages/_index.md` as preferred rather than required.

The previous wording ("so your package is usable by the whole Pulumi community") implied a package without published SDKs is less usable, which reads badly against the Terraform-bridged listings already in the Registry that ship none. It now states the install trade-off and points at `pulumi package add` for local SDK generation.

That matches how the Registry's own check behaves: in `sdk_install_probe.py` the plugin install is `blocking=True` while every per-language SDK probe is `blocking=False`, and an `absent` SDK isn't counted even as an advisory failure.

@claude #update-review

content/docs/iac/guides/building-extending/packages/publishing-packages.md
jkodroff: Fixed.

content/docs/iac/guides/building-extending/packages/publishing-packages.md
jkodroff: Fixed.

content/docs/iac/guides/building-extending/packages/publishing-packages.md
jkodroff: Fixed.

PR #21394, opened 2026-09-03
Generate the Pulumi package schema reference from upstream

### LINKED ISSUE #21390: Auto-generate Pulumi Schema Docs
Investigate whether we can devise a process to auto-generate the Pulumi Schema docs from the pulumi/pulumi source code or a published SDK, similar to how we do with other reference pages like ESC schemas, policy packs, etc..

If it's easy to keep them versioned by following patterns elsewhere, we should to that.

jkodroff:
Closes #21390.

The schema reference at [/docs/iac/guides/building-extending/packages/schema/](https://www.pulumi.com/docs/iac/guides/building-extending/packages/schema/) was 427 hand-maintained lines reproducing types that live in three other repos. Its content hadn't been touched since June 2025, and it had drifted. This generates it instead.

## What it was getting wrong

| | |
|---|---|
| **Wrong** | `Config` documented a `required` key — the wire key is `defaults`. `Alias` documented `name` and `project`, both removed upstream. `Type` marked `type` **and** `$ref` required, which is impossible; they're mutually exclusive. `Resource.inputProperties` carried a copy/pasted description of a different field. The Node.js table documented `pluginName` and `pluginVersion`, gone from `NodePackageInfo`. |
| **Missing** | `dependencies`, `allowedPackageNames`, `supportPack`, `listInputs`, `returnType`, `multiArgumentInputs`, `isOverlay`, `overlaySupportedLanguages`, and most of the Go and Java language options. |
| **Stale claims** | Four sections read "Not yet supported in any target language" for extensions that exist: `CSharpResourceInfo`, `NodeObjectInfo`, and the Python and Java `PropertyInfo` structs. |
| **Voice** | Descriptions were pasted from Go doc comments, so they read "Name is the unqualified name of the package…". |

## Where the content comes from now

**Core reference** — `pkg/codegen/schema/pulumi.json` in pulumi/pulumi. That file is the Pulumi Package Metaschema: `//go:embed`-ed into every `pulumi` binary and validated against verbatim by `pulumi schema check`. If it disagrees with the page, the page is wrong. Its descriptions are already written as prose.

**Language extensions** — each language's own `*PackageInfo` / `*PropertyInfo` / `*ResourceInfo` / `*ObjectInfo` Go struct. The metaschema *has* `*LanguageSpec` `$defs`, but they're a stale subset — its `goLanguageSpec` lists four properties where `GoPackageInfo` has fifteen — so the structs are the truth.

| Section | Repo | Path |
|---|---|---|
| Core type reference | pulumi/pulumi | `pkg/codegen/schema/pulumi.json` |
| Go, Node.js, Python | pulumi/pulumi | `pkg/codegen/{go,nodejs,python}/importer.go` |
| C# | pulumi/pulumi-dotnet | `pulumi-language-dotnet/codegen/importer.go` |
| Java | pulumi/pulumi-java | `pkg/codegen/java/{package_info,importer}.go` |

## Three workflows, one per repo

Those repos release on their own cadences, so each gets its own workflow, writing only its own data files, fired by that repo's release. `git add` takes an explicit file list for the same reason — a pulumi-dotnet release must never open a PR that also rewrites the Java section.

- `package-schema-docs.yml` — pulumi/pulumi, via a new fan-out job in the existing `pulumi-cli.yml` release dispatcher.
- `package-schema-docs-dotnet.yml` / `package-schema-docs-java.yml` — subscribe to the `pulumi-dotnet-sdk` / `pulumi-java-sdk` `repository_dispatch` types those repos **already send**. A dispatch fires every workflow subscribed to its type, so **no upstream change is needed**.

Each section on the page states the release it was generated from. Files are fetched one at a time from `raw.githubusercontent.com` at the release tag: no checkout, no Go toolchain, no auth.

## The editorial layer

`data/package_schema_overrides.yaml` holds what upstream can't decide: which parts of the sources become documented types, what each is called, which anchor it gets, and the handful of properties upstream leaves with no description at all (five C# fields, two Python ones). The generator **fails on an override that stops matching anything upstream**, so that layer shrinks as upstream fills its own gaps instead of quietly rotting.

## Rendering

Reuses the `.api-param*` stacked layout from `openapi.css` that the REST API reference uses, which gives each description the full content width — the longest one here (`$ref` on `Type`) was a paragraph crushed into a table cell between the left nav and the TOC.

Two things are load-bearing and worth knowing if you touch the shortcode:

- **Headings carry explicit ids** and are emitted **outside** `<pulumi-choosable>`. The docs TOC isn't Hugo's `.TableOfContents` — `theme/src/ts/misc.ts` scans `#main` for `h2, h3` *that have an id*, and skips only `.hidden` ancestors. An inactive choosable is hidden by a `:not(.active)` rule instead, so a heading inside one would show up in "On this page" once per language.
- **`data/package_schema/anchors.yaml` is a contract.** The shortcode fails the build if a render stops emitting one of the 21 fragments the hand-written page published. Nothing in CI validates URL fragments (`check-links.js` doesn't look at them), and `provider-configuration.md` deep-links `#default`.

## Also

- Refreshed the example schema, which pinned the AWS provider at v3/v4 (~five years stale).
- Sentence-cased the two `##` headings; both anchors are unchanged.
- Linked `pulumi schema check`, which validates against exactly this metaschema and was previously linked from neither direction.

## Verification

- `make test-unit` — 28 new tests covering the pointer/type resolution and the Go struct scanner, including that a malformed struct body throws rather than silently dropping properties.
- `make lint` clean; `make lint-prose` down to two pre-existing wordiness warnings in the untouched intro.
- Full `hugo` build clean. All 21 contract anchors present in the output; deliberately breaking the contract fails the build as intended.
- Re-running all three generators produces byte-identical files, so a release that changed nothing here opens no PR.

## Upstream follow-ups (not blocking)

- `pulumi.json` lacks `plainInputs` and `extensionParameterization`, which `schema.go` has. The metaschema root is `additionalProperties: false`, so a schema emitting `extensionParameterization` fails `pulumi schema check` today.
- `pulumi/pulumi`'s own `docs/references/metaschema.md` is regenerated by hand and is stale (missing `listInputs`); nothing in CI checks it for drift.

## Review notes

The `data/package_schema/*.json` files are generated — review them the way you'd review the upstream change they came from. `content/.../schema.md`, `scripts/fetch-package-schema.js`, `data/package_schema_overrides.yaml`, the two layouts and the three workflows are the hand-written parts.

https://claude.ai/code/session_01AurPbqvauGSv3v7C8CCM1s


CamSoper:
@copilot resolve the merge conflicts in this pull request

CamSoper:
@claude #update-review

CamSoper (approved):

PR #21395, opened 2026-09-03
Rebuild /terraform/ around the August 2026 Terraform release

djperlovsky:
### Proposed changes

Rebuilds `/terraform/` (`content/migrate/terraform.md`), which went stale after the [August 2026 release](https://www.pulumi.com/releases/terraform-state-backend-modules-hcl/).

**Why the page needed more than a copy refresh.** Its premise was inverted by the release. The page argued *migrate off of Terraform*; the release says the opposite — your state, modules, and HCL are things Pulumi runs. Its examples had also aged out, advertising `nodejs8.10` Lambdas and `"${aws_s3_bucket.mybucket.id}"` interpolation syntax.

**What changed.** The page now opens on *"Keep your Terraform. Gain the platform around it."* and leads with the three release capabilities: Terraform state in Pulumi Cloud, modules consumable across languages, and HCL as a first-class Pulumi language. Conversion didn't disappear — it moved into an "Adopt at the pace that suits you" section as one of six paths, alongside *keep your HCL* and *run both side by side*, with [Pulumi vs. Terraform](https://www.pulumi.com/docs/iac/comparisons/terraform/) linked as the honest "here's where Terraform fits better" option.

**How it's built.** Converted from the bespoke `layouts/migrate/terraform.html` to `layout: template-page`, composed from eight existing `template-partials/*` sections. That picks up the shared design system, responsive fingerprinted images, and makes the content frontmatter-editable. The old layout had no other consumer and is deleted (−113 lines). The three section illustrations are the release's own SVGs, already in `assets/fingerprinted/`.

`url: /terraform` is unchanged, so the URL and its search rankings carry over.

**Customer proof is framed as scale** — Wiz (1M+ resources), BMW (20,000+ resources), Mercedes-Benz (many teams, many clouds). Starburst was considered and dropped: the case-study card auto-pulls the case study's quote, and Starburst's quote contrasts against Terraform, which reads oddly under a "keep your Terraform" headline.

### Verification

- `hugo --minify` production build clean
- `make lint` — 0 errors, Prettier clean
- `make lint-prose` (Vale) — 0 errors, 0 warnings, 0 suggestions
- All 8 sections render non-empty; all 7 images load; mobile (375px) stacks correctly
- Copy written against the brand guide's voice, writing-style, and terminology sections via the brand MCP server

### Notes for review

Three deliberate calls worth a second opinion:

[...]

PR #21396, opened 2026-09-04
apply-update: a resolved finding can come back when its fix is reverted

CamSoper:
First live update-lane error on v3 (pulumi/docs#21395). The lane had recorded F2 as `fixed` when the fix commit landed; the author reverted that commit and accepted the finding as-is. The model's `accept` was refused with "F2 is not an open finding" because the contract only knew open rows, so the lane errored twice with no way back except `#new-review`.

A ✅ Resolved finding is addressable now: `reopen`, `accept`, `hold`, `retext`, and `promote` may name one. The row is rebuilt from the evidence record (bucket and original text; 🚨 and the annotation-stripped cell when there is none), leaves the ✅ table, and the lane's own `fixed` record is removed so the finding blocks again. A human disposition that landed meanwhile is kept. `resolve` and `concede` on a resolved row stay rejected. `update.md` documents the action.

Replayed against the failed run's real patch and live cards: F2 reopens and lands on the brief as accepted as-is, header recomputes, cards and evidence validate clean. Four regression tests.

https://claude.ai/code/session_01CawS2ZRwZUsa3Y8bgvMMyf

### DIFF
--- a/.claude/commands/docs-review/references/update.md
+++ b/.claude/commands/docs-review/references/update.md
@@ -248,6 +248,7 @@ If the author deletes the 1/M comment via the GitHub UI, the next re-entrant run
    {"id": "F5", "action": "hold",    "reason": "evidence: the docs say otherwise"},
    {"id": "F8", "action": "accept",  "reason": "author: internal figure, shipping as-is", "bulk": false},
    {"id": "F6", "action": "promote", "to": "outstanding", "reason": "also in social copy"},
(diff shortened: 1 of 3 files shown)


CamSoper:
@claude #update-review

All four fixed: L433 retext now builds the Do-this block from nothing on a reopened row; L86 promote to the row's current bucket is a no-op (only downward raises); L263 resolve renders `— ✅ <annotation>` and concede `— concede: <reason>`, and `strip_resolution()` splits on those markers (legacy cells fall back to the last separator) — tested with em dashes on both sides; L82 the docstring lists hold, accept, and reopen.

PR #21408, opened 2026-09-04
Context API docs: use the orgless GetGraphQuerySchema

lblackstone:
## Why

`GetGraphQuerySchema` serves the same Context API schema for every organization and requires authentication but no organization. This lets an agent fetch the current primer before choosing an organization for a query. The organization-scoped `GetGraphSchema` is deprecated in its favor.

## What

The guide and launch blog now use `pulumi api GetGraphQuerySchema`. The guide also clarifies how `pulumi api` selects an organization for queries, what a `402 Payment Required` response means, and how to refresh the cached API specification.


lblackstone:
@pulumi-eon /review

PR #21453, opened 2026-09-08
Support form: file Intercom conversations instead of tickets

MeganYTan:
## Summary

- Switches the `/api/support` Lambda from filing an Intercom ticket to filing an Intercom conversation, verified against the live API (contact search/create → `POST /conversations` → `PUT /conversations/{id}` to set `custom_attributes`, since attributes aren't accepted at creation time).
- Subject and description are folded into the conversation's message body; `pulumi-org` and `priority` go through the follow-up `custom_attributes` update.
- Renames the response/log field `ticketId` → `conversationId` throughout (`handler.ts`), and the error codes `ticket_creation_failed`/`support_request_ticket_failed` → `conversation_creation_failed`/`support_request_conversation_failed`. The frontend (`theme/src/ts/support-form.ts`) only reads `body.ok`, never `ticketId`, so this is not a breaking change externally.
- Lowers `INTERCOM_TIMEOUT_MS` 2500ms → 2000ms: the conversation flow can take up to 4 sequential Intercom calls (vs. 3 for tickets), and 4 × 2500ms would exceed the Lambda's 10s timeout with no margin.
- Removes the now-dead `intercomTicketTypeId` / `INTERCOM_TICKET_TYPE_ID` config plumbing (`supportForm.ts`, `index.ts`, both `Pulumi.www-testing.yaml` and `Pulumi.www-production.yaml`) — conversations have no ticket type.
- Updates `intercom.test.ts` and the handler-level tests in `validation.test.ts` for the new call sequence and payload shapes.

## Test plan

- [x] `yarn test-support-form` — all 55 tests pass
- [x] Verified the real request/response shapes against the live Intercom API by hand (contact search, conversation create, conversation attribute update) before writing the code
- [x] Confirm in the Intercom workspace that `pulumi-org`/`priority` custom attributes are defined on the `conversation` model (they were already present when tested against `www-testing`'s credentials)
- [x] Manual end-to-end test of `/support/new/` against a deployed `www-testing` stack


CamSoper:
@claude implement both of the suggested fixes as described and include a comment somewhere (perhaps the infra file itself -- but I'm willing to accept AGENTS.md if the text is extremely concise) mentioning the relationship between the contract and llms.txt. We want to nudge agents to updating both at the same time.

#update-review

CamSoper (approved):
Deployed e87abe0 to www-testing and regression-tested: API contract probes (405/400/422/413/403, honeypot dropped without an Intercom call), plus one browser-path and one agent-path submission (the latter following the /llms.txt instructions). Both logged `support_request_accepted` with `attributesSet: true` and `ipSource: viewer`, and the conversations landed in the testing Intercom workspace with org and priority set. Local suites pass (56 infra / 40 theme).

The one open ⚠️ is real but cosmetic: the Hugo comment leaves an extra blank line before the support bullet in rendered /llms.txt. Accepting as-is; `*/ -}}` fixes it whenever that file is next touched.

PR #21465, opened 2026-09-08
docs(insights): note removal of the org-scoped GetGraphSchema operation

lblackstone:
The guide's schema section still described it as deprecated but working; it now says the operation returns 404 and points callers at `GetGraphQuerySchema`.

Everything else in the docs already uses `GetGraphQuerySchema`, and the REST API reference regenerates from the published OpenAPI spec at build time, so no other content changes.


lblackstone:
@pulumi-eon /review

lblackstone:
@pulumi-eon /review

fallimic (approved):

PR #21506, opened 2026-09-09
docs: document no-subcommand exit code in the exit-codes page

shashankvarma499:
The CLI Exit Codes page does not mention what happens when `pulumi` is run with no subcommand. Since v3.254.0 a bare `pulumi` invocation prints help and exits with a non-zero code (previously it exited 0), which silently broke downstream health checks that used this pattern.

This adds a short note documenting the behavior and pointing to `pulumi version` (or `pulumi --help`) as the reliable way to check that the binary is installed and working.

Refs pulumi/pulumi#24301.

### DIFF
--- a/content/docs/iac/cli/exit-codes.md
+++ b/content/docs/iac/cli/exit-codes.md
@@ -39,4 +39,4 @@ The exact non-zero value of a failing exit code gives us some information about 
 
 ## Running the CLI with no subcommand
 
-Running `pulumi` with no subcommand prints the CLI help and then exits with a non-zero exit code. This behavior changed in Pulumi CLI v3.254.0; earlier versions exited with code 0. If your automation runs a bare `pulumi` command to check that the binary is installed and working, use `pulumi version` or `pulumi --help` instead, both of which exit with code 0.
+Running `pulumi` with no subcommand prints the CLI help and exits with code 1, the generic error code. This behavior changed in Pulumi CLI v3.254.0; earlier versions exited with code 0. If your automation runs a bare `pulumi` command to check that the binary is installed and working, use [`pulumi version`](/docs/iac/cli/commands/pulumi_version/) or `pulumi --help` instead, both of which exit with code 0.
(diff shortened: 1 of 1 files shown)


content/docs/iac/cli/exit-codes.md
```diff
@@ -36,3 +36,7 @@ The exact non-zero value of a failing exit code gives us some information about
 | Canceled run                      | 8         | The operation is canceled before completion, for example due to `pulumi cancel`, a user interrupt (such as Ctrl+C), or cancellation initiated through the Automation API. |
 | Timeout                           | 9         | The operation fails because it does not complete within an expected time window. |
 | Internal CLI error                | 255       | An unexpected internal condition occurs inside the Pulumi CLI itself that does not fit another category. |
+
+## Running the CLI with no subcommand
+
+Running `pulumi` with no subcommand prints the CLI help and then exits with a non-zero exit code. This behavior changed in Pulumi CLI v3.254.0; earlier versions exited with code 0. If your automation runs a bare `pulumi` command to check that the binary is installed and working, use `pulumi version` or `pulumi --help` instead, both of which exit with code 0.
```
CamSoper: Two small tweaks, bundled as a suggestion so it can be applied in one click. Since this is the exit codes page, let's name the actual code (it's 1, the generic error row in the table above -- verified against v3.262.0), and link `pulumi version` to its command reference page.

```suggestion
Running `pulumi` with no subcommand prints the CLI help and exits with code 1, the generic error code. This behavior changed in Pulumi CLI v3.254.0; earlier versions exited with code 0. If your automation runs a bare `pulumi` command to check that the binary is installed and working, use [`pulumi version`](/docs/iac/cli/commands/pulumi_version/) or `pulumi --help` instead, both of which exit with code 0.
```

---

CamSoper (approved):
Thanks for this -- it's a real gap, and the kind of thing that only shows up when a health check breaks at 2am.

Checked the claims against the pulumi CHANGELOG (the change is in 3.254.0 via pulumi/pulumi#23848) and confirmed empirically on v3.262.0: bare `pulumi` exits 1, `pulumi --help` and `pulumi version` both exit 0. Vale is clean on the new lines (the two warnings it reports are on pre-existing content). Heading is sentence case, terminology checks out against the brand guide.

Approving. Thanks for the update!

PR #21531, opened 2026-09-10
fix(review): key the glow-up body gate on table rows, drop the declined placeholder, alert on publish refusals

CamSoper:
### Proposed changes

Three glow-up workers in a row passed the scope gate and were then refused by `publish-gate.py`'s Backlog accounting, so no glow-up PR opened on 2026-09-09 or 2026-09-10 and nobody was told: [install/_index.md](https://github.com/pulumi/docs/actions/runs/34361912504), [get-started/terraform/_index.md](https://github.com/pulumi/docs/actions/runs/34361915729), [guides/clouds/aws/elb.md](https://github.com/pulumi/docs/actions/runs/34487248304).

Two of the three refusals were gate false positives, the third was a composer trap. This PR fixes both, gives the model the gate as a self-check, and makes the next refusal loud.

**Gate keyed on rows, not substrings.** `glowup_body_accounting()` decided whether an id was "in" a Backlog table with `"`id`" in section`, so a reason cell that cross-referenced another row ("see `findings-f17`", "executed under `pr21457-findings-12`") counted as that id appearing in both tables. The elb body did that six times and was refused six times. Membership is now the backticked id that opens a row's first cell -- the same parse `record-page-findings.py`'s `declined_reasons()` already uses, so what the gate accepts is what the findings record reads back. Re-running the new check over the three refused bodies (downloaded from the handoff/snapshot artifacts): elb and install come back clean.

**No more placeholder row in Backlog declined.** The composer ended that table with `<TODO: any further banked finding ... or delete this row>`, whose only correct disposition was deletion, and the gate refused the body whenever the model left it in. That was the terraform failure. The row is gone; the HTML comment above the table carries the instruction. The terraform body fails the new check on that placeholder alone.

**Self-check.** `compose-pr-body.py --check-accounting --body-file .pr-body-draft.md --backlog .glowup-backlog.json` runs the publish gate's accounting over the draft and exits 2 on any violation, the way the skill already has the model self-check its diff with `verify-glowup-scope.py`. The skill's glow-up Validate step and the worker prompt now require it.

**Slack on refusal.** The publish job posts one line to #docs-ops when the publish gate, the guardrail gates, or the publish step fails, quoting the gate's `::error::` lines (stderr is teed to `.publish-gate.err`). Fail-open, skipped for the report lane.

Tests: `make test-review-pipeline` passes; `test_compose_pr_body.py` gains cross-reference and placeholder cases.

Follow-up once merged: re-dispatch the three refused glow-ups on master.

n/a

### Related issues (optional)

Follows #21469 (fix-mode Vale + fixpoint loop), which is what made glow-up bodies large enough to trip the substring check.

https://claude.ai/code/session_01J1jQUJLbDHt895GQEwMUoW

---
_Generated by [Claude Code](https://claude.ai/code/session_01J1jQUJLbDHt895GQEwMUoW)_

### DIFF
--- a/.claude/commands/review-existing-content/SKILL.md
+++ b/.claude/commands/review-existing-content/SKILL.md
@@ -560,7 +560,9 @@ Each banked item is split in two, and the split is the point:
 1. **Work the backlog first.** Execute every banked finding, or explicitly
    decline it with one line of reasoning. Every item lands in exactly one of
    the PR body's two tables — **Backlog executed** (pre-stubbed, one row per
-   item; fill "What changed") or **Backlog declined**. No silent drops.
+   item; fill "What changed") or **Backlog declined** (move the whole row,
+   keeping its id cell, and fill "Why not executed"). No silent drops, and
+   no row in both tables.
 1. **Then the secondary sweep**: apply the improvement taxonomy from
    `.claude/commands/glow-up.md` §5 — style, structural fixes, code
    formatting, terminology, links, image/diagram flags (flag-only, as ever),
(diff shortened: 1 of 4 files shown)


CamSoper:
@claude #update-review

Fixed both findings in e36b14b6 by taking the low-confidence suggestion's shape: the gate call now writes both streams to `.publish-gate.err` (`> .publish-gate.err 2>&1 || rc=$?`), replays the file with `cat`, and exits with the gate's status. That captures `fail()`'s stdout `::error::` lines (the blocker) and guarantees the file is complete before a non-zero status ends the step (the low-confidence one), with no change to `publish-gate.py`. Verified locally against the real terraform body from run 34361915729: the violation text lands in the file and the exit code is preserved.

---

CamSoper:
@claude #update-review

Took the low-confidence suggestion in ba469c75: the first "Publish gate" step now uses the same redirect / `cat` / re-raise shape as the publish-step call, so a refusal on either path lands its `::error::` lines in `.publish-gate.err` for the Slack alert to quote.

---

PR #21557, opened 2026-09-11
Update lane: quoted protocol is not a request, explicit refresh wins the race, pointer goes to the author

CamSoper:
Three things one week of v3 traffic showed, plus the doc drift the `REVIEW_V3_COMMENTS=1` flip would otherwise leave behind.

- **A quoted protocol is not a request.** `claude-update.yml` / `claude-new.yml` match `@claude` + the hashtag as substrings, so an explainer that quoted `@claude F<n>: … #update-review` in backticks (#21535) dispatched two full Opus runs. New `mention-gate.py` strips fenced blocks, inline spans, and blockquoted lines first; both workflows run it as step one (default-branch sparse checkout, body via env) and every later step gates on its verdict. `workflow_dispatch` runs are untouched.
- **An explicit `#update-review` wins the race.** Push-then-comment used to lose: the auto-refresh dispatch cancelled the author's run through the per-PR concurrency group and their reasoning was never adjudicated (#21535, 20:35). The auto-refresh job now yields when a live explicit request in the last 15 minutes still has a run in flight. Fails open.
- **The prose pointer goes to the author only.** It nudged a maintainer for a status note that mentioned F2 (#21395).
- Docs: CONTRIBUTING describes the two cards as the default and the monolith as legacy; `scripts/review-v3/README.md` names the real `review-worklist.py` flags; the `pr-review` skill refuses to write a legacy monolith on a v3 PR.

Tests: `mention-gate.py --self-test` + `test_mention_gate.py` (wired into `make test-review-pipeline`), resolve-handler pytest + self-test updated. Machinery only.

### DIFF
--- a/.claude/commands/docs-review/scripts/mention-gate.py
+++ b/.claude/commands/docs-review/scripts/mention-gate.py
@@ -1,0 +1,286 @@
+#!/usr/bin/env python3
+"""Decide whether a comment really asks for a review refresh.
+
+The `@claude … #update-review` / `#new-review` workflows fire on a plain
+substring match in their `if:` expressions (GitHub expressions can't strip
+markdown). That is enough to make a comment that merely *quotes* the reply
+pattern — an explainer with `@claude F2: … #update-review` in backticks, a
+quote-reply of the card's own instructions — dispatch a full model run
+(#21535: two of four card versions were unrequested). This script is the
+second, precise check the workflows run before spending anything: it
+removes fenced code blocks, inline code spans, and blockquoted lines, then
+looks for the mention and the hashtag in what is left.
+
+Two entry points, both side-effect free:
+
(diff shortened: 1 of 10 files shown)


CamSoper:
@claude F1: fixed — the inline-span pattern no longer uses re.S, so a stray backtick can't pair across a paragraph break and swallow a live request; self-test and pytest cover it #update-review

CamSoper:
@claude F3: fixed — the yield check now excludes exactly the login the workflow trigger excludes (claude[bot]) instead of a [bot]-suffix heuristic with a workprentice prefix allowlist; self-test and pytest cover claude[bot], workprentice[bot], another bot, and a human. F2: accept — the card's own instructions use fenced blocks; indented-block quoting is a follow-up #update-review

PR #21560, opened 2026-09-11
Smoke: multi-stack guidance on the stacks concept page (do not merge)

CamSoper:
End-to-end smoke of the v3 review surface on an unlabeled PR after today's variable flips. Deliberately flawed; will be closed, never merged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JLhMYkjKBgBXE8aWs2bv1V

PR #21561, opened 2026-09-11
Validator: give Step B a token so diff-based exemptions actually run

CamSoper:
The initial lane's validate step (Step B) ran `validate-pinned.py check --pr` without `GH_TOKEN`, so the validator's own `gh pr diff` failed silently and every diff-based exemption was dead. `internal-link-existence` then flagged a review for quoting the PR's own broken link. v2 hid this behind the splicer and model fallback; v3 has no soft floor, so the first v3 review that quoted a dead link errored (#21560).

- Step B gets `GH_TOKEN`.
- A failed `gh pr diff` inside the validator is now a `::warning::`, not a silent empty diff.
- New test: every workflow step invoking `validate-pinned.py check --pr` must carry a token.

### DIFF
--- a/.claude/commands/docs-review/scripts/test_validator_step_token.py
+++ b/.claude/commands/docs-review/scripts/test_validator_step_token.py
@@ -16,7 +16,8 @@ import pytest
 import yaml
 
 REPO_ROOT = pathlib.Path(__file__).resolve().parents[4]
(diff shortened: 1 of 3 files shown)


CamSoper:
F1 addressed in c3850d10da: the scan now covers both .yml and .yaml workflow files. Merging as admin (machinery only).

PR #21563, opened 2026-09-11
Review: deterministic dead-internal-link check on every PR

CamSoper:
Hugo doesn't validate plain markdown links and the Hugo pre-step is skipped for content-only PRs, so a `/docs/...` link to a page that doesn't exist reached the reviewer only if the model noticed. Smoke PR #21560 added one and the review never flagged it.

- `link-check-diff.py`: resolves every `/docs/` and `/blog/` link on added lines of content files (page file, PR-added file, or `aliases:` entry), appends dead ones to `.hugo-build.json.link_integrity` in the shape the composer already pre-stubs as 🚨 rows, and records a `link_check` block. Fence state comes from the file at head, since a hunk alone can't tell.
- Composer reads `link_integrity` even when Hugo was skipped; the brief's mechanical line reports the check's outcome, including "did not run".
- Workflow step after the Hugo pre-step; a crash never fails the review, it is named on the brief.
- Tests: pytest for the checker and the v3 card, `--self-test`, wired into `make test-review-pipeline`.


CamSoper:
@claude F1: fixed — image references and any target with a file extension are skipped as assets, not pages. F2: fixed — a local miss is probed on production with a HEAD request; only a 404/410 counts as dead, 2xx/3xx is live, and a timeout or 5xx is recorded as unknown and never reported (verified live: the author, series, cloud-rest-api, pkg, and tag paths from the finding all answer 200; the smoke PR's target answers 404) #update-review

PR #21578, opened 2026-09-12
Add Dev Center results to docs search as a tab

cnunciato:
Adds a Dev Center tab to docs search, alongside Docs, Registry, and Blog. The hourly `update-search-index` cron now fetches `/dev/search-index.json` (published by pulumi/marketing-web), normalizes each record onto our Algolia schema with `section: "Dev Center"` (dropping the non-searchable `content` field), and merges it into the `production` index. Because the cron uses `replaceAllObjects`, Dev Center removals drop out on the next run automatically, and a Dev Center fetch failure is non-fatal so it never blocks the docs/registry publish.

Dev Center blog posts and series are filtered out, since those are already surfaced in the Blog tab. Same-site Dev Center hrefs are normalized to relative paths, and `getItemUrl` now tolerates already-absolute (off-site) hrefs. The search tabs now render from the shared icon sprite (the same Phosphor set as the main nav), and the "All results" tab is relabeled "All".

Verified against a scratch Algolia index built from the live indexes:

image

Closes [MA-757](https://linear.app/pulumi/issue/MA-757/surface-dev-center-in-docs-search-results).


jeffmerrick (approved):
Looks good to me!

jkodroff (approved):
Nice improvement, and the comments are very human-readable.

PR #21630, opened 2026-09-15
docs: Update deployment settings docs for reworked UI

### LINKED ISSUE #20510: Update deployment settings docs after UI tweaks
Depends upon https://github.com/pulumi/pulumi-service/issues/46426

jkodroff:
Updates the deployment settings docs to match the stack Settings > Deploy page reworked in pulumi/pulumi-service#46723.

Fixes #20510

## Changes
- Renamed toggles: dependency caching, skip dependency installation, custom executor image
- Pulumi Cloud role select and Deployment runner pool selection (runner pools, permissions)
- Custom environment variables section, executor root path, Terraform remote-execution source/role
- New `"named"` option for `{{< pulumi-cloud >}}` (templates, lint, AGENTS.md); used to mark customer-managed deployment runner pools as Business Critical


CamSoper:
@claude Merged master in to clear conflicts in AGENTS.md and layouts/partials/cloud-availability-body.html. Kept master's V6 edition wording and its `safeHTML` pipes, re-added the `named` form on top, and piped the new named sentence through `safeHTML` too so it matches. Worth a look at the rendered callouts on the preview. #update-review

CamSoper (changes requested):
- `layouts/shortcodes/pulumi-cloud.html` L38: In the `named` example in layouts/shortcodes/pulumi-cloud.html, change "Business Critical" to "Enterprise". customer-managed-runners is `available_from: enterprise`, so that's what the callout actually renders.
- `content/docs/deployments/concepts/settings/dependency-caching.md` L28: Confirm the toggle label matches the shipped Settings → Deploy page, and reply on F1.
- `content/docs/deployments/concepts/settings/environment-variables.md` L32: Confirm that editing or renaming a secret variable requires re-entering its value in the new UI.
- `content/docs/deployments/concepts/settings/runner-pools.md` L29: Confirm the self-hosted empty-pool behavior: a link to create a pool, or an ask-your-admin message.
- `content/docs/deployments/guides/private-sources.md` L29: Confirm the label "Skip package manager dependency installation" (it's used in three places in this PR).
- `content/docs/deployments/concepts/settings/custom-executor-images.md` L28: Move the closing "For guidance on choosing between a pre-run install hook and a custom image…" paragraph above `## Executor root path`, so it isn't read as root-path advice.
- `content/docs/deployments/concepts/settings/runner-pools.md` L31: No change needed: the role section already lived on this page, so keeping it there (renamed) is fine.
- `content/docs/deployments/operations/permissions.md` L40: Replace "select" as a noun ("the **Pulumi Cloud role** select", "the **Deployment runner pool** select", "leave the select on") with "list".
- `content/docs/iac/get-started/terraform/terraform-remote-execution.md` L249: Confirm the exclusion list for Terraform remote-execution stacks, particularly runner pools on self-hosted installations, where there's no hosted pool.

jkodroff:
@claude #update-review

Pushed 5f5b50e addressing every item. Details below.

**Fixed**

- **F10** — right, and the rebase theory didn't hold up: `origin/master` has no "Business Critical" anywhere in `data/` or `layouts/`, and the `"named"` argument plus its doc block are *added* by this PR, so the string was branch-authored rather than stale-from-master. `customer-managed-runners` is `available_from: enterprise` (`data/pulumi_pricing.yaml:405`), so the example now reads "the Enterprise edition".
- **F6** — moved the "Deployment execution environment" see-also back above `## Executor root path`.
- **F7** — good catch, and the fix went further than moving the heading. The role control lives under **Advanced settings** in the console (`deploy-settings.component.html:779`, inside `advanced-settings__list` opened at `:634`), beside the controls that each already have their own settings page. So it got one: `content/docs/deployments/concepts/settings/pulumi-cloud-role.md` (weight 95, nav entry, `_index.md` listing). Removed the section from `runner-pools.md`, narrowed that page's `meta_desc`, repointed all three deep-links, and added the missing "open **Advanced settings**" locator the old prose lacked.
- **F8** — all five noun uses of "select" are now "dropdown" (house precedent: 30 uses of "dropdown" across `content/docs/` vs. 1 of "list").
- **F9** — the list was correct but **incomplete**. Verified every branch in `deploy-settings.component.html`: also excluded are skipping dependency installation (`:635`) and OIDC (`:838`), both now listed, and the role is tied to the custom-roles condition the console enforces (`:621`). The runner-pools exclusion you flagged is correct regardless of self-hosted — `:1009` gates on `!isTerraformRemoteExec && canConfigureAgent()`, which excludes it before self-hosted is considered.
- **Style** — both taken. The future-tense sentence moved to the new page in present tense, which makes the inline one-click suggestion on `runner-pools.md` L35 stale; please don't apply it. `private-sources.md` L68 now reads "so Pulumi encrypts it and keeps it out of logs".

**Confirmed, no change needed**

- **F1** — confirmed against the shipped Settings → Deploy UI: the toggle reads exactly **Cache dependencies to speed up deployments**.
- **F5** — confirmed: **Skip package manager dependency installation**, under **Advanced settings**. Settles all three occurrences.
- **F2** — confirmed in `pulumi-service`, and specifically the rename case you doubted. `environment-variables.component.ts:194` emits "Existing secret values are never shown. Re-enter the value to save this change.", and `environment-variables.component.spec.ts:297-305` renames `TOKEN` → `API_TOKEN` without re-entering a value, then asserts `getPendingEnvironmentVariables()` is `undefined` — a rename alone is rejected until the value is re-entered. The sentence is accurate as written.
- **F3** — confirmed in source. `deploy-settings.component.ts:879`: `shouldShowAgentPoolEmptyState = isSelfHosted() && agentPoolsLoaded && agentPools.length === 0`. `deploy-settings.component.html:1040-1054` then branches on `canManageAgentPools`: a "Create a deployment runner pool" link, else "Ask an organization admin to create one." That is exactly what the page describes.

One note for the human reviewer: `runner-pools.md` on master had a `## Role assignment` section, so its `#role-assignment` anchor no longer resolves. That anchor was already being renamed to `#pulumi-cloud-role` by this PR before the move, so nothing that worked at HEAD~ breaks that wasn't breaking already.

jkodroff:
@claude #update-review

Pushed 92dbcc8 for F11 — good catch, and you were right that a whole-page marker would over-claim.

`custom-roles` is `available_from: pro` (`data/pulumi_pricing.yaml:802-808`), so rather than hedge in prose I moved both custom-roles sentences into their own `## Custom roles` section carrying `{{< pulumi-cloud "custom-roles" "named" />}}`. That way the edition is derived from the pricing data rather than asserted by hand, per AGENTS.md ("not hand-written prose"), and it can't drift if custom roles move editions. The page's main subject — choosing a role — isn't gated, so it stays unmarked.

While moving the sentences I also kept the page's two closing paragraphs above the new heading, so they don't get captured by it — the same failure F6 flagged on `custom-executor-images.md`.

jkodroff:
Note for whoever reviews this — **do not apply the remaining inline style suggestion** on `content/docs/deployments/concepts/settings/pulumi-cloud-role.md` L22.

It was generated against the pre-F11 version of the page. Its body restores "you can create custom roles with specific permissions tailored to what the deployment needs to do" into the main flow — the ungated sentence F11 asked to gate, which now sits in the `## Custom roles` section behind the `pulumi-cloud` marker. Clicking it would regress F11.

The wordiness it actually targets ("accomplish") is already gone from the page, so there's nothing left to fix. Recorded as not-applicable rather than ignored.

CamSoper (approved):

PR #21634, opened 2026-09-15
Move AWS guides to Dev Center tutorials

cnunciato:
Deletes the nine AWS service guides now hosted as Dev Center tutorials at `pulumi.com/dev/tutorials/aws-*` (ecs, ecr, elb, vpc, eks, api-gateway, iam, lambda, cdk), adds 301s from their canonical URLs and every legacy alias via `scripts/redirects/aws-guides-redirects.txt`, and repoints the AWS Guides landing and provider-comparison pages at the new locations. The intro and Choosing a provider pages stay. 

⚠️ Head's up! ⚠️ 

This pairs with pulumi/marketing-web#165. Do not merge before that PR is both merged and deployed. 

Closes #216.


content/docs/iac/guides/clouds/aws/_index.md
```diff
@@ -35,21 +35,21 @@ The guides use the following packages:
 ## Getting started
 
 - [Choosing a provider](/docs/iac/guides/clouds/aws/providers/)
-- [Pulumi CDK Adapter](/docs/iac/guides/clouds/aws/cdk/)
+- [Pulumi CDK Adapter](/dev/tutorials/aws-cdk/)
```
jkodroff, line 38: We have a policy that says "no links from docs to marketing" (although links from marketing to docs are ok) that I think is a good idea. I don't know that we thought that much about links from docs to Dev Center.

I think it's fine conceptually, but maybe we should add some LLM instructions on this. Specifically:

1. Are links from docs to dev center ok?
2. If they are ok, what are the guidelines around where we should put those links? For example, I know there's examples for other clouds we list in this section. Should we send them to a landing page in Dev Center?

I won't block on this b/c I doubt these guides get a ton of traffic and this change is clearly not harmful, but we should have a think on this and develop a site-wide strategy so that the guidance is easy to follow.

jkodroff (approved):
Approved with comments.

content/docs/iac/guides/clouds/aws/_index.md
CamSoper: Yeah, I think linking from docs to tutorials and examples is fine. We shouldn't be linking to blog posts, product pages, or -- I think -- what-is docs. (That last one I could see going either way.)

cnunciato: Agree. Dev Center is intended to carry the same level of technical credibility and quality as the docs, so we should be comfortable linking between these. What Is pages, not so much today, but once we're done with the reboot we're starting this/next week, that may change, and we can reevaluate if we like.

Issue #21641, opened 2026-09-15
Phase 2 of the /pr-review queue: scheduled pr-queue.yml, published board, and queue:* labels the Sentinel can gate on


CamSoper:
## Where phase 1 leaves us

The refactored `/pr-review` runs `scripts/review-v3/collect.py` -> `analyze.py` -> `render.py` on the approver's machine (or web session) and publishes the board as a private Artifact. Every approver pays the collect cost, collisions between open PRs are only visible when someone runs it, and the verdict lives nowhere GitHub's own UI can show.

## Sketch

A scheduled workflow, `pr-queue.yml` (hourly, plus `pull_request` events on open / synchronize / labeled / closed), that:

1. Runs `collect.py` and `analyze.py` with the repo token (`gh` is on the runner, so the adapter's `gh` backend applies; `--owner any` so the board is everyone's). The judge step is a model call and stays out of the deterministic scripts; either run it once here as its own job (`claude-code-action`, writing `.pr-review-judgments.json` for `analyze.py --judgments`) or leave judge rows unjudged on the shared board and let the skill judge on demand.
2. Publishes `queue.json` and `board.html` to the review-evidence bucket under `pr-review/queue/` (same key scheme and degradation contract as `record-evidence.py`: `PR_REVIEW_EVIDENCE_URI` unset means local files plus a warning, never a failure), so the board is one URL the whole team reads.
3. Stamps labels from the verdict: `queue:stamp`, `queue:judge`, `queue:route:<role>` (marketing / docs-guild / tools), `queue:blocked`, and `queue:collides-with-#N` per overlap pair. Labels are the one place the verdict shows in GitHub's UI, and `queue:route:<role>` doubles as the request to that team.
4. Lets the Sentinel gate on collisions: a new gate (G6, "no unresolved collision") is red while `queue:collides-with-#N` names an open PR whose hunks overlap this one's, and turns green when N merges or the pair is re-analyzed as same-file. Directional link conflicts (adds links to an alias-only URL another open PR removes) should be red too, since they are the ones that shipped green three times in the triage session that motivated this.

The skill becomes a thin reader plus the act layer: `/pr-review` fetches the published `queue.json` (falling back to a local collect when it is older than the head it is looking at), renders, and acts. `act.py`'s per-PR preflight stays exactly as it is.

## Why not now

[...]

Issue #21640, opened 2026-09-15
resolve-handler: record actor==author dispositions as `author-accepted` and keep the ⚠️ row visible on the brief


CamSoper:
## Problem

`scripts/review-v3/resolve-handler.py` records every `/resolve F<n> accepted` (and `not-applicable` / `deferred` / `refuted`) with `actor: <login>` and nothing else. When the actor is the PR author, the finding stops blocking, the author card's header flips to "nothing blocks merge", and `review:no-blockers` lands -- the same green as a maintainer's answer.

That is how #21586 went green: `workprentice[bot]` accepted its own reviewer's finding (F44, six rewritten links keep fragments that match no heading on the destination) with "will fix in a follow-up" and no issue link. The queue in the refactored `/pr-review` (`scripts/review-v3/analyze.py`, reason code `self-accepted:F<n>`) flags this after the fact, but the pipeline itself still treats the author's word as the reviewer's.

The state block has no role field, so a reader has to re-fetch the PR author and compare logins to notice (`review_state.py` validates `actor` as a bare string; `resolve-handler.py` uses `is_author` only for the permission gate at the apply step).

## Proposal

- In `resolve-handler.py::handle`, when `actor == gh.get_pr_author()` and the disposition is one of `accepted` / `not-applicable` / `deferred` / `refuted`, record the disposition as `author-accepted` (a new value in `review_state.DISPOSITIONS`; `review-worklist.py` shares the tuple and needs the same addition).
- Sentinel G2 keeps treating it as answered (the author did answer), but the brief's `### ⚠️ Check these before approving` table keeps the row visible with an `author-accepted` marker, so the human approver sees what the author waved through. `compose-review.py` / `apply-update.py` are the writers; `sentinel._card_rows` already reads the table.
- The evidence object mirrors the value (bump `schema_version` per the README's transition rule, or add it to the disposition enum in `evidence-schema.json` and teach `validate-evidence.py`).
- `analyze.py` in the queue can then drop its login comparison and key on the disposition value.

## Not in scope

Changing who may resolve. The author may still answer; the point is that the answer stays visible to the approver instead of dissolving into a green label.

Filed from the `/pr-review` refactor PR; see that PR's description for the queue side.

PR #21642, opened 2026-09-15
Refactor /pr-review into a batch adjudication queue

CamSoper:
### Proposed changes

`/pr-review` was a one-PR, 10-step wizard that assumed `gh` and held a single PR across several AskUserQuestion round-trips. It is now a batch adjudication surface: `/pr-review` renders a queue of every open PR the approver can act on, sorted into **stamp / judge / route / blocked** with collision clusters pinned first; `/pr-review N` renders one PR as that row expanded. One data model (`queue.json`), one act layer, two renderings (board Artifact, `--terminal`). The skill keeps its name; the 35 references to it across workflows and docs still resolve.

Here is today's board, produced by the new scripts against the live queue: https://claude.ai/artifact/5q8ucqdDVwE5dVpi9KUWRg (every open non-draft PR with exactly one verdict and its reasons; collision clusters and the directional link conflicts on top; the PRs waiting on another reviewer folded into a compact list at the bottom).

**Deterministic scripts, all in `scripts/review-v3/` so `make test-review-pipeline` covers them** (no model calls; the model writes only the per-row judgment file the analyzer merges):

- `gh_client.py` -- GitHub adapter: `gh` subprocess, REST with `GITHUB_TOKEN`/`GH_TOKEN` (what a web session has), or a snapshot directory of endpoint JSON (tests, and the way to hand MCP-fetched data to the scripts). Owns the `author:app/<slug>` rewrite for GitHub App authors.
- `pr_review_config.py` -- `~/.pr-review.yml`: `me:` lanes, `stamp_max_lines`, `stale_date_days`, `link_fixes`. With no config file, lanes come from the approver's GitHub team memberships. Routing itself stays in `.github/review-routing.yml`.
- `collect.py` -- facts per open PR: files with patches, `mergeable_state` (re-asked while GitHub says `unknown`), check rollup, reviews, requested reviewers (bots dropped), the approver's own login and team memberships, the parsed pinned review via `review-worklist.py` (both surfaces, the only parser), `REVIEW_STATE`, triage's `TRIAGE_PROSE` comment, the reviewed-head SHA, preview URL and per-page links, and the trust axes / risk tier / AI-suspect logic ported from the four retired shell scripts. Cached per (PR, head SHA, updated_at).
- `analyze.py` -- one verdict per PR, a closed vocabulary of reason codes, and row actions. Cross-PR: collision clusters (union-find over shared paths; `overlap` when base-side hunks intersect, else `same-file`, with a merge order), directional conflicts (adds links to a URL that exists only as a Hugo `aliases:` entry while another open PR removes links from it), duplicate candidates, stale blog dates, self-accepted findings, stale reviews, stale brief summaries. The stamp bar is the one in the design doc, all gates required; heightened scrutiny caps at judge. **Handed off:** a PR whose requested reviewers include a human who isn't the approver is waiting on them, not on the queue; it keeps its verdict but leaves the groups, and collisions against it are advisory (`:theirs`).
- `render.py` -- board, detail, and terminal from the same `queue.json`. Server-side rendered through one `esc()`, queue inlined as JSON, never calls GitHub. The board opens with a folded manual and a **Do next** strip, then the rows grouped owner -> domain, then the collision detail. Reason chips split into a primary tier and a folded "why" tier, each with a tooltip explaining that code's own value; filter chips carry their counts; handed-off PRs render as a compact "Waiting on others" list. `--artifact` emits the fragment form the Artifact tool wraps.
- `act.py` -- plan -> preview -> execute. `--stamp` runs a per-PR preflight immediately before each squash-merge (head unchanged, `mergeable_state` in {clean, blocked} with green checks, no changes-requested review) and continues the batch after a refusal. `--route` requests review and posts the row's defects; `--request-changes` sends a judge row back to its author; `--chain C1`, `--unblock`, `--fix`, `--close`, `--refresh`, `--rerun`, `--render`, `--deploy`. Attribution footer on every posted comment except the approval body.

**Skill docs.** `SKILL.md` is rewritten around the five steps and carries the v3 guard verbatim, plus one section stating what makes a row "mine". `reading-the-board.md` is the manual for the rendered page. `action-menus.md` is the row action bar plus the bot-branch push policy. `action-preview-templates.md` and `execution-results.md` fold into the act preview and report and are deleted, as are the four shell scripts. The dashboard's PR section is retired in favour of a one-line pointer.

**Tests** -- `test_gh_client.py`, `test_collect.py`, `test_analyze.py`, `test_render.py`, `test_act.py`, `test_routing.py`, `test_sentinel.py`, plus each script's `--self-test`.

Verified locally: `make lint` clean, `make test-review-pipeline` green, a full live collect -> analyze -> judge -> render cycle against today's open PRs, and `act.py --dry-run` against a real plan.

### Since the first draft

Work from live cycles against the real queue, and the gaps they surfaced.

**Who can be asked for what**

- **Send-back only reaches an author who reads it.** Authors carry `can_revise`: people can, and so can an agent bot that answers reviews (WorkPrentice, Copilot). A PR opened by a workflow run -- pulumi-bot's content-review, glow-up and regen lanes, dependabot -- carries `author:generated` and offers "close it out" instead. `act.py` refuses `--request-changes` on such a row without `--force`, `--close N` alone closes with a comment built from the row's judgments, and a `request-changes` recommendation on a generated row becomes a `close`.
- **Approving says whether it merges.** A bot row leads with "approve & merge" and offers "approve, don't merge"; a human-authored row leads with "approve, no merge". `act.py` takes `N:merge` / `N:no-merge` per PR so a mixed batch stays one command.
- **Approving answers the findings.** `--stamp` posts one `/resolve F<n> <disposition>: <why>` per judged finding before the approval, so the review's own state records why each finding closed instead of the merge walking over it.

**Routing and the merge gate**

- **An unreadable team lookup is no longer a missing team.** A 403 on `orgs/{org}/teams/{slug}` routed every lane to its SLA person and chipped `route:no-team`. It now routes to the team the config names and chips `route:team-unverified`.
- **No `~/.pr-review.yml` takes its lanes from your GitHub teams.** Every routing team you belong to contributes the subjects its role owns, so a first run shows route rows with no local setup.
- **blog routes to `pulumi/docs-blog-review`.** The team exists now, so Sentinel requires a docs-blog-review approval on blog PRs rather than a docs-marketing-review one.
- **A change with no team gate at all is any approver's** (`gate:none`). A mechanical docs or blog edit resolves to no required role, so Sentinel passes G3 with "no human approval required" and nobody is waiting; the queue was routing those rows anyway.
- **Any review team may approve a link-only sweep** (`gate:any-team`). New `link_only: { approval: any-team }` in `.github/review-routing.yml`: a diff whose every changed line differs only in a link still needs a human, but not a particular lane's human, because the question it raises is whether the target resolves and still says what the sentence claims. The link-only bar moved into `sentinel.py` so the gate and the queue cannot disagree about what a sweep is. `approval: lane` restores the old behavior, and the mechanical bar is untouched.

**Making the page readable**

- **The board carries its own manual**, folded at the top, and the same content ships as `pr-review:references:reading-the-board`.
- **Do-next cards say what they do**, name their PRs, and press those rows' own buttons, so a card and a contrary row decision can never both be lit and the composed command cannot contradict itself. A consolidation, whose command carries a written reason no row button has, tags its rows instead.
- **Disposition badges say why, not who**: "not a real issue", "fair, not blocking", "doesn't apply", "needs the author", "no author to ask".
- **Tooltips describe their own value**, not the family it belongs to, everywhere on the page: chips, verdicts, buttons, filters, folds. A test walks every title on a rendered board and fails any that is empty, too thin, or the same as the text it is attached to.
- **A row with no findings says so and shows the diff.** It names why nothing reviewed the PR (trivial, frontmatter-only, oversized, draft, bot author), and a diff small enough to read renders in place rather than sending you to GitHub for ten lines.
- Filter chips carry their counts, blocked rows with a mechanical unblock get their own Do-next cards, and the reviewer's guide's own summary sentence is no longer truncated.

### Second pass over the board

A working session driving the live queue, and the things that broke under it.

- **The filter chips were dead.** The expand/collapse lever borrowed the filter chip's class, so the setup loop read `filters[undefined]` and threw before `apply()` ran: every row rendered whatever its chip said, and clicking a chip did nothing. Filter loops select on `[data-filter]` now, with a test that fails on a bare `.fchip` selector.
- **Every verdict starts visible.** The stampable rows arrive with their approve-and-merge button selected, so hiding them by default meant the composed command acted on PRs the page never showed. That chip is also called `stampable` now, since it names rows rather than a saved set.
- **One selection per decision.** The same button can render twice (compact row and expanded card); selection is keyed by PR plus command and painted on every button carrying that key. A collision chain's follow-up has no row button of its own until the row is actually stuck, so the card claims it instead of naming a target nothing can press -- which is why the card could never light.
- **An unjudged finding reads like a judged one**: the claim on top, a badge for where it stands (`probably not real`, `worth a look`, `nobody has ruled`), one sentence of the review's reasoning, the diff lines, and the review's full note folded away. It used to dump the whole paragraph under a label that told an approver nothing.
- **A workflow-authored row says the send-back is dead, not the fix.** `author:generated` meant "close it out" was the only move; push to the branch, apply the drafted fixes, or ask Claude on the PR are all still open, and only the changes-requested review has no audience.
- **Folds and their defaults.** The reviewer's guide, preview links, diff quotes and the collisions section start open; the manual and the folded reasons start closed. One lever in the filter bar moves every panel on the board, each row carries the same lever for itself, and the two folds whose `display:flex` summary drops the browser's marker draw their own chevron.
- **The header says what it did**, not which flags it took: `showing: rows in your lanes` and `your lanes: every lane (nothing pinned)` when there is no `~/.pr-review.yml` and team memberships are unreadable, with tooltips naming the cause and the fix.
- **Tooltips survive a code that carries detail.** `checks:red:sentinel` matched no help key and rendered bare; the lookup keys off the state and names the failing checks in the sentence, and the test walks the multi-segment forms.
- **A row summary is never template boilerplate.** A generated PR body's first prose line is often a machine note to a machine ("do not edit it"); the summary prefers the body's own "Fixes applied" section, table row included.
- **Planning is read-only, so the skill no longer asks permission to plan.** One question, after the preview, about doing the writes.

Rebased onto `master` at 6003efcde3, which carries #21674's sentinel work; `make test-review-pipeline` and `make lint` are green on the combined tree.

n/a

### Related issues (optional)

Filed, not built:

- #21640 -- `resolve-handler.py` should record actor==author dispositions as `author-accepted` and keep the ⚠️ row visible on the brief.
- #21641 -- phase 2: a scheduled `pr-queue.yml` that publishes `queue.json` and the board to the evidence bucket and stamps `queue:*` labels the Sentinel can gate on.

https://claude.ai/code/session_01DmcivKmxnxxhAxSsQijVvf

---


CamSoper:
**Sentinel check is red, and it isn't this PR's diff.** Recording what happened so it's on the PR rather than only in a session.

The `sentinel` job runs `scripts/review-v3/sentinel.py` from the **default branch**, never from the PR head -- that's deliberate (`pull_request_target` + checking out PR code would be a credential-exfil vector, and `test_sentinel.py` asserts the workflow never references a head ref). So nothing in this diff is running there.

What changed was the `sentinel:preview` label. In report-only mode that label is what tells the evaluator to also maintain the pinned gate-status comment, and that write path crashes:

```
subprocess.CalledProcessError: Command '['gh', 'api', '--method', 'POST', 'repos/pulumi/docs/issues/21642/comments', '-f', 'body=<!-- SENTINEL_STATUS -->...']' returned non-zero exit status 1
```

Every gate had already been evaluated at that point. The crash discards the verdict and fails the check, and because `check=True` raises with argv and nothing else, the log carries 2KB of comment markdown and no reason. It reproduced identically on two heads (`d0def344`, `b583a525`), so it isn't a flake.

**What I did**

- Removed `sentinel:preview` so the check stops failing. The preview comment is worth having, but not at the cost of a permanently red check on a PR that's up for review. One click to restore once the fix below is on master.
- Ported the fix into this PR, since it already touches `sentinel.py`:
  - `Gh._run` now surfaces what the API actually said, so the next failure names its own cause.
  - The pinned status comment is written through a wrapper that reports a write failure instead of raising. The check run is the merge gate; the comment is a courtesy on top of it, and it shouldn't be able to fail a verdict that's already computed.
  - Tests for both; `make test-review-pipeline` and `make lint` green.

**What's still unknown:** why the POST returns non-zero. The job has `issues: write` and the same token's reads succeed earlier in the same process, so it's most likely a 403 or a body the API rejects -- the stderr that would say which is exactly what the old code threw away. The first failure after this merges will name it.

**One datapoint since:** this very comment was posted through the same endpoint with the same body shape and it went through, so the API and the markdown are fine -- which points at the workflow's `GITHUB_TOKEN` rather than at the request. Worth checking whether the org's default workflow permissions still let a `pull_request_target` job write issue comments here.

**Superseded by #21675, and my fix is reverted here.** That PR has the actual root cause: `POST /repos/{o}/{r}/issues/{n}/comments` on a PR number is governed by the **pull_requests** scope, not `issues`, so the workflow needed `pull-requests: write` -- which also explains why the same request succeeded from a user token. It carries the stderr surfacing too, fixes the identical latent bug in `staging-deploy-pr.yml`, and handles a consequence I had missed: `get_team_membership` catches `CalledProcessError` to read a clean 404 as "not a member", so raising a new exception type out of `_run` -- which my commit did -- would have turned every non-member into a G3 error on every governed PR. Reverted in 4e2ef46; `sentinel.py` here is back to just this PR's `link_only`/any-team work.

---

CamSoper:
@claude I pushed fixes for F2 and F4, and I'm accepting F3 as-is. #update-review

**F2 — fixed.** Correct catch, and worth naming the cause: #21674 added `blog:` to `teams:` and `sla:` on master, this branch added the same two keys, and the rebase applied both with no textual conflict. Duplicate YAML keys, last-one-wins, same value on both sides -- right by luck. Both of this branch's copies are gone.

**F4 — fixed.** `/.pr-review-evidence/` and `/.pr-review-state/` now sit in the same ignore block, with a comment pointing at the two scripts that write them. Same argument the block already makes: they land in the repo root before the `make lint` re-gate, and an unignored one fails `prettier --check .`.

**F3 — accepting as-is.** The mismatch is real: `staging-deploy.sh` writes the pending status at the `--head-sha` it was handed, while the dispatched run resolves at its own `github.sha`. I'm leaving it, because the alternative is worse. The status should describe *what was actually built*, and the run builds the branch tip it finds when it starts -- pinning the final status back to the dispatch-time SHA would assert a deploy of a tree that run never built, which is exactly the lie Sentinel G4 exists to catch. The cost of leaving it is a pending status stranded on a superseded commit, which nothing reads: the merge box shows statuses for the current head only, and that head gets its own dispatch, its own pending, and its own resolution. A stale marker on a dead SHA beats a confident marker on the wrong tree.

---

CamSoper:
@claude I pushed a fix for F5 #update-review

Correct, and the sloppiness was mine: `record-evidence.py`'s own contract is "write two local files, ALWAYS, regardless of whether S3 is reachable," so `.pr-review-evidence/` appears on every run. Only `sla-sweep.py`'s `.pr-review-state/` is conditional on no bucket being configured. The comment now says which is which instead of flattening both into the same condition.

---

PR #21645, opened 2026-09-15
Add the agentic infrastructure operating model whitepaper

jeffmerrick:
### Proposed changes

Adds `content/whitepapers/agentic-infrastructure-operating-model.md` with the figures, PDF download, and reusable shortcodes it needs.

- The figure SVGs have their `font-family` rewritten to `'Monaspace Neon Var'` and `'Inter'`, since the theme registers the variable face under the `Var` name and Figma pins every `<tspan>` to an absolute `x`/`y` - without the rewrite the labels fall back and drift out of their boxes.
- `svg-figure` inlines the SVG rather than loading it through ``, as an ``-loaded SVG renders in an isolated document with no access to the page's `@font-face` rules. Named `svg-figure` and not `figure` because Hugo ships a built-in `figure` that blog posts already use.
- `pullquote` is restyled at the component level, so `/what-is/what-is-ai-infrastructure/` and `/blog/seven-rules-ai-native-software-factory/` change too.
- The whitepapers section no longer uses the template-page component system. `.template-sidebar-content` and `.template-hero--centered` are gone from `_templates.scss` (both were whitepapers-only), and what's left lives in `theme/src/scss/_whitepapers.scss` next to `_case-studies.scss`.
- `template-counter-cards.html` takes optional `align` and `accent` params. Defaults are unchanged, so `/product/infrastructure-as-code/` renders as before.
- Meta images: the `whitepapers` section is enabled, and the `title` template takes an optional label pill. `TEMPLATE_REVISION.title` bumped so only title cards re-render.
- `static/logos/tech/wiz.{svg,png}` deleted - grep finds no references - and `content/case-studies/wiz.md` points `customer_logo` at the new `.svg`, which the homepage card resolves through.
- `/whitepapers/` leads with a full-width card for the newest paper, driven by two new optional front matter keys: `cover:` for the preview art and `card_label:` for the overline.
- The other two whitepapers pick up section rules, and the enterprise paper's executive summary moves into a `panel`.

Known gaps

- The PDF is ~12 MB.
- `yarn.lock` carries an unrelated transitive bump that predates the branch.


cnunciato (approved):
This doc's been reviewed by several folks internally, so I'm approving on their behalf. We can address necessary feedback/issues in a follow-up.

PR #21647, opened 2026-09-15
docs: self-hosted changelog entry for the OpenSearch 3.7 upgrade

aacotroneo:
Adds a **September 2026** entry to the [self-hosted changelog](https://www.pulumi.com/docs/administration/self-hosting/changelog/) for the OpenSearch 3.7 upgrade landing in [pulumi-self-hosted-installers#1350](https://github.com/pulumi/pulumi-self-hosted-installers/pull/1350), which moves every installer variant to 3.7.0 — the Docker Compose quickstart, the EKS and GKE Helm deployments, the ECS AWS-managed domain, and the AKS component.

## Framing

Flagged as a breaking change because the installer **default** moved, but the note leads with the mitigation rather than the hazard: the service stays fully backwards compatible with older OpenSearch versions, so operators can update the service and keep pointing at an existing 2.x cluster. Nobody is forced to upgrade on this release.

Per-installer upgrade steps are deliberately **not** restated here — they differ meaningfully per variant (managed AWS domain vs. Helm vs. plain container) and #1350 documents each one in its own installer's README. Duplicating them in the changelog would guarantee drift.

Uses the `{{< notes type="warning" >}}` convention already established by the April 2026 SSRF and November 2025 BYOK entries.

## Follows on from

[#21346](https://github.com/pulumi/docs/pull/21346) (merged), which updated the supported-version guidance to "v2.x or v3.x" and recorded that Pulumi Cloud runs v3.7 — the compatibility statement this entry relies on.

Note the entry goes under the existing `## 2026` heading, above `### August`; the `### September` further down the file belongs to `## 2024`.


aacotroneo:
@claude review

aacotroneo:
@claude what do you think now?

CamSoper (approved):
LGTM.

CamSoper:
@aacotroneo 

> what do you think now?

~~To get claude to update the pinned review, you must include `#update-review` in your comment. It enables a special workflow that gives claude extra permissions. See below for an example.~~

Never mind, I see now that the v3 review errored. I've requested a new one. When it shows up, the review will have the appropriate instructions.

CamSoper:
@claude #update-review

aacotroneo:
> @aacotroneo
> 
> To get claude to update the pinned review, you must include `#update-review` in your comment. It enables a special workflow that gives claude extra permissions. See below for an example.~
> 

Thanks @CamSoper !! I didn't run that; I was kinda wondering what the proper way was. Thanks!

PR #21648, opened 2026-09-15
List pulumi-context-api on the Agent Skills page and in the well-known index

lblackstone:
The `pulumi-context-api` skill (pulumi/agent-skills#56) was missing from the Agent Skills page and from `.well-known/agent-skills/index.json`, so agents that discover skills through either had no way to find it. This lists it in both, and also adds `pulumi-migrate-from-discovered-stack`, which the page listed but the index did not.

Every digest in the index is refreshed to the SHA-256 of the current `SKILL.md` on agent-skills `main`; the existing values matched no revision of the files they point at.


lblackstone:
Closing for now — parking this work. The branch is retained, so this can be reopened unchanged if we come back to it.

🤖 AI-assisted (Claude Code)

PR #21668, opened 2026-09-16
Restyle the pricing page support plan table

jeffmerrick:
### Proposed changes

Screenshot 2026-09-16 at 11 03 55 AM

Restyles the support plan comparison on `/pricing/` to match the edition comparison table above it.

* `support_plans` benefit cells in `data/pulumi_pricing.yaml` are bools instead of the strings `Included` / `Not included`, so they render through `pricing/value.html` like the edition cells do.
* `eligibility_note` goes through `safeHTML` so it can carry an `&nbsp;` - note: the field is raw HTML now, so an `&` in one has to be escaped.
* Below `lg` the table reflows to one block per benefit rather than scrolling horizontally. `lg` and not `md` where the edition table breaks, since six columns of prose need more room than six columns of check marks.
* `.table-plain` added to `theme/src/scss/_tables.scss` to opt a table out of the site-wide gridlines and gray header fill. The support table is its only consumer.
* Dashes in both tables lightened from `text-gray-primary` to `text-gray-600`, the lightest step on the scale still over the 3:1 contrast minimum for a non-text state indicator.
* The section heading uses `heading-2`, and the FAQ heading drops to the same size.

At 1024px the five plan columns are about 155px wide, so longer cells like "Chat, support portal, email, and Slack" wrap to two or three lines.


jeffmerrick:
@claude I pushed fixes for all findings #update-review

kramhuber (approved):

cnunciato (approved):
Nice. The mobile treatment looks great!

PR #21674, opened 2026-09-16
Review v3: make the merge gate satisfiable — reviewer routing, staging evidence, authorized waives, pinned gate status

CamSoper:
### Proposed changes

Everything here came out of the report-only Sentinel dry run over 14-16 Sep, where **every governed PR measured would have been blocked** — mostly on gates whose inputs nothing was producing. Four changes, all aimed at the same thing: make the gates possible to satisfy before anyone flips `REVIEW_V3_SENTINEL` to `'1'`.

**1. Reviewer routing gets a token that can actually request teams.** The routing step has existed since Phase 3 behind `REVIEW_V3_ROUTING`, so nothing ever exercised it. It would not have worked when flipped: a *team* review request resolves `org/slug` against the org on the `requested_reviewers` endpoint, and the repo-scoped `GITHUB_TOKEN` can't -- the same reason the Sentinel mints `PULUMI_BOT_TOKEN` for its team reads. Every `--add-reviewer` would have fallen into the `|| echo "::warning::"` branch and routed nobody, silently, on every PR. Now mints the org-scoped token from ESC (the job already had `id-token: write`) and logs what it requested, or what it *would* have requested when the token is missing.

**2. G4 verifies the deploy, not the report of it.** The commit status is a separate API call after `gh run watch` returns, so a cancelled runner, a lost token, or a hand-run deploy that never went through `/deploy-staging` all leave a green deploy with no status. `_staging_evidence` now accepts either the `staging/pulumi-test-io` status **or** a successful `testing-build-and-deploy.yml` run at the same head SHA. A failed run-history read degrades to "no evidence" (red) rather than `action_required` -- red is already the conservative answer, and escalating would misreport an API hiccup as a corrupt PR.

**3. Staging evidence now has a supply.** New `staging-deploy-auto.yml`: every same-repo, non-draft PR that `route-pr.py` says needs staging evidence gets a deploy dispatched on open/push, skipping when the head already has one. Shares the `staging-stack` concurrency group with the comment lane, and both now call `scripts/review-v3/staging-deploy.sh` so the two can't drift on what counts as evidence. `/deploy-staging` goes back to being what it reads like: a retry. Neither lane checks out or runs PR code -- the checkout is pinned to the base default branch, same invariant as the Sentinel's.

**4. `review:waived` is authorized, not merely applied.** A waive skips every gate but G4, so it has to be at least as privileged as approving something. The actor must be an **active member of a team in `teams:`** -- any of them, not just the one the PR routes to, because scoping it to the required team would make a waive exactly as hard to get as the approval it bypasses, killing the case it most needs to cover (the required approver is the PR's own author). Fails closed on an unreadable actor, a failed lookup, or a non-member, and refuses **out loud** in the summary rather than ignoring the label.

**5. A pinned gate-status comment.** One `<!-- SENTINEL_STATUS -->` comment per PR: a row per gate with its state and, when red, that gate's own remediation text. The body is a pure function of the verdict -- no timestamps, no run ids -- so an unchanged verdict produces a byte-identical body and the PATCH is skipped. Enforcing mode maintains it on every governed PR; report-only mode only on a PR labelled `sentinel:preview`, so the dry run stays invisible to anyone who didn't opt in.

### Testing

- 12 new `test_sentinel.py` cases: the run-history witness (accepted / failed-run ignored / success-among-several / lookup failure blocks rather than errors), waive authorization (any routing team, non-member refused, unreadable actor and failed lookup both refused), and the status comment (upsert, no churn on an unchanged verdict, single-line table cells, report-only banner).
- `make test-review-pipeline` passes in full.
- `make lint` could not run in this environment (`make ensure` needs Hugo and Node 24, neither present). The markdown touched is meta files, which AGENTS.md exempts from formatting rules, but CI is the real check here.

n/a

### Related issues (optional)

n/a

---

**Before merging, create the new label** (triage/Sentinel label calls are batched and `gh` rejects the whole call if one name doesn't exist):

```
gh label create "sentinel:preview" --color d4c5f9 --description "Preview the Sentinel gate-status comment while it is still report-only"
```

**Still not addressed, deliberately:** the approver pool. `pulumi/docs-tools` has one member, so infra PRs authored by that member still have no eligible G3 approver — a waive is now the only path, which is exactly the case the "any routing team" rule above keeps open. That's a people change, not a code one.

https://claude.ai/code/session_019ZBZZNG6LNiMMLh14ofeJ4


CamSoper:
@claude #update-review

CamSoper:
@claude #update-review

Fixed all three. Pushed `5c78594b`.

**F4** — confirmed at the source and reproduced before fixing. Built a real sparse clone of the repo and ran `classify_mechanical` over a one-line link addition pointing at `/docs/iac/concepts/stacks/`:

```
FULL   -> (True, [])
SPARSE -> (False, ['content/docs/foo.md: added link does not resolve (/docs/iac/concepts/stacks/)'])
```

Exactly as described, bogus reason text included.

Of the two remedies you offered I took neither literally. Adding `content/` to the sparse list would put 860 MB back and undo the optimization; dropping the repo root would skip link validation entirely, which fails *open* on link validity — a PR adding a genuinely broken link could then classify mechanical and skip review. Instead `_resolve_content_target` now probes `git cat-file -e HEAD:<candidate>` before giving up. `sparse-checkout` scopes the worktree, never the object store, so the commit's trees are there in both layouts; the probe is exact rather than heuristic, so it runs ahead of the existing `git grep` fallback. Both layouts now agree, and a target that genuinely doesn't exist still fails in both.

You were also right that `test_sparse_checkout_covers_everything_the_evaluator_loads` could never have caught this — it scrapes `_load`/`_HERE` literals out of `sentinel.py`, and this is a runtime *data* dependency reached through the classifier. Added `test_condition_links_under_a_sparse_worktree`, which builds an actual sparse clone and asserts both directions. Worth noting that registering it in `main()`'s explicit list was load-bearing: a test defined but not added there is silently skipped, which is the same hazard `test_sentinel.py`'s harness docstring documents.

**F1** — correct, and the fix is yours verbatim. `main()` upserts whenever `--report-only` is absent, and `render_status_comment` has explicit branches for a non-governed automation lane and for drafts, so "every governed PR" was wrong. Now reads "every PR the Sentinel evaluates".

**F3** (from the reviewer's card) — also correct, and a bug I introduced while documenting the very behavior it breaks. `concurrency: staging-stack` was at workflow level, so every PR event joined the shared queue before the routing step could rule it out, and a docs-only push could sit pending and displace a queued infra deploy that GitHub then drops silently. Split into a `route` job with no group and a `deploy` job that carries it, gated on `needs.route.outputs.deploy`. Only a run that will actually deploy takes a slot now.

`make test-review-pipeline` passes in full; the mechanical-bar suite is at 56 checks.


---

PR #21676, opened 2026-09-16
Review v3: don't drop a superseded review on the floor

CamSoper:
### Proposed changes

Push while a v3 review is in flight and the review is lost silently, leaving a `review:in-progress` label nothing will ever clear. Observed on #21642 (run 35157124406, 2026-09-16): the model ran 22:21-22:25, two pushes landed at 22:23, the publish job's freshness guard correctly refused the stale handoff at 22:26, and every step after it was skipped. The guard cleaned up the check-run and the spinner but not the label, and nothing restarts a review on push (only open, ready_for_review, and an explicit dispatch do). The PR sat there until someone ran the workflow by hand.

Three commits, each revertable on its own:

**Reset the label and re-dispatch when a publish is superseded.** The decision moves into `scripts/review-v3/handoff_guard.py` (publish / redispatch / rest, mirroring `publish_guard.py`) and the workflow owns the side effects. A superseded run now rests at `review:stale` -- the one state whose documented meaning is actually true here, and the one every reader (pr-context's already-reviewed skip, address-review, pr-review, the reconcile cron) already treats as "needs a refresh, don't auto-loop"; a cleared PR would be indistinguishable from one triage never touched. A new `redispatch` job then starts a fresh review at the live head. It sits outside the `claude-review-<pr>` concurrency group on purpose: the run it dispatches enters that group with cancel-in-progress and would cancel the job that dispatched it. It also waits until the head has been quiet for three minutes first, so a retry fired into a push train doesn't just get superseded again. A `supersede_depth` dispatch input caps the chain at two automatic retries; after that the guard rests, rewrites the spinner to say how to refresh by hand, and the check-run summary says the same. Also deletes the `#new-review` confirmation comment on this path (it was only cleaned up on the publish path) and stops the check-run summary from claiming "a newer run owns this PR" when none did.

Cost: one full review per supersession, never per push. Worst case per ready-transition is three full reviews. The alternative -- re-trigger on every `synchronize` and let cancel-in-progress sort it out -- would burn a partial model run for every push in a train, and collides with `mark-stale` / `auto-refresh`, which already own that event. Rejected, with the reasoning in the job's design note.

**Stamp the handoff with the SHA the run actually checked out.** The handoff carried the live head as read by pr-context, while the checkout used `workflow_run.head_sha` / the dispatcher's `head_sha` input. When a push landed between trigger and pr-context those diverged, and the guard (live head vs. handoff SHA) would pass a review of commit A off as a review of commit B. Now the handoff carries the checked-out SHA, so the guard answers the question it claims to. Tightens the guard; does not weaken it.

**Reconcile cron sweeps orphaned `review:in-progress`.** The cron only reconciled the two terminal-success labels, so any run that died without finalizing (job-level timeout, lost runner, this bug) left in-progress forever. Both lanes post a `CLAUDE_PROGRESS` spinner before setting the label and delete or rewrite it on every terminal path, so an in-progress label 30+ minutes old with no live spinner is an orphan: move it to `review:stale`. Run-list mapping is deliberately not used (dispatched runs report the default branch as head and carry no PR number).

Tests: `scripts/review-v3/test_handoff_guard.py` pins the decide/settle truth table plus the workflow invariants in the `test_sentinel.py` style: every publish step still gates on the guard, the redispatch job has no concurrency group and never touches PR code, the claude-review group is unchanged, depth flows through the dispatch inputs, and `claude-new.yml`'s dispatch still satisfies the input set. `make lint` and `make test-review-pipeline` are green. The Vale mirror was not consulted because no prose content changed; the brand MCP was not needed.

Docs updated: the publish job and redispatch job design notes, the guard step comment, `.github/labels-pr-review.md`, `scripts/review-v3/README.md`, and the label table in `address-review/SKILL.md`.

n/a

### Related issues (optional)

Observed on #21642. Builds on #21674 and #21675 (both merged).

https://claude.ai/code/session_01MyBVMH3g8HZzqVmWfRNqLm

---
_Generated by [Claude Code](https://claude.ai/code/session_01MyBVMH3g8HZzqVmWfRNqLm)_


CamSoper:
@claude #update-review

Fixed F2 in 210a2e2c. You're right that the default `run` shell has no pipefail -- I'd assumed it did. Rather than fail toward publishing, the step now sets `pipefail` and refuses any action that isn't publish / redispatch / rest, so a broken guard fails the step and the existing publish error trap takes over (review:error, red check, "errored while publishing" spinner). Publishing on a guard we couldn't run would mean stamping cards with a SHA nobody checked; loud is the safer default here. `test_publish_guard_fails_loudly_when_the_script_does` pins both the pipefail and the case-guard.

---

PR #21703, opened 2026-09-17
Simplify event linking with a single event_url field

cnunciato:
Collapses the event linking model to a single frontmatter field. An event hosted elsewhere now sets `event_url`, and its shape decides everything: an `https://` URL links out to a third party, a `/` path points at another Pulumi page. Its presence drives the list card and RSS link, meta-refreshes the generated `/events/<slug>/` stub to the real destination, and marks that stub `noindex` (and drops it from the sitemap), since the stub is only a build by-product, never a real page.

Removes `url_slug` (which never controlled the slug), the `external` and `internal_landing_page` booleans, and the `block_external_search_index: false` default. Content is migrated in the second commit; same-domain absolute URLs are normalized to `/` paths (so the Neo launch pages become internal landing pages, which is what they always were).

Supersedes #21594. The schema loader now skips structured data only for `event_url` stubs and emits it for every real event page, so no indexable event loses its Event JSON-LD — and the four previously-misconfigured `external: true` + indexable pages (two Neo, two third-party) are no longer indexable dead-end stubs.

Behavior change worth a look: internal landing-page stubs now meta-refresh to their target instead of rendering a thin noindexed page. Verified against a `hugo --minify` build — external stubs redirect + noindex + no schema, internal stubs redirect to their `/` path, normal events keep schema and stay indexed, and the sitemap excludes every `event_url` stub.

Draft for design review before the field rename churns 250 files.


jeffmerrick (approved):
Yep, this makes more sense, looks good!

PR #21705, opened 2026-09-17
Replace /case-studies with /customers, backed by a customer registry

jeffmerrick:
### Proposed changes

`/case-studies/` is now `/customers/`: a grid of every customer Pulumi shows publicly, filterable by industry, with case studies linked from the cells that have one.

- `data/customers.yaml` is the registry every customer surface reads (the grid, logo banners and grids, testimonials, case-study cards, meta images). 97 customers, 34 with a case study, ordered most recognizable first. The build fails on an unknown id, an industry not in `data/customers_industries.yaml`, a malformed color, or a missing `<id>.svg` / `<id>-on-dark.svg` pair.
- Case studies declare `customer: <id>` and get name, URL, and logos from the registry; the retired per-page keys (`customer_logo`, `logo_bg_color`, etc.) are `make lint` errors. `industry:` stays on the page because Hugo's taxonomy reads front matter, and lint fails when it disagrees with the registry.
- Industries: `blockchain` is renamed `technology` (its old term URL redirects), `energy` ("Energy & utilities") is new, and every industry name is now sentence case.
- Every moved page and industry term page carries an alias for its old `/case-studies/` URL.
- `content/industry/<id>/_index.md` stubs exist so every `/customers/industry/<id>/` page renders whether or not a case study is filed there; lint keeps the directory equal to the data file. An industry with no customers is noindexed.
- A registry entry can carry a canonical `quote:`. On the grid that customer gets a double-width cell showing it, and a testimonial section that names the customer without its own quote falls back to it.
- Every customer logo on every page links to the case study when there is one and to `/customers/` otherwise. The homepage logo section is a static grid of twelve (`logo_grid` template type) instead of the 35-logo carousel.
- Logo lists elsewhere keep only customers in the registry, so `/proserv/ai/` drops ten small AI logos. The `customer_logos` blocks on the what-is and topics pages were front matter no layout read, and are removed.
- 12 customers are still PNGs (`raster_holdouts` in the registry); that list is meant to only shrink.
- `scripts/customers/trim-logo.mjs` trims logo viewBoxes so every logo renders at the same optical area, and `derive-on-dark.mjs` makes a white variant when a brand ships none.
- The old `static/logos/customers/` files and the per-customer `layouts/partials/customer-logos/` partials are deleted; nothing references them. Same for the stacked carousel and the layouts no content used (`ai-solutions`, `pulumi-up-2022`, `awsx/aws`, `product/secrets-management`).
- Case-study social cards regenerate with the registry logo (template revision bumped), and the industry cards say "Customers".
- `data/footer.yml` and `data/header_nav.yaml` now point at `/customers/`, which also changes the Dev Center's nav and footer through marketing-web's sync.
- The docs-review skill and `triage-classify.py` route `content/customers/` and both data files to the blog review domain.
- `yarn.lock` picks up the `css-select@6` entries the svgo bump left out.

### Related issues (optional)

Resolves MA-766


data/header_nav.yaml
```diff
@@ -103,11 +103,11 @@ items:
             description: Security, compliance, and support for teams
             icon: buildings
             track: header-enterprise-solutions
-          - label: Case studies
-            href: /case-studies/
+          - label: Customers
+            href: /customers/
             description: How Snowflake, Mercedes-Benz, and others use Pulumi
```
cnunciato: Not sure I _love_ this, but it feels both correct and a little less boring (to me at least). 😄 

```suggestion
            description: Get a glimpse of who's using Pulumi and hear their stories
```

jeffmerrick: Good catch!

cnunciato (approved):
This is great. Amazing improvement. ✨

cnunciato (approved):
🚀

PR #21708, opened 2026-09-17
Sentinel: close the gate holes, separate ownership from classification, fix the deploy queue

CamSoper:
### Proposed changes

Started as four routing fixes. Then subagents reviewed the Sentinel end to end and found the gates were secured **inversely to how much they're relied on** — so this also carries the fixes for that, plus the review's own findings and a `/simplify` pass. Absorbs #21734 (closed; commit `77442fd7` preserved via merge) and #21739.

Nothing here is enforcing yet (`REVIEW_V3_SENTINEL` is `report`), which is the only reason it's one PR instead of an incident.

---

## Part 1 — Routing: one field was answering several questions

```
matrix     -> what kind of file is this   (which review criteria apply)
overrides  -> who owns it                 (who gets asked)
approval   -> who may clear the gate      (who can approve)
staging    -> what must be demonstrated   (already split, #21701)
```

#### 1a. Ownership and classification were the same field

Two live misroutes reported the same morning, both reproduced before changing anything:

- **#21723** -- a Hugo bug blanking the Responses section of the [Cloud REST API reference](https://www.pulumi.com/docs/reference/cloud-rest-api/deployments/). `layouts/` is `domain:frontend` → marketing "who own how the site looks". True of marketing chrome, false of the templates that render the API docs. Went to `docs-marketing-review`, who can't review it.
- **#21718** -- drifted samples on a Get Started page. `content/docs/` is one subject with one owner → `docs-guild`. Get Started is part of the marketing funnel.

Same shape: **a subject whose path prefix spans more than one owner.**

`overrides:` is an ordered path→role list consulted before the matrix, per path, first match wins, with a **required `why`**:

```yaml
overrides:
  - paths: [layouts/partials/openapi/]
    role: tools
  - paths: [content/docs/**/get-started/]
    role: marketing
```

**#21723 proves this belongs at the ownership layer and not in the classifier:** its domain was *correct* -- it's a Hugo template and genuinely wants the Hugo/dark-mode criteria -- and only the approver was wrong. Reclassifying it would have fixed routing by breaking the review.

**The rule: `classify_path()` says what a file IS. Don't bend it to say who owns it.**

New `**/` matcher support (zero or more directories) so `content/docs/**/get-started/` covers all seven get-started trees *and* the top-level one.

#### 1b. A push that widens the path set never asks the new team

The Sentinel resolves required roles from **live** paths on every evaluation. Triage's routing step ran only at open / ready. So a widening push introduced a required approver nobody was told about.

Not "re-request on every push" (that's the noise the sticky rule prevents) -- **"ask a team that has never been asked on this PR"**:

| Situation | Before | After |
|---|---|---|
| Team already reviewed (dropped from `requested_teams`) | not re-pinged | not re-pinged |
| Team a human deliberately un-requested | not re-pinged | not re-pinged |
| Team the live paths now require, never asked | **never pinged** | asked, once |

Reads the timeline's `review_requested` events, not current `requested_teams` -- GitHub drops a team from current state the moment a member reviews, so current state can't tell "never asked" from "asked and answered". Exact-slug matching, so `docs-tool` can't satisfy `docs-tools`.

Now three steps — resolve (on `GITHUB_TOKEN`) / mint token / request — with the last two gated on there being something to ask for. The ESC OIDC exchange was otherwise paid on every push to produce a token that's unnecessary on nearly all of them.

#### 1c. Who clears the gate is a separate question from who is asked (was #21734)

G3 required an approval from a member of **every** matrix-resolved team, so a PR spanning subjects needed a quorum. #21630 is the case: docs pages + the shortcode they use + the lint rule that validates it → three teams, waiting on marketing for a one-line template change the docs reviewer had already read.

```yaml
approval: { scope: any-team, admins_satisfy: true }
```

- **`scope: any-team`** -- a member of any team in `teams:` satisfies G3. `scope: lane` is the old rule and stays the default when absent.
- **`admins_satisfy: true`** -- a repo admin's approval clears it, read from GitHub's effective permission rather than a hand-kept list. This concedes what an admin can already do at the merge box instead of reporting a block the repo doesn't impose.

Bot approvals never count, at any scope, admin or not.

**These three depend on each other.** Under `scope: any-team` a misroute no longer *blocks* -- it just means the wrong team gets pinged and the right one never hears. That makes 1a and 1b load-bearing rather than cosmetic: the reviewer *request* becomes the only mechanism putting the right eyes on a PR.

#### 1d. `review:reroute` was added, then removed

Originally a targeted escape hatch: G3 accepts any routing team on one PR. Under `approval.scope: any-team` that's the **identical predicate** -- inert, not dormant. Removed entirely. A documented, labelled escape hatch that no-ops on day one is the same defect class #21701 deleted `auto_approve` for.

---

### 🔴 A fork PR could reach the org-scoped token

`staging-deploy-auto.yml` and `staging-deploy-pr.yml` interpolated `${{ github.event.pull_request.head.ref }}` directly into `run:` blocks. `${{ }}` is textual substitution performed *before* bash parses the line, and `git check-ref-format` accepts every shell metacharacter except a space. A branch named ``a`curl …` `` executes. Both workflows hold ESC OIDC credentials. (Shipped separately as #21739 and merged in.)

### 🔴 G1 and G2 were forgeable by the PR author

`_find_comment` matched its marker **anywhere in any comment by anyone**, and returned the oldest hit. A PR author who posted an ordinary comment containing `<!-- CLAUDE_REVIEW_AUTHOR -->` before the review lane posted the real card owned G1 and G2 for the life of the PR.

Reproduced: a card with three unanswered 🚨 went from `G2 red` to `G2 ok, blocking=[]` purely on comment ordering. `update_strip` then wrote the ⛔ banner into the forgery, so the real card never showed it either.

### 🔴 An approval survived a push

G3 counted any approving review regardless of the commit it was left on. Approve a one-line typo fix, then push anything. Approvals are now filtered to `commit_id == head_sha`, and a stale one is named in the summary rather than silently dropped.

### 🟠 Staging evidence was forgeable

G4 accepted any green `staging/pulumi-test-io` status at the head, whoever wrote it. Anyone with write access could hand-write the one gate with no waiver. Now restricted to `github-actions[bot]` and `pulumi-bot`, with the writer named in the summary.

### 🟠 A missing team token deadlocked the whole repo, silently

GitHub answers **404** for an org resource a token can't see, byte-identical to "not a member". The ESC fetch is `continue-on-error`. So an expired `GH_TOKEN_TEAM_READ` made every approver read as a non-member: every PR red with *"no qualifying human approval yet"* — blaming the approver — and `review:waived` refused by the same lookup. One probe now names the cause.

### 🟠 The `F?` placeholder passed G2, and 🟡 `parse_state` took the *first* state block

A card that escaped renumbering read as "every finding answered" with a live blocking finding on screen. And the composer appends the real `REVIEW_STATE` last, while a file path in a finding row can embed a forged one.

### 🟡 Three labels and one review action weren't in the workflow's event filter

`automation/merge` (present ⇒ success, **no gates evaluated**) and `sentinel:preview` are applied by their lanes *just after* opening the PR, so the `opened` run predated them. And G5 reads the oversized ack out of the approving review's **body**, so the natural human sequence — approve, see the red gate, edit the review to add the string — fired no event at all.

---

## Part 3 — The deploy queue, and the staging paths

**`await-in-progress.js` only waited on `in_progress` runs.** A *queued* run is invisible to that filter. Run 100 queues while 101 runs → 101 sees nothing and proceeds; 100 then starts, sees 101, discards it because `101 > 100` — both enter `pulumi up` and one takes the 409 the script exists to prevent. Also removed the `branch` filter (the stack is per-workflow) and bounded the wait at 45 minutes.

**`staging_evidence.paths`: 28 paths → 7.** The gate exists for one thing — proving `pulumi up` lands. Every PR already runs `make ci_pull_request`, the same pipeline in preview mode, so a build script is exercised for real on the PR that changes it; a second nine-minute deploy of a shared, lock-contended stack re-proved nothing. What a preview cannot do is **apply**. What's left is `infrastructure/`, `run-pulumi.sh`, the `make ci_push` chain into it, `await-in-progress.js`, and the two deploy workflows.

One honesty note, now in the config: a staging deploy runs `testing-build-and-deploy.yml`, so evidence for `build-and-deploy.yml` is partial — the two are structural twins differing only in the GitHub Environment that swaps the stack. Kept anyway; partial beats none for the workflow that deploys production.

---

## Part 4 — Docs that contradicted the code

- `CONTRIBUTING.md`: "any write-access human can apply `review:waived`" — the waive path requires routing-team membership.
- `CONTRIBUTING.md`: "truly mechanical changes need no human at all" — they skip the *model review*, not the approver.
- `sentinel.py` promised four gates and listed five.
- `domain:infra` does **not** imply a staging run and `domain:frontend` does not preclude one. Three files welded the matrix to `staging_evidence.paths`.
- `automation/merge` was undocumented — the label that turns the merge gate off entirely.
- `infrastructure-deployment.md` credited only `/deploy-staging` with making G4's evidence.
- `analyze.py`'s `REASON_CODES` still advertised `gate:none`.

---

## Part 5 — The pre-merge review, and a `/simplify` pass

The review's three findings are answered ([thread](https://github.com/pulumi/docs/pull/21708#issuecomment-5735861612)). **F1 was a real bug**: a failed timeline read made every team look never-asked, re-requesting all of them on every push. **F3 was a test that could not fail** — it asserted something `load_config` already rejects on the very call it made.

The `/simplify` pass then caught **two regressions of my own**:

- **`_card_rows`.** I removed the `F?` filter from a *shared* helper to serve G2. `sla-sweep.py` counts rows absent from `REVIEW_STATE` as unanswered and `F?` is never a key in it — so N placeholders counted N times against an author with no id to answer with, feeding `review:author-stalled` and the 21-day auto-close. `collect.py` put them on the `/pr-review` board. Filter restored; G2 opts in.
- **`require_org_read`** guarded `get_team_membership` but not `get_repo_permission` three methods below it — the same 404-means-two-things conflation the probe exists to stop, reintroduced one method down. Both go through one helper now.

And a wrong explanation of my own: the F1 comment blamed `| sort -u` for masking gh's exit status. `set -euo pipefail` is on in that step, so the pipe was never the problem — `|| true` discarded the status. Fix unchanged, comment corrected.

Plus: widening the injection test to `*.yaml` (one workflow was exempt from a test whose docstring promised every workflow) and adding `inputs.pr_number` surfaced three more unhardened sites, fixed here. A second copy of `publish_guard.decide`'s head check that had already drifted on SHA case. Two dead-pattern tests collapsed into one that checks *per pattern* rather than per entry. Two identical `frozenset`s, two copies of the same path validator, comments listed three times per run, a test compiling ~25,500 regexes where 2 would do, and a dead exception class.

Every new assertion was verified to fail against the code it guards.

---

## Deliberate design, documented rather than reversed

| | Why it stays |
|---|---|
| Fork PRs skip G1/G2 | A fork can't get a review run at head; the gate would be permanently red. |
| Authors dispose of their own findings | The v3 premise — `accepted: <reason>` is a recorded decision, not a bypass. |
| `not_governed` skips G4 too | Those lanes post their own approval and arm auto-merge. |
| `link_only.approval: any-team` | Inert under `scope: any-team`, kept as the narrower rule if the scope tightens. Unlike `review:reroute`, it predates this PR and `analyze.py` reads it directly. |

### Skipped, with reasons

- **Comparing qualified `org/slug`** in the ever-asked check, which would drop the new bare-slug config constraint. It rests on the timeline payload carrying the org, which I can't verify — this environment has no `gh` CLI. Wrong guess here silently re-pings everyone, so it needs someone who can check the payload.
- **Moving the ever-asked decision from bash into `route-pr.py`.** Architecturally right — it would delete the string-pinned assertions and the config constraint — but too large for this branch.
- **A structural injection test** (parse YAML, walk `run:` scalars, allowlist safe expressions). Reports **131 hits repo-wide** today, mostly numeric ids. Can't land without a cleanup far wider than this PR.

### Still open before `REVIEW_V3_SENTINEL` goes to `1`

- **`pulumi/docs-blog-review` is unproven end to end.** Three of four routing teams have confirmed `pulumi-bot` requests since routing flipped on. A team that doesn't resolve fails as a swallowed `::warning::`.
- **A failed team request is swallowed by design.** The cover would be Sentinel-side: a red G3 that names a required team never asked. Needs its own timeline read.

### Related issues

Follows #21701. Absorbs #21734 and #21739. Fixes the routing behind #21723 and #21718; prompted in part by #21630.

https://claude.ai/code/session_01XMtbujfu7YTZAAAQZz1pFe


CamSoper:
@claude all three findings answered — F1 fixed (it was real), F2 fixed, F3 replaced. Detail below so the re-review has the context. #update-review

**F1 — fixed, and it was a real bug.** A failed timeline read made every required team look never-asked, so one transient 500 re-requested all of them, on every push. Now the read's exit status is tested rather than discarded, and a failed read requests nobody (routing is an assist; the Sentinel enforces the matrix either way).

One correction to my own first attempt at this: I wrote that `| sort -u` masked gh's exit status. That's wrong — `set -euo pipefail` is on at the top of that step, so the pipe propagates gh's failure fine. `|| true` discarded the status and `2>/dev/null` discarded the reason. The fix was right; the comment explaining it named the wrong mechanism, and is corrected.

**F2 — fixed.** The `synchronize` guard was dropped from two steps and pinned on one. It's now three steps (see below) and the test walks all of them.

**F3 — correct, and replaced rather than deleted.** That test couldn't fail: `load_config` already rejects a malformed slug on the very call the test made. Replaced with the invariant the step actually depends on and nothing was checking — bare-slug uniqueness across `teams:`, since the ever-asked check greps `${team##*/}`.

Every new assertion was verified to fail against the pre-fix code before I trusted it. Given F3 was itself a test that couldn't fail, once seemed insufficient.

---

A `/simplify` pass after that turned up two more of mine worth flagging to a reviewer:

- **`_card_rows` regression.** I'd deleted the `F?` filter from a shared helper to let G2 see placeholders. `sla-sweep.py` and `collect.py` also call it — placeholders were counting as unanswered findings against authors who have no id to answer with, feeding `review:author-stalled` and the 21-day auto-close. Filter restored; G2 opts in.
- **`require_org_read` asymmetry.** It guarded `get_team_membership` but not `get_repo_permission` three methods below — the same 404-means-two-things conflation the probe exists to prevent. Both go through one helper now.

Also worth a reviewer's eye: widening the injection test's glob to `*.yaml` and adding `inputs.pr_number` surfaced three more unhardened sites in `claude-code-review.yml` and `claude-update.yml`, fixed here. I also removed that test's claim that "a new lane cannot reintroduce it" — it's a denylist matched against lines, and the structural version reports 131 hits repo-wide, so the claim wasn't true.

Three things I deliberately did **not** do, all noted in the commit: comparing qualified `org/slug` in the ever-asked check (rests on a timeline payload shape I can't verify without `gh` in this environment), moving the ever-asked decision out of bash into `route-pr.py` (right call, too large for this branch), and the structural injection test (blocked on those 131 pre-existing hits).

---

PR #21710, opened 2026-09-18
Add template-first paths to the onboarding prompt

cnunciato:
Reworks the paste-in onboarding prompt (`static/onboard.md`). Adds a template-first path as the default: the user picks a cloud (AWS, Azure, or GCP), what to build (a minimal starter, static website, serverless app, or container service), and a language, and the agent scaffolds a known-good template from github.com/pulumi/templates — the most predictable route to a first successful `up`. Custom builds and migrations are still available on request.

Also makes the rest of the flow more reliable and more instructive: installs a durable CLI on the user's PATH across macOS, Linux, and Windows, with per-shell setup for zsh, bash, fish, and PowerShell rather than leaving `pulumi` unresolvable in new shells; names Pulumi's supported languages and offers a cloud-appropriate default set; explains projects, stacks, configuration, preview, and update inline as they happen, each with a docs link; verifies the deployment from a stack output; surfaces the account-claim step before inviting the user to run any `pulumi` commands themselves (their shell has to claim and `pulumi login` first); and closes by suggesting next steps and linking the docs, blog, and Dev Center.

Closes [MA-790](https://linear.app/pulumi/issue/MA-790/update-onboardingmd).


jeffmerrick (approved):

static/onboard.md
```diff
@@ -1,19 +1,58 @@
...
+
+## 1. Install the CLI and put it on the PATH
+
+The user should be able to run `pulumi` now and in new shells, so onboarding
+must leave a real CLI on their PATH — not just a per-session `npx` shim. First
+check whether it is already installed: `pulumi version` (this does not touch
+Pulumi Cloud). If that works, continue.
+
+If `pulumi` is not found, install it for the user's platform. Neither installer
+modifies the PATH, so a fresh shell won't find `pulumi` yet — set that up next.
```
kramhuber: Did you run into issues with `path`? Both installers appear to handle persistent PATH setup already. Should we let them handle that and keep the manual setup here as a fallback? That would save the user a separate PATH decision. 

We would still need to make pulumi available in the agent’s current shell, since the Unix installer cannot update its parent shell.

static/onboard.md
```diff
@@ -49,48 +114,117 @@ If the migration skill install does not succeed, continue with
...
-have from the block, and tell them to claim after you finish, since claiming
-locks the organization during the claim process. If you are working on behalf
-of the user rather than beside them, relay the link in your response.
-
-## 7. Point to the next step
-
-Ask the user what they want to do next. Useful directions are more resources in
-the project, credentials and secrets in ESC (`pulumi-overview` Level 3), or the
-provider catalog at pulumi.com/registry.
+Surface the claim URL right after the first success, framed as saving their
```
kramhuber: Should we show the claim instructions only when the CLI prints a claim link? Otherwise, the agent could continue with the user’s existing account or help them log in. That condition would also apply to the claim references in steps 5 and 7

static/onboard.md
```diff
@@ -1,19 +1,58 @@
...
+  full path.
+- Permanently: offer to add that directory to the PATH for future shells —
+  never edit anything automatically, only with the user's go-ahead. Detect the
+  shell and use its own target and syntax rather than assuming zsh (`$SHELL`
+  names the login shell on macOS/Linux; on Windows you're in PowerShell):
+  - zsh: add `export PATH="$HOME/.pulumi/bin:$PATH"` to `~/.zshrc`.
+  - bash: the same line in `~/.bashrc` (Linux) or `~/.bash_profile` (macOS).
+  - fish: run `fish_add_path $HOME/.pulumi/bin`.
+  - PowerShell: add `$env:Path += ";$HOME\.pulumi\bin"` to the file at
+    `$PROFILE`, or persist it with `[Environment]::SetEnvironmentVariable('Path',
```
kramhuber: `$env:Path` contains every command directory available to the running terminal: system directories, the user’s saved directories, and any temporary directories added for the agent session. Would saving that whole list into the user’s permanent PATH preserve temporary agent directories and duplicate system entries?

The Windows installer already adds Pulumi’s directory to the user’s saved PATH without copying that combined list. Could we remove this extra command and rely on the installer?

static/onboard.md
cnunciato: I did; the current setup did not work for me. Opening a new terminal session would leave me without a `pulumi` not on the path. (I reported this in my initial feedback as well.)

static/onboard.md
cnunciato: Good catch on both counts — fixed in ccfc3971a6. Dropped the `SetEnvironmentVariable('Path', "$env:Path;…", 'User')` line: persisting the process `$env:Path` would have baked temporary agent-session dirs into the user PATH and duplicated system entries, exactly as you describe. And you're right that the Windows install script already adds Pulumi's dir to the saved user PATH, so I removed the extra command (and the incorrect "neither installer modifies the PATH" line) and now rely on the installer for Windows — the per-shell PATH setup is macOS/Linux only.

static/onboard.md
cnunciato: Agreed — fixed in ccfc3971a6. The claim flow now only applies when the CLI actually prints a claim block (i.e. the user had no saved credentials). Step 6 is gated on that; if the user is already on their own account there's nothing to claim, so we skip it and just ensure they can `pulumi login`. Threaded the same condition through the claim references in steps 5 and 7.

cnunciato:
Got @kramhuber's 👍🏻 to merge! @jeffmerrick We can wire up the UI on this tomorrow, starting with `/docs`.

Issue #21702, opened 2026-09-17
Document the Pulumi Deployments status lifecycle


jkodroff:
Follow-up to pulumi/pulumi-service#19462, which asked for "an extensive list of deployment statuses and an explanation of what they mean."

The **API reference** half is handled in pulumi-service: the per-value descriptions on `JobStatus` and `StepStatus` in the Java IDL have been rewritten as customer-facing prose, and they flow to https://www.pulumi.com/docs/reference/cloud-rest-api/deployments/ automatically once that ships to prod and the docs site rebuilds.

What's still missing is the **narrative** — the parts that don't fit in a per-field description.

## What to write

A conceptual page, suggested home `content/docs/deployments/operations/`, alongside the existing `deployment-queue.md` (which already documents the 7-day queue expiry → `skipped` rule). It should cover:

### 1. The deployment → job → step hierarchy and state machine

A deployment contains jobs; jobs contain steps. A deployment reports its job's status. Steps use a narrower set of values (no `accepted`, no `skipped`).

The authoritative transition diagram already exists in the service source and can be transcribed — `cmd/service/services/workflows/controller.go`, doc comment on `doStepStateTransition`:

```
Jobs:
             +-------------+--------> failed
             |             |            ^
             |             |            |
not started -+-> accepted -+-> running -+-> succeeded
                     ^     |      ^     |
                     |     |      |     |
                     +-----+      +-----+
```

Note `skipped` does **not** appear there — it's applied by the queue, not the workflow engine.

### 2. A status vocabulary mapping table

This is the highest-value piece. The same deployment state has three different names depending on where you look, and nothing documents it. A webhook consumer today receives `queued` — a value that appears in no published enum.

| REST API | Raw webhook payload | Console label |
|---|---|---|
| `not-started` | `queued` | Queued (or Paused, when deployments are paused) |
| `accepted` | `queued` | Running |
| `running` | `running` | Running |
| `succeeded` | `succeeded` | Succeeded |
| `failed` | `failed` | Failed |
| `skipped` | `skipped` | Skipped |

Sources: `cmd/service/services/queue/webhooks/output.go` (`getDeploymentOutputStatus`) and `cmd/console2/src/app/util/deployments.ts` (`transformStringDeploymentStatus`).

Worth noting: the drift-detection webhook does **not** apply this collapse — it reports the raw `JobStatus`.

### 3. There is no `canceled` status

[...]

PR #21723, opened 2026-09-18
Render API response bodies that aren't objects

jkodroff:
### The bug

`layouts/partials/openapi/response-schema.html` emits markup only under `{{ if $resolved.properties }}`. Any 200 response whose body isn't an object has no `properties`, so the whole block is skipped and the page renders an empty Responses section — the schema is right there in the spec, and nothing reaches the reader.

**39 operations across 13 tags** are affected: Environments 12, Organizations 6, Deployments 4, Webhooks 4, VCS Integrations 3, Users 2, and one each in AI, AuditLogs, CloudSetup, InsightsAccounts, Miscellaneous, Stacks, StackUpdates.

The worst case is [`GET /api/deployments/{deploymentId}/status`](https://www.pulumi.com/docs/reference/cloud-rest-api/deployments/#get-apideploymentsdeploymentidstatus), which returns a bare `JobStatus`. All six values *and their descriptions* ship in the spec and are simply never displayed:

> **Responses**
> 200 &nbsp; OK

That endpoint is what pulumi/pulumi-service#19462 points at when it reports that deployment statuses are undocumented. They weren't missing — they were unrenderable.

### The fix

Three branches added to `response-schema.html`:

- **Bare enum** → the existing `enum-values.html` partial, the same one the object path already uses for an enum-valued property. No new rendering logic.
- **Array / map / scalar** → a type label, using the same notation as the property type column in `property-rows.html` so the two read alike (`array[Foo]`, `map[string]Bar`, links to the schema pages).
- **Array of `$ref`** → also expands the item schema's fields. Those are what the reader came for; a bare `array[Foo]` label would just move the question one click away. 18 of the 39 get a full field table this way.

`resolve-schema.html` had to start carrying `enum`, `items` and the resolved leaf schema — it previously returned only `properties`/`required`/`description`/`type`/`schemaName` and dropped the rest. Purely additive, so existing callers are unaffected. The leaf schema (`raw`) is needed because `enum-values.html` reads the sibling `x-pulumi-model-property` extension (`enumFieldNames` / `enumFieldComments`) off the schema object it's handed, not off the bare enum array.

### Verification

Built locally against the **live prod spec** (`scripts/fetch-openapi-spec.sh`), so this is the template change on its own:

| | |
|---|---|
| `#get-apideploymentsdeploymentidstatus` | was empty → now all six values with descriptions |
| `#responses-3` (GetDeployment, an object body) | unchanged — regression check on the `properties` path |
| Environments page | 12 previously-blank responses now labelled; `array[WebhookResponse]` expands its fields |

No new Hugo warnings or errors. (The two `openapi: missing intro` warnings for `ai-agents` and `invites` are pre-existing on master.)

### Related

- pulumi/pulumi-service#49677 rewrites the `JobStatus` / `StepStatus` descriptions from Go-internals prose into real documentation. Independent of this PR — that one fixes *what* the text says, this one fixes *whether it appears at all* on the bare-enum endpoint. They compose.
- pulumi/docs#21702 tracks the narrative deployment-status docs.


jkodroff:
/resolve F1 fixed: ac2935909c scopes `color: inherit` to a badge inside `.api-schema-label`, with the dark counterpart because the dark `.api-schema-label a:hover` rule outweighs the light override on its own. Verified with headless Chrome against a local server — on environments, organizations and deployments the response badge link and the property-row link now compute to the same color in both themes (`rgb(30,87,235)` light, `rgb(163,208,255)` dark), and hover no longer flips.
/resolve F2 deferred: filed #21740. Both copies set a variable as a side effect — `$refSchemaName` (property-rows L43/51, read at L90) and `$itemSchemaName` (response-schema L38, read at L74) — to drive recursive `$ref` expansion. A Hugo partial can emit markup or `return` a value, not both, so extraction means returning `{html, schemaName}` and building the notation as a string, and it puts `property-rows.html` — every property on every page of the API reference — in the blast radius of a bug fix. Worth doing on its own.

jkodroff:
@claude #update-review

CamSoper:
Routing to @pulumi/docs-marketing-review: this is a frontend change, not mine to approve.

What the queue flagged:
- `warnings:1:F1`

---

CamSoper (approved):
LGTM.

PR #21729, opened 2026-09-18
Add a WorkPrentice-scoped section to AGENTS.md, plus WORKPRENTICE.md

CamSoper:
### Proposed changes

Per the thread with @jkodroff: WorkPrentice keeps shipping comparison pages between two products that aren't ours (most recently `content/docs/iac/comparisons/cdk-vs-terraform.md`). Josh asked if we could just put the rule in AGENTS.md.

**`AGENTS.md`** gets a short `## WorkPrentice` section that names the bot and tells every other agent to skip it -- AGENTS.md is read by everything, and we don't need Claude Code following instructions written for someone else's bot. It points at the new file and nothing else.

**`WORKPRENTICE.md`** (new, root, 19 lines) has three rules:

- **Comparisons compare Pulumi to something.** Every page under `content/docs/iac/comparisons/` already does `h1: Pulumi vs. X`, so the rule is "match the neighbors." The legacy `content/what-is/` comparison pages are called out as not-precedent so the rule isn't DOA against content we already shipped.
- **A new content genre gets an issue, not a PR.**
- **A brown-M&Ms canary:** PR descriptions get signed `-- Dr. Sbaitso, DDS`. Unmissable, zero blast radius, tells us at a glance whether the file was read. (An obsolete 1991 speech-synth AI co-signing the AI-generated PRs felt right. The dentistry degree is not canon.)

I left `cdk-vs-terraform.md` alone -- it does cover Pulumi, it's just framed with us as the third wheel, so the fix is a retitle plus an alias and that's a separate decision from the rule.

**Testing:** `make lint` doesn't run in this container (no Hugo, no `node_modules`), but both files are outside its reach anyway: `scripts/lint/lint-markdown.js` walks `content/` only, and `*.md` is in `.prettierignore`.

n/a

### Related issues (optional)

n/a -- came out of a Slack thread. Context: the `cdk-vs-terraform.md` commit history and https://github.com/pulumi/docs/pull/20756.

https://claude.ai/code/session_01T2YJ3vWtbxPTL2YPkYMx23


CamSoper:
@workprentice Would you mind reviewing this PR, please?

jkodroff (approved):

PR #21734, opened 2026-09-18
Loosen Sentinel G3 to any routing team or a repo admin

CamSoper:
### Proposed changes

G3 required an approval from a member of **every** matrix-resolved team, so a PR that spans subjects needed a quorum. [#21630](https://github.com/pulumi/docs/pull/21630) is the case that prompted this: it touches docs pages, the Hugo shortcode those pages use, and the lint rule that validates the shortcode, so the matrix resolved docs-guild + marketing + tools, and the PR sits waiting on marketing for a one-line template change the docs reviewer has already read.

This splits the two questions the matrix was answering at once. The matrix still decides who gets **asked** -- triage requests that team, the SLA sweep chases it, the reviewer brief names it. A new `approval:` section decides who can **clear** the gate:

```yaml
approval: { scope: any-team, admins_satisfy: true }
```

- **`scope: any-team`** -- a member of any team in `teams:` satisfies G3, whatever the matrix routed. `scope: lane` is the old rule and stays the default when the section is absent.
- **`admins_satisfy: true`** -- a repository administrator's approval clears the gate on its own, read from GitHub's effective permission for the approver (`repos/{repo}/collaborators/{user}/permission`) rather than a hand-kept list that would rot. This concedes what an admin can already do at the merge box instead of reporting a block the repo doesn't actually impose.

Side effect worth naming: the reviewer brief already tells reviewers "any member's approval satisfies the merge gate," which was wrong under the lane rule on any multi-subject PR. It's now true.

What doesn't change:

- Bot approvals never count, at any scope, admin or not.
- Every governed PR still resolves to at least one role and still needs a human approval.
- `link_only.approval: any-team` stays as the narrower rule -- subsumed while the scope is `any-team`, live again if it goes back to `lane`.
- Routing output (`route-pr.py` roles/teams) is untouched, so reviewer requests and SLA escalation behave exactly as before.

Implementation notes:

- The admin lookup runs only when the team rule already came up short, so an ordinary PR costs no extra API calls; it rides the same org-scoped token as the team-membership endpoint (both need more reach than the default `GITHUB_TOKEN`), and a lookup failure errors the gate to `action_required` rather than reporting a block it couldn't verify.
- `routing.py` validates the new keys the same way it validates everything else -- unknown key, bad `scope`, or non-boolean `admins_satisfy` is a hard error.
- Tests: `make test-review-pipeline` passes (490 pytest cases + every `--self-test`), including new coverage for the any-team scope, the admin rule on and off, a bot admin, and a failed permission lookup. `make lint` couldn't run in this container (no Hugo/node deps), but nothing here touches content markdown.

### Related issues (optional)

Prompted by [#21630](https://github.com/pulumi/docs/pull/21630).

https://claude.ai/code/session_01QJeNNAXvc8WzXsdBTMoHmq

---
_Generated by [Claude Code](https://claude.ai/code/session_01QJeNNAXvc8WzXsdBTMoHmq)_


CamSoper:
Merged into #21708 and closing here — commit `77442fd7` is preserved intact, so this landed as a merge rather than a re-type.

Both branches rewrote the same G3 block, the same config sections, and the same tests, so these were always going to conflict. Better to resolve it once.

**They compose well.** `overrides:` (who gets *asked*) and `approval.scope` (who can *clear* the gate) are the same separation-of-questions argument applied at two different points:

```
matrix     -> the subject a path belongs to
overrides  -> who owns it, when ownership doesn't follow the subject
approval   -> who may clear G3, now a separate question again
```

They also need each other. Under `scope: any-team` the reviewer **request** becomes the only thing putting the right eyes on a PR — a misroute no longer blocks, it just means the wrong team gets pinged and the right one never hears. That makes #21708's two routing-accuracy fixes (`overrides:`, and asking a newly-required team when a push widens the path set) load-bearing rather than cosmetic.

**One thing didn't survive the merge: `review:reroute`.** #21708 had added it as a targeted escape hatch — G3 accepts any routing team on one PR, so a misroute wouldn't need `review:waived`. Under `approval.scope: any-team`, G3 accepts any routing team on *every* PR, so the label isn't "narrower and dormant" — it's the identical predicate, doing nothing. Removed, along with its gate branch, summary line, workflow label-filter entry, label row, README paragraph, and tests.

Reasoning: shipping a documented, labelled escape hatch that no-ops on day one is the same defect class this branch has been clearing. #21701 deleted `auto_approve` for reporting a pass GitHub disagreed with; a lever that silently does nothing is that same lie pointed the other way. It's one commit to restore alongside anything that sets `approval.scope` back to `lane`.

`link_only.approval: any-team` is deliberately **not** removed on the same logic — it already exists, costs nothing dormant, and your header already documents it as the narrower rule that goes live again if the scope changes. Adding a new inert lever and keeping an existing one are different trades.

Conflicts were two, both unions rather than choices: `TOP_LEVEL_KEYS` took both `overrides` and `approval`; G3 took this PR's `all_teams` / `resolution.any_team` shape with the reroute local dropped.

`make test-review-pipeline` and `make lint` pass on the merged tree.

---

Issue #21733, opened 2026-09-18
Claim verifier: 27% of verified verdicts cite only the page under review


CamSoper:
## What I found

In the content-review claims index (`claims/` prefix of the ledger bucket, 176 pages as of 2026-09-18), **1,075 of 3,986 `verified` verdicts (27%) cite only the page's own file as their source** -- 830 of them at `high` confidence.

By claim type:

| type | self-sourced | of all verified | share |
|---|---|---|---|
| behavior | 429 | 1051 | 41% |
| feature | 254 | 544 | 47% |
| api-surface | 258 | 1087 | 24% |
| cross-reference | 15 | 52 | 29% |
| version | 15 | 61 | 25% |
| url | 91 | 934 | 10% |
| numerical | 9 | 90 | 10% |
| entity-spec | 3 | 65 | 5% |

Method: a `verified` verdict counts as self-sourced when every part of its `source` field (split on `;` / ` and `) contains the page's own `path`. Typical evidence text: "The doc itself states at L82: ...".

This is exactly the failure the verifier prompt's source-discipline section names ("The content under review never verifies its own technical claims ... citing the reviewed file as the `source` of a `verified` is circular"). The rule is prose-only, and like the generated-from-data rule before it got a gate, it looks advisory in practice.

## Why it matters

- The "N claims verified clean" figure on reviewer briefs and in ledger health is inflated by roughly a quarter. A reviewer rubber-stamping "Facts: verified" is trusting a check that didn't happen.
- These claims also sit in the nightly reverify pool looking healthy.
- Whole-file content-review runs are the worst case, since the page is both the claim source and the nearest thing to read.

## Caveats

I haven't hand-audited a sample. Some fraction are self-descriptive claims ("the example uses X") where the right verdict is `not-a-claim`, not a missed verification. Either way `verified` is the wrong label, but the fix differs.

## Proposed next steps

1. Sample ~50 of the 1,075 (stratified by type) and classify: checkable against product source / should be `not-a-claim` / legitimately self-evident.
1. Based on the split, add the mirror-image gate in `_finalize_verdict`: a `verified` whose only source is the reviewed file becomes `not-a-claim` or gets one re-hop against an independent source, the same machinery as the self-reference gate for `contradicted`.
1. Measure before/after on the fork corpus -- this moves brief confidence numbers and the reverify pool, so it needs its own benchmark rather than riding along with another change.

Related: the `contradicted` side of the same rule surfaced on #21720.

[...]

Issue #21732, opened 2026-09-18
Pre-merge review: give the claim verifier the PR body's rationale for changed values


CamSoper:
## Problem

When a content-review bot PR (glow-up or fix lane) changes a factual value, the PR body already says why and where the new value came from. The pre-merge review's claim verifier never sees it, so it re-derives from scratch and can land on the opposite answer.

Concrete case: #21720. The glow-up bumped an example `Pulumi.yaml` pin from `version: 0.10.0` to `1.4.0`. Its "Backlog executed" table cited the receipt (`gh release list -R pulumi/pulumi-terraform-provider` shows v1.4.0, 2026-08-28), and the same run's verdict (c13, `framing-drift`) was written to the claims index in S3 two minutes before the review started. The review's verifier got thin one-line claims, routed them to web search, found the live copy of the same page still showing `0.10.0`, and stamped all three edits `contradicted`. A human had to dispute it before the update lane ran the `gh` check and conceded.

## Proposal

Thread the PR body's rationale for changed lines into the verifier prompt, the same way `fetch_impl_refs` in `verify-claims.py` already threads "this docs PR cites implementing change(s)" into pass1/pass3.

- Scope it to bot-authored content-review PRs to start (`content-review/*` labels), where the body is composed deterministically and has a stable table shape.
- Treat it as a **lead, not evidence**: the verifier gets the pointer ("author says: checked `gh release list -R pulumi/pulumi-terraform-provider`") and runs the check itself. Author assertion doesn't establish a verifiable claim -- that's already the rule in the update lane's dispute handling and should stay the rule here.
- Optional second source: the page's entry in the claims index (`claims/<slug>.json`). A fresh verdict on the same entity from the run that produced the PR is a strong hint about which source to read.

## Open questions

- Matching body rows to claims: by line range, or by value (`0.10.0` -> `1.4.0`)? The executed table cites line numbers, which should be enough.
- Prompt-injection surface: the body of a bot PR is trusted-ish, a human PR body is not. Another reason to start with bot PRs only.
- Worth measuring on the fork before shipping: replay #21720, #21394, #21552, #21602 and confirm the false contradictions clear while the two real regressions (#21552, #21602) still get caught.

## Update, 2026-09-18: parked pending data

[...]

PR #21736, opened 2026-09-18
Persist the review's claim records and verdict metadata

CamSoper:
### Proposed changes

Tracing the false `contradicted` verdicts on #21720 hit a wall: nothing keeps the pre-model claim records. `.candidate-claims.json` and `.verified-claims.json` die with the runner, and the evidence trail only keeps the rendered verdict, so there was no way to tell afterward what claim `type` or `source_hint` sent a version pin to web search.

- `claude-code-review.yml` uploads `.candidate-claims.json`, `.verified-claims.json`, and `.fetched-urls.json` as `review-claims-pr<N>-run<id>` (30 days), right after verification so a later timeout or model failure can't lose them. `include-hidden-files: true`, since they're dotfiles.
- Trail records in the evidence object now carry `claim_id`, `type`, `confidence`, `framing`, `framing_note`, `turn_cap_exhausted`, and `source_discipline_gate` when the verdict has them. All optional and additive -- records already in S3 still validate, and the update lane carries the trail through verbatim.
- `test_workflow_sanity.py` now fails any `upload-artifact` step that lists a dotfile without `include-hidden-files: true`.

The uploaded files aren't redacted, but everything in them derives from the public diff, unauthenticated page fetches (capped at 8 KB x 30 URLs), and repo-scoped `gh` reads. The API key only ever travels as a request header.

The accepting side (schema, validator, evidence page) landed first in #21741, because the publish job validates evidence with the default branch's validator.

`make test-review-pipeline` passes (1075 tests).

### Related issues (optional)

Follow-up from the #21720 trace. Related: #21732, #21733.


CamSoper:
@claude I pushed fixes for F1 and F2 in 2669f4b064. F1: the duplication stays (the composer and the validator live in different trees and the publish job runs the validator from the default branch), but a new test pins the two vocabularies together so they can't drift silently. F2: agreed, dropped to 30 days to match the other debug bundles. #update-review

PR #21737, opened 2026-09-18
Stop the claim verifier contradicting a PR against its own live page

CamSoper:
### Proposed changes

On #21720 the pre-merge review stamped three correct edits `contradicted` (a glow-up refreshing an example `terraform-provider` pin from `0.10.0` to `1.4.0`, which is the latest release). Two of the three verdicts cited the live published copy of the page under review as their source. That page still shows the old value, so the check is a tautology: any PR that changes a number "contradicts" it. A human had to dispute the findings before the update lane ran `gh release list` and conceded.

It isn't a one-off. Across the 117 v3 review records in the ledger, 66 verdicts are `contradicted`; 9 cite the reviewed page's own live URL and 9 more rest on other pulumi.com pages alone. The verifier prompt already forbids both, but only the generated-from-data rule was enforced in code.

**Source-discipline gates (`verify-claims.py`)**

- A `contradicted`/`mismatch` whose only source is the reviewed page's own live URL, a `contradicted` whose only sources are editorial pulumi.com pages, and a `contradicted` citing only the reviewed file's repo path each get **one re-check** in the pass1 (`gh`) lane, told exactly what the first pass leaned on and why it isn't evidence.
- An independently sourced answer stands, whichever way it goes. Otherwise the verdict is downgraded to `unverifiable`, stamped `source_discipline_gate`, keeps the original reasoning, and asks the composer for an author question instead of a 🚨.
- A re-check rather than a blunt downgrade, because "the PR changed what the live page says" is a useful change detector. It just isn't a judge.
- Exempt: generated reference pages (the prompt's existing carve-out), and a claim that is *about* the cited URL (the pass2 dead-link case). Inert in `--dry-run` and without a repo root.

**Routing**

- A `version` claim with a non-URL `source_hint` now routes pass1 instead of pass3. `claim-extraction.md` tells the extractor to put the package name in `source_hint`, and the router treated any bare name as "named external source, go web search". So a pin for a Pulumi package without "pulumi" in its name (`terraform-provider`) got web-searched, where the top hit is pulumi.com's own page.

**Prose kept in sync**

- Verifier prompt and `fact-check.md` now say which rules are harness-enforced. `fact-check.md` was missing the self-reference rule entirely, and its worked-example count was stale.
- `claim-extraction.md`: worked example 14 for a `Pulumi.yaml` `packages:` block (the `version` binds to the package, `parameters` names the wrapped provider), and the `version` row prefers the `pulumi/<repo>` hint form.

### Verification

`make test-review-pipeline` passes, including 61 new tests built on the real ledger verdicts from #21720, #21552, #21602, #21394, and #21509.

Live replay of the #21720 claims (real model calls, same inputs):

| | master | this branch |
|---|---|---|
| `version` claims with hint `terraform-provider` | all pass3 (web search) | all pass1 (`gh`) |
| `contradicted` citing the page's own live URL | 3 of 5 | 0 of 5 |
| re-check observed live | n/a | yes: self-reference caught, re-verified against `gh release list`, `verified` |

Fork CI smoke (CamSoper/pulumi.docs#246, the exact #21720 diff, one run): all three `1.4.0` claims routed pass1 and came back `verified` against `gh release list -R pulumi/pulumi-terraform-provider`; zero `contradicted`; validator clean. The routing fix kept the claims off web search, so the re-check hop itself didn't fire in CI -- its live evidence is the local replay above.

Known residual: when the *extracted claim text* misdescribes the page ("pins the `random` provider to 1.4.0"), the verifier can still return `mismatch` against the page itself. That's an extraction error, and worked example 14 is the fix for it, but it's prompt-level and the replay can't prove it.

Also worth knowing: the `pulumi/<repo>` hint form yields entity keys like `version/pulumi-pulumi-gcp` (vs `version/pulumi-gcp` for the bare form). That quirk predates this PR -- worked example 4 already recommends the `pulumi/<repo>` form -- but more claims will take that shape now, so claims-index continuity shifts for them. Not touching `entity_key.py` here.

`reverify-claims.py` calls `process_claim`, so the nightly lane gets the re-check too. Its own own-corpus demotion still runs afterward.

### Related issues (optional)

Follow-ups filed: #21732 (give the verifier the PR body's rationale), #21733 (27% of `verified` verdicts cite only the page under review). Companion: #21736 (persist the claim records this trace couldn't find).


CamSoper:
@claude F1: accepting as-is -- it's the right trade. Version claims that carry a bare non-URL hint are rare (version is ~60 of ~4,000 claims in the index, and most carry no hint at all), pass1 escalates to web search when gh can't close it, and the failure this replaces was a false blocker that needed a human dispute to clear. An extra hop on the occasional Node.js pin is cheap next to that. If PR #21736's claim artifacts later show it costing real money, a narrower rule is easy to add with data in hand. #update-review

PR #21739, opened 2026-09-18
Injection guard: a one-line `run:` is a mapping entry too

CamSoper:
### Proposed changes

**Rebased onto `master`, and now a different PR than it was.** The original three workflow fixes shipped in #21708, which squash-merged three minutes before this branch's last fix landed. Everything this PR used to carry is on `master` already — except one thing, which is the gap this PR's own review found.

So the branch was rebuilt on current `master` and carries a single-file diff.

### The remaining gap

`test_no_workflow_interpolates_pr_controlled_text_into_a_shell` exempts lines that look like a YAML mapping entry, because `env:` bindings are the safe form it steers people toward and `concurrency: group: name-${{ … }}` is a run-name that never reaches a shell.

A one-line `run:` is a mapping entry too. So this passed the guard:

```yaml
run: echo ${{ github.event.pull_request.head.ref }}
```

That is exactly the defect class the guard exists to catch.

**How it got there, since the sequence matters.** The exemption originally read `^KEY:\s*\$\{\{` — the interpolation had to follow the colon immediately. That caught a single-line `run:` *by accident*, and false-positived on `concurrency: group:`. Widening it to "any mapping entry" fixed the false positive and opened this hole. Keying it on the **name** closes both:

```python
EXECUTABLE_KEYS = {"run", "script", "args", "entrypoint", "cmd"}
key = re.match(r"^([A-Za-z_][A-Za-z0-9_-]*):(\s|$)", stripped)
if key and key.group(1) not in EXECUTABLE_KEYS:
    continue
```

`script` is in there for the same reason `run` is — `actions/github-script` executes its value as JavaScript. `env:` stays exempt because it's the target pattern; `if:` because GitHub evaluates it, never bash.

### Already on master, not in this diff

The review's other finding (F1 — the guard globbed `*.yml` only, so `scheduled-upstream-sync.yaml` was exempt from a test whose docstring promised every workflow) was found independently by a `/simplify` pass and merged with #21708. Widening that glob and adding `inputs.pr_number` to the expression list also surfaced three more unhardened sites in `claude-code-review.yml` and `claude-update.yml`, fixed there.

### What this guard is and isn't

The docstring used to claim it "covers every workflow … so a new lane cannot reintroduce it." That claim is gone, because it wasn't true. This is a denylist of expressions matched against lines: it can't distinguish a `run:` block from a `concurrency:` block except by heuristic, and the expression list is extended by hand.

The structural version — parse the YAML, walk every `run:` scalar, allow only a short list of non-author-controlled expressions — reports **131 hits repo-wide** today, mostly numeric ids and booleans. It can't land without a cleanup far wider than this PR, so the docstring now says that instead of overclaiming.

### Testing

`make test-review-pipeline` and `make lint` pass on the rebased branch. Verified against the exact case: a single-line `run:` carrying an author-controlled expression fails the guard before this change and is caught after, in a `.yml` file and in the `.yaml` one.

### Related

Follows #21708 (merged), which carries the original three workflow fixes and F1.

https://claude.ai/code/session_01XMtbujfu7YTZAAAQZz1pFe


CamSoper:
@claude both findings are correct. F1 and F2 are fixed — but on #21708, not on this branch. Reasoning below. #update-review

**F1 — correct, already fixed.** A `/simplify` pass on #21708 found this independently and widened the glob to `[*wf_dir.glob("*.yml"), *wf_dir.glob("*.yaml")]`. Widening it also turned up three more unhardened sites (`claude-code-review.yml`, `claude-update.yml` ×2) once `inputs.pr_number` went on the list, all fixed there.

**F2 — correct, and the `/simplify` pass had made it worse.** Worth spelling out because the sequence matters:

The exemption originally read `^KEY:\s*\$\{\{` — the interpolation had to follow the colon immediately. That caught a single-line `run: echo ${{ ... }}` *by accident*, and false-positived on `concurrency: group: name-${{ ... }}`, which is a run-name and never executed. I widened it to "any mapping entry" to kill the false positive, which fixed that and opened exactly the hole you name.

Fixed by keying the exemption on the **name** instead of the shape, with a carve-out from the carve-out:

```python
EXECUTABLE_KEYS = {"run", "script", "args", "entrypoint", "cmd"}
key = re.match(r"^([A-Za-z_][A-Za-z0-9_-]*):(\s|$)", stripped)
if key and key.group(1) not in EXECUTABLE_KEYS:
    continue
```

`script` is in there for the same reason `run` is — `actions/github-script` executes its value as JavaScript. `env:` stays exempt because it's the safe form this test steers people toward, and `if:` because GitHub evaluates it, never bash.

Verified against the exact case each finding describes: a single-line `run:` carrying an author-controlled expression is now caught, in a `.yml` file and in `scheduled-upstream-sync.yaml`.

**One thing you should know about the docstring.** It claimed "covers every workflow … so a new lane cannot reintroduce it." That claim is gone. This is a denylist of expressions matched against lines — it cannot tell a `run:` block from a `concurrency:` block except by heuristic, and the list has to be extended by hand. The structural version (parse the YAML, walk every `run:` scalar, allowlist non-author-controlled expressions) reports **131 hits repo-wide** today, mostly numeric ids and booleans, so it can't land without a cleanup far wider than this PR. The docstring now says that instead.

---

**Branch note for a human:** this PR's commit `34eb89f5` is already merged into #21708, and both fixes above are there. This branch has no unique content left. It should either be closed in favour of #21708, or someone should say the word and I'll put the two fixes on `camsoper/fix-staging-deploy-injection` so it can still merge on its own — I've been asked to develop on the other branch, so I haven't pushed here.

---

CamSoper:
@claude rebased onto master. F1 is fixed on master already; F2 is fixed here and is now this PR's entire diff. Title and body rewritten to match. #update-review

Worth flagging because it isn't visible from the diff alone: #21708 squash-merged at 21:49 UTC and the F2 fix landed at 21:53, so **master currently carries this guard with the F2 hole open**. Everything else this branch used to hold — the three workflow `env:` rewrites and the F1 glob fix — went in with that merge.

Rebasing normally would have been wrong here. The squash means git can't see that master already contains this branch's content, and the branch's `test_sentinel.py` predates #21708, so a merge or replay would have reverted a chunk of merged work. The branch was rebuilt on current master instead and carries one commit.

---

PR #21741, opened 2026-09-18
Accept verdict metadata on evidence trail records

CamSoper:
### Proposed changes

Prerequisite for #21736. The evidence schema and `validate-evidence.py` reject unknown keys on trail records, and the review's publish job validates with the default branch's copy, not the PR head's. So a PR that makes the composer emit new trail fields fails its own review at "Record evidence" -- which is exactly what happened on #21736 (`evidence.trail[0] has unexpected property 'claim_id'`).

This lands the accepting side first:

- `evidence-schema.json` and `validate-evidence.py` accept seven optional trail keys: `claim_id`, `type`, `confidence`, `framing`, `framing_note`, `turn_cap_exhausted`, `source_discipline_gate`. Unknown keys are still rejected; legacy records validate unchanged.
- `render-evidence-html.py` shows them on the existing muted route line when present. No new CSS; rows without them render exactly as before.

Nothing emits these fields yet, so this is a no-op until #21736 merges. `make test-review-pipeline` passes.

### Related issues (optional)

Merge before #21736.


CamSoper:
@claude I pushed a fix for F1 in 88db3b5 -- fair catch. The page now renders the framing shape with or without a note (`framing: shifted — <note>`, or just `framing: shifted`), and `none` renders nothing. Self-tests cover all three. #update-review

PR #21743, opened 2026-09-18
docs: Replace azure-native version support section with API versions

jkodroff:
The Azure provider guide had a versioning section describing support status for azure-native v1/v2/v3. That framing doesn't reflect reality, so it's removed rather than corrected. In its place: what readers actually need from that spot — how Azure Native handles Azure's multiple date-based API versions.

## Changes

- Removed the "azure-native versioning" section (major version table + support claims)
- Added "Azure API versions": default per-resource API versions, generating a local SDK for a specific version, and the generic resource type
- Linked to the registry's [Azure Native version guide](/registry/packages/azure-native/version-guide/)


CamSoper (approved):

CamSoper:
/resolve F1 accepted: Dropping it is intended: the PR description says the support claims for v1 and v2 didn't reflect reality, which is why the table is removed rather than corrected. Version questions now go to the registry's Azure Native version guide, which the new section links and which returns 200.

---

PR #21748, opened 2026-09-19
pr-review: drop the Do next strip; chain and consolidate become row buttons

CamSoper:
### Proposed changes

The `/pr-review` board opened with a "Do next" strip: one card per batch (approve the set, send back, route, each unblock) naming PRs as plain text, plus a chain card and a consolidate card that had no row button at all. Every card was just the row buttons it named, pressed together, and it sat at the top of the page where you couldn't see the rows it was talking about -- so judging a batch meant scrolling down to find each row and back up to the card. The stampable rows already arrive selected (the default command merges them either way), and every other batch is the row buttons it would have pressed. The strip bought nothing a row doesn't.

This removes it and puts the two moves that had no row button onto rows:

- **`analyze.py`**: `attach_cluster_actions()` puts the chain on its lead row as a decision (`--chain C1`, only where `only_collisions_hold()`, since `--chain` approves the lead with `--force`; it `covers` the next link) and the consolidation on the newest sweep. Recommendations and cluster buttons are rebuilt in `merge_judgments()` too, since a judged row can change which member leads. `do_next` stays as the batch list `--terminal` prints; the board no longer renders it.
- **`render.py`**: the strip, its card JS (targets/claims/lead, the partial-lit tally), CSS and help section are gone. A row button can carry `data-covers`; while a chain is lit, the covered row reads "covered by the chain from #N", counts as decided on the progress line, and a decision picked there puts the chain out. On a chain lead the chain outranks a bare approve for the coloured slot.
- **Tests**: the jsdom composer scenarios are rewritten around rows and the chain (lit, contradicted from the covered row, re-lit, cleared); render and analyze tests cover the row buttons and their rebuild. `make test-review-pipeline` is green.
- **Docs**: `SKILL.md`, `reading-the-board.md`, `action-menus.md` and the review-v3 README describe the row buttons instead of the strip.

Verified in a headless browser: pressing the chain on the lead row composes `--chain C1`, marks the next link covered, and the progress line counts both.

n/a

### Related issues (optional)

n/a

https://claude.ai/code/session_01NQDJm74v2U9qB9hSpS5fXk

---
_Generated by [Claude Code](https://claude.ai/code/session_01NQDJm74v2U9qB9hSpS5fXk)_


CamSoper:
The three stale `Do-next` comments this review flagged are fixed in 7c0ef86 -- the `only_collisions_hold` docstring, the `test_analyze.py` section banner, and the `test_rows_are_in_pr_number_order_inside_their_group` docstring.

The run before that errored without publishing: the brief's one ⚠️ row was missing its closing table pipe, so `v3-finding-grammar` refused the card and nothing went out. The push then triaged as `already-reviewed` because `review:error` was still set, so this is the explicit retry. The normalizer gap that let a missing pipe take down a whole review is fixed in [pulumi/docs#21756](https://github.com/pulumi/docs/pull/21756).

@claude #update-review

---

CamSoper:
@claude I pushed a fix for F1 in 7900618 #update-review

`covers` is now set only in the `merges_on_stamp(lead)` branch -- `"covers": [nxt] if merges else []` -- which is the first of the two options F1 offered. An empty list renders no `data-covers` attribute, so the covered-row marking never fires for a human-authored lead, and the label and help (which already described the two-run path) are unchanged. Added a regression test asserting the human-lead chain is still offered, still labelled "approve, then unblock #N", and covers nothing; it fails if `covers` goes back to unconditional.

One note for the pipeline, not for this PR: F1's Where cell reads `scripts/review-v3/analyze.py` L199, but the code it describes is at L1195 -- L199 is a docstring line inside `parse_hunks`. That wrong anchor is why this push didn't auto-refresh the card: `auto-refresh-gate.py` correctly reported `hunk scripts/review-v3/analyze.py:1162-1168 is outside outstanding finding lines` and stood down, so the fix path silently fell back to this comment.

---

PR #21757, opened 2026-09-19
docs-review: flag a finding whose line anchor isn't in the PR's diff

CamSoper:
### Proposed changes

A finding's `Where` cell is the only machine-readable pointer from the card back into the code, and `auto-refresh-gate.py` acts on it: it refuses to dispatch a refresh unless **every** hunk of the author's push overlaps an outstanding finding's line range. So a wrong line number doesn't just misdirect the reader -- it silently breaks the "push a fix and the card refreshes itself" path the card itself documents. The author pushes a correct fix, no banner appears, the finding stays open, and nothing says why.

That happened on [pulumi/docs#21748](https://github.com/pulumi/docs/pull/21748). F1 cited `analyze.py` L199 for code at L1195 (L199 is a docstring line in `parse_hunks`, ~996 lines away). The gate did its job perfectly on bad input:

```
auto-refresh gate: fire=false — hunk scripts/review-v3/analyze.py:1162-1168 is outside outstanding finding lines
```

The fix had to be claimed with an explicit `#update-review` comment instead. The wrong anchor is still visible in the published card, whose resolved-finding link points at `...R199`.

**What changed:**

- **`build-evidence.py`**: optional `--pr-diff`. `parse_diff_ranges()` pulls the head-side span of each hunk per file (findings cite lines in the head version); `anchor_note()` marks any finding whose cited line falls outside its own file's changed lines. The tolerance is ±3, matching `auto-refresh-gate.SLACK_LINES` so the two can't disagree about what's reachable. Flagged findings get `anchor_ok: false` plus a note naming the ranges that *did* change.
- **`render-evidence-html.py`**: the evidence page shows `file:199 ⚠︎ unverified`, so a line number the pipeline couldn't place never reads as authoritative.
- **`validate-evidence.py`**: learns the two keys (the schema is closed via a hand-rolled `additionalProperties:false`, so they'd otherwise be rejected) and holds them to their meaning -- `anchor_ok` is only ever `false`, and never appears without a non-empty note.
- **`claude-code-review.yml`**: fetches the PR diff in the v3-build step and passes it.

**Advisory on purpose.** A bad anchor degrades one automation; refusing to publish over it would cost the whole review -- which is the failure mode v3's fail-closed design already over-serves, and the one [pulumi/docs#21756](https://github.com/pulumi/docs/pull/21756) exists to reduce. Nothing here can fail the build: an unfetchable or unreadable diff leaves the check off, and a finding in a file the PR doesn't touch is never judged, since a pre-existing issue or a caller in an untouched file may legitimately point outside the diff.

Two hazards worth calling out, both caught by re-reading the diff before pushing and both fixed: the v3-build step runs under `set -euo pipefail`, where `[ -s file ] && VAR=...` exits the step when the file is empty -- which would have meant `ready` never got set and **nothing published**, precisely the bug class this PR is meant to remove. It's an `if` block now. The step also had no `GH_TOKEN`, so the `gh api` fetch would have failed silently and left the check permanently dead.

**Verification.** `make test-review-pipeline` green (pytest across `docs-review`, `review-v3`, `review-admin`, plus every `--self-test`). New self-test coverage in `build-evidence.py` (hunk parsing, the slack boundary, the out-of-range case, the file-not-in-diff exemption, and that no `--pr-diff` leaves findings untouched) and in `validate-evidence.py` (the flag accepted, `anchor_ok: true` rejected, a missing or empty note rejected). Run against the real #21748 diff: L199 is flagged and names the true ranges (`101-107, 1151-1209, 1216-1239, 1589-1595`); L1195 comes back clean.

n/a

### Related issues (optional)

n/a

https://claude.ai/code/session_0121HyoY69ME8PrXurw4BNTi

---
_Generated by [Claude Code](https://claude.ai/code/session_0121HyoY69ME8PrXurw4BNTi)_


CamSoper:
@claude I pushed fixes for F1 and F2 in ba23ba8 #update-review

Both were real, both reproduced before fixing, and both regression tests fail without their fix.

**F1** — right, and it's the exact false positive the brief called out as the wrongness that would matter. `anchor_note`'s docstring already promised the pre-existing exemption; only the file-not-touched half was implemented. It takes the bucket now and returns early for `preexisting`. Reproduced: a `preexisting` finding at `a.md` L8 against a diff touching 100-120 came out `anchor_ok: false`.

**F2** — right, and worse than unreachable code: with `b/` required, `+++ /dev/null` never matched, so on a deletion `path` kept the *previous* file and the deleted file's `@@ -N,M +K,0 @@` landed on it as a bogus `(K, K)` span. Reproduced: `{'a.md': [(5, 8), (0, 0)]}` where it should be `{'a.md': [(5, 8)]}`. The prefix is optional now.

One observation for the pipeline, not this PR. The anchors on this review were correct, and auto-refresh still stood down: `hunk build-evidence.py:212-218 is outside outstanding finding lines`. The gate requires *every* hunk to overlap an outstanding anchor, so a fix that also touches the line above the flagged one, or adds its own regression tests, can't qualify. Fixing an anchor-shaped finding properly seems to guarantee the explicit-comment path.

---

PR #21782, opened 2026-09-21
Show the Sentinel's gate-status comment on every PR in preview mode

CamSoper:
### Proposed changes

While `REVIEW_V3_SENTINEL` is `report`, the Sentinel's pinned gate-status comment only appeared on PRs wearing `sentinel:preview` -- and the only PRs wearing it were the ones three workflows labelled themselves (content-review, glow-up, link sweeps). That exercises the renderer and not a single reader, which is the half worth testing before the gate gets teeth. So the comment now posts on every PR the Sentinel evaluates, in both modes, the label is retired, and the preview disclaimer is explicit about what it is.

**The disclaimer.** `[!WARNING]` alert, heading suffixed "(preview)", check-run titled "Preview -- would be: X (not enforced yet)":

> ## ⚠️ Preview mode -- this check is NOT blocking your merge
>
> The Sentinel is in **preview (report-only) mode**: its check-run concludes `neutral` whatever the gates below say, so nothing here gates this PR. The rows are informational and **safe to ignore for now**. It *would* have concluded `failure`.
>
> **This will be enforced in the near future.** Once it is, every row below has to be green before this PR can merge. The rows are the sign-offs a merge needs -- some yours, some a reviewer's or a deploy's. Tell us in `#docs` if one looks wrong.

Both halves are tested. "Safe to ignore" alone trains people to scroll past the comment; "will be enforced" alone reads as a blocked PR. Tests also keep the banner from saying "you can merge" or framing a red row as the author's homework -- G3 is somebody else's approval, G4 a deploy.

**Evaluator** (`scripts/review-v3/sentinel.py`):

- `PREVIEW_LABEL` and the verdict flag it set are gone; `--status-comment` writes in both modes. `--update-strip` stays enforcing-only -- it edits the author card, which is someone else's comment.
- Report-only stamping of title/summary/conclusion was duplicated on two return paths; now `_stamp_report_only`.
- Two bugs the wider audience would have hit: preview headlines said "before this can merge" under a banner saying nothing blocks you (now "would block this merge once the Sentinel is enforced"), and a waived PR in preview reached that headline at all, because the waive branch read `conclusion` -- always `neutral` in preview. It reads `would_be` now.

**Label retirement.** `content-review-article.yml` and `check-links.yml` no longer apply `sentinel:preview` (the link-sweep lane's labelling step is deleted), and `review-sentinel.yml` drops it from the label-event filter -- which removes the "the lanes label just after opening, so the `opened` run predates the label" wart instead of working around it. `.github/labels-pr-review.md` gains a retired-labels section with the `gh label delete`; CONTRIBUTING.md and `scripts/review-v3/README.md` describe the new behavior. **Still to do by hand:** delete the label from the repo.

`make test-review-pipeline` and `make lint` pass. No behavior change while `REVIEW_V3_SENTINEL` is unset -- the job doesn't run.

n/a

### Related issues (optional)

n/a

https://claude.ai/code/session_01WPaDWBuHvBQikQiZYv9meR


CamSoper:
@claude #update-review

CamSoper:
@claude #update-review

Merged master to pick up the review pipeline fixes from #21791. No content changes in this push -- please refresh against the new head.

CamSoper:
@claude #update-review

Fixed in 13bd121:

- F1: real bug. The gateless branches now render the unstamped summary (new `base_summary` on the verdict), so a not-governed PR in preview says "Not governed -- ..." again. Also dropped the "every row below has to be green" sentence from the banner when there's no gate table under it. New test: `test_not_governed_preview_comment_still_says_not_governed`.
- F3: both assertions now match `\bowe[sd]?\b` instead of the bare substring.

F2: accepting as-is -- a draft isn't evaluated, so there's no "would be" conclusion for a preview banner to report, and "Draft PRs are not gated" is true in both modes. Stamping it would put a "would be: neutral" banner over a comment with no gates.

CamSoper:
@claude #update-review

CamSoper:
@claude F4: unintended -- good catch. Restored in 45034d9.

The test went out with the F3 phrasing asserts by accident, not by decision. I resolved the merge with `git checkout --ours` on the test file, which takes my whole copy of it and discards 13bd121's non-conflicting addition along with the two copy assertions I actually meant to drop. The scripted edit I wrote to trim that test then silently no-op'd, because `str.replace` doesn't fail when it finds nothing -- so the loss left no trace.

`test_not_governed_preview_comment_still_says_not_governed` is back, minus the `"every row below"` copy assert (that phrase is prose I've already rewritten three times in this review; the `| Gate |` assert covers the same branch structurally). Verified it actually catches the regression rather than passing by accident: clearing `base_summary` on the verdict drops "Not governed" from the comment and leaks "PREVIEW MODE" into it, which is exactly F1.

#update-review

---

PR #21784, opened 2026-09-21
Stop /pr-review re-offering a base merge that already failed; add --ask-fix

CamSoper:
### Proposed changes

Two fixes to the `/pr-review` board, both out of one report: **merge base & retry** on #21485 looked like it did nothing.

It wasn't doing nothing. `--unblock` tried the merge, hit a real conflict in `content/blog/mlops-huggingface-llm-aws-sagemaker-python/index.md` (master and the branch had each rewritten an adjacent link-reference line), aborted as designed, and reported it -- to that run's stdout and nowhere else. The next collect saw the same `mergeable:dirty`, so the next board offered the same button, which failed the same way. The board had no memory of a refused merge, so the one button on the page that could do nothing did it indefinitely.

**1. A conflicted `--unblock` now leaves a record.**

- `act.py` posts the conflicted files on the PR under `<!-- PR_REVIEW_UNBLOCK_CONFLICT -->`, keyed to the head it tried. Best-effort: a failed comment never turns a failed merge into a raised exception.
- `collect.unblock_conflict` reads it back, but only while it still describes the current head -- so any push spends the record and the button returns.
- The row then carries `unblock:refused:conflict` and offers the send-back instead, reusing the path that already existed for a branch `act.py` may not push to. No new vocabulary.
- Provenance follows `sentinel._find_comment`'s rule (exact marker line in the body's first three) with a different roster, because `act.py` comments as whoever ran the board, not as the review app.

**2. A stuck workflow row gets a second way out: `--ask-fix N`.**

It already offered **fix it yourself** (`/address-review`, an interactive run that never joins the batch). It now also offers **ask @claude to fix them** -- one comment reading `@claude fix F1 and F3 #update-review` with the findings' summaries beneath, handing the same list to the agent already watching the PR. It's an ordinary write, so unlike the handoff it batches with everything else.

- Only the review's *open* rows are named. A style or pre-existing finding is the review's own take-it-or-leave-it, and asking for it would turn an optional note into a push.
- A row with nothing open is refused at plan time rather than posting `@claude fix  #update-review`.
- The two do the same job, so they share an exclusive group (`data-exclusive="fix"`) and lighting either puts the other out. `clearRow` can't pair them, since a handoff is deliberately not a decision.

**Tests** (`make test-review-pipeline`, 532 passing): the conflict record and its head-keying, the refusal to re-offer, `--ask-fix`'s body / bucket rules / empty-row refusal, and a jsdom click sequence for the pairing across the handoff/fragment boundary.

`make lint` clean.

n/a

### Related issues (optional)

The conflict on #21485 is resolved separately, by hand, on that branch -- it needed someone to say which side won, which is exactly what `--unblock` refuses to guess.

---

https://claude.ai/code/session_01NBGUXcSpcsEjRQBQ93foTa

---
_Generated by [Claude Code](https://claude.ai/code/session_01NBGUXcSpcsEjRQBQ93foTa)_


CamSoper:
@claude pushed fixes for F1 and F2 in fe74f487 #update-review

**F2** was the good catch, and it was worse than it looked: `_unblock_conflict_authors` asked the token who it was with `gh.me()`, but `collect()` had already resolved the approver just above it (`--approver` if given, else `gh.me()`). So a run started with `--approver`, or one against a backend that can't answer `GET /user`, checks a different login than the rest of the queue -- or none -- the conflict record goes unseen, and the board re-offers the same dead **merge base & retry**. That's the exact bug this PR exists to delete, walking back in through a side door. The resolved approver is now threaded through `collect_pr`, and both sides of the comparison go through `norm_login` the way `act.py` writes the record. Added coverage for the case-normalization and the bot-authored path.

**F1** confirmed -- the block was a lazy continuation of the regens bullet and split `workprentice[bot]` into a second list. Moved it below the `push_allowed` sentence, just before the handoff section, blank lines both sides.

`make test-review-pipeline` and `make lint` clean before pushing.

---

PR #21785, opened 2026-09-21
Rewrite /address-review for the v3 review surface

CamSoper:
### Proposed changes

`/address-review` was half-migrated. `review-worklist.py` has been v3-aware for a while -- it detects the author card, parses `REVIEW_STATE`, extracts `F<n>` ids, reads the reviewer brief, and knows the `author-answer` / `reviewer-check` buckets. The skill's prose still described v2 exclusively. Net effect: an agent following the skill got a correct v3 worklist and then applied v2 instructions to it. Nothing crashed, it just burned model calls on `#update-review` for things `/resolve` does for free, treated blocking ❓ items as advisory, and called PRs "ready" on the wrong gate.

This makes v3 the described path and keeps v2 as an explicit legacy section.

**`.claude/commands/address-review/SKILL.md`**

- Surface detection is explicit now, and tests for `CLAUDE_REVIEW_AUTHOR` **first** -- the v3 card keeps `<!-- CLAUDE_REVIEW 1/1 -->` as a legacy alias on its first line, so the obvious check reads every v3 PR as v2. That one bit me while writing this.
- Ids are `F<n>`, stable across re-reviews, not synthetic `bucket:L<n>`.
- Real bucket order: 🚨 outstanding and ❓ author-answer both block merge; ⚠️ reviewer-check is advisory and lives on the brief, addressed to the approver.
- `REVIEW_STATE` is the ledger; `.review-worklist-<PR>.json` is a cache for style items (no `F<n>`) and in-progress notes.
- Adds the `/resolve F<n> <disposition>[: reason]` lane with its actual constraints: PR author or write/maintain/admin, fails closed on an unreadable permission, ignored for bot accounts, reason required for `deferred`/`accepted`/`not-applicable`. Includes a table for picking between push / `/resolve` / `#update-review`.
- Documents `author-accepted`. A non-`fixed` disposition by the PR author collapses to it with the original preserved, and stays visible to the human approver. Worth knowing before you tell someone to refute their own finding.
- "Done" is the Sentinel where it runs -- findings answered, routed team approved, CI green, no conflict -- not `--require-clean` alone. G3 is never the author's to clear, and a push can only dismiss approvals.
- Freshness uses `pinned-comment.sh last-reviewed-sha` instead of grepping the head sentinel out of a fetched body.

**`references/dispositions.md`** -- where a disposition actually lives, the two write lanes, the `author-accepted` rule, and per-bucket rules for ❓ and ⚠️.

**`references/watching.md`** -- mechanical lanes skip the model review but never the approver, and `#new-review` is called out as the only thing that migrates a v2 PR (people will reach for it as a retry otherwise).

**Two stale rollout claims corrected:** `output-format.md` said the v2 surface "remains the default" and `AGENTS.md` called v3 a staged rollout. It's been on repo-wide since the 2026-09-14 flip.

#### v2 straggler roster

Verified by reading the comments on every open PR carrying a review marker, not by search -- GitHub's comment index is lossy (#21760 has a textbook v3 card and turns up in no marker search at all). **17 open PRs still carry only a v2 monolith:**

| Blocking findings | PRs |
|---|---|
| none (`review:no-blockers`) | #20631, #20949, #21039, #21083, #21149, #21172, #21198, #21210, #21412, #21418, #21432, #21434, #21485, #21487, #21491, #21554 |
| open 🚨 (`review:outstanding-issues`, draft) | #20914 |

No bulk `#new-review` in this PR, deliberately. Sentinel grandfathers a v2 review current at head (`sentinel.py:816`) and scores G2 by counting 🚨 bullets (`:902`), so the sixteen clean ones pass as "legacy review clean" and merge today. Regenerating them costs a full review each and resets any ledger the author already built.

Two things the roster turned up that are worth knowing:

1. **v2 PRs never migrate on their own.** #21485 was refreshed twice *after* the flip, most recently today, and came back v2 both times -- `claude-update.yml` grandfathers the v2 refresh path on purpose. Attrition clears these only when they merge or close, so a long-lived PR stays v2 indefinitely.
2. **A legacy PR with open 🚨 has no way to answer them.** `/resolve` is gated on the v3 surface, so G2 only clears when the findings disappear from the comment. #20914 is the one in that state. That's the case where `#new-review` actually buys something, and the skill now says so.

Separately: #21485's own review notes it's carrying `needs-author-response` and `review:in-progress` with nothing outstanding. Looks like a label-reconciler miss, not fixed here.

#### Verification

- `make lint` -- 1794 files parsed, 0 errors; Prettier clean.
- `make test-review-pipeline` -- all pytest suites, standalone harnesses, and `--self-test` flags pass (including `review-worklist.py` and `sentinel.py`).
- `make build` / `make serve` were not run: Hugo isn't installed in this container, so `make ensure` can't complete. Docs-only change under `.claude/` plus one `AGENTS.md` line, so there's no template or content surface to render.

n/a

### Related issues (optional)

n/a

https://claude.ai/code/session_01AbRpu4W5wyCbMneG1NbTwk

---
_Generated by [Claude Code](https://claude.ai/code/session_01AbRpu4W5wyCbMneG1NbTwk)_


CamSoper:
@claude #update-review

**F1 — fixed** (`5a8229c`). You were right and I had it backwards. `cmd_fetch` goes through `fetch_pinned_bodies` → `list_pinned_comments`, which selects on `^<!-- CLAUDE_REVIEW N/M -->` only; the union-of-role-cards code I'd read is in `cmd_clear`. So the author card comes back (legacy alias on line 1) and the brief doesn't. Dropped the union clause, kept the marker-ordering advice -- it holds for the same reason it always did -- and noted that `review-worklist.py` fetches the brief itself.

**F2 — fixed too** (`5a8229c`), even though it's on the reviewer's card rather than mine. It's the exact species of rot this PR exists to remove, so leaving it would have been funny in the wrong way. The sentence now says to read the mode off the check run instead of trusting any documented rollout state. Checking that before writing it turned up something I'd have gotten wrong: `neutral` covers draft PRs as well as report-only, so "neutral means report-only" would have been a fresh bad claim inside the fix for a bad claim. Both are now stated.

**One more, unprompted** (`7e90f80`): that first push should have tripped the auto-refresh gate and didn't, because fixing F1 and F2 together puts a hunk on a ⚠️ line, and `parse_anchor_ranges` only anchors on the author card's 🚨 and ❓ sections -- so the mixed push gets declined as a whole. My own Step 6 said "every hunk on a flagged line," which is what made that look free. Corrected in both SKILL.md and dispositions.md, since a skill that misdescribes the gate will keep costing people a round.

`make lint` clean on all three.

---

CamSoper:
@claude #update-review

Retrying the errored run, and answering the ⚠️ item while I'm here.

**Reviewer-check item (Step 3 count line) — fixed** (`0c8258a`). It's right: `total` is `len(items)` and `remaining` filters only on `optional`, which reviewer-check items aren't — so the template announced a number smaller than `--require-clean` measures. Folded them into the total and stated the counting rule so the next person doesn't re-derive it.

Also in that commit: this round left the PR carrying `review:error` and `review:no-blockers` simultaneously, and Step 1 presents the state labels as mutually exclusive. That's true by design, but a run that dies *after* writing its cards falsifies it in practice, and an agent reading that table top-down takes the first row that matches. Step 1 now says to believe the cards and their `CLAUDE_REVIEW_HEAD` over the labels when they disagree.

Three things from this round that look like pipeline bugs rather than anything about this diff, flagged rather than touched:

1. **`high_water` went backwards, 2 → 1, and `F1` was reused.** The round-1 F1 (the `fetch` union claim) and F2 (the Sentinel sentence) are gone, and the new reviewer-check item was issued `F1`. Per `scripts/review-v3/README.md` the ids are "monotonically increasing per PR, never reused," with the high-water mark travelling in the evidence object, and "the update lane preserves existing IDs." Practical cost: my round-1 comment saying "F1 — fixed" now sits above an open, unrelated `F1`.
2. **The run errored after publishing.** Both cards were rewritten correctly at `7e90f80` — head advanced, findings cleared, new item added — and *then* the lane posted "review errored." Whatever failed, it failed downstream of a good publish.
3. **The two state labels above.** Probably a consequence of 2, but `review-label-reconcile.yml` didn't clear it.

Happy to open issues for any of these if they aren't already known.

---

CamSoper:
@claude #update-review

**Fixed** in `1508820`, and the finding is right — my previous round is what introduced it. `extract_items_v3` emits only 🚨 / ❓ / ⚠️ / ✏️ items, all `optional: False`; `optional: True` lives solely in the v2 `BUCKETS` table. So there is no 💡 row on v3 at all, and "remaining excludes only the optional ones, which is 💡 pre-existing alone" was v2 logic sitting in a v3 paragraph. Step 3 now says nothing is optional on v3 (with v2's behavior in parentheses), Step 4's walk ends at ⚠️ with 💡 marked v2-only and the v3 route spelled out, and dispositions.md matches.

Separately, and not a finding about this diff -- **the cards raced on this PR and a recorded disposition was lost.** Two runs composed against `0c8258a` roughly a minute apart and published out of order:

- composed 20:26:14 → **v1**, one open 🚨 F1, `high_water: 1`, `findings: {}`
- composed 20:26:53 → **v2**, that F1 resolved, `REVIEW_STATE` carrying `"F1": {"disposition": "fixed", "sha": "0c8258a22339"}`

The v2 card published at 20:27. The **older** v1 card then overwrote it at 20:30, and the `fixed` disposition went with it. Same shape as the earlier failure, where `apply-update` rejected a patch naming `F2` because a different run had already replaced the cards it was written against -- and the same shape as the `high_water` 2 → 1 regression and the `review:error` + `review:no-blockers` pair.

`publish_guard.py` guards exactly this for the Sentinel check-run ("Overlapping runs for one PR are therefore normal, and this guard is what keeps a stale verdict from overwriting a fresh one"). The card publish path has no equivalent: no head/run-id check, and nothing rejecting a card whose `high_water` is below the prior evidence's, even though `record-evidence.merge_latest` already takes a `max()` on the S3 side.

Not touching it here -- it's shared pipeline code and this is a docs PR. Flagging it because a lost disposition is worse than a crash: the crash announced itself.

---

CamSoper:
@claude #update-review

**Fixed** in `e28dfd8` — the walk example now reads `[1/5]`, matching the count line above it. My round-2 edit changed 4 items to 5 and left the example behind.

Sending this as a mention rather than relying on a push refresh, because the fix lands on a ⚠️ line and `parse_anchor_ranges` only anchors on the author card's 🚨 and ❓ sections. The gate declined the push (`auto-refresh: skipped`), which is the behavior this PR documents.

**The cards raced again, and a second disposition was lost.** Same PR, same head `1508820`:

- composed **20:49:07** → v2, `REVIEW_STATE` carrying `"F1": {"disposition": "fixed", "sha": "1508820834ad"}`
- composed **20:48:08** → v1, `"findings": {}`

The v2 card published first. The **older** v1 card overwrote it at 20:51:52, and the `fixed` disposition went with it — same as the 20:27 → 20:30 loss earlier. Two runs were in flight because the push tripped the auto-refresh gate *and* I mentioned `#update-review` a minute later; that's my error and this repo's own skill warns against it, but a caller mistake shouldn't be able to delete a recorded disposition.

Worth stating plainly: **a lost disposition is worse than the crash that started this thread.** The crash announced itself with `review:error` and a comment. This just quietly empties the block the Sentinel reads for G2, and nothing anywhere says it happened.

Still not touching the pipeline from this PR. The evidence is all on the thread if someone wants to pick it up.

---

CamSoper:
The card race diagnosed above now has a fix in #21788 (draft) -- a stale-publish guard at `pinned-comment.sh upsert --role`, with the two losses from this thread as regression tests. Keeping it out of this PR since it's shared pipeline code.

---

PR #21786, opened 2026-09-21
Fix CloudFront invalidation globs that skip directory roots

jeffmerrick:
CloudFront invalidation paths ending in `/*` match everything below a directory but not the directory itself, so `/dev/*` never invalidated `/dev/`. On the 2026-09-21 marketing-web deploy every path under `/dev/` picked up the new build immediately while `/dev/` served a 22-minute-old object (`age=1324`) until the thirty-minute cache policy expired it.

* `invalidate-dev-cache.yml`: `/dev/` and `/assets/` added to the default paths, in both the `workflow_dispatch` input default and the `PATHS` fallback. `/assets/` is there to keep the two surfaces listed the same way - nothing is served at that exact path.
* `run-pulumi.sh`: the main deploy had the same hole, where the section landing pages were never getting invalidated. Every section is now listed as both `/x/` and `/x/*`, one pair per line. `/docs/`, `/blog/`, `/product/`, `/pricing/` and `/contact/` are all real pages.
* `/index.html` and `/` stay as they were in `run-pulumi.sh` - the root has no glob form short of `/*`, which would also sweep `/fingerprinted/*`.
* Won't show up until the next marketing-web deploy - `/dev/` should come back without a climbing `age`.


.github/workflows/invalidate-dev-cache.yml
```diff
@@ -13,8 +13,9 @@ on:
   workflow_dispatch:
     inputs:
       paths:
+        # Keep in sync with the PATHS fallback below.
         description: Space-separated CloudFront paths to invalidate
-        default: "/dev /dev/* /assets /assets/*"
+        default: "/dev* /assets*"
```
cnunciato: But `/dev/` is actually (implicitly) `/dev/index.html` — and this change would be a bit "greedier" in that it'd invalidate things like `/devops` (not that we have this, but you get the idea).  

If the issue is that `/dev/` was left invalidated, I'd sooner add `/dev/` than wildcard anything after `/dev`.

jeffmerrick: @cnunciato Ah, got it! I changed it to just add the paths instead of the wildcard.

cnunciato (approved):

PR #21787, opened 2026-09-21
Add LTS policy to public-facing docs

guineveresaenger:
Fixes https://linear.app/pulumi/issue/PVD-4015/update-public-docs. 

This is part of the policy rollout for long-term support on select providers.


CamSoper:
@claude #new-review

CamSoper:
@claude #new-review

CamSoper:
@claude #new-review

The earlier review runs on this PR errored because of bugs in the review pipeline, not anything in this PR. Those are fixed in #21791, and I merged master into this branch so the fix applies here. Re-running the review from scratch.

content/docs/support/provider-support-policy.md
```diff
@@ -0,0 +1,75 @@
...
+        weight: 3
+        identifier: support-provider-support-policy
+---
+
+Pulumi actively maintains the latest released major version of every provider it publishes.
+For a set of widely used providers, Pulumi also ships security updates for the previous major version for up to 12 months, or until the next major version is released, whichever comes first.
+
+## Providers covered
+
+- `pulumi-aws`
```
jkodroff: These should be links to the providers in the Registry and use the package names rather than the repository names. Users don't generally know the repository names.

content/docs/support/provider-support-policy.md
```diff
@@ -0,0 +1,75 @@
...
+- `pulumi-azure-native`
+- `pulumi-azuread`
+- `pulumi-command`
+- `pulumi-docker-build`
+- `pulumi-gcp`
+- `pulumi-kubernetes`
+- `pulumi-tls`
+- `pulumi-vault`
+
+## What ships on the long-term support branch
```
jkodroff:
```suggestion
## What ships on the long-term support (LTS) branch
```

Add the acronym here since we use it below and leave it to the user to figure out what it stands for.

content/docs/support/provider-support-policy.md
```diff
@@ -0,0 +1,75 @@
...
+
+- Fixes for CVEs affecting the provider's code or its direct dependencies.
+- Fixes for broken internal infrastructure on the LTS branch needed to ship the above.
+
+An LTS release does not contain:
+
+- New upstream API versions.
+- New resources or new resource properties.
+- Performance improvements.
+- Non-security bug fixes, including customer-reported bugs.
```
jkodroff:
```suggestion
- Non-security bug fixes.
```

I think we can just leave it as "bug fixes".

content/docs/support/provider-support-policy.md
```diff
@@ -0,0 +1,75 @@
...
+- Compatibility with newer versions of the Pulumi engine, the Pulumi SDKs, or language runtimes.
```
jkodroff: This is framed like it might not be backwards compatible, which I do not think is what we mean here. What I think we mean here is "it may not take advantage of new features in the CLI or SDK".

content/docs/support/provider-support-policy.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Engine support
+
+Providers will be compatible with the version of Pulumi they were frozen at unless otherwise prompted via security fixes.
```
jkodroff, line 47: This can be combined into the previous section - it's part of the same "contract" we're setting.

content/docs/support/provider-support-policy.md
```diff
@@ -0,0 +1,75 @@
...
+
+At any time, exactly one major version of a covered provider is the LTS version. Older majors receive no patches.
+
+| Support phase | Which version | Duration | What ships |
+| --- | --- | --- | --- |
+| **Current major (N)** | The latest released major. If v3 is the newest release, v3 is N. | Until the next major, N+1, ships. | New features, new resources, bug fixes, security patches, dependency upgrades, and new upstream API versions. |
+| **Long-term support (N−1)** | The major immediately before the current one. If v3 is N, then v2 is N−1. | Up to 12 months from the general availability of N, or until N+1 ships, whichever comes first. | Security updates in the provider's code or its direct dependencies. |
+| **End of life (N−2 and older)** | Every older major. | — | Nothing. |
+
+### What the policy guarantees
```
jkodroff, line 49: This duplicates the previous section. (Feels like we're belaboring the point.)

content/docs/support/provider-support-policy.md
```diff
@@ -0,0 +1,75 @@
...
+
+- Exactly one long-term support major exists per provider at any time.
+- When N+1 ships, it becomes the new N, and the old N becomes the new LTS major.
+  The old long-term support major reaches end of life on the same day, regardless of how much of its 12-month window remained.
+- Within the long-term support major, only the latest minor version receives patches. Earlier minors are frozen.
+
+## How to consume LTS releases
+
+- LTS releases follow the same distribution channels as current-major releases (i. e. the relevant SDK package registries for each language), but are tagged differently.
```
jkodroff:
```suggestion
- LTS releases are available through the same distribution channels as current-major releases: standard language package feeds like https://npmjs.org, PyPI, etc.
```

content/docs/support/provider-support-policy.md
```diff
@@ -0,0 +1,75 @@
...
+
+Providers will be compatible with the version of Pulumi they were frozen at unless otherwise prompted via security fixes.
+
+## Summary
+
+At any time, exactly one major version of a covered provider is the LTS version. Older majors receive no patches.
+
+| Support phase | Which version | Duration | What ships |
+| --- | --- | --- | --- |
+| **Current major (N)** | The latest released major. If v3 is the newest release, v3 is N. | Until the next major, N+1, ships. | New features, new resources, bug fixes, security patches, dependency upgrades, and new upstream API versions. |
```
jkodroff: This table can be simplified. If we want to provide an example, we can do that in a sentence or bullet point or two.

content/docs/support/provider-support-policy.md
guineveresaenger: The provider at this version may not work with more modern versions of the engine. That _can_ happen. You may need to use an older version of the engine.

content/docs/support/provider-support-policy.md
guineveresaenger: I moved the summary to the end (where it should be).

guineveresaenger:
Blocked until explicit approval from Product

tyler-dunkel:
@guineveresaenger as this is a public repo, can we link the public GitHub issue as opposed to our linear one?

content/docs/support/provider-support-policy.md
```diff
@@ -0,0 +1,75 @@
...
+        weight: 3
+        identifier: support-provider-support-policy
+---
+
+Pulumi actively maintains the latest released major version of every provider it publishes.
+For a set of widely used providers, Pulumi also ships security updates for the previous major version for up to 12 months, or until the next major version is released, whichever comes first.
+
+## Providers covered
+
+- [`pulumi-aws`](/registry/packages/aws/)
```
jkodroff:
```suggestion
- [AWS](/registry/packages/aws/)
```

(Similarly for the others in this list - use the package name instead of the repo name.)

content/docs/support/provider-support-policy.md
```diff
@@ -0,0 +1,75 @@
...
+        name: Provider support policy
+        parent: support-home
+        weight: 3
+        identifier: support-provider-support-policy
+---
+
+Pulumi actively maintains the latest released major version of every provider it publishes.
+For a set of widely used providers, Pulumi also ships security updates for the previous major version for up to 12 months, or until the next major version is released, whichever comes first.
+
+## Providers covered
```
jkodroff, line 17: The section header plus the intro doesn't make it clear that there are 2 categories of provider support.

I would do something like:

- Introduce the concept of long-term support (LTS) in the intro.
- Explain in this section header that this is the list of providers covered by LTS. (Just giving the header with no intro paragraph to explain what the list is feels a little too terse.)

content/docs/support/provider-support-policy.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Summary
+
+At any time, exactly one major version of a covered provider is the LTS version. Older majors receive no patches. **N** is the latest released major version: if v3 is the newest release, then v3 is N, v2 is N−1, and v1 and earlier are end of life.
+
+| Support phase | Duration | What ships |
+| --- | --- | --- |
+| **Current major (N)** | Until N+1 ships. | New features, new resources, bug fixes, security patches, dependency upgrades, and new upstream API versions. |
+| **Long-term support (N−1)** | Up to 12 months from N's release, or until N+1 ships. | Security updates in the provider's code or its direct dependencies. |
+| **End of life (N−2 and older)** | — | Nothing. |
```
jkodroff: "nothing" is too harsh and recalls an angry Willy Wonka. Suggest making this bullet points instead of a table since it's not that data heavy. We can just say "Older versions are not supported and will not receive updates" to soften the language.

content/docs/support/provider-support-policy.md
guineveresaenger: What exactly isn't clear? 
"For a set of widely used providers, Pulumi also ships" --> "Providers covered".

I'm not sure I understand what's missing.

PR #21788, opened 2026-09-21
Stop a stale review run from overwriting a newer card

CamSoper:
### Proposed changes

Two review runs on one PR are routine: a push trips the auto-refresh gate while a `#update-review` mention dispatches the update lane, triage chains re-dispatch, someone mentions twice. Nothing serializes their card writes, so whichever finishes **writing** last wins -- even when it composed first. The author card carries `REVIEW_STATE`, so losing that race deletes recorded dispositions with no error, no label, and nothing in the thread saying it happened.

Caught it twice on #21785 inside half an hour, both at the same head `1508820`:

| composed | published | `REVIEW_STATE.findings` |
|---|---|---|
| 20:49:07 | 20:49 | `{"F1": {"disposition": "fixed", "sha": "1508820834ad"}}` |
| **20:48:08** | **20:51** | `{}` |

The older composition published second and won. The Sentinel's G2 input went blank, and the only evidence was the disposition no longer being there. The earlier `apply-update: contract violation: F2 is not an open finding` crash on that same PR was the same race announcing itself louder -- a patch validated against cards another run had already replaced underneath it.

**`handoff_guard.py` can't catch this.** It compares the handoff SHA against the live head:

```python
if live_sha.lower() == handoff_sha.lower():
    return "publish", "handoff is current"
```

Head unchanged → publish, unconditionally. That is exactly the same-head case both losses took. `publish_guard.py` already solves the equivalent problem for the Sentinel check-run -- *"Overlapping runs for one PR are therefore normal, and this guard is what keeps a stale verdict from overwriting a fresh one"* -- and the cards simply never got the equivalent.

This adds it at `pinned-comment.sh upsert --role`, the one choke point every card write passes through, so both lanes are covered without touching either workflow.

**Composition time, not revision.** A `#new-review` regeneration legitimately republishes v1 over a v2 card, so comparing `vN` would block the one flow whose entire job is to reset -- while its composition time is correctly *newer* than what it replaces. Every card already carries that timestamp in its `<sub>Review vN · updated … </sub>` stamp. There's a test pinning the regeneration case specifically.

**Failure posture.** Refusing exits 0, matching `publish_guard`: the newer card is the right state and this run has nothing left to do, so it stands down with a `::notice::` rather than failing a job over having been beaten to it. Every unreadable stamp fails open -- legacy bodies, hand-written fixtures, a changed stamp format, even a missing `python3` (verified under `set -e`) -- because a guard must never cost a review its publish. `--allow-stale-overwrite` is the break-glass.

#### What this does not fix

Only the card write. A stale run still writes its evidence object to S3, so the S3 mirror can hold a composition the PR rejected. Per `scripts/review-v3/README.md` the PR is the disposition source of truth and S3 is the telemetry mirror, so the damaging half is closed -- but the divergence is real and worth a follow-up. Left out deliberately to keep this reviewable.

Also unaddressed: the `high_water` 2 → 1 regression and the `review:error` + `review:no-blockers` label pair seen on the same PR. Both look like downstream symptoms of this race rather than separate defects, but I haven't proven that.

#### Verification

- `make test-review-pipeline` -- passes, including six new cases: the #21785 loss verbatim, normal newer-over-older publish, the `#new-review` reset, fail-open on unstamped bodies, the break-glass, and the brief card.
- `make lint` -- 0 errors, Prettier clean.
- `test_role_upsert_never_invokes_spine_floor` asserted "reads no comment body at all" as a proxy for "no spine floor". The guard reads exactly one, so it now counts instead -- reading more than the card it owns is still the spine floor leaking in.
- `make build` not run: Hugo isn't installed in this container so `make ensure` can't complete. No content or template surface changed.

n/a

### Related issues (optional)

Found while working #21785.

https://claude.ai/code/session_01AbRpu4W5wyCbMneG1NbTwk

---
_Generated by [Claude Code](https://claude.ai/code/session_01AbRpu4W5wyCbMneG1NbTwk)_


CamSoper:
@claude #update-review

All three fixed in `8c992e5`. F1 especially -- that was a hole in the guard, not a nit, and I'd missed it.

**F1.** Confirmed against the code: `refresh_counts` never touches the `<sub>… updated` stamp, so `/resolve` wrote REVIEW_STATE while the card kept claiming it was last written at the previous *composition* time. Which makes the disposition erasable by exactly the mechanism the guard was added to stop:

```
20:50  review composes and publishes, stamp = 20:50
20:52  another run composes (slow), stamp = 20:52
20:55  /resolve F1 fixed  -> REVIEW_STATE written, stamp STILL 20:50
20:56  slow run publishes -> 20:52 > 20:50, allowed, disposition gone
```

`resolve-handler.py` now restamps `updated` on both cards whenever a disposition lands. `vN` and the head are left alone deliberately -- a `/resolve` isn't a new review round and doesn't move the head -- and a body with no parseable stamp round-trips unchanged, since the guard already fails open on those. I put the restamp on the resolve path rather than inside the shared `refresh_counts`, because `apply-update.py` writes its own stamp and doesn't need a second one.

**F2.** Right that a silent fail-open is the wrong default *here* of all places: "no newer card" and "couldn't look" were indistinguishable in the log. Now emits a `::warning::` naming the role and PR, then publishes.

**F3.** Fixed -- `card_stamp` returns `<epoch><TAB><ISO8601>` and the notice prints the ISO times it actually compared. There's a test asserting no bare epoch seconds survive in that message.

Three new tests: the restamp (plus `vN`/head untouched and the no-stamp round-trip) in `resolve-handler --self-test`, and the ISO notice and read-failure warning in `test_pinned_comment.py`. `make test-review-pipeline` and `make lint` both clean.

Sending this as a mention rather than leaning on the push: all three findings were ⚠️ items on the brief, which don't anchor the auto-refresh gate, and it declined the push (`auto-refresh: skipped`) exactly as expected.

---

PR #21789, opened 2026-09-21
Harden the deploy against transient 504s and stack-lock conflicts

CamSoper:
### Proposed changes

Two independent fixes for the six Build and deploy failures on 2026-09-21, plus two commits addressing review and audit findings.

**1. Retry the Vale and s5cmd downloads** (`dc04b52`)

[Run 35618729067](https://github.com/pulumi/docs/actions/runs/35618729067) died on `curl: (22) The requested URL returned error: 504` fetching the Vale tarball. That step was a bare `curl -fsSL` with no retry, so one transient 504 from GitHub's release CDN discarded a run that had already spent minutes on checkout, toolchain setup and cache restore.

Adds retry flags to every Vale and s5cmd download across the workflows — 17 call sites in 14 files, plus `.devcontainer/scripts/post-create.sh`, which was the last bare copy in the repo.

> **Correction.** An earlier version of this description claimed `--retry-all-errors` is what makes curl retry a 504 at all. That is wrong: a 504 is already in curl's transient-error set, so plain `--retry 3` would have caught the failure that motivated this PR. `--retry-all-errors` broadens retrying to connection-level failures (refused, reset) that are *not* in that set, which is the actual reason it's here. A second inaccuracy: the first commit message says the flags differ from `scripts/fetch-openapi-spec.sh` "with one deviation" — there are two, since `--retry-delay 5` is also new (those scripts use exponential backoff). The tradeoff is that a wrong version pin or moved release URL burns ~15s over four attempts before failing, with an identical error message.

**This does not cover** the other two failures that morning, which were 504s *inside* `peaceiris/actions-hugo` fetching the Hugo tarball. That action does its own retry (three attempts, all 504) and takes no flags from us. Making it resilient means caching the binary or wrapping the action — a separate change.

**2. Retry a deploy that failed only on the stack lock** (`55cd947`)

Runs [35625043036](https://github.com/pulumi/docs/actions/runs/35625043036), [35625107209](https://github.com/pulumi/docs/actions/runs/35625107209) and [35625169056](https://github.com/pulumi/docs/actions/runs/35625169056) all failed within 40 seconds of each other with `error: [409] Conflict: Another update is currently in progress`, each after ~6 minutes of building and syncing.

`scripts/await-in-progress.js` serializes the deploy by waiting on lower-numbered runs, but caps the wait at 45 minutes and then proceeds regardless. All three runs left that wait together and walked into a `pulumi up` that still held the lock.

This adds the backstop at the layer where the lock actually lives: `retry_on_stack_lock` wraps the refresh and the update with exponential backoff (30s doubling, 5 attempts). Each wrapped call gets its own budget, so a single deploy can spend up to ~15 minutes there in the worst case — time `scripts/ci-build-duration-alert.sh` cannot see, since it subtracts only the queue wait.

It retries that one error and nothing else, matched against an `error:` line carrying the 409 code rather than a bare substring of the whole output. The service returns the 409 *before* touching any resource, so nothing is half-applied and a later attempt re-plans against current state. Any other failure returns immediately. It also covers a lock the turnstile can never see: a `scripts/laptop-deploy.sh` run, or a console operation.

`preview` is deliberately left unwrapped — it doesn't take the update lock, evidenced on the F2 thread below and conceded by the review.

**3–4. Review and audit findings** (`b621df5`, `91711f4`)

Three independent adversarial reviews plus the pre-merge review raised five things, all fixed here:

- The tests proved the logic but not the **wiring** — deleting `retry_on_stack_lock` from both call sites left every test passing. Two tests now read `run-pulumi.sh` directly and fail if either call site is unwrapped.
- A non-numeric `PULUMI_LOCK_MAX_ATTEMPTS` **hung the deploy**: `[ 1 -ge five ]` exits 2, which `if` reads as false, so the cap never tripped and the doubling delay parked the job until the six-hour timeout. Both overrides are validated now.
- The 409 match was an **unanchored substring**, so a recovered warning quoting the phrase would have retried a permanent failure for minutes.
- A test fed the validation a bad delay, got the real 30s default back, and **actually slept it** — 30 seconds added to every `make test`. Measured 30.5s → 0.386s after stubbing `sleep` in the harness.
- `--retry-max-time 300`, because `Retry-After` overrides `--retry-delay` uncapped (measured: `Retry-After: 600` left curl sleeping at 75s), and none of the 14 jobs sets `timeout-minutes`.

### Testing

`scripts/pulumi-lock-retry.test.js` — 13 tests, picked up by `scripts/run-unit-tests.sh` and so by `make test`, following the bash-from-node pattern in `scripts/deploy-bucket-naming.test.js`. **Full suite: 67 passing, 0.585s.**

The tests assert **call counts**, not just exit statuses — "failed once" and "retried three times then failed" are indistinguishable by status alone, and the first draft of the non-lock test passed against a deliberately broken guard for exactly that reason.

Verified non-vacuous against mutations at each round, most recently: revert the anchored 409 match, revert the numeric validation, unwire both call sites, unwire only refresh. Each fails at least one test. `spawnSync` carries a 15s timeout so a runaway loop surfaces as a failed assertion rather than a hang.

**Not verified:** nothing here has run against a real locked stack. The retry logic is covered by fakes only.

### What this deliberately does not do

Neither commit touches `await-in-progress.js` — that was #21800, now merged, which stops three permanently-wedged `queued` runs from forcing every deploy to burn the turnstile's full 45-minute cap.

The structural fix — replacing the turnstile with a GitHub `concurrency:` group on both this workflow and `testing-build-and-deploy.yml` — remains out of scope. It's a larger change touching production deploys and deserves its own PR.

Also known and not addressed here: `pulumi up` is the publish step (it swings the live CloudFront origin to the bucket named in `origin-bucket-metadata.json`, and `infrastructure/index.ts` reads only `.bucket`, ignoring the `timestamp` and `commit` in the same file). An older run that retries after a newer one has published can therefore flip the origin backwards. That hazard exists today without this change, but the retry makes it reachable more often and silently. Being worked separately.

### Related issues

None filed; this came out of a #docs-ops investigation into the 2026-09-21 failures.

https://claude.ai/code/session_01Eif28XV3WrnCmpfdNK3HZK


CamSoper:
Answers to the two ⚠️ items on the reviewer's guide, with receipts, so nobody has to re-derive them.

## F1 -- `--retry-all-errors` is deliberate, but my stated reason for it was wrong

The PR description claimed `--retry-all-errors` is what makes curl retry a 504. That's false, and good catch for making me check it. A 504 is already in curl's transient-error set, so plain `--retry 3` would have caught the failure that motivated this PR.

Measured on curl 8.5.0 against a local server returning a fixed status, counting actual requests:

| invocation | HTTP 504 | HTTP 404 |
|---|---|---|
| `-fsSL` (what these steps had before this PR) | 1 request | not tested |
| `-fsSL --retry 3` | 4 requests | 1 request |
| `-fsSL --retry 3 --retry-all-errors` | 4 requests | 4 requests |

So the flag's real job here is the connection-level failures -- refused, reset -- which aren't transient by curl's definition and which plain `--retry` leaves uncovered. Those are plausible ways a release-CDN fetch dies, so I'm keeping it deliberately rather than by accident.

The cost you flagged is real and I'd rather state it than hand-wave it: a wrong version pin or a moved release URL now burns roughly 15 seconds across four attempts before failing, with an identical error message. That seemed a fair trade against losing a six-minute build to a connection reset. It also matches what `scripts/fetch-openapi-spec.sh`, `fetch-esc-schemas.sh` and `fetch-audit-log-events.sh` already do, so it isn't a new convention.

No code change. PR description corrected, since the wrong claim was the actual defect here.

## F2 -- `preview` left unwrapped is deliberate

`pulumi preview` doesn't take the update lock, so there's no 409 for it to hit. Direct evidence from the incident itself, run [35625043036](https://github.com/pulumi/docs/actions/runs/35625043036):

```
17:13:29  Previewing update (pulumi/www-production)
17:13:56  Resources: ~1 to update, +-1 to replace      <- preview completed fine
17:13:56  Updating (pulumi/www-production)
17:13:57  error: [409] Conflict: Another update is currently in progress.
```

The preview ran to completion while the stack was contended, and the 409 landed one second later on the update. `refresh` is a different story, which is why it is wrapped -- run [35625169056](https://github.com/pulumi/docs/actions/runs/35625169056) took its 409 on `Refreshing`, not `Updating`.

Your second reason holds independently: `preview` is only reachable through `scripts/ci-pull-request.sh`, which is the PR path, not the merge train that produced the collision.

---

Not sending these through `#update-review` -- the author card has no open items and nothing is blocking merge, so re-adjudicating would spend review budget to change nothing. Shout if you'd rather have it on the record there.

---

CamSoper:
@claude F1 and F2 both answered below -- full measurements and log excerpts are in https://github.com/pulumi/docs/pull/21789#issuecomment-5768469973. #update-review

**F1 -- accepting as-is, with a corrected rationale.** The finding is right that `--retry-all-errors` retries hard failures too. Keeping it anyway, but my original justification was wrong and the PR description has been fixed. Measured on curl 8.5.0 against a local fixed-status server: plain `--retry 3` already retries a 504 four times (it's in curl's transient-error set), so `--retry-all-errors` was never what made the 504 case work. What it actually buys is retrying connection-level failures -- refused, reset -- which are not transient by curl's definition and which a release-CDN fetch can plausibly hit. Cost is as you describe: a wrong pin or moved URL burns ~15s over four attempts with an identical error message. Accepting that trade, and matching the existing form in `scripts/fetch-openapi-spec.sh`, `fetch-esc-schemas.sh` and `fetch-audit-log-events.sh`.

**F2 -- disagreeing: `preview` doesn't take the update lock, so there is no 409 for it to hit.** From run 35625043036 during the incident: `Previewing update` started 17:13:29 and completed at 17:13:56 while the stack was contended; the 409 landed at 17:13:57 on `Updating`. `refresh` is genuinely different and is wrapped -- run 35625169056 took its 409 on `Refreshing`, not `Updating`. Your own second reason also holds independently: `preview` is only reachable via `scripts/ci-pull-request.sh`, the PR path, not the merge train that produced the collision. No change needed.

---

CamSoper:
@claude #update-review

CamSoper:
@claude F3 fixed in 91711f4 -- you were right, and I measured it before and after. #update-review

The test fed `PULUMI_LOCK_RETRY_DELAY="soon"` to validation I'd just added, which correctly falls back to the real 30s default, and then the retry paid it against the real `sleep`. Measured: **30.5s for this file, 0.386s after the fix** -- the full unit suite goes from ~31s to 0.585s. Your read of the harness comment was exact: "Keep the suite fast; the real defaults are 5 attempts / 30s" was a contract that test was breaking, and I added it without ever timing the suite.

Took your suggested fix -- `sleep() { :; }` in the `runSnippet` program. Nothing here tests timing (the backoff schedule is documented in `run-pulumi.sh` and asserted by reading it, not by waiting for it), so the stub costs no coverage and immunizes the whole harness rather than just this test. `calls === 2` and the "not a number" stderr assertion are unchanged, as you predicted.

One thing the fix introduced that I closed in the same commit, since it bears on whether the F3 regression test still has teeth: a bug in the attempt cap means an unbounded loop, and the real doubling delay was what kept such a loop *parked* rather than hot. With `sleep` stubbed it spins, so reverting the validation would have **hung** the suite instead of failing it. `spawnSync` now carries a 15-second timeout, so a runaway snippet is killed and surfaces as a failed assertion. Verified: with the validation reverted, the suite fails in 0.6s rather than hanging.

Re-verified non-vacuous against four mutations -- revert the anchored 409 match, revert the numeric validation, unwire both call sites, unwire only refresh. Each fails at least one test.

---

CamSoper:
@claude #new-review

CamSoper:
@claude #new-review

PR #21790, opened 2026-09-21
Add Pulumi HCL tabs to the Terraform Get Started track (MA-594 cluster A)

CamSoper:
### Proposed changes

The Terraform Get Started track is written for Terraform users, but until now only `first-look.md` carried an HCL tab. This adds one to the remaining seven language choosers in the track, so a reader who wants to keep writing HCL can follow the whole track without switching languages.

Cluster A of MA-594. Only the five files below are touched; other clusters are other PRs.

| File | Chooser | What the HCL tab says |
|---|---|---|
| `begin.md` | Choose your language | No extra language runtime -- the CLI fetches the HCL language plugin on first use (Pulumi CLI 3.256.0+) |
| `terraform-providers.md` | Example: Random Pet Names | A `runtime: hcl` project using `terraform { required_providers }`, plus the `pulumi/`-prefixed source for a native Pulumi provider |
| `terraform-modules.md` | AWS S3 bucket module | A `module` block invoking the registry module, run as a component resource |
| `terraform-modules.md` | Load a module at runtime | HCL has no SDK-generation step to opt out of, so the distinction doesn't arise -- plus the one case it genuinely doesn't cover |
| `reference-state.md` | Reference local state files | `terraform_remote_state` against the `local` backend |
| `reference-state.md` | Reference remote state | Same data source against the `remote` backend, with the token as a sensitive variable |
| `convert-hcl.md` | Converted TypeScript example | Prose, not code -- see below |

Every chooser's option list and its `choosable` bodies were checked to match exactly: 7 choosers, 7 new HCL bodies, no empty tabs.

Two things worth calling out:

**`convert-hcl.md` is a prose tab on purpose.** `pulumi convert --from terraform --language hcl` is rejected upstream, so there is no conversion to show. The tab says that plainly, gives the reason (re-homing every resource onto a different provider would show a delete and a create on the next preview), and points at `/docs/iac/languages-sdks/hcl/`. I verified the refusal rather than taking it on faith:

```
$ pulumi convert --from terraform --language hcl --out ./out
error: cannot convert a Terraform program to the "hcl" language: pulumi-hcl runs
Terraform directly, so no conversion is needed; converting would re-home every
resource onto a different provider and show a delete and create for each one on
the next preview
```

**One drive-by fix.** `terraform-providers.md`'s existing "Terraform equivalent" block ended with `random_shuffle.az.results`, while the resource reference two lines above correctly used `.result`. The provider only exposes `result` -- with `.results` the program fails to bind:

```
Unsupported attribute; This object does not have an attribute named "results".
```

Since my HCL tab sits right beside that block and a reader compares the two, I fixed the one line rather than shipping a matching pair that disagree.

### Verification

Ran against Pulumi CLI **3.263.0** (`curl -fsSL https://get.pulumi.com | sh`), `pulumi login --local`, scratch projects outside the repo.

**Plugin downloads worked.** This is the thing a previous session in a restricted environment could not get past, so to be specific: the HCL language plugin, the bridged and native providers (`random`, `aws`, `terraform`), and remote Terraform Registry modules all resolved and downloaded fine. `pulumi install` returned 0 in every project. One narrow exception, below.

|---|---|---|---|
| 1 | `begin.md` -- prerequisites bullet | Prose only; CLI version claim matches `languages-sdks/hcl/_index.md` | n/a |
| 2 | `terraform-providers.md` -- `required_providers` + `random_shuffle` + `aws_elb` | `pulumi install` + `pulumi preview` | **Pass** -- clean preview, 4 resources, outputs resolve |
| 3 | `terraform-providers.md` -- `pulumi/random` exact-version variant | `pulumi install` + `pulumi up` | **Pass** -- 2 created, real output values |
| 4 | `terraform-modules.md` -- registry S3 `module` block | `pulumi install` + `pulumi preview` | **Pass (partial)** -- module fetched, registered as `components:index:S3Bucket`, outputs bound; full preview blocked only by AWS credentials |
| 5 | `terraform-modules.md` -- runtime-load `module` block | Same | **Pass (partial)** -- same, same reason |
| 6 | `terraform-modules.md` -- `module` block semantics generally | Equivalent local module, `pulumi install` + `pulumi up` | **Pass** -- 3 created, component + child resource + output |
| 7 | `reference-state.md` -- `terraform_remote_state`, `local` backend | Real hand-written `terraform.tfstate`, `pulumi preview` | **Pass** -- 0 errors, both outputs resolve to the real values |
| 8 | `reference-state.md` -- `terraform_remote_state`, `remote` backend | `pulumi preview` with a placeholder token | **Pass (partial)** -- binds, maps to `terraform:state:getRemoteReference`, reaches app.terraform.io; fails only on `unauthorized` |
| 9 | `convert-hcl.md` -- prose tab | Reproduced the upstream refusal (quoted above) | **Pass** |

On the three **partial** rows: in each case the HCL parsed, bound, and planned, and the only failure was a live credential the sandbox doesn't have (AWS STS `GetCallerIdentity`, or an HCP Terraform token). That's the ceiling without a real cloud account, and it's not a defect in the examples. Two supporting negative tests confirm binding is real rather than deferred: a bogus module output name is rejected (`unknown node "module.s3_bucket.output.totally_bogus_output"`), and so is `.results` on `random_shuffle`.

Also verified by negative test, because it's a claim the runtime-load tab makes: `source = var.module_address` is rejected with `Variables not allowed; Variables may not be used here.` The tab says so rather than implying HCL can do the dynamic-address case.

**What I could not run:** `pulumi convert --from hcl --language typescript` as a bind-check. Converter-plugin resolution goes to `api.github.com`, which returns 403 through this environment's proxy. It didn't matter -- `pulumi preview` is the stronger check and it worked everywhere. Same 403 blocked `make ensure` at the theme's `@pulumi/design-tokens` dependency (a `codeload.github.com` tarball), so no full `make build` here; lint doesn't need it.


CamSoper:
@claude #update-review

All three items addressed in f742b36.

**F2 -- fixed.** Good catch, and the cause is worse than it looks: `reference-terraform-state` is the project name from the YAML tab of the *local state* section (L189). The remote-state section's own YAML tab uses `terraform-remote-state-with-yaml` (L397). So the HCL tab had borrowed a namespace from a different section's different-language tab. Dropped the prefix:

```bash
$ pulumi config set --secret tfe_token <your-token>
```

**F3 -- fixed.** Agreed the two sentences read as a contradiction side by side. Both were defensible on their own (`pulumi install` does write provider descriptors to `sdks/<provider>/hcl.sdk.json`, which is why `languages-sdks/hcl/_index.md` says "SDKs"), but on a page whose other tabs generate real typed SDKs via `pulumi package add`, "fetch the SDKs" invites exactly the wrong inference. Both spots now say "fetch the providers". Left the "no SDK-generation step" sentence alone since that one is the point of the section.

**F1 -- confirmed, and now linked.** The behavior is documented in two published Pulumi pages:

- `content/docs/iac/languages-sdks/hcl/_index.md` L68: "Providers resolve the same way as OpenTofu: an unqualified source such as `random` (or a fully-qualified `hashicorp/random`) is looked up in the OpenTofu registry and bridged into Pulumi automatically -- no `required_providers` entry is required."
- `content/docs/iac/languages-sdks/hcl/hcl-language-reference.md` L374, under Required providers: "an unqualified `aws` resolves to `registry.opentofu.org/hashicorp/aws` and is bridged into Pulumi automatically."

Upstream says the same thing in [pulumi-hcl `docs/providers.md`](https://github.com/pulumi/pulumi-hcl/blob/master/docs/providers.md), whose default-resolution example is commented `# Resolves to "registry.opentofu.org/hashicorp/aws"`.

Rather than just reply, I linked the sentence to the reference's Required providers section and mirrored its wording, so the claim carries its own source from now on.

`make lint` passes -- 1794 files, 0 errors, Prettier clean.

---

content/docs/iac/get-started/terraform/convert-hcl.md
```diff
@@ -1174,6 +1174,22 @@ outputs:
...
+
+[Pulumi HCL](/docs/iac/languages-sdks/hcl/) has nothing to convert: it runs the configuration at the top of this page as-is. Add a `Pulumi.yaml` next to your existing `.tf` files and leave the files alone.
+
+```yaml
+name: my-infrastructure
+runtime: hcl
+description: Existing Terraform configuration, run by Pulumi
+```
+
+`pulumi convert --from terraform --language hcl` refuses the job for this reason, and the refusal is worth understanding. Converting would rewrite each resource onto a different provider, and Pulumi would read the rewritten resources as new ones — so the next `pulumi preview` would show a delete and a create for every resource in your configuration. Keeping the HCL avoids that entirely: the same files, the same providers, the same resource addresses.
```
tatcoo-pulumi: nit, but "the refusal is worth understanding" is a Claudism, suggest dropping

tatcoo-pulumi (approved):
Minor comment, LGTM

content/docs/iac/get-started/terraform/convert-hcl.md
CamSoper: Good catch, dropped in d1bada8. It was telling the reader the next sentence matters instead of just letting it say the thing.

---

CamSoper:
@claude re-verify d1bada8 #update-review

The only change since the last review is a one-clause deletion in `content/docs/iac/get-started/terraform/convert-hcl.md` L1187, addressing a reviewer nit — "and the refusal is worth understanding" removed, nothing else on the line or in the file touched. No open findings were outstanding, so `auto-refresh` skipped and the card is pinned to the previous head.

---

CamSoper:
@claude #new-review

This PR was rebuilt onto the #21806 layout, so its diff no longer resembles the one the pinned review covers: the five `iac/get-started/terraform/*` files are gone, and the HCL content now lives in three `integrations/terraform/` pages. It also absorbed the module-guide tab from #21801. The PR description has the full file table and verification. A fresh review is more useful than updating findings anchored to files that no longer exist.

---

CamSoper:
@claude answers to F1–F4, all confirmed from primary sources and testing (CLI 3.263.0, language-hcl 0.18.0). No text changes needed. #update-review

**F1 + F2: why the converter refuses, and which provider resources would move to.** The primary source is the Pulumi CLI's own error, captured verbatim by running the command:

```
$ pulumi convert --from terraform --language hcl --out ../out
error: cannot convert a Terraform program to the "hcl" language: pulumi-hcl runs Terraform directly, so no conversion is needed; converting would re-home every resource onto a different provider and show a delete and create for each one on the next preview
```

That states both halves of the sentence: resources move to a different provider, and the next preview shows a delete and create for each. The mechanism, observed in previews:
- Running `.tf` files directly under `runtime: hcl` uses the **bridged** Terraform provider (unqualified `aws` → `registry.opentofu.org/hashicorp/aws` via `terraform-provider`), registering types like `aws:index:S3Bucket` / `aws:index:Vpc`.
- The Terraform converter targets Pulumi's **native** providers. `pulumi convert --from pcl --language hcl` over an equivalent program emits `source = "pulumi/aws"` and `aws_ec2_vpc`, i.e. type `aws:ec2/vpc:Vpc`.
- A different type token means a different URN. A type change with no alias previewed as create + delete in our testing (the alias test on `refactoring-with-aliases`, PR #21798), which is exactly what the CLI message warns about.

**F3: `hcl_module`'s `source` is an ordinary argument.** Upstream `pulumi-hcl`:
- `pkg/server/module_resource.go:52` — "`hcl:index:Module` resource: a component whose module source arrives as a plain input at Construct time, rather than being baked into a per-module schema."
- `pkg/server/module_resource.go:203` — `construct` reads it with `req.Inputs.GetOk("source")`.
- `docs/mlc.md:67` — "the dynamic `hcl:index:Module` resource (referenced as `hcl_module`)".

Tested: with `source = var.module_address` pointing at a local module, preview returned the module's output (`shout: "HELLO"`). By contrast, a native `module` block with `source = var.x` fails with "Variables not allowed". One nuance, consistent with the sentence as written: `source` must be a plain string by construct time (`module requires a plain string "source" input`), so a variable works, and an unknown resource output wouldn't. The docs only claim "a variable".

**F4: `terraform_remote_state` with `backend = "remote"`.** Upstream `pulumi-hcl` `pkg/hcl/run/terraform_remote_state.go`, `lowerRemoteStateInvoke`: `case "remote"` lowers to `terraform:state:getRemoteReference` (`packages.RemoteReferenceToken`) and maps exactly the `organization`, `hostname`, `token`, and `workspaces` fields. The token constant is in `pkg/hcl/packages/builtins.go:48`. Tested: the tab's program lowered to `terraform:state:getRemoteReference` and reached app.terraform.io, which returned "unauthorized" for the dummy token. The request went all the way to Terraform Cloud.

---

PR #21793, opened 2026-09-21
Add Pulumi HCL tabs to the stash concept page (MA-594 cluster C)

CamSoper:
### Proposed changes

Adds an `hcl` tab to all six language choosers on `content/docs/iac/concepts/stash.md`, using Terraform's built-in `terraform_data` resource -- which, per the [HCL language reference](https://www.pulumi.com/docs/iac/languages-sdks/hcl/hcl-language-reference/#terraform-built-in-resources), lowers onto the Pulumi engine's built-in stash resource.

**The one thing worth reviewing carefully:** a bare `terraform_data` does *not* reproduce the behavior this page documents. I found this while verifying, not while writing.

The page's premise is that `output` "persists the original `input` value even after the `input` property is changed." That holds for the native Stash resource -- I confirmed it against a YAML program. It does not hold for `terraform_data`, which keeps Terraform's semantics, where `output` is the computed value of `input` and follows it on every update. The engine stores the right thing (state shows `output` pinned at the original), but the HCL plugin surfaces `output` the Terraform way, so the program reads back the current value:

```
~ stashed_value: "Hello, World!" => "Goodbye, World!"
```

So every tab pairs `terraform_data` with `lifecycle { ignore_changes = [input] }`, which pins the created value and gives the stateful `output` the page describes. I verified the block is load-bearing with a negative control: drop it and `output` tracks `input` again.

That has a knock-on effect I documented in the first tab rather than papering over: because `input` is ignored, `terraform_data.<name>.input` reports the pinned value too, rather than echoing the most recent one. The page's intro advertises both output properties; an HCL program gets one of the two, not both.

Other per-tab notes:

- **Secrets** -- `sensitive()` on the input is enough. Secretness propagates to the stack output without `sensitive = true` on the `output` block, which matches what the other tabs' comment claims, so I left it off.
- **First deployment user** -- HCL has no function that reads the current user, so the tab takes a variable and shows `TF_VAR_deployer="$USER" pulumi up`.
- **Stable random value** -- `uuid()` stands in for the other tabs' fictional `generatePassword` helper; it genuinely produces a fresh value per run, so the example demonstrates the real thing.
- **Complex values** -- kept the `instanceType` map key from the sibling tabs and noted that map keys are not `snake_case`-converted (confirmed in output).

The documented `ignore_changes`-on-`triggers_replace` caveat does not apply here: these examples ignore changes to `input`, not `triggers_replace`. I did verify the two combine cleanly when using `triggers_replace` to rotate a stashed value, which is what the last tab points at.

### Verification

Pulumi CLI **v3.263.0**, local file backend. The HCL was extracted programmatically from the committed file into scratch projects (so what ran is exactly what ships), then deployed for real -- `terraform_data` needs no cloud credentials.

**Plugin downloads worked.** The HCL language plugin resolved and downloaded without issue; `pulumi install` returned 0 on every project. No `api.github.com` problems, and no `pulumi convert` was used.

| # | Example | Verified by | Result |
|---|---|---|---|
| 1 | Create a stash | `pulumi up`; then changed the literal and re-ran | **PASS** -- `stashed_value: "Hello, World!"`, still `"Hello, World!"` after input changed to `CHANGED` |
| 2 | Stashing complex values | `pulumi up` | **PASS** -- nested object round-tripped; `instanceType` key preserved verbatim |
| 3 | Stashing secret values | `pulumi up`; `stack output`, `--show-secrets`, state inspection | **PASS** -- `api_key: [secret]`, plaintext only with `--show-secrets`, `ciphertext` in state |
| 4 | Capturing the first deployment user | `pulumi up` as `alice`, re-ran as `bob` | **PASS** -- `original_deployer: "alice"` both times |
| 5 | Recording the initial creation time | `pulumi up`, slept, re-ran | **PASS** -- `first_deployed: "2026-09-21T22:35:35Z"` unchanged |
| 6 | Preserving a stable random value | `pulumi up`, re-ran, compared with `--show-secrets` | **PASS** -- same UUID both runs, masked as `[secret]` |

Supporting checks:

| Check | Result |
|---|---|
| Negative: bogus attribute on `terraform_data` | **PASS** -- rejected (`An argument named "bogus_attribute" is not expected here`), so binding is real |
| Negative control: remove `lifecycle` block | **PASS** -- `output` tracks `input` again, confirming the block is required |
| Control: native Stash in YAML, input changed | **PASS** -- `output` stays original, confirming the divergence is HCL-specific |
| `triggers_replace` rotation + `ignore_changes` | **PASS** -- replaces the stash and re-evaluates the input |
| `make lint` | **PASS** -- 1794 files parsed, 0 errors; Prettier clean |
| `make lint-prose` (Vale) | **PASS** -- 0 errors, 0 warnings, 0 suggestions |
| Hugo render (v0.157.0, pinned version) | **PASS** -- all 6 choosers carry `hcl`; all 6 `hcl` tabs render non-empty with `language-hcl` blocks and rendered prose, no empty gaps |

### Notes and anything left undone

- A **full site build was not run locally.** `make ensure` cannot fetch `@pulumi/design-tokens` in this environment (403), so the theme CSS is absent and `hugo` fails in the asset pipeline on an unrelated page. I worked around it for lint with `yarn install --ignore-scripts`, and validated the shortcodes by rendering this page in an isolated Hugo site with the repo's real `chooser`/`choosable` shortcodes at the pinned Hugo version. CI will cover the full build.
- **Scope:** only `stash.md` is touched. The prose-only "Updating stashed values" section has no chooser, so it is unchanged; the last HCL tab points at `triggers_replace` for the rotation case instead.
- Nothing was left unverified, and no example needed to be dropped.

https://claude.ai/code/session_01EXwQEz2q64T5qdxYymFxPQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01EXwQEz2q64T5qdxYymFxPQ)_


CamSoper:
@claude rebased onto master and pushed a second-pass review in 617f983 #update-review

- Rebased onto current master (clean; no content conflicts).
- New HCL paragraph under "Updating stashed values": the shared text names `replacementTrigger` and says a changed `input` shows up in the `input` output, and neither holds for an HCL reader. The HCL spelling is `triggers_replace`, and because every HCL tab ignores changes to `input`, both `input` and `output` stay pinned until replacement. Verified with real `pulumi up` runs on the local backend: changing `input` and re-running reported "2 unchanged" with both outputs still at the original value; setting `triggers_replace = "v2"` then `"v3"` replaced the stash each time.
- Three sentences rewritten as plain statements ("The `lifecycle` block is what makes this a stash", "One consequence: …", "… is what holds the stash …"). Meaning unchanged.

---

CamSoper:
@claude F2: yes, observed directly, both halves. Re-run just now on CLI 3.263.0 / language-hcl 0.18.0, local backend, with `terraform_data` + `lifecycle { ignore_changes = [input] }` and outputs `echo_input = terraform_data.my_stash.input`, `stashed = terraform_data.my_stash.output`:

1. First `pulumi up` with `input = "first"`: `echo_input: "first"`, `stashed: "first"`.
2. Changed the literal to `input = "second"`, second `pulumi up`: **`2 unchanged`**, `echo_input: "first"`, `stashed: "first"`.
3. `pulumi stack export` shows the resource (`pulumi:index:Stash`) carries exactly two attributes, `input` and `output`, both still `"first"`, so nothing on the resource echoes the new expression.

So "`terraform_data.my_stash.input` also returns the pinned value" and "no attribute of the resource echoes the current `input` expression" are both observed behavior, not inference. The sentence stays as written. #update-review

---

tatcoo-pulumi (approved):

PR #21794, opened 2026-09-21
Add Pulumi HCL tabs to Pulumi Cloud provider pages (MA-594 cluster I)

CamSoper:
### Proposed changes

Cluster I of MA-594: adds Pulumi HCL tabs to the pages that manage Pulumi Cloud resources with the `pulumiservice` provider. Seven of the nine audited chooser blocks got an `hcl` choosable; two didn't, for a reason described below.

#### Per-file summary

| File | Block | What was added |
| --- | --- | --- |
| `content/docs/deployments/concepts/drift.md` | Setting it up via the Pulumi Service Provider | `pulumiservice_drift_schedule` + `schedule_id` output |
| `content/docs/deployments/concepts/schedules.md` | Pulumi Cloud Service provider | `pulumiservice_deployment_schedule` + `schedule_id` output |
| `content/docs/deployments/concepts/ttl.md` | Pulumi Cloud Service provider | `pulumiservice_ttl_schedule` + `schedule_id` output |
| `content/docs/deployments/concepts/webhooks.md` | Create a Webhook | `pulumiservice_webhook` (organization form) |
| `content/docs/deployments/concepts/webhooks.md` | Headers | **no HCL tab** -- see below |
| `content/docs/deployments/guides/dependent-stack-updates.md` | Deployment webhook destinations | two `pulumiservice_webhook` resources with `format`/`filters` |
| `content/docs/deployments/guides/dependent-stack-updates.md` | Pulumi Auto Deploy | four `auto-deploy_auto_deployer` resources wired through `.ref` |
| `content/docs/esc/concepts/webhooks.md` | Create a Webhook in a Pulumi IaC Program | `pulumiservice_webhook` with `environment_name` |
| `content/docs/esc/concepts/webhooks.md` | Headers | **no HCL tab** -- see below |

Nothing outside these six files was touched.

#### The two blocks left alone

Both "Headers" choosers show how to compute and verify the `Pulumi-Webhook-Signature` HMAC **inside a webhook receiver** -- `crypto.createHmac`, `hmac.new`, `hmac.New(sha256.New, ...)`. That's general-purpose runtime code running in your own service, not a resource declaration. HCL has no HMAC primitive and no notion of an HTTP request handler, so there's nothing to write that wouldn't be invented syntax. Left them as `javascript,typescript,python,go`.

#### Token surface: `index:*`

I documented the `pulumiservice:index:*` resources rather than the newer `pulumiservice:api/*` ones. Both exist in the schema; neither carries a `deprecationMessage`. Two reasons for `index:*`: every sibling tab already on these pages uses it (the YAML tabs say `type: pulumiservice:index:DriftSchedule`), so a reader flipping between tabs sees the same resource, and a mixed surface within one chooser would read as an accident rather than a choice. Worth revisiting as a separate pass if/when `api/*` becomes the recommended surface -- that's a decision bigger than this cluster.

#### Notes on the HCL itself

- Resource names follow the `pulumiTokenToHCL` rule (drop `index`, snake_case): `pulumiservice:index:TtlSchedule` -> `pulumiservice_ttl_schedule`.
- `required_providers` keys equal the package name, and Pulumi-sourced packages take an **exact** semver. Versions used: `pulumiservice` **1.3.0**, `auto-deploy` **0.0.4** (both current at time of writing).
- Property and enum wire values (`format = "pulumi_deployments"`, `filters = ["update_succeeded"]`, `pulumi_operation = "update"`) come from `pulumi package get-schema`, not from memory.
- The Auto Deploy example is the odd one: the package name `auto-deploy` is hyphenated, so the hyphen carries into both the `required_providers` key and the resource type -- `auto-deploy_auto_deployer`. That looks wrong but is the only form that resolves; `auto_deploy_auto_deployer` fails with `resolving resource type ... not found`, and a `auto_deploy` key sends the resolver to `hashicorp/auto` on the OpenTofu registry. Verified both ways round, and called out in the tab's prose.

### Verification

Every example was extracted **from the committed markdown** (not retyped), scaffolded into a scratch project outside the repo, then `pulumi install` + `pulumi preview` against a `--local` backend. Pulumi CLI v3.263.0; HCL language plugin v0.18.0.

Plugin downloads worked. `pulumi install` returned 0 for every example, for both the native `pulumiservice` provider and the `auto-deploy` component package.

| Example | `pulumi install` | Bound? | Stopped at |
| --- | --- | --- | --- |
| drift.md -- DriftSchedule | 0 | yes | access token |
| schedules.md -- DeploymentSchedule | 0 | yes | access token |
| ttl.md -- TtlSchedule | 0 | yes | access token |
| webhooks.md -- Webhook | 0 | yes | access token |
| dependent-stack-updates.md -- webhook destinations | 0 | yes | access token |
| dependent-stack-updates.md -- Auto Deploy | 0 | yes | access token |
| esc/webhooks.md -- Webhook | 0 | yes | access token |

"Access token" is the honest ceiling: there's no Pulumi access token in this environment, so the provider can't do anything real. Every example got as far as `pulumi access token not found`, which is past binding.

**Negative tests.** Binding is the only thing provable here, so it needed proving rather than assuming:

- **Bogus resource type** -- `pulumiservice_wibblefish` is rejected with `unknown resource type`, never reaching the auth boundary.
- **Misspelled property** -- `display_nmae` is rejected with `Missing required argument: display_name`.
- **Per-example mutation** -- each of the seven programs above was copied, its first property renamed to `zz_bogus_prop`, and re-previewed. All seven failed with a schema error (`Missing required argument`) *instead of* the access-token error. So the clean runs really are binding, per example, not short-circuiting somewhere earlier.

Also verified empirically, since the Auto Deploy tab asserts it: `pulumi config set organization <org>` does feed an HCL `variable "organization"`.

Per the task note, `pulumi convert --from hcl` was not used as a bind check.

**Lint:** `make lint` passes (1794 files, 0 errors; prettier clean). `make lint-prose` (Vale, advisory) reports 54 findings across the six files -- **0 of them on lines this PR adds**; cross-referenced findings against the diff's added line ranges to confirm. The pre-existing ones are heading/title sentence-case, "simply", and future tense, left alone as out of scope. `yarn install` succeeded; `make ensure` wasn't needed for lint.

N/A.

### Related issues (optional)

MA-594 (cluster I).

#### Left undone

- The two "Headers" blocks, as described above.
- Provider versions are pinned literally in the docs (`1.3.0`, `0.0.4`) because Pulumi-sourced packages don't accept constraints. These will age; worth knowing if the repo grows a version-freshness check.

https://claude.ai/code/session_01Tpnbg1hBp64facWRG2vP8T

---
_Generated by [Claude Code](https://claude.ai/code/session_01Tpnbg1hBp64facWRG2vP8T)_


CamSoper:
@claude rebased onto master and pushed a second-pass review (eff22f7, d432de8) #update-review

- Rebased onto current master (clean).
- **`esc/concepts/webhooks.md`: the environment-webhook sample was broken in every tab, not just HCL.** It omitted `projectName`, and a preview against pulumiservice fails the provider's own check: `projectName and environmentName must both be specified for environment webhooks`. Added `projectName`/`project_name` to all five tabs. The "just omit `environmentName`" guidance for an organization webhook was wrong too (the provider rejects a project name with no environment or stack: `projectName needs to be empty if this is meant to be an organization webhook`); it now says to omit both.
- **`dependent-stack-updates.md` Auto Deploy tab:** replaced a required `variable "organization"` (and its `pulumi config set` instruction) with `pulumi.organization`, matching the sibling tabs' `pulumi.getOrganization()` / `${pulumi.organization}`. Preview registered all four `AutoDeployer` components and their child webhooks.
- **Dropped the seven `version` pins** on `pulumi/` sources. A `pulumi/` source takes an exact version *if* one is given, but none is required; without one, `pulumi install` resolves the latest release. The docs convention is unpinned (13 of 17 `required_providers` blocks in content/docs omit `version`), the sibling tabs don't pin, and a pin rots silently. The explanatory note is now one identical sentence on every HCL tab that declares a `pulumi/` source.

All resource types and arguments were checked against the pulumiservice schema, with a negative control (`auto_remediatex` → "Unsupported argument… Did you mean auto_remediate") proving validation ran.

---

tatcoo-pulumi (approved):

PR #21795, opened 2026-09-21
Add Pulumi HCL tabs to guides basics (MA-594 cluster G)

CamSoper:
### Proposed changes

Cluster G of MA-594: adds Pulumi HCL to the language choosers on the two "basics" guides. Eight choosers, eight new `{{% choosable language hcl %}}` bodies -- no tab was added without its body.

#### `content/docs/iac/guides/basics/how-pulumi-works.md`

Four choosers: the three program snapshots (two buckets → add tags → rename a bucket) and the architecture diagram.

The page is conceptual, so the HCL tabs stay close to what the other languages show but say what's actually true of HCL rather than paraphrasing the TypeScript answer:

- Resource labels are identifiers, so the logical names are `media_bucket` / `content_bucket`, not the hyphenated names the walkthrough prose uses. The HCL tab says so instead of quietly disagreeing with the `pulumi up` output block.
- The rename tab notes that a `moved` block renames without replacing -- the HCL-specific answer to what that section is about.
- The architecture diagram gets an HCL language host running `main.tf`.

**One thing worth a reviewer's eye:** I forked the "Next, we run `pulumi up`" paragraph per language. That paragraph leans on the inline `pulumi-language`, `pulumi-host`, `pulumi-bucket`, and `pulumi-new-bucket` shortcodes, and none of them lists `hcl`. I checked the chooser component -- when the global selection isn't among a chooser's options, `preferredOrDefault` falls back to *displaying the first option* rather than rendering empty. So without the fork, a reader who picked HCL would have been told "this program is written in **JavaScript**... launches the **Node** language host." Forking the paragraph inside the content file fixes that without touching the shared shortcodes, and lets the HCL copy describe the model correctly: HCL has no object construction, so there's no constructor call to return from -- `resource` blocks become registrations as the configuration is evaluated.

Those four shortcodes are consumed *only* by this page, so adding `hcl` entries to them is also a viable fix if a reviewer prefers it -- I kept to the two in-scope files. Note the same fallback already mis-serves Java and YAML readers via `pulumi-bucket` / `pulumi-new-bucket`, which list neither; that's pre-existing and untouched here.

#### `content/docs/iac/guides/basics/organizing-projects-stacks/_index.md`

Four choosers: stack references, ESC, repo alignment, and in-project code organization.

- **Stack references** use the `pulumi_stack_reference` resource, with `${var.org}/networking/${pulumi.stack}` and outputs read by their exported map keys (`outputs["vpcId"]`, not `vpc_id` -- map keys aren't translated).
- **ESC** arrives as ordinary config, so the HCL tab reads it through `variable` blocks named to match the `pulumiConfig` keys.
- **Repo layout / code organization** follow HCL's real model rather than the TypeScript one: a project is a directory of `.tf` files merged into a single root module, there's no entrypoint and no imports, and the unit of reuse is a `module` block, not a package. The layout also lists `sdks/`, the provider descriptors `pulumi install` writes and that the HCL docs say to commit.

### Verification

Pulumi CLI **3.263.0** (`curl -fsSL https://get.pulumi.com | sh`), `pulumi login --local`. **Plugin downloads all worked** -- the HCL language plugin, the bridged `terraform-provider` (random, aws), and `terraform_data`'s built-in stash all resolved; `pulumi install` returned 0 every time. Per the task note I did *not* use `pulumi convert --from hcl` as a bind check.

|---|---------|--------------|--------|
| 1 | how-pulumi-works — two buckets (`aws_s3_bucket` ×2, empty `{}` bodies) | `pulumi preview`: parses and binds, reaches AWS credential check. Empty-body `{}` form + label→logical-name behavior confirmed by a real `pulumi up` with `random_pet` stand-ins | ✅ pass (see note A) |
| 2 | how-pulumi-works — `Pulumi.yaml` with `runtime: hcl` | Used as the manifest for every scratch project below | ✅ pass |
| 3 | how-pulumi-works — add `tags` to `media_bucket` | `pulumi preview` reaches credential check; `tags` confirmed present on `aws:index/s3Bucket:S3Bucket` in the provider schema (`pulumi package get-schema terraform-provider hashicorp/aws 6.66.0`), and `taggz` confirmed absent as a negative control | ✅ pass (see note A) |
| 4 | how-pulumi-works — rename `content_bucket` → `app_bucket` | Real `pulumi up` then `pulumi preview` with `random_pet`: yields `+ 1 to create / - 1 to delete`, matching the narrative | ✅ pass |
| 5 | how-pulumi-works — `moved` block note | Real `pulumi preview` with the `moved` block in place: `3 unchanged`, no create/delete | ✅ pass |
| 6 | how-pulumi-works — architecture diagram | Mermaid, no HCL to execute | n/a — prose/diagram |
| 7 | how-pulumi-works — forked "Next, we run `pulumi up`" paragraph | Prose | n/a — prose |
| 8 | organizing — `pulumi_stack_reference` | Real end-to-end `pulumi up`: a `networking/prod` producer exporting `vpcId` + `privateSubnetIds`, then a `clusters/prod` consumer resolving `${var.org}/networking/${pulumi.stack}` and reading both back correctly | ✅ pass |
| 9 | organizing — ESC config via `variable` blocks | Real `pulumi up`: `pulumi config set` for a camelCase scalar and a `list(string)`, both bound and read (`subnet_count: 2`) | ✅ pass |
| 10 | organizing — multi-repo directory layout | Directory tree; `sdks/<provider>/hcl.sdk.json` confirmed created by `pulumi install` | ✅ pass |
| 11 | organizing — multi-file project (`networking.tf` / `clusters.tf` / `workloads.tf`) | Real `pulumi up`: all three files merged into one root module, cross-file references resolved | ✅ pass |
| 12 | organizing — local `module` block (`source = "./modules/vpc"`, `cidr_block`) | Real `pulumi up` of the snippet exactly as written: instantiated as a component resource | ✅ pass |

**Negative tests** (confirming the positives mean something): a bogus `bogus_attribute` on `random_pet` and a bogus `not_a_thing` on `pulumi_stack_reference` were both rejected with `Unsupported argument`. On the schema side, `taggz` is absent from `aws_s3_bucket` where `tags` is present.

**Note A — the one limitation.** There are no AWS credentials in this environment, so examples 1 and 3 could not be driven to a real `pulumi up`. They are verified two ways instead: `pulumi preview` gets all the way past parse and bind to the STS credential check (so the HCL and the resource type are good), and attribute validity is confirmed directly against the downloaded provider schema. Attribute-level validation for AWS resources happens *after* provider configuration, so a credential-free preview can't catch a bad attribute -- hence the schema check rather than claiming more than I verified.

### Linting

- `make lint` — **passes** (0 errors, Prettier clean).
- `make lint-prose` (Vale, advisory) — ran with a locally fetched Vale 3.9.1. Three findings landed on lines I added; I fixed the one that was mine (`Pulumi.DirectionalReferences`, "as described below"). The other two — `Pulumi.NarrativeWe` ("Next, we") and `Pulumi.Difficulty` ("it just means") — are inherited verbatim from the sibling paragraph the HCL copy mirrors. The whole page is written in that voice and Vale flags the original too; making only the HCL tab read differently would be worse than the nag. Left as-is deliberately.
- `make ensure` was not used (`@pulumi/design-tokens` 403s here, as expected); `yarn install --ignore-scripts` was enough for lint.

Hugo isn't installed in this environment, so I couldn't do a render pass. Instead I wrote a structural check over both files asserting every chooser's option list has a matching `choosable` body and that open/close tags balance — **all 8 choosers OK, tags balanced.** Worth a visual check on a preview deploy anyway, particularly the forked paragraph.

### Nothing left undone

All 8 chooser blocks in scope are done. No HCL construct had to be skipped or invented. Only the two files in the slice are touched.

n/a — Pulumi HCL is GA (requires Pulumi CLI 3.256.0+).

### Related issues (optional)

MA-594 (cluster G).

https://claude.ai/code/session_018QHQA9xzpF8Rqy9EhRcnoZ

---
_Generated by [Claude Code](https://claude.ai/code/session_018QHQA9xzpF8Rqy9EhRcnoZ)_


CamSoper:
@claude rebased onto master and pushed a second-pass review in d602ab6 #update-review

Changes since the last review, all verified against CLI 3.263.0 / language-hcl 0.18.0 and `pulumi-hcl` source:

- **`how-pulumi-works.md` naming note:** it said labels are snake_case because "HCL resource labels are identifiers." That isn't a constraint — `resource "aws_s3_bucket" "media-bucket"` previews fine with logical name `media-bucket`. Reworded: snake_case is the HCL convention, and the label becomes the logical name. Also notes that the unqualified `aws` provider is bridged (`hashicorp/aws` 6.66.0), so `pulumi up` shows type `aws:index:S3Bucket` rather than the shared output's `aws:s3/bucket:Bucket`.
- **HCL execution paragraph:** replaced "there is no constructor call to return from" with what the host actually does — reads every `.tf` file, builds a dependency graph, and walks it (`pkg/hcl/run/run.go`, `Engine.Run` → `g.Walk`). Preview registered `content_bucket` before `media_bucket`, confirming there is no top-to-bottom order.
- **Shared prose false for HCL/YAML:** "the call to aws.s3.Bucket causes a new resource registration" → "the media bucket's definition causes…".
- **Language hosts section:** "This binary is distributed with the Pulumi CLI" is false for HCL — the CLI bundle has no HCL executor; it downloads `pulumi-language-hcl` on first run. Also added that Pulumi YAML and Pulumi HCL have no separate language SDK.
- **Resource providers section:** added that Pulumi HCL has no per-provider SDK package; `pulumi install` downloads the plugin and writes `sdks/<provider>/hcl.sdk.json` (confirmed on disk).
- **`organizing-projects-stacks/_index.md`:** rewrapped the added HCL paragraphs to the file's 120-column convention and tightened two sentences.

End-to-end checks: the stack-reference sample ran a real `up` on a file backend (a `networking/prod` HCL stack exporting `vpcId`/`privateSubnetIds`, read back by the `clusters` sample); renaming a label with a `moved` block gave "2 unchanged", without it create + delete.

---

tatcoo-pulumi (approved):

PR #21796, opened 2026-09-21
Add Pulumi HCL tabs to ESC guides and what-is pages (MA-594 cluster H)

CamSoper:
### Proposed changes

Cluster H of MA-594: adds a Pulumi HCL choosable to **7 chooser blocks across 6 files** -- two ESC guides and four `what-is` pages. Every chooser that gained `hcl` in its option list got its matching `{{% choosable language hcl %}}` body in the same edit, so there are no empty tabs. The four `what-is` choosers are page-scoped self-closing (`/ >}}`); each of those pages has exactly one `choosable` group, which I checked across the whole file rather than just the named section.

No prose was added or changed -- the diff is entirely chooser option strings and fenced `hcl` code blocks.

**Per file:**

| File | Chooser | What the HCL tab shows |
|---|---|---|
| `content/docs/esc/guides/pulumi-iac/adopt-esc-for-config-and-secrets.md` | The starting point | `variable` blocks for the plaintext value and the secret. ESC feeds stack config, which is the highest-priority source HCL variables read from, and `sensitive = true` is what makes the value a Pulumi secret -- the HCL analogue of `requireSecret`. |
| `content/docs/esc/guides/pulumi-iac/migrate-from-stack-references.md` | The starting point | `pulumi_stack_reference` plus `locals` reading `outputs["vpcId"]` -- the before-state the page migrates away from. |
| `content/docs/esc/guides/pulumi-iac/migrate-from-stack-references.md` | Update the consuming program | Typed `variable` blocks (`string`, `list(string)`) replacing the stack reference. |
| `content/what-is/is-kubernetes-infrastructure-as-code.md` | Can you manage Kubernetes with Pulumi | The page's three numbered steps: `eks_cluster`, a Kubernetes provider pointed at `kubeconfig_json`, and a Helm v4 chart. |
| `content/what-is/what-is-a-cloudflare-secret.md` | Conclusion | `cloudflare_workers_script` with a `secret_text` binding, reading account ID and secret from config. |
| `content/what-is/what-is-aws-secrets-manager.md` | Conclusion | `aws_secretsmanager_secret` plus `aws_secretsmanager_secret_version`. |
| `content/what-is/what-is-infrastructure-as-code.md` | How does infrastructure as code work? | A single `aws_s3_bucket`, matching the one-resource TS/Python tabs. |

### Verification

Pulumi CLI **3.263.0** installed from `get.pulumi.com`. Each example was extracted **from the committed markdown** (not from my drafts) into a scratch project outside the repo -- `Pulumi.yaml` with `runtime: hcl` plus the `.tf` file -- then `pulumi login --local`, `pulumi install`, `pulumi preview`.

Every row also got a **negative test**: a bogus attribute added to the resource, confirming the program is really binding against the provider schema rather than just parsing. All of them were rejected at registration (`Unsupported argument; An argument named "..." is not expected here`) -- one even suggested the correct spelling, which only happens with a loaded schema.

| Example | Result | How verified |
|---|---|---|
| ESC config `variable` blocks | **pass** | Clean `preview` + `up`. Config set via `pulumi config set <project>:<var>` resolved into both variables; `sensitive = true` rendered the output as `[secret]`. |
| `pulumi_stack_reference` | **pass** | Clean `preview` + `up` against a real local producer stack. `outputs["vpcId"]` / `outputs["publicSubnetIds"]` resolved to the exact exported values. Negative test rejected. |
| Post-migration `variable` blocks | **pass** | Clean `preview` + `up`. `list(string)` decoded correctly from `pulumi config set --path`. |
| Cloudflare `workers_script` | **pass** | Clean `preview` -- resolved to `cloudflare:index:WorkersScript`. Cloudflare's provider doesn't validate credentials at plan time, so this one previews green end to end. Negative test rejected (both at resource level and inside the `bindings` block). |
| S3 bucket | **pass (partial)** | Binds fully; `preview` reaches a live STS `GetCallerIdentity` call and fails only on `InvalidClientTokenId`. No AWS credentials here -- that's the ceiling, not a defect. Negative test rejected. |
| AWS Secrets Manager (both resources) | **pass (partial)** | Same: both resource types bind (verified in isolation so the STS failure couldn't mask the second one), `preview` stops at credentials. Negative test rejected. |
| EKS + Kubernetes + Helm v4 | **pass (partial)** | All three bind; `preview` stops at `Invalid credentials configured` for AWS. Installed the AWS CLI so the EKS component got past its prerequisite check. `kubeconfig_json` confirmed against the package schema (`pulumi package get-schema eks@4.3.0` lists `kubeconfigJson`), since credentials block runtime output evaluation. Negative test rejected. |

**Plugin downloads:** all worked. The HCL language plugin, the bridged `terraform-provider` (AWS, Cloudflare), and the native `pulumi/eks` + `pulumi/kubernetes` packages all downloaded and `pulumi install` returned 0. One caveat: **version *resolution*** goes through `api.github.com`, which 403s in this environment, so `pulumi/eks` and `pulumi/kubernetes` are pinned to exact versions in the Kubernetes example. That's not just a workaround -- the HCL reference states Pulumi-sourced packages require an exact version rather than a constraint, so the pin is correct either way. Pinned CDN downloads work fine.

**Two schema findings worth flagging for a reviewer:**

- The Helm v4 chart's HCL type name is `kubernetes_helm.sh_v4_chart` -- with a literal dot, derived from the `kubernetes:helm.sh/v4:Chart` token. Looks odd; it's what the plugin actually accepts, and `kubernetes_helm_v4_chart` errors with a "did you mean" pointing at it.
- On the Cloudflare page, the current bridged provider exposes `cloudflare_workers_script` (plural). `cloudflare_worker_script` no longer exists. The page's existing TS/Python/Go tabs still use the older `WorkerScript` / `secretTextBindings` API, so **the HCL tab is correct but no longer matches its neighbours**. Bringing those three tabs up to date is out of scope for this cluster -- flagging rather than silently widening the PR.

### Checks

- `make lint` -- **passes** (1794 files, 0 errors; prettier clean).
- `make lint-prose` (Vale 3.9.1) -- 0 errors, 6 warnings, **all on pre-existing prose** (`just` at two untouched lines, `accessor` x4 in a paragraph I didn't change). Left alone per convention. The diff adds no prose for Vale to flag.
- Full `hugo --minify` build -- **could not run.** It fails in the theme asset pipeline (`fingerprint: <nil> can not be transformed`) because `make ensure` can't fetch `@pulumi/design-tokens` here (the same 403). Unrelated to this diff. Instead I confirmed `hcl` is a first-class `LanguageKey` in `theme/stencil/src/components/chooser/chooser.tsx` with label `"HCL"` and a logo, and the pattern matches the ~20 existing pages under `content/docs/iac/concepts/resources/options/`.

### Nothing left undone

All 6 files, all 7 chooser blocks. No constructs had to be skipped for lack of working syntax.

### Related issues

MA-594 (cluster H)

https://claude.ai/code/session_014uqRHhXdBCck7uUnkePcqE

---
_Generated by [Claude Code](https://claude.ai/code/session_014uqRHhXdBCck7uUnkePcqE)_


CamSoper:
@claude #update-review

**F1: disputing.** `kubernetes_helm.sh_v4_chart` is the correct type name, and the suggested underscore form would break it.

I verified this by running pulumi-hcl's own converter rather than reasoning about it. `PulumiResourceTokenToHCL` on the real tokens:

```
kubernetes:helm.sh/v4:Chart     -> kubernetes_helm.sh_v4_chart
kubernetes:helm.sh/v3:Release   -> kubernetes_helm.sh_v3_release
kubernetes:apps/v1:Deployment   -> kubernetes_apps_v1_deployment
kubernetes:core/v1:Service      -> kubernetes_core_v1_service
```

`camelToSnake` maps `/` to `_` and lowercases, but has no dot handling, so the dot in the `helm.sh/v4` API group survives by design. That is why `apps/v1` and `core/v1` come out clean while `helm.sh/v4` does not -- the comparison to those two examples is the thing that doesn't hold, since neither has a dot in its group.

The underscore form also fails resolution, not just naming. `resourceLookupKey` strips underscores, and `tokenSearchKey` strips `/` and `_` but not `.`:

```
kubernetes_helm.sh_v4_chart  -> "helm.shv4chart"   matches the schema side
kubernetes_helm_sh_v4_chart  -> "helmshv4chart"    no match
```

**The real constraint is narrower than the finding states.** The dot doesn't make the type unusable, it makes the resource unusable in a *dotted traversal* -- you can declare it, but you can't write `kubernetes_helm.sh_v4_chart.nginx_ingress.foo`. This sample never does; it wires ordering with `depends_on`. I added a comment saying so in 8e64b5c, since a reader who tries to reference the chart would otherwise hit it cold.

**F2: partially fixed, and one part is yours to call.**

Fixed the part that was unambiguous: the tab declared `variable "secrets-demo"`, whose hyphen can't be used with `var.` dot access and forced the `var["secrets-demo"]` index form you flagged. Renamed to `secrets_demo` (and `accountId` to `account_id`), so it now uses plain dot access like every other HCL example in the docs. That removes the indexing question rather than answering it.

The provider mismatch is a real inconsistency and I'm leaving it for a human: the HCL tab declares no `required_providers`, so it resolves to the bridged Terraform `cloudflare` provider, while the TypeScript, Python, and Go tabs use the Pulumi provider (the Go tab pins `sdk/v5`). That's `cloudflare_workers_script` + `script_name` + `bindings` against `WorkerScript` + `name` + `secretTextBindings`. Aligning them means picking `pulumi/cloudflare` and pinning a version, which I can't verify from here -- provider plugin resolution goes through `api.github.com`, which 403s in this environment.

---

CamSoper:
@claude rebased onto master and pushed a second-pass review (19006b9, 485589b) #update-review

- **`what-is-a-cloudflare-secret.md`: the config command no longer set the HCL variable.** The previous round renamed the HCL variable to `secrets_demo`, but the page tells readers to run `pulumi config set secrets-demo --secret`. After setting `secrets-demo`, preview fails with `variable "secrets_demo" is required but no value was provided`. Added an HCL-specific `pulumi config set secrets_demo --secret` inside the HCL tab.
- **`is-kubernetes-infrastructure-as-code.md` helm comment gave advice that doesn't work.** It said to refer to the chart "with `depends_on` rather than a dotted traversal", but `depends_on` is itself a traversal: `depends_on = [kubernetes_helm.sh_v4_chart.nginx_ingress]` fails with `unknown node "kubernetes_helm.sh_v4_chart"` (`graph.TraversalKey` keys references on their first two segments). The comment now says other blocks can't refer to it.
- **Same page, prose contradicting its own HCL tab:** "Pulumi programs are TypeScript, Python, Go, .NET, or Java" and "Pulumi programs are written in real languages (…)" are both scoped to general-purpose languages now; the Kubernetes/Helm two-apply limitation is attributed to Terraform, and the page notes teams can write the same Pulumi program in HCL and still deploy in one `pulumi up`.
- **ESC guides:** noted that HCL variable names match config keys without the `myapp:` prefix (so the camelCase is deliberate), and added the HCL equivalent of "use the accessor that matches…" (the variable's `type` does the accessor's job; secret config stays secret without `sensitive`). Verified `lookupConfig` in `pulumi-hcl` `run.go` and with a real `pulumi up`.

---

tatcoo-pulumi (approved):

PR #21797, opened 2026-09-21
Add Pulumi HCL tabs to core concepts: stacks, config, secrets, resource names (MA-594 cluster B)

CamSoper:
### Proposed changes

Pulumi HCL is GA, but the four highest-traffic concept pages had no HCL tab. This adds one to **all 22 language choosers** across stacks, config, secrets, and resource names, so a reader who wants to keep writing HCL can follow the core concepts without switching languages. Every construct on these pages maps onto an HCL block, so no chooser needed a "not supported" punt except where the mapping genuinely doesn't exist (noted below).

#### `content/docs/iac/concepts/stacks.md` — 9 choosers

| Section | HCL tab |
| - | - |
| Stack outputs | `output` blocks |
| Output JSON serialization | `output "x"` / `output "o"` producing the same documented `pulumi stack output` results |
| Getting the current stack | `pulumi.stack`, plus `pulumi.project` / `pulumi.organization` |
| Stack references | the `pulumi_stack_reference` resource and its `outputs` map |
| Exporting a cluster's kubeconfig | `output` from a module |
| Consuming a stack reference across environments | stack ref → native `pulumi/kubernetes` provider → namespace |
| `requireOutput` | indexing `outputs` — a missing name is an `Invalid index` error at deployment time |
| `getOutput` + `apply` | `try` / `lookup` with a fallback, then ordinary interpolation to derive a value |
| `getOutputDetails` | note tab, like YAML's — HCL values aren't `Output`-wrapped, so there's nothing to unwrap |

The cross-environment example uses the `pulumi/`-prefixed native Kubernetes provider, and the tab says why: the bridged `hashicorp/kubernetes` provider takes `config_path` / `host` rather than `kubeconfig`. I found that the hard way -- the bridged provider rejects `kubeconfig` outright.

The **Stack README** chooser is left alone per the task's JUDGMENT marker.

#### `content/docs/iac/concepts/config.md` — 5 choosers

`variable` blocks are the config surface. The tabs cover the basic declaration (required vs. `default` vs. `sensitive`), object type constraints for structured config, and nested attribute/index access. Two caveats I verified rather than assumed:

- **Variables only read the project's own namespace.** `aws:region` is *not* reachable as `var.region` — I confirmed it resolves to the `default` instead. Provider settings go in a `provider` block, optionally fed by a project variable.
- **Only the root module reads stack config.** For the component-library-scoped chooser, a module declares its own `variable` blocks and the caller supplies them; leaving one unset is an error even when `mylib:name` is set.

#### `content/docs/iac/concepts/secrets/_index.md` — 4 choosers

`sensitive = true` on a `variable` or `output`, plus the `sensitive()` / `issensitive()` functions, and the SSM `SecureString` example fed from a sensitive variable.

#### `content/docs/iac/concepts/resources/names.md` — 4 choosers

The resource's second label is its logical name; the nested `pulumi { name = "..." }` block overrides it (useful since an HCL label can't contain a hyphen). Physical names are an ordinary resource argument, distinct from the `pulumi` block's `name`. One note worth flagging for review: **Pulumi HCL needs no `deleteBeforeReplace`** — it keeps Terraform's delete-first replacement ordering by default, and `create_before_destroy = true` is what opts into Pulumi's create-first order. I verified this against a real replacement rather than reading it off the mapping table.

### Verification

All examples were run against **Pulumi CLI 3.263.0** in scratch projects outside the repo (`Pulumi.yaml` with `runtime: hcl` + `.tf` files, `pulumi login --local`, `pulumi install`, `pulumi preview`). `pulumi convert --from hcl` was not used as a bind check, per the task note about `api.github.com`.

AWS-backed examples normally stop at live STS credentials in this environment. Setting `skip_credentials_validation` / `skip_requesting_account_id` / `skip_metadata_api_check` on a scratch-only provider block gets past that and exercises **real schema binding**, so these are full passes rather than partials. Those skip flags are scaffolding only -- they are not in any doc snippet.

| # | File / example | How verified | Result |
| - | - | - | - |
| 1 | stacks: `output "url"` (`aws_instance.web.public_dns`) | preview, AWS schema bound | pass |
| 2 | stacks: `output "x"` / `output "o"` | preview; outputs `"hello"` and `{num: 42}` — matches documented CLI output | pass |
| 3 | stacks: `locals { stack = pulumi.stack }` | preview; resolved to `dev` | pass |
| 4 | stacks: `pulumi_stack_reference` + `outputs["x"]` | `up` on a producer stack, then preview on a consumer | pass |
| 5 | stacks: `output "kubeConfig"` from `module.cluster` | preview with a local module | pass |
| 6 | stacks: stack ref → `pulumi/kubernetes` → `kubernetes_namespace` | preview; namespace planned, no cloud creds needed | pass |
| 7 | stacks: `requireOutput` via `outputs["vpcId"]` | preview | pass |
| 8 | stacks: `try(...)` + interpolation + `aws_s3_object` | preview; produced `logs/10.0.1.7.log` | pass |
| 9 | stacks: `getOutputDetails` note tab | prose only, no code | n/a |
| 10 | config: `variable` name / lucky / secret | preview; `Joe`, `42`, `[secret]` | pass |
| 11 | config: `provider "aws" { region = var.region }` | preview | pass |
| 12 | config: module with its own `variable` block | preview, plus negative test below | pass |
| 13 | config: `object({active, nums})` from `--path` config | preview; `active: true`, `nums[0]: 1` | pass |
| 14 | config: nested `api` object + `headers["authorization"]` | preview; endpoint / `30` / `Bearer token123` all correct | pass |
| 15 | secrets: `sensitive = true` / `sensitive()` prose list | preview of `sensitive()` + `issensitive()` → `true` | pass |
| 16 | secrets: `aws_ssm_parameter` SecureString | preview; SsmParameter planned | pass |
| 17 | secrets: `dbPassword` variable + sensitive output | preview; masked as `[secret]` | pass |
| 18 | secrets: `name` + `dbPassword` variables | preview; `BroomeLLC` plain, password masked | pass |
| 19 | names: logical name from label + `pulumi { name }` | `up`; `pulumiresourcename` → `my-role`, URN carries the override | pass |
| 20 | names: `name = "my-role-001"` | preview, AWS schema bound | pass |
| 21 | names: `name = "my-role-${pulumi.project}-${pulumi.stack}"` | preview; rendered `my-role-verify2-dev` | pass |
| 22 | names: `aws_s3_bucket` + `aws_s3_object` via `.id` | preview | pass |

**Negative tests** (confirming binding really happened, not silent acceptance):

| Test | Result |
| - | - |
| Bogus attribute on `random_pet` | rejected: `Unsupported attribute` |
| Bogus attribute on `aws_s3_bucket` | rejected: `Unsupported attribute` |
| Missing required `assume_role_policy` on `aws_iam_role` | rejected: `Missing required argument` |
| Missing key in a stack reference's `outputs` | rejected: `Invalid index` — this is what backs the `requireOutput` claim |
| Bogus attribute on a `pulumi_stack_reference` | rejected: `Unsupported attribute` |
| Module variable left unset while `<project>:name` is set in config | rejected: `variable "name" is required` — backs the root-module-only claim |
| `aws:region` set in config, read as `var.region` | falls through to the variable's `default` — backs the namespace claim |
| Replacement ordering with and without `create_before_destroy` | default deletes original first; `create_before_destroy = true` creates first — backs the `deleteBeforeReplace` note |

**Plugin downloads:** all worked. The HCL language plugin, the bridged `terraform-provider` (random, aws, kubernetes), the native `pulumi/kubernetes` provider, and a local module all resolved; `pulumi install` returned 0 every time. One gotcha worth knowing: a stale `sdks/<provider>/hcl.sdk.json` from an earlier install pins the previously resolved provider, so switching a source from bridged to `pulumi/` needs that directory cleared first.

**Repo checks:**

- `make lint` — passes (1794 files, 0 errors; Prettier clean).
- `make lint-prose` (Vale) — **0 findings introduced by this change.** Two `write-good.ThereIs` findings in my own new prose were rewritten. 38 pre-existing findings on these pages are untouched.
- Hugo `v0.157.0` build — all four files parse and assemble cleanly. The build then fails in the asset pipeline (`fingerprint`), because `make ensure` can't fetch `@pulumi/design-tokens` in this environment (403). I confirmed this is environmental and not a content problem by deliberately breaking one `{{% choosable %}}` close tag: that fails earlier, at the assemble phase, with a shortcode error. The real build getting past assemble means the shortcodes are sound.
- Chooser/body parity was machine-checked: every chooser's declared option list exactly matches the `choosable` bodies inside it, across all four files. No `hcl` option exists without a body.

### Left undone

- **Stack README chooser** (`stacks.md`, "Stack README") — untouched per the task's JUDGMENT marker. HCL could express it with `file()`, but it was flagged as a judgment call rather than a mechanical add.
- The inline `{{< pulumi-getstack >}}` shortcode (`layouts/shortcodes/pulumi-getstack.html`) has no `hcl` entry, so it renders nothing when HCL is selected. That's the existing behavior for `java` and `yaml` too, so this isn't a regression -- and the shortcode is outside this cluster's file list. Worth a follow-up if someone wants those three filled in.
- Nothing was blocked by unsupported syntax. No block was skipped for lack of a working construct.

### Related issues (optional)

MA-594 (cluster B). Cluster A is `camsoper/ma-594-cluster-a`.

https://claude.ai/code/session_01KJbscnHPmKxPDgMzQBTmEq

---
_Generated by [Claude Code](https://claude.ai/code/session_01KJbscnHPmKxPDgMzQBTmEq)_


CamSoper:
@claude I pushed fixes for F2 and F3 in fb6538d.

**F2** — you were right, and my line was wrong in a way that its own example contradicted. HCL's grammar is `Ident = (ID_Start | '_') (ID_Continue | '-')*` (`hclsyntax/scan_tokens.rl`, and the spec's `Identifier = ID_Start (ID_Continue | '-')*`), so hyphens are legal in labels and in traversals — which is why `variable "my-secret-value"` / `var.my-secret-value` in this same PR parses fine. The aside now just says what the `pulumi { name }` option does: give the resource a logical name that differs from its label, to match the `my-role` the other tabs on that page use.

**F3** — dropped the pin. Thirteen of the seventeen `required_providers` blocks in `content/docs/` carry no `version`; the four that do are either teaching the syntax (the HCL reference pages) or using an obvious placeholder (`1.2.3` in publishing-packages). A real pinned version on a concept page is the kind that rots silently, so a copy-pasting reader now lands on the current provider.

#update-review

---

CamSoper:
@claude rebased onto master and pushed a second-pass review in 6ac9d6f #update-review

Changes since the last review (verified with CLI 3.263.0 / language-hcl 0.18.0 on the local backend):

- **`stacks.md` `getOutput` HCL example — broken in exactly the case it claims to handle.** `try(outputs["privateIp"], null)` followed by `"logs/${local.ip}.log"` fails when the output is missing: `Invalid template interpolation value; The expression result is null`. Reproduced against a local stack reference. The fallback is now `"unknown"`; preview gives `log_key: "logs/unknown.log"`.
- **`config.md` namespaced-config tab** implied `aws:region` config was unusable and had to move into a provider block. In fact, with no `provider "aws"` block the default provider reads `aws:region` itself (set `eu-central-1`; `data.aws_region.current.region` returned it), while an explicit empty `provider "aws" {}` does *not* (`invalid AWS Region: `). The tab now says both and uses `.region` (`.name` is deprecated in hashicorp/aws 6.x).
- **`config.md` nested-values tab** claimed `content-type`'s "hyphen is not a valid attribute name". False: `var.api.headers.content-type` returned `application/json`. Replaced with the real reason `headers` is a `map(string)`: a map accepts any key; an object fixes its attributes.
- **`names.md`** said the `aws_iam_policy_document` data source is "declared here", but no example declares it, so a copy-paste failed. It now says the examples read it from a data source named `assume_role` that isn't shown.
- Minor prose cleanup in the `config.md` modules tab.

Checked and unchanged: `sensitive = true` produces real Pulumi secrets (nothing plaintext in `stack export`, derived values encrypted), and every `pulumi config set` key on the secrets page matches what its HCL tab reads.

---

tatcoo-pulumi (approved):

PR #21798, opened 2026-09-21
Add Pulumi HCL tabs to operations and CI pages (MA-594 cluster F)

CamSoper:
### Proposed changes

MA-594 cluster F: adds `hcl` to 12 language choosers across seven operations pages. Every chooser that gained `hcl` got a matching `{{% choosable language hcl %}}` body in the same edit -- no empty tabs.

**Per-file summary**

| File | Choosers | What the HCL tab shows |
| - | - | - |
| `operations/stack-management/refactoring-with-aliases.md` | 4 | Renaming (`moved` block + `pulumi { aliases }`), re-parenting (moving a declaration into a module), type change (`aliases = [{ type = ... }]`), cross-stack/project move |
| `operations/stack-management/protecting-against-undesired-changes.md` | 1 | Config-driven `pulumi { protect = var.protect }`, plus how `prevent_destroy` differs |
| `operations/stack-management/removing-resources-without-deleting-them.md` | 1 | `pulumi { retain_on_delete = true }`, and why a `removed` block can't do this job |
| `operations/container-image-reproducibility.md` | 3 | Digest pinning, registry-backed BuildKit cache, build/deploy stack split |
| `operations/continuous-delivery/github-actions.md` | 1 | Checkout + `pulumi/actions`, no language setup, no install step |
| `operations/continuous-delivery/gitlab-ci.md` | 1 | `pulumi/pulumi-base` image, no dependency install |
| `operations/continuous-delivery/google-cloud-build.md` | 1 | `pulumi/pulumi-base` builder, no dependency install |

**Left untouched on purpose:** the "Wrapping resources in a component" chooser in `refactoring-with-aliases.md`. Every sibling tab there leans on an SDK-specific root-stack sentinel (`pulumi.rootStackResource`, `NoParent`), and the HCL equivalent -- putting `{ no_parent = true, name = "..." }` on each child inside the module source -- is a judgment call about how far to restate the refactor for a language where modules *are* components. It needs a call from someone closer to the HCL story, not a guess from me.

### Verification

Pulumi CLI 3.263.0, HCL language plugin v0.18.0, local filesystem backend. Every HCL block in the diff was extracted verbatim and run through `pulumi preview`; the resource-option examples were additionally exercised as real `pulumi up` round-trips against credential-free resources (`random_pet`), since the whole point of an alias or a protect flag is what it does on the *second* run.

| - | - | - | - |
| 1 | Rename via `pulumi { aliases = [{ name }] }` | `up` creating `old_db`, rename to `primary_db` with alias, `preview` | **pass** -- 2 unchanged, no replacement |
| 2 | Rename via `moved` block | Same state, `moved { from = ... to = ... }`, `preview` | **pass** -- 2 unchanged |
| 3 | Rename baseline (no alias) | Same state, neither alias nor `moved` | **pass** -- correctly shows `+1 create / -1 delete`, confirming 1 and 2 are doing real work |
| 4 | Re-parent via `moved` into a module | `up` at root, then `moved { to = module.new_parent.random_pet.key }`, `preview` | **pass** -- only the new component wrapper is created; the resource itself is unchanged |
| 5 | `aliases = [{ type }]` | `preview` on a bound program | **pass** |
| 6 | `aliases = [{ stack, project }]` | `preview` on a bound program | **pass** |
| 7 | `aliases = [{ parent_urn }]` / `{ no_parent, name }` | `preview` on a bound program | **pass** (bind only; referenced in prose, not shown as a snippet) |
| 8 | `pulumi { protect = var.protect }` | `up` false, `pulumi config set` true, `up` (diff shows `~protect`), then drop the declaration | **pass** -- delete refused with the unprotect instructions |
| 9 | `lifecycle { prevent_destroy = true }` | `up`, then force a replacement; separately, drop the declaration | **pass** -- replacement blocked; guard lifts when the argument leaves the program, exactly the contrast the prose draws |
| 10 | `pulumi { retain_on_delete = true }` | `up`, drop the declaration, `up` | **pass** -- `delete[retain]`, provider delete never called |
| 11 | `removed { lifecycle { destroy = false } }` | `preview` | **pass (negative)** -- rejected, error points at `pulumi state delete`, which is what the prose now says |
| 12 | Container: digest-pinned ECS task definition | `preview` of the full program | **pass (partial)** -- clean preview, 5 resources, all properties bind. Creating the ECR repo needs live AWS credentials |
| 13 | Container: `cache_from` / `cache_to` | `preview` of the full program | **pass (partial)** -- clean preview, 4 resources. Same credential ceiling |
| 14 | Container: build/deploy split | Built a real `app-image/production` stack, exported `digest`, consumed it from the deploy stack | **pass** -- stack reference resolved `local/app@sha256:ffcf2924...` end to end |
| 15 | `docker-build_image` actually builds | Started a local Docker daemon, `pulumi up` on a `FROM scratch` image | **pass** -- real image, real `digest` and `ref` outputs |
| 16 | CI: HCL needs no language runtime | `pulumi preview` inside `pulumi/pulumi-base:3.263.0` (CLI only), checkout with `sdks/` committed, cold plugin cache, no install step | **pass** -- plugins downloaded on demand, preview succeeded |
| 17 | Negative test: bogus `pulumi` attribute | `pulumi { aliasez = ... }` | **pass (negative)** -- rejected with "Did you mean aliases?" |

**Plugin downloads worked.** The HCL language plugin, the bridged `terraform-provider` shim, the bridged AWS provider, and the native `pulumi/docker-build` provider all downloaded fine. Remote modules and `pulumi install` returned 0.

**Harness deltas** -- three mechanical differences between what ran and what's in the diff, none affecting the syntax being documented:

- AWS examples ran with a `provider "aws"` block setting `skip_credentials_validation` and friends. Without it the provider fails at STS before any resource input is bound, so nothing would be proven. The docs snippets omit it, matching the sibling language tabs.
- `pulumi/docker-build` ran with `version = "0.0.22"` pinned. Unpinned, plugin resolution hits `api.github.com`, which 403s from this sandbox -- and that failure mode ("could not find latest version") is itself proof the `pulumi/` source resolved correctly. The docs snippet leaves the version off rather than shipping a number that rots.
- The placeholder `resource "database" "..."` blocks (mirroring `new Database(...)` in the sibling tabs) were bind-checked with `random_pet` substituted. The `pulumi {}` option shapes are byte-identical.

**Things worth flagging**

- `pulumi { parent = module.some_module }` is documented in the HCL language reference, but in my test it silently did not re-parent: the resource stayed at the stack root with no error and no parent prefix on its URN. I avoided it and wrote the re-parenting tab around `moved` blocks, which do work. Possibly worth a `pulumi-hcl` issue -- flagging rather than filing, since I may be holding it wrong.
- Consuming a native Pulumi package from HCL has two sharp edges the tab now calls out: the `required_providers` local name becomes the resource type's prefix (so it has to stay `docker-build`, giving `docker-build_image`), and single-object inputs use `=` while list-of-object inputs are repeated blocks. Getting either wrong fails with an unhelpful "not found".
- `pulumi destroy` on a program using `prevent_destroy` needs `--run-program`, because the guard is implemented as a delete hook. Real gotcha, but out of scope here -- noting it in case it belongs on a Terraform-compatibility page.

### Checks

- `make lint` -- passes (1794 files, 0 errors; Prettier clean).
- `make lint-prose` -- 18 warnings across the seven files, **all pre-existing**; I attributed every warning against the diff hunks to confirm none are mine. Two warnings my own prose introduced (`Pulumi.Difficulty` on "just", `write-good.TooWordy` on "therefore") are fixed.
- Rendering -- `make build` can't run here (`@pulumi/design-tokens` 403s), so I rendered the seven pages through Hugo 0.157.0 with the repo's real shortcodes and asserted on the HTML: 24 rendered `pulumi-chooser` elements carry `hcl`, every declared option has a non-empty `pulumi-choosable` body, and no body lacks an option. No empty tabs.
- The Vale frontmatter advisory flags pre-existing non-sentence-case titles on `container-image-reproducibility.md` and `google-cloud-build.md`. Left alone: the second is a proper noun, and retitling the first is an editorial change beyond this cluster that would collide with other MA-594 slices.

https://claude.ai/code/session_016y8zUGZ8U2GAHdXefqD7ZW

---
_Generated by [Claude Code](https://claude.ai/code/session_016y8zUGZ8U2GAHdXefqD7ZW)_


Voyagerroc-Lab (commented):
Review summary

The HCL tabs are paired with non-empty choosable bodies across all added chooser entries, and the examples cover the operational semantics rather than only syntax. The reported preview/up round-trips specifically validate aliases, protect, retain_on_delete, CI-only HCL flows, and negative validation cases.

The site build check is still running; the local build limitation is documented in the PR. I found no content defect in the reviewed diff. I attempted an approval, but the current review token does not have explicit repository approval permission; this comment records the completed review.

CamSoper:
@claude F3: the source for the hidden-`.pulumi`-job claim is the page's own example pipeline, not GitLab's docs. It's a description of the YAML immediately below the sentence, so it's checkable against the file rather than against an external reference.

Every language tab in that chooser defines `.pulumi` and pulls it into all three jobs, e.g. the TypeScript tab:

```yaml
# Shared setup: enter the program directory and install dependencies.
.pulumi:
  before_script:
    - cd infra
    - npm ci

preview:
  extends: .pulumi
  ...
```

The `extends` link in that sentence is there so a reader who hasn't met GitLab's job-inheritance keyword can look it up — it isn't the citation for the claim. The `rules` link in the preceding clause is the one that points at GitLab's reference, and it covers the `rules` claim, which is presumably how the two got crossed.

I pushed fixes for F1, F2, and F4 in 59056fd. You were right that my earlier fix overcorrected — it made Pulumi HCL the sole exception to "installs dependencies," but C# and Java have no separate install step either. Verified per page: the C# and Java tabs on `github-actions.md` set up a runtime and go straight to `pulumi/actions`; on `google-cloud-build.md` they run Pulumi from `pulumi/pulumi-dotnet` and `pulumi/pulumi-java` with no install command; on `gitlab-ci.md` their `.pulumi` jobs are `cd infra` only. All three intros now name the three cases — TypeScript/Python/Go install, C#/Java restore during the Pulumi run, Pulumi HCL has nothing to install — which also lines up with each page's closing sentence. The "in between" wordiness went with it.

#update-review

---

CamSoper:
@claude rebased onto master and pushed a second-pass review in cb4e6df #update-review

Changes since the last review (all verified with real `pulumi up` runs on the local backend, language-hcl 0.18.0):

- **`refactoring-with-aliases.md` re-parenting tab — the alias alternative was wrong.** It said `aliases = [{ parent_urn = "urn:pulumi:..." }]` alone moves a resource into a module. It doesn't: a resource declared in a module takes the module's name as a prefix on its Pulumi name (`new_parent.moved_db`), so `parent_urn` alone previewed "+ create / - delete". With `{ name = "<old name>", parent_urn = "..." }` the resource was kept and only the component wrapper changed. The tab now shows that two-key form as a code block for the module-to-module case. (A `moved` block between two modules with different sources also previewed as delete + create, which is why that case uses the alias form.)
- **`google-cloud-build.md` intro** listed only the five language-runtime images while the HCL tab uses `pulumi/pulumi-base`; added one sentence naming it.

Also confirmed rather than changed: with a fresh `PULUMI_HOME` and committed `sdks/` (no `pulumi install`), CI preview downloads the HCL plugin and provider and succeeds, while removing `sdks/` fails with "unresolved package spec(s)… run `pulumi install`" — so the tabs' "commit the `sdks/` descriptors" instruction is correct and necessary. `protect`, `retain_on_delete`, rename/stack/project/type aliases, `moved`, and the `prevent_destroy` vs `protect` distinction each behaved exactly as the pages describe.

---

tatcoo-pulumi (approved):

PR #21800, opened 2026-09-21
Ignore wedged runs the deploy turnstile can never outlive

CamSoper:
### Proposed changes

Every production deploy since 2026-09-18 has taken ~54 minutes instead of ~9. This is why, and the fix.

> **Note:** this branch was force-pushed after the first review. It was originally cut from #21789's branch by mistake, so the PR showed that PR's 16 files as well — 18 total. It is rebased onto `master` now and contains only the two files below. The first review read the wrong diff; the second reflects reality.

**What's wrong**

`scripts/await-in-progress.js` serializes deploys by waiting for every run of the same workflow with a lower run ID. Its only early exit is:

```js
if (recent.length === 0) { break; }
```

A run that never completes therefore blocks it permanently — it stays lower-id than every new run forever, the loop always runs out its 45-minute cap, and the script then proceeds **unserialized**, which is the exact condition it exists to prevent.

Three such runs have accumulated here, all reported by the API as `queued`, none of which ever started:

| Run | Workflow | Queued since |
|---|---|---|
| [23509983446](https://github.com/pulumi/docs/actions/runs/23509983446) | `build-and-deploy.yml` | 2026-03-24 |
| [31117840729](https://github.com/pulumi/docs/actions/runs/31117840729) | `build-and-deploy.yml` | 2026-08-06 |
| [32907599016](https://github.com/pulumi/docs/actions/runs/32907599016) | `testing-build-and-deploy.yml` | 2026-08-25 |

Only 31117840729 could be cancelled. The other two refuse cancel, force-cancel and delete alike:

```
409 Cannot cancel a workflow run that has not been queued yet
403 Could not delete the workflow run
```

There is no API that clears them, so filtering them here is the only available remedy. (Removing them entirely needs a GitHub Support ticket; this makes them harmless in the meantime.)

**When it started**

They became load-bearing on 2026-09-18, when #21708 added `queued` and `waiting` to the status query. That was a *correct* fix for a real hole — a queued run genuinely can hold the stack — but it also made these three visible for the first time. Before it the query was `const status = "in_progress"`, and they were invisible.

The effect is measurable: the 40 production runs since that commit have a median wall time of **54.3 minutes** (clustered 52–59) against **~9 minutes** before it. That is not a build-time distribution; it is a constant, and the constant is `maxWaitMs`. The three `[409] Conflict: Another update is currently in progress` failures on 2026-09-21 were the first symptom anyone traced back here.

**The fix**

`dropStaleRuns()` filters runs older than 48 hours out of the wait list before the id comparison, and logs each one it skips.

On the threshold: six hours was the obvious number and it was wrong. GitHub's six-hour limit is a *job* timeout whose clock starts when the job gets a runner — queue time isn't counted — so a run that waited for capacity can be older than six hours while genuinely holding the stack. 48 hours clears any plausible queue wait plus a full six-hour job, and is still two orders of magnitude below the ages actually observed. It's a heuristic, not a proof, so the worst case is worth stating: the wait loop already gives up after `maxWaitMs` and proceeds regardless, so this filter can only change *when* a run stops waiting, never *whether* it does.

Each skip is announced once per deploy as a `::warning::` naming the run, its status and its age. Once, not per poll — GitHub renders at most 10 warning annotations per step, and 45 repeats would push the "giving up on the queue and proceeding" warning off the end of the list, which would make the diagnostic the thing hiding the diagnosis. The previous code reported a bare count (`Found 5 other job(s)`), so a queue holding a run from March read exactly like a busy afternoon; that silence is why this took three days to find.

`@octokit/rest` moves to a lazy require inside `waitForInProgressRuns`, and the entry point is guarded by `require.main === module`, so the test can reach `dropStaleRuns` without the dependency installed and without the script talking to GitHub on import. The `node ./scripts/await-in-progress.js` call in `scripts/ci-push.sh` is unaffected — verified in both directions.

### Testing

`scripts/await-in-progress.test.js`, picked up by `scripts/run-unit-tests.sh` and so by `make test`. 13 tests here; full suite **67 passing**.

Coverage includes the real wedged run, a run older than the six-hour job timeout but inside the bound (which must still be waited on), both fail-safe paths, the log-once-per-deploy behavior across 45 simulated polls, and that suppressing the log doesn't suppress the filtering.

Verified non-vacuous against seven mutations — never drop, drop everything, revert the threshold to 6h, log on every poll, never log, drop unparseable timestamps, ignore the `created_at` fallback. Each fails at least one test; the suite is clean on the real code.

### Related issues

None filed. Came out of the #docs-ops investigation into the 2026-09-21 build-and-deploy failures, alongside #21789, which retries the resulting 409s at the Pulumi layer. Now that this branch is rebased, the two are genuinely independent and neither depends on the other.

https://claude.ai/code/session_01Eif28XV3WrnCmpfdNK3HZK


CamSoper:
@claude Pushed fixes for F1 and F2, and rebased the branch -- details below. #update-review

**First, the branch was wrong.** I cut it from #21789's branch instead of `master`, so the diff you reviewed carried that PR's 16 files as well. It's rebased onto `master` now and contains only `scripts/await-in-progress.js` and its test. Force-pushed, so the review you just did read a diff that no longer exists. Sorry for the wasted pass.

**F1 -- fixed, and you were right about the reasoning as well as the number.**

Six hours was the obvious threshold and it was wrong for exactly the reason you gave: the limit is a *job* timeout whose clock starts when the job gets a runner, so queue time isn't counted and a run can be older than six hours while genuinely executing. Calling that "safe by construction" in the comment was wrong twice over -- wrong bound, and a guarantee I couldn't make.

The bound is 48 hours now: past any plausible queue wait plus a full six-hour job, still two orders of magnitude below the ages actually observed (27 to 181 days). The comment states the worst case rather than claiming a proof -- the wait loop already gives up after `maxWaitMs` and proceeds regardless, so this filter can only change *when* a run stops waiting, never *whether* it does. New test: a run 12h old is kept, since 12h is not proof of anything.

I didn't lean on `retry_on_stack_lock` as the justification, as your note offered -- it lives in #21789 and isn't on this branch, so it would have been a guarantee borrowed from a PR that hasn't merged.

**F2 -- fixed.** `dropStaleRuns` now takes a `warned` set of already-announced run ids that `waitForInProgressRuns` holds across poll iterations, so each wedged run is reported once per deploy instead of ~45 times. Filtering still runs every poll; only the logging is deduplicated. Your point about the 10-annotation cap is the part that made this worth fixing rather than tidying -- the repeats would have pushed the "giving up on the queue and proceeding" warning off the end, so the diagnostic would have been the thing hiding the diagnosis.

Tests: 45 simulated polls produce one warning rather than 45, distinct runs each get their own, and suppressing the log doesn't suppress the filtering.

Both fixes re-verified non-vacuous by reverting each one as a mutation and confirming it fails a test. Full suite 67 passing.

---

PR #21801, opened 2026-09-21
Add Pulumi HCL tabs to migration and adoption guides (MA-594 cluster E)

CamSoper:
### Proposed changes

MA-594 cluster E: adds an HCL tab to the language choosers across the seven migration and adoption guides, so a reader who keeps their config in HCL sees a real path instead of silently getting another language's tab.

**16 HCL tabs added across 7 files. No chooser gained an `hcl` option without a matching `{{% choosable language hcl %}}` body** -- I wrote a checker over all `chooser`/`choosable` pairs in the touched files and it reports 0 mismatches.

Per file:

- **`from-terraform.md`** (2 blocks) -- the highest-value page here, because with Pulumi HCL there's nothing to rewrite. The converter tab says exactly that, and shows the verbatim error you get if you ask the converter for HCL anyway. The module tab shows that from HCL a Terraform module is just a `module` block, with no `pulumi package add` and no generated SDK.
- **`from-arm.md`** (3 blocks) -- coexist-with-ARM reads the deployment through an `azurerm` data source; convert shows `--language hcl`; import shows an `import` block. Also fixed the prose that claimed `pulumi convert --from arm` returns ".NET, TypeScript, Python, Go, Java, or YAML" -- HCL is a valid target, so it's in the list now.
- **`from-cloudformation.md`** (3 blocks) -- `data "aws_cloudformation_stack"`, an `aws_cloudformation_stack` resource with a heredoc template, and an `import` block.
- **`from-serverless.md`** (2 blocks) -- stack-output reference, plus the full Serverless-equivalent app (DynamoDB, IAM, Lambda, API Gateway v2, Lambda permission).
- **`from-kubernetes.md`** (3 of 5 -- see "Left undone") -- `ConfigFile`, `ConfigGroup`, and YAML rendering.
- **`migration/import/_index.md`** (2 blocks) -- the generated program with `pulumi { protect = true }`, and the `import` resource option via `pulumi { import_id }`. The first tab also points at `pulumi import --from hcl <statefile>`, which is the bulk path for HCL users.
- **`use-terraform-module.md`** (1 block) -- the S3 module as a `module` block.

One thing worth a reviewer's eye: several of these pages carry prose *outside* the chooser that only makes sense for the SDK languages ("this will generate a Pulumi program...", the `pulumi package add` walkthrough). Left alone, that copy flatly contradicts the HCL tab sitting right above it. I scoped those passages to the languages they apply to rather than duplicating them into every tab.

### Verification

Every example was built as a scratch `runtime: hcl` project outside the repo and run through `pulumi install` + `pulumi preview`, or `pulumi up` where no cloud credentials were needed. Pulumi CLI v3.263.0, HCL language plugin v0.18.0.

- **pass** -- completed with no errors
- **pass (partial)** -- HCL parsed, provider/type resolved, resource graph bound; stopped at live cloud credentials. This is the environment ceiling, not a defect.

| # | File | Example | How verified | Result |
|---|------|---------|--------------|--------|
| 1 | from-terraform | `Pulumi.yaml` + `.tf`, `pulumi install` | `pulumi preview` on a random-provider project | pass |
| 2 | from-terraform | `pulumi convert --language hcl` is rejected | ran it; error captured verbatim | pass |
| 3 | from-terraform | `terraform-aws-modules/vpc/aws` 5.19.0 `module` block | module downloaded, `components:index:Vpc` registered, `vpc_id` resolved; stopped at AWS STS. Inputs cross-checked against the module's own `variables.tf` | pass (partial) |
| 4 | from-terraform | local `module` block (`source = "./..."`) | `pulumi preview`, full graph, no creds needed | pass |
| 5 | from-arm | `azurerm_resource_group_template_deployment` data source + storage account/container/blob | parsed and bound; stopped at Azure auth. Every type and attribute checked against `pulumi package get-schema terraform-provider hashicorp/azurerm` | pass (partial) |
| 6 | from-arm | `pulumi convert --from arm --language hcl` | **not run** -- see below. The `Pulumi.yaml` shape shown is verified | not run |
| 7 | from-arm | `import` block + `azurerm_storage_account` | parsed and bound; stopped at Azure auth; schema-checked | pass (partial) |
| 8 | from-cloudformation | `data "aws_cloudformation_stack"` + `aws_instance` | parsed and bound; stopped at AWS STS; schema-checked | pass (partial) |
| 9 | from-cloudformation | `aws_cloudformation_stack` + heredoc `template_body` | parsed and bound; stopped at AWS STS; schema-checked | pass (partial) |
| 10 | from-cloudformation | `import` block + `aws_vpc` | parsed and bound; stopped at AWS STS; schema-checked | pass (partial) |
| 11 | from-serverless | CFN data source + `aws_sqs_queue` | parsed and bound; stopped at AWS STS | pass (partial) |
| 12 | from-serverless | full app: DynamoDB, IAM role/policy, Lambda, API Gateway v2 (api/integration/route/stage), Lambda permission | parsed and bound, incl. `jsonencode` and interpolation; stopped at AWS STS. All 10 types + attributes schema-checked | pass (partial) |
| 13 | from-kubernetes | `kubernetes_yaml_v2_config_file` | type resolved to `kubernetes:yaml/v2:ConfigFile`, inputs converted; stopped at "cluster is unreachable" | pass (partial) |
| 14 | from-kubernetes | `kubernetes_yaml_v2_config_group` | type resolved to `kubernetes:yaml/v2:ConfigGroup`; stopped at cluster | pass (partial) |
| 15 | from-kubernetes | `render_yaml_to_directory` + Deployment + Service | **`pulumi up` succeeded** and wrote real manifests to `yaml/1-manifest/` | pass |
| 16 | import guide | `aws_vpc`/`aws_subnet` with `pulumi { protect = true }` | parsed and bound; stopped at AWS STS | pass (partial) |
| 17 | import guide | `pulumi { import_id }` and `import` block on `aws_security_group` | parsed and bound; stopped at AWS STS | pass (partial) |
| 18 | use-terraform-module | `terraform-aws-modules/s3-bucket/aws` 4.1.2 `module` block | module downloaded and bound; stopped at AWS STS. `bucket_prefix`, `tags`, `s3_bucket_arn`, `s3_bucket_id` all confirmed in the module source | pass (partial) |

**Negative tests** (all confirm the checks above actually bite):

| Test | Result |
|------|--------|
| bogus resource attribute (`random_pet`) | rejected -- "Unsupported argument" |
| bogus key inside a `pulumi { }` block | rejected -- so `protect` / `import_id` are really validated |
| bogus module output (`module.vpc.not_a_real_output`) | rejected -- "unknown node" |
| bogus stack-reference output key | rejected -- "Invalid index" |
| camelCase key where snake_case is required (`targetPort`) | rejected by the k8s provider |
| `pulumi convert --from terraform --language hcl` | rejected with the documented message |
| bogus `azurerm`/`aws` attribute | **not** rejected -- provider auth short-circuits input validation, which is why those rows were schema-checked against `get-schema` instead of trusted to `preview` |

`pulumi_stack_reference` got a full round trip: I stood up a producer stack with a real output and confirmed the consumer read the actual value back, and that a bogus key fails.

**Plugin downloads.** Worked: HCL language plugin, bridged Terraform providers via the OpenTofu registry (`random`, `aws`, `azurerm`, `helm`), remote registry modules, and native `pulumi/kubernetes`. Blocked in this environment by a proxy 403 on `api.github.com` / `get.pulumi.com`: every converter plugin (`terraform`, `arm`, `kubernetes`) and `pulumi/azure-native`.

`make lint` passes (0 errors, Prettier clean). Vale reports 0 findings on lines this PR adds; the pre-existing warnings on these pages are untouched. Hugo's assemble stage parses all seven files with no shortcode errors -- I confirmed that stage is actually load-bearing by breaking a shortcode on purpose and watching it fail. A full `hugo` build can't finish here because the theme CSS pipeline needs `@pulumi/design-tokens`, which is behind the same 403.

### Left undone

**Two Helm blocks in `from-kubernetes.md` have no HCL tab, deliberately.** The native Helm component types spell as `kubernetes_helm.sh_v3_chart` and `kubernetes_helm.sh_v3_release` -- the plugin's own "did you mean" suggests exactly that. They can be *declared*, but the dot makes them impossible to *reference*:

```
error: unknown node "kubernetes_helm.sh_v3_release"
```

Both doc examples need a reference (`wordpress.getResource(...)`, `wordpress.status.namespace`), so there's no way to write a working HCL tab. Rather than invent syntax or ship a crippled example, I left those two choosers alone. This looks like a `pulumi-hcl` bug -- the `helm.sh` module segment isn't sanitized into a valid HCL identifier -- and is probably worth an upstream issue.

**`pulumi convert --from arm --language hcl` (from-arm.md):** the tab documents the command and the resulting project shape, but not a converted program body. I couldn't run the ARM converter here (403), so rather than fabricate output and present it as the converter's, I left the program out. Worth filling in by anyone who can run the converter.

**The configuration-transformations chooser in `from-kubernetes.md` is untouched**, as specified -- transformations are callbacks, not something the `pulumi` block expresses.

Nothing outside the seven listed files was modified.

### Related issues (optional)

MA-594

https://claude.ai/code/session_01Q3U6eAGcmAVoouJFHdTW8N

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q3U6eAGcmAVoouJFHdTW8N)_


CamSoper:
@claude F1: disagreeing — `hcl` **is** a valid `--language` target. The two pages don't contradict each other; the restriction on `from-terraform.md` is specific to `--from terraform`, not to `--language hcl`.

Verified against CLI 3.263.0.

`--language hcl` works. Converting a trivial PCL program (using the built-in `pcl` front end, which needs no downloaded converter plugin):

```
$ pulumi convert --from pcl --language hcl --out ../out2
Converting from pcl...
Converting to hcl...
$ ls ../out2
Pulumi.yaml  main.tf
```

`Pulumi.yaml` came out with `runtime: hcl` and `main.tf` with the converted `resource`/`output` blocks. So the HCL language plugin is a real conversion target that `pulumi convert` will render into.

The Terraform case is a single deliberate guard, with its own message:

```
$ pulumi convert --from terraform --language hcl --out ../out3
error: cannot convert a Terraform program to the "hcl" language: pulumi-hcl
runs Terraform directly, so no conversion is needed; converting would
re-home every resource onto a different provider and show a delete and
create for each one on the next preview
```

That's a `--from terraform` guard, and the reason it exists (no conversion needed, and converting would re-home resources) doesn't apply to ARM. `from-terraform.md` documents that specific rejection correctly; `from-arm.md` describes a different, unguarded combination.

What I think misled the check: `pulumi convert --help` prints `Valid target languages: typescript, python, csharp, go, java, yaml`. That string is stale — it's informational and isn't used for validation. `--language` resolves against any installed language plugin, which is why `hcl` renders above despite not being listed.

One honest limit: I could not run `--from arm --language hcl` end to end here, because fetching the ARM converter plugin needs `api.github.com`, which 403s in this sandbox:

```
error: load converter plugin: install "arm": could not find latest version for
provider arm: 403 HTTP error fetching plugin from
https://api.github.com/repos/pulumi/pulumi-converter-arm/releases/latest
```

It failed loading the converter (the `--from` side), never reaching the language selection. Since `--from` and `--language` are independent — the converter emits PCL, the language plugin renders it — and I've shown the language side accepts `hcl`, the ARM tab's command holds. Happy to be corrected by anyone who can run the ARM converter.

I also pushed fixes for F3 and F4 in 1ae9f53:

- **F3** — wrapped the `getResource` paragraph in a choosable for the five SDK languages plus YAML, matching the ARM-ID paragraph in `from-arm.md`.
- **F4** — `managed_policy_arns` does still work today: the unqualified `aws` source resolves to `registry.opentofu.org/hashicorp/aws` 6.66.0, and a negative control shows argument validation runs *before* any credential check (a bogus argument dies with "Unsupported argument" and never reaches STS, while `managed_policy_arns` passes validation and fails only on dummy credentials). But it's deprecated upstream and the example pins no version, so I switched it to `aws_iam_role_policy_attachment` — idiomatic, survives the eventual removal, previews with the same clean binding.

#update-review

---

CamSoper:
@claude rebased onto master and pushed a second-pass review in 1211970 #update-review

Rebase: one conflict in `from-terraform.md` (kept master's `/docs/integrations/terraform/modules/` link and this PR's HCL tab). The module guide left this PR — #21806 moved it to `integrations/terraform/modules.md`, which #21790 now owns and carries this PR's tab.

Review fixes (details and evidence in the commit message and PR description):

- **`import/_index.md` generated-code tab was wrong.** It showed hand-written Terraform-provider blocks (`aws_vpc`, `aws_subnet`) with two resources and different CIDRs from the siblings. The import file names `aws:ec2/vpc:Vpc`, but the bridged `aws` provider registers `aws:index:*` types, so those blocks wouldn't match imported state. Now shows what the HCL generator actually emits (`pulumi/aws`, `aws_ec2_vpc` / `aws_ec2_subnet`, all three resources), produced with `pulumi convert --from pcl --language hcl`.
- **`from-arm.md` tabs moved from `azurerm` to `pulumi/azure-native`**, matching the siblings and the converter's output. The read tab used the Pulumi logical name instead of the real deployment name; the adopt tab's `azurerm_storage_account` contradicted "copy-paste the program `pulumi convert` generated". Adopt now previews `= import`.
- **Shared prose the HCL tabs made untrue is scoped away:** the `pulumi convert --from terraform` lead-in, the Kubernetes "exports the frontend's cluster IP" (HCL's `resources` map is untyped, so only `id` is visible), and "the full power of a programming language".
- `from-serverless.md` stage comes from `pulumi.stack` like every sibling; `from-kubernetes.md` render section uses the siblings' logical names and ran `pulumi up` end to end.

---

tatcoo-pulumi (approved):

PR #21802, opened 2026-09-21
Re-score Sentinel G4 when a staging deploy finishes

CamSoper:
### Proposed changes

Sentinel gate **G4 infra-evidence** requires a successful staging deploy at the PR's current head, and it is the one gate documented as **not waivable**. Nothing ever re-ran the Sentinel when that deploy finished, so an infra PR could sit indefinitely on a red gate whose condition it had already met.

PR #21789 is the worked example: Sentinel scored G4 red at 22:41 with no deploy yet; the staging deploy ([run 35659427886](https://github.com/pulumi/docs/actions/runs/35659427886)) went green at 22:44; nothing re-read it; the gate cleared at 23:05 only because someone hand-dispatched `review-sentinel.yml` with `pr_number: 21789`. **This PR automates that manual action.**

#### The one-liner does not work, and I can prove it

The tempting fix is adding `"Build and deploy testing"` to `review-sentinel.yml`'s `workflow_run:` list. It is a silent no-op, twice over:

1. **The event never fires.** GitHub emits no `workflow_run` cascade for a run dispatched with `GITHUB_TOKEN`, and both deploy lanes (`staging-deploy-auto.yml` and `/deploy-staging`, via `staging-deploy.sh`) dispatch with `GITHUB_TOKEN`. This repo already paid for that lesson in #21696 — `staging-status.yml` is a `workflow_run` listener on exactly this workflow, with exactly this wiring, and it has never once fired for a PR-branch deploy.
2. **Even if it fired, it could not name the PR.** `review-sentinel.yml`'s "Resolve PR number" step reads `github.event.workflow_run.pull_requests[0].number`; a dispatched run's event does not populate it. The step would resolve empty, print "No PR number resolvable from this event", and every subsequent step is `if: steps.pr.outputs.pr != ''` — so the trigger would fire and do nothing.

`test_staging_sweep.py::test_adding_the_deploy_to_the_sentinels_workflow_run_list_is_not_the_fix` pins that workflow **out** of the Sentinel's list, with the reasoning, so the no-op cannot be re-landed later as a fix.

On your third question — `workflow_run` only firing for default-branch workflow files — that one does not bite: the listener is on master and fires for runs on any branch. It is the token provenance, not the branch, that kills it.

#### How I verified it, concretely

Everything below is against the live `pulumi/docs` API, not reasoning from docs.

**1. The cascade genuinely never fires.** Over the entire history of `staging-status.yml`: **95 runs — 94 were master *pushes*** (correctly skipped by its own job `if`) **and 1 was a hand-run backfill.** In the same history there are **28 dispatched PR-branch deploys** of "Build and deploy testing". They produced **zero** listener runs: no listener run's `head_sha` matches any of those 28 deploys, and exactly one listener run ever got past its `if` (the manual backfill). Corroborating detail: PR #21789's head `55cd947f2` carries 14 commit statuses and **not one of them is `staging/pulumi-test-io`**, despite a green deploy — the listener never ran for it. Its G4 passed on witness (2), the run record, and only after the manual dispatch.

**2. One correction to the premise.** The REST representation of run 35659427886 *does* populate `pull_requests` with `#21789`, so the "empty array" assumption in the brief is not reliably true for the REST API (the header comment in `staging-status.yml` asserting it is overstated). It is moot here — the event never arrives, so there is nothing to read the field off of — but I would rather flag it than quietly lean on a claim I found to be wrong. The fix does not depend on it either way: the sweep resolves the PR from the head branch and passes `pr_number` explicitly, which is deterministic.

**3. The sweep produces the right answer on production data.** I replayed `staging-sweep.py` in dry-run against a snapshot captured from the live API (real runs, real commit statuses, real open PRs), over the 6-hour window containing the incident:

```
POST repos/pulumi/docs/actions/workflows/review-sentinel.yml/dispatches
     {"ref": "master", "inputs": {"pr_number": "21789"}}
POST repos/pulumi/docs/statuses/55cd947f2…  {"state": "success", "context": "staging/pulumi-test-io", …}
```

It resolved run 35659427886 → branch `camsoper/loving-brown-wsse8v` → **PR #21789**, confirmed that SHA is still the PR's head, and dispatched the Sentinel with that number — byte-for-byte the manual action that cleared the gate. Two other deploys in the window (`jeffmerrick/invalidate-dev-slash-path`) correctly got their status written with **no** poke, because their branch has no open PR.

**4. It is idempotent on that same real data.** I applied sweep 1's writes back into the snapshot and re-ran: sweep 2 finalized 0, poked 0, and sent **no requests at all** — all three SHAs reported `already-recorded`. That matters because this runs every 10 minutes over a 6-hour window.

**What I could not verify end to end:** I cannot make GitHub actually fire a `schedule` trigger from here, so "cron fires, the job starts, `GITHUB_TOKEN` is accepted on `POST /statuses` and `POST /actions/workflows/…/dispatches`" is verified by shape and by the fact that `staging-status.yml` and `staging-deploy-auto.yml` already perform those exact calls with that exact token under the same `permissions:` block — not by observing a scheduled run. First real confirmation will be a scheduled run appearing in the Actions tab after merge. Worth watching once.

#### The fix

A `schedule` is the only trigger immune to **both** documented failure modes — branch age (#21676: a `workflow_dispatch` runs the workflow file from the ref it was dispatched at, so a job added on master does not exist for older branches) and token provenance (#21696). That is not my invention; it is the writer `staging-deploy.sh` and `staging-status.yml` have both been asking for in prose since September:

> If a pending status ever comes back, it needs a writer that CANNOT miss — a `schedule` sweep that finalizes any pending status whose run has completed, not an event cascade.

So `staging-status.yml` gains a 10-minute `sweep` job running the new `scripts/review-v3/staging-sweep.py`, which finalizes every dispatched deploy the cascade never reported:

- lists recent "Build and deploy testing" runs, keeping `workflow_dispatch` + completed + non-default-branch (a master push deploy is nobody's evidence — same filter the listener's `if` applies);
- groups by head SHA, because two deploys of one SHA are one fact: a success anywhere in the group wins, matching what `_staging_evidence` reads;
- writes the terminal `staging/pulumi-test-io` status when that SHA does not already carry that verdict from an authorized writer;
- pokes `review-sentinel.yml` with an explicit `pr_number` when the deploy **succeeded** and that SHA is still the open PR's current head.

Design points worth a reviewer's eye:

- **Idempotence keys on the verdict, not on the presence of a status.** A failed deploy's green retry still flips the gate; a hand-written status from a non-allowlisted writer does not count as "handled" (same writer allowlist `sentinel.py` uses when reading one as evidence, so nobody with push access can park a status to suppress the real one).
- **The poke is sent before the status write.** The status is the idempotence key, so sealing it first would make a failed poke permanent — the exact failure mode this workflow exists to prevent.
- **Only a success pokes**, and only at the PR's current head. A failed deploy leaves G4 red, which it already was; a superseded SHA was already re-scored by the push that moved it.
- The `workflow_run` listener and the `run_id` backfill both **stay**. When the cascade does fire it is free latency, and the two converge without coordinating.
- No `pip install` in the sweep job — the script imports `gh_client` and the stdlib only. It runs 144×/day.
- No concurrency group, no PR code, checkout pinned sparse to the default branch — same invariants as the job beside it.

**Considered and rejected:** dispatching the deploy with `PULUMI_BOT_TOKEN` instead of `GITHUB_TOKEN` so the existing cascade fires. It is a smaller diff, but it puts an org-scoped PAT and an ESC session into the unattended lane (whose own header flags that combination as the higher-risk one), it would need doing in both lanes, and it trades a deterministic mechanism for a GitHub behavior I cannot verify without shipping it. The sweep is slower and duller and cannot silently stop working.

**Cost:** up to one 10-minute cron interval of latency on top of a ~9-minute deploy, and 144 short runs/day. Happy to stretch the cron if that is the wrong trade.

Also updated: G4's remedy text described only "it never ran or it failed" — the case this PR fixes (a deploy that already succeeded) had no sentence. It now says the sweep picks it up and no action is needed.

#### Testing

- `make test-review-pipeline` — passes. New suite `scripts/review-v3/test_staging_sweep.py`, 23 tests, auto-discovered by both the pytest lane and the `--self-test` lane; full review-v3 suite 555 passed.
- `make lint` — passes (1794 files, 0 errors; prettier clean).

Coverage includes candidate selection, the lookback window, per-SHA verdict grouping, the three contract description strings, poke-before-status ordering, second-sweep-does-nothing, failure-then-green-retry, the unauthorized-writer case, failed/stale-head/no-PR non-poke paths, dry-run, both degradation paths, the workflow wiring, and the regression test pinning the one-liner out of `review-sentinel.yml`.

### Related issues (optional)

Came out of the #docs-ops investigation into the 2026-09-21 build-and-deploy failures. Independent of #21789 and #21800 — no overlapping files; neither branch was touched.

https://claude.ai/code/session_01JmcykGPCw3JYcFHEFxwzwX

---
_Generated by [Claude Code](https://claude.ai/code/session_01JmcykGPCw3JYcFHEFxwzwX)_


CamSoper:
@claude #update-review

Fixed in e001bce:

- F1: `open_pr_for_branch` now raises on a failed `GET /pulls` instead of returning `None`. `sweep()` catches it, logs a warning, records `pr-lookup-failed`, and `continue`s without writing the status, same path as `poke-failed`. So a transient error gets retried next cycle instead of being sealed as `no-open-pr`. New test `test_a_failed_pr_lookup_is_retried_rather_than_sealed_as_no_open_pr` fails on the old code and passes now.
- F2: went with the equality test. Importing from `sentinel.py` isn't an option since it pulls in yaml via `routing` and the sweep job deliberately installs nothing. So the comment now says why the constants are re-spelled, and `test_the_shared_constants_match_sentinel` pins all three (`STAGING_WORKFLOW_FILE`, `STAGING_STATUS_CONTEXT`, `STAGING_STATUS_WRITERS`) equal to sentinel's.

`make test-review-pipeline` passes.

---

CamSoper:
@claude #new-review

PR #21803, opened 2026-09-22
Refuse to publish an older deploy over a newer one

CamSoper:
### Proposed changes

Production deploys can publish an **older** commit over a newer one, silently, and nothing detects it. This adds a guard that refuses.

#### The mechanism

`scripts/sync-and-test-bucket.sh` builds a per-run S3 origin bucket and writes `./origin-bucket-metadata.json` naming it. `infrastructure/index.ts` reads that file at program load (~line 220) and feeds `.bucket` into the CloudFront distribution's origin (~line 236, and the `origins` block at ~1170). **So `pulumi up` on `www-production` IS the publish step** -- it swings the live origin to whatever the local metadata file happens to name. It reads only `.bucket`; the `timestamp` and `commit` the sync script also writes are ignored, and nothing anywhere compares the incoming build against what is already live.

Deploys are serialized only probabilistically. `scripts/await-in-progress.js` waits for runs with a **lower** run id and gives up after 45 minutes. An older run that outlives that cap -- or that was invisible to the turnstile -- reaches `pulumi up` after a newer run has already published, and the origin flips backwards. The site then serves older content until the next push or one of the three daily scheduled rebuilds, which across the overnight gap between the 4:23pm and 7:17am slots is up to ~8 hours.

Nothing catches it. `post-deployment-health-check.yml` runs only on success and checks liveness, never which commit is live. #21789 makes this both more reachable and quieter: retrying a `pulumi up` that failed with `[409] Conflict` turns a run that used to fail loudly into one that can succeed late and regress the site with every run green.

I verified each link in that chain against the code before writing anything -- see "How I verified it fires" below for the part that matters most.

#### What this does

`scripts/check-publish-ordering.js` runs immediately before `pulumi up` in `run-pulumi.sh` (and in `laptop-deploy.sh`). It compares the incoming build's **GitHub Actions run id** against the run id of the bucket the stack currently names as its origin, and on production it exits 1 rather than let an older run publish.

Run ids are now recorded in the metadata document, which grows two fields:

```diff
 {
     "timestamp": 1758400000000,
     "commit": "a1b2c3d4...",
+    "runId": 35266996109,
+    "runAttempt": 1,
     "bucket": "www-production-pulumi-docs-origin-push-a1b2c3d4-k3f9j2",
     "url": "http://..."
 }
```

The `printf` block that built that JSON inline moved to `render_origin_bucket_metadata()` in `common.sh`, so the document's validity is unit-testable without a full site build. `runId`/`runAttempt` are JSON numbers or `null`, never strings.

### Design choices, and what I rejected

**Ordering key: run id.** Run ids are assigned by GitHub, strictly increasing per repo, and need no clock and no git history. They are also *precisely* the ordering `await-in-progress.js` already promises ("a run only ever waits for runs with a LOWER id"), so this enforces at publish time the invariant the turnstile tries to arrange at queue time. Enforcing the same invariant twice at two different points is the whole idea -- the turnstile can fail open, this can't.

- **Rejected: wall-clock `timestamp`.** It's written when the sync finishes, not when the publish happens, so a slow early run can carry a *later* timestamp than a fast later one and sail straight through a naive "reject older timestamp" check -- while two runners a few seconds out of sync fail one that should pass. It's kept as a **warn-only** secondary signal for when run ids aren't available on both sides, which is exactly the rollout window.
- **Rejected: commit ancestry (`git merge-base --is-ancestor`).** Most semantically correct, and unusable here. `testing-build-and-deploy.yml` checks out at the default depth of 1, so the history isn't there. `staging-deploy.sh` dispatches that workflow at arbitrary **PR branch refs**, where "is an ancestor of master" is false for entirely healthy deploys. And force-pushes break it. (Production does have `fetch-depth: 0`, so it's available *there* -- but a guard that works on one of two deploy paths and needs a second mechanism for the other is worse than one rule that works on both.)

**Source of truth for "what is live": two reads, each authoritative for its half.**

1. *Which* bucket is live -> the stack's own `originS3BucketName` output. Written by the same `pulumi up` that swings the origin, so it can't disagree with the distribution short of hand-edited state.
2. *What's in it* -> `metadata.json` read directly out of that bucket over S3. The sync script uploads a copy into every bucket it builds, so the bucket describes itself. That stays correct even after a manual pin, and it's uncached.

- **Rejected: `https://www.pulumi.com/metadata.json`.** Not for the obvious reason -- that path has its own zero-TTL cache behavior in `index.ts`, so it isn't edge-cached. The problem is a layer down: swinging the origin is a CloudFront **distribution config** change, and those take 15-20 minutes to propagate (this repo's own docs say so). For that whole window the public endpoint can still answer from the previous origin -- and that window is exactly the one this check runs in, since the race it detects is two deploys minutes apart. Of the three candidates it's the only one that's systematically stale precisely when it's being asked.
- **Rejected: stack outputs alone** (exporting commit/runId from `index.ts`). One read instead of two, but it's the variant the "stale after a manual change" failure mode actually bites: pin `originBucketNameOverride`, run `up`, and the bucket output and the metadata outputs describe different things. Reading the bucket's own copy can't desync.

**Scope.** Keyed off the stack name, because a rule that's right for production is wrong for a scratch environment:

| Stack | Behavior | Why |
| --- | --- | --- |
| `www-production` | **abort** (exit 1) | The live docs site. Silent, site-wide, persists for hours. |
| `www-testing` | **warn** (`::warning::`, continue) | Deploys here are *deliberately* unordered: `/deploy-staging` dispatches at PR branch refs and "next merge to master resets pulumi-test.io" is documented behavior. Failing those would make the lane flaky for a reason the PR author can't act on. Blast radius is one reviewer's preview. |
| dev stacks, `staging`, PR previews | **skip** | Single owner, no ordering to violate. PR previews run `pulumi preview`, never `up`. |

`resolveMode()` is unit-tested against the literal stack names, including a case asserting that bare `"production"` does **not** match -- `sync-and-test-bucket.sh` has a latent instance of exactly that bug today (it compares `stack --show-name` against `"production"` while the stack is named `www-production`, so its bucket-tagging branch never fires). Not fixing that here; flagging it as a follow-up.

**Abort vs. warn on production.** Abort, and the usual objection doesn't apply. "Abort" normally means somebody has to understand something at 2am; this one doesn't, because **when the guard fires, the content that should be live already is** -- a newer run published it. The losing run has nothing to contribute. Whoever reads the red build confirms the named live origin looks right and closes the tab. That's why the message leads with it in as many words rather than making you infer it:

```
=================================================================
  This deploy would move the live site BACKWARDS. Stopping.
=================================================================

  incoming: www-production-...-push-aaaaaaaa-fqwwnrm2, commit aaaaaaaa, run 35266933458
  live:     www-production-...-push-bbbbbbbb-fqwwnrm9, commit bbbbbbbb, run 35266996109

  run 35266933458 would publish over run 35266996109, which is newer and already live.

  Nothing is broken and nothing needs fixing right now. A newer run
  already published, so the site is serving the content it should be.
  This run raced it, lost, and is being stopped before it can undo the
  result. Confirm the live origin above looks right and you are done.
  ...
```

**Failing open.** Every unknown resolves to "allow, and say so": no prior deploy, a cleaned-up live bucket, missing or unparseable metadata, a metadata file with no run id (which is what every bucket built before this lands looks like -- it self-heals after one deploy). The one thing that does *not* fail open is not knowing which stack we're publishing to: that exits 2, because a silently skipped check reads exactly like a passing one.

The happy path logs too, deliberately -- `publish-ordering: OK -- run 102 is newer than the live run 101` plus both sides -- so "is this thing actually running?" is answerable from any deploy log.

### The rollback escape hatch

**The ordinary rollback is unaffected and needs nothing.** `git revert` and push produces a *new* commit in a *newer* run, so it's newer by this check's reckoning and publishes normally. Same for pinning `originBucketNameOverride` (Method 2 in BUILD-AND-DEPLOY.md), which the check skips outright since the metadata file isn't what gets published in that case.

The only rollback it stops is re-publishing an *older* build **as-is** -- re-running an earlier "Build and deploy" run, or dispatching at an older ref. `originBucketNameOverride` is the wrong lever for that: it's sticky config that has to be remembered and unset, which is a bad thing to hand someone under pressure. So it gets a companion:

- A `workflow_dispatch` input on "Build and deploy": **"Publish even if a newer deploy already published"**. Check the box, run the workflow, done.
- `ALLOW_OUT_OF_ORDER_PUBLISH=true` in the deploy environment, for anything not going through that form.

Both are named in the failure message, so nobody has to remember either. GitHub renders an unchecked boolean input as the *string* `"false"`, so presence isn't enough -- there's a test for that specifically.

### How I verified the guard actually fires

A guard that never fires is worse than none, so this got three passes, not one.

**1. Direct execution against a staged regression.** Fake `pulumi`/`aws` on `PATH`, a live bucket at run `35266996109`, an incoming build at run `35266933458` (the two run ids from the 2026-09-17 collision in `await-in-progress.js`'s header):

| Scenario | Exit | Output |
| --- | --- | --- |
| production, older over newer | **1** | the `BACKWARDS` block above + `::error::` |
| production, newer over older | 0 | `OK -- run 35266996109 is newer than the live run 35266933458` |
| `www-testing`, same regression | 0 | `::warning::... Not failing the deploy` |
| dev stack `christian` | 0 | `no shared publish ordering to protect. Skipping.` |
| no `pulumi` on PATH at all | **2** | `Refusing to guess: a skipped check reads exactly like a passing one.` |

**2. Mutation testing**, because passing tests only prove the tests pass. I broke the guard three ways and confirmed the suite catches each:

| Mutation | Result |
| --- | --- |
| `decidePublish` always returns `regression: false` | **6 failures**, incl. the end-to-end exit-1 test |
| `publish_run_id()` always returns `null` (ordering key never recorded) | **2 failures** |
| `MODES_BY_STACK` keyed on `"production"` instead of `"www-production"` | **10 failures** |

That third one is the realistic bug -- it's the mistake already sitting in `sync-and-test-bucket.sh` -- and it's the one that would produce a guard that runs on every deploy and never fires.

**3. The suite itself**: 33 new tests in `scripts/check-publish-ordering.test.js`, picked up by `run-unit-tests.sh` and so by `make test`, in three layers:

- `decidePublish()` as a pure function: newer/older/equal run ids, same commit, same bucket, missing run id, timestamp skew.
- `main()` with injected deps: stack scoping, both escape hatches, the pin, and every way live state can be unavailable (no output, unreadable bucket, non-JSON body, pre-rollout metadata with no run id).
- One **end-to-end subprocess run of the real script** asserting `exit 1` and the message, since layers 1 and 2 both stub out the part that does the firing. Modelled on `scripts/deploy-bucket-naming.test.js` -- no network, no AWS, no Pulumi backend.

Plus four tests on `render_origin_bucket_metadata()` itself, including one asserting the document stays parseable when `GITHUB_RUN_ID` is garbage (the values are interpolated unquoted so consumers read numbers, so a non-numeric value would emit JSON that breaks the ordering check, `list-recent-buckets.sh`, and the cleanup retention window together).

`infrastructure/` has no test tooling that fits this -- its only suite is `test-support-form`, a `tsc`-then-`node --test` pass over `support-form/*.test.ts` gated by `check-test-suites-compiled.js`, and the guard has no business being a Lambda-adjacent TypeScript module. It's plain Node in `scripts/`, matching `await-in-progress.js`.

**`make lint`**: clean (1794 files parsed, 0 errors; prettier clean). **`./scripts/run-unit-tests.sh`**: 100/100 (67 existing + 33 new). `make test`'s `test-programs` half needs Hugo and a network, so it wasn't run here.

### What this doesn't cover

- **A newer run whose `pulumi up` is still in flight.** Its outputs aren't written yet, so this reads the pre-race state and allows. That case is already handled loudly by the Pulumi stack lock (`[409] Conflict`). This covers the case the lock can't: a newer run that has already finished.
- **`pulumi up` typed by hand**, or `deploy-dev-stack.sh` (dev stacks are out of scope anyway). The guard sits in `run-pulumi.sh` and `laptop-deploy.sh`.
- Worth noting `laptop-deploy.sh` has an unconditional `exit` partway through today, so its deploy path is dead code. The call is placed where it belongs regardless; not touching that `exit` here.

### Notes for reviewers

- **Interaction with #21789.** That PR retries `pulumi up` on `[409] Conflict`. When it lands, this call belongs **inside** the retry loop, not above it -- the run that won the conflict is exactly the newer run this exists to avoid overwriting, so a single pre-loop check would be stale by the time the retry succeeds. It's cheap and idempotent, so per-attempt costs nothing. There's a comment at the call site saying so; whoever resolves the merge should read it.
- **A `concurrency:` group on `build-and-deploy.yml`** would eliminate out-of-order publishes by construction rather than probabilistically, and is the better long-term fix. It was considered and deferred as a larger change. This stands on its own either way: if the concurrency group lands, this becomes a cheap assertion that it's working, which is worth keeping -- the turnstile was also supposed to make this impossible.
- **Follow-up:** `sync-and-test-bucket.sh` compares `pulumi stack --show-name` against `"production"` while the production stack is named `www-production`, so its bucket-tagging branch is dead. Out of scope here; happy to file it.

### Related issues

Came out of the #docs-ops investigation into the six build-and-deploy failures on 2026-09-21. Related: #21800 (merged), #21789 (open).

https://claude.ai/code/session_01TTAYyob9EMtK1erw9DdChV

---
_Generated by [Claude Code](https://claude.ai/code/session_01TTAYyob9EMtK1erw9DdChV)_


CamSoper:
@claude #new-review

CamSoper:
@claude #update-review

Fixed F1 in 77862379df: the comment now says CloudFront config changes "can take several minutes to propagate", and the same edit is applied to the figure it cites at `BUILD-AND-DEPLOY.md` L1571.

PR #21804, opened 2026-09-22
Open the v3 style block to model-found nits

CamSoper:
### Proposed changes

The v3 author card's `#### Style suggestions` block carried Vale's output and nothing else, so a nit the review found itself had no author-facing home. Everything else on that card blocks merge: 🚨 overshoots on a one-character fix, ❓ is the composer's deterministic `unverifiable` lane, and a hand-written `[style]` bullet desynced the brief's Vale-derived count.

So those findings ended up on the reviewer's brief. [#21787](https://github.com/pulumi/docs/pull/21787) F3 is the clean example -- *"Typo: `i. e.` has a stray space ... Trivial fix for the author"*, shipped on the card headed "not for the author". It made a human relay a one-character fix, and asked them to rubber-stamp a ⚠️ row with no judgment in it.

This opens the block:

- **`[nit]` bullets** -- same shape and non-blocking treatment as `[style]`, for a fix that's mechanical, whose actor is the author, and that leaves a reviewer nothing to decide. Qualifying test and shape in `output-format.md` §Nits; `ci.md` step 10 routes the editorial pass to it.
- **`style-advisory-provenance`** matches every `[style]` bullet against `.vale-findings.json`, and rejects `[nit]` outside v3. Opening the block recreated one tier down the bypass `style-blocker-provenance` exists to close: an unverified finding rendered as `[style]` reads as linter output, and the block gets quoted out of context often enough for that to matter.
- **The counts follow the card.** The brief's rubber-stamp **Style** line and `style_suggestions_count` are re-derived from the bullets actually published, in both the initial and re-entrant lanes, instead of being pinned to Vale's pre-editorial count. (In `apply-update.py` the count moved out of the `prior`-carry loop into `assemble_evidence`, so every caller gets a live number, not just `main()`.)
- **The block always composes on v3** -- the model needs a stable anchor to append under, and asking it to author the H4 and caption verbatim is how you get a malformed block. `build-evidence.py` drops the whole block again when nothing landed, so an empty one never reaches a published card.

Also fixes **`style-render-mode`**, found while working here: it scoped its scan to a `⚠️ Low-confidence` section the v3 author card doesn't have, making it a silent no-op across the whole v3 surface while the rules table advertised it as running there.

Validator schema v23 → v24. The v2 golden fixture is byte-identical -- the v3 fixtures were regenerated and their diff is the caption and the Style line's pluralization.

**Testing.** `make test-review-pipeline` passes; `scripts/lint/lint-markdown.js` reports 0 errors across 1794 files. 13 new tests in `test_style_advisory_provenance.py` (both provenance directions, the shared bullet walker, ✏️-annotator interop, the `style-render-mode` regression) and 6 in `test_compose_v3.py` (the always-composed anchor, the empty-block drop, the count refresh in both lanes). Also ran the real CI path end to end -- compose → `validate-pinned check` → `build-evidence` → `validate-evidence` -- confirming a `[nit]` passes clean and a mis-tagged `[style]` is caught with an actionable hint. Full `make lint` could not run in this environment (no Hugo binary); nothing here touches Hugo content.

N/A

### Related issues (optional)

Follows from the review on [#21787](https://github.com/pulumi/docs/pull/21787).

---

https://claude.ai/code/session_01S9DNvjWrQqcjc93tc6h9rp

---
_Generated by [Claude Code](https://claude.ai/code/session_01S9DNvjWrQqcjc93tc6h9rp)_


CamSoper:
@claude #update-review

Fixed in f7705055:

- F1: style-render-mode now bounds its scan at the next heading or card furniture (`##`/`###`/`####`, the 📎 line, `<!--` markers), so a `<details>` below the style block no longer reads as a collapsed style block. `##### <path>` groups stay inside the bound. Regression test added (`test_style_render_mode_ignores_details_below_the_block`), and it fails against the unbounded version.
- The `[nit]` on test_compose_v3.py L584: dropped the extra blank line, two now.

---

CamSoper:
@claude #new-review

PR #21805, opened 2026-09-22
Let any review team run /deploy-staging, and test the gate

CamSoper:
### Proposed changes

`/deploy-staging` accepted active members of `teams.tools` only. That team has exactly one member, so Sentinel G4 -- a gate with no waiver -- depended on one person being available. This widens it to active membership in **any** team under `teams:` in `.github/review-routing.yml`, which is the same set and the same any-team rule G3 already applies to approvals. Two gates asking "does this repo trust you with this PR" shouldn't answer differently.

Widening costs nothing in blast radius: the fork guard already limits the command to branches in this repo, so everyone it newly admits could push the same code here anyway.

**Refusals now distinguish two cases.** A 404 from every lookup means "you're not on any of these teams" and names them; anything else means the check itself couldn't be made and says so. Both refuse -- fail-closed is unchanged -- but they call for different next actions, and the old code reported an ordinary non-member as `lookup-failed`, telling someone the tooling was broken when it was working correctly. Each lookup is now logged with its result, so a refusal is readable from the Actions tab.

The routing config is still read through the contents API with no `?ref=`, so it resolves against the default branch and a PR can't widen `teams:` in its own diff to admit itself. That was true before and incidental; it's the whole gate now, so it's commented.

G4's remedy text and `scripts/review-v3/README.md` said tools-team. They now say what the gate does.

**Tests.** `scripts/review-v3/test_staging_gate.py`, picked up by `make test-review-pipeline`. They extract the step's `run:` block from the workflow YAML by step name and execute it against a fake `gh`, rather than testing a transcribed copy that would keep passing after the workflow changed underneath it. Coverage: every team in the routing config clears the gate (parametrized from the config, so a new team tests itself), a non-member is refused and told who to ask, `pending` isn't membership, a lookup failure gets the cautious wording, a partial outage neither strands a member of a team that did answer nor tells an unverifiable actor they're not a member, and the fork guard still refuses a review-team member.

Verified non-vacuous against five mutations -- tools-only again, lookup failure treated as permission, `pending` accepted, fork guard removed, per-team logging silenced -- each fails at least one test.

The first of those tests earned its keep immediately: the team list was originally read with a multi-line `python3 -c` whose lines began at column 0, which terminates a YAML block scalar. The workflow file wouldn't have parsed at all.

Worth stating plainly: this workflow has ~990 runs and every one short-circuited on the `if:`. Nobody has ever typed `/deploy-staging`, so none of this code had executed before now. Its own header records the last thing that lay in wait behind that fact -- a missing `pull-requests: write`, found only because the same mistake surfaced in #21642.

**Testing:** full `make test-review-pipeline` suite green. `make lint` doesn't reach these files -- `.github`, `*.md` and `scripts` are all in `.prettierignore`, and `lint-markdown.js` only inspects `content/`.

**Note for review:** #21802 edits the same G4 message string, so whichever of these lands second takes a small conflict there. Nothing else overlaps.

N/A

### Related issues (optional)

N/A

https://claude.ai/code/session_019WYGqWVdGh5QFbNJFBeBXk

---
_Generated by [Claude Code](https://claude.ai/code/session_019WYGqWVdGh5QFbNJFBeBXk)_


CamSoper:
@claude #update-review

Fixed F1 in fd353de. Good catch -- the G4 `Gate(...)` call is 430 chars, so 370 of that 800-char window was the `else:` branch's comments, and I confirmed the old slice fails on an unrelated "tools-team" comment dropped in there.

Went a step past the suggested `))` slice, since a `))` inside the message would cut it short. `g4_red_message()` now parses sentinel.py with `ast`, finds the `Gate("G4 infra-evidence", "red", ...)` call itself, and asserts there's exactly one. Checked both ways: unrelated text next to the call passes, and regressing the message itself to "a tools-team member" still fails.

One correction for the brief: it says the new test file has 13 tests. It has 17 at this head -- the second commit added four (exact refusal formatting, a token that can't see the teams, one invisible team next to a visible one, and an empty `PULUMI_BOT_TOKEN`).

---

CamSoper:
@claude #new-review

